PR #138, opened 2012-09-07
SPR-9275 Cache Abstraction Exception Handling

billkoch:
Introduced FailSafeCache.

FailSafeCache is intended to wrap existing cache implementations and
gracefully handle any Exceptions thrown by the native cache.  The
AbstractCacheManager is responsible for wrapping the native cache(s) IF
the gracefullyHandleExceptions property is _true_.  Otherwise, the
behavior of the AbstractCacheManager remains unchanged.

Issue: SPR-9275


snicoll:
Wow, I wasn't even aware of that PR. You're probably aware that SPR-9275 has been implemented in 4.1 in a slightly different approach. Thanks for the PR anyway!

PR #142, opened 2012-09-10
Add JUnit Assumptions to core test sources

rwinch:
As per discussion in this weeks stand up I have added a utility class that allows assuming JDK version in order for the test to run. Also see the [junit-assumptions-demo branch](https://github.com/rwinch/spring-framework/tree/junit-assumptions-demo) to see this in action (since there are no use cases in the actual build now).

Previously it was cumbersome for tests to assert if the current JDK was at
least JDK 1.7.

There is now a utility class for JUnit assumptions that provides a method
that allows tests to assume the current JDK is at least JDK 1.7. If the
assumption is not met, a AssumptionViolatedException is thrown to indicate
the test should be skipped.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,6 +51,10 @@ configure(allprojects) {
 configure(subprojects) { subproject ->
     apply from: "${rootProject.projectDir}/publish-maven.gradle"
 
+    if(project.name != 'spring-core') {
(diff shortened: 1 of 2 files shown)


sbrannen:
@rwinch, since this PR is over 6 years old, I am going to close it.

However, if you feel that we still need to do something in this regard, please feel free to speak up.

PR #162, opened 2012-10-09
SPR-9335: catch exceptions explicitly where possible

dsyer:
This change restores the old behaviour in Spring 3.0, as near
as I can tell, without breaking the Groovy aspect changes
introduced in 3.1.

Since no-one has come up with a test case for all the catch blocks
and there is no test project for the WAS case (I tried) we have to
rely on the users in the JIRA thread to test this change.


rstoyanchev:
Closing as the related issue has been fixed.

PR #215, opened 2013-01-13
ShallowEtagHeaderFilter can now handle the If-Match HTTP Header for cond...

eugenp:
...itional GET requests

Few notes: 
- this is also being tracked on JIRA: [SPR-10164](https://jira.springsource.org/browse/SPR-10164)
- the changes (can only) deal with read operations (GET); the **shallow implementation** cannot handle conditional PUT operations, since it would have to determine IF the filter chain can be executed and STOP it in case it should not be - which is not possible 
- the **if-else nesting** in `updateResponse` is a bit to deep (3 levels) - I can extract out the logic handling the `If-None-Match` and the logic handling `If-Match`
- the original implementation only handled `If-None-Match` so I had to rename a few variables to accomodate the new code path for `If-Match`

Thanks.


nolines:
Hi, collegues. Any news about this issue?

cebbens:
@eugenp Any update on this?

eugenp:
Hey @cebbens - I haven't looked into this one for a while now, so no updated. If I do run into this usecase again, or if you have - I'm happy to jump back in.

PR #222, opened 2013-01-25
SPR-10217 Implement JUnit 4 Support using Rules

marschall:
Currently JUnit 4 support is provided by SpringJUnit4ClassRunner which
is a custom BlockJUnit4ClassRunner. There is no support for using other
runners like Theories or Parameterized or 3rd party runners like
MockitoJUnitRunner. A runner based approach does not seem to offer much
promise as runners are not composable, a custom Spring version of every
runner has to be developed and maintained.

With JUnit 4.9+ the preferred way to implement such behavior is to use
rules. Unlike runners there can be several ones of them and they can be
composed. In theory TestExecutionListener could be deprecated and be
replaced with standard JUnit rules but this seems to be a bit on the
drastic side.

This proposed implementation is using both a class rule and a method
rule. The class rule creates the TestContextManager, runs all the class
level callbacks and class level checks. The method rule runs all the
instance level callbacks and method level checks. I did not see a way
to implement the current functionality offered by
SpringJUnit4ClassRunner using only one rule.
Using two rules has the advantage that the implementation is cleaner
because it better separates the concerns. However it has the
disadvantage that it's harder to set up because both a method rule and
a class rule are needed. This also increases the potential for
misconfiguration.

The method rule has to be a MethodRule instead of a TestRule because
only the former has access to the test object with we need to perform
injection. This interface used to be deprecated once but doesn't seem
to be anymore. This creates a certain risk that it will be deprectated
again and potentially be remvoed in the future. An additional drawback
is that MethodRule unlike TestRule can only be defined in fields and
not methods. This is an unfortunate consequence of the implementation
of org.junit.runners.model.TestClass. As JUnit does not do
Field#setAccessible(true) this means that tests will have to be defined
in public fields.

Another minor issue is that tests not run because of IfProfileValue
will still show up in the Eclipse test tree, just blank.

In conclusion while the given implementation has some downsides I don't
see any other possible implementations given the current state of
affairs in JUnit.
- Add SpringJUnitClassRule for all the class level processing
- Add SpringJUnitMethodRule for all the method level processing
- Add tests for the rules

SPR-10217

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/junit4/SpringJUnitClassRule.java
+++ b/spring-test/src/main/java/org/springframework/test/context/junit4/SpringJUnitClassRule.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2004-2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.test.context.junit4;
+
+import static org.junit.Assume.assumeTrue;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.springframework.test.annotation.ProfileValueUtils;
+import org.springframework.test.context.TestContextManager;
+
+
+/**
+ * <p>
+ * {@code SpringJUnitClassRule} is a custom {@link TestRule} which provides
+ * functionality of the <em>Spring TestContext Framework</em> to standard
+ * JUnit 4.9+ tests by means of the {@link TestContextManager} and associated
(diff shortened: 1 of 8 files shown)


erizzo:
The comments in SpringJUnitMethodRule mention that ExpectedException is not supported; is that indeed true? If so, can you elaborate a bit?

sbrannen:
The Javadoc for `SpringJUnitMethodRule` with regard to `ExpectedException` is admittedly misleading at a glance.

If you take a look at the imports in `SpringJUnitMethodRule`, you will notice that the class referenced in the Javadoc is `org.springframework.test.annotation.ExpectedException` (i.e., the `@ExpectedException` annotation that was supported in earlier versions of the Spring TestContext Framework).

So fear not: JUnit's `org.junit.rules.ExpectedException` (i.e., the `ExpectedException` Rule) should be fully supported when used in conjunction with `SpringJUnitMethodRule`.

sbrannen:
For what it's worth, `@ExpectedException` was permanently removed from the `spring-test` module in Spring Framework 4.0 (see [SPR-10499](https://jira.spring.io/browse/SPR-10499)).

sbrannen:
@marschall and @eeichinger, thanks so much for all of your ideas and collaboration on this!

I've just committed support for this in Spring Framework 4.2 RC1 here: https://github.com/spring-projects/spring-framework/commit/d1b1c4f888b5ab5c60cf4beb87f577143bdbebe7.

Cheers,

Sam

sbrannen:
Though... perhaps I should attribute the `Assume` idea in `ProfileValueChecker` to @eeichinger instead of @marschall. ;)

PR #227, opened 2013-02-01
SPR-10250 - BufferedImageHttpMessageConverter redefine content type of HttpOutputMessage

kpankowski:
More details here: https://jira.springsource.org/browse/SPR-10250


rstoyanchev:
Setting the Content-Type on ResponseEntity should work in more recent versions.

PR #232, opened 2013-02-13
Add media types to possible handler method params

acogoluegnes:
The requested media types are now valid arguments for Spring MVC
controller methods. They are computed by the
ContentNegotiationManager (so using the HTTP Accept header).
Controllers can ask for the requested media types by adding
a Collection<MediaType>\- or a List<MediaType>-typed method argument.

Issue: SPR-9980

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java
@@ -157,6 +157,10 @@ import java.util.concurrent.Callable;
  * (Servlet-only, {@literal @MVC 3.1-only})
  * for preparing a URL relative to the current request's host, port, scheme,
  * context path, and the literal part of the servlet mapping.
+ * <li>{@link java.util.Collection} or {@link java.util.List} of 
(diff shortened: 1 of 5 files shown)


ahsanfz:
While this is an old issue, I think it is a genuine usecase and a viable alternative to the Variant mechanism (request.getVariant() call )  that can be found in Jersey implementations.

I am migrating an application from Jersey to Spring MVC and I stumbled upon this. For the language part in the Variant implementation, the LocaleResolver was a good alternative. 

For the  other 2 dimensions of MediaType and encoding that is found in the Jersey implementation, this seems a good fit. 

In my case I want to write to a StreamingResponseBody and the old implementation was using Variants to determine encoding. The encoding was specified the the accept header (MediaType/MimeType) and the parameter portions contains the charset. Using the resolver above I can get the encoding here where I need to know if before so that I can specify the correct encoding that the client specified to write with in my OutputStream/Writer.

In my case I am using text/plain and the client can ask me various encoding in order of his preference and I can accomadate it in my Response. 

Does that make sense for you guys. 

I think this resolver belongs in Spring by default.

PR #266, opened 2013-03-24
[SPR-10412] Support generics in JmsTemplate

marschall:
The `#convertAndSend` and `#receiveAndConvert` methods and the
`MessageConverter` interface currently are not type safe. Everywhere
`Object` is passed around casts have to be made.

This pull request tries to address the issue by:
- split `MessageConverter` into two interfaces, `MessageDecoder` and
  `MessageEncoder` so that Java 8 lambdas can be used
- add generics to `MessageConverter`, `MessageDecoder` and
  `MessageEncoder`
- for every `#convertAndSend` method add a new one that has a
  `MessageEncoder` as a second argument
- for every `#receiveAndConvert` method add a new one that
  has a `MessageDecoder` as a second argument
- add generics to the new `#convertAndSend` and
  `#receiveAndConvert` methods

MessageConverter and MessageDecoder are not be best names but the best
ones I could come up with.

Maybe `MessageConverter` and `MessageDecoder` should be annotated with
`java.lang.FunctionalInterface` but this would have to be done over the
entire code base eg. `org.springframework.jdbc.core.RowMapper`.

As usual, I'm not one hundred percent certain when it comes to wild
cards.

And tests are missing.

Issue: SPR-10412
https://jira.springsource.org/browse/SPR-10412

### DIFF
--- a/spring-jms/src/main/java/org/springframework/jms/core/JmsOperations.java
+++ b/spring-jms/src/main/java/org/springframework/jms/core/JmsOperations.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


snicoll:
Thanks for the PR and sorry for the time it took to get back to you. We're not keen to add a bunch of overloaded methods like this to `JmsTemplate` itself. `JmsTemplate` is usually used internally with a more defined API where such rich type is exposed at that level.

PR #283, opened 2013-05-11
Added test case to show broken bean lookup with type prediction.

odrotbohm:
Added test case to Spr8954Tests that explicitly prevents bean creation on type lookup. This results in beans not being found for a "beans by type" call.

Refactored existing test cases to use assert methods correctly.

Issues: SPR-10517, SPR-8954, SPR-9177

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,6 +63,9 @@ configure(allprojects) { project ->
 	test {
 		systemProperty("java.awt.headless", "true")
 		systemProperty("testGroups", project.properties.get("testGroups"))
+		scanForTestClasses = false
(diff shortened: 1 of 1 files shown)


MrKolt01:
Hi! Any ideas why there are no commits here?

sbrannen:
No, I have no idea what happened to the commits in the PR, and the linked merge commit (9ff9d45283b5aad6eb8b97775a406a4fbdb74782) seems to be the wrong commit as well.

🤷‍♂️

sbrannen:
It appears to have been related to issue #13594 and PR #16.

PR #285, opened 2013-05-15
Add method that execute batch update by named param

teddyjoo:
Prior to this commit, BatchSqlUpdate class didn't override 
updateByNamedParam() method but overrided update() method only.

Now BatchSqlUpdate class support both method.
This change would make BatchSqlUpdate class more consistent
because parent SqlUpdate class supports both method.

To add new method, existing codes from SqlUpdate class and 
BatchSqlUpdate class were reused.

Issue: SPR-10435


snicoll:
Unfortunately the CLA has not been signed so there's nothing we can do. @teddyjoo if you're still around and you're willing to sign the CLA, please add a comment here and we can reopen.

PR #293, opened 2013-05-26
Add support for MethodExecutor in FunctionReference

obecker:
This commit enhances FunctionReference to also support MethodExecutor
instances for evaluation (besides static java.lang.reflect.Method
instances). This enables dynamic SpEL function definitions where the
function to be executed is not known at compile time.

Additionally this commit
- removes all mentions of lambda function definitions in comments since
  they are not implemented
- fixes the type name in the message of the thrown
  SpelEvaluationException in case the value of the function has the wrong
  type and cannot be executed

Issue SPR-10589


rstoyanchev:
Closing as outdated, along with related issue. The Javadoc issue seems to have been cleaned up since.

PR #301, opened 2013-06-24
Support WritableResource property values

philwebb:
Update DefaultResourceLoader to always attempt FileResource resolution
before URLResource, allowing casting of the result to WritableResource.

Also update ResourceEditorRegistrar to register a property editor for
WritableResources.

Issue:


snicoll:
Sorry this was overlooked for so long @philwebb. This got superseded in the meantime as of `5.0.2`.

PR #309, opened 2013-07-04
Add scaleToUse to org.springframework.jdbc.core.StatementCreatorUtils.setParameterValueInternal.

GeorgeVan:


snicoll:
Unfortunately the CLA isn't signed and we can't proceed with this change. We can reconsider if the CLA is signed.

PR #312, opened 2013-07-09
Request streaming for Apache HttpClient

poutsma:
- Added org.springframework.http.StreamingHttpOutputMessage, which
  allows for a settable request body (as opposed to an output stream).
- Added http.client.HttpComponentsStreamingClientHttpRequest, which
  implements the above mentioned interface, mapping setBody() to a
  setEntity() call on the Apache HttpClient HttpEntityEnclosingRequest.
- Added a 'bufferRequestBody' property to the
  HttpComponentsClientHttpRequestFactory. When this property is set to
  false (default is true), we return a
  HttpComponentsStreamingClientHttpRequest instead of a (request
  buffering) HttpComponentsClientHttpRequest.

Issue: SPR-10728

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/HttpComponentsStreamingClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/HttpComponentsStreamingClientHttpRequest.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright 2002-2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


spring-web/src/main/java/org/springframework/http/client/HttpComponentsStreamingClientHttpRequest.java
```diff
@@ -0,0 +1,168 @@
...
+	@Override
+	public void setBody(Body body) {
+		assertNotExecuted();
+		this.body = body;
+	}
+
+	@Override
+	protected OutputStream getBodyInternal(HttpHeaders headers) throws IOException {
+		throw new UnsupportedOperationException(
+				"getBody not supported when bufferRequestBody is false");
```
dreambrother: +1
I've tested FormHttpMessageConverter and it is broken for HttpComponentsClientHttpRequestFactory with setBufferRequestBody set to false

bclozel: Could you create an issue on https://jira.spring.io and explain the problem there?
Thanks!

dreambrother: It's already created
https://jira.spring.io/browse/SPR-11199

bclozel: ok thanks!

PR #355, opened 2013-09-17
Beanbuilder spring4

jeffscottbrown:
I am working with Juergen to get the bean builder code to a state such that it can be included in Spring 4.  The code represented in this pull request has been tested with Grails 2.4 snapshots and so far everything works well (more testing is pending). If we can get something into the Spring code base so that a build snapshot can be published that will greatly simplify work that we need to do to verify and finalize this code.  As it is all of the testing is being done in my local development environment where I have built this latest snapshot code and installed it into my local maven repo for consumption by Grails snapshots and related plugins.  Having this snapshot published to a public repo will allow us to do more thorough testing in our CI environment.  Please let me know whatever I can do to help move forward with this.


jhoeller:
Closing this old PR since it got superseded by our reworked `GroovyBeanDefinitionReader` etc in #11783.

PR #361, opened 2013-09-25
Add "ignoreResourceNotFound" attribute to the @PropertySource

vvictorlee:
Currently, a file is required to exist if it is added to the @PropertySource annotation. If the application already 
has the properties it needs, it should be possible to launch it without the associated file being present:

```
@PropertySource(value = "classpath:/notexists.properties", ignoreResourceNotFound = true)
```

Adding new attribute 'ignoreResourceNotFound' to @PropertySource,
user can decide to ignore the not-existing resource.
- default is false
- The ignoreResourceNotFound is supported for XML-based application
  context by context:property-placeholder

ConfigurationClassParser is changed to load properties regarding option 'ignoreResourceNotFound', 
and PropertySource is changed to have new member 'boolean ignoreResourceNotFound'. 
Thank you.

issue: SPR-10932


snicoll:
@vvictorlee we've added such a feature in the meantime, sorry this PR was overlooked.

PR #368, opened 2013-09-30
Issue: SPR-10919

thomasdarimont:
Added support to ConfigurationClassParser to deal with @Bean annotated Java 8 default Methods. Introduced getInterfaces() Method to SourceClass to be able to consider @Bean annotated Bean factory methods on interfaces. Added additional processing step to doProcessConfigurationClass(…) to process information in interfaces. Introduced getRawCandidateMethods(…) to ConstructorResolver that is able to consider default Methods in addition to the previous raw candidates if executed on Java 8 or later. Added test case for feature.

All existing and new tests pass.


snicoll:
This PR is now outdated by #15547

PR #467, opened 2014-02-24
Fix PathVariable encoding if request charset is not set

snicoll:
This commit simply adds a `@PathVariable` decoding test to reproduce a
scenario exposed in `SPR-11474`

Issue: SPR-11474


snicoll:
Closing a PR that I created a couple of months after joining 😬 - Kidding aside, this got superseded by #16543

PR #481, opened 2014-03-05
Add merge support to nested collection properties

cbaldwin74:
Prior to this change, the merge attribute on nested collections was not
handled properly which would result in the typical replacement
behaviour. This is because during the merging of the parent and child
beans the parent bean would have nested BeanDefinition objects while the
child would have a property with the nested name
(beanA.beanB.propertyName). When it came time to look for matching the
property names it would be looking for the nested name which would not
exist in the parent.

The implementation of MutablePropertyValues.addProperty(PropertyValue)
will now try to find a chain of nested BeanDefinition objects that match
the nested property name and walk down the chain of beans and find the
property and then merge if applicable.

A JUnit test has been added to check for the new behaviour.

Issue: SPR-9552

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/MutablePropertyValues.java
+++ b/spring-beans/src/main/java/org/springframework/beans/MutablePropertyValues.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2013 the original author or authors.
+ * Copyright 2002-2014 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@cbaldwin74 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=481)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=481) to manually synchronize the status of this Pull Request.

snicoll:
Closing as the [related issue](https://github.com/spring-projects/spring-framework/issues/14186#issuecomment-453395223) has been declined

PR #507, opened 2014-04-01
Fix return zero size array with empty String parameter in StringUtils.delimiitedListToStringArray

youngjee:
Prior to this commit, zero size String array returns if StringUtils.delimitedListToStringArray is called with parameter empty String ("")

I think, after call StringUtils.delimitedListToStringArray("", ",", null);

it should return one size String array -> [""]

but it returns zero size String array.

StringUtils.delimitedListToStringArray(null, ",", null) & StringUtils.delimitedListToStringArray("", ",", null)

should returns differently. However there are same results.

There's no logic for empty String(""), but just is for null parameter.

I added if statement for "" parameter. The result will be one size String array.

And also in SPR-11126, the issue will be fine.

Issue: SPR-11126

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -1045,6 +1045,9 @@ public abstract class StringUtils {
 		if (str == null) {
 			return new String[0];
 		}
+		else if ("".equals(str)) {
+			return new String[] {""};
+		}
 		if (delimiter == null) {
 			return new String[] {str};
 		}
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@youngjee Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=507)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=507) to manually synchronize the status of this Pull Request.

PR #561, opened 2014-06-13
Added a cache to transaction manager lookup to alleiviate locking issue...

serioussam:
Added a cache to tranbsaction manager lookup to alleiviate locking issues when we have more than one transaction manager present in application.


JanStureNielsen:
:D @FancyCoder0 -- I suggest you review @snicoll comment just prior; his algorithm solved the problem 4 years prior...

PR #566, opened 2014-06-24
In case that reqeust url includes "=" char in front of first semicolon,

MoonJongMin:
In case that reqeust url includes "=" char in front of first semicolon,
PathVariable do not work properly.
Problem is that @PathVariable value includes matrixVariable String.

For example, if Spring MVC Controller method has @RequestMapping(value =
/bar/{foo}", method = RequestMethod.GET) and request url is
http://localhost:8080/context/bar/abc;key=xyz, we expect @PathVariable
foo == "abc" and @MatrixVariable key == "xyz". Actually this works well.
But when request url is
http://localhost:8080/context/bar/name=abc;key=xyz,
despite expectations that @PathVariable foo == "name=abc" and
@MatrixVariable key == "xyz", @PathVariable foo == "name=abc;key=xyz".

I modified so that @PathVariable foo may have "name=abc".

Issue : SPR-11897

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/RequestMappingInfoHandlerMapping.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/RequestMappingInfoHandlerMapping.java
@@ -143,7 +143,7 @@ public abstract class RequestMappingInfoHandlerMapping extends AbstractHandlerMe
 		for (Entry<String, String> uriVar : uriVariables.entrySet()) {
 			String uriVarValue = uriVar.getValue();
 
-			int equalsIndex = uriVarValue.indexOf('=');
+			int equalsIndex = uriVarValue.lastIndexOf('=');
 			if (equalsIndex == -1) {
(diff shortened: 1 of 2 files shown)


FancyCoder0:
Dear Jongmin Moon,

We are working on identifying redundant development and duplicated pull requests. We have found a pull request: https://github.com/spring-projects/spring-framework/pull/569 which might be potentially duplicate to this one. We would really appreciate if you could help us to validate and give us feedback. 

Sincerely,
Luyao

rstoyanchev:
I've processed the fix under #569 instead which is better for backwards compatibility.

pivotal-issuemaster:
@MoonJongMin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=566)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=566) to manually synchronize the status of this Pull Request.

PR #569, opened 2014-06-27
Fixed [SPR-11897]

boazj:
Please refer to [SPR-11897](https://jira.spring.io/browse/SPR-11897).

The gist of the bug is that in the case of posting "a=a;b=c" as a path variable when shouldRemoveSemicolonContent is false, the matrix parameters are evaluated as [a->a, b->c] and the path variable remains "a=a;b=c" whilst in the case of "aa;b=c" the matrix parameters are evaluated as [b->c] and the path variable is passed as "aa" thus leading to inconsistent and undocumented behavior when a '=' sign is present in the pre-matrix (before the first semicolon) part of the path.
This fix assumes that parsing the pre-matrix part of the path was a desired feature (due to lack of official comment on SPR-11897) but that the returned path variable should omit what spring parsed as matrix variables.

I've added several tests to validate this and fix other tests that expected it (for example: in case of having a path variable that it's first character is a semicolon, the test expected the entire value and not an empty string - I changed it to an empty string as all these values were parsed as matrix variables)

Commit message:
Fixed SPR-11897 in a way that does not harm the feature (?) of parsing key=value in
the pre-matrix part of the path as though it was a matrix variable
- added test to validate that scenario

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/RequestMappingInfoHandlerMapping.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/RequestMappingInfoHandlerMapping.java
@@ -151,12 +151,14 @@ public abstract class RequestMappingInfoHandlerMapping extends AbstractHandlerMe
 			String matrixVariables;
 
 			int semicolonIndex = uriVarValue.indexOf(';');
-			if ((semicolonIndex == -1) || (semicolonIndex == 0) || (equalsIndex < semicolonIndex)) {
-				matrixVariables = uriVarValue;
-			}
-			else {
-				matrixVariables = uriVarValue.substring(semicolonIndex + 1);
+			if (semicolonIndex != -1) {
 				uriVariables.put(uriVar.getKey(), uriVarValue.substring(0, semicolonIndex));
+			}
+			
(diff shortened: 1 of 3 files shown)


FancyCoder0:
Dear Boaz Jan,

We are working on identifying redundant development and duplicated pull requests. We have found a pull request: https://github.com/spring-projects/spring-framework/pull/566 which might be potentially duplicate to this one. We would really appreciate if you could help us to validate and give us feedback.

Sincerely,
Luyao

rstoyanchev:
I've applied the suggested fix manually since the original PR was made a long time ago.

pivotal-issuemaster:
@boazj Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=569)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=569) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@boazj Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=569)!

PR #570, opened 2014-06-27
SPR-11875: Performance regression for custom autowireBean calls with many properties

trask:
Please see SPR-11875.

This commit reverts e1d11ec99d40051ecdb60191c49bc587373a031d and uses ConcurrentReferenceHashMap to fix SPR-8956 memory leak instead.

Autowire benchmark https://github.com/trask/spring-autowire-benchmark is back in line with Spring 3.1.4 after this.


snicoll:
Closing as the [related issue](https://github.com/spring-projects/spring-framework/issues/16494) has been fixed.

PR #575, opened 2014-07-01
edit exception list type as Exception instead of Throwable

milgarubanjuk:
I've changed the exception list type from Throwable to Exception. It's more clear when @ExceptionHandler annotation use.

Issue: https://jira.spring.io/browse/SPR-11933


rstoyanchev:
The exception is now also used in WebFlux which does support Throwable.

PR #581, opened 2014-07-05
Adding cacheSharedConsumers configuration for CachingConnectionFactory

cshannon:
There could be use cases to allow more fine grained control in the CachingConnectionFactory so the user can specify what is and isn't cached. Consumers, shared consumes, durables, no durables, etc could potentially be configured separately so that some types are cached and some not cached.  Therefore I split up the configuration of caching of consumers and shared consumers so either type can be configured separately.  Durable/non-durables could also be split up too in another commit.

Issue: SPR-11958

### DIFF
--- a/spring-jms/src/main/java/org/springframework/jms/connection/CachingConnectionFactory.java
+++ b/spring-jms/src/main/java/org/springframework/jms/connection/CachingConnectionFactory.java
@@ -90,6 +90,8 @@ public class CachingConnectionFactory extends SingleConnectionFactory {
 	private boolean cacheProducers = true;
 
 	private boolean cacheConsumers = true;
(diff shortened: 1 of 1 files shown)


snicoll:
Sorry it took so long to get back to you about this. #16574 has an interesting discussion about the use cases. However, we haven't heard of similar interest of the community about this in a long while so I wonder if we should pursue.

If we decide to do so, we'd need you to update the PRs with additional tests.

snicoll:
Thanks for the PR but we consider `CachingConnectionFactory` to be feature complete at this point and we don't intend to extend it any further. Please consider using a fully-featured JMS connection pools for such advanced scenario.

PR #604, opened 2014-07-29
add testNullSafeToStringWithXSSValue testcase to objectUtilsTests

hangsublim:
In the absence of a message string for the error code, this is the message that gets written as the error.
So ObjectUtils class, nullSafeToString method parameter can contains XSS attack script value.
So I added convertXSSValue method and testCase.

When the vaule contains script then the method converting the value safe.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ObjectUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ObjectUtils.java
@@ -562,7 +562,7 @@ public abstract class ObjectUtils {
 			return NULL_STRING;
 		}
 		if (obj instanceof String) {
-			return (String) obj;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@hangsublim Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=604)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=604) to manually synchronize the status of this Pull Request.

snicoll:
@hangsublim thanks for the PR but a generic method such as `nullSafeToString` can't do that kind of transformation as it could be used with trusted input and such change may not be welcome.

PR #614, opened 2014-08-01
MockHttpServletRequest does not handle rfc formatted dates.

jmnarloch:
This change within MockHttpServletRequest implementation allows to
use a formatted date headers. This opens the possibility to prepare
the tests that are as close as possible to the behaviour of real
client/server communication.

Issue: SPR-12056


bclozel:
Fixed in 43e36e2dee
Thanks!

jmnarloch:
Cool thanks!

PR #615, opened 2014-08-02
Registering custom argument resolvers within the ExceptionHandlerExcepti...

jmnarloch:
...onResolver.

By default the ExceptionHandlerExceptionResolver completely ignores
the custom arguments resolvers that were registered within the
web application. This patch registers within
ExceptionHandlerExceptionResolver the same arguments resolvers that
are used by the controllers, allowing to inject parameters into
@ExceptionHandler annotated methods.

Issue: SPR-12058


rstoyanchev:
I've incorporated this slightly differently but effectively the same also taking care of the MVC namespace side. Thanks for the PR!

PR #631, opened 2014-08-21
Add Travis CI configuration

candrews:
Add basic Travis CI configuration.

Travis will test pull request and branches to ensure they compile and pass tests.

Go to https://travis-ci.org/profile/ and flip the switch to "ON" for "spring-projects/spring-framework" then merge this pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: java
+
+sudo: required
+dist: trusty
+
+# required for oraclejdk9 to be available
+group: edge
+
+jdk:
+  - oraclejdk9
+  - oraclejdk8
+
+matrix:
+  - allow_failures:
+    # allow java9 builds to fail until Java 9 is actually released (estimated to be March, 2018)
+    - jdk: oraclejdk9
+
+script:
+ - java -version
+ - ./gradlew check


snicoll:
@candrews looking at my assigned PR. That doesn't seem to work very well on your fork. Is there a way to increase the timeout or something?

.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: groovy
```
sbrannen: Why did you choose `groovy` for the language instead of `java`?

.travis.yml
```diff
@@ -0,0 +1,5 @@
...
+jdk:
+- oraclejdk8
```
sbrannen: I'd also recommend using a more recent version of JDK 8, like we do in the [JUnit 5 build](https://github.com/junit-team/junit5/blob/master/.travis.yml).

.travis.yml
candrews: I'm honestly not sure... this was a long time ago. I've changed it to `java` as I agree that makes more sense.

.travis.yml
candrews: Travis CI is using 1.8.0_65 as is... that doesn't seem so bad. In the interest of saving build time, it seems best not do `sudo apt-get update && sudo apt-get install oracle-java8-installer` as your junit example does. What do you think?

sbrannen: > Travis CI is using 1.8.0_65 as is... that doesn't seem so bad. In the interest of saving build time, it seems best not do sudo apt-get update && sudo apt-get install oracle-java8-installer as your junit example does. What do you think?

If Travis CI is now using JDK 8 Update 65 as the default, then I agree: that should be fine.

For the JUnit 5 build, that wasn't the case when we set up the build: Travis CI used to default to an early JDK 8 release that didn't work for us.

snicoll:
Closing as we’re investigating moving to a different CI system.

PR #647, opened 2014-09-16
Removing problematic creation of unnecessary RequestConfig

marenzo:
The explicit (possibly again) creation of RequestConfig for HttpContext is wrong, since if someone passes default RequestConfig to the HttpClient object (via setDefaultRequestConfig, as seen here: http://hc.apache.org/httpcomponents-client-4.3.x/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html#setDefaultRequestConfig(org.apache.http.client.config.RequestConfig)) it will override it and making the developers to override for nothing the createHttpContext each time, instead of just using the already provided feature of setting default RequestConfig.

Instead of this condition, and making people overriding this class for just re-configuring RequestConfig for each HttpComponentsClientHttpRequestFactory, maybe it should be replaced with other conditions that checks if there's ANY RequestConfig, and if not then to create a new one.

Anyway, currently this condition is wrong and as my point of view it is kind of a bug.


pivotal-issuemaster:
@marenzo Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=647)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=647) to manually synchronize the status of this Pull Request.

rstoyanchev:
This has been since superseded by work done in #17144 and #17184.

PR #653, opened 2014-09-26
SPR-12081: Use a more efficient buffer implementation in ShallowEtagHeaderFilter

candrews:
https://jira.spring.io/browse/SPR-12081

This PR is composed of 2 commits.

The first commit doesn't change the API of ShallowEtagHeaderFilter so it's safe to use in 4.1.x. It offers an improvement over the current state by eliminating most of the copying and all of the array reallocations.

The second commit changes the API of ShallowEtagHeaderFilter so it probably has to wait for 4.2. It eliminates the creation of the whole response byte[].

I'm sure this PR isn't perfect; I look forward to the comment and discussion. For example, I'm not too thrilled about the "instanceof" check I had to do in DigestUtils in order to do the optimized digest, but I couldn't think of a better way.

Thanks in advance!

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/DigestUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/DigestUtils.java
@@ -16,6 +16,8 @@
 
 package org.springframework.util;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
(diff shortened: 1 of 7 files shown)


bclozel:
Hi @candrews 
No need to do anything, I'll push your commits along other enhancements (I think we might be able to reuse this in oher places as well).
I'll let you know when it makes master.

bclozel:
Hi @candrews 
Your contribution is in master now - I've just squashed commits into one and polished a bit. Thanks a lot!

candrews:
Thanks, @bclozel!

PR #671, opened 2014-10-23
Fix First invocation of org.springframework.core.task.AsyncListenableTaskExecutor executes in same thread as caller

MoonJongMin:
With relation to SPR-12358, as Juergen Hoeller and Sébastien Deleuze commented in the Spring JIRA, adding a special delay time in the executor looks enough. We can think that the thread in the AsyncTaskExecutor will call callback method. But if a developer puts pretty much time spending code block between AsyncTaskExecutor's submitListenable and ListenableFuture's addCallback, that will not be enough. It might be the same problem.
So I slightly changed ListenableFutureCallbackRegistry.java in order to run a callback in a new thread when AsyncTask is already finished. At least this will prevent being blocked.

Issue: SPR-12358

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/concurrent/ListenableFutureCallbackRegistry.java
+++ b/spring-core/src/main/java/org/springframework/util/concurrent/ListenableFutureCallbackRegistry.java
@@ -48,7 +48,7 @@ public class ListenableFutureCallbackRegistry<T> {
 	 * @param callback the callback to add
 	 */
 	@SuppressWarnings("unchecked")
-	public void addCallback(ListenableFutureCallback<? super T> callback) {
+	public void addCallback(final ListenableFutureCallback<? super T> callback) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@MoonJongMin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=671)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=671) to manually synchronize the status of this Pull Request.

snicoll:
@MoonJongMin thanks for the PR but we can't create new threads as you suggest in a class of the core framework. If you're still experiencing this issue, please create a new issue with a small sample that we can run ourselves. That will help us debug and find a more appropriate place to try to fix the issue you're experiencing.

PR #673, opened 2014-10-23
Some methods in the org.springframework.core.task.AsyncListenableTaskExecutor class need to be added.

MoonJongMin:
org.springframework.core.task.AsyncListenableTaskExecutor has submitListenable(Callable<T>) method.
This return ListenableFuture object after
ListenableFutureTask.execute(ListenableFutureTask future).
As Juergen Hoeller mentioned in the
SPR-12358(https://jira.spring.io/i#browse/SPR-12358),
if the async task has already been fully executed at the time of the
addCallback call,
Spring's ListenableFutureCallbackRegistry is executing the callback in
the caller's thread
since the async thread can't be used for it anymore.
I think AsyncListenableTaskExecutor is needed to be improved in order to
add callback
before ListenableFutureTask.execute(ListenableFutureTask future) method
is called.

ISSUE : SPR-12364

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/scheduling/commonj/WorkManagerTaskExecutor.java
+++ b/spring-context-support/src/main/java/org/springframework/scheduling/commonj/WorkManagerTaskExecutor.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
 import java.util.concurrent.FutureTask;
+
 import javax.naming.NamingException;
 
 import commonj.work.Work;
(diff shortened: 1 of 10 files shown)


pivotal-issuemaster:
@MoonJongMin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=673)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=673) to manually synchronize the status of this Pull Request.

snicoll:
Thanks for the PR but we'd rather not add overloaded methods like this. JmsTemplate is usually used internally with a more defined API where such rich type is exposed at that level. The infrastructure is usually used internally in a higher-level API that can expose the rich type.

PR #712, opened 2014-12-27
SPR-7509 JSP tag spring:eval does not apply @NumberFormat and @DateTimeF...

bedla:
...ormat formatting on the expression when using the var attribute

Hi, I am sending you pull request witch fixes SPR-7509 according to discussion. Hope you will like it. Thx Ivos


rstoyanchev:
Closing along with related issue.

PR #719, opened 2015-01-14
Improve Netty4ClientHttpRequestFactory configurability and memory usage (SPR-12623)

flozano:


rstoyanchev:
Netty request has been removed in `main` together with `AsyncRestTemplate`.

PR #726, opened 2015-01-25
SPR-6343 ChildOf annotation.

janesser:
How to reference parent-bean from annotated spring, like with component-scan?

The question came up during work life when we were trying to shift towards component-scan. Found that "historical" issue with most of the most given.

Here some pull-request extended with some comments and a test-case.


janesser:
See https://github.com/spring-projects/spring-framework/pull/729

PR #737, opened 2015-02-11
Fix Animal Sniffer

wilkinsona:
Jasper Reports’ transitive dependency on spring-context (via
castor-xml which is a new dependency in 6.0.3) was being mapped by
Gradle to a dependency on the spring-context project. For reasons that
I do not fully understand this was causing -source and -javadoc jars
to be added to the project's compile classpath which is used by the
Animal Sniffer Ant task. When the task runs these jars do not exist
which causes it to fail. This commit fixes the problem by adding an
exclusion of org.springframework:spring-context to the Jasper Reports
dependencies in spring-context-support and spring-webmvc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -143,7 +143,6 @@ configure(allprojects) { project ->
 
 		inputs.dir sourceSets.main.output.classesDir
 		inputs.dir copyJavaApiSignature.to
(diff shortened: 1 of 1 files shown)


snicoll:
Merged, Thanks a bunch Andy!

PR #742, opened 2015-02-23
Add strategy for handling Stomp client errors (SPR-12732)

jaymes-bearden:
As mentioned in SPR-12732, the current implementation does not allow for sufficient information to be return in a STOMP ERROR frame to the client during errors resulting from either client message error parsing, or output channel processing (perhaps through the use of an output channel interceptor).

This commit enhances the StompSubProtocolHandler by adding the StompSubProtocolErrorHandler strategy with a default implementation matching the existing behavior.

Developers will be able to configure the StompSubProtocolHandler with customized versions of the error handler strategy to delegate error handling behavior. For example, this can allow developers to have different behavior for different types of exceptions.

Issue: SPR-12732


jaymes-bearden:
Wrong branch. Resubmitting pull request against the 4.1.x branch

PR #743, opened 2015-02-23
Add strategy for handling Stomp client errors (SPR-12732)

jaymes-bearden:
As mentioned in SPR-12732, the current implementation does not allow for sufficient information to be return in a STOMP ERROR frame to the client during errors resulting from either client message error parsing, or output channel processing (perhaps through the use of an output channel interceptor).

This commit enhances the StompSubProtocolHandler by adding the StompSubProtocolErrorHandler strategy with a default implementation matching the existing behavior.

Developers will be able to configure the StompSubProtocolHandler with customized versions of the error handler strategy to delegate error handling behavior. For example, this can allow developers to have different behavior for different types of exceptions.

Issue: SPR-12732

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/messaging/AbstractStompSubProtocolErrorHandler.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/messaging/AbstractStompSubProtocolErrorHandler.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2002-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.socket.messaging;
+
+import org.springframework.messaging.simp.stomp.StompCommand;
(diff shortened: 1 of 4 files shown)


rstoyanchev:
We might need to expose another option on `StompWebSocketEndpointRegistration` and also not to forget on the equivalent XML namespace (probably under the `<stomp-element>` element). 

Configuration details aside, the proposed contract should work for exceptions on the inbound channel processing side (i.e. handleMessageFromClient). However there can be exceptions on the outbound channel side (within handleMessageToClient) in which case there is also a Message that we were trying to send. Also there can be exceptions in a different component. The main one I have in mind is StompBrokerRelayMessageHandler (search for the `sendStompErrorFrameToClient` method) resulting in in an ERROR frame being sent. I can imagine other parts of the code potentially doing something similar in the future, i.e. sending an ERROR to the client.

So overall I like the idea of a StompSubProtocolErrorHandler configured on the StompSubProtocolHandler but I think the contract needs more work to accommodate all use cases -- (1) exception on the inbound handling side, (2) exception on the outbound side (in which case we have a Message that we were trying to send), and (3) an ERROR frame sent by someone on the outbound channel. In all these cases we might want to ask a StompSubProtocolErrorHandler to get involved.

Yet one more comment that the StompBrokerRelayMessageHandler does not need to send the message on the WebSocketSession itself. It's enough to help to create a WebSocketMessage and perhaps also indicate if we should proactively close the connection. Currently we always do but if we are going to provide some flexibility we should think of all the options that might apply.

datton:
Sounds good!

I'll look for more exit points for error frames and code up those enhancements too. I'll see if I can address the other posted comments too.

jaymes-bearden:
Sounds good!

I'll look for more exit points for error frames and code up those enhancements too. I'll see if I can address the other posted comments too.

rstoyanchev:
This is now available with commit https://github.com/spring-projects/spring-framework/commit/696a010e81c678f449320c3573cc26a0558baa07.

PR #744, opened 2015-02-23
Permit easier overriding of mapping

ryanskow:
Following the feature enabled by SPR-7812, we overrode `RequestMappingHandlerMapping.getCustomMethodCondition()` to add a custom `RequestCondition` that does some evaluation of custom headers and custom annotations on `@RequestMapping` endpoints. (The end result is beyond what is possible using `headers` on `@RequestMapping`.) We then discovered `WebMvcConfigurationSupport` instantiates its own `RequestMappingHandlerMapping` in `requestMappingHandlerMapping()`, requiring overriding of that protected method to use our new subclass. 

The existing `requestMappingHandlerMapping()` not only instantiates the mapping, but performs all of the default configuration of it. Initially, we copied that method intact into our subclass, replacing only the `new RequestMappingHandlerMapping()` with our own subclass. However, this duplicates the default config code in our code and exposes the possibility that we may miss changes to that config in future releases of Spring.

We did a simple refactor to extract the application of default configuration to into a separate method so overriders of `requestMappingHandlerMapping()` can simply pass their custom implementation into the new `configureDefaultHandlerMapping()`.

Example override in `WebMvcConfigurationSupport` subclass:

```
@Bean
public RequestMappingHandlerMapping requestMappingHandlerMapping() {
    RequestMappingHandlerMapping handlerMapping = new RequestMappingHandlerMapping() {
        protected RequestCondition getCustomMethodCondition(Method method) {
            // do custom request condition processing
        }
    };
    configureDefaultHandlerMapping(handlerMapping);

    return handlerMapping;
}
```

Issue: [SPR-12746](https://jira.spring.io/browse/SPR-12746)

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/WebMvcConfigurationSupport.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/WebMvcConfigurationSupport.java
@@ -223,7 +223,7 @@ public class WebMvcConfigurationSupport implements ApplicationContextAware, Serv
 	 */
 	@Bean
 	public RequestMappingHandlerMapping requestMappingHandlerMapping() {
-		RequestMappingHandlerMapping handlerMapping = new RequestMappingHandlerMapping();
+		RequestMappingHandlerMapping handlerMapping = createRequestMappingHandlerMapping();
 		handlerMapping.setOrder(0);
 		handlerMapping.setInterceptors(getInterceptors());
 		handlerMapping.setContentNegotiationManager(mvcContentNegotiationManager());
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Thanks, how about a `createRequestMappingHandlerMapping` protected method that would be invoked by the base class to obtain the instance prior to configuring it? 

That way the sub-class has a clear extension point vs overriding the @Bean method and replicating what it does, which even if it is simple at the moment, leaves the possibility that the base class might do more in the future.

ryanskow:
Thank you for the idea!  We tried that approach first actually, but ran into issues when running within Spring Boot.  We found a way around that now and used your suggestion.  Please see the updated pull request.

rstoyanchev:
This is now merged, thanks!

PR #745, opened 2015-02-24
Provide access to tasks scheduled with the ScheduledTaskRegistrar

tobiasmh:
https://jira.spring.io/browse/SPR-12748

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/config/ScheduledTaskRegistrar.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/config/ScheduledTaskRegistrar.java
@@ -1,5 +1,6 @@
+
 /*
- * Copyright 2002-2014 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-context/src/main/java/org/springframework/scheduling/config/ScheduledTaskRegistrar.java
```diff
@@ -122,6 +122,14 @@ public void setTriggerTasks(Map<Runnable, Trigger> triggerTasks) {
 	public void setTriggerTasksList(List<TriggerTask> triggerTasks) {
 		this.triggerTasks = triggerTasks;
 	}
+	
+	/**
+	 * Return trigger tasks as a list of {@link TriggerTask}
+	 * @since 4.2
+	 */
+	public List<TriggerTask> getTriggerTaskList() {
+		return this.triggerTasks;
```
snicoll: Regardless of whether we want to do this or not, we shouldn't expose the raw mutable list. This also means that your tests must be rewritten. 

Before doing that, let's discuss on the Jira issue what we want to do here. Thanks!

tobiasmh: If you guys are happy with moving forward on an implementation for this after the discussion takes place in Jira, I am happy to rewrite the code as needed.

snicoll: Yes, that's ok. Thanks!

snicoll:
Merged with some updates. Thanks.

PR #746, opened 2015-02-25
@AuthenticationPrincipal in handler parameters

fromi:
By adding this `AuthenticationPrincipalArgumentResolver` to the arguments resolvers, `@AuthenticationPrincipal` should be supported and we will be able to inject Spring Security principal object.
It will be useful when using `@MessageMapping` to handle remote STOMP messages over websockets.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SimpAnnotationMethodMessageHandler.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SimpAnnotationMethodMessageHandler.java
@@ -299,6 +299,7 @@ public class SimpAnnotationMethodMessageHandler extends AbstractMethodMessageHan
 		resolvers.add(new HeaderMethodArgumentResolver(this.conversionService, beanFactory));
 		resolvers.add(new HeadersMethodArgumentResolver());
 		resolvers.add(new DestinationVariableMethodArgumentResolver(this.conversionService));
+		resolvers.add(new AuthenticationPrincipalArgumentResolver());
 
 		// Type-based argument resolution
 		resolvers.add(new PrincipalMethodArgumentResolver());


fromi:
Seems like there are two interfaces named "HandlerMethodArgumentResolver"...

sbrannen:
As an aside [AuthenticationPrincipalArgumentResolver](http://docs.spring.io/autorepo/docs/spring-security/current/apidocs/org/springframework/security/web/bind/support/AuthenticationPrincipalArgumentResolver.html) is a class from the Spring Security project and can therefore not be referenced from the core Spring Framework.

fromi:
I solved my problem by creating a custom HandlerMethodArgumentResolver. I copied the "supportsParameter" method from AuthenticationPrincipalArgumentResolver, and most of "resolveArgument" method from PrincipalMethodArgumentResolver.
I am sad with this solution though!

rstoyanchev:
Spring Security 4 is almost GA. It has the resolver you need and much more:
https://github.com/spring-projects/spring-security/blob/master/messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java

https://spring.io/blog/2015/02/26/spring-security-4-0-0-rc2-released

/cc @rwinch

fromi:
Fantastic, I saw that it was supposed to handle WS security in a better way, but I did not knew if @AuthenticationPrincipal was going to be managed.
Any info if this is compatible with current spring boot version, or when it will be referenced in a new spring boot release ?

rwinch:
@fromi Spring Security 4 is and will be compatible with the current Spring Boot version.

I don't know how Spring Boot plans to handle the upgrade to Spring Security 4 since it is a non-passive release. My guess is that you will need to override the Spring Security version (i.e. provide the spring-security.version property in your pom). @philwebb may be able to comment more authoritatively on this.

PR #747, opened 2015-02-27
Fix SPR-11759, added support of List<String> injection from YAML propert...

cooniur:
Related to this issue: https://jira.spring.io/browse/SPR-11759
Also related to this discussion: https://github.com/spring-projects/spring-boot/issues/501

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
@@ -300,6 +300,10 @@ public abstract class YamlProcessor {
 					buildFlattenedMap(result,
 							Collections.singletonMap("[" + (count++) + "]", object), key);
 				}
+
+				if (value instanceof List) {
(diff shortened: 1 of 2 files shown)


snicoll:
It's not that easy. You just broke collection injection (which is excercised by the test you had to change).

cooniur:
I'd like to do more test. Thanks!

On Friday, February 27, 2015, Stéphane Nicoll <notifications@github.com>
wrote:

> It's not that easy. You just broke collection injection (which is
> excercised by the test you had to change).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/spring-projects/spring-framework/pull/747#issuecomment-76382452>
> .
>


-- 
Sent from Gmail Mobile

cooniur:
@snicoll 
Thanks for pointing this out! I'll try to find another solution.

rubesMN:
Did this ever get in?

cooniur:
@rubesMN 
No, it hasn't happen yet.

PR #748, opened 2015-02-27
Fix SpEL Compilation mode IMMEDIATE.

ericbottard:
Require 1 interpreted pass only (instead of 2)

cc @aclement


snicoll:
The Javadoc of `IMMEDIATE` sates:

> In immediate mode, expressions are compiled as soon as possible (usually after 1 interpreted run).

The current code looks legit to me from that perspective.

PR #750, opened 2015-03-02
HttpMessageConverters should support streaming

poutsma:
All HttpMessageConverters should support StreamingHttpOutputMessage.
Specifically, the BufferedImageHttpMessageConverter and
FormHttpMessageConverter should do so.

Issue: SPR-12715


rstoyanchev:
Merged, thanks.

PR #751, opened 2015-03-03
Organize reference documentation

bclozel:
- modularize the reference documentation
- update asciidoctor gradle plugin to 1.5.2


bclozel:
Hey @rwinch and @sdeleuze 

I know you guys worked on this at some point. I've modularized the documentation and upgraded the asciidoctor gradle plugin. We're still relying here on the docbook plugin as well, since pdf and epub asciidoctor backends are still in alpha version.

Anything I should know before merging this?

@poutsma sorry I could not reuse your PR, as rebasing it against master was a nightmare...

rwinch:
@bclozel  Thanks for reaching out. 

In the event you are interested...Historically the reason we used a single file was tooling limitations at the time. This has obviously changed though.

One thing I would consider is making each document start with a heading of one and using [leveloffset](http://asciidoctor.org/docs/user-manual/#relative-leveloffset). The reason is that this allows the context of what heading level is used is independent of the file being included making it easy to render the file in other contexts.

rstoyanchev:
While I agree in general with the change, I think reverting back to the previous complete split along chapters makes it too fine grained to aid in understanding.

I'd like to suggest we consider a less fine-grained split along the lines of what makes sense when editing. For example: overview, what's new, core, testing, data access, web mvc, portlet mvc, websocket, integration, appendices. 

This is not too different from the idea with JIRA components to aid in categorizing issues. It's also how we might want to organize the documentation if we were to create a more helpful "landing page" for the documentation rather than the current Table of Contents which requires no less than 48 page down hits on my laptop (never mind a smaller resolution device).

bclozel:
Thanks @rstoyanchev @rwinch , I'll update this PR with the suggested changes.

bclozel:
I've revisited this PR with suggested changes.

## Documentation Layout

The documentation is now organized in bookparts, so there are now less files to track. LevelOffsets for bookparts don't need to be incremented, as they're already starting at level 0. Still, I separated some parts that were long/important enough to be in their own file; those are imported with an incremented leveloffset.

Here is the new documentation layout:
```
index
  |-- overview[]
  |
  |-- whats-new[]
  |
  |-- core[]
  |   |-- core-beans[leveloffset=+1]
  |   |-- core-aop[leveloffset=+1]
  |-- testing[]
  |
  |-- data-access[]
  |
  |-- web[]
  |   |-- web-mvc[leveloffset=+1]
  |   |-- web-view[leveloffset=+1]
  |-- integration[]
  |
  |-- appendix[]
      |-- appx-spring-tld[leveloffset=+1]
      |-- appx-spring-form-tld[leveloffset=+1]
```

## Remaining issues

When viewing individually a bookpart (e.g. looking at `core` in its own HTML rendered page):

1. `[partintro]` are not rendered in individual pages, since the bookpart is not a child of a book document when being rendered indivually

```
asciidoctor: ERROR: partintro block can only be used when doctype is book
and it's a child of a book part. Excluding block content.
```

2. links to other bookparts don't work; they probably need to be updated:

```
<<section, My section>> must be changed into <<filename#section, My section>> 
```

Of course, both issues are solved in single document mode.

## Notes

This is a first stab at reorganizing our documentation, and I believe this must be done iteratively. Breaking down our huge `index.adoc` file will help us and avoid PR decay, merge conflicts, etc. And should give some fresh air to our file editors and browsers.

@rwinch I know we're now relying on docbook, but I was wondering if something needed to be done re asciidoctor/asciidoctor-gradle-plugin#52 if we want to support multi-page documentation natively?

Thoughts?

rwinch:
@bclozel 

> I know we're now relying on docbook, but I was wondering if something needed to be done re asciidoctor/asciidoctor-gradle-plugin#52 if we want to support multi-page documentation natively?

As I imagined it, this feature would not have supported native multi page documentation. It was to update the Gradle plugin to do the integration with docbook for the user.

As soon as there is multi page html output available, then this would not actually be necessary.

PR #757, opened 2015-03-18
Fix StackOverflowError due to recursion in DefaultManagedAwareThreadFactory

gedrox:
Bug easily reproducible by passing invalid JNDI name leading to recursion in newThread method


jhoeller:
Fixed in a slightly extended revision of DefaultManagedAwareThreadFactory, to be pushed ASAP.

jhoeller:
JIRA reference: https://jira.spring.io/browse/SPR-12830

PR #758, opened 2015-03-19
Allow subclass adaptation of @Scheduled runnable creation.

ManicRooks:
Small change in ScheduledAnnotationBeanPostProcessor to allow subclass adaptation of @Scheduled runnable creation. This allows generic behaviour for scheduled tasks to be added in one place.


FancyCoder0:
Dear ManicRooks,

We are researchers working on identifying redundant development and duplicated pull requests. We have found there is a pull request: https://github.com/spring-projects/spring-framework/pull/1827 which might be a potentially duplicate to this one. We would really appreciate if you could help us to validate and give us feedback.

Sincerely,
Luyao

PR #759, opened 2015-03-20
Support {/...} patterns in UriComponents(Builder)

poutsma:
This commit introduces support for "Path Segment URI Variable
expansion", see https://tools.ietf.org/html/rfc6570#section-3.2.6.
In practice, this means that URI template variables prefixed with a '/'
are treated like path segments and - as such - will encode any '/'
found. For example: {/foo} expanded with "bar/baz" with result in
"bar%2F".

Issue: SPR-12750

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/HierarchicalUriComponents.java
+++ b/spring-web/src/main/java/org/springframework/web/util/HierarchicalUriComponents.java
@@ -22,11 +22,11 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


rstoyanchev:
I've gone for a slightly different approach, essentially detecting `{/var}` syntax at the point when `path(String)` is added to the builder, splitting it up, and translating to a sequence of path and path segments.

rstoyanchev:
Actually I've realized what I did isn't going to work for all cases. Re-opening..

poutsma:
@rstoyanchev I've updated this PR with a new approach. Please take a look.

rstoyanchev:
Looks good, merged.

PR #772, opened 2015-04-07
SPR-12893 Support OkHttp as an implementation for ClientHttpRequestFactory / AsyncClientHttpRequestFactory

lukiano:


rstoyanchev:
This is now merged.

PR #773, opened 2015-04-09
Performance Optimization on getResultSetValue

DR9885:
Performance Optimization, using ".equals" is signifigantly slower than "==". 

When using "BeanPropertyRowMapper", we ran load tests and found significant lack of performance when calling "getColumnValue". I noticed that "getColumnValue" simply just calls "JdbcUtils.getResultSetValue". These tests were run on an table with roughly 70 columns.

After making change I ran three tests
1) with existing BeanPropertyRowMapper
2) with optimized BeanPropertyRowMapper
3) with a Custom RowMapper, mapped exactly to the class.
# Benchmark Results

1) round: 0.38 [+- 0.11], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 1, GC.time: 0.01, time.total: 7.98, time.warmup: 0.28, time.bench: 7.70
2) round: 0.23 [+- 0.02], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 0, GC.time: 0.00, time.total: 4.80, time.warmup: 0.22, time.bench: 4.58
3) round: 0.22 [+- 0.02], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 0, GC.time: 0.00, time.total: 4.62, time.warmup: 0.19, time.bench: 4.43

As you can see the optimized version makes BeanPropertyRowMapper almost as fast as a custom made RowMapper.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/JdbcUtils.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/JdbcUtils.java
@@ -144,52 +144,51 @@ public abstract class JdbcUtils {
 		Object value;
 
 		// Explicitly extract typed value, as far as possible.
-		if (String.class.equals(requiredType)) {
-			return rs.getString(index);
-		}
-		else if (boolean.class.equals(requiredType) || Boolean.class.equals(requiredType)) {
-			value = rs.getBoolean(index);
-		}
-		else if (byte.class.equals(requiredType) || Byte.class.equals(requiredType)) {
-			value = rs.getByte(index);
-		}
-		else if (short.class.equals(requiredType) || Short.class.equals(requiredType)) {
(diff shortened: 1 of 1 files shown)


jhoeller:
Could you please create a JIRA issue for this one? We have quite a few equals comparisons between Class references in the codebase, so this might be worth revisiting on a wider scale. In any case, a JIRA issue will make sure it doesn't get lost...

Juergen

DR9885:
Opened a Ticket: https://jira.spring.io/browse/SPR-12926

panchenko:
This PR can be closed, as the corresponding change was performed for the whole codebase in b4095c3e1d61fc7279e6b2ca105ec60c3cd7128c

PR #775, opened 2015-04-10
Add isNotEmpty convenience method to StringUtils

kamilszymanski:
In some (most?) cases `isNotEmpty(someString)` expresses your intentions better than `hasLength(someString)` and is more readable and less error-prone than `!isEmpty(someString)`


sbrannen:
Thanks for the suggestion.

I implemented it slightly differently in conjunction with https://jira.spring.io/browse/SPR-12909.

PR #776, opened 2015-04-10
FutureAdapter should wrap RuntimeExceptions

poutsma:
RuntimeExceptions thrown from FutureAdapter.adapt() should be wrapped in
an ExecutionException, not thrown as is.

Issue: SPR-12887


rstoyanchev:
This is merged. Thanks.

PR #777, opened 2015-04-15
Fix javadoc error with multiple profiles

mikeroda:
When multiple profiles are specified, they are treated as an OR condition. The javadoc incorrectly says "and/or".


sbrannen:
The Javadoc was perhaps misleading though technically correct.

Strictly speaking, `A and/or B` implies `(A && B) || (A || B)`.

Of course that is logically equivalent to `A || B`. :smirk:

In any case, I have revised the Javadoc in commit 96e6406b4b45c223f85257d09f03c069be62ec20.

Thanks for bringing this to our attention!

PR #778, opened 2015-04-17
SPR-12003 StringUtils.commaDelimitedListToSet sorts resulting list

ameyjadiye:
Fix for SPR-12003, the sorting is extra and not necessary computation towards the output, we should use  LinkedHashSet. Performance of LinkedHashSet is always better than TreeSet.


jhoeller:
I've already added this as a direct commit for commaDelimitedListToSet as well as for removeDuplicateStrings. Thanks for the pull request, in any case!

PR #779, opened 2015-04-18
HttpMessageConverterExtractor.extractData should TRY each HttpMessageConverter

jhcao23:
In `HttpMessageConverterExtractor.extractData` method, it loops for all `HttpMessageConverter` to return the converted result. This is correct but not perfect. Sometimes, it's not always possible to determine `canRead` with only 'return type' and `MediaType`. So my suggestion is, within the `for` loop, 'TRY' each `HttpMessageConverter` - I mean wrap the `return` statement within a try-catch. As I said, it's possible that one or more `HttpMessageConverter` `canRead` but in practice it `throws IOException`, then with the round robin loop, it may ultimately find the last one `HttpMessageConverter` which `canRead` and really convert the data into the expected return type. If none of them did the expected conversion, let `RestClientException` throw as usual.

So my suggested solution is:

``` Java
    public T extractData(ClientHttpResponse response) throws IOException {

        MessageBodyClientHttpResponseWrapper responseWrapper = new MessageBodyClientHttpResponseWrapper(response);
        if(!responseWrapper.hasMessageBody() || responseWrapper.hasEmptyMessageBody()) {
            return null;
        }
        MediaType contentType = getContentType(responseWrapper);

        for (HttpMessageConverter<?> messageConverter : this.messageConverters) {
            if (messageConverter instanceof GenericHttpMessageConverter) {
                GenericHttpMessageConverter<?> genericMessageConverter = (GenericHttpMessageConverter<?>) messageConverter;
                if (genericMessageConverter.canRead(this.responseType, null, contentType)) {
                        logger.debug("Reading [" + this.responseType + "] as \"" +
                                contentType + "\" using [" + messageConverter + "]");
                    }
                    try{
                        return (T) genericMessageConverter.read(this.responseType, null, responseWrapper);
                    }catch(IOException|HttpMessageNotReadableException e){
                            logger.debug("FAILURE of Reading [" + this.responseType + "] as \"" +
                                    contentType + "\" using [" + messageConverter + "]", e);

                        }
                    }
                }
            }
            if (this.responseClass != null) {
                if (messageConverter.canRead(this.responseClass, contentType)) {
                        logger.debug("Reading [" + this.responseClass.getName() + "] as \"" +
                                contentType + "\" using [" + messageConverter + "]");
                    }
                    try{
                        return (T) messageConverter.read((Class) this.responseClass, responseWrapper);
                    }catch(IOException|HttpMessageNotReadableException e){
                            logger.debug("FAILURE of Reading [" + this.responseType + "] as \"" +
                                    contentType + "\" using [" + messageConverter + "]", e);

                        }
                    }
                }
            }
        }
        throw new RestClientException(
                "Could not extract response: no suitable HttpMessageConverter found for response type [" +
                        this.responseType + "] and content type [" + contentType + "]");
    }   
```


bclozel:
Hi @jhcao23
Could you open an [improvement issue on JIRA](https://jira.spring.io/browse/SPR/) in order to discuss that change? This pull request cannot be merged anyway - it does not contain a commit with the proposed change.

I'm not sure I'm getting the rationale behind that change:
* what kind of converter would require this? Could you tell us a bit more?
* spring framework (and other projects) implement a great deal of message converters; maybe [one of those converters](https://github.com/spring-projects/spring-framework/tree/master/spring-web/src/main/java/org/springframework/http/converter) can give you some hint on how to solve your issue?
* I'm a bit reluctant to rely on canRead + exceptions for control flow here; the behavior can be hard ot predict here.

Thanks!

PR #782, opened 2015-04-22
SPR-12945: Add merged RequestPostProcessor to front on merge

rwinch:
Previously MockHttpServletRequestBuilder merge method would append the
parent's (default) RequestPostProcessor implementations to the end. This
means that the default RequestPostProcessor implementations would override
values set by previous RequestPostProcessor implementations.

This commit ensures that the default RequestPostProcessor are preformed
first so that additional RequestPostProcessor implementations override
the defaults.

Issue: SPR-12945


rstoyanchev:
This is now merged. Note that this PR seems to be based on the 4.1.x branch. I cherry-picked the commit to the master branch targetting 4.2.

PR #794, opened 2015-05-07
Introduce Marshalling MessageConverter

poutsma:
This commit introduces a messaging.converter.MessageConverter that
marshals to/from XML using the abstractions provided in the OXM module.

Issue: SPR-12726


rstoyanchev:
This is now merged, thanks! Only removed the constructor with marshaller and unmarshaller since setters should still work for that less common case.

PR #795, opened 2015-05-08
Remove duplicate separators when combining

poutsma:
Prior to this commit, AntPathMatcher would not correctly combine a path
that ends with a separator with a path that starts with a separator.
I.e. /foo/ + /bar combined into /foo//bar.

This commit removes the duplicates separator. It also improves
RequestMappingInfo's String representation.

Issue: SPR-12975


sbrannen:
This has been merged into `master`. Thanks!

PR #797, opened 2015-05-10
Conditional Error Message Formatting

marschall:
Create error message strings only if an actual error has happened. This
seems to be the common pattern in the code base base and saves both CPU
and memory. We found only very few places where unconditional error
message formatting happens.
- fix unconditional error message formatting in
  AdviceModeImportSelector
- fix unconditional error message formatting in AnnotationAttributes
- fix unconditional error message formatting in
  ReadOnlySystemAttributesMap, this also fixes a potential
  NullPointerException

I did sign the CLA.

Issue: SPR-13007


sbrannen:
This has been merged in `master`. Thanks!

PR #799, opened 2015-05-15
SPR-13033: Use @RequestMapping of correct class

ractive:
`MvcUriComponentsBuilder::fromMethodCall` creates wrong URLs with derived controller classes. The `@RequestMapping` of the declaring class of the method that is called is used instead of the `@RequstMapping` of the given controller class.

The same issue exists for the `fromMethod(Method method, ...)` methods. You'd also need to add the class of the controller to get the correct type level `@RequestMapping`. I'm not sure how you want to handle such API breaking changes though...

https://jira.spring.io/browse/SPR-13033


rstoyanchev:
This is now merged. I've made sure all method accept a controllerType. The existing `fromMethod` is deprecated. Thanks!

PR #800, opened 2015-05-18
Support OkHttp as (Async)ClientHttpRequestFactory

poutsma:
This commit introduces support for OkHttp
(http://square.github.io/okhttp/) as a backing implementation for
ClientHttpRequestFactory and AsyncClientHttpRequestFactory.

Issue: SPR-12893


rstoyanchev:
This is now merged. Thanks!

PR #801, opened 2015-05-19
Parse @Scheduled attributes as long instead of int.

echatman:
Trivial change: use `Long.parseLong` instead of `Integer.parseInt` because the assigned fields are of type `long`, not `int`. The current code is causing errors on startup when I try to use a value too big for an `int`.


snicoll:
Thanks. Have you [signed the CLA](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md)?

echatman:
Yes

snicoll:
Merged, thanks!

PR #802, opened 2015-05-19
Fix missing space in assertion message

luontola:
An assertion failure from `MockMvcResultMatchers.jsonPath` used to give an assertion message as follows:

```
java.lang.AssertionError: JSON pathhealthChecks.dummy
Expected: is "OK"
     but: was null
    at org.springframework.test.util.MatcherAssertionErrors.assertThat(MatcherAssertionErrors.java:80)
    at org.springframework.test.util.JsonPathExpectationsHelper.assertValue(JsonPathExpectationsHelper.java:92)
    at org.springframework.test.web.servlet.result.JsonPathResultMatchers$1.match(JsonPathResultMatchers.java:56)
    at org.springframework.test.web.servlet.MockMvc$1.andExpect(MockMvc.java:152)
```

This pull request fixes the missing space before the expression, so that the assertion message will be `JSON path healthChecks.dummy`


rstoyanchev:
Merged, thanks!

PR #808, opened 2015-05-28
SPR-13080 SseEventBuilder needs event rather than name

litpho:
As per the W3C recommendation (http://www.w3.org/TR/eventsource/#event-stream-interpretation) the event type field is called "event".
The SseEmitter.SseEventBuilder class erroneously builds it using "name" as a prefix.

Issue created as https://jira.spring.io/browse/SPR-13080


rstoyanchev:
Merged thanks!

PR #809, opened 2015-05-29
Long waiting on jms connection recovering when stopping application

nivolg:
This fix prevent long waiting on jms connection recovering when stopping application.


snicoll:
Have you seen there is a `BackOff` support now? You can basically put whatever value you want there and it will use that instead of the default. I'd rather use that instead. Let me know if I am missing anything.

nivolg:
I have seen BackOff and I use expotential backoff with max recovery interval 5 minutes. It what we need.
But problem is when jms destination is not available for a long time, consumer start sleeping continuously

nivolg:
Consumer sleep continuosly for 5 minutes and when I stop application at this moment then it freeze until consumer wake up. It actual for a fixed backoff with interval 30 sec for example.

PR #810, opened 2015-05-30
Add nullToEmpty convenience method to StringUtils

kamilszymanski:
It allows you to replace common patterns like:

``` java
if(string == null) {
    string = "";
}
doSomethingWith(string);
```

or:

``` java
return string == null ? "" : string;
```

with a `nullToEmpty(string)` method call.


bclozel:
There hasn't been much demand for this, so I'm closing this PR as a result.
Thanks for your contribution!

PR #815, opened 2015-06-09
Allow StopWatch to be started on construction

magJ:
Simplify starting stopwatch after construction

Return "this" from the start method of the stopwatch,
so that a stopwatch can be constructed and started in one line.
Most of the time you use the stopwatch you want to start it
right away, this saves a line of code and makes the stopwatch
a preferable approach to a simple system time comparison.

Issue: SPR-13106

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StopWatch.java
+++ b/spring-core/src/main/java/org/springframework/util/StopWatch.java
@@ -86,28 +86,6 @@ public class StopWatch {
 		this.id = id;
 	}
 
-	/**
-	 * Construct a new stop watch and optionally start an unnamed task.
-	 * @param start if true starts an unnamed task.
-	 * @since 4.2
-	 */
-	public StopWatch(boolean start) {
-		this("", start);
-	}
-
-	/**
-	 * Construct a new stop watch with the given id and optionally start an unnamed task.
-	 * @param id identifier for this stop watch.
-	 * @param start if true starts an unnamed task.
-	 * @since 4.2
-	 */
-	public StopWatch(String id, boolean start) {
-		this.id = id;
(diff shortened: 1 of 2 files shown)


Shredder121:
I'd leave the original method in there, but I don't know what the Spring Team generally does?

magJ:
I’m not sure what you mean by leave the original method in?
Since the modified method has the same signature then we have have both.
Originally I implemented it with a new distinct constructor, but after feedback I changed it to this builder patternish design.

Shredder121:
Binary compatibility is a delicate thing.
[It can all be found in JLS 7 Ch. 13](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.15).
So adding a return type is not binary compatible.

I also like the builder style much better, but I don't know how Spring views binary compatibility.

magJ:
@Shredder121 Looks like the policy is to try and maintain binary compatibility between minor versions. Although the wording seems to indicate that exceptions are often made. I doubt this change would warrant an exception.
http://docs.spring.io/spring-security/site/docs/current/reference/html/introduction.html#release-numbering

Shredder121:
I guess you're right. Sorry for interfering.
Then disregard my comment.

magJ:
Closing since this open PR annoys me.

pivotal-issuemaster:
@magJ Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=815)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=815) to manually synchronize the status of this Pull Request.

PR #817, opened 2015-06-14
Make AbstractAdaptableMessageListener getResponseDestination aware of response JMS Destination if set

bartprokop:
I was trying to implement a kind of router - retrieve a message in JmsListener annotated method, create a new message inside it (that basically embraces the request message) and send it to dynamically chosen queue, based on some processing logic. Unfortunately SendTo annotation doesn't allow for dynamic destination choosing and my incoming messages can't have ReplyTo field set. I've ended witch the following code, that works:

```
@JmsListener(destination = "inbound.queue")
//@SendTo("Only static queue possible")
public Message router(Message src, Session s) throws JMSException {
    TextMessage dst = s.createTextMessage(((TextMessage) src).getText());
    // based on some processing logic, we want dst to be placed in some queue
    // This works, but is a "hack":
    src.setJMSReplyTo(new MQQueue("some.queue"));
    // This doesn't work but is much more clean solution provided with this pull request
    dst.setJMSDestination(new MQQueue("some.queue"));
    return dst;
}
```

There are some good reasons why modifying request (incoming) message is bad idea. First of all, it is good to treat inbound messages as immutable objects and not mess with changing their properties. Second, not all listener use cases are suitable for request/reply patterns. The original meaning for JMSDestination is informative one (to be used by receiving party to check what queue message originates from). Therefore message brokers ignore those properties when sending messages, while try to keep it informative on message reception. So reusing JMSDestination for hinting Spring Container about overriding SendTo annotation is not abusive and will allow to write a more clean code.

### DIFF
--- a/spring-jms/src/main/java/org/springframework/jms/listener/adapter/AbstractAdaptableMessageListener.java
+++ b/spring-jms/src/main/java/org/springframework/jms/listener/adapter/AbstractAdaptableMessageListener.java
@@ -304,7 +304,11 @@ public abstract class AbstractAdaptableMessageListener
 
 	/**
 	 * Determine a response destination for the given message.
-	 * <p>The default implementation first checks the JMS Reply-To
+	 * <p>The default implementation first checks the JMS {@link Destination}
+	 * of retrieved response and uses it if not {@code null}.
+	 * This allows to dynamically override {@link SendTo} annotation value
+         * as well as the JMS Reply-To {@link Destination} of the handled request.
+	 * <p>Then the default implementation checks the JMS Reply-To
 	 * {@link Destination} of the supplied request; if that is not {@code null}
 	 * it is returned; if it is {@code null}, then the configured
 	 * {@link #resolveDefaultResponseDestination default response destination}
@@ -321,6 +325,9 @@ public abstract class AbstractAdaptableMessageListener
 	 */
 	protected Destination getResponseDestination(Message request, Message response, Session session)
 			throws JMSException {
+		if (response.getJMSDestination() != null) {
+			return response.getJMSDestination();
+                }
 
 		Destination replyTo = request.getJMSReplyTo();
 		if (replyTo == null) {


snicoll:
The problem is that you are not allowed to change the `JMS_DESTINATION` field of a `javax.jms.Message` per spec.

Excerpt of the javadoc:

> JMS providers set this field when a message is sent. This method 
> can be used to change the value for a message that has been received.

I agree this sounds nice but I am not too keen to pursue this as it's not spec compliant.

bartprokop:
Hi Stéphane,

I fully agree with your point of view. Could you please consider the following:
 - The JMS spec doesn't forbid to set the JMS Destination. It only says, that it is set when JMS provider dispatch message to the queue and also that setter is provided to allow it to be set to meaningful value on message reception.
 - My pull requests will cover only those corner cases when Message is returned from listener.
 - I can alter my pull request to nullify the JMSDestination after retrieval - note that when calling session.createXXXMessage it is set to null (or not set at all) - checked with IBM MQ and Active MQ clients. I'm aware that this point can be perceived as anti-pattern. However this would make completely no difference to underlying JMS provider.
 - I believe that all of JMS implementations clearly ignore it on dispatch.
 - I was aware of JMS spec, when proposing this change and I still think it is more elegant and purposeful than modifying request message.

snicoll:
I don't think I agree with your interpretation of the spec (cfr. [javadoc of `setJMSDestination`](http://docs.oracle.com/javaee/7/api/javax/jms/Message.html#setJMSDestination(javax.jms.Destination)))

> This message cannot be used by clients to configure the destination of the message. This method is public to allow a JMS provider to set this field when sending a message whose implementation is not its own

Using that field for what you need is clearly against the statement above IMO (hint: Spring Framework is not a JMS provider) and that's not a programming model we want to put in the framework. 

Could you please create an issue in Jira with your actual need and we can take it from there? Juergen and I just discussed this and maybe we can envision a specific objet that holds the message and the destination rather than simply the destination. It's not too late to include that in 4.2 if we can quickly find a good solution for this.

Thanks!

bartprokop:
Right, EE7 is more strict here.

I think there are three cases (when using JmsListener annotated method model):
1. request-reply scenario. The request have a JMSReplyTo set. I've actually exploited this to achieve the point 3 below.
2. processing messages and sending result back to predefined queue - using SendTo annotation.
3. processing message but the decision where to send response is determined on processing request. There is no elegant solution to do this. I can always return void and use JMSTemplate or hack request JMSReplyTo - but this is not elegant solution.

You are right that a kind of "tuple" that contains a Message and Destination returned from JmsListener annotated method would do the job. Actually the tuple do not have to contain Message object could be anything that can be converted to Message in accordance to Spring programming model. Does it also makes sense to allow return array of messages to be dispatched?

I'll create relevant JIRA ticket ASAP.

Cheers,

Bart

bartprokop:
JIRA ticket SPR-13133 created.

PR #818, opened 2015-06-15
Header values are optional in Stomp 1.1 and 1.2

jking-roar:
Contributor License Agreement


rstoyanchev:
Merged, thanks!

PR #819, opened 2015-06-17
fix StringIndexOutOfBoundsException in AntPathMatcher for leading *

martinlippert:


rstoyanchev:
This has been merged, thanks! Note that I rebased the commit on top of the latest in master, so you might have to reset your fork (since you submitted from the master branch).

PR #820, opened 2015-06-18
Reduce BeanDefinition cloning from isFactoryBean

philwebb:
Update AbstractBeanFactory.isFactoryBean to only create the merged local
bean definition when absolutely necessary.


jhoeller:
After #23336 and #23337, this seems to be covered and not worth optimizing there anymore.

PR #821, opened 2015-06-18
Use method meta-data in getTypeForFactoryMethod

philwebb:
Update AbstractAutowireCapableBeanFactory.getTypeForFactoryMethod to
use already loaded method meta-data when deducing the type of a factory
method.


jhoeller:
In the meantime, the cached factory method return type is a `ResolvableType` which takes a generic method signature into account. It seems generally sensible to immediate resolve the `Method` handle there (rather than inferring the declared return type from the ASM metadata) since the actual `Method` is subsequently introspected for other purposes as well. Resolving the ASM-declared type name first (which involves a `ClassLoader` call) and then the full `Method` handle shortly thereafter (which comes with the resolved return type `Class` as well as a potential generic declaration) is not likely to provide a performance benefit within the current container boostrap arrangement anymore.

PR #826, opened 2015-06-24
Refactor WebSocket int. tests to work w/ Jetty 9.3

sbrannen:
Recent builds of Jetty 9.3 require that Jetty's own ServletContext
implementation be supplied to WebSocketServerFactory's init() method.
Otherwise, the Jetty server will fail to start with the exception
message: "Not running on Jetty, WebSocket support unavailable".

This commit refactors AbstractSockJsIntegrationTests and all
WebSocketTestServer implementations in order to support this new
requirement. Specifically:
- WebSocketTestServer defines a new getServletContext() method.
  TomcatWebSocketTestServer and UndertowTestServer simply return a
  MockServletContext; whereas, JettyWebSocketTestServer returns the
  ServletContext created internally by Jetty's ServletContextHandler.
- The setup() method in AbstractSockJsIntegrationTests has been
  refactored so that the WebApplicationContext is supplied the
  appropriate ServletContext, after the WebSocketTestServer has been set
  up.

Issue: SPR-13162

### DIFF
--- a/spring-websocket/src/test/java/org/springframework/web/socket/JettyWebSocketTestServer.java
+++ b/spring-websocket/src/test/java/org/springframework/web/socket/JettyWebSocketTestServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2014 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


sbrannen:
This has been merged into `master` slightly differently in e8c8d2a6ad1d75b72bb680a2bf76043a7a8cee79.

PR #829, opened 2015-06-29
Add an ObjectMapper constructor to MappingJackson2JsonView

candrews:
Instead of having to use the default constructor then calling #setObjectMapper(ObjectMapper), this constructor allows the MappingJackson2JsonView to be created and configured with the desired object mapper in one step.


sdeleuze:
I think that makes sense, could you modify your commit to add a similar `XmlMapper` constructor to `MappingJackson2XmlView` as well?

candrews:
@sdeleuze I've updated the PR - thanks for the feedback!

sdeleuze:
Merged in master with this polished commit https://github.com/spring-projects/spring-framework/commit/d4c74ad34645119f2f18fc813fe5b9e141560210

PR #831, opened 2015-07-01
SPR-13183 Send BindingResult Errors in ResponseEntityExceptionHandler

bonhamcm:
When a RESTful API method returns a 400 status due to validation errors,
those errors are not returned. This commit changes the handleBindException()
and handleMethodArgumentNotValid() methods to return the BindingResult
errors as the response entity.

Issue: SPR-13183

I have signed the SpringSource CLA.


rstoyanchev:
Thanks for the PR but as discussed in JIRA this is expected to be an application choice.

PR #832, opened 2015-07-03
Configure embedded database with properties 

akalotkin:
Spring embedded database support isn't very configurable at the moment.
These changes related to SPR-12455.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -30,12 +30,14 @@ import org.springframework.util.Assert;
 
 /**
  * Helper class that encapsulates the specification of a method parameter, i.e.
- * a Method or Constructor plus a parameter index and a nested type index for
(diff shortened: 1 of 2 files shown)


sbrannen:
@akalotkin, thanks for taking the time to implement this and issue the pull request!

It definitely looks like this might be a viable way to solve the problem; however, we cannot modify existing contracts for interfaces. Otherwise, we will likely break existing clients of the API.

Specifically, we cannot add new methods to the `ConnectionProperties` interface. 

As an alternative, consider introducing a new sub-interface -- for example, something like `ConfigurableConnectionProperties`.

Please rework your code to ensure that you are not changing existing APIs, squash your changes into a single commit (rebased on `master`), and then update this pull request (i.e., with a forced push).

Thanks,

Sam

sbrannen:
Hi @akalotkin, did you mean to close this pull request?

akalotkin:
No. I'm commiting my updates.

sbrannen:
Ahhh. OK.

I also saw that you opened #833.

PR #834, opened 2015-07-09
DelegatingFilterProxy looks at "registered" WebApplicationContexts

rwinch:
Previously when users registered an additional Filter in the
DispatcherServlet the DelegatingFilterProxy would not find it unless
the context attribute was set. The value for which was difficult to
remember and added quite a bit of additional "if" statements to setting
up Spring Security successfully.

This change allows registering additional known WebApplicationContext
instances. The DispatcherServlet uses this mechanism to register its
context. With that change DelegatingFilterProxy searches for the first
registered WebApplicationContext that contains the bean name.

This change also sets up the ability for Spring Security to lookup
Beans defined in the child context within the Tag libs.

Issue: SPR-13191

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
+++ b/spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2014 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,9 +17,12 @@
 package org.springframework.web.context.support;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
@@ -55,6 +58,7 @@ import org.springframework.web.context.request.WebRequest;
  * This helper class is just the most generic way to access the root context.
  *
  * @author Juergen Hoeller
+ * @author Rob Winch
  * @see org.springframework.web.context.ContextLoader
  * @see org.springframework.web.servlet.FrameworkServlet
  * @see org.springframework.web.servlet.DispatcherServlet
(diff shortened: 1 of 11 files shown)


spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
```diff
@@ -125,6 +136,63 @@ public static WebApplicationContext getWebApplicationContext(ServletContext sc,
...
+	 * Gets all the known {@link WebApplicationContext} for this web app. The order is the
+	 * root context (if specified) followed by any context registered using
+	 * registerWebApplicationContext in the order they were registered.
+	 *
+	 * @param sc ServletContext to find the web application context for. Cannot be null.
+	 * @return the known {@link WebApplicationContext} for this web app. Cannot be null,
+	 * but may be empty.
+	 * @since 4.2
+	 */
+	public static List<WebApplicationContext> getRegisteredWebApplicationContext(ServletContext sc) {
```
sbrannen: This method should be named `getRegisteredWebApplicationContexts` (i.e. plural).

spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
```diff
@@ -125,6 +136,63 @@ public static WebApplicationContext getWebApplicationContext(ServletContext sc,
 		return (WebApplicationContext) attr;
 	}
 
+	/**
+	 * Registers a custom {@link WebApplicationContext} for this web app.
```
sbrannen: Spring's Javadoc convention is to use an imperative statement for the list line -- in other words, "Register ..." instead of "Registers ...". If you don't follow that rule, you run the risk of getting Juergenized. :wink:

spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
```diff
@@ -125,6 +136,63 @@ public static WebApplicationContext getWebApplicationContext(ServletContext sc,
...
+	public static void registerWebApplicationContext(ServletContext sc, String attrName, WebApplicationContext context) {
+		Assert.notNull(sc, "ServletContext must not be null");
+		Assert.notNull(attrName, "attrName must not be null");
+		Assert.notNull(context, "WebApplicationContext must not be null");
+
+		sc.setAttribute(attrName, context);
+
+		@SuppressWarnings("unchecked")
+		LinkedHashSet<String> attrNames = (LinkedHashSet<String>) sc.getAttribute(APPLICATION_CONTEXT_ATTRIBUTES_ATTRIBUTE);
+		if(attrNames == null) {
```
sbrannen: Spring requires a space between `if`, `for`, etc. and the following parenthesis.

rwinch: Thanks! This should be fixed now. Is there an IntelliJ code formatter available?

spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
rwinch: Good catch! This is fixed now.

spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
rwinch: Being Juergenized via proxy...I like it! This is fixed now :)

sbrannen:
Very interesting proposal!

spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java
sbrannen: There's a [Code Style](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style) page.

rwinch:
Closing since this was resolved in https://github.com/spring-projects/spring-framework/commit/1fcd465f2d281a432eee79adfe1d4d8dd7ec3630

PR #836, opened 2015-07-10
Add support for JdbcTempalte.queryForObject(sql,<primitive>.class)

rwinch:
Previously using jdbcTemplate.queryForObject(sql, int.class) produced
a TypeMismatchDataAccessException stating it could not convert an Integer
to an int.

This update allows conversion to the respective primitives instead of just
the wrapper Objects.

Issue: SPR-13220

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/NumberUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/NumberUtils.java
@@ -32,6 +32,7 @@ import java.util.Set;
  *
  * @author Juergen Hoeller
  * @author Rob Harrop
+ * @author Rob Winch
  * @since 1.1.2
  */
 public abstract class NumberUtils {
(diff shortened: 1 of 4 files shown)


panchenko:
Wouldn't it be simpler to convert primitive class to the corresponding wrapper initially (when initializing SingleColumnRowMapper#requiredType) ?

rwinch:
@panchenko Thanks for your feedback. If I understand the question correctly, you wonder why we don't just convert the value in the constructor and / or setter? If so

*  `NumberUtils.convertNumberToTargetClass(Number, Class<T>)` is used in multiple places (and may be used in more places in the future), so I think that is the correct place to fix this.
*  `convertValueToRequiredType(Object,Class<?>)` which should work for primitive values as well For example, while mapRow passes in `this.requiredType`, the class `SingleColumnRowMapper` is non final so anyone that extends the class can pass in a different type to the method.
* As far as I know there is no easy way to convert from the primitive class to its corresponding wrapper class, so this actually involves quite a bit of code.

Does this help? If not, can you please clarify? Perhaps I am not understanding your proposal.

PR #838, opened 2015-07-16
Reference more recent resources

mdeinum:
There is a new version of AspectJ in Action which is now mentioned in the
_Further Resources_ section of the AOP chapter.

Issue: SPR-13238


snicoll:
Merged, thanks.

PR #840, opened 2015-07-17
Reset Pragma header when caching

felixbarny:
Spring Security by default sets a `Pragma: no-cache` header (see [CacheControlHeadersWriter](https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/header/writers/CacheControlHeadersWriter.java)). Even if explicitly calling `WebContentGenerator#applyCacheSeconds` that header is still present, preventing the browser from caching the resource. So `cacheForSeconds` should remove the Pragma header if it has been set.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/support/WebContentGenerator.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/support/WebContentGenerator.java
@@ -436,6 +436,9 @@ public abstract class WebContentGenerator extends WebApplicationObjectSupport {
 			}
 			response.setHeader(HEADER_CACHE_CONTROL, headerValue);
 		}
+		if (response.containsHeader(HEADER_PRAGMA)) {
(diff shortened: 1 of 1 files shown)


bclozel:
This has been resolved with 09e3fc40e5c24 in [SPR-13252](https://jira.spring.io/browse/SPR-13252).
Thanks!

By the way did [you sign our CLA](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#sign-the-contributor-license-agreement)?

felixbarny:

bclozel:
Thanks @felixbarny 
One more question - did you discover this as a regression when testing the latest RCs with an existing application? Or did you use those HTTP caching features for the first time in an app and found this bug?

(just trying to prepare for a smooth upgrade for all Spring developers...)

felixbarny:
I've discovered this after upgrading from spring-security 3 to 4 as with version 4, the Pragma header is set by default.

PR #843, opened 2015-07-22
Improve DateHeaders in MockServletRequest/Response

bclozel:
Prior to this change, calling the `setDateHeader` method on a
Spring Test MockHttpServletResponse instance would just store the given
long value in a Map, not writing it as a formatted date String.
Also, calling `getDateHeader` on a MockHttpServletRequest would not
support date strings and could not parse those values.

This can be problematic when testing features related to date headers
such as "Expires", "If-Modified-Since", "Last-Modified", etc.

This commit adds formatting and parsing capabilities to Servlet Mocks
for date strings in HTTP headers.

When formatting dates to Strings, the date format used is the one
preferred by the HTTP RFC. When parsing date Strings, multiple date
formats are supported for better compatibility.

Issue: SPR-11912

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
@@ -24,6 +24,8 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.UnsupportedEncodingException;
 import java.security.Principal;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -36,6 +38,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.TimeZone;
+
 import javax.servlet.AsyncContext;
 import javax.servlet.DispatcherType;
 import javax.servlet.RequestDispatcher;
@@ -119,6 +123,18 @@ public class MockHttpServletRequest implements HttpServletRequest {
 
 	private static final ServletInputStream EMPTY_SERVLET_INPUT_STREAM =
 			new DelegatingServletInputStream(new ByteArrayInputStream(new byte[0]));
+
+	/**
+	 * Date formats as specified in the HTTP RFC
+	 * @see <a href="https://tools.ietf.org/html/rfc7231#section-7.1.1.1">Section 7.1.1.1 of RFC 7231</a>
+	 */
+	private static final String[] DATE_FORMATS = new String[] {
+			"EEE, dd MMM yyyy HH:mm:ss zzz",
+			"EEE, dd-MMM-yy HH:mm:ss zzz",
+			"EEE MMM dd HH:mm:ss yyyy"
+	};
+
+	private static final TimeZone GMT = TimeZone.getTimeZone("GMT");
 
 
 	private boolean active = true;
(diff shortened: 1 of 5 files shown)


spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
```diff
@@ -912,6 +928,9 @@ public long getDateHeader(String name) {
 		else if (value instanceof Number) {
 			return ((Number) value).longValue();
 		}
+		else if (value instanceof String) {
+			return parseDateHeader(name, (String) value);
+		}
 		else if (value != null) {
 			throw new IllegalArgumentException(
 					"Value for header '" + name + "' is neither a Date nor a Number: " + value);
```
sbrannen: Need to update exception message as follows:

`"Value for header '" + name + "' is not a Date, Number, or String: " + value`

spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
```diff
@@ -921,6 +940,21 @@ else if (value != null) {
...
+			SimpleDateFormat simpleDateFormat = new SimpleDateFormat(dateFormat, Locale.US);
+			simpleDateFormat.setTimeZone(GMT);
+			try {
+				return simpleDateFormat.parse(value).getTime();
+			}
+			catch (ParseException ex) {
+				// ignore
+			}
+		}
+		throw new IllegalArgumentException("Cannot parse date value \"" + value +
```
sbrannen: AFAIK, in Spring we typically enclose _quoted_ values within exception messages in single quotes.

sbrannen:
What do you think about adding a `setDateHeader()` method in `MockHttpServletRequest` as suggested in the JIRA issue?

spring-web/src/test/java/org/springframework/mock/web/test/MockHttpServletRequest.java
```diff
@@ -912,6 +927,9 @@ public long getDateHeader(String name) {
 		else if (value instanceof Number) {
 			return ((Number) value).longValue();
 		}
+		else if (value instanceof String) {
+			return parseDateHeader(name, (String) value);
+		}
 		else if (value != null) {
 			throw new IllegalArgumentException(
 					"Value for header '" + name + "' is neither a Date nor a Number: " + value);
```
sbrannen: same as previous comment from `spring-test` mock.

spring-web/src/test/java/org/springframework/mock/web/test/MockHttpServletRequest.java
```diff
@@ -921,6 +939,21 @@ else if (value != null) {
...
+			SimpleDateFormat simpleDateFormat = new SimpleDateFormat(dateFormat, Locale.US);
+			simpleDateFormat.setTimeZone(GMT);
+			try {
+				return simpleDateFormat.parse(value).getTime();
+			}
+			catch (ParseException ex) {
+				// ignore
+			}
+		}
+		throw new IllegalArgumentException("Cannot parse date value \"" + value +
```
sbrannen: same as previous comment from `spring-test` mock.

bclozel:
Thanks @sbrannen for this review! I've fixed all the issues you pointed at.
I've updated the javadoc for `addHeader` with the supported types (the implementation details it was describing were wrong anyway). The supported date formats are listed on the `getDateHeader` method.

sbrannen:
Good deal. Thanks!

PR #846, opened 2015-07-23
SPR-13235: Deprecate Velocity support

cynicLT:


snicoll:
This relates to [SPR-13235](https://jira.spring.io/browse/SPR-13235)

Looking at the commit, you are changing a bunch of unrelated stuff such as import statements. You should limit this PR to what you're supposed to change.

jhoeller:
Thanks for the effort to try to help with upcoming tasks. However, please note that we're usually not accepting deprecations or removals as pull requests. Such steps come with fine-tuned deprecation messages and often javadoc pointers to alternative functionality, so we prefer doing them ourselves.

PR #847, opened 2015-07-23
SPR-13229: Remove Tiles 2 support

cynicLT:


jhoeller:
Thanks for the effort to try to help with upcoming tasks. However, please note that we're usually not accepting deprecations or removals as pull requests. Such steps come with fine-tuned deprecation messages and often javadoc pointers to alternative functionality, so we prefer doing them ourselves.

PR #849, opened 2015-07-28
Clarify documentation on @Ordered

chrylis:
The `LOWEST_PRECEDENCE` is not only the default value for `@Ordered` but is also applied to any compared elements that are unannotated.

See spring-projects/spring-boot#3599.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/Order.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/Order.java
@@ -26,8 +26,9 @@ import org.springframework.core.Ordered;
 /**
  * Annotation that defines ordering. The value is optional, and represents order value
  * as defined in the {@link Ordered} interface. Lower values have higher priority.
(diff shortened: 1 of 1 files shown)


chrylis:
Ack, I knew I shouldn't have trusted the "helpful" PR suggestion.

sbrannen:
@chrylis, have you signed the SpringSource CLA?

chrylis:
Yes, I have, under me at Gmail. Previous contributions in a number of places, particularly Cloud Connectors.

snicoll:
As I've said [here](https://github.com/spring-projects/spring-boot/issues/3599#issuecomment-125860493) this is the behaviour of the comparator, not `@Order` itself. I welcome any other opinion but it feels wrong from my perspective to add it there. 

Maybe we could add `@see OrderComparator` to make that link more explicit?

sbrannen:
Done slightly differently in 1ff3af6da37896e229e54fc8592673918f9203f4.

PR #850, opened 2015-07-30
SPR-13293

m-szalik:
Issue https://jira.spring.io/browse/SPR-13293


m-szalik:
It has been changed already.

PR #851, opened 2015-07-31
Remove type parameter of ResponseBodyEmitter

making:
Fix wrong example of `ResponseBodyEmitter` in reference doc


snicoll:
Merged, thanks!

PR #852, opened 2015-08-04
Introduce `EnumToEnumConverter`

quaff:
sometime we need copy a shadow package for export remote service
interface, service interface depend domain object which include domain
enum type, it is necessary to support conversion between origin Enum
type to shadow Enum.


snicoll:
Thanks. Have you signed the CLA?

quaff:
@snicoll , I didn't do that, What should I do ?

snicoll:
Do what's described [in the contributing page](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#sign-the-contributor-license-agreement).  Thanks!

quaff:

snicoll:
:+1:

quaff:
I have updated this commit to avoid swallow IllegalArgumentException.

philwebb:
This seems like quite a specific use-case to add to the `DefaultConversionService`

PR #855, opened 2015-08-06
Update NumberUtils.java - constructor -> valueOf method

vladad:
Replace calling the constructor of Number classes by calling valueOf method.


sbrannen:
Thanks for the tip!

I've committed something strikingly similar to `master`... :wink: 

... along with some general polishing. See my latest three commits in case you're interested.

PR #860, opened 2015-08-17
Register Hibernate4Module and Hibernate5Module when available

candrews:
Issue: https://jira.spring.io/browse/SPR-13355, https://jira.spring.io/browse/SPR-13356


candrews:
I think we may want to disable Feature.USE_TRANSIENT_ANNOTATION and enable Feature.FORCE_LAZY_LOADING - I find such a configuration to be much more expected. However, my PR just does the default configuration.

sdeleuze:
I close this pull request, since the related issues are resolved.

PR #868, opened 2015-09-01
Fix error in function org.springframework.http.HttpHeaders.getContent…

auk:
Function org.springframework.web.client.MessageBodyClientHttpResponseWrapper.hasMessageBody() returns wrong positive answer when http response has no body and no Content-Length header.

Important, that function org.springframework.http.HttpHeaders.getContentLength() returns -1 if header Content-Length does not exist in response and. But we rely only on 0 as answer.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/MessageBodyClientHttpResponseWrapper.java
+++ b/spring-web/src/main/java/org/springframework/web/client/MessageBodyClientHttpResponseWrapper.java
@@ -61,10 +61,7 @@ class MessageBodyClientHttpResponseWrapper implements ClientHttpResponse {
 				responseStatus == HttpStatus.NOT_MODIFIED) {
 			return false;
 		}
-		else if(this.getHeaders().getContentLength() == 0) {
-			return false;
-		}
-		return true;
+		return this.getHeaders().getContentLength() > 0;
 	}
 
 	/**


bclozel:
Hi @auk 
Have you [signed the CLA](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#sign-the-contributor-license-agreement)?

auk:
Yes, twice already
For project lead Rossen Stoyanchev and you Brian Clozel.

bclozel:
This has been fixed in [SPR-8016](https://jira.spring.io/browse/SPR-8016) with a different approach.

fat-panda:
This is still a problem for us - https://jira.spring.io/browse/SPR-13414
We have a response 202 without a body, and fails because of the `	if (getHeaders().getContentLength() == 0) {` in hasMessageBody 

What are the workaround for this?

According to the RFC https://tools.ietf.org/html/rfc7230#section-3.3.3
 Any 2xx (Successful) response to a CONNECT request implies that
       the connection will become a tunnel immediately after the empty
       line that concludes the header fields.  **A client MUST ignore any
       Content-Length** or Transfer-Encoding header fields received in
       such a message.

Thanks

snicoll:
I can see @jhoeller reopened the related issue. @fat-panda commenting here is not a great way to get help, please comment on the Jira issue next time please.

bclozel:
Hello @fat-panda ,

I'm sorry I don't understand your comment nor why it applies to this issue.
Can you elaborate a bit more on this?

* which spring framework version are you using?
* could you share the HTTP request and responses for this case?
* could you share code snippets or a small application that replicates the issue? (even using https://httpbin.org/ as a remote server)

Please drop a comment on the jira issue itself - it's easier for everybody to track down things on the original issue.

Thanks!

fat-panda:
Sorry didn't realise this was closed when posting previously.
This is no longer an issue - solved by using postForEntity

PR #870, opened 2015-09-05
RequestParam plus MultipartFIle

rocky-peng:
if there is a controller method with code like this:

```
public void uploadFile( @RequestParam(value = "file",required = true) MultipartFile file){
    //some code
}
```

and there is no header which name is "Content-Type" in the request,then the method "resolveName" will throw MultipartException which is from the method "assertIsMultipartRequest" even if the value of attribute which name is "required" is true. 

The expected response should be throw MissingServletRequestParameterException i think. Because it is convenient for coder to execute the same code when there is no enough parameters.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/annotation/RequestParamMethodArgumentResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/method/annotation/RequestParamMethodArgumentResolver.java
@@ -163,35 +163,26 @@ public class RequestParamMethodArgumentResolver extends AbstractNamedValueMethod
 				WebUtils.getNativeRequest(servletRequest, MultipartHttpServletRequest.class);
 		Object arg;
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@rocky-peng Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=870)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=870) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@rocky-peng Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=870)!

PR #874, opened 2015-09-17
Fix Configuration docs

kamilszymanski:


snicoll:
`@Import` is actually accurate. It allows you  to refer to configuration resources (typically XML configuration files).

kamilszymanski:
But it doesn't allow you to inject imported beans.

IMHO the following statement is invalid (the @Import part): 
```
Bean definitions imported from XML can be injected using {@code @Autowired} or {@code @Import}:
```

snicoll:
Right, sorry. More coffee needed over here :)

PR #875, opened 2015-09-18
Fix AllowedOrigins according to RFC 6454

GeorgeCGV:
- Transformation of the origin into uri components used 
  to fail with a `NumberFormatException` when origin string
  contained a character after the port definition like `[port]/`.
  - The slash straight after the port will ruin the integer parsing.
    `DefaultCorsProcessor.processRequest` will fail during call to `WebUtils.isSameOrigin`.
- Origin  validity check was considering 
  'http://domain1.com' and 'http://domain1.com/' to be different. The same applies to allowed origins comparison. This doesn't meet RFC 6454 standard.

> 3.2.1. Examples
> 
> ```
> All of the following resources have the same origin:
> 
> http://example.com/
> http://example.com:80/
> http://example.com/path/file
> ```

**Notes**
1. In order to improve performance the `allowedOrigins` collection can convert provided `Strings` directly to `UriComponents` or parsing for the header `origin` and `allowedOrigins` can be further simplified.
2. I'm not totally sure about `UriComponentsBuilder.fromHttpRequest` but based on the code it may fail the same way as `UriComponentsBuilder fromOriginHeader` did. It depends what value can be stored in `X-Forwarded-Host`.

Issue: Nothing has been reported so far. However, the issue becomes noticeable: [SocketRocket](https://github.com/square/SocketRocket/pull/256)

_I have signed and agree to the terms of the SpringSource Individual
Contributor License Agreement._

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -345,27 +345,26 @@ public class UriComponentsBuilder implements Cloneable {
 	 * @see <a href="https://tools.ietf.org/html/rfc6454">RFC 6454: The Web Origin Concept</a>
 	 */
 	public static UriComponentsBuilder fromOriginHeader(String origin) {
-		UriComponentsBuilder builder = new UriComponentsBuilder();
+		final UriComponentsBuilder builder = new UriComponentsBuilder().scheme(null).host(null).port(null);
 
-		Matcher matcher = URI_PATTERN.matcher(origin);
-		if (matcher.matches()) {
-			String scheme = matcher.group(2);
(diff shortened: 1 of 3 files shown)


rstoyanchev:
Thanks for the pull request. Would you mind creating a ticket in our [JIRA issuer tracker](jira.springsource.org/browse/SPR)?

GeorgeCGV:
@rstoyanchev sure, here we go: [Jira Issue - SPR-13478](https://jira.spring.io/browse/SPR-13478).

sdeleuze:
Resolved by [this commit](https://github.com/spring-projects/spring-framework/commit/9c66dfa7b5f610ddeb69b9dc10baa227fbddcbf1), so I close this PR. Thanks for raising this issue.

PR #879, opened 2015-09-28
Update Value.java

pkaq:
# {systemProperties.myProp} should be ${systemProperties.myProp}


sbrannen:
The documentation is correct as it is.

The example correctly demonstrates how to use `@Value` with a [SpEL expression](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/expressions.html#expressions-beandef-annotation-based).

PR #881, opened 2015-10-06
Add description for two exceptions which handled by DefaultHandlerExceptionResolver in reference document

kazuki43zoo:
- MissingPathVariableException (Available since 4.2)
- NoHandlerFoundException (Available since 4.0)

Please check.


rstoyanchev:
This is merged.

PR #884, opened 2015-10-10
 prevent duplicate "/"

veryyoung:
The request url will remove the duplicate duplicate "/" by using UrlPathHelper.getLookupPathForRequest(request), but the RequestMappingInfo won't.

And "directPathMatches"  will be null at AbstractHandlerMethodMapping.lookupHandlerMethod.

For example, when I request ///c, the lookupPath will be /c, it won't match the HandlerMethod //c


rstoyanchev:
Is there a ticket for this in the JIRA issue tracker? It appears to be a duplicate of https://jira.spring.io/browse/SPR-12975.

PR #886, opened 2015-10-11
Handling null in BufferingClientHttpResponseWrapper

jmnarloch:
BufferingClientHttpResponseWrapper#getBody brakes the contract for ClientHttpResponse and whenever the response body is null (for instance when 204 is being returned) it throws IllegalArgumentException, while any other implementation is returning null.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/BufferingClientHttpResponseWrapper.java
+++ b/spring-web/src/main/java/org/springframework/http/client/BufferingClientHttpResponseWrapper.java
@@ -16,13 +16,13 @@
 
 package org.springframework.http.client;
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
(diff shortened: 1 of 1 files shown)


jhoeller:
I'm picking this up in SPR-13563, clarifying getBody's contract there. I'm wondering whether the null return value is actually intended, since the interface javadoc doesn't mention it, and the only implementation ever returning null seems to be HttpComponents(Async)ClientHttpResponse. All the others seem to return an empty body instead, or am I missing something?

jhoeller:
Resolved through SPR-13563. Please reopen that JIRA issue if there is anything left to be addressed.

jmnarloch:
Cool, thanks!

The IAE was exactly what we were struggling with, but your solution is way more elegant.

Thanks again.

PR #891, opened 2015-10-16
SPR-13579: Support conversion from Strings to Java 8 Durations.

josephw:
Implement conversion from `String`s to `java.time.Duration` instances using `Duration`'s
`parse` method. This isn't the greatest format, but it simplifies wiring up
a bean that uses `Duration` in its API rather than (for example) long for millis.
Includes tests.


cskinfill:
Is this supported in spring now?

bclozel:
Yes, check https://jira.spring.io/browse/SPR-6706

PR #892, opened 2015-10-16
Spring Jdbc add Postgresql ?-contained operator support

arturgspb:
ISSUE: https://jira.spring.io/browse/SPR-13582

Postgresql have a many functional oparators such as "?" (for example jsonb operators http://www.postgresql.org/docs/9.4/static/functions-json.html)

Spring jdbc processing this case not correct. When run sql

```
SELECT '["1","2","3"]'::jsonb ? '4'
```

I see exception like this:

```
class org.springframework.dao.InvalidDataAccessApiUsageException
SQL [SELECT '["1","2","3"]'::jsonb ? '4']: given 1 parameters but expected 0
```

In pgjdbc this bug fixed in 2014 (https://github.com/pgjdbc/pgjdbc/pull/227)

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterUtils.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterUtils.java
@@ -152,6 +152,12 @@ public abstract class NamedParameterUtils {
 					}
 				}
 				if (c == '?') {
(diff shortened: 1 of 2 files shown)


arturgspb:
Changes merged https://github.com/spring-projects/spring-framework/commit/11d1f050e9bb02c70086176b841e6a766f21fb4e

PR #896, opened 2015-10-23
Add null check to enable creation of triggers without JobDetail.

bendiscz:
This change allows SimpleTriggerFactoryBean and CronTriggerFactoryBean to create triggers without associated job details. This is something that we need in our project (we are using spring to configure triggers, but we associate them with jobs programmatically) and it worked fine in Spring 3.x.

Issue: [SPR-13604](https://jira.spring.io/browse/SPR-13604)

License Agreement.


snicoll:
Thanks for the PR: this was fixed as part of https://github.com/spring-projects/spring-framework/commit/29700658d4a6c337105c722f7b02857231c02020 (I see no reason to not set the `jobDataMap` if the details are not set).

PR #900, opened 2015-10-27
merge plugin: fix broken reference to merged projects if gradle daemon used

vladsoroka:
Caching 'attachedProjectsEvaluated' flag using static plugin field may broke reference to merged projects for subsequent build executions if gradle daemon used


snicoll:
Thanks, have you signed the CLA?

vladsoroka:
yep, I've just signed it

PR #904, opened 2015-10-28
Fix CronTrigger example in documentation

LinkedList:
The original example would fire every second of the 15th minute, not
just once as stated in the paragraph before.

Issue: SPR-10474

License Agreement.


jhoeller:
Applied as part of a slightly wider revision. Thanks for raising this!

PR #907, opened 2015-11-01
Add support to any Iterable types on PropertyAccessor

ljtfreitas:
Sometimes, for better clarity of the application model, we create objects that encapsulate a Collection, and this object comes to represent a collection of a certain type in the application model. To simplify `for-each` operations, this object must implement `java.lang.Iterable`. Example:

``` java
public class MyType {
    @NotNull
    private String name;
}

public class MyTypesGroup implements Iterable<MyType> {
    private Collection <MyType> myTypes = new ArrayList <MyType> ();

    @Override
    public Iterator<MyType> iterator() {
         return myTypes.iterator();
    }

    // model methods
}

public class MyAggregateType {
     @Valid
     private MyTypesGroup myTypesGroup MyTypesGroup = new ();
}
```

Databind, get and set properties operations, typically used with `BeanWrapper`, they are not supported for these objects (in the above example, something like "myTypesGroup[0].name"). 

In my use case, with Hibernate Validator, the path will be generated "myTypesGroup[0].name" (Hibernate Validator au), occurring an exception when the Spring tries to extract the property value (to generate the error message on the Spring MVC)

This pull request adds `java.lang.Iterable` as a supported type for reading properties by `PropertyAccessors`.

Thanks (and sorry for any English errors :))

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
@@ -678,7 +678,28 @@ public abstract class AbstractNestablePropertyAccessor extends AbstractPropertyA
 						Object convertedMapKey = convertIfNecessary(null, null, key, mapKeyType, typeDescriptor);
 						value = map.get(convertedMapKey);
 					}
-					else {
+					else if (value instanceof Iterable) {
+						// Apply index to Iterator in case of other Iterable implementation.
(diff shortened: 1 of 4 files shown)


jhoeller:
As part of the wider binding and validation theme in 6.1, I've picked up this one in a custom fashion, unified with our `Set`/`Collection` handling (just doing the size check after the iteration rather than before).

Your pull request was a great starting point there, thanks for your efforts! Sorry for not revisiting it earlier...

PR #912, opened 2015-11-07
SPR-12538 AsyncRestTemplate interceptors

jmnarloch:
Hi,

I would like to propose this change that introduces the interceptors functionality to the AsyncRestTemplate.

The implementation mimics the existing one for RestTemplate though is adjusted to handle ListenableFuture instead.

This is an entry point from which I would like to start and open the discussion. If you see some issues with this implementation just let me know. This change does not maintain the backward compatibility between the RestTemplate interceptors and I don't see a way to achieve that goal.

The motivation behind adding this is the fact that is really convenient to write extensions that automatically plug-in into the AsyncRestTemplate. Spring Cloud/Spring Cloud Sleuth already does that for RestTemplate and enables for instance request tracing.

Related issue: https://jira.spring.io/browse/SPR-12538

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/InterceptingAsyncClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/InterceptingAsyncClientHttpRequest.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2002-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.http.client;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpRequest;
+import org.springframework.util.StreamUtils;
+import org.springframework.util.concurrent.ListenableFuture;
+import org.springframework.util.concurrent.ListenableFutureAdapter;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * A {@link AsyncClientHttpRequest} wrapper that enriches it proceeds the actual request execution with calling
+ * the registered interceptors.
+ *
+ * @author Jakub Narloch
+ * @see InterceptingAsyncClientHttpRequestFactory
+ */
+class InterceptingAsyncClientHttpRequest extends AbstractBufferingAsyncClientHttpRequest {
+
+    private AsyncClientHttpRequestFactory requestFactory;
+
+    private List<AsyncClientHttpRequestInterceptor> interceptors;
+
+    private URI uri;
+
+    private HttpMethod httpMethod;
+
+    /**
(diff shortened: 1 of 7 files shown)


spring-web/src/main/java/org/springframework/http/client/AsyncInterceptingClientHttpRequest.java
```diff
@@ -0,0 +1,99 @@
...
+    private class AsyncRequestExecution implements AsyncClientHttpRequestExecution {
+
+        private Iterator<AsyncClientHttpRequestInterceptor> nextInterceptor = interceptors.iterator();
+
+        @Override
+        public ListenableFuture<ClientHttpResponse> executeAsync(HttpRequest request, byte[] body) throws IOException {
+            if(nextInterceptor.hasNext()) {
+                AsyncClientHttpRequestInterceptor interceptor = nextInterceptor.next();
+                ListenableFuture<ClientHttpResponse> future = interceptor.interceptRequest(request, body, this);
+                return new IdentityListenableFutureAdapter<ClientHttpResponse>(future);
```
jmnarloch: I've added this wrapper to streamline the execution of the callbacks, though LIFO callback execution could still break the ordering, making the client code handler being invoked prior the interceptors callbacks.

jmnarloch:
@jhoeller @rstoyanchev @poutsma may I ask you on your opinion on this topic?

spring-web/src/main/java/org/springframework/http/client/InterceptingAsyncClientHttpRequest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.concurrent.ExecutionException;
+
+/**
+ * A {@link AsyncClientHttpRequest} wrapper that enriches it proceeds the actual request execution with calling
+ * the registered interceptors.
+ *
+ * @author Jakub Narloch
+ * @see InterceptingAsyncClientHttpRequestFactory
+ */
+public class InterceptingAsyncClientHttpRequest extends AbstractBufferingAsyncClientHttpRequest {
```
poutsma: This class does not have to be public, I believe.

spring-web/src/main/java/org/springframework/http/client/InterceptingAsyncClientHttpRequest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        private Iterator<AsyncClientHttpRequestInterceptor> nextInterceptor = interceptors.iterator();
+
+        @Override
+        public ListenableFuture<ClientHttpResponse> executeAsync(HttpRequest request, byte[] body) throws IOException {
+            if (nextInterceptor.hasNext()) {
+                AsyncClientHttpRequestInterceptor interceptor = nextInterceptor.next();
+                ListenableFuture<ClientHttpResponse> future = interceptor.interceptRequest(request, body, this);
+                return new IdentityListenableFutureAdapter<ClientHttpResponse>(future);
+            } else {
```
poutsma: Please refer to [Spring Code Style](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style) to format your code: `else` goes on a new line!

poutsma:
Looks good to me, except for a few minor comments.

jmnarloch:
Thanks, I've addressed all the issues. If you would consider merging this PR, please also take a look at https://github.com/spring-projects/spring-framework/pull/915, there are completely independent, but provide complementary functionality, so that you would be able to intercept a request and for instance log the response payload or do rewriting of the request body.

jmnarloch:
Any luck with this pull request?

dgomesbr:
+1 since it would help spring-cloud/spring-cloud-commons#36

rstoyanchev:
When 4.2.4 goes out this week, we'll be creating a 4.2.x branch and the master branch will be open for 4.3 changes. We'll merge this PR as soon as that's the case.

Thanks.

rstoyanchev:
This is now merged. Thanks! 

Note there were a couple of tweaks in AsyncRequestExecution to make sure it allows decoration of the request.

PR #913, opened 2015-11-07
Avoid char[] allocation when generates cache key

Gordiychuk:
_org.springframework.aop.framework.autoproxy.AbstractAutoProxyCreator#getCacheKey_ generates to many string during work with context and create context, especially when in context present many beans(in my case 1200+) . 

String concatenation leads to two problems:
1) Excessive heap garbage due to throw-away string creation
2) Excessive CPU due to computing hash code for the newly created strings.
It occurs because DefaultListableBeanFactory#doGetBeanNamesForType iterates by all beans in context and try find correspond bean when specified only class and bean name absent(I think it occurs in 80% of cases when uses java configurations).

We can avoid it problem use only beanName bacause they unique in context(exception only bean creates via FactoryBean because in it case beanName for factory and result bean can be same)

Issue: [SPR-13655](https://jira.spring.io/browse/SPR-13655)

I check fix on starting application scenario, you can see results below

**Without with:**
char[] allocation:
spring_allocation_string_by_start

Hot methods:
spring_hot_methods_by_start

**With current fix**
char[] allocation:
spring_allocation_string_by_start_after_fix

Hot methods:
spring_hot_methods_by_start_after_fix

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/autoproxy/AbstractAutoProxyCreator.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/autoproxy/AbstractAutoProxyCreator.java
@@ -42,6 +42,7 @@ import org.springframework.beans.BeansException;
 import org.springframework.beans.PropertyValues;
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
+import org.springframework.beans.factory.FactoryBean;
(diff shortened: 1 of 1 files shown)


Gordiychuk:
I not foud special module where I can put JMH benchmarks for it fix, that why they absent in commit. What do you think about create special module with benchmarks? It module will safe from performance degradation(I plan to show one of these cases in the near future) and also allow demonstrate performance improvements

PR #920, opened 2015-11-13
Consider MANIFEST.MF classpath when scanning

philwebb:
Update PathMatchingResourcePatternResolver to also consider classpaths
provided by MANIFEST.MF when dealing with the System ClassLoader. It
turns out that such entries are not exposed by URLClassLoader.getURLs().

Issue:


jhoeller:
I've rolled a refactored variant of this into master now...

PR #922, opened 2015-11-16
Added quotes around attribute value.

uthark:
There are missing quotes in XML configuration example


uthark:
License Agreement.

snicoll:
Merged, thanks!

PR #924, opened 2015-11-21
TypeDescriptor.equals tests equality of annotations

nathanpiper:
The current implementation of TypeDescriptor.equals merely tests for the presence of annotations between the compared type descriptors, as apposed to also comparing annotation property values.

This ultimately lead to problems with the GenericConversionService cache returning the "wrong" conditional converter in my application.

License Agreement.

Issue: SPR-13714


jhoeller:
Thanks for the pull request! I've rolled this in a slightly revised fashion, including some general rearrangments in the affected tests.

PR #925, opened 2015-11-22
SPR-13713: Add mandatory behavior to SimpleNamespaceContext

leoarnold:
SimpleNamespaceContext did not implement the following behavior
mandated by javax.xml.namespace.NamespaceContext (as per Java 8):
- # getPrefix and #getPrefixes with Argument `null` did not throw
  
  an IllegalArgumentException
- The iterator returned by #getPrefixes(boundPrefix) was modifiable,
  i.e. failed to throw an UnsupportedOperationException upon #remove
- # clear did not clear the map `namespaceUriToPrefixes`, hence after
  # clear an iterator returned by #getPrefixes could be non-empty
- Some mandatory functionality was implemented, but lacked unit tests

There were also code smells:
- # getPrefixesInternal used to modify the field `namespaceUriToPrefixes`
  
  despite being a getter method
- After removing all prefixes bound to a namespace URI, the map
  `namespaceUriToPrefixes` still contained the URI as key mapped to
  an empty list.

The unit testing class was extensively rewritten to ensure implementation
of the full contract. All assertions now use Hamcrest matchers for
type safety and better error messages in case of failure.

Changes to the class SimpleNamespaceContext:
- All missing functionality (described above) was added
- # getPrefixesInternal is now a clean accessor
- The field `namespaceUriToPrefixes` is now mutated via new private
  methods #addPrefixToNamespaceUri and #removePrefixFromNamespaceUri
- After removing all prefixes bound to a namespace URI, the URI is also
  removed (as key) from the map `namespaceUriToPrefixes`
- The private method #getPrefixesInternal was renamed to
  # getPrefixesList as a more telling name

Issue: SPR-13713

License Agreement.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/xml/SimpleNamespaceContext.java
+++ b/spring-core/src/main/java/org/springframework/util/xml/SimpleNamespaceContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2012 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


jhoeller:
Rolled into master as part of a wider rewrite now, addressing all your concerns and reusing your unit tests. Thanks for the pull request!

PR #926, opened 2015-11-22
Modify jsp tag library list in reference document

kazuki43zoo:
I've fixed a some mistakes of [SPR-13520](https://jira.spring.io/browse/SPR-13520).


snicoll:
Merged, thanks!

PR #929, opened 2015-12-02
SPR-13752: MockHttpServletResponse.setIntHeader Content-Length

bedrin:
[SPR-13752: MockHttpServletResponse.setIntHeader does not support 'Content-Length' header](https://jira.spring.io/browse/SPR-13752)

License Agreement.


jhoeller:
Rolled in as part of a slightly wider revision. Thanks for raising this!

PR #932, opened 2015-12-05
Fix JavaDoc of @EventListener

kazuki43zoo:
I've fixed the [SPR-13763](https://jira.spring.io/browse/SPR-13763).


jhoeller:
Applied as part of slightly wider javadoc polishing. Thanks for spotting this!

PR #934, opened 2015-12-10
Format-string method called with 2 arguments but is given 3.

mnhock:
License Agreement.


snicoll:
Merged, thanks!

PR #937, opened 2015-12-18
improved content length exception message

deki:
In case ShallowEtagHeaderFilter is not used, the message was confusing.


bclozel:
Merged with 554bf4958de

Thanks!

PR #939, opened 2015-12-20
Fix markup in reference section 9.4.2

sniewczas:
Trivial markup fix in the 'Built-in PropertyEditor implementations' section of core-validation.

Before: 
before

After: 
after


snicoll:
Merged, thanks!

PR #947, opened 2016-01-02
Rejected async task on spring-web will now end instead of timeout.

icha024:
Handles RejectedExecutionException on spring-web async call, now
asyncWebRequest end and return immediately instead of waiting until
timeout.

Closing the request immediately on error to avoid request time-out triggering a second exception.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/context/request/async/WebAsyncManager.java
+++ b/spring-web/src/main/java/org/springframework/web/context/request/async/WebAsyncManager.java
@@ -21,6 +21,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Thanks for the PR. Could you please [create a ticket](jira.springsource.org/browse/SPR). Have you already signed the [contributor agreement](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#sign-the-individual-contributor-license-agreement-icla)?

icha024:
https://jira.spring.io/browse/SPR-13836

License Agreement.

rstoyanchev:
Merged, thanks! 

Note that the PR was rebased, slightly modified to comply with conventions, and squashed so you'll have to reset your master branch. In general it's better to submit from a branch since it's very likely your PR will be rebased if nothing else, so you can just remove the branch when the PR is accepted.

PR #949, opened 2016-01-05
Use the official SonarQube Gradle plugin

bclozel:
Since the build has been upgraded to Gradle 2.10, it's showing the
following message:
"The 'sonar-runner' plugin has been deprecated and is scheduled to be
removed in Gradle 3.0. please use the official plugin from SonarQube"

This commit makes use of the "org.sonarqube" Gradle plugin and uprgrades
the Jacoco agent version to 0.7.5.

_Note_: Since Jacoco 0.7.5 changed its binary format, **it is required to
upgrade the Java plugin** on our SonarQube instance.

See: https://sonar.spring.io/updatecenter/updates
See: https://jira.sonarsource.com/browse/SONARJAVA-1091

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,10 @@ buildscript {
 	}
 }
 
+plugins {
+	id "org.sonarqube" version "1.1"
+}
+
 ext {
 	linkHomepage = 'https://projects.spring.io/spring-framework'
 	linkCi = 'https://build.spring.io/browse/SPR'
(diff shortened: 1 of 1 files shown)


bclozel:
Hey @snicoll, I've tested this successfully on my laptop.

We have to make sure that our "Java" SonarQube plugin is up to date.
If not, an update is required (and probably a restart of the instance).

PR #953, opened 2016-01-12
Add PDF MediaType and Value

mjgpy3:
License Agreement.


PeterSt66:
Have a look at MediaTypeUtils which is used during MediaType constructing, you might need to add it there too.

rstoyanchev:
Thanks for the PR but this was already added https://jira.spring.io/browse/SPR-13831.

PR #955, opened 2016-01-19
[SPR-13876] - fix ServletServerHttpRequest URI encoding issue, provid…

sammyhk:
…e test case and fix test cases related


rstoyanchev:
This is now merged, thanks for the pull request!

PR #956, opened 2016-01-24
Fix AntPathMatcher.isPattern for URI templates

m5k10:
I'm not sure but I think AntPathMatcher.isPattern("/{bla}.jpg") should return true.


bclozel:
Doing that change fails the following test, so this raises more question about AntPathMatcher behavior:

```
org.springframework.core.io.support.ResourceArrayPropertyEditorTests > testSystemPropertyReplacement
```

Could you create an issue on jira.spring.io for this?

soldierkam:
Done https://jira.spring.io/browse/SPR-14763

bclozel:
Fixed with #22959

PR #958, opened 2016-01-25
Speed up AntPathMatcher by doing prefix-match

m5k10:
Many ant style patterns has prefix that can be used to determine if there is no need to do complicated and costly matching. 
For example pattern "/static/css/*" have prefix "/static/css/" that can be used to say that path "/api/resource/1234" will not match. 

This was tested with https://gist.github.com/soldierkam/c80e433aa9fbec35df5a
Before:

| Benchmark | Mode | Cnt | Score | Error | Units |
| --- | --- | --- | --- | --- | --- |
| MyBenchmark.isMatching | thrpt | 5 | 425995,288 ± | 162966,573 | ops/s |
| MyBenchmark.isNotMatching | thrpt | 5 | 864690,507 ± | 130643,210 | ops/s |
| MyBenchmark.suffix | thrpt | 5 | 748551,270 ± | 17229,496 | ops/s |

After:

| Benchmark | Mode | Cnt | Score | Error | Units |
| --- | --- | --- | --- | --- | --- |
| MyBenchmark.isMatching | thrpt | 5 | 483815,001  ± | 105669,816 | ops/s |
| MyBenchmark.isNotMatching | thrpt | 5 | 10208718,438 ± | 125879,376 | ops/s |
| MyBenchmark.suffix | thrpt | 5 | 624841,454  ± | 349001,674 | ops/s |

Result: over 10x faster for isNotMatching and 17% slower for "suffix" case.

I tested it with https://github.com/spring-projects/spring-boot/tree/master/spring-boot-samples/spring-boot-sample-data-rest (wrk -c 40 -d 30 -t 2 http://127.0.0.1:8080/api/hotels) but results are almost the same (before: 310.15req/seq, after: 305.41req/sec).
For https://github.com/spring-projects/spring-boot/tree/master/spring-boot-samples/spring-boot-sample-actuator-ui (wrk -c 40 -d 30 -t 2 http://localhost:8080/css/bootstrap.min.css) there is small perf improvement - before: 2237.90req/seq, after: 2353.93req/sec
For real world application (that contains many more endpoints) I've got 17% more throughput for endpoint serving static resource (baseline is at 1812req/sec).

Let me know what you think.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java
+++ b/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java
@@ -85,7 +85,7 @@ public class AntPathMatcher implements PathMatcher {
 
 	private volatile Boolean cachePatterns;
 
-	private final Map<String, String[]> tokenizedPatternCache = new ConcurrentHashMap<String, String[]>(256);
+	private final Map<String, PreprocessedPattern> tokenizedPatternCache = new ConcurrentHashMap<String, PreprocessedPattern>(256);
 
 	final Map<String, AntPathStringMatcher> stringMatcherCache = new ConcurrentHashMap<String, AntPathStringMatcher>(256);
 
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks @soldierkam for this PR! I'd be happy to review more if you've got other ideas :+1: 

Indeed, I ran a few JMH benchmarks myself and found that this leads to a consistent performance boost for many cases. `AntPathMatcher` is a central piece of our infrastructure, so I'm not keen on backporting this change to the latest version of the 4.2.x line. This will be shortly available as a `4.3.0-BUILD-SNAPSHOT` version.

Merged as e77ff3c for [SPR-13913](https://jira.spring.io/browse/SPR-13913)

PR #959, opened 2016-02-08
DefaultListableBeanFactory getBeansOfType initialize beans when includeNonSingletons is false

sewolf:
Hi,

I have a question:  
in DefaultListableBeanFactory the method call getBeansOfType(MyClass.class, false, true) will initialize lazy singleton beans, but the call getBeansOfType(class, true, true) not. 
Is this a correct behaviour?

Regards,
Sergej


pivotal-issuemaster:
@sewolf Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=959)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=959) to manually synchronize the status of this Pull Request.

snicoll:
@sewolf sorry it took so long to get back to you but a question disguised in a PR is a bit unusual. What you've described is not the correct behavior (the second call should also initialize lazy-init singletons). I can't know for sure as your test reuses the same bean factory instance. If you're experiencing an issue, we'd rather get a description of it with a small sample that reproduces the problem.

PR #962, opened 2016-02-14
404 rather than 405 or 200

matthew-pearson:
Check that the path is valid and resolvable before checking that the http method is supported. For invalid or unresolvable paths, always respond with a 404.
Maybe OPTIONS should have 200 status, but I think the others would be better as 404.
I've raised this as a bug on [Jira](https://jira.spring.io/browse/SPR-13944)


rstoyanchev:
Merge. Note that I've rebased on top of the latest so you local fork will need to be reset with --hard. In general it's a good idea to use a branch for pull requests that you can then simply delete.

PR #965, opened 2016-02-17
Distinguish between unsatisfied constructor and factory method deps

wilkinsona:
As discussed with @snicoll, and in support of Spring Boot 1.4's goal of providing improved diagnostics when application context refresh fails, this PR introduces specific exceptions for unsatisfied factory method dependencies and unsatisfied constructor dependencies.


snicoll:
@jhoeller has fixed this one already.

PR #969, opened 2016-02-18
Introduce specific exceptions for field and method autowiring failures

wilkinsona:
Previously, a BeanCreationException was used when autowiring of a field or method failed. Details of the field or method were included in the exception message. This meant that anyone who wanted to access
information about the field or method had to resort to the error-prone approach of parsing the exception’s message.

This pull request introduces two new exceptions, FieldAutowiringException and MethodAutowiringException, that provide access to the Field or Method for which autowiring has failed.


snicoll:
@jhoeller I think that PR is outdated now, right?

OnlyRebbits:
Maybe 


发自 网易邮箱大师
在2016年03月05日 05:45，Stéphane Nicoll 写道:

@jhoeller I think that PR is outdated now, right?

PR #970, opened 2016-02-18
Consistent descriptions for ConfigurationClass

wilkinsona:
Previously, a ConfigurationClass created from AnnotationMetadata
and a ConfigurationClass created from a class would have subtly
different descriptions. Given a class named com.example.Foo, the
former’s description would be “com.example.Foo”, whereas the latter’s
description would be “class com.example.Foo”.

This commit updates ConfigurationClass to make the description
consistent, preferring the description without “class” in it.


snicoll:
Thank you!

PR #972, opened 2016-02-21
Support for @ExceptionHandler exclusions

LinkedList:
Added support for excluding an exception from being handled in a method
annotated with @ExceptionHandler.

Issue: SPR-13932

License Agreement.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/bind/annotation/ExceptionHandler.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/annotation/ExceptionHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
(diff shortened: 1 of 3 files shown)


spring-web/src/main/java/org/springframework/web/bind/annotation/ExceptionHandler.java
```diff
@@ -129,4 +130,9 @@
 	 */
 	Class<? extends Throwable>[] value() default {};
 
+	/**
+	 * Exceptions excluded from handling by the annotated method.
+	 * @since 5.0
```
PascalSchumacher: I guess this should be 4.3?

LinkedList: I don't know. 
The Jira issue has Fix Version as '5.0 Backlog' so I thought it's planned for 5.0.

PR #974, opened 2016-02-22
`are are` might be `are all`?

Youmoo:
License Agreement.


snicoll:
No, I think `are` is just mentionne twice. I've fixed that in https://github.com/spring-projects/spring-framework/commit/a94ffbb17a7525a3e3fa0e3c377e8bdafac0e331 - Thanks for the PR!

PR #980, opened 2016-02-26
Remove @BootstrapWith from @WebAppConfiguration

philwebb:
Update @WebAppConfiguration so that it no longer directly specifies
a TestContextBootstrapper. This allows third parties to use the
annotation in combination with their own bootstrapper.

BootstrapUtils now provides the detection logic for when
WebTestContextBootstrapper should be used.

Issue: SPR-13991


snicoll:
This issue has been fixed so I am closing this PR.

PR #981, opened 2016-02-27
Introduce composed annotations for @RequestMapping

sbrannen:
This commit introduces the following common composed annotations for
@RequestMapping in Spring MVC and Spring MVC REST.
- @GetMapping
- @PostMapping
- @PutMapping
- @DeleteMapping
- @PatchMapping

Issue: SPR-13992

### DIFF
--- a/spring-test/src/test/java/org/springframework/test/web/servlet/samples/context/AsyncControllerJavaConfigTests.java
+++ b/spring-test/src/test/java/org/springframework/test/web/servlet/samples/context/AsyncControllerJavaConfigTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


sdeleuze:
:+1:

rstoyanchev:
Looks good overall. I'm wondering if we shouldn't also drop `produces` on POST, PUT, and PATCH? Also drop both `consumes` and `produces` on DELETE? It's less common to produce on those methods to begin with plus those conditions are for disambiguation purposes. They are not always needed, and you still have consumes.

sbrannen:
@rstoyanchev, thanks for the review!

Those are exactly the topics I wanted to discuss with you: produces vs. consumes.

It's a bit of a challenge getting it right and not being too restrictive.

Regarding `produces` for POST, PUT, and PATCH, some people actually return the new/updated resource in the body in order to avoid having to submit a followup GET based on the 'location' header. So I thought we would want to allow such use cases. The same holds true for DELETE, though perhaps that's not so common.

Regarding `consumes` for DELETE, I also pondered whether we should include that. I doubt many people would submit a body with a DELETE request, but then again... I just didn't want to be too restrictive.

Let's come to a consensus during the team call -- OK?

PR #982, opened 2016-02-27
Introduce composed annotations for web scopes

sbrannen:
This commit introduces the following common composed annotations for
web scopes.
- @RequestScope
- @SessionScope
- @ApplicationScope

Issue: SPR-13993


gognamunish:
Hi sbrannen,

Are there any plans for @ViewScoped as in JSF?

sbrannen:
@gognamunish,

No, we are not considering an `@ViewScope` annotation for JSF, not because we don't want to but rather because there is no straightforward way for Spring to tie into that.

PR #984, opened 2016-03-01
Support *Aware for @ComponentScan custom filters

philwebb:
Support a limited set of *Aware interfaces for TypeFilters created
via the `@ComponentScan` annotation.

Issue: SPR-14009


snicoll:
Looks like this is an outdated proposal that got included in the meantime, see https://github.com/spring-projects/spring-framework/issues/18581#issuecomment-453436952

PR #987, opened 2016-03-05
fix the SPR-14022

lqjack:


rstoyanchev:
Closing along with related ticket.

PR #989, opened 2016-03-05
Upgrade to OkHttp 2.7.5

royclarkson:
Upgrade dependency to the latest OkHttp 2 version


bclozel:
Taken care of with 406d06118
Thanks!

PR #990, opened 2016-03-06
fix convert BigDecimal to Integer/Short/Byte maynot overflow

anguo-wenz:
fix NumberUtils throws Exception when convert a BigDecimal to Long but sucess when convert to Integer
http://stackoverflow.com/questions/35822941/numberutils-throws-exception-when-convert-a-bigdecimal-to-long-but-sucess-when-c


jhoeller:
Could please create an SPR issue on our JIRA, referring to this pull request? We aim to address such consistency issues in the 4.3 timeline, so this one is certainly a candidate there...

andyjojo:
Hi Jhoeller

issue created
https://jira.spring.io/browse/SPR-14041

Thanks.

PR #991, opened 2016-03-08
Add ContextCustomizer support to spring-test

philwebb:
Allow libraries to contribute ContextCustomizers that can update
ApplicationContexts create by spring-test before they are run.

A Customizer may be provided via a ContextCustomizerFactory which is
registered with `spring.factories`. Each factory is consulted whenever
a new ApplicationContext needs to be created by spring-test. Factories
may inspect various details about the test and either return a new
ContextCustomizers or `null`.

ContextCustomizers are similar to ApplicationContextInitializers and
may perform any number of tasks, including bean registration.

Issue: SPR-13998

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/support/AbstractContextLoader.java
+++ b/spring-test/src/main/java/org/springframework/test/context/support/AbstractContextLoader.java
@@ -28,11 +28,13 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextException;
 import org.springframework.context.ApplicationContextInitializer;
 import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.support.GenericApplicationContext;
 import org.springframework.core.GenericTypeResolver;
 import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 import org.springframework.core.env.PropertySource;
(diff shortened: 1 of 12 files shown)


spring-test/src/main/java/org/springframework/test/context/support/AbstractContextLoader.java
```diff
@@ -132,6 +135,9 @@ protected void prepareContext(ConfigurableApplicationContext context, MergedCont
 		TestPropertySourceUtils.addPropertiesFilesToEnvironment(context, mergedConfig.getPropertySourceLocations());
 		TestPropertySourceUtils.addInlinedPropertiesToEnvironment(context, mergedConfig.getPropertySourceProperties());
 		invokeApplicationContextInitializers(context, mergedConfig);
+		for (ContextCustomizer contextCustomizer : mergedConfig.getContextCustomizers()) {
+			contextCustomizer.prepareContext(context, mergedConfig);
+		}
```
sbrannen: I'm not very comfortable having a `ContextCustomizer` being called from a `prepareContext()` method when a `customizeContext()` method exists as well.

sbrannen:
Merged into `master` in https://github.com/spring-projects/spring-framework/commit/beb01e674e33d6302390acd8b6ec39963f3f80e1

PR #994, opened 2016-03-09
delegate PayloadApplicationEvent type resolution to the PayloadApplicationEvent itself

radai-rosenblatt:
this fix allows writing custom application events that implement ResolvableTypeProvider, as an alternative to extending ApplicationEvent. see for example this custom event implementation - https://github.com/radai-rosenblatt/spring-events/blob/master/src/main/java/net/radai/ContentsChangedEvent.java

Signed-off-by: radai-rosenblatt radai.rosenblatt@gmail.com

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
+++ b/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
@@ -369,7 +369,7 @@ public abstract class AbstractApplicationContext extends DefaultResourceLoader
 		else {
 			applicationEvent = new PayloadApplicationEvent<Object>(this, event);
 			if (eventType == null) {
-				eventType = ResolvableType.forClassWithGenerics(PayloadApplicationEvent.class, event.getClass());
+				eventType = ((PayloadApplicationEvent)applicationEvent).getResolvableType();
 			}
 		}
 


snicoll:
I am actually confused that it does not work right now. I'll have a look.

radai-rosenblatt:
the project i linked to can be made to compile against either the latest release or a dev build (in pom properties). the test fails vs the latest release, but passes after this fix is applied.

see this SO question for background - http://stackoverflow.com/questions/35883022/spring-generic-application-event-failing-to-reach-destination

snicoll:
That was a note to acknowledge the pull request. Something must be wrong in the existing tests. Thanks for the PR, have you signed the CLA?

snicoll:
I've actually fix that in https://github.com/spring-projects/spring-framework/commit/8e24a4153c83fb4ba3d8c8a24503c2a6adf35e58 with a review of the current tests suite. I thought at test was exercising this scenario but that wasn't the case. Thanks!

radai-rosenblatt:
thanks for the fix. for future reference - "the CLA" ?

snicoll:
https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#sign-the-individual-contributor-license-agreement-icla

PR #995, opened 2016-03-10
wrapping to 90 characters in spel root directory files

cristiancota:


snicoll:
Sorry but there is no reason to strictly wrap at 90 chars. The current code looks of from my perspective. Thanks for the PR anyway.

cristiancota:
I've done it because of this: https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#mind-the-whitespace

And thanks for the replay! :+1:

PR #996, opened 2016-03-10
Support servlet listener activate attribute

philwebb:
Add a ServletTestExecutionListener.ACTIVATE_LISTENER attribute which
can be set on the TestContext to trigger activation of the listener
even if a `@WebAppConfiguration` is not present.

Issue: SPR-14035

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/web/ServletTestExecutionListener.java
+++ b/spring-test/src/main/java/org/springframework/test/context/web/ServletTestExecutionListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-test/src/main/java/org/springframework/test/context/web/ServletTestExecutionListener.java
```diff
@@ -98,6 +98,16 @@
 	public static final String CREATED_BY_THE_TESTCONTEXT_FRAMEWORK = Conventions.getQualifiedAttributeName(
 		ServletTestExecutionListener.class, "createdByTheTestContextFramework");
 
+	/**
+	 * Attribute name for a {@link TestContext} attribute which indicates that
+	 * that the {@link ServletTestExecutionListener} should be activated. When
+	 * not specified activation occurs when the {@linkplain TestContext#getTestClass()
+	 * test class} is annotated with {@link WebAppConfiguration @WebAppConfiguration}.
+	 * @since 4.4
```
kazuki43zoo: 4.4 -> 4.3 ?

sbrannen: Yes, that will likely be 4.3. Good catch.

sbrannen:
This has been merged into `master`: https://github.com/spring-projects/spring-framework/commit/8222ff465deea4c93d1e5bfd2b9a12f9c05099dc

Cheers!

PR #998, opened 2016-03-11
Add converters between Enum and Integer

quaff:
see SPR-14039


snicoll:
Merged with a polish commit in f422e49

PR #1003, opened 2016-03-12
Recognize string interning feature

marschall:
We use the validation support in Spring WS to validate SOAP requests.
Profiling has shown that we have a lot of caught
SAXNotRecognizedException during normal program execution. They come
from AbstractXMLReader.getFeature which is called from
ValidatorHandlerImpl#validate. ValidatorHandlerImpl checks for the
value of the string interning feature. Throwing a
SAXNotRecognizedException is interpreted the same way as returning
false.

Recognizing the feature and returning false preserves the semantics but
gets rid of the SAXNotRecognizedException during normal control flow.
- recognize the "http://xml.org/sax/features/string-interning" feature
  as disabled

Issue: SPR-14047


marschall:
License Agreement.

marschall:
@jhoeller did a much better fix

PR #1006, opened 2016-03-14
SPR-14050 - ReflectionTestUtils doesn't detect proxy

mdeinum:
This commit adds proxy detection for the set-/getField methods
on the ReflectionTestUtils. Instead of accessing the field on the
passed in targetObject the AopTestUtils class is used to obtain the
possible ultimateTargetObject and use that for accessing the
fields.

I've signed the SpringSource ICLA...


sbrannen:
Resolved in 9439a008d7a4ec7359fc024fce60453995f2ebd1.

PR #1008, opened 2016-03-17
Add a way to disable SSL verification

selkhateeb:
This is similiar to curl --insecure (-k) where sometimes you don't want ssl verification.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
@@ -18,11 +18,21 @@ package org.springframework.web.client;
 
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.net.HttpURLConnection;
 import java.net.URI;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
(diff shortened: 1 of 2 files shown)


bclozel:
Hi @selkhateeb - looks interesting!
Could you create an issue on jira.spring.io for this? (project Spring Framework, component "Web:client")?

Also, I think adding this directly on the RestTemplate won't fit with all our supported clients (netty, okhttp, etc). This could be done only on the `SimpleClientHttpRequestFactory`, as there are already some configurations facilities.

pivotal-issuemaster:
@selkhateeb Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1008)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1008) to manually synchronize the status of this Pull Request.

ahrytsiuk:
Hello @selkhateeb, @bclozel as I see this issue is not resolved yet, and this PR looks abandoned. I can propose my help. 

@bclozel can you please guide me? Should I create a Jira ticket first?

bclozel:
Hi @ahrytsiuk 

After checking this PR once more, it doesn't seem to be the right approach.

First, `RestTemplate` is now in maintenance mode so we're not adding new features to it. Also, this is a feature that really belongs at the request factory level - and this really depends on the underlying client library. You should be able to achieve the same by configuring an SSL context on the client/factory that fits your need.

As a result, I'm closing this PR.

ahrytsiuk:
@bclozel I see your concern. 

Maybe it's a good idea to add some autoconfiguration into Spring Boot project which can handle this functionality depends on underlying HTTP library?

bclozel:
I think you can already configure that on the client using the `RestTemplateBuilder`.

PR #1009, opened 2016-03-17
MockMvc HtmlUnit support shares CookieManager

rwinch:
Previously MockMvc builders failed to share the WebConnection used for
managing cookies in the MockMvcWebConnection. This meant that the various
CookieManagers would have different states.

This commit ensures that the WebConnection is set on the
MockMvcWebConnection.

Fixes SPR-14066


rstoyanchev:
@rwinch thanks for the quick pull request. I'm wondering should we also update the constructors of MockMvcWebConnection to the effect that a WebClient is always expected? Presumably a MockMvcWebConnection is created for use with a WebClient and it makes sense to expect the reference to that client vs creating a new instance?

rwinch:
@rstoyanchev Thanks for the feedback. You are right it would be valuable to require `WebClient` in the constructor. I have updated the PR with the changes.

rstoyanchev:
This is now merged.

PR #1011, opened 2016-03-18
Support HTTP range requests in Controllers

bclozel:
Prior to this commit, HTTP Range requests were only supported by the
ResourceHttpRequestHandler when serving static resources.

This commit adds a new ResourceMethodProcessor that both resolves
Resource-related Controller handler arguments and return types.

A Controller handler like this can now handle Range requests for
return types that extend Resource or HttpEntity<Resource>:

```
@RequestMapping("/example/video.mp4")
public Resource handler() { }
```

Issue: SPR-13834


bclozel:
After a chat with @rstoyanchev, I'll try to address that in a different way and regroup that support into the `ResourceHttpMessageConverter`.

bclozel:
@rstoyanchev @poutsma I've just updated this PR. Let me know what you think.

rstoyanchev:
Regarding the dead code, yes we should be using the resolved media types, as the Javadoc above says: 
> then expands to check against the request via ContentNegotiationManager#resolveMediaTypes

Otherwise overall I like the outcome of the current approach.

PR #1012, opened 2016-03-18
Be explicit about the time unit in WebRequest

malkusch:
I just had to dig into the implementation to understand why `WebRequest.checkNotModified()` did not provide the expected behaviour with seconds. This documentation update is more explicit about the time unit.


malkusch:
License Agreement.

bclozel:
Merged with 160410c148 - thanks!

PR #1015, opened 2016-03-23
Changed declaration order of some variables in GenericTypeResolver.resolveReturnTypeForGenericMethod()

dreis2211:
Changed the order to don't declare variables that might not be needed. If declaredTypeVariables.length == 0 is true, the method returns early.


dreis2211:

PR #1016, opened 2016-03-25
Polish MimeTypeUtil to handle only delimiter.

tmyjoe:
I've encounter `java.lang.ArrayIndexOutOfBoundsException:` when only ";" passed in Accept Header and found `MimeTypeUtils` throws the exception.

It think it's an unexpected behavior, so added a logic to throw the dedicated exception :)

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
@@ -219,6 +219,10 @@ public abstract class MimeTypeUtils {
 		if (!StringUtils.hasLength(mimeType)) {
 			throw new InvalidMimeTypeException(mimeType, "'mimeType' must not be empty");
 		}
+		if (mimeType.startsWith(";")) {
+			throw new InvalidMimeTypeException(mimeType, "does not contain any type");
+		}
+
 		String[] parts = StringUtils.tokenizeToStringArray(mimeType, ";");
 
 		String fullType = parts[0].trim();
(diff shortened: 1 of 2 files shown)


tmyjoe:
License Agreement.

rstoyanchev:
Would you mind also creating a [ticket in JIRA](jira.springsource.org/browse/SPR)?

tmyjoe:
I've created a ticket for this.
https://jira.spring.io/browse/SPR-14124

spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
```diff
@@ -219,6 +219,10 @@ public static MimeType parseMimeType(String mimeType) {
 		if (!StringUtils.hasLength(mimeType)) {
 			throw new InvalidMimeTypeException(mimeType, "'mimeType' must not be empty");
 		}
+		if (mimeType.startsWith(";")) {
```
ryanskow:
```
if (parts.length == 0) {
            throw new InvalidMimeTypeException(mimeType, "'mimeType' must not be empty");
}
```

spring-core/src/test/java/org/springframework/util/MimeTypeTests.java
```diff
@@ -165,6 +165,11 @@ public void parseMimeType() throws Exception {
 	}
 
 	@Test(expected = InvalidMimeTypeException.class)
+	public void parseMimeTypeNoType() {
+		MimeTypeUtils.parseMimeType(";");
```
ryanskow: Leading whitespace still breaks:
`MimeTypeUtils.parseMimeType(" ;");`

rstoyanchev:
Thanks for the PR. I've implemented as suggested by @ryanskow since that addresses the exact issue with the ArrayIndexOutOfBoundsException and more conclusively.

PR #1017, opened 2016-03-25
Improve HTTP support for JDK client and HttpInvoker server

bclozel:
Two commits that improve HTTP support when using the native JDK HTTP architecture.

The first commit prevents the RestTemplate from closing the underlying HTTP connection when closing an HTTP response and completely drains the content of the response. Those are the two conditions to really leverage HTTP persistent connections support in HTTP 1.1.

The second commit works around a JDK API that seems to artificially prevent gathered writes when using the HttpInvoker server implementation.

Issue: SPR-14040

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StreamUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StreamUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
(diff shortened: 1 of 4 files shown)


hadrienk:
This render InputStreamResource created by ResourceHttpMessageConverter unusable.

bclozel:
@hadrienk Could you create a new issue on jira.spring.io and reference `SPR-14040`?
Can you also describe how that change makes that `InputStreamResource` unusable with a code sample?

Thanks!

PR #1018, opened 2016-03-25
Remove unnecessary condition in ResourceHttpRequestHandler.handleRequest()

dreis2211:
Hey,

just stumbled upon this. `Resource.contentLength()` returns a primitive long, so the condition

```
Long contentLength = resource.contentLength();
if (contentLength != null) {
    // Something
}
```

seems to be unnecessary because it will be always true.

Keep up the good work!
Cheers


dreis2211:

jhoeller:
Coincidentally, this got fixed as part of a wider revision yesterday. Good catch in any case :-)

PR #1019, opened 2016-03-27
Mocked Request queryString

acc15:
Fixes mocked Request queryString.


rstoyanchev:
Normally you need to [open a ticket|jira.springsource.org/browse/SPR] to explain the issue. 

In the Servlet API a request parameter may be a query parameter or also a name-value pair from the body of an HTTP POST. If you want query string parameters, you need to add them to the target URL when creating a request builder and that gives you both a query string and Servlet API request parameters. If you use `addParam` that will translate to request parameters only and we can't assume if they are query string parameters or not.

PR #1020, opened 2016-03-30
Prevent methods of GroovyObject from becoming transactional

laszlocsontos:
As Jürgen pointed out, we exclude GroovyObject methods for a couple of other
TransactionAttributeSource implementations already, but not for
AnnotationTransactionAttributeSource since we expected it to be clearly
annotation-driven.

Yet, if @Transactional is applied at the class-level, in this case Groovy methods will be considered transactional.
- Added an integration test to reproduce the issue.
- Added exclusion to method computeTransactionAttribute() in AbstractFallbackTransactionAttributeSource, this way, the fact that a method should be excluded can be cached.

Issue: SPR-14095

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -993,6 +993,7 @@ project("spring-webmvc-portlet") {
 
 project("spring-test") {
 	description = "Spring TestContext Framework"
+	apply plugin: "groovy"
 
 	dependencies {
 		compile(project(":spring-core"))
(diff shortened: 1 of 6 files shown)


jhoeller:
As outlined in JIRA, I've opted for a slightly different approach. Your pull request was nevertheless very helpful; thanks for taking the time!

PR #1021, opened 2016-03-30
Backport SPR-14095 to 4.2.x

laszlocsontos:
AnnotationTransactionAttributeSource applies class-level metadata to user-level methods only

Issue: SPR-14095
(cherry picked from commit b7819e6)

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -552,6 +552,7 @@ project("spring-tx") {
 		optional("com.ibm.websphere:uow:6.0.2.17")
 		testCompile("org.aspectj:aspectjweaver:${aspectjVersion}")
 		testCompile("org.eclipse.persistence:javax.persistence:2.0.0")
+		testCompile("org.codehaus.groovy:groovy-all:${groovyVersion}")
 	}
 }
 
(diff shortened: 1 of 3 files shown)


jhoeller:
I meant to address this for 4.3 only since it subtly changes the applicability of class-level metadata not only for Groovy objects but also for other kinds of synthetic methods. This can be hard to debug if somebody happens to be affected by an unexpected case, so I'd rather avoid a backport here. Method-level metadata works fine on 4.2.x even with Groovy objects, after all.

laszlocsontos:
Thanks for the explanation! Basically, I'd like to get rid of that nasty workaround (see _GroovyAwareAnnotationTransactionAttributeSource_ on [SPR-14095](https://jira.spring.io/browse/SPR-14095)) I have temporarily, as it adds some unwanted complexity (having to replace _tx:annotation-driven_ and configure TX handling manually).

> it subtly changes the applicability of class-level metadata not only for Groovy objects but also for other kinds of synthetic methods. 

Could we possibly introduce a flag and keep this functionality switched off by default?

> Method-level metadata works fine on 4.2.x even with Groovy objects, after all.

Do you mean that SPR-14095 doesn't affect 4.2.x, but only 4.1.x and 4.3.x? If I remove your changes from _AbstractFallbackTransactionAttributeSource_, test method _transactionAttributeDeclaredOnGroovyClass()_ in _AnnotationTransactionAttributeSourceTests_ fails.

jhoeller:
I meant that the use of the Transactional annotation at the method level works fine even with Groovy objects. It's just class-level sharing of transaction metadata across all methods which doesn't work as expected since the convention over-applies it even to GroovyObject methods.

I'm not a great fan of flags for such purposes. This is a typical example where the change does make sense by default; it would just be good to let it go through a release candidate phase. We have 4.3 RC1 coming up along with 4.2.6 next week, so this is by no means far away...

snicoll:
Alright, let's close this one and keep the change for 4.3 only. RC1 is out by the way...

PR #1026, opened 2016-04-04
Fix iso date time format

Mischi:
According to https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html the ISO 8601 time zone letter is `X` instead of `Z`. `Z` stands for RFC 822 time zone.

With this, we are able to parse dates which are created in JavaScript using `new Date().toISOString()`.

CC: @davidreher

### DIFF
--- a/spring-context/src/main/java/org/springframework/format/datetime/DateFormatter.java
+++ b/spring-context/src/main/java/org/springframework/format/datetime/DateFormatter.java
@@ -48,7 +48,7 @@ public class DateFormatter implements Formatter<Date> {
 	static {
 		Map<ISO, String> formats = new HashMap<DateTimeFormat.ISO, String>(4);
 		formats.put(ISO.DATE, "yyyy-MM-dd");
-		formats.put(ISO.TIME, "HH:mm:ss.SSSZ");
+		formats.put(ISO.TIME, "HH:mm:ss.SSSX");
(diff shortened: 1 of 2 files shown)


cswzxly:
where is the JdbcTemplate ? can't find; was it changed position?

davidreher:
@cswzxly I think you are in the wrong thread. We are not using JdbcTemplate in our commits, nor did we move it ...

spring-context/src/main/java/org/springframework/format/datetime/DateFormatter.java
```diff
@@ -49,7 +49,7 @@
 		Map<ISO, String> formats = new HashMap<DateTimeFormat.ISO, String>(4);
 		formats.put(ISO.DATE, "yyyy-MM-dd");
 		formats.put(ISO.TIME, "HH:mm:ss.SSSZ");
```
brunokrebs: So why haven't you changed this one as well?

davidreher: To be honest, I just didn't see it. I will change this!

pivotal-issuemaster:
@Mischi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1026)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1026) to manually synchronize the status of this Pull Request.

PR #1027, opened 2016-04-04
Fix fypo.

Youmoo:
License Agreement.


snicoll:
Merged, thanks!

PR #1029, opened 2016-04-11
Fix 500 error caused by If-Modified-Since=0 HTTP header

nacmacfeegle:
License Agreement.

Prior to this commit a 500 error would be bubbled up to the user when the
If-Modified-Since header was passed to the backend with a value of 0.

In this fix, the zero value is explicitly handled (in same way as with a null
value) by returning -1 to the caller.

Issue: SPR-14144


jhoeller:
As discussed on SPR-14144, I've taken a slightly broader approach. Thanks for raising this!

PR #1032, opened 2016-04-13
Fix logging issue with MessageConverter

mogren:
This commit fixes an issue with MappingJackson2HttpMessageConverter where it always logs a Warning, even if another message converter can handle it. Example:

2016-04-13 13:22:30.760  WARN [app-service] 15947 --- [AppClient-1] .c.j.MappingJackson2HttpMessageConverter : Failed to evaluate deserialization for type [simple type, class com.company.product.component.protobuf.Client$Configurations]: com.fasterxml.jackson.databind.JsonMappingException: Can not find a (Map) Key deserializer for type [simple type, class com.google.protobuf.Descriptors$FieldDescriptor]

Issue: SPR-14163


mogren:
This is not the correct solution, will close the PR.

PR #1037, opened 2016-04-16
Use the defined constant in MergedContextConfiguration

izeye:
This PR simply uses the defined constant instead.


snicoll:
Thanks!

PR #1042, opened 2016-04-23
Added cron expression validation method for CronSequenceGenerator

Enigo:
It's handy to know in advance whether or not expression that is
passed to CronSequenceGenerator or CronTrigger constructor would
not results in IllegalArgumentException. The only way to do it
now is to try\catch an instance creation but it's kinda ugly.

---

License Agreement.

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/support/CronSequenceGenerator.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/support/CronSequenceGenerator.java
@@ -149,6 +149,22 @@ public class CronSequenceGenerator {
 		return calendar.getTime();
 	}
 
+	/**
+	 * Indicates whether the specified cron expression can be parsed into a
+	 * valid cron sequence generator
+	 * @param cronExpression the expression to evaluate
+	 * @return a boolean indicating whether the given expression is a valid cron
+	 *         expression
+	 */
+	public static boolean isValidExpression(String cronExpression) {
+		String[] fields = StringUtils.tokenizeToStringArray(cronExpression, " ");
+		return validateCronFields(fields);
+	}
+
+	private static boolean validateCronFields(String[] fields) {
+		return fields != null && fields.length == 6;
+	}
+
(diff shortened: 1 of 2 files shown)


spring-context/src/main/java/org/springframework/scheduling/support/CronSequenceGenerator.java
```diff
@@ -262,7 +274,7 @@ private void reset(Calendar calendar, List<Integer> fields) {
 	 */
 	private void parse(String expression) throws IllegalArgumentException {
 		String[] fields = StringUtils.tokenizeToStringArray(expression, " ");
```
sbrannen: With the current implementation, you are tokenizing the string twice.

We would ideally prefer to avoid that.

Enigo: Thanks for reply! Updated my commit.

spring-context/src/main/java/org/springframework/scheduling/support/CronSequenceGenerator.java
```diff
@@ -149,8 +149,24 @@ public Date next(Date date) {
...
+		return validateCronFields(fields);
+	}
+
+	private static boolean validateCronFields(String[] fields) {
+		return fields != null && fields.length == 6;
+	}
+
 	private void doNext(Calendar calendar, int dot) {
-		List<Integer> resets = new ArrayList<Integer>();
+		List<Integer> resets = new ArrayList<>();
```
sbrannen: `master` is currently for Spring Framework 4.3 which supports Java 6, 7, and 8.

We therefore cannot use the `<>` operator.

Enigo: Oh, all right! Updated.

sbrannen:
This has been merged into `master` and subsequently polished.

Thanks!

PR #1043, opened 2016-04-24
Support non-blocking receiveTimeout.

erik-wramner:
AbstractPollingMessageListenerContainer#receiveMessage currently blocks
indefinitely both on -1 and 0 as receiveTimeout. This is because
MessageConsumer#receive blocks if the timeout is 0 and it is not
possible to poll without a timeout. This commit fixes that by calling
receiveNoWait if the timeout is 0 as receive(0) blocks.
In some cases (specifically AQ) this can yield better performance.

Issue: SPR-14212

### DIFF
--- a/spring-jms/src/main/java/org/springframework/jms/listener/AbstractPollingMessageListenerContainer.java
+++ b/spring-jms/src/main/java/org/springframework/jms/listener/AbstractPollingMessageListenerContainer.java
@@ -417,7 +417,8 @@ public abstract class AbstractPollingMessageListenerContainer extends AbstractMe
 	 * @throws JMSException if thrown by JMS methods
 	 */
 	protected Message receiveMessage(MessageConsumer consumer) throws JMSException {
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks, have you signed the CLA?

jhoeller:
I'm in the process of implementing this slightly differently, in sync with JmsTemplate's receive timeout handling. Thanks for the pull request in any case!

erik-wramner:
Yes, I have signed the CLA. Do you want me to post the confirmation 
number?

-Erik

erik-wramner:
Great. One thing though. I intend to submit another issue with pull 
request for adding a client side timeout in 
DefaultMessageListenerContainer#receiveMessage. I'll save the long 
motivation for the issue, but it helps us a lot with performance using 
AQ JMS. That client timeout only works really well if receiveMessage in 
the base class calls receiveNoWait when configured with a zero timeout. 
How the exact implementation is done doesn't matter, I look forward to 
testing your solution.

PR #1044, opened 2016-04-25
MockWebResponseBuilder defaults cookie domain

rwinch:
Previously MockWebResponseBuilder would use the cookie domain of the cookie
received from MockMvc response. This caused problems because the cookie
domain can be null, but HtmlUnit forbids a null domain.

This commit defaults the domain of the cookie to the domain of the
WebRequest.

Issues SPR-14169

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockWebResponseBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockWebResponseBuilder.java
@@ -29,6 +29,7 @@ import com.gargoylesoftware.htmlunit.WebResponse;
 import com.gargoylesoftware.htmlunit.WebResponseData;
 import com.gargoylesoftware.htmlunit.util.NameValuePair;
 
+import org.apache.http.impl.cookie.BasicClientCookie;
 import org.springframework.http.HttpStatus;
 import org.springframework.mock.web.MockHttpServletResponse;
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


rwinch:
cc @rstoyanchev @sbrannen

sbrannen:
@rwinch,

Thanks for the quick work!

I haven't cloned your PR locally to try it out, but the implementation looks solid, and the test reflects what I think one would expect (i.e., that the cookie domain defaults to the request domain).

So, from my perspective, feel free to merge it.

rstoyanchev:
I agree but either way I'll do that before RC2. Thanks a lot Rob!

rwinch:
@sbrannen @rstoyanchev 

Thanks for the feedback. I decided to improve the test by only setting required fields (i.e. fields populated by the constructor) on the Cookie. This will help to ensure we don't have problems for any other fields. I have pushed those changes to the branch.

chrylis:
Is the current-request-host guaranteed to be just `localhost` for requests on this path? If not, a cookie returned from request to `foo.example.com` without a domain set will be valid for `*.example.com`.

rwinch:
> Is the current-request-host guaranteed to be just localhost for requests on this path?

The host is the value that the user passed into MockMvc and can be any value.

> If not, a cookie returned from request to foo.example.com without a domain set will be valid for *.example.com.

Previously I didn't see a way around this. I found a way to work around the problem. See the updated PR.

rstoyanchev:
Merged, thanks!

PR #1046, opened 2016-04-28
"final" for public static fields in ReflectionUtils

panchenko:


panchenko:
License Agreement.

PR #1047, opened 2016-04-29
Proposal for easing the development of annotation driven static method advices

laszlocsontos:
# Proposal for easing the development of annotation driven static method advices and pointcuts

When developmening adivces around custom annotations, I'd like to have a simple AOP support API.
Preferably, I'd like to have the annotation passed to the interceptor to be able process its
attributes in a type-safe way.
### Interceptor implementation

```
@interface LogUsage {
  String usageTypeDescription();
}
```

```
public class LogUsageMethodAdvice extends AnnotationMethodAdvice<LogUsage> {
  @Override
  public Class<LogUsage> getAnnotationType() {
    return LogUsage.class;
  }

  @Override
  protected void duringFinally(LogUsage annotation, Method method, Object[] args, Object target) {
    String usageTypeDescription = annotation.usageTypeDescription()
    // Log usage
  }
}
```
### Service bean

```
@Service
public class SummarizationServiceImpl implements SummarizationService {
  @Override
  @LogUsage(usageTypeDescription = UsageService.TYPE_SUMMARY)
  Document summarize(Document document, SummaryConfiguration summaryConfiguration) {
  // App logic.
  }
}
```
### AOP configuration

```
     xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd">

  <bean id="logUsageMethodAdvice" class="app.usage.LogUsageMethodAdvice" />

  <bean id="logUsageMethodAdvisor" class="org.springframework.aop.support.annotation.AnnotationMatchingMethodPointcutAdvisor">
    <constructor-arg name="methodAnnotationType" value="app.usage.LogUsage" type="java.lang.Class" />
    <constructor-arg name="advice" ref="logUsageMethodAdvice"/>
  </bean>

  <bean class="org.springframework.aop.framework.autoproxy.DefaultAdvisorAutoProxyCreator" />

</beans>
```
## Implementation

There was two new API classes added to make this happen:
- AnnotationMatchingMethodPointcutAdvisor which is a convinience advisor support wrapping the existing AnnotationMatchingPointcut.
- AnnotationMethodAdvice which is the base class of annotation driven method advices. It contains all the possible hooks for capturing the before, afterReturning, afterThrowing and duringFinally events on the intercepted method call. All these method receive the method's annotation in a type safe way.

**Issue:** [SPR-14227](https://jira.spring.io/browse/SPR-14227)

---

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/support/annotation/AnnotationMatchingMethodPointcutAdvisor.java
+++ b/spring-aop/src/main/java/org/springframework/aop/support/annotation/AnnotationMatchingMethodPointcutAdvisor.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.aop.support.annotation;
+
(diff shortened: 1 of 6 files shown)


reflexdemon:
:+1:

PR #1050, opened 2016-05-05
Remove potential XSS vulnerability in Freemarker documentation

fj:
Since the user controls the value of `spring.status.value`, one wouldn't want to echo unescaped values here in this form. Otherwise one could submit a value of, say, `"/> <script>alert("pwned")</script>` or any other arbitrary JS.

This patch changes the `input` tag to use the HTML-escaped value rather than the raw value.


bclozel:
Thanks @fj !

PR #1052, opened 2016-05-08
Fix CacheManager example in Javadoc

ferstl:
Use `CacheManager#setCaches()` instead of `CacheManager#addCaches()` in the Javadoc example.


ferstl:

snicoll:
Merged, thanks for the PR!

PR #1054, opened 2016-05-10
Add MediaType IMAGE_SVG

sgrillon14:
When you do a Web Service to Expose a svg (example: web service expose flags of the world)

in my RestControler I use this syntax:
@RequestMapping(value = "/{country}", method = RequestMethod.GET, produces = MediaType.IMAGE_SVG_VALUE) 

instead of

@RequestMapping(value = "/{country}", method = RequestMethod.GET, produces = "image/svg+xml")

is more smart :)


YanshuoShang:
yes,this is more smart

pivotal-issuemaster:
@sgrillon14 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1054)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1054) to manually synchronize the status of this Pull Request.

sgrillon14:
I had personal licence (I'm an individual contributor)

pivotal-issuemaster:
@sgrillon14 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1054)!

sgrillon14:
@snicoll, what do you think this PR? (is it simple change but more smart)

snicoll:
None, otherwise I would have added a comment already.

bclozel:
Thanks for this contribution, but I don't see this as a very common media type in Spring applications. There's no specific support for that in Spring Framework, so it's mainly about the convenience of having it in the `MediaType` enum already vs storing in a static attribute that media type.

Arguably, we already have a lot of those, but we should try to limit ourselves there or we're in danger of having an even larger enum.

sgrillon14:
OK, it's a shame for the SVG because this standard is more and more used (to generate SVG side Back-End, for example flags, APP6 military symbols, games (dominos, playing cards, ...) ...

PR #1055, opened 2016-05-11
MockMvcWebConnection stores cookies from response

rwinch:
Previously MockMvcWebConnection did not update the cookie manager with the
cookies from MockHttpServletResponse. This meant that newly added cookies
are not saved to the cookie manager and thus are not presented in the next
request.

This commit ensures that MockMvcWebConnection stores the response cookies
in the cookie manager.

Issue: [SPR-14265](https://jira.spring.io/browse/SPR-14265)


rwinch:
When possible, please review @sbrannen @rstoyanchev

sbrannen:
I took a quick look, and it appears reasonable to me.

rstoyanchev:
Merged, thanks.

PR #1059, opened 2016-05-14
fix SPR-14269: make markBeanAsCreated(String beanName) thread safe

anguo-wenz:


snicoll:
Thanks for the PR, the issue has been fixed, see the jira issue for more details.

PR #1060, opened 2016-05-14
Treat Kotlin nullable as non-required

rocketraman:
Modify the core MethodParameter class to return the Kotlin nullability
of a parameteter (cached for performance).

Where `isOptional` is used, also check for `isKotlinNullable` i.e.
values are not considered required if they are Kotlin nullables:
- spring-messaging: named value method arguments
- spring-web: named value method arguments
- spring-webmvc: request parts

This means that Kotlin client code no longer has to explicity specify
"required=false" for Kotlin nullables -- this information is inferred
automatically by the framework.

Issue: SPR-14165

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Optional;
 
 import org.springframework.util.Assert;
+import org.springframework.util.KotlinUtils;
 
 /**
  * Helper class that encapsulates the specification of a method parameter, i.e. a {@link Method}
@@ -329,6 +330,16 @@ public class MethodParameter {
 		return (isOptional() ? nested() : this);
 	}
 
+	/**
+	 * Return whether this method parameter is declared as a "nullable" value, if supported by
+	 * the underlying language. Currently the only supported language is Kotlin.
+	 * @since 5.0
+	 */
+	public boolean isNullable() {
+		return KotlinUtils.isKotlinPresent() &&
+				KotlinUtils.isKotlinClass(getContainingClass()) &&
+				KotlinUtils.isNullable(this.parameterIndex, this.method, this.constructor);
+	}
 
 	/**
 	 * Set a containing class to resolve the parameter type against.
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@ buildscript {
 	dependencies {
 		classpath("org.springframework.build.gradle:propdeps-plugin:0.0.7")
 		classpath("org.asciidoctor:asciidoctor-gradle-plugin:1.5.2")
+		classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.0.4"
 		classpath("io.spring.gradle:docbook-reference-plugin:0.3.1")
 		classpath("ws.antonov.gradle.plugins:gradle-plugin-protobuf:0.9.1")
 	}
@@ -71,6 +72,7 @@ configure(allprojects) { project ->
 	ext.junitPlatformVersion   = '1.0.0-M2'
 	ext.log4jVersion           = '2.7'
 	ext.nettyVersion           = "4.1.6.Final"
+	ext.kotlinVersion          = "1.0.4"
 	ext.okhttpVersion          = "2.7.5"
 	ext.okhttp3Version         = "3.4.1"
 	ext.poiVersion             = "3.15"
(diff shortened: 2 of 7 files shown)


rocketraman:
Notes on submission / pending work:

* I put the kotlin-specific code into a separate module `spring-core-kotlin` to avoid adding the kotlin compiler to the `spring-core` module (inspired from `spring-beans-groovy`).
* I wasn't sure where to add tests for this. My first though was `MethodParameterTests`, but currently it contains very little in the way of tests. Guidance is appreciated.

Otherwise the implementation turned out to be quite simple and concise.

sdeleuze:
Thanks for this PR, I only had a quick look but I would be in favor of not creating a new `spring-core-kotlin`, instead you could add kotlin dependency as an optional one. After that I will make a detailed review of your PR.

rocketraman:
> Thanks for this PR, I only had a quick look but I would be in favor of not creating a new spring-core-kotlin, instead you could add kotlin dependency as an optional one. After that I will make a detailed review of your PR.

Done!

rocketraman:
NOTE: Another place we could use this functionality is when `@Inject` or `@Autowired` is used, the Kotlin nullable information could be used to determine if the value is required or not.

rocketraman:
@sdeleuze Ping. Looks like it needs another rebase, but would appreciate a review.

spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -329,6 +338,19 @@ public MethodParameter nestedIfOptional() {
 		return (isOptional() ? nested() : this);
 	}
 
+	/**
+	 * Return whether this method parameter is declared as a Kotlin
+	 * nullable.
+	 * @since 4.4
```
sdeleuze: Let's target 5.0 for this feature

spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -329,6 +338,19 @@ public MethodParameter nestedIfOptional() {
...
+	 */
+	public boolean isKotlinNullable() {
```
sdeleuze: Maybe `isNullable()`would be less specific and let the door opened to supporting this is other languages or with specific annotations

spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -50,6 +51,12 @@
  */
 public class MethodParameter {
 
+	private static final boolean kotlinPresent;
```
sdeleuze: Let's encapsulate that in a KotlinUtils class

spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -67,6 +74,8 @@
 
 	private volatile Class<?> parameterType;
 
+	private volatile Boolean isKotlinNullable;
```
sdeleuze: I think I would try initially without caching the result, unless you have specific concern about the performances

spring-core/src/main/kotlin/org/springframework/core/MethodParameterKotlin.kt
```diff
@@ -0,0 +1,49 @@
...
+import kotlin.reflect.jvm.kotlinFunction
+
+/**
+ * Helper internal class that provides Kotlin-specific extensions to [MethodParameter].
+ *
+ * @author Raman Gupta
+ * @since 4.4
+ * @see MethodParameter
+ */
+object MethodParameterKotlin {
```
sdeleuze: Could we turn this into a regular java `KotlinUtils` class with maybe an additional method to know if Kotlin is present? I think I would like to restrict Kotlin code to only test classes.

sdeleuze:
Sorry for the delay and thanks for your updated PR. I have added comments on the relevant lines of your commit.

Could you also rebase that on top of the new master (Spring Framework 5.0) and add some Kotlin unit tests to check that works?

Good idea to support that also for `@Inject`, feel free to add that to your PR and update the title of this JIRA issue to make is less specific.

rocketraman:
> Good idea to support that also for @Inject, feel free to add that to your PR and update the title of this JIRA issue to make is less specific.

It looks more complicated to do this [1], so I'd prefer to have this pull request reviewed as-is for now, just to keep things as simple as possible. We can add the `@Inject` functionality in a subsequent pull.

> Sorry for the delay and thanks for your updated PR. I have added comments on the relevant lines of your commit.
> 
> Could you also rebase that on top of the new master (Spring Framework 5.0) and add some Kotlin unit tests to check that works?

I have updated the pull with all of your comments taken into account. Thanks!

[1] It looks like a variation of `RequiredAnnotationBeanPostProcessor` is needed.

rocketraman:
Ping @sdeleuze for review.

sdeleuze:
Superseeded by #1242

PR #1062, opened 2016-05-19
SPR-14291: SimpleMappingExceptionResolver doesn't catch exceptions being thrown from Spring Formatters

jcagarcia:
I've just create the following issue on JIRA https://jira.spring.io/browse/SPR-14291 and its related question in StackOverflow 

http://stackoverflow.com/questions/37322182/exceptionhandler-doesnt-catch-exceptions-being-thrown-from-spring-formatters

I've just included the provided implementation that checks if the **cause exception** matches with some exception registered on `SimpleMappingExceptionResolver`.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/SimpleMappingExceptionResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/SimpleMappingExceptionResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jcagarcia:
So, if this changes will not be applied, will be necessary to apply [this](https://github.com/DISID/disid-proofs/blob/master/spring-boot-exception-handling/SimpleMappingExceptionResolver/src/main/java/com/springsource/petclinic/web/exceptions/CauseAdviceSimpleMappingExceptionResolver.java) workaround everytime I want to throw a custom exception from Spring Formatter to be able to check cause exception?

Regards,

rstoyanchev:
I've closed this because it needs more discussion. I'm sure this PR solves the problem you need it to solve well but there are other cases that need to be considered before this goes in.

jcagarcia:
Hi @rstoyanchev , that's perfect! 

I've updated the JIRA ticket too with some useful information.

Regards and thanks for your support!

PR #1065, opened 2016-05-25
Fix comments on private fields in TableMetaDataContext

benelog:
It seems that comments on private fields in TableMetaDataContext is copied from CallMetaDataContext. This PR changes unsuitable comments for TableMetaDataContext.


snicoll:
Merged, thanks!

PR #1067, opened 2016-05-28
Allow consutractor injection at the CompositeDatabasePopulator

kazuki43zoo:
Issue: SPR-14313

I've fixed [SPR-14313](https://jira.spring.io/browse/SPR-14313).
Please review this.

Thanks.


jhoeller:
Rolled into 4.3 in a slightly different fashion, in particular with an overloaded Collection-based variant and without delegation to other methods from the constructors. I've also added your unit tests, just slightly revising the style towards instance-level mocks which do not need to be reset afterwards.

kazuki43zoo:
Hi @jhoeller , thanks for fix & polishing !!

PR #1069, opened 2016-05-30
Fix SPR-14315:Flush headers before return

zhoulifu:
Flush headers before method return in StreamingResponseBodyReturnValueHandler#handleReturnValue

Issue: SPR-14315


zhoulifu:
License Agreement.

rstoyanchev:
I've merged this and also updated the ResponseBodyEmitter accordingly.

zhoulifu:
Thank you!

PR #1072, opened 2016-06-03
UTF-8 as default for "MULTIPART_FORM_DATA" conversion

RG9:
FormHttpMessageConverter should use UTF-8 on default for converting "MULTIPART_FORM_DATA" to handle Eastern Europe characters. It could be overriden, however in default causes unnecessary bug.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/FormHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/FormHttpMessageConverter.java
@@ -104,7 +104,7 @@ public class FormHttpMessageConverter implements HttpMessageConverter<MultiValue
 		this.supportedMediaTypes.add(MediaType.MULTIPART_FORM_DATA);
 
 		this.partConverters.add(new ByteArrayHttpMessageConverter());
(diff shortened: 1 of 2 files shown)


bclozel:
Could you open an issue on jira.spring.io and explain in more details why this should be done? Encoding matters are often complex, so we need to make sure the default should be changed there instead of keeping the Servlet specification default encoding.

Note, this seems to be somewhat related to [SPR-7789](https://jira.spring.io/browse/SPR-7789), [SPR-9096](https://jira.spring.io/browse/SPR-9096), and [SPR-11035](https://jira.spring.io/browse/SPR-11035).

Also, have you signed the [CLA](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#sign-the-individual-contributor-license-agreement-icla)?

RG9:
Thank you help.

Here is an issue [SPR-14338](https://jira.spring.io/browse/SPR-14338).

benelog:
It is fixed by https://github.com/spring-projects/spring-framework/commit/cc7781ecf323f7c9fbc6959bfabf73df613c274e

RG9:
Yes, thank you, now I can only write some unit tests :)

PR #1074, opened 2016-06-07
Dual setDeserializers method

djechelon:
For simple duality with setSerializers, useful when configuring via XML app context

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperFactoryBean.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperFactoryBean.java
@@ -273,6 +273,16 @@ public class Jackson2ObjectMapperFactoryBean implements FactoryBean<ObjectMapper
 	}
 
 	/**
+	 * Configure custom deserializers. Each deserializer is registered for the type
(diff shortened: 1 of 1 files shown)


djechelon:
Please forgive me if I didn't follow any of the contribution guidelines, but this is a trivial improvement, so trivial that I have hand-written the code and omitted testing. It is meant as a proposal and is to be reviewed by Spring developers because I don't have a Spring development environment ready to go.

jhoeller:
Thanks for the suggestion! I've raised https://jira.spring.io/browse/SPR-14337 for this, about to be included in this week's 4.3 GA release. As a side note, the change turned out to be a bit more extensive since we also need a corresponding method in our Jackson2ObjectMapperBuilder.

djechelon:
Just for sake of completeness here is the only way to achieve the same result with existing Spring code

```
            <bean class="org.springframework.http.converter.json.Jackson2ObjectMapperFactoryBean">
                <property name="deserializersByType">
                    <map key-type="java.lang.Class">
                        <entry>
                            <key>
                                <value>java.lang.String</value>
                            </key>
                            <bean class="Some.Deserializer" />
                        </entry>
                    </map>
                </property>
```

And yea, sorry, but I was pretty sure to have found `deserializers()` method in the builder, unfortunately I was wrong :smile:

PR #1078, opened 2016-06-13
Fix getFieldValues to allow proper extending

mhlz:
I'm currently using Spring 4.3.0.RC2 (via Spring Boot 1.4.0.M3) and I noticed that in some circumstances CORS headers were added multiple times, which causes the request to fail.

I noticed that the DefaultCorsProcessor does perform a check to see if the headers were already added, however that check didn't seem to work.

I traced the problem back further to ServletServerHttpResponse and its class ServletResponseHttpHeaders which extends HttpHeaders. The DefaultCorsProcessor uses the getAccessControlAllowOrigin function to check for the headers. That function in turn uses the protected function getFieldValues to build a comma delimited string. This function accesses this.headers directly, instead of using the get function which causes this problem, since ServletResponseHttpHeaders only overwrites the get function.

I changed the protected getFieldValues function to use the get function which should fix the problem, but is a fairly trivial change.

As a workaround I registered the CorsFilter manually instead of relying on Spring Boots autoconfiguration (which uses an HttpInterceptor instead of a filter)

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -16,31 +16,19 @@
 
 package org.springframework.http;
 
+import org.springframework.util.Assert;
+import org.springframework.util.LinkedCaseInsensitiveMap;
+import org.springframework.util.MultiValueMap;
+import org.springframework.util.StringUtils;
+
 import java.io.Serializable;
 import java.net.URI;
 import java.nio.charset.Charset;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.EnumSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.TimeZone;
+import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
-import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


mhlz:
License Agreement.

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -16,32 +16,20 @@
 
 package org.springframework.http;
 
+import org.springframework.util.Assert;
```
benelog: You can refer  [Spring-Framework-Code-Style](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style#import-statements) for the ordering of import statements.

mhlz:
I fixed the import order in that file.

pivotal-issuemaster:
@mhlz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1078)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1078) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mhlz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1078)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1078) to manually synchronize the status of this Pull Request.

sbrannen:
This change has already been made in the `4.3.x` branch.

https://github.com/spring-projects/spring-framework/blob/e0883f5eaaca7004fab8a0b5cc3909c21b47290f/spring-web/src/main/java/org/springframework/http/HttpHeaders.java#L1147-L1150

PR #1082, opened 2016-06-17
Move <a> out of @code

nkjackzhang:
It looks like the &lt;a&gt; in the comment doesn't work, so I move the &lt;a&gt; out of @code.

License Agreement.


snicoll:
merged, thanks!

PR #1084, opened 2016-06-17
SPR-14433 - Inline variable that is used only once

nkjackzhang:
The variable canonicalName is used only once, and there is no need
to call canonicalName() when 
"alreadySeen != null && alreadySeen.contains(beanName)" is true, 
so inline it.

License Agreement.


jhoeller:
Thanks for raising this! Unfortunately I haven't noticed your pull request before... In any case, I've applied this refinement to all our actively maintained 4.x branches right away, in a slightly different way (since we prefer avoiding nested calls, we still have a variable declaration but in the latest possible place now).

PR #1085, opened 2016-06-17
Fix ambiguous static import

nkjackzhang:
Since there is also a method org.mockito.Matchers.startsWith(String prefix), so the method 
startsWith(String) is ambiguous for the type TestPropertySourceUtilsTests, and as I don't know
which one you want to use, so I just change the static import for 
org.hamcrest.CoreMatchers.containsString to use org.hamcrest.CoreMatchers.startsWith explicitly.

License Agreement.


sbrannen:
Thanks!

PR #1087, opened 2016-06-20
Is there a better way to solve this problem？

hxy1991:
When I call "request.getParameter" in the javax.servlet.Filter, request will be parsed into "parts" field.After then,the request reaches "org.springframework.web.multipart.commons.CommonsMultipartResolver.parseRequest" and will be parsed again,but it can not get any request parameter values.
The method in my controller is "public String saveImage(@RequestParam("imageFile") MultipartFile file, @RequestParam Integer userId)".
The logic in my filter proceed as follows.
[code]
public void doFilter(ServletRequest request, ServletResponse response, FilterChain filterChain) throws IOException, ServletException {
    Integer userId = request.getParameter("userId")
    if (isauthorized(userId)) {
        filterChain.doFilter(request, response);
        return;
    }
    ((HttpServletResponse) response).sendError(HttpServletResponse.SC_FORBIDDEN);
}
[/code]
I solved this problem temporarily by change "CommonsMultipartResolver.parseRequest",but it is not rigorous.Is there a better way to solve this problem？
Thanks.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/multipart/commons/CommonsMultipartResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/multipart/commons/CommonsMultipartResolver.java
@@ -155,6 +155,27 @@ public class CommonsMultipartResolver extends CommonsFileUploadSupport
 		String encoding = determineEncoding(request);
 		FileUpload fileUpload = prepareFileUpload(encoding);
 		try {
+			Collection<Part> parts = request.getParts();
+			if(parts != null){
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Sorry but this is not a place for asking questions. Please follow [the guidelines](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#search-stack-overflow-first-discuss-if-necessary).

PR #1088, opened 2016-06-20
SPR-14370 refactoring

djechelon:
Made propertySource protected so that getProperty() can be overriden by
copy&paste

Discussion in Jira


snicoll:
The related issue is now closed. Thanks for the PR!

PR #1098, opened 2016-07-04
SPR-13826 - Move to JDBC 4.1

eddumelendez:
See SPR-13826

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/simple/AbstractJdbcInsert.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/simple/AbstractJdbcInsert.java
@@ -476,31 +476,20 @@ public abstract class AbstractJdbcInsert {
 					@Override
 					public Object doInConnection(Connection con) throws SQLException, DataAccessException {
 						// Do the insert
-						PreparedStatement ps = null;
-						try {
-							ps = con.prepareStatement(getInsertString());
+						try (PreparedStatement ps = con.prepareStatement(getInsertString())) {
 							setParameterValues(ps, values, getInsertTypes());
 							ps.executeUpdate();
 						}
-						finally {
-							JdbcUtils.closeStatement(ps);
-						}
 						//Get the key
-						Statement keyStmt = null;
-						ResultSet rs = null;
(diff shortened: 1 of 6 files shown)


jmithmstr:
Please, do not use try-with-resources as exceptions from close() are not ignored. And I do not want those exceptions thrown up when everything but the close was OK.

spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
```diff
@@ -1067,9 +1067,8 @@ public int update(String sql, Object... args) throws DataAccessException {
 			logger.debug("Calling stored procedure" + (sql != null ? " [" + sql  + "]" : ""));
 		}
 
-		Connection con = DataSourceUtils.getConnection(getDataSource());
 		CallableStatement cs = null;
-		try {
+		try (Connection con = DataSourceUtils.getConnection(getDataSource())) {
```
snicoll: We're not sure yet we want to do that. And we probably won't. Can you push force an update that doesn't use try/resource please?

eddumelendez:
@snicoll PR updated. Thanks

jhoeller:
@eddumelendez I share the concerns about try-with-resources. We do not just have transactional logic in `DataSourceUtils.releaseConnection` but also custom logging with no rethrowing in `closeStatement` and `closeResultSet`. Since there is no user-level benefit in changing those to try-with-resources, as far as I can see, let's rather leave this as-is.

Thanks for your efforts in any case!

jhoeller:
@eddumelendez I appreciate your efforts but I share @jmithmstr's concerns about try-with-resources. We have some custom logging logic in our close delegates, even transactional logic in `DataSourceUtils.releaseConnection`... Since there is no user-level benefit in changing those to try-with-resources, let's rather leave this as-is.

jhoeller:
As for SPR-13826, that issue is more about relying on JDBC 4.1 features such as `getParameterType` than about the resource handling code. Since unconditional use of such JDBC operations requires full support in all common JDBC drivers out there, it's driven by the drivers and not by the JDBC API level in JDK 8.

PR #1100, opened 2016-07-07
Declare the injection point as a Provider

nkjackzhang:
It seems like there is no injection point in the example, so I just add setter to inject the Provider.

License Agreement.


snicoll:
Merged, thanks.

PR #1101, opened 2016-07-07
Fix typo in javadoc

nkjackzhang:
Fix typo in javadoc and delete extra white space line.

License Agreement.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StopWatch.java
+++ b/spring-core/src/main/java/org/springframework/util/StopWatch.java
@@ -67,7 +67,6 @@ public class StopWatch {
 	/** Total running time */
 	private long totalTimeMillis;
 
-
 	/**
 	 * Construct a new stop watch. Does not start any task.
 	 */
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/util/StopWatch.java
```diff
@@ -289,7 +283,6 @@ public String toString() {
 		return sb.toString();
 	}
 
-
```
snicoll: That one should actually stay to separate inner classes from methods.

snicoll:
Thanks for the PR but the current code looks fine as it is. Changing the second `{@code}` would link to the method itself. I think the intention was to use code and have a separate `@see` below.

nkjackzhang:
But in javadoc, there exist extra "#", may be we can just delete the extra characters.

snicoll:
Ah good point, that's a glitch introduced in a6a6aed17f3ef9e9f774f25c14c7e6e5bb5dd2f3 - Fixed in https://github.com/spring-projects/spring-framework/commit/c425e774a3fc96f205832541f630dd205373b940

nkjackzhang:
Thanks!

PR #1106, opened 2016-07-11
SPR-14456 - Asynchronous tests of controller that returns StreamingResponseBody

darklynx:
This pull requests add 3 tests to `AsyncTests.java`. They demonstrate testing of asynchronous request handling if controller returns an instance of `StreamingResponseBody`. This differs from testing of controller that return `Callable<?>` or `DeferredResult<?>` very much.
## Background story

While testing my Spring MVC application I came across your great examples of how to test asynchronous request processing using `spring-test` library in [AsyncTests.java](https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/test/java/org/springframework/test/web/servlet/samples/standalone/AsyncTests.java) file.

My application however is designed to steam a huge amount of data and thus controllers return instances of `StreamingResponseBody` interface.

Unfortunately the examples from `AsyncTests.java` did not work as expected. During second step after the `asyncDispatch(mvcResult)` method my expectations against `content()` always failed claiming that response is empty whereas I've expected some data to be written to output stream.

It took me some time to debug and figure out that the `asyncDispatch(...` method is actually returns back only the modified version of original request, and the `perform(...` method creates a new instance of `MvcResult` with a fresh version of `MockHttpServletResponse`. The logic inside controllers was writing the data into original HTTP response generated by initial `perform(...` method, while I was validating the content of HTTP response generated by second `perform(...`.

If somebody else is looking for a good example to test controller with streaming body, it would be great if they could find it in the source code of original library.

### DIFF
--- a/spring-test/src/test/java/org/springframework/test/web/servlet/samples/standalone/AsyncTests.java
+++ b/spring-test/src/test/java/org/springframework/test/web/servlet/samples/standalone/AsyncTests.java
@@ -17,6 +17,7 @@
 package org.springframework.test.web.servlet.samples.standalone;
 
 import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
@@ -26,6 +27,7 @@ import org.junit.Test;
 
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.test.web.Person;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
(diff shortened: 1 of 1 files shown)


snicoll:
Thank you for the PR. Have you signed the CLA?

darklynx:
@snicoll I only read the `CONTRIBUTING.md` after I have already submitted the PR, very structured and self-explaining document (+1). I'll sign the CLA tonight and post corresponding message here. Do I need to create a ticket for this PR in the Spring JIRA or since the changes do not fix any bug and do not add any new features to the library this step can be omitted?

snicoll:
Usually you should always go through Jira and then submit your PR with a reference to the issue you've created. In that particular case, I am not sure but maybe @rstoyanchev has an opinion?

darklynx:
License Agreement.

rstoyanchev:
This is now merged, thanks. Normally I would prefer a ticket indeed that records the change but for a test improvement we can do without.

darklynx:
I have just created an issue in Spring JIRA: https://jira.spring.io/browse/SPR-14456 :)
I'm not sure how to tag this request now, but probably the ticket can now be closed as well.

PR #1113, opened 2016-07-16
Applied skeletal implementation to interface migration automated refactoring

khatchad:
This is a semantics-preserving refactoring that migrates existing method implementations in classes to the corresponding implemented interface methods as `default` methods. The tool **does not** add new code; it only rearranges _existing_ code.
- We are evaluating a research prototype automated refactoring Eclipse plug-in called [Migrate Skeletal Implementation to Interface](https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring). We have applied the tool to your project in the hopes of receiving feedback.
- The approach is very conservative. That may mean that not all changes that can be made were made. Please feel free to continue the refactoring manually if you wish.
- We only migrated methods declared in abstract classes with the hopes of such methods being suitable default methods in corresponding interfaces.
- The source code should be semantically equivalent to the original.

Thank you for your help in this evaluation! Any feedback you can provide would be very helpful. In particular, we are interested if _each_ of the proposed changes are helpful or not.


bclozel:
Hi @khatchad 

Thanks for this interesting PR. I've taken a look at those changes and I right now I'm not keen on merging this for the following reasons:

1. As an integration framework, our code has to do "strange things" in order to guard against libraries limitations or support multiple versions. Automated analysis/refactoring tools tend to have a lot of false positives on our codebase
2. It is really important for us to keep a clean codebase and we have a strict "0 package tangle" policy. Moving code (and imports) around is rather risky. I haven't run Sonar on your PR but I think this might be an interesting feedback for your tool.
3. It seems there are currently [some performance limitations to default methods](http://www.scala-lang.org/blog/2016/07/08/trait-method-performance.html), so I wouldn't consider moving all that code to default methods without a good reason.
4. We're in the middle of big changes for Spring 5.0, so your PR is already outdated

I'm closing this for now and I'll let other team members reopen this in case they think we can do something here.

khatchad:
Hi @bclozel. Thank you for this feedback! Indeed, running sonar before and after the refactoring is something we've not previously considered and would be of interest. Also, please let us know if you would ever be interested in applying the refactoring to your new Spring 5.0 branch.

PR #1115, opened 2016-07-17
Fix javadoc reference

KingBuzzer:
Method is called nextBackOff() and not nextBackOffMillis().


KingBuzzer:
License Agreement.

snicoll:
Good catch, thanks!

PR #1116, opened 2016-07-18
SPR-14471 - Add autowiring support to SpringBeanJobFactory

mdeinum:
This commit introduces autowiring support to the SpringBeanJobFactory.
It will call the autowireBean method if an AutowireCapableBeanFactory has
been injected. The autowiring will be done on either the Job or the delegate
of the DelegatingJob which contains a Runnable.

A test has been added to verify the correct behavior of the autowiring and
as well as retaining the initial support of using the job data for wiring dependencies.

Issue: [SPR-14471](https://jira.spring.io/browse/SPR-14471)


Hronom:
Why this not accepted yet?

mdeinum:
Implemented as part of SPR-17323 (d9e72535328ff56e447a9ec69b95263da8a8495b)

mdeinum:
Implemented as part of SPR-17323 (d9e72535328ff56e447a9ec69b95263da8a8495b)

PR #1123, opened 2016-07-27
Refactor AbstractResponseBodyFlushProcessor states

violetagg:
With the current state machine
- the implementation can hang after the last element when executing
  on Jetty.
- in some cases there will be no flush after the last
  Publisher<DataBuffer>.


rstoyanchev:
This is now merged, thanks!

PR #1124, opened 2016-07-28
Add a bean aware CacheManager.

phillescott:
When trying to select a SSMCache based on a spring.active.profile the CacheManager was unable to select the correct cache based on the dynamic nature of the lookup. This CacheManager offers a sufficient level of abstraction and a standardised way of dynamically selecting a cache.

New cache manager that resolves the cache name against the Cache type
beans registered in the application context.

Issue: SPR-14519

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/support/BeanAwareCacheManager.java
+++ b/spring-context/src/main/java/org/springframework/cache/support/BeanAwareCacheManager.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cache.support;
+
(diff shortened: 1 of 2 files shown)


GleyderC:
Hello,

trigger@ifttt.com has been retired. Use trigger@recipe.ifttt.com to trigger your Email Channel Recipes.

—The IFTTT Team

snicoll:
@phillescott thank you so much for the PR Phill. I've commented on Jira.

phillescott:
Yea no problems.

Cheers for the feedback

phillescott:
Although this will still lookup the cache by cache name not the name of the cache bean in the application context.

snicoll:
That's much better IMO. If you create a `Cache` bean named `foo` and it's `Cache#getName` returns `bar` I would expect `bar` to be the valid name. If you feel otherwise, just align the cache name and the bean name.

phillescott:
Yes I do agree. It just we have a use case where were dynamically selecting a cache based on profile (it's an SSM cache and the cache name must match the SSM cache) so I was looking up by bean name as it would be constant. No worries though; it's an edge case that I accounted for and just thought I'd submit PR for it.

Cheers for being helpful with the responses

snicoll:
I have no idea what "SSM cache" is but if it's your code (or you are willing to extend it) you could something like this:

```
public class SsmCache implements Cache, BeanNameAware {

  private String name;

  @Override
  public String getName() {
    return this.name;
  }

  @Override
  public void setBeanName(String name) {
    this.name = name;
  }
}
```

That way you are sure that `getName()` always match the bean name. That's another way to achieve the same use case.

phillescott:
Ah sorry its a SimpleSpringMemcached (
https://www.javacodegeeks.com/2013/06/simple-spring-memcached-spring-caching-abstraction-and-memcached.html)
implementation.

I've managed to sort it through an alias they supply. I was just too keen
to commit something haha!

Enjoy your weekend.

PR #1128, opened 2016-08-01
Improved StringUtils

Akirus:
Improved following StringUtils methods:
- mergeStringArrays:
  now this method have a linear complexity instead of quadratic.
- splitArrayElementsIntoProperties:
  optimized case when charsToDelete param is empty

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -30,6 +30,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.TimeZone;
+import java.util.HashSet;
 
 /**
  * Miscellaneous {@link String} utility methods.
@@ -810,8 +811,9 @@ public abstract class StringUtils {
 		}
 		List<String> result = new ArrayList<>();
 		result.addAll(Arrays.asList(array1));
+		Set<String> stringSet = new HashSet<>(result);
 		for (String str : array2) {
-			if (!result.contains(str)) {
+			if (!stringSet.contains(str)) {
 				result.add(str);
 			}
 		}
@@ -889,9 +891,7 @@ public abstract class StringUtils {
 			return array;
 		}
 		Set<String> set = new LinkedHashSet<>();
-		for (String element : array) {
-			set.add(element);
-		}
+		Collections.addAll(set, array);
 		return toStringArray(set);
 	}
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@Akirus Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1128)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1128) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Akirus Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1128)!

spring-core/src/main/java/org/springframework/util/StringUtils.java
```diff
@@ -810,8 +811,9 @@ public static TimeZone parseTimeZoneString(String timeZoneString) {
 		}
 		List<String> result = new ArrayList<>();
 		result.addAll(Arrays.asList(array1));
+		Set<String> stringSet = new HashSet<>(result);
```
a-derevyanko: You can create HashSet from Arrays.asList(array1) (the "result" list is not required). 

But this way is not clear. I see an easier way:
/*
Set<String> result = new HashSet<>();
Collections.addAll(result, array1);
Collections.addAll(result, array2);
*/

Akirus: but your way is not equal to original, just becouse result can contains same items if they was present in a first array.

a-derevyanko: It is a Set. Not List.

Akirus: just look at original code
```
List<String> result = new ArrayList<>();
		result.addAll(Arrays.asList(array1));
		for (String str : array2) {
			if (!result.contains(str)) {
				result.add(str);
			}
		}
		return toStringArray(result);
```

a-derevyanko: Sorry, you was right. But HashSet wrapping looks like overengineering (original result.addAll(Arrays.asList(array1)) too :)). Have you tested, how your improvement works on large String arrays? 

---
But you did the same mistake like me) If second array contains duplicate items, you put to  "result" _both_ of duplicates (because your stringSet dont know about new items in result).

Akirus: yes, i did mistake) i can say that it would be really faster on large string arrays, but slower on small arrays.

spring-core/src/main/java/org/springframework/util/StringUtils.java
```diff
@@ -1068,7 +1068,8 @@ public static Properties splitArrayElementsIntoProperties(
 		List<String> result = new ArrayList<>();
 		if ("".equals(delimiter)) {
 			for (int i = 0; i < str.length(); i++) {
-				result.add(deleteAny(str.substring(i, i + 1), charsToDelete));
+				if(isEmpty(charsToDelete) || charsToDelete.indexOf(str.charAt(i)) > -1)
```
guonudt: I think it should be
 `charsToDelete.indexOf(str.charAt(i)) == -1`

PR #1136, opened 2016-08-17
SPR-14593: Use HttpClient constructor that supports SSL connections

olsson:
Fix for SPR-14593, using the HttpClient constructor that does support SSL, instead of one that forces port 80.


bclozel:
Merged with 64bc0ca744781af2bf

Thanks!

PR #1138, opened 2016-08-24
Improve performance of assertion in StompSubProtocolHandler.convertCo…

dreis2211:
…nnectAcktoStompConnected()

Closes SPR-14624


jhoeller:
Applied in a revised fashion as per my comments on JIRA.

PR #1141, opened 2016-08-25
Added Host property to HttpHeaders

poutsma:
Added getHost/setHost to HttpHeaders, returning/taking a
InetSocketAddress.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -803,7 +803,7 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 		if (value == null) {
 			return null;
 		}
-		int idx = value.indexOf(':');
+		int idx = value.lastIndexOf(':');
 		String hostname = null;
 		int port = 0;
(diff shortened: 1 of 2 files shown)


poutsma:
@bclozel: You've indicated the need for access to the Host header before. Can you please check this PR and see if it is to your liking?

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -781,6 +782,47 @@ public long getExpires() {
...
+	 * Return the value of the required {@code Host} header.
+	 * <p>If the header value does not contain a port, the returned
+	 * {@linkplain InetSocketAddress#getPort() port} will be {@code 0}.
+	 */
+	public InetSocketAddress getHost() {
+		String value = getFirst(HOST);
+		if (value == null) {
+			return null;
+		}
+		int idx = value.indexOf(':');
```
bclozel: Will that work with IPv6 addresses though?
Wondering if `lastIndexOf` is enough or if more checks are required for addresses like `[2001:db8::ff00:42:8329]:8080`.

bclozel:
Squashed/merged with ab7107c4c5b

Thanks!

PR #1142, opened 2016-08-25
Refactored SseEvent to ServerSideEvent

poutsma:
Refactored mutable SseEvent class into immutable ServerSideEvent with
builder.


poutsma:
@sdeleuze: Could you take a look at this PR? I renamed `SseEvent` to `ServerSideEvent` to make the name less redundant. It's also become immutable, having a builder to create new instances. I've also tried to re-align the class properties closer to the events described in the spec, so that `reconnectTime` becomes `retry`, and `name` becomes `event`.

bclozel:
@poutsma [Server-Sent Events|https://developer.mozilla.org/fr/docs/Server-sent_events]?

poutsma:
Ugh, guess I should go back on holiday. Let me come up with a new PR, this time changing the name to `ServerSentEvent`

PR #1143, opened 2016-08-25
Refactored SseEvent to ServerSentEvent

poutsma:
- Renamed SseEvent to ServerSentEvent to make the name less redundant.
- ServerSentEvent is now immutable, having a builder to create new instances.
- Realigned the class properties to more closely match the events
  described in the spec, so that `reconnectTime` becomes `retry`, and
  `name` becomes `event`.


poutsma:
My second attempt. @sdeleuze: could you take a look a this one instead of #1142?

spring-web/src/main/java/org/springframework/http/codec/ServerSentEvent.java
```diff
@@ -0,0 +1,209 @@
...
+    /**
+     * Return a builder for a {@code SseEvent}.
+     *
+     * @param <T> the type of data that this event contains
+     * @return the builder
+     */
+    public static <T> Builder<T> builder() {
+        return new BuilderImpl<>();
+    }
+
```
rstoyanchev: How about using the builder method to provide the data which is the most likely case:
````java
public static <T> Builder<T> data(T data) {
    return new BuilderImpl<>().data(data);
}
````
and also some option for no data:
````java
public static Builder<Object> withoutData() {
    return new BuilderImpl<>();
}
````
It goes from:
`ServerSentEvent.<String>builder().data("foo")`

to:
`ServerSentEvent.data("foo")`

**Comment by sdeleuze** [comment]
👍 For renaming to `SseEvent` to `ServerSentEvent` (I was hesitating between both options) and making it immutable with the builder.

I am also in favor of @rstoyanchev proposal to make it more convenient to use, but maybe not exactly with the same method names. What about:

```java
public static <T> Builder<T> builder(T data) {
    return new BuilderImpl<>().data(data);
}
```

And:
```java
public static Builder<Object> builder() {
    return new BuilderImpl<>();
}
```

Regarding the field renaming, I have mixed feelings. `name` and `reconnectTime` match with the description of the fields from the spec and are consistent with our current `SseEmitter.SseEventBuilder`. On the other way, I can see the benefit of just sticking to the SSE field names. It maybe interesting to see @rstoyanchev and @bclozel take on this.

poutsma:
I've added a `public static <T> Builder<T> builder(T data)` method.

As for the field renaming: the spec does not contain the concept of a `reconnectTime`, only a [`reconnection-time`](http://www.w3.org/TR/eventsource/#concept-event-stream-reconnection-time). `SseEmitter.SseEventBuilder` does use `reconnectTime`, unfortunately, but I don't think that alone is reason to stick with it. Also, `retry` is a lot shorter than `reconnectTime` or `reconnectionTime`, which is more important when used in a builder API.

The spec also doesn't contain the concept of a "event name"; the `event` field is used to denote the event *type*. If we change it, I feel we should change the methods that expose this field to be named `type`; not `name`.

In summary: the way `SseEmitter.SseEventBuilder` exposes SSE fields is supposed to reflect the "full names" of these properties (as defined in the spec), but unfortunately it doesn't do so. In the `ServerSentEvent` class, I'd rather use correct names than be consistent with an existing wrong names. I also believe it's a low easier for people to understand what SSE fields they are changing when their names reflect the data on the write, as opposed to the full name defined in the spec.

poutsma:
We could even consider deprecating `SseEmitter.SseEventBuilder#reconnectTime` in favor of a new `retry` or `reconnectionTime` method. Same for `name`, in favor of `event` or `type`.

sdeleuze:
@poutsma Good point for event name versus type, I was reading an outdated version.

Since we are talking about an `ServerSentEvent` which is already an event, I would be in favor using `type`, `event.event()` would be confusing I think even if it make sense in the serialized data. I have no strong opinion about `reconnectTime` versus `retry` versus `reconnectionTime`.

sdeleuze:
Hum, maybe something to add. Since we are using a `Duration`type unlike in `SseEmitter', I think we could avoid the `Time` prefix. So I would be in favor of something like `retry`, `reconnect` or `reconnection`.

rstoyanchev:
I don't think 100% consistency with SseEmitter is all that important. I do like the idea of aligning with the actual names that appear on the wire so retry and event sound good to me. 

I don't see much wrong with `reconnectTime` to be honest. I find it more descriptive on its own and unqualified although with Duration in the picture, `retry(Duration)` sounds pretty good too. We could use the Java 8 upgrade as a reason to add `retry(Duration)` and deprecate `reconnectTime`.

As for event vs name that was a mistake from the start https://jira.spring.io/browse/SPR-13080. I would be in favor of deprecating it and adding `event(String)`.

poutsma:
> Since we are talking about an ServerSentEvent which is already an event, I would be in favor using type, event.event() would be confusing I think even if it makes sense in the serialized data. I have no strong opinion about reconnectTime versus retry versus reconnectionTime. 

Well, for consistency's sake, we should either use long names or field names, both as defined in the spec. So we either use `reconnectionTime` and `type`, or `retry` and `event`. Mixing them up would confuse people. I vote for using the latter, since the names are shorter and reflect the data on the wire.

The fact that it might be confusing to have an `event` field in a `ServerSentEvent` class is not that concerning to me. These names come straight out of the spec, so any confusion that might cause is not caused by us, but by the spec authors.

>Since we are using a Durationtype unlike in `SseEmitter`, I think we could avoid the `Time` suffix. So I would be in favor of something like `retry`, `reconnect` or `reconnection`.

The latter two names cannot be found in the spec. I prefer to keep the class close to the spec.

poutsma:
Ok, I think we've come to the conclusion that short field names are the way to go. So I believe this PR can be merged.

PR #1154, opened 2016-09-02
Addes static start method to HttpHandler adapters

poutsma:
This commit adds a static "start" method to the Reactor and RxNetty
HttpHandler adapters, to facilitate easy starting from a main method.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorHttpHandlerAdapter.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorHttpHandlerAdapter.java
@@ -16,10 +16,12 @@
 
 package org.springframework.http.server.reactive;
 
+import java.net.InetSocketAddress;
 import java.util.function.Function;
 
 import reactor.core.publisher.Mono;
 import reactor.ipc.netty.http.HttpChannel;
+import reactor.ipc.netty.http.HttpServer;
 
 import org.springframework.core.io.buffer.NettyDataBufferFactory;
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


poutsma:
@rstoyanchev, @sdeleuze, @bclozel: I would love your input on this PR. The idea is to make it a bit easier for people to start a Reactor or RxNetty server from a `public static void main(String[])` method, but without creating a full-fledged abstraction like we did in `HttpServer`. I did not add it to the `ServletHttpHandlerAdapter`, as we can expect users to be more comfortable with the servlet deployment model.

rstoyanchev:
We have as much for copy+paste in the docs now https://github.com/spring-projects/spring-framework/blob/master/src/asciidoc/web-reactive.adoc#manual-bootstrapping. Perhaps we can add it in the docs as well to make it easier to come across. It might even be easier to notice in the Javadoc vs a static start method that looks strange in the adapter.

poutsma:
I think we should do both have the Javadoc as well as the static method. Personally, I would prefer to call a static method than copy-paste code from Javadoc, especially when doing demos, as it's less error prone. But I can also see the reason why other people wouldn't.

As for the static method "looking strange" in the adapter: it's not that strange, as the method actually uses the adapter in its logic. I am happy to put the static method in another class, but I don't think there are better locations. Creating a new class for this seems overkill.

bclozel:
Obviously, I'm not the most objective here - I'd rather have Spring Boot deal with setting up the infrastructure or leave full control to the developers and properly document things.

As our infra will get more complex (TLS, protocols, etc), I suspect we'll have a harder time  providing such facilities.

poutsma:
Ok, I will close this PR then.

PR #1156, opened 2016-09-03
SPR-10502 - PropertyPlaceholder multiple locations resolution

fdesu:
This commit fixes an issue, when in XML defined config there is context:property-resolver with
the value if form of value="${placeholder}" thus it should be resolved with system property. Issue occurs when property already resolved, but appropriate splitting by comma already had passed.

Issue: https://jira.spring.io/browse/SPR-10502

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/config/AbstractPropertyLoadingBeanDefinitionParser.java
+++ b/spring-context/src/main/java/org/springframework/context/config/AbstractPropertyLoadingBeanDefinitionParser.java
@@ -23,6 +23,9 @@ import org.springframework.beans.factory.support.BeanDefinitionBuilder;
 import org.springframework.beans.factory.xml.AbstractSingleBeanDefinitionParser;
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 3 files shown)


fdesu:
Please review that simple fix.
I have also commented appropriate JIRA page.

fdesu:
@snicoll Can you please review and maybe push further this PR?
Thank you in advance.

snicoll:
I am not sure I like that change. I was about to try to resolve the placeholders in the parser itself rather but it looks like the editor may also resolve the array when translating the string to a `Resource`. I am sure @jhoeller has an opinion about the best course of action.

fdesu:
Thanks a lot for your opinion @snicoll. Agreed, with you, but, yes, let's just wait for an opinion from @jhoeller

fdesu:
@snicoll 
I've updated PR with your suggestions, and added some more tests to cover changes.
Could you please take a look?

fdesu:
@jhoeller resolved the original issue with more extensive changes

PR #1158, opened 2016-09-06
Fix typo in asciidoc

nkjackzhang:
Change servlet name in the code to "example", because the servlet name "example" is mentioned below.

### DIFF
--- a/src/asciidoc/web-mvc.adoc
+++ b/src/asciidoc/web-mvc.adoc
@@ -171,7 +171,7 @@ Java EE Servlet configuration in a Servlet 3.0+ environment:
 
 		@Override
 		public void onStartup(ServletContext container) {
-			ServletRegistration.Dynamic registration = container.addServlet("dispatcher", new DispatcherServlet());
+			ServletRegistration.Dynamic registration = container.addServlet("example", new DispatcherServlet());
 			registration.setLoadOnStartup(1);
 			registration.addMapping("/example/*");
 		}


snicoll:
Thanks for the PR but that's not a typo (look at the type of the servlet). `/example` is the mapping of the custom `DispatcherServlet`.

sbrannen:
@snicoll, @nkjackzhang is correct.

Look at the subsequent text:

>  In the preceding example, all requests starting with `/example` will be handled by the
`DispatcherServlet` instance named `example`.

The instance is currently named "dispatcher" (via the `servletName` parameter to `ServletContext.addServlet(String, Servlet)`), and @nkjackzhang's PR changes it to "example" to fix the inconsistency.

kazuki43zoo:
@snicoll, i vote @sbrannen 's comment.

Explanation for xml based configuration at under about 10 lines says as follow:

> Below is the web.xml equivalent of the above code based example:
>
> ```xml
> <web-app>
>     <servlet>
>         <servlet-name>example</servlet-name>
>         <servlet-class>org.springframework.web.servlet.DispatcherServlet</servlet-class>
>         <load-on-startup>1</load-on-startup>
>     </servlet>
> 
>     <servlet-mapping>
>         <servlet-name>example</servlet-name>
>         <url-pattern>/example/*</url-pattern>
>     </servlet-mapping>
> 
> </web-app>
> ```

snicoll:
@kazuki43zoo sorry, I missed that context. This has been merged in `master` and `4.3.x`. Thanks!

PR #1159, opened 2016-09-07
Fix docs

neetkee:
Old links were not available.


bclozel:
Merged with 42f101fb5bfa3 and 5496f11f3c550

PR #1162, opened 2016-09-07
Change View.render method to take Map<String, ?>

poutsma:
This commit introduces support for model & view rendering in web.reactive.function. The first three commits or this PR contain changes in web.reactive.result.view necessary to facilitate this.

@rstoyanchev: could you take a look at this, especially the first three commits? If you are OK with the changes there, then I will make a JIRA issue to link to.

### DIFF
--- a/spring-web-reactive/src/main/java/org/springframework/web/reactive/result/view/AbstractView.java
+++ b/spring-web-reactive/src/main/java/org/springframework/web/reactive/result/view/AbstractView.java
@@ -28,9 +28,7 @@ import reactor.core.publisher.Mono;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.http.MediaType;
-import org.springframework.ui.ModelMap;
 import org.springframework.util.Assert;
-import org.springframework.web.reactive.HandlerResult;
 import org.springframework.web.server.ServerWebExchange;
 
 /**
@@ -84,28 +82,28 @@ public abstract class AbstractView implements View, ApplicationContextAware {
 		return applicationContext;
 	}
 
-
 	/**
 	 * Prepare the model to render.
-	 * @param result the result from handler execution
+	 * @param model Map with name Strings as keys and corresponding model
+	 * objects as values (Map can also be {@code null} in case of empty model)
 	 * @param contentType the content type selected to render with which should
 	 * match one of the {@link #getSupportedMediaTypes() supported media types}.
 	 * @param exchange the current exchange
 	 * @return {@code Mono} to represent when and if rendering succeeds
 	 */
 	@Override
-	public Mono<Void> render(HandlerResult result, MediaType contentType,
+	public Mono<Void> render(Map<String, ?> model, MediaType contentType,
 			ServerWebExchange exchange) {
 
 		if (logger.isTraceEnabled()) {
-			logger.trace("Rendering view with model " + result.getModel());
+			logger.trace("Rendering view with model " + model);
 		}
 
 		if (contentType != null) {
 			exchange.getResponse().getHeaders().setContentType(contentType);
 		}
 
-		Map<String, Object> mergedModel = getModelAttributes(result, exchange);
+		Map<String, Object> mergedModel = getModelAttributes(model, exchange);
 		return renderInternal(mergedModel, exchange);
 	}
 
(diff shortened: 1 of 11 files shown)


rstoyanchev:
@poutsma, overall the right changes with a couple of comments.

For the contentType parameter and the charset there should be a default charset property in AbstractView and also in ViewResolverSupport so it can be configured conveniently. The default value should probably be UTF-8 rather than ISO-8859-1 as we don't need to stick to Servlet defaults and even the Servlet spec is [discussing changes](https://java.net/projects/servlet-spec/lists/jsr369-experts/archive/2016-09/message/3).

For ModelAndView do we really need a builder? The current ModelAndView seems perfectly adequate with its constructors and chained methods. It results in shorter lines and there is no need for a build step. A couple of other smaller details are the ability to create a ModelAndView without a view which could come up if the method signature is ModelAndView but the controller returns conditionally with different outcomes. The second is the ability to add an attribute without specifying a name, i.e. falling back on conventions. In short I see no motivation to change much the present ModelAndView and the consistency would be helpful as well.

poutsma:
>For the contentType parameter and the charset there should be a default charset property in AbstractView and also in ViewResolverSupport so it can be configured conveniently. The default value should probably be UTF-8 rather than ISO-8859-1 as we don't need to stick to Servlet defaults and even the Servlet spec is discussing changes.

Ok, I will try and add those.

>For ModelAndView do we really need a builder? The current ModelAndView seems perfectly adequate with its constructors and chained methods. It results in shorter lines and there is no need for a build step.

The goal was to make this new `ModelAndView` immutable, to better facilitate functional programming and serverless architectures. That meant getting rid of the setters and the `addObject` methods. Without those, `ModelAndView` becomes less pleasant to use (especially with more than one object in the model), so I thought that having a builder would be better:

#### Constructor
```java
Map<String, Object> model = new HashMap<>();
model.put("foo", "bar");
model.put("baz", "qux");
ModelAndView mav = new ModelAndView("view", model);
```

#### Builder
```java
ModelAndView mav = ModelAndView.viewName("view")
	.modelAttribute("foo", "bar")
	.modelAttribute("baz", "qux").build();
```

I also thought that a builder would be preferable to having eight overloaded constructors (which is what `ModelAndView` currently has).

>A couple of other smaller details are the ability to create a ModelAndView without a view which could come up if the method signature is ModelAndView but the controller returns conditionally with different outcomes.

Couldn't that use case be supported by just returning a model from the controller? I know that we had this variant in `ModelAndView` before, but I believe that was done mostly to facilitate the inheritance-based MVC, where every `Controller` had to return `ModelAndView`. I doubt if it's still necessary in @web.reactive with its flexible signatures. Requiring a view(name) in `ModelAndView` certainly does make a lot of sense in `web.reactive.function`, where there is no method signature reflection going on, so we cannot deduce the view name.

>The second is the ability to add an attribute without specifying a name, i.e. relying on {{Conventions.getVariableName(attributeValue)}}. I think both of those are still relevant.

I will add those.

poutsma:
@rstoyanchev: I've updated the PR to fix the issues you mentioned. Can you take another look?

rstoyanchev:
Okay I understand much better now the motivation for the ModelAndView changes. 

As we have been discussing offline, between generics (e.g. `ModelAndView<String>`, `ModelAndView<View>`) and the builder, the result is more tailored for use on the functional side and is much less convenient for use in annotated methods. The generics do bring benefits on the functional side for further composition but have no benefit on the annotations side. Also the builder seems like too much as opposed to:

````java
return new ModelAndView("view", "foo", "bar");
````
or more generally this (which could also be immutable):
````java
return new ModelAndView("view", new ModelMap("foo", "bar").addAttribute("baz", "qux"));
```
Perhaps on the functional side we don't have to stick to ModelAndView and use some alternative terminology along the lines of "render", "template", or other that fits well next to the other methods in the BodyBuilder of Response.

> Couldn't that use case be supported by just returning a model from the controller? I know that we had this variant in ModelAndView before, but I believe that was done mostly to facilitate the inheritance-based MVC, where every Controller had to return ModelAndView.

ModelAndView could still be preferred in the method signature as the return type but the method could return without a view name (e.g. after validation failure) or with both ModelAndView on redirect providing a fresh model. In any case if it is possible to return no view at all in general, then no reason for ModelAndView to be more restrictive.

poutsma:
I will remove the latter two commits from this PR, so that the non-controversial parts can be merged. In a separate branch, I will then introduce a new "template rendering" type (similar to the `ModelAndView` contained in commit 4 of this PR) in `web.reactive.function`, but with a different name. This way, w.r.f can have the immutable, builder variant, while we can continue to use the constructor-based variant in the annotation-based code.

poutsma:
@rstoyanchev Are you ok with merging this PR in its current form?

rstoyanchev:
:+1:

poutsma:
JIRA is at https://jira.spring.io/browse/SPR-14685

PR #1166, opened 2016-09-12
Is the first word "method" needless?

nkjackzhang:
From context, It seems that the first word "method" is needless,


sbrannen:

Thanks

PR #1167, opened 2016-09-12
Request/ResponseBodyAdvice based solution to provide hints to codecs

sdeleuze:
In order to continue the discussion began in [SPR-14557](https://jira.spring.io/browse/SPR-14557) and PR https://github.com/spring-projects/spring-framework/pull/1163, here is a proposal to provide hints like `@JsonView` from annotated controllers to codecs.

I have proposed this mechanism because I think:
- We need a pluggable mechanism
- I guess `Request/ResponseBodyAdvice` will still be needed in Reactive world, and hints are related to request/response body
- I am not in favor using `ResolvableType` source because to do so we need to pass the stream type and not just the element type to the codecs, and that create some issues by exposing specific Reactive type information (dealing with `Foo` versus Flux<Foo>`versus`Observable<Foo>`) at codec level)
- I am in favor of handling hints in a consistent way for our 3 main use cases : annotation based controller model, new functional model introduced by [SPR-12954](https://jira.spring.io/browse/SPR-12954) and the new Reactive HTTP client at codec level.

A possible alternative is to provide a lighter pluggable mechanism specific to providing hints.

Thanks in advance for your feedbacks.

### DIFF
--- a/spring-web-reactive/src/main/java/org/springframework/web/reactive/config/WebReactiveConfiguration.java
+++ b/spring-web-reactive/src/main/java/org/springframework/web/reactive/config/WebReactiveConfiguration.java
@@ -29,6 +29,7 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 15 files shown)


sdeleuze:
I am going to propose a new PR based on @rstoyanchev feedback.

PR #1168, opened 2016-09-12
Add Resource.readableChannel()

poutsma:
Added getReadableByteChannel() to Resource, which returns a
java.nio.ReadableByteChannel. The default implementation uses
Channels.newChannel() to create a channel based on what is returned from
getInputStream(). Subclasses have more effecient, file-based
implementations.

Issue: SPR-14698

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/codec/ResourceEncoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ResourceEncoder.java
@@ -17,7 +17,7 @@
 package org.springframework.core.codec;
 
 import java.io.IOException;
-import java.io.InputStream;
+import java.nio.channels.ReadableByteChannel;
 import java.util.Map;
 
 import reactor.core.publisher.Flux;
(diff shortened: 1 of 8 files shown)


poutsma:
For your consideration, @jhoeller and @rstoyanchev.

jhoeller:
Hmm, maybe shorten the method name to `readableChannel()`, along the lines of our existing `contentLength()` and `lastModified()` methods? Also, let's position it right after `getFile()`...

poutsma:
@jhoeller: I've updated the PR accordingly. Can you take another look?

jhoeller:
Thumbs up from my side, except for the now outdated title and commit comment :-)

@rstoyanchev, looking good to you as well? If so, let's get it in.

poutsma:
> Thumbs up from my side, except for the now outdated title and commit comment :-)

Ah, of course, I will change that as well.

rstoyanchev:
This is now merged, thanks @poutsma!

PR #1169, opened 2016-09-13
SPR-14705 Add support for CommonsVfs resources.

ptahchiev:


nigelsim:
Is there a problem using the getSize method for getting the content length?
    @Override
    public long contentLength() throws IOException {
        return this.fileObject.getContent().getSize();
    }

snicoll:
@ptahchiev thanks for the PR but there hasn't been a lot of traction of the community and the related issue was closed as a result. I am going to close this as well.

PR #1171, opened 2016-09-13
Improve exception handling when clients disconnect.

violetagg:
Issue: SPR-14538

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractResponseBodyFlushProcessor.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractResponseBodyFlushProcessor.java
@@ -74,8 +74,8 @@ abstract class AbstractResponseBodyFlushProcessor implements Processor<Publisher
 
 	@Override
 	public final void onError(Throwable t) {
-		if (logger.isErrorEnabled()) {
-			logger.error(this.state + " onError: " + t, t);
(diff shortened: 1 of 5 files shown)


rstoyanchev:
@violetagg regarding the logging and handling of error signals in the HttpHandler adapters (like `ServletHttpHandlerAdapter` and `UndertowHttpHandlerAdapter`), it's true that won't ever be used if an `ExceptionHandlingWebHandler` is installed which is the case by default with the `WebHttpHandlerBuilder`. However in theory one may not use that builder or may install other exception handlers and so the presence of `ExceptionHandlingWebHandler` is not 100% guaranteed.

Unless there is an issue with keeping that logic in place as a stop-gap measure, also at ERROR level since it really should not happen. I would even go add similar logging and 500 response status to the adapters for Reactor and RxNetty.

WDYT?

violetagg:
Hi,

I reverted the changes in ```ServletHttpHandlerAdapter ``` and ```UndertowHttpHandlerAdapter ```. Should we track the changes in Reactor and RxNetty here or in another issue?

Thanks,
Violeta

rstoyanchev:
Thanks. Fine by me to do it here. Minor in terms of changes and very closely related.

violetagg:
Can you check the new changes?

Thanks,
Violeta

rstoyanchev:
This is now merged in. Thank!

PR #1173, opened 2016-09-14
Add WebApplicationInitializers for Web Reactive

poutsma:
This commit introduces three new WebApplicationInitializers for use with
Spring Web Reactive:
- The AbstractServletHttpHandlerAdapterInitializer registers a
  ServletHttpHandlerAdapter that wraps a user-provided HttpHandler.
- The AbstractDispatcherHandlerInitializer registers a
  ServletHttpHandlerAdapter that wraps a DispatcherHandler (or any
  WebHandler). The handler is provided with an application context.
- The AbstractAnnotationConfigDispatcherHandlerInitializer is a
  subclass of the above, creating an
  AnnotationConfigWebApplicationContext based on provided `@Configuration`
  classes.

Issue: SPR-14713

### DIFF
--- a/spring-web-reactive/src/main/java/org/springframework/web/reactive/support/AbstractAnnotationConfigDispatcherHandlerInitializer.java
+++ b/spring-web-reactive/src/main/java/org/springframework/web/reactive/support/AbstractAnnotationConfigDispatcherHandlerInitializer.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.reactive.support;
+
+import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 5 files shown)


poutsma:
For your consideration, @jhoeller and @rstoyanchev.

poutsma:
Updated to reflect your comments, @rstoyanchev: 

 - The two dispatcher-handler initializers no longer subclasses from AbstractContextLoaderInitializer, so there is no root context.
 - The PR now uses `ApplicationContext` instead of `WebApplicationContext`.

rstoyanchev:
Should we have a default implementation of {{getServletMappings()}} to return "/"? That should be good enough for the common case.

Aside from that looks good to me!

poutsma:
We don't have that in the existing `AbstractDispatcherServletInitializer`, which this class is based on. I decided to keep it consistent.

rstoyanchev:
Here is my reasoning against consistency in that regard.

On the reactive side, the Servlet API and its configuration model are not front and center. The Servlet runtime is the main thing of interest. In order to run on it you need `ServletHttpHandlerAdapter` to adapt to Reactive Streams, no more no less. Having other Servlets should be an exceptional situation rather than a means of solving common scenarios. Perhaps this will be easier to see in hindsight once we expose equivalent means (support for context paths at least).

Another motivation for my comment is to keep things simple in the path. Similar to the discussion around the Servlet contextPath in #1174, dealing with Servlet mappings "underneath", yet not exposed in the programming model would probably make things even more challenging. Especially since there is no API to introspect the current Servlet mapping for a request. 

I would even consider switching from `String[]` to just `String` for a single, prefix-based,  optional Servlet mapping. We could then have the option to explicitly set a property on `ServletHttpHandlerAdapter` from the ServletContextInitializer to let it know how it is mapped.

poutsma:
Third time's a charm! Here is an updated version with a default, single servlet mapping which defaults to `/`.

rstoyanchev:
Merged!

PR #1174, opened 2016-09-14
Consider servlet context path in path matching

poutsma:
Prior to this commit, the reactive HttpRequestPathHelper used the path
as returned from HttpRequest.getURI().getRawPath() as lookup path.
However, when running in a servlet environment, this lookup path
represents the entire path (including servlet context), which is not
desirable.

This commit adds a getPath() method to HttpRequest, defaulting to return
getURI().getRawPath(), but with an override in ServletServerHttpRequest
to return the proper lookup path (obtained from UrlPathHelper).

Issue: SPR-14714

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,5 +39,15 @@ public interface HttpRequest extends HttpMessage {
 	 * @return the URI of the request (never {@code null})
 	 */
 	URI getURI();
+
+	/**
+	 * Return the path of the request.
+	 * <p>Default implementation returns {@link URI#getRawPath()} of the URI provided by
+	 * {@link #getURI()}.
+	 * @return the path of the request
+	 */
+	default String getPath() {
+		return getURI().getRawPath();
+	}
 
 }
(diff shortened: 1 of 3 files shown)


poutsma:
For your consideration, @rstoyanchev.

poutsma:
I've changed the PR to no longer use the UrlPathHelper, @rstoyanchev. Could you take another look?

rstoyanchev:
Thanks @poutsma. This is better but on further through I have concerns still. 

The Javadoc on `ServletServerHttpRequest.getPath()` now properly explains that it removes the contextPath. Yet it's inconsistent, if only a nuisance but also potentially problematic, to discover that `request.getPath()` looks like a shortcut for `request.getUri().getPath()` but isn't. It could also be misleading in a more subtle way too considering that `URI.getPath()` returns a decoded path whereas we use `getRawPath` internally.

What do you think about adding a `getContextPath` method to `ServerHttpRequest` instead? It would expose any contextPath at which the application may be installed at the HTTP runtime level which for non-Servlet containers would typically return an empty string but who knows may be one of the other runtimes may eventually support a contextPath. The `HttpRequestPathHelper` would then use this method in its `getLookupPathForRequest` implementation to leave out the `ServerHttpRequest` contextPath.

If and when we do introduce our own concept of a contextPath within Spring Web Reactive (i.e. multiple apps at separate context paths) by definition that would have to be managed above the `HttpHandler` level. At that time `ServerWebExchange` would probably have a `getContextPath` method that incorporates the contextPath we manage plus any that might come from the HTTP runtime.

I think that would be transparent and flexible enough vs the only other feasible alternative I can think of to completely remove the Servlet container's contextPath from the URI and add an `X-Forwaded-Prefix` header. We would eventually be forced to look up the very `X-Forwarded-Prefix` header we created in order to create a proper externally facing link. Why hide it when we know we will need in the same JVM.

PR #1175, opened 2016-09-14
SPR-14664 Add byte range requests support

bclozel:
Add byte range requests support for static resources in the reactive
web project.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ResourceUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ResourceUtils.java
@@ -18,11 +18,16 @@ package org.springframework.util;
 
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.util.OptionalLong;
+
+import org.springframework.core.io.InputStreamResource;
+import org.springframework.core.io.Resource;
 
 /**
  * Utility methods for resolving resource locations to files in the
@@ -382,6 +387,25 @@ public abstract class ResourceUtils {
 	 */
 	public static void useCachesIfNecessary(URLConnection con) {
 		con.setUseCaches(con.getClass().getSimpleName().startsWith("JNLP"));
+	}
+
+	/**
+	 * Determine, if possible, the contentLength of the given resource
+	 * without reading it.
+	 * @param resource the resource instance
+	 * @return the contentLength of the resource
+	 */
+	public static OptionalLong contentLength(Resource resource) {
+		// Don't try to determine contentLength on InputStreamResource - cannot be read afterwards...
+		// Note: custom InputStreamResource subclasses could provide a pre-calculated content length!
+		if (InputStreamResource.class != resource.getClass()) {
+			try {
+				return OptionalLong.of(resource.contentLength());
+			}
+			catch (IOException ignored) {
+			}
+		}
+		return OptionalLong.empty();
 	}
 
 }
--- a/spring-core/src/main/java/org/springframework/core/codec/ResourceEncoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ResourceEncoder.java
@@ -50,7 +50,7 @@ public class ResourceEncoder extends AbstractSingleValueEncoder<Resource> {
 	}
 
 	public ResourceEncoder(int bufferSize) {
-		super(MimeTypeUtils.ALL);
+		super(MimeTypeUtils.APPLICATION_OCTET_STREAM, MimeTypeUtils.ALL);
 		Assert.isTrue(bufferSize > 0, "'bufferSize' must be larger than 0");
 		this.bufferSize = bufferSize;
 	}
--- a/spring-core/src/main/java/org/springframework/core/codec/ResourceRegionEncoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ResourceRegionEncoder.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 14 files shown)


spring-core/src/main/java/org/springframework/util/ResourceUtils.java
```diff
@@ -384,4 +389,23 @@ public static void useCachesIfNecessary(URLConnection con) {
...
+		if (InputStreamResource.class != resource.getClass()) {
+			try {
+				return OptionalLong.of(resource.contentLength());
+			}
+			catch (IOException ignored) {
+			}
+		}
+		return OptionalLong.empty();
+	}
+
```
rstoyanchev: Could we consider a new method on Resource for this? The fact that a utility method takes a Resource and calls its property is a sign. It looks like a variant of the current `getContentLength()` that returns the content length optionally if known. So perhaps `OptoinalLong getContentLengthIfKnown()` with a default implementation returning empty.

/cc @jhoeller, @poutsma

rstoyanchev:
Only one minor comment about on ResourceUtils. Aside from that looks very good and once the `ServerHttpMessageWriter` is in place it will be even more encapsulated and re-usable from both the `ResourceWebHandler` and for `ResponseBody` methods.

spring-core/src/main/java/org/springframework/util/ResourceUtils.java
jhoeller: I don't mind an addition to `Resource`, in particular now where we can do default methods. However, I feel that we're overusing `Optional`. On such core abstractions, can we rather do it in a more traditional style through a `Long` return type which may simple be `null` if not known?

To be clear, using `Optional` in a functional API is perfectly sensible. I just don't want it to creep into Java interfaces at a lower level, in particular not in existing interfaces with a traditional design mindset in other methods... and not really in new interfaces in the same packages either.

rstoyanchev: For consistency with the existing method, possibly returning `long` with -1 to indicate unknown. It looks like some (e.g. UrlConnection) can return -1.

bclozel:
I just tried reusing the `ServerHttpMessageWriter` and it seems we can't really move the `HttpRange` information from `ResourceWebHandler` because:

* In `ResourceWebHandler`, in case an `IllegalArgumentException` is thrown while parsing the HTTP range  information, we not only need to set an HTTP status (which could be possible with a `ResponseStatusException`), but also add an HTTP response header.
* In `ResourceWebHandler`, we're setting the response status to `HttpStatus.PARTIAL_CONTENT` if this is a byte range request; we can't do that from within an `HttpMessageWriter` which only holds a reference to a `ReactiveHttpOutputMessage` (no `setStatus` method).

Given that `ResourceHttpMessageWriter` already delegates to `ResourceRegionHttpMessageWriter`, which delegates to the `ResourceRegionEncoder`, I'm wondering if adding another level of wrapping and hints is really helping the situation in the end.

rstoyanchev:
@sdeleuze in light of @bclozel's comment I'd like to propose something I was already thinking about. Instead of `ServerHttpMessageWriter` having a method to prepare hints, it should have a method to write except taking all the inputs that are available on the server side (e.g. the type of the stream, the ServerHttpRequest and ServerHttpResponse). It can then prepare hints and delegate to the next writer, or compose on the input/output stream, set the status, short-circuit the process, and so on.

bclozel:
@rstoyanchev @sdeleuze to me, this `ServerHttpMessageWriter` contract would look more and more like a full `WebHandler` with a few things added on top. This makes sense for this use case, but I'm not sure I'm comfortable adding this right now (I've explained in my previous comment the level of indirection we're talking about if we do that).

I'd rather duplicate a few lines of code in `ResourceWebHandler` and a `HandlerResultHandler` implementation rather than introducing that abstraction if we're not confident about it. Will it be useful for more use cases?

rstoyanchev:
Well I'm not suggesting it because of the Resource use case alone. I was independently thinking of making this suggestion. Plus we are not talking about a major change to what's already in place that we are comfortable with. I'm simply suggesting to replace `resolveHints` with a `write`, so pretty much the same story but a little more control. 

If we accept that an `HttpMessageWriter` is intentionally limited to what is relevant to both client and server and that a `ServerHttpMessageWriter` is what `HttpMessageWrite` could be if we could pass to it the input available on the server side, then the contracts should be largely the same with a few extra inputs. In short I think resolvHints constrains things a little more than necessary, so you can only produce hints given server-side inputs, and I don't see a strong reason for that.

sdeleuze:
@rstoyanchev Is your proposal to replace `ServerHttpMessageWriter#resolveWriteHints()` by something like
`Mono<Void> ServerHttpMessageWriter#write(Publisher<? extends T> inputStream, ResolvableType streamType, ResolvableType elementType, MediaType mediaType, ServerHttpRequest request, ServerHttpResponse response, Map<String, Object> hints)` where `hints` could be used to read existing hints as well as adding eventually new ones before delegating to the nested writer?

rstoyanchev:
@sdeleuze yes that's what I was thinking about.

sdeleuze:
Here is a draft just on the writer to discuss: https://github.com/sdeleuze/spring-framework/commit/b17f596f60fac7af458e6cf5b3cfcb911d0c3744

Main points are:
 - `hints` are not provided anymore to `canDecode()` / `canWrite()` because we don't have a dedicated method to get them anymore (not an issue for my POV, hint should just be used to customize the output)
 - `ServerHttpMessageWriter` now has a more powerful write method
 - For `AbstractServerHttpMessageWriter` I made various tries because letting implementers directly implement `write()` or `writeInternal()` lead to not so obvious and error prone implementations and I was not happy with that. I ended-up with this `beforeWrite()` solution that allows easily to provide hints or set the status. For really advanced use cases, it is still possible to override `write()` if needed.

Any thoughts ?

rstoyanchev:
Looks good to me, even a bit simpler with hints now prepared and managed internally.

sdeleuze:
These improvements are now in master.

@bclozel Let me know if there is any issue while rebasing your changes.

PR #1184, opened 2016-09-24
HttpClient instance should be shared for each call

m5k10:
I made simple mvc controller like this:

``` java
private final WebClient webClient;

    public ProxyController() {
        ClientHttpConnector httpConnector = new MyReactorClientHttpConnector(ClientOptions.create());
        webClient = new WebClient(httpConnector);
    }

    @RequestMapping("/proxy")
    @ResponseBody
    @ResponseStatus(code = HttpStatus.OK)
    public Mono<String> proxy() {
        return webClient
            .perform(get("http://localhost/static.json").accept(APPLICATION_JSON))
            .extract(ResponseExtractors.body(String.class));
    }
}
```

After few seconds running app start to throw too many open file exception (lsof command revealed that app use many file descriptors with name 'eventpoll'). 
This patch fix this problem, but I think ClientHttpConnector and WebClient needs to implement close method that will delegate to HttpClient#close

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpConnector.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpConnector.java
@@ -21,6 +21,7 @@ import java.util.function.Function;
 
 import reactor.core.publisher.Mono;
 import reactor.ipc.netty.config.ClientOptions;
+import reactor.ipc.netty.http.HttpClient;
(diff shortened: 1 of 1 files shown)


bclozel:
I've merged this in 8143e5f; but keep an eye on [SPR-14750](https://jira.spring.io/browse/SPR-14750) as changes are coming to the Reactor Netty client which will remove the existing `shutdown()` method.

PR #1187, opened 2016-09-28
ExceptionHandlingWebHandler should log exception at "error" level

m5k10:
This handler send status 500, so in logs should be information "why".


sdeleuze:
Thanks for this PR, I have done this changes + other log refinement as part of https://github.com/spring-projects/spring-framework/commit/324c310cbd3a95da73d36a6f868a538742a10c4f.

PR #1192, opened 2016-10-02
SPR-14597: Fix CachingResourceResolver's handling of same path with different locations

heffebaycay:
When calling `CachingResourceResolver#resolveUrlPathInternal` with the same `resourceUrlPath` param but different list of search locations, the same cache key will be assigned to both resources.

This effectively breaks Spring's `VersionResourceResolver`. Considering two calls for the same `ressourceUrlPath` 'sprite.png', with the following locations:
- First call: class path resource [static/alpha/style.css]
- Second call: class path resource [static/beta/style.css]

The first call will resolve to the alpha 'sprite.png' ('sprite-1cf43f6ba5cbc71b4b2d040f2a358f3e.png') and cache this value under a key that only depends on 'sprite.png'.

The second call will use the cached value and thus won't resolve to the beta 'sprite.png' file.

This commit changes the way the cache keys are generated. Cache keys now depends of both the `ressourceUrlPath` and the `locations` parameters.

Issue: [SPR-14597](https://jira.spring.io/browse/SPR-14597)

Note: I have signed Pivotal Contributor License Agreement.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/resource/CachingResourceResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/resource/CachingResourceResolver.java
@@ -19,10 +19,13 @@ package org.springframework.web.servlet.resource;
 import java.util.List;
 import javax.servlet.http.HttpServletRequest;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import org.springframework.cache.Cache;
 import org.springframework.cache.CacheManager;
 import org.springframework.core.io.Resource;
 import org.springframework.util.Assert;
+import org.springframework.util.DigestUtils;
 
 /**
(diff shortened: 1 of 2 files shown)


bclozel:
I'm closing this PR in favor of #1199, since hashing the list of resource locations into the cache key may not be consistent nor efficient. I've improved the way we're resolving/rewriting links within resources to use absolute paths instead of relative paths.

Your PR was very helpful and underlines the limitations of the current implementation.
Don't hesitate to voice your opinion on the dedicated issue or on #1199 directly.

Thanks!

heffebaycay:
Alright, thanks for improving the fix!

rstoyanchev:
Indeed the Resource description does not overcome the issue for some Resource implementations like `ServletContextResource` for example.

PR #1193, opened 2016-10-03
Log unresolved exceptions at error level

sdeleuze:
This PR change the log level of unresolved exceptions from debug to error, in order to avoid the "no exception displayed by default" behaviour that we have currently. Are you ok with this change @rstoyanchev?


rstoyanchev:
@sdeleuze yes. Might as well also change from debug to error in the various HTTP handler adapters (e.g. ReactorHttpHandlerAdapter), see https://github.com/spring-projects/spring-framework/commit/b2ccdfbc9e112e48f001b0dffa1280e6e7fe7122. 

Also PR #1187 is for the same request as this.

sdeleuze:
Merged (https://github.com/spring-projects/spring-framework/commit/324c310cbd3a95da73d36a6f868a538742a10c4f)

PR #1194, opened 2016-10-03
Handle async operation events

violetagg:
Problem:
The following exception is observed when there is async operation
timeout: 

```
java.lang.IllegalStateException: It is invalid to call isReady() when the response has not been put into non-blocking mode
    at org.apache.coyote.Response.isReady(Response.java:616) ~[tomcat-embed-core-8.5.5.jar:8.5.5]
    at org.apache.catalina.connector.OutputBuffer.isReady(OutputBuffer.java:677) ~[tomcat-embed-core-8.5.5.jar:8.5.5]
    at org.apache.catalina.connector.CoyoteOutputStream.isReady(CoyoteOutputStream.java:155) ~[tomcat-embed-core-8.5.5.jar:8.5.5]
    at org.springframework.http.server.reactive.ServletServerHttpResponse$ResponseBodyProcessor.isWritePossible(ServletServerHttpResponse.java:169) ~[spring-web-5.0.0.BUILD-SNAPSHOT.jar:5.0.0.BUILD-SNAPSHOT]
```

Current Implementation:
The async operation events sent by the web container are not
propagated to the internal implementation.
When timeout/error happens and if the application does not complete the
async operation, the web container will complete it. At that point if
the application tries to read/write, the operation will fail with an
exception (above) that there is not async operation started.

Proposed Solution:
When there is an async operation timeout or the async operation failed
to complete, calls to
- AbstractRequestBodyPublisher.onError,
- AbstractResponseBodyProcessor.onError,
- AbstractResponseBodyFlushProcessor.onError

will be made. As a result of these calls the async operation will be
completed and no more invocations of read/write will be made.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractResponseBodyFlushProcessor.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractResponseBodyFlushProcessor.java
@@ -120,7 +120,7 @@ abstract class AbstractResponseBodyFlushProcessor implements Processor<Publisher
 
 	}
 
-	private void cancel() {
+	protected void cancel() {
 		this.subscription.cancel();
 	}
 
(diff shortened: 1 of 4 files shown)


rstoyanchev:
This is now merged, thanks! Note that I've applied an extra commit with minor, style-related changes.

PR #1197, opened 2016-10-05
SPR-14779: providing type-safe static methods like Collections.emptyLi…

hakamairi:
…st() to Comparators.


jhoeller:
See the JIRA issue: I've taken a slightly different approach.
Thanks for raising this and for the pull request in any case!

PR #1198, opened 2016-10-07
SPR-14745 SchedulerFactoryBean's setOverwriteExistingJobs does not reliably work in a cluster

apixandru:
In relation to [SPR-14745](https://jira.spring.io/browse/SPR-14745) 

While commit doesn't actually fix the issue, it extracts the way that the jobs, triggers and calendars are being registered so that they can be registered differently in a subclass. This can be used as a workaround until the issue is fixed at some later point.

I tried making as few changes as I could in order to register the triggers in another class while keeping everything 100% backwards compatible.

It would be great if this was backported to 4.3.x so that we can subclass SchedulerFactoryBean and override the way that the triggers are created to not run into the cluster race condition.


apixandru:
License Agreement.

PR #1201, opened 2016-10-12
Add reactive multipart request support

sdeleuze:
This draft PR is a first Multipart support implementation here for high-level feedbacks and discussion.

It does not include RequestMapping level support, integration tests do not work yet with Reactor and RxNetty and `NioMultipartResolver` need to be rewritten using `Flux.create()` instead of `ReplayProcessor`.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/ServerWebExchange.java
+++ b/spring-web/src/main/java/org/springframework/web/server/ServerWebExchange.java
@@ -24,6 +24,7 @@ import java.util.function.Consumer;
 
 import reactor.core.publisher.Mono;
 
+import org.springframework.http.codec.multipart.Part;
 import org.springframework.http.server.reactive.ServerHttpRequest;
 import org.springframework.http.server.reactive.ServerHttpResponse;
 import org.springframework.util.MultiValueMap;
@@ -81,6 +82,12 @@ public interface ServerWebExchange {
 	 * {@code "application/x-www-form-urlencoded"} or an empty map.
 	 */
 	Mono<MultiValueMap<String, String>> getFormData();
+
+	/**
+	 * Return the form parts from the body of the request or an empty {@code Mono}
+	 * if the Content-Type is not "multipart/form-data".
+	 */
+	Mono<MultiValueMap<String, Part>> getMultipartData();
 
 	/**
 	 * Return a combined map that represents both
--- a/spring-web/src/test/java/org/springframework/http/server/reactive/MultipartIntegrationTests.java
+++ b/spring-web/src/test/java/org/springframework/http/server/reactive/MultipartIntegrationTests.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2002-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.http.server.reactive;
+
+import java.net.URI;
+import java.util.Optional;
+
+import static org.junit.Assert.*;
+import org.junit.Test;
+import reactor.core.publisher.Mono;
+
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.buffer.DataBuffer;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.http.codec.multipart.Part;
+import org.springframework.web.server.ServerWebExchange;
+import org.springframework.web.server.WebHandler;
+import org.springframework.web.server.adapter.HttpWebHandlerAdapter;
+
+public class MultipartIntegrationTests extends AbstractHttpHandlerIntegrationTests {
+
+	@Override
+	protected HttpHandler createHttpHandler() {
+		HttpWebHandlerAdapter handler = new HttpWebHandlerAdapter(new CheckRequestHandler());
+		return handler;
+	}
+
+	@Test
+	public void getFormParts() throws Exception {
+		RestTemplate restTemplate = new RestTemplate();
+		RequestEntity<MultiValueMap<String, Object>> request = RequestEntity
+				.post(new URI("http://localhost:" + port + "/form-parts"))
+				.contentType(MediaType.MULTIPART_FORM_DATA)
+				.body(generateBody());
+		ResponseEntity<Void> response = restTemplate.exchange(request, Void.class);
+		assertEquals(HttpStatus.OK, response.getStatusCode());
+	}
+
+	private MultiValueMap<String, Object> generateBody() {
+		HttpHeaders fooHeaders = new HttpHeaders();
+		fooHeaders.setContentType(MediaType.TEXT_PLAIN);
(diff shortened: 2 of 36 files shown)


sdeleuze:
I have updated this PR witgh the following changes:
 - Tests now run fine with Reactor Netty and RxNetty
 - `getAllParts()` now caches the resulting mono to allow multiple invocation
 - There is a single `Part` object suitable usable for file and non file parts
 - We now use `Flux.create()` instead of `ReplayProcessor`
 - `NioPart#transferTo()` now uses zero copy  when possible (when a temporary file has been previously created)

@rstoyanchev Could you please make a high-level review of this commit?

rstoyanchev:
@sdeleuze, apologies for suggesting it but ServerHttpRequest doesn't seem like the right level for exposing multipart support. Having MultipartResolver take ServerHttpRequest as input is already an indication that it's a higher level concern. Then MultipartResolver has to be injected into every request implementation, which is rather repetitive and artificial. It's also a cyclical dependency even at the package level since the http package is logically at a level below the web package and should not depend on anything from it.

It seems to be a better fit for being exposed at the ServerWebExchange level and to avoid a circular dependency again, MultipartResolver and Part should be in server while with maybe a server.multipart sub-package for the implementations. For initialization a MultipartResolver could be a field of HttpWebHandlerAdapter and passed into every created ServerWebExchange. This is very similar to WebSessionManager and WebSession.

For the caching, the cache operator should really be called at construction time or otherwise concurrent calls to getParts can each create a separate pipeline. See how DefaultWebExchange does it for WebSession.

For when the request is not multipart, could we be more lenient and return an empty stream from getParts and a Mono with empty map from getAllParts? It removes the need to check isMultipartRequest first.

Do we need Part.getContentType if we have getHeaders since there are plenty of methods already on part? 

I haven't looked into the implementation library but I presume it is non-blocking indeed? It seems to be returning java.io.InputStream.

sdeleuze:
No worry indeed, I agree that `ServerHttpRequest` is not the right level to implement that for the reasons you mentioned. I also agree with your remarks on caching and being more lenient if the request is not multipart.

We can remove `Part.getContentType()` since this is just a shortcut to `Part.getHeaders().getContentType()` (with maybe a Javadoc hint to help users for file parts).

Currently I think JSON parts will be natively treated by the underlying library like a part with a `String` based value since it has no filename header. Even if in `NioPart` I allowed to get the content both as a `String` or as a `Publisher<DataBuffer>`, that could be not optimal.

Not sure if that makes sens, but should we add some `HttpMessageReader` based capabilities with some `<T> Flux<T> getContentAsStream(Class<T> type)` and  `<T> Mono<T> getContentAs(Class<T> type)` methods available on `Part` ? Or should we delegate that part to a composite `MultipartHttpMessageReader` that could be initialized with a `List<HttpMessageReader>` like we do on the writing side for SSE?

About the NIO Multipart library, it seems to me that the request parsing is done in a non blocking way but sadly we can only access to data as `String` or `InputStream` depending on the part type, so it may be blocking I suppose. I have implemented an optimisation to be able to perform a zero-copy transfer if the `InputStream` is a `FileInputStream`.

I can take care of the changes discussed here if you are ok.

I am going to create an issue on their GitHub to ask them more insight about the blocking behavior implied by `InputStream` and that fact that non-file part content is only accessible as a String.

rstoyanchev:
Okay, I'm not too sure what you mean with the HttpMessageReader based capabilities. I would think that Part should only expose basic ways of getting to the data, but once you've updated the PR maybe it will be more clear.

sdeleuze:
I have pushed an updated PR, and asked [a new question](https://github.com/synchronoss/nio-multipart/issues/4#issuecomment-254652854) to NIO Multipart project lead about how to perform non-blocking processing with their current `InputStream` / `OutputStream` based API.

sdeleuze:
@rstoyanchev Could you please review the last version of this PR?

spring-web/src/main/java/org/springframework/web/multipart/reactive/nio/NioMultipartResolver.java
```diff
@@ -0,0 +1,167 @@
...
+					throw Exceptions.bubble(ex);
+				}
+
+			}, (e) -> {
+				try {
+					parser.close();
+				}
+				catch (IOException ex) {
+					throw Exceptions.bubble(e);
+				}
```
rstoyanchev: Should this notify the parser about the error which calls `error` on the emitter, so that subscribers for the parts will get the error?

spring-web/src/main/java/org/springframework/web/multipart/reactive/nio/NioMultipartResolver.java
```diff
@@ -0,0 +1,167 @@
...
+
+				@Override
+				public void onNestedPartFinished() {
+				}
+
+				@Override
+				public void onError(String message, Throwable cause) {
+					emitter.error(new MultipartException(message, cause));
+				}
+			});
```
rstoyanchev: An inner class perhaps given the number of methods?

spring-web/src/main/java/org/springframework/web/server/ServerWebExchange.java
```diff
@@ -118,4 +122,19 @@
...
+	 */
+	Flux<Part> getParts();
+
+	/**
+	 * Lazily parse all parts of a multipart request.
+	 * @return a {@link Mono} with a map containing all the parts parsed, with the name as key,
+	 * or a {@link Mono} emitting an error the current request is not a multipart one.
+	 */
+	Mono<MultiValueMap<String, Part>> getAllParts();
+
```
rstoyanchev: These could be ordered right below `getPrincipal()` above with `getAllParts` ahead of `getParts` as the more common case. Same for decorator and default implementations.

spring-web/src/test/java/org/springframework/http/server/reactive/MultipartIntegrationTests.java
```diff
@@ -0,0 +1,164 @@
...
+						else if (part.getName().equals("barPart")) {
+							assertEquals("barPart", part.getName());
+							Optional<String> filename = part.getFilename();
+							assertFalse(filename.isPresent());
+							assertEquals("bar", part.getValue().block());
+						}
+						else {
+							fail();
+						}
+					})
```
rstoyanchev: Could `exchange.getParts()` with `collectToMap` be used to get all parts first and then invoke a method that asserts all parts, which can also be called from `assertGetAllParts`?

rstoyanchev (changes requested):
Having `ServerWebExchange` to expose multipart methods and delegate to `MultipartResolver` internally looks alright. In fact the MultipartResolver seems to belong in the http package. It's like an HttpMessageReader and could even be used with an HttpMessageReader to further parse the content of a part. I suggest moving `web.multipart.reactive` to `http.reactive` right next to the codec package.

spring-web/src/main/java/org/springframework/web/multipart/reactive/nio/NioMultipartResolver.java
```diff
@@ -0,0 +1,167 @@
...
+
+	@Override
+	public Mono<MultiValueMap<String, Part>> resolveAllParts(ServerHttpRequest request) {
+		return resolveParts(request)
+				.collectList()
+				.map(parts -> {
+					MultiValueMap<String, Part> partsMap = new LinkedMultiValueMap<>();
+					parts.forEach(part -> partsMap.add(part.getName(), part));
+					return partsMap;
+				});
```
sdeleuze: Reactor `collectMultimap()` is using a `Collection` for the value type while `LinkedMultiValueMap` is taking a `Map<K, List<V>>` parameter, that's why I did not use it in my first implementation, but after more thoughts I was able to use it in my updated commit thanks to `Collectors.toMap()` to change the value type of the map.

spring-web/src/main/java/org/springframework/web/server/ServerWebExchange.java
sdeleuze: I changed the order and also proposed new names in order to be more consistent with `getFormData()`: `Mono<MultiValueMap<String, Part>> getFormParts()` and `Flux<Part> getFormPartsAsFlux()`. Are you ok with these new names?

sdeleuze:
@rstoyanchev  I took in account your feedbacks and:
 - Changed the package to `org.springframework.http.codec.multipart`
 - Polished the Javadoc
 - Renamed `getAllParts()` to `getFormParts()`
 - Renamed `getParts()` to `getFormPartsAsFlux()`
 - Integrated the `MultipartHttpMessageWriter` commit

rstoyanchev:
This is now in master.

PR #1207, opened 2016-10-14
Add HeadMapping shortcut annotation for @RequestMapping(method = RequestMethod.HEAD)

SympathyForTheDev:

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/bind/annotation/HeadMapping.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/annotation/HeadMapping.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.bind.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.core.annotation.AliasFor;
+
+/**
+ * Annotation for mapping HTTP {@code HEAD} requests onto specific handler
+ * methods.
+ *
+ * <p>Specifically, {@code @HeadMapping} is a <em>composed annotation</em> that
+ * acts as a shortcut for {@code @RequestMapping(method = RequestMethod.HEAD)}.
+ *
+ * <h5>Difference between {@code @HeadMapping} &amp; {@code @RequestMapping}</h5>
+ * <p>{@code @HeadMapping} does not support the {@link RequestMapping#consumes consumes}
+ * attribute of {@code @RequestMapping}.
+ *
+ * @author Damien Bouclier
+ * @since 4.3
+ * @see GetMapping
+ * @see PostMapping
+ * @see PutMapping
+ * @see DeleteMapping
+ * @see PatchMapping
+ * @see RequestMapping
+ */
+@Target(ElementType.METHOD)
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@DevWantJustHaveFun Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1207)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1207) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@DevWantJustHaveFun Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1207)!

rstoyanchev:
@DevWantJustHaveFun as of 4.3 we have built-in support for HTTP HEAD and OPTIONS so that any GET mapped method should also support HEAD with no extra code. I'm wondering what your case is?

DevWantJustHaveFun:
Hello, thanks for your response, in some projects, I have this kind of code 

` 
@RequestMapping(value = "/ping", method = RequestMethod.HEAD)
@ResponseStatus(value = HttpStatus.OK)
public void ping()
{
    // pong ....
}
` 
I use this with a load balancer (Haproxy), the goal of this method is to check the health of my springboot application.

I want just to add a new annotation shortcut like "@GetMapping"

` 
@HeadMapping(value = "/ping")
@ResponseStatus(value = HttpStatus.OK)
public void ping()
{
    // pong ....
}
` 

If you have a better syntax for my case, please tell me. 
Thanks,

rstoyanchev:
This will work:
````java
@GetMapping("/ping")
public void ping()
{
// pong ....
}
````

jhoeller:
I'm rather strongly opposed to adding an explicit HEAD mapping since we generally want HEAD to be handled through `@GetMapping` (considering HEAD simply as a bodyless GET variant), and since a custom HEAD mapping for special-purpose scenarios can be expressed through the general `@RequestMapping` annotation anyway.

DevWantJustHaveFun:
Ok, I have tested your solution it's work (with @ResponseStatus(value = HttpStatus.OK))

but I have two points 

1/ it's not obvious  in the code, if I see that in an API, I will try to call the endpoint with a "GET" method not a "HEAD" 

2/ With my solution if you call the endpoint with a GET method you will get an error 405 "Method Not Allowed"


> curl -i -X HEAD http://localhost:8080/ping

> HTTP/1.1 200 OK
> Set-Cookie: JSESSIONID=KHbbUS4TsdPxBUbD6NUT0vJCuxwGuuJjWI2PjlTh; path=/
> Content-Length: 0
> Date: Mon, 17 Oct 2016 13:16:21 GMT
> Set-Cookie: SERVERID=; Expires=Thu, 01-Jan-1970 00:00:01 GMT; path=/
> 

>  curl -i -X GET http://localhost:8080/ping

> HTTP/1.1 405 Method Not Allowed

jhoeller:
Good points, those are indeed reasons to avoid `@GetMapping.` However, if you'd like such a custom tailored HEAD endpoint, what's wrong with `@RequestMapping(method=HEAD)`? The whole point of `@GetMapping` and co is a convenience for common cases, with `@RequestMapping` remaining as the common underlying mapping mechanism (that `@GetMapping` and co are literally composed from). I would argue that a custom ping endpoint does not qualify as such a common case, and that the common way of handling HEAD is indeed as a combined GET/HEAD endpoint along the lines of `@GetMapping.`

Also, you could easily compose a custom `@HeadMapping` annotation yourself, possibly even as a custom `@PingMapping` variant for your particular use case...

DevWantJustHaveFun:
@jhoeller Thanks for your reponse.

Annotations  shorcut like `@GetMapping`  are cool because more readable and they reduce boilerplate.
They are more like annotations defined in the JSR-339, and the @HEAD annotation exist in the JSR-339.

I have no problem with the usage of @RequestMapping(method=HEAD) but I thought this method was forgotten in the Spring implementation of the `@RequestMapping` shortcut annotations

rstoyanchev:
> With my solution if you call the endpoint with a GET method you will get an error 405 "Method Not Allowed" but not with the @GetMapping, for me it's look better because my "ping" method should be used only with "HEAD"

I think that only points to the fact that HTTP HEAD is perhaps not used for it's intended to do. The definition of HEAD is that it behaves exactly like GET minus the body. So I would ask why not use HTTP GET for the ping to begin with?

DevWantJustHaveFun:
Ok, I will continue to use my custom annotations in my own project, you can decline the pull request, if it's not relevant for the framework.
Anyway thanks for your time.

PR #1210, opened 2016-10-18
DefaultDataBuffer improvements

violetagg:
DefaultDataBuffer.grow(int):
- Copy only the remaining data
- Update readPosition/writePosition


poutsma:
Merged.

PR #1211, opened 2016-10-18
Better encapsulation for CORS default permit configuration

sdeleuze:
This PR allows better encapsulation for CORS user-level configuration, and improve the `CorsRegistration` Javadoc as discussed in [SPR-14798](https://jira.spring.io/browse/SPR-14798).

I did not implement the `CorsConfiguration #PERMIT_ALL_CORS_CONFIGURATION` proposal because the properties won't be immutable, so I chose the previous proposal using `applyDefaultPermitConfiguration()` method name instead of `initEmptyToPermitAll()`.

I chose this name because this is not totally a permit all strategy, not all methods are allowed and it also set `allowCredentials` and `maxAge` to defaults that make sense for most users.

### DIFF
--- a/spring-web-reactive/src/main/java/org/springframework/web/reactive/config/CorsRegistration.java
+++ b/spring-web-reactive/src/main/java/org/springframework/web/reactive/config/CorsRegistration.java
@@ -19,18 +19,14 @@ package org.springframework.web.reactive.config;
 import java.util.ArrayList;
 import java.util.Arrays;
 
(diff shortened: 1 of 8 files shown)


rstoyanchev:
This is now merged after some minor polish. 

One thing to note in particular to check is CorsConfiguration used to say (on its constructor and type-level) that it does not allow any requests except for GET and HEAD. However as far as I can see no cross-origin requests will be allowed at all as long as origins is set to null. So I have updated that to say so.

PR #1212, opened 2016-10-19
add some utility methods for StringUtils

SympathyForTheDev:
this PR add some useful method in StringUtils like "isNotEmpty", 
this kind of method is more readable and avoid thing like this 

> import static org.springframework.util.StringUtils.isEmpty;
> (...)
> if (!isEmpty(someString)) {
> // some code with this string
> }

OR 

> import static org.springframework.util.StringUtils.isEmpty;
> import static org.springframework.util.StringUtils.trimWhiteSpace;
> (...)
> if (!isEmpty(trimWhiteSpace(someString))) {
> // some code with this string
> }

Also I have update the StringUtils unit test

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -95,8 +95,7 @@ public abstract class StringUtils {
      * @return true if string is not empty
      * @since 5.0.0
      */
-    public static boolean isNotEmpty(CharSequence str)
-    {
+    public static boolean isNotEmpty(Object str) {
         return !isEmpty(str);
     }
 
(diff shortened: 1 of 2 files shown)


DevWantJustHaveFun:
@snicoll  it's necessary to create a jira ticket for this PR ?

snicoll:
We prefer to have things to other way around: discussion for non trivial changes in Jira and then a PR. This avoid you spending time on something we wouldn't ultimately include. 

Can you please create a Jira issue?

snicoll:
On a second look on the framework and having double check with @jhoeller we won't accept that change unfortunately. We have several shorter options for those and we don't think it's worth adding them in the core framework. Thanks for the PR though!

DevWantJustHaveFun:
@snicoll  thanks for your response, just out of curiosity what is the shorter options ?

snicoll:
`isNotEmpty is longer than `!isEmpty`. As for ` isBlank`, we got `hasText` vs `!hasText` already.

PR #1213, opened 2016-10-19
ServletHttpHandlerAdapter improvements

violetagg:
- Handle the request from a different thread instead of the current
  container thread
- Handle AsyncListener.onComplete event

Issue: SPR-14772


rstoyanchev:
This has been merged. I've expanded on the commit message a bit to reflect the background.

PR #1218, opened 2016-10-24
SPR-14818 - Add exception MissingServletRequestHeaderException

AustinSong:
Reason : I am implementing a Restful API and needed to add a custom header value to all responses.  Some endpoint contained required request headers and wanted to customize the output for the specific situation of missing header values, by handling a more specific exception than ServletRequestBindingException.

General usage benefit: This is more consistent with the implementation of required request parameters.  There exists a MissingServletRequestParameterException for the specific situation of missing required request parameters.  A more specific exception is beneficial.

Issue: SPR-14818

I have agreed to the ICLA.


mathieuales:
Great !
Which is the minimal release for this new feature ?

snicoll:
@mathieuales the PR is still open (and so is [the Jira issue](https://jira.spring.io/browse/SPR-14818). In other words, it's not available yet.

radarsh:
Why is this issue still unresolved?

AustinSong:
https://jira.spring.io/browse/SPR-14818

The JIRA ticket seems to be tied to a different pull request, is that intentional?  Are we looking to go with a more generic message?

FancyCoder0:
Dear Austin,

We are working on identifying redundant development and duplicated pull requests. We have found there is a pull request: https://github.com/spring-projects/spring-framework/pull/1653 which might be a potentially duplicate to this one. We would really appreciate if you could help us to validate and give us feedback.

Sincerely,
Luyao

AustinSong:
Hey Luyao,

These pull requests do appear to be duplicates, with some caveats.  The other [#1653](https://github.com/spring-projects/spring-framework/pull/1653) appears to be more of a generic exception covering several different missing header values.  While an improvement, it still leaves ambiguity.  This pull request specifically,  is only dealing with the case of missing headers.  I believe it would be beneficial to include specific exceptions being more explicit for common cases.  The fact that it's done for request parameters leads me to also believe this is the favorable approach.

Hope this helps!

Best,
Austin

izeye:
[The referenced Jira issue](https://jira.spring.io/browse/SPR-14818) has been resolved, so this PR looks closable.

snicoll:
Indeed, good catch again @izeye

PR #1221, opened 2016-10-24
SPR-14773: CssLinkResourceTransformer not working with gzipped CSS

mnhock:
CssLinkResourceTransformer is not working if you are using gzipped CSS
files. The links will not be modified within a CSS file, if you have a
gzipped file. The problem is that the links could not be find due to
reading the gzipped content. The gzipped file should be decompressed
first or if available the uncompressed version should be used.

Issue: SPR-14773

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/resource/GzipResourceResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/resource/GzipResourceResolver.java
@@ -22,6 +22,7 @@ import java.io.InputStream;
 import java.net.URI;
 import java.net.URL;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


bclozel:
The main goal behind the `GzipResourceResolver` is to resolve resources that were gzipped at build time, in order to save CPU and avoid the container to compress those resources while being served to clients.

I'm closing this PR as this change would lead to:
* resolving a pre-gzipped resource
* uncompressing it at runtime
* transforming it at runtime
* in this case, not even re-compressing it, which leads to an incorrect response since the gzip HTTP response header is still there and the body is not compressed anymore

I've addressed this issue in [SPR-14773](https://jira.spring.io/browse/SPR-14773).

PR #1224, opened 2016-10-26
Add HttpHeaders#get/setContentDisposition()

sdeleuze:
As requested in [SPR-14408](https://jira.spring.io/browse/SPR-14408), this PR adds a method to get the `filename` parameter of the `Content-Disposition` header.

@bclozel I have not yet included support for `filename*` encoded headers, because I am not sure we should support this. See [this comment](https://jira.spring.io/browse/SPR-14547?focusedCommentId=133006&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-133006) for more details. I will not merge this PR before your feedback.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -20,7 +20,6 @@ import java.io.Serializable;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -676,6 +675,8 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 	 * for {@code form-data}.
 	 * @param name the control name
 	 * @param filename the filename (may be {@code null})
+	 * @see #setContentDisposition(ContentDisposition)
+	 * @see #getContentDisposition()
 	 */
 	public void setContentDispositionFormData(String name, String filename) {
 		setContentDispositionFormData(name, filename, null);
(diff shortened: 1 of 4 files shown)


sdeleuze:
After more reading, `filename*` is likely to be the right way to go for modern browsers, so I am going to update this PR with support for decoding it.

sdeleuze:
@rstoyanchev @poutsma I think this PR is ready to be reviewed by one of you before going to master.

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -682,6 +683,41 @@ public void setContentDispositionFormData(String name, String filename) {
...
+						}
+						else if (attribute.equals("filename*") ) {
+							return decodeHeaderFieldParam(value);
+						}
+					}
+				}
+			}
+		}
+		return null;
+	}
```
rstoyanchev: The Stream API could make this more readable, something like (not tested):
````java
String contentDisposition = getFirst(CONTENT_DISPOSITION);
if (contentDisposition == null) {
	return null;
}
return Arrays.stream(StringUtils.tokenizeToStringArray(contentDisposition, ";"))
		.skip(1)
		.map(part -> StringUtils.tokenizeToStringArray(part, "="))
		.filter(pair -> {
			List<String> attributeNames = Arrays.asList("filename", "filename*");
			return (pair.length == 2 && attributeNames.contains(pair[0]));
		})
		.map(pair -> {
			String value = "";
			return new String[] { pair[0], value};
		})
		.map(pair -> "filename".equals(pair[0]) ? pair[1] : decodeHeaderField(pair[1]))
		.findFirst();
````

File: spring-web/src/main/java/org/springframework/http/HttpHeaders.java [hunk 2]
```diff
@@ -1387,6 +1425,45 @@ static String encodeHeaderFieldParam(String input, Charset charset) {
...
 	}
 
+	/**
+	 * Decode the given header field param as describe in RFC 5987.
+	 * <p>Only the US-ASCII, UTF-8 and ISO-8859-1 charsets are supported.
+	 * @param input the header field param
+	 * @return the encoded header field param
+	 * @see <a href="https://tools.ietf.org/html/rfc5987">RFC 5987</a>
+	 */
+	static String decodeHeaderFieldParam(String input) {
```
rstoyanchev: This should be a private field, if testing is the only reason, which is the case as far as I can see.

rstoyanchev (changes requested):

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
rstoyanchev: Also I'm wondering if it's worth having a ContentDisposition representation along the lines of `HttpHeaders#getContentDisposition()`, such as [for example](https://jersey.java.net/apidocs/2.8/jersey/org/glassfish/jersey/media/multipart/ContentDisposition.html).

sdeleuze: That would extend the scope of this feature, but yes I think that would be a good idea, more future proof than getContentDispositionFilename(). Do you think we should keep the builder pattern?

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
sdeleuze: It seems Java 8 stream API is not designed to collect successive pair, so I have kept the array based implementation for now.

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
sdeleuze: I have made both `encodeHeaderFieldParam()` and `decodeHeaderFieldParam()` `private static` methods in my last commit.

sdeleuze:
@rstoyanchev I have pushed an updated commit with `HttpHeaders#getContentDisposition()`. I have implemented the support for the disposition type and `name`, `filename`, `filename*`and `size` parameter.

If we don't have date parsing according to [RFC 822](https://tools.ietf.org/html/rfc822#section-5.1) already implemented, I would suggest to wait an eventual user request to implement `creation-date`, `modification-date` and `read-date` support (my guess is that they are not very commonly used, and in any case this is outside of the scope of the original issue).

I have keep an array based parsing instead of the `Stream` based one you proposed because [Java 8 streams seems not really design to collect successive pairs](http://stackoverflow.com/questions/20470010/collect-successive-pairs-from-a-stream) + I think current implementation will be faster.

I have made the 2 encode and decode methods private as requested.

Could you have another look to this updated PR and send me your feedback if any?

rstoyanchev:
Sorry for the delay. A few last tweaks. Let's make ContentDisposition a top-level type like HttpMethod, MediaType, HttpRange. The parse logic can then be a static method in Content-Disposition following the same pattern as HttpRange. Similarly the setContentDisposition should take a ContentDisposition and the actual logic to format to a String can be in ContentDisposition, again like HttpRange. In other words let's encapsulate all the Content-Disposition related logic inside `ContentDisposition` and HttpHeaders will simply delegate to it.

sdeleuze:
No worry, thanks for your feedback, I will update the PR accordingly.

rstoyanchev:
Okay, feel free to commit straight to master after those changes.

sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/99a8510ace46af9b05b822e7c65f08aae885ca98

PR #1231, opened 2016-11-06
PropertyEditor for java.time.Duration

findepi:
java.time.Duration is useful type for bean properties that represent timeouts
or other time limits. It's easily instantiatable using Java configuration and
XML configuration as well (e.g. with a SpEL). However, a property editor is
necessary to provide Duration value in a Property-based configuration override
file.

For example, this allows a bean with a definition like:

```xml
<bean id="a" class="A">
    <property name="duration" value="#{T(java.time.Duration).ofSeconds(2)}"/>
</bean>
```

to have its property reconfigured with:

```xml
<context:property-override properties-ref="p"/>

<util:properties id="p">
    <prop key="a.dur">12 s</prop>
</util:properties>
```
(where the actual properties might actually be sourced from an external configuration file.)


findepi:
I'm aware of https://jira.spring.io/browse/SPR-13721 but that did not provide out-of-the-bxo String-to-Duration conversion for overrides. I.e. without the proposed changes, the example set forth in the PR description won't work. As the property overrides are often used for configuration, this PR adds, as an additional bonus, support for simple representation that is both really human readable and natural.

findepi:
I still believe this patch brings something good to Spring. However, I guess it's not going to be merged (am I right, @jhoeller?). Therefore closing it.

PR #1235, opened 2016-11-09
Support URI with https scheme and port for SSL client use with Netty

mp911de:
Do not check on non-specified port when scheme is https. Enables SSL for https URIs with a specified port.

Issue: SPR-14889


bclozel:
Merged with 6261106

Thanks!

PR #1236, opened 2016-11-11
SPR-14901 Allow customization of STOMP message header encoding

dreis2211:
Fixes https://jira.spring.io/browse/SPR-14901


rstoyanchev:
Could we go without the STOMP codec registration? I had the impression you were okay with the latest results?

dreis2211:
Finally had some time to work on this again. Please also see my latest comment in the ticket.

rstoyanchev:
This is now merged and backported to 4.3.x. 

Note that I've added synchronization around the cache updates and an additional access-only cache map since the LinkedHashMap cannot be accessed and updated concurrently.

PR #1237, opened 2016-11-14
Add SockJS clients for EventSource transport

slovdahl:
There were no transport implementations available in spring-websocket
that were able to use the EventSource transport method. This commit
adds EventSource transports based on RestTemplate, Jetty and Undertow.

This commit also extracts some common code that can be shared between
XHR and EventSource transports to the `AbstractHttpTransport` class.

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/sockjs/client/AbstractEventSourceTransport.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/sockjs/client/AbstractEventSourceTransport.java
@@ -16,8 +16,8 @@
 
 package org.springframework.web.socket.sockjs.client;
 
+import java.io.ByteArrayOutputStream;
 import java.net.URI;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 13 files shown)


slovdahl:
Please note that the actual parsing of multiline events is still missing in all three client implementations. I gave it a try but still haven't succeeded in writing a spec-compliant multiline parser. Any help with writing the parsers would be much appreciated!

rstoyanchev:
What is the motivation is for the extra transport? The choice of the current transports is by design and not because other could not be added.

The JavaScript https://github.com/sockjs/sockjs-client has all these transports in order to cover a wider range of browsers. The EventSource one is used only in very specific cases as shown in [this table](https://github.com/sockjs/sockjs-client#supported-transports-by-browser-html-served-from-http-or-https).

Outside the browser however, the browser compatibility is not a concern and it's mainly about having a HTTP streaming as a WebSocket alternative (e.g. if not supported) and long polling if streaming is prevented by proxies for example.

slovdahl:
I do understand that the EventSource transport isn't as generally useful as the other transports. It's mostly useful in different kinds of test situations. I wrote them as part of a master thesis where I compare different HTTP push mechanisms.

rstoyanchev:
Generally not opposed to having such a transport, but in the proposed implementation adds close to 2K lines of code which is hard to justify or reconcile intuitively for the given change since on the server side, in the transport handler hierarchy, the concrete implementations are minimal. From what I can see the RestTemplate, Jetty, and Undertow  `EventSourceTransport` implementations duplicate their XhrTransport equivalents which is a non-starter.

PR #1238, opened 2016-11-16
Router without generics

sdeleuze:
To be discussed ...


sdeleuze:
As discussed during the meeting, there is maybe something to rethink about filters. It is not obvious how you chain multiple filters currently for example.

sdeleuze:
Closing this PR for now, it is just here for discussion and better understanding of the use cases for `RouterFunction#and()`, `RouterFunction#andSame()` and `RouterFunction#filter()`.

PR #1240, opened 2016-11-21
Add RedirectView to Spring Web Reactive

sdeleuze:
Main differences with the Spring MVC original implementation:
 - Default redirect HTTP code is 303 See Other since we can assume
   all HTTP clients support HTTP 1.1 in 2016
 - No more http10Compatible property, use statusCode instead
 - By default the redirect is relative to the context path

@rstoyanchev Could you please review this PR to give me your feedback on the choices I made? Notice that `sessionId` should be correctly encoded for Servlet based engines.

### DIFF
--- a/spring-web-reactive/src/main/java/org/springframework/web/reactive/result/view/RedirectView.java
+++ b/spring-web-reactive/src/main/java/org/springframework/web/reactive/result/view/RedirectView.java
@@ -1,0 +1,365 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.reactive.result.view;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.nio.charset.Charset;
+import java.util.Collections;
+import java.util.Locale;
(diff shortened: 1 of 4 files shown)


rstoyanchev:
@sdeleuze overall looks good. 

A few comments with hindsight in mind on an all new RedirectView. Starting with the overloaded constructors designed to be one liners in controller methods, perhaps we could replace those with a builder? Not sure if we even need the default constructor "for use as a bean" since what's typically configured in Spring configuration is the UrlBasedViewResolver.

On that note, the present UrlBasedViewResolver exposes all (or most?) properties that RedirectView as so that it can pass them each time it creates a RedirectView instance at runtime. At present I see only redirectStatusCode and redirectHosts. I'm also not a fan of replicating as many properties. Perhaps we can think of some better mechanism. Perhaps a property that accepts a `java.util.function.Function<String, RedirectView>`?

For exposeModelAttributes let's skip that part altogether including appendQueryProperties and related methods. The automatic appending of attributes to the query has been a long standing source of issues and one of the drivers for RedirectAttributes (aside from flash attributes). Instead we can make expandUriTemplateVariables the primary way of appending to the query on a redirect. It's simple, expressive, and explicit. 

For the ServletBasedRedirectUrlEncoder this is raising a good point for SPR-14529 that the solution for URL-rewriting should probably be at the ServerHttpResponse level and allow any such underlying runtime mechanism to also be included. This is what we did for contextPath support on ServerHttpRequest. So feel free to leave this part out for now and @Ignore any tests related to it.

spring-web-reactive/src/main/java/org/springframework/web/reactive/result/view/RedirectView.java
```diff
@@ -0,0 +1,521 @@
...
+
+	/**
+	 * Constructor for use as a bean.
+	 */
+	public RedirectView() {
+	}
+
+	/**
+	 * Create a new RedirectView with the given URL.
+	 * <p>The given URL will be considered as relative to the the current context path.
```
sdeleuze: This is a typo, I will fix that thanks for your feedback.

sdeleuze:
Merged as https://github.com/spring-projects/spring-framework/commit/2e994f7dc38b5520555720dac215064949ccfabf.

PR #1242, opened 2016-11-21
Treat Kotlin nullable as non-required

sdeleuze:
Superseeds #1060

@jhoeller Could you please review the polished version of this pull request, since this is the first time we introduce Kotlin code in our unit tests and since it adds a new `Optional<Boolean> MethodParameter#isNullable()` method that I would like to validate with you?

Issue: https://jira.spring.io/browse/SPR-14165


sdeleuze:
I merged a refactored version that reuses `MethodParameter#isOptional()`, with more tests and polish.

PR #1244, opened 2016-11-24
Fix wrong synchronization

quaff:
1. synchronization lock should on class not instance if try to access static variable
2. use double-checked locking instead


snicoll:
@quaff please create a Jira issue for this.

jhoeller:
This is just a class in our tests. The production variant `StandardTypeConverter` had synchronization refined already (https://jira.spring.io/browse/SPR-14465), and we're considering general reuse of a shared `DefaultConversionService` instance now (https://jira.spring.io/browse/SPR-14948).

PR #1245, opened 2016-11-24
Make the signature of ReactiveHttpOutputMessage#writeAndFlush(…) more flexible

danielfernandez:
This solves [SPR-14952](https://jira.spring.io/browse/SPR-14952) by applying the suggested solution and cascading the change through the interface hierarchy (and clients).

Original signature of `ReactiveHttpOutputMessage#writeAndFlushWith(…)`:

```java
Mono<Void> writeAndFlushWith(Publisher<Publisher<DataBuffer>> body);
```

Modified signature:

```java
Mono<Void> writeAndFlushWith(Publisher<? extends Publisher<DataBuffer>> body);
```

Which makes this method callable with `Flux<Flux<DataBuffer>>` objects as arguments. Which is what is obtained for example by means of calling `Flux#window(…)`:

```java
final Flux<DataBuffer> dataStream = ...;
// Flux<T>#window(...) returns Flux<Flux<T>>
exchange.getResponse().writeAndFlushWith(dataStream.window(3));
```

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpResponse.java
+++ b/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpResponse.java
@@ -52,7 +52,7 @@ public class MockServerHttpResponse implements ServerHttpResponse {
 
 	private Publisher<DataBuffer> body;
 
-	private Publisher<Publisher<DataBuffer>> bodyWithFlushes;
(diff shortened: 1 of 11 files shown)


sdeleuze:
Merged and polished with https://github.com/spring-projects/spring-framework/commit/8d26c738a016d9aa57fc3b080fe58f0c44fb341d in order to also relax type checks on `DataBuffer`. Thanks for your contribution.

PR #1246, opened 2016-11-25
ThreadPoolExecutorFactoryBean add ability to prestart threads

fededonna:
Sometimes you want the executor given by ThreadPoolExecutorFactoryBean to be already hot. Thread initialization can be accomplished with MethodInvokingBean, but that is a little bit tricky.


snicoll:
@fededonna sorry it took so long and thank you for making your first contribution to Spring Framework.

PR #1250, opened 2016-11-30
update STOMP support to reactor-netty

smaldini:


rstoyanchev:
This is now merged. Thanks!

Note that I refactored the ReactorNettyTcpClient input a bit introducing a simple ReactorNettyCodec to hold the encoding and decoding delegates. Effectively that replaces the MessageHandlerConfiguration with the Scheduler now always created internally.

PR #1252, opened 2016-12-01
Add twice to ExpectedCount as convenience method

drumonii:


spring-test/src/main/java/org/springframework/test/web/client/ExpectedCount.java
```diff
@@ -78,6 +79,13 @@ public static ExpectedCount once() {
 	}
 
 	/**
+	 * Exactly once.
```
kartoffelsup: The comment sounds a bit wrong to me. :o

drumonii: Doh! Copy and paste error... I'll update that

rstoyanchev:
This is now merged, thanks!

PR #1255, opened 2016-12-06
SPR-12031 Support @ContextConfiguration at method level

fdesu:
Prior Spring TestContext Framework had not supported
@ContextConfiguration annotation on the method-level.

This commit enables @ContextConfiguration and
@ContextHierarchy to be the method-level annotations.
Now TCF will emit fully configured throw-away
TestContext for the particular test method.

Issue: [SPR-12031](https://jira.spring.io/browse/SPR-12031)

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/util/MetaAnnotationUtils.java
+++ b/spring-test/src/main/java/org/springframework/test/util/MetaAnnotationUtils.java
@@ -17,6 +17,7 @@
 package org.springframework.test.util;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -26,6 +27,7 @@ import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.core.style.ToStringCreator;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
+import org.springframework.util.ReflectionUtils;
 
 /**
  * {@code MetaAnnotationUtils} is a collection of utility methods that complements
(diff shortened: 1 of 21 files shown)


spring-test/src/main/java/org/springframework/test/util/MetaAnnotationUtils.java
```diff
@@ -105,7 +110,7 @@
 		}
 
 		// Declared locally?
-		if (AnnotationUtils.isAnnotationDeclaredLocally(annotationType, clazz)) {
+		if (isAnnotationDeclaredLocally(annotationType, clazz)) {
```
sbrannen: Production code is not allowed to use static imports.

sbrannen (changes requested):
Please apply the requested changes in order to reduce the amount of noise in this commit.

Thanks!

fdesu:
@sbrannen 
Can you please advice if [such declaration](https://github.com/spring-projects/spring-framework/pull/1255/files#diff-d8ea9b54580f6b93b075124892369a62R197) permissible. Or should I pull this classes in the _different files_/_into some class as inner classes_. 
Asking because it is forbidden by the Code Style.

sbrannen:
> Can you please advice if such declarations permitted. Or should I pull this classes in the different files/into some class as inner classes.
> Asking because it is forbidden by the Code Style.

It's definitely forbidden in production code. In test code, I think there are few places where that style is used. But I don't want that in the `spring-test` module. So please convert them to static nested classes within the test class.

Thanks

fdesu:
@sbrannen 
I amended previous commit and hope that current one is easier to review

PR #1257, opened 2016-12-06
SPR-14988: @GetMapping method annotation uses consumes attribute from @RequestMapping class annotation

bedrin:
[SPR-14988: @GetMapping method annotation uses consumes attribute from @RequestMapping class annotation](https://jira.spring.io/browse/SPR-14988)


rstoyanchev:
This is now merged, thanks!

PR #1262, opened 2016-12-07
SPR-13508 - Add i18n support to ScriptTemplateView

pgrimard:
The current ScriptTemplateView does not provide support for internationalization using a message resource bundle on the classpath. This PR allows configuring a resource bundle basename on the ScriptTemplateConfigurer and then load that resource bundle at render time using the HttpServletRequest locale.

Two new methods have been added to ResourceBundleMessageSource allowing to return a ResourceBundle as a Map for a given basename and locale, or simply for a given locale. The resulting map can then be passed to the render function from ScriptTemplateView where it can be used to replace message placeholders in the template.

Issue: SPR-13508

I have signed the CLA.


sdeleuze:
@pgrimard Thanks for this PR, I am currently evaluating various option to fulfill that need as part for [SPR-15064](https://jira.spring.io/browse/SPR-15064). I will keep you informed.

pgrimard:
That's great!

PR #1263, opened 2016-12-07
SPR-14995 - Add some new matchers for query parameters into MockRestRequestMatchers

schjan79:
Hello,
as I have seen there isn't any request matcher for query parameters in the project spring-test.
I'd like to use it in the project spring-cloud-contract and imho the place of these code pieces are here.
If you have the same feelings please merge this pull request. :)
The factory methods for query parameters also support charset for URI decoding or use the default one (UTF_8)

Thanks.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
@@ -18,6 +18,7 @@ package org.springframework.test.web.client.match;
 
 import java.io.IOException;
 import java.net.URI;
+import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
 import javax.xml.xpath.XPathExpressionException;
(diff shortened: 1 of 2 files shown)


sbrannen:
Thanks for the pull request.

For new features like this, however, we require a JIRA issue.

So, please [create a new JIRA issue](https://jira.spring.io/secure/CreateIssue!default.jspa) against the _Spring Framework_ project and _Test_ component and describe your proposal there. Then update this PR to reference the JIRA issue.

schjan79:
#SPR-14995

rstoyanchev:
This is now merged in master and is also backported to 4.3.x. 

Note that I've removed the overloaded methods taking a charset. A URI is expected to be UTF-8 encoded to begin with. More importantly the query string cannot be parsed reliably after decoding for example if the parameter name or value contains a reserved char like "=" or "&" it would change the result of the parsing. It's just fine to match the encoded values consistent with the remaining methods in the same class. No other request component has 4 overloaded methods and certainly query parameters aren't the most important one.

PR #1268, opened 2016-12-14
Refactor the usage of Undertow ByteBufferPool

violetagg:
- lazy allocate the PooledByteBuffer, only if there is a request body
  for reading
- close the PooledByteBuffer once the request finishes


rstoyanchev:
This is merged. Note that I found a completion listener on HttpServerExchange and replaced the close method with that.

PR #1273, opened 2016-12-22
Ordering by root context first before exporting a json view of contexts

bijukunjummen:
This is likely a fairly pointless PR - absolutely no issues if it is felt that this is likely not required.

I noticed that the `/beans` actuator endpoint in a spring-boot app returns the list of beans in a random order, this can be traced back to the behavior of the LiveBeansView class. This PR orders it parent context first before printing it as a json


snicoll:
Thank you for the PR!

We already had several requests to order things in actuator endpoints and those requests were rejected for the most part. Maybe this one  would not. Could you please create an issue in Spring Boot? We'll see where the change needs to be made at that point.

PR #1276, opened 2016-12-26
SPR-15054: Add Kotlin extensions for function Web API

sdeleuze:
PR ready to be merged after @jhoeller approval.


lqjack:
why introduce kotlin to the spring framework ? just for simpility?

sdeleuze:
@lqjack Kotlin allows to write more efficient code, and is also relevant for both object-oriented and functional programming. Notice that Kotlin support is like any other technology supported in Spring Framework, it is available optionally if you are using it, but if you don't use Kotlin you won't notice any change.

sdeleuze:
Merged

geoand:
I love the way Kotlin is being introduced in a non-intrusive manner into Spring!
Two awesome technologies working together seamlessly just makes our lives simpler :)

PR #1278, opened 2016-12-29
Build - Update Apache Licence 2.0 name and URL

stonio:
Use same Apache Licence 2.0 formal name and URL across all Spring projects.

References:
* https://www.apache.org/licenses/LICENSE-2.0#apply
* https://github.com/spring-projects/spring-data-build/blob/master/pom.xml#L42


snicoll:
Merged, thanks!

PR #1280, opened 2016-12-31
SPR-15075 - Remove String#toCharArray from ScriptUtils

marschall:
ScriptUtils contains two calls to String#toCharArray for the sole
purpose to iterating over all chars in a String. Not only is this
unnecessary and can be replaced with String#charAt it also causes
additional allocator and heap pressure because String#toCharArray
rather than returning the backing array (which is gone in Java 9)
creates a copy.

 - remove String#toCharArray from ScriptUtils and replace with
   String#charAt

Issue: SPR-15075


marschall:
I did sign and agree to the CLA.

snicoll:
@marschall thanks for the PR, no need for those confirmation message about the CLA, we have a bot that checks that automatically based on your account.

snicoll:
Fixed in https://github.com/spring-projects/spring-framework/commit/35bd00beaf25943521b5650ef6e004291b19eff7 - Thanks for the PR!

sbrannen:
👍

PR #1281, opened 2016-12-31
Remove use of Boolean constructors

marschall:
Codacy warns us that there are several references to Boolean
constructors in the tests. Direct usage of the Boolean constructor is
discouraged and even deprecated in Java 9 [1]. Boolean constructor use
can easily be replaced with the constant instances.

 - replace references to Boolean constructors with boolean constants in
   JSP tag tests
 - update the copyright year where necessary
 - BooleanComparatorTests is intentionally left unchanged as it should
   also work with the non-constant instances correctly

 [1] http://download.java.net/java/jdk9/docs/api/java/lang/Boolean.html#Boolean-boolean-

Issue: SPR-15076


snicoll:
Fixed in https://github.com/spring-projects/spring-framework/commit/19e77cd140d6efd2736a43b8688a8a3ffb6f2a7b

PR #1282, opened 2016-12-31
Avoid use of double constructor of BigDecimal

marschall:
Codacy warns about an Error Prone [1] use of the double constructor of
BigDecimal in tests. The reason given is that it is a source of
precision loss if the number does not have an exact double
representation. The recommendation is to use the String constructor of
BigDecimal instead as it does not require using a lossy argument.

 - replace usage of the double constructor of BigDecimal with the
   String constructor of BigDecimal in JdbcTemplateQueryTests
 - update the copyright year

 [1] http://errorprone.info/bugpattern/BigDecimalLiteralDouble

Issue: SPR-15077


snicoll:
Fixed in https://github.com/spring-projects/spring-framework/commit/3f97ab183e08ee0001ee98b9614770dad8d29155

PR #1286, opened 2017-01-06
SPR-15107 Updating MySQLMaxValueIncrementer to not rely on MYISAM

trisberg:
We should no longer rely on MYISAM for the sequence table since this
engine might not always be available. After this change the storage
engine used by the sequence table can be MYISAM or INNODB since the
sequences are allocated using a separate connection without being
affected by any other transactions that might be in progress.

Issue: SPR-15107


trisberg:
Rebased and squashed the commits. Added a `useNewConnection` flag to allow for falling back on original functionality by setting this flag to false.

trisberg:
We might consider having the default for "useNewConnection" to be false in 4.3.x to maintain backwards compatibility and then flip that to true in 5.0.

PR #1292, opened 2017-01-11
SPR-15188 - If backing Cache is down @CacheResult does not seamlessly call method and return.

dbrimley:
I wish to have the JCache annotation `@CacheResult` work seamlessly when the backing cache is down or unavailable at present this is not possible.  

I've been able to intercept cache down exceptions during the GET cache call that is executed in the before method phase, this is because the `CacheResultInterceptor` at line 53 correctly defers to the `AbstractCacheInvoker` for the GET which in turns handles exceptions with the ErrorHandler.

However in the post method call phase where we attempt to place a value into the cache, it fails and throws an exception further up to the caller, because `CacheResultInterceptor` at line 62 deals directly with the cache rather than going through `AbstractCacheInvoker`

Therefore this change...

Allows ErrorHandler to intercept cache.put AFTER the method call in same manner that cache.get allows for ErrorHandler before method call.  Also adding ErrorHandler interceptor to put on ExceptionCache.

This then allows us to handle ANY cache exceptions in the same manner, such as the cache being down in both PRE and POST method call phases.

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/jcache/interceptor/CacheResultInterceptor.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/jcache/interceptor/CacheResultInterceptor.java
@@ -59,7 +59,7 @@ class CacheResultInterceptor extends AbstractKeyCacheInterceptor<CacheResultOper
 
 		try {
 			Object invocationResult = invoker.invoke();
-			cache.put(cacheKey, invocationResult);
+			doPut(cache, cacheKey, invocationResult);
 			return invocationResult;
 		}
 		catch (CacheOperationInvoker.ThrowableWrapper ex) {
@@ -88,7 +88,7 @@ class CacheResultInterceptor extends AbstractKeyCacheInterceptor<CacheResultOper
 			return;
 		}
 		if (filter.match(ex.getClass())) {
-			exceptionCache.put(cacheKey, ex);
+			doPut(exceptionCache, cacheKey, ex);
 		}
 	}
 
--- a/spring-context-support/src/test/java/org/springframework/cache/jcache/interceptor/CacheResultInterceptorTest.java
+++ b/spring-context-support/src/test/java/org/springframework/cache/jcache/interceptor/CacheResultInterceptorTest.java
@@ -1,0 +1,186 @@
+package org.springframework.cache.jcache.interceptor;
+
+import org.junit.Test;
+import org.springframework.cache.Cache;
+import org.springframework.cache.interceptor.*;
+
+import javax.cache.annotation.CacheMethodDetails;
+import javax.cache.annotation.CacheResult;
+import java.lang.reflect.Method;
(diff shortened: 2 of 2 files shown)


dbrimley:
One more thing, when is the M5 5.0.0 Milestone out?

snicoll:
Could you please push tests that exercise that failing use case please?

dbrimley:
I've added a new test for `CacheResultInterceptor` that now tests that a method body value is returned correctly when the cache is in error.  It also checks that the `CacheErrorHandler` is called correctly twice in this scenario, once for the pre method get cache failure and once for the post method put cache failure.

So with your `CacheResultInterceptor` it will fail, with my proposed fix it will pass.

Apologise if the the format or anything else is not quite to standard.

Happy to take further advice and tidy up the test.

Cheers
David

dbrimley:
@snicoll Do you want me to tidy up according to the codacy/pr failure?

snicoll:
Nope that's fine. To be honest, I am not sure I've understood what the problem is so I need to review the code first.

dbrimley:
@snicoll At a high level, it's about providing the same error handling semantics as are provided if you use the @Cachable annotation, the behaviour with @Cacheable is that, on cache failure it always calls the Error Handler for the get (pre-method) and the put (post-method).

@CacheResult presently only calls the error handler on the get (pre-method) but not on the put (post-method), which has the result of short-circuiting the call.

I think it's a small oversight on the handling of errors in the @CacheResult annotation.

My pull request fixes this and has @CacheResult behave the same way as @Cacheable.

Fixing this will also allow @CacheResult to play nicely with circuit breaker patterns.

snicoll:
@dbrimley thanks for the contribution. We already had an existing test class for that stuff so I've just improved that (see my polish commit).

dbrimley:
:thumbsup:

Pleased to help.

linead:
Could this be backported to the 4.3.x?

This forced me to back out our annotation based caching as without being able to use this you end up with a system that fails catastrophically when the cache is down and the provided mechanism of a custom error handler allows for the get and put operations to be non-fatal on error; it will always fall over when the described issue happens with cache result. 

Happy to contribute towards this, I find it strange that cache failures would by default cause fatal errors (to a request).

snicoll:
It has been backported to 4.3.x already.

linead:
Thanks for the backport and sorry for missing that, that's awesome news to me :)

I just wonder if it needs to be fatal, the system is still in a working state and the only indication of a failure goes direct back to the user, also without the cache being part of the healthcheck we can't even have a failed node be removed from the rotation.

snicoll:
I think I already answered to that. The default is binary, either you hide the exception or you don't. If you care (and that's perfectly fine if you do), configure your own `ErrorHandler`.

linead:
no worries :+1: thanks to both of you for getting this fixed :)

PR #1293, opened 2017-01-11
SPR-14544 Introduce PathPatternParser for optimized path matching

bclozel:


rstoyanchev (commented):
@bclozel definitely a step in the right direction judging by the reduced code in existing places where we deal with path matching. Aside from the detailed review comments here is some higher level feedback.

From a configuration perspective things are a bit unclear right now. `AbstractHandlerMapping` still takes a `PathMatcher` but that's not really used (once you factor in the review comments). Logically it should be used in both of the pattern registry things but those want a `PathPatternParser` not just a `PathMatcher`. So it's unclear what would a user actually need to configure?

`PathPatternMappingRegistry` and `PathPatternRegistry` are fashioned a bit too much after the needs of `AbstractUrlHandlerMapping` and `PatternsRequestCondition` respectively. The common thread in both I think is the need to store parsed patterns, to find and sort matches for given a path, as well as to deal with pattern variants (e.g. trailing slash, file extensions, etc). The other features found in those registries like the mappings from pattern to handler, or combining/comparing entire registries I think are concepts that are too specific to `AbstractUrlHandlerMapping` and `PatternsRequestCondition`. 

I think it would be useful to think of a single, more generally applicable component that serves as registry of patterns. If it is configurable with a `PathPatternParser` and other options related to trailing slash and file extension patterns, and is also decoupled from the concept of a handler map, then that's what a `HandlerMapping` could be configurable with replacing existing properties such as `PathMatcher` and the others for trailing slash, and so on. Just a thought.

I also would like to see if we can apply this pattern registry globally across annotated methods. Currently we go one `RequestMappingInfo` at a time but what if we first checked across all patterns and narrowed it down to a list of `RequestMappingInfo`s whose patterns match the request? Then we just need to ask each of those infos to match themselves in full to the request.

aclement:
Rossen said: 
> Currently we go one RequestMappingInfo at a time but what if we first checked across all patterns and narrowed it down to a list of RequestMappingInfos whose patterns match the request?

I'm not a total expert on all the interconnected elements here but the comment reminds me of my first pass at the matching where instead of a set of distinct parsed path objects, there was one tree structure built from parsing all paths (or some coherent group of paths), branches in the tree only existing where paths diverged ( https://github.com/aclement/pathmatcher ). This talks to that use case the guy had on stackoverflow with a lot of overlapping mappings. With such a structure it means if you had 10 paths with the same prefix, we would only match that prefix once against an incoming path to determine if we should continue matching the rest of each pattern against that path or we should dismiss them all immediately.  I kept this direction in mind in the parser/pattern matcher code, so it could grow from the current design into the tree structure, if we went in this direction.

rstoyanchev:
> I'm not a total expert on all the interconnected elements here but the comment reminds me of my first pass at the matching where instead of a set of distinct parsed path objects, there was one tree structure built from parsing all paths

This is very much the way I think of what Brian did with the registry of patterns. One component to keep all patterns and return the sorted matches for a request. So once they're in one place using some tree representation for better performance makes perfect sense.

The `SimpleUrlHandlerMapping` is naturally suited for this approach because it already has a handler map with all patterns in once place. 

The `RequestMappingHandlerMapping` on the other hand maintains a collection of `RequestMappingInfo`s, one for each `@RequestMapping` method, and we are currently matching each info individually because the match is more complex -- it involves multiple patterns per annotation and other conditions. It wouldn't be hard to build a single registry of all patterns for all annotated methods. Then we can perform a lookup across all patterns first and that will significantly narrow the  set of infos that need to be matched individually. Currently we actually [do something very similar](https://github.com/spring-projects/spring-framework/blob/944d17f23773d179b1d788507e8d27ad26777d71/spring-web-reactive/src/main/java/org/springframework/web/reactive/result/method/AbstractHandlerMethodMapping.java#L307-L314) currently with a map of non-pattern URLs so we can perform a "direct lookup" first and only then fall back on matching every info. Of course a lot of mappings tend to use URI variables... So we'd effectively be replacing that with a the pattern-based, tree mechanism.

> it could grow from the current design into the tree structure, if we went in this direction.

I wouldn't worry much about matching to this PR. If you create such a component it's easy to incorporate and redo this PR.

Thanks Andy!

aclement:
> I wouldn't worry much about matching to this PR. If you create such a component it's easy to incorporate and redo this PR.

I stepped back from the tree option to bring down complexity and get something done that supports the new syntax which was more compatible with the existing codebase. Actually, also because I wasn't confident enough to the the real spring changes that it would have necessitated, which Brian has started doing here. Personally I'd probably not rework this PR (my part of it) right now but instead in a later PR (in a later release?) do that refactoring that goes from 'something that iterates over parsed patterns and accumulates matches' to 'something that calls the tree matcher once and gets all matches in one go'. This latter approach also works well for pre-processing the candidate being matched against the patterns for even more speed.  Maybe it depends where we land on any further changes Brian makes, or if you say we must have that tree approach in to make this a coherent piece of work.

rstoyanchev:
Brian's changes are a sketch only at this point but they show precisely what parts need to be touched and it all looks very manageable. We know understand the problem quite well, we just need to make up our minds what the APIs exactly to expose for configuration purposes and what SPIs for consumption within the framework.

> Personally I'd probably not rework this PR (my part of it) right now..

We can work towards having a registry with all patterns with the understanding that further optimizations may be possible later. In other words I agree it makes sense to leave that as a later concern.

bclozel:
@poutsma @rstoyanchev 

I've just updated this PR with the changes we discussed:
* use a `PathPatternRegistry` instance in `AbstractHandlerMapping` and for all its implementations. Those can choose to couple it with a map with `PathPattern` as key and  handlers or mapping info as values.
* the `ParsingPathMatcher` should not be used by default in Spring MVC

With those changes, it seems `PathPatternRegistry` methods are looking more and more like Collection-related methods + parsing/search features. I haven't found a way to extract a meaningful interface out of it for now.

Should we proceed right away with it so that we can use it in the functional WebFlux variant? We can always adapt the contract with that feedback in mind.

rstoyanchev (commented):
@bclozel some minor code-level comments. Overall I agree the way this is integrated looks pretty good now  to evolve from here. That said I do think it's worth discussing a couple more things.

In summary of the code-level comments, I think it would make sense to accept consistently String-based patterns wherever registrations are made and currently it's a mix. Likewise providing matching String-based getters also returning the original set, i.e. without the variants. Such getters are useful for informational purposes and it's most intuitive if they're consistent with the input. If one wants the actual `PathPatern`s it would make more sense to go through the registry at which point getting variant patterns should be less of a surprise.

As a related comment perhaps we can consolidate the management of variant pattern preferences in the registry itself so that there are no extra properties related to things like trailing slashes and suffix pattern matching in places anywhere outside the `PathPatternRegistry` itself and `PathMatchConfigurer`.

I am also wondering if we should configure the same `PathPatternRegistry` instance throughout as we did for `PathMatcher`? That would make sense in terms of pattern variant preferences which ideally should be configured once and used everywhere. 

What about the registry itself though which has specific patterns registered? Should it be shared like a `PathMatcher` was? That could mean the configuration needs to be separated from the actual registry containing patterns. Or with a single registry one could imagine matching patterns once across `HandlerMapping`s and making the result available to them for use as they see fit. This is to the changes for annotated methods in `AbstractHandlreMethodMapping` to narrow down based on pattern first but at a higher level still. That would eliminate calculations about how to order `HandlerMapping`'s based on performance ([one such ticket](https://jira.spring.io/browse/SPR-11541) comes to mind).

bclozel:
Hey @rstoyanchev - Sorry, I did merge this PR before I saw the notification about your additional review.
I did polish a bit those changes with 09d18f2, taking into account many of your comments (thanks!).

I believe a couple of those couldn't be applied in a straightforward way.
I agree, the getter/setter methods aren't symmetrical; but this could actually be a feature of our support. After all, we're registering patterns against a configuration we provided. When asking for the registered patterns, I think we should give the full picture to developers by default. Pattern variants aren't checked on-the-fly anymore, but actively registered at startup. Maybe those setter/getter methods should change name for register/list?

I think you pointed to the big limitation of the current approach. We're gathering configuration flags and the patterns themselves into a single class. So sharing an instance of that one means sharing patterns. But we can't share those between `HandlerMapping` instances, and certainly not with `UrlBasedCorsConfigurationSource`.
This is the part I need to improve in the current design.

PR #1295, opened 2017-01-12
Fix construction of UndertowServerHttpRequest.uri

violetagg:
Ensure there is no "?" at the end of the generated URI when there is
no query available.


rstoyanchev:
Merged, thanks!

PR #1296, opened 2017-01-13
Fix SPR-15136 - Exception if no RequestDataValueProcessor is present

danielfernandez:
This modifications fix the way the reactive AbstractView retrieves the
RequestDataValueProcessor bean, correctly returning null if there
is no bean of such type at the Application Context.

This avoids an exception at the reactive RedirectView (which extends
AbstractView) when trying to post-process the URL generated for 
redirection, when no RequestDataValueProcessor exists.


rstoyanchev:
This is now merged, thanks for the fix!

PR #1297, opened 2017-01-14
SPR-15140: @RequestParam injects a raw undecoded string for reactive HTTP requests

eiiches:
When using spring-web-reactive, %-encoded strings were injected as-is into @RequestParam variables, which does not coincide with spring-webmvc behaviour.

Example: 

```java
@ResponseBody
@GetMapping("/search")
public String search(@RequestParam("q") String q) {
    return q;
}
```
```bash
curl -s 'http://localhost:8080/search?q=%20%2B+%C3%A0'
```

Spring Web Reactive: `%20%2B+%C3%A0`
Spring Web MVC: `(space)+ à`

(`(space)` is a space character, GitHub seems to remove leading whitespaces...)

---

The cause is that `AbstractServerHttpRequest.getQueryParams()` is returning undecoded name-value pairs, while the interface method `ServerHttpRequest.getQueryParams()` specifies (in javadoc) to return *decoded* ones.

This commit fixed the `AbstractServerHttpRequest` implementation to correctly return
decoded name-value pairs and added an unit test.

https://jira.spring.io/browse/SPR-15140

(I have signed the CLA.)

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractServerHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractServerHttpRequest.java
@@ -16,7 +16,10 @@
 
 package org.springframework.http.server.reactive;
 
+import java.io.UnsupportedEncodingException;
 import java.net.URI;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


rstoyanchev:
This is merged, thanks for the report and pull request. 

Note that I've added an integration test as well and fixed the resulting issues. Also I replaced the use of `URLDecoder` which is not really for urls despite its name (see Javadoc) and replaced it with `UriUtils` which is what we commonly use.

PR #1304, opened 2017-01-24
Support i18n and nested template in ScriptTemplateView

sdeleuze:
This commit changes the 3rd parameter passed to the rendering
function from String url to RenderingContext renderingContext.

RenderingContext contains 4 properties:
 - ApplicationContext applicationContext
 - Locale locale
 - Function<String, String> templateLoader
 - String url

Issue: [SPR-15064](https://jira.spring.io/browse/SPR-15064)


sdeleuze:
@pgrimard Feel free to have a look and say me if that would fit with your need.

pgrimard:
I like the approach you've taken with `RenderingContext`.  The template loader is a nice way to keep templates modular and composable within the render function.  With the introduction of rendering context, the `template` variable seems a little less important.  I see it as a convenience if the render function doesn't need the added functionality of the rendering context, but ultimately the same template could be retrieved simply using the rendering context.  So replacing the `template` argument with the rendering context could be an option, although a breaking change.

sdeleuze:
Indeed it could have been an option initially, but that would be too much breaking now, so I think I prefer the solution where we provide the content of the current template, and provide the `RenderingContext` facility for more advanced use cases (i18n, nested templates, etc.).

pgrimard:
Makes sense 👍

sdeleuze:
Merged via [this commit](https://github.com/spring-projects/spring-framework/commit/2d95199466c52e26ff4c5c3e327e8ae6114e807f).

pgrimard:
This will be release as part of Spring 5?

sdeleuze:
Yes, available in a few minutes in Spring Framework 5 snapshots for both Spring MVC and Spring Web Reactive, and in Spring Framework 5 RC1 and RELEASE.

PR #1305, opened 2017-01-24
Use separate reactive HttpHandler for Tomcat/Jetty

violetagg:
Introduce separate adapters TomcatHttpHandlerAdapter/JettyHttpHandlerAdapter
so that each adapter can use the Tomcat/Jetty APIs for reading/writing
with ByteBuffer.


rstoyanchev:
This is now merged. Thanks!

PR #1307, opened 2017-01-27
SPR-15196 - Check for null on the argument instead of the message

Shredder121:
Don't know if this fix is _truly trivial_, but let me know if a JIRA ticket is required.

I did already sign the CLA, so that should be good.


snicoll:
Thanks for the PR!

Quickly discussed this one with @jhoeller and we should probably revisit that class to avoid that kind of mistake. I've created [SPR-15196](https://jira.spring.io/browse/SPR-15196)

Shredder121:
Good to hear that it's picked up so quickly!

Shredder121:
👍

PR #1308, opened 2017-01-28
Fix trailing / handling in UriComponents(Builder)

sdeleuze:
Issue: SPR-15124


rstoyanchev:
Thanks I have opted for a slightly different fix focusing specifically on the area that introduced the issue, i.e. preserving the trailing slash after parsing the path of a URI template.

The empty path segment approach unfortunately would be inconsistent with how those are treated in other parts, i.e. when not at the end. For example in UriComponentsBuilderTests, the test {{pathSegmentsSomeEmpty}}.

PR #1310, opened 2017-02-02
Fix Javadoc in UriComponentsBuilder

hdpe:
The Javadoc for `UriComponentsBuilder.fromHttpUrl` contains an example that uses the wrong method.

If not paying attention, like me, this might lead you to believe that `fromHttpUrl` could be used with relative URIs, when in fact this leads to a `java.lang.IllegalArgumentException: ... is not a valid HTTP URL`.

This PR alters the Javadoc example to demonstrate correct use of the method.

I'm sure this is an Obvious Fix but I've signed the CLA anyway.


rstoyanchev:
Merged, thanks.

PR #1313, opened 2017-02-03
SPR-15218 - Add logging to processPropertySource() (Obvious Fix)

amdw:
This change set amends ConfigurationClassParser.processPropertySource() to log a message at debug level when a PropertySource with ignoreResourceNotFound = true cannot be loaded for any reason.

Detailed justification for this change, including a link to an illustrative project, may be found on the JIRA:

https://jira.spring.io/browse/SPR-15218

I believe this qualifies as an "obvious fix" which does not require a CLA signature according to the following document:

https://cla.pivotal.io/about#obvious-fixes


jhoeller:
I've adopted your suggestion in a slightly different fashion, including alignment with `PropertiesLoaderSupport` (as used by `<context:property-placeholder>`). Thanks for raising this!

PR #1314, opened 2017-02-03
Introduce JSON streaming support

sdeleuze:
This commit introduces JSON streaming support which
consists of serializing HTTP request with
`application/stream+json` media type as line delimited JSON.

It also optimize `Flux` serialization for `application/json` by
using `flux.collectList()` and a single Jackson invocation
instead of one call per element previous strategy.
This change result in a x4 throughput improvement
for collection with a lot of small elements.

Issues: [SPR-15095](https://jira.spring.io/browse/SPR-15095), [SPR-15104](https://jira.spring.io/browse/SPR-15104)


sdeleuze:
Merged with https://github.com/spring-projects/spring-framework/commit/6b9b0230c4de394348c42b6f5964fc66eb659dd3

PR #1315, opened 2017-02-03
Add ability to set a caffeine ticker

ulich:
This is useful for easier testing of cache implementations. Normally you would have to configure a completely new cache in the tests (e.g. something as described here: http://dolszewski.com/spring/multiple-ttl-caches-in-spring-boot/). Then it makes no more sense to actually write a test at all...

With spring boot you can use a CacheManagerCustomizer to just change the ticker, still testing the real configuration of the cache.

```java
    private static FakeTicker fakeTicker = new FakeTicker();

    @Configuration
    public static class TestConfig {
        @Bean
        public CacheManagerCustomizer<CaffeineCacheManager> customizer() {
            return cacheManager -> cacheManager.setTicker(fakeTicker::read);
        }
    }

    @Test
    public void testRefresh() {
        (...)
        fakeTicker.advance(1, TimeUnit.HOURS);
        (...)
    }
```

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ import com.github.benmanes.caffeine.cache.CacheLoader;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.CaffeineSpec;
 
+import com.github.benmanes.caffeine.cache.Ticker;
 import org.springframework.cache.Cache;
 import org.springframework.cache.CacheManager;
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@ulich Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1315)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1315) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ulich Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1315)!

snicoll:
I am not too keen to make that concept a first class citizen. This really looks like testing only and something that should not pollute the production code.

@ben-manes do you have any opinion about this?

ben-manes:
If you change your configuration by increasing the expiration and your test breaks by not advancing enough, isn't that a bad testing method? Similarly you might want to disable caching to make tests simpler. I would guess that it's preferable to make tests explicit so this method might be undesirable.

snicoll:
Thanks Ben. That confirms my initial reaction.

ulich:
thank you for taking your time!

i was rather thinking of more like acceptance testing, like testing that my app returns a cached value and it also refreshes the cache at some point. maybe for some reason the configuration of `expireAfterWrite` gets lost by accident or is no longer supported in a newer version (and you configured it via properties in spring-boot, so no compiler checks), then you have an app that never ever refreshes the value again. there is no way of ensuring this right now.

what I basically want to test is the cache configuation and if my configuration does what i want.

for not having to adapt the test when the expiry time is changed: it is just like any other test where you rely on some constant value: you could create a constant out of it and reuse it in the tests...

Maybe you could rethink now that I elaborated. But I am also not angry if it does not get merged :)

PR #1321, opened 2017-02-09
SPR-15238 - Use getOrDefault method

jonathanborenstein:
Use the getOrDefault method for Maps in Java 8.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/ResolvableType.java
+++ b/spring-core/src/main/java/org/springframework/core/ResolvableType.java
@@ -1229,7 +1229,7 @@ public class ResolvableType implements Serializable {
 	public static ResolvableType forMethodParameter(MethodParameter methodParameter, ResolvableType implementationType) {
 		Assert.notNull(methodParameter, "MethodParameter must not be null");
 		implementationType = (implementationType != null ? implementationType :
-				forType(methodParameter.getContainingClass()));
+			forType(methodParameter.getContainingClass()));
 		ResolvableType owner = implementationType.as(methodParameter.getDeclaringClass());
 		return forType(null, new MethodParameterTypeProvider(methodParameter), owner.asVariableResolver()).
 				getNested(methodParameter.getNestingLevel(), methodParameter.typeIndexesPerLevel);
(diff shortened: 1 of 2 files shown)


jonathanborenstein:
https://jira.spring.io/browse/SPR-15238

jhoeller:
See my comment on SPR-15238. These two cases are good examples for `getOrDefault` enforcing early construction/retrieval of default values even if we're likely to encounter a cache hit where we don't need the default value in the first place.

jonathanborenstein:
https://github.com/jonathanborenstein/spring-framework/commit/79d8cf0acb76069e46e2544a7400313a6cefbb64

I did add two more uses of it, but didn't make a pull request. Would it be applicable in either of those situations?

jhoeller:
Class loading is also pretty expensive, so that's not an ideal use there either.

`AbstractSqlParameterSource.java ` with its `TYPE_UNKNOWN` constant is the only that I see as a good tradeoff. No need for a pull request for that one, I'll roll that it along with some other fine-tuning today.

jonathanborenstein:
Fair enough. Do I get any credit for the change?

jhoeller:
I'll reopen your SPR-15238 ticket, repurposing it for this particular change.

jonathanborenstein:
Also, about the getOrDefault method, are both the get and default value initiated at the same time, which is the reason that it is slower?

For example 

<code>public Class<?> loadClass(String name) throws ClassNotFoundException {
  		if (shouldShadow(name)) {		  		if (shouldShadow(name)) {
 -			Class<?> cls = this.classCache.get(name);		 +			return this.classCache.getOrDefault(name, doLoadClass(name));</code>

Why would doLoadClass(name) be more expensive? I thought it would only be called if a null value was found.

jhoeller:
Well, the problem is that - with Java method call semantics - all parameter values are being resolved, then passed into the method. In order to get deferred execution of {{doLoadClass(name)}} here, it would have to be wrapped in a `Supplier` or the like, to be called back only when the default value is needed.

`Map` actually has a variant of this in `default V computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction)`. That's only useful for adding values to the map though, not for fallback default values.

PR #1323, opened 2017-02-12
Add a Kotlin RouterFunction implementation

melgenek:
The reason for this PR is to give opportunity to create functional controllers in Kotlin without implementing `RouterFunction.route` function but writing mappings inside the controller.
Insired by [Twitter Finatra](https://github.com/twitter/finatra/blob/develop/examples/hello-world/src/main/scala/com/twitter/hello/HelloWorldController.scala) "server as a function principle" and [Spek](http://spekframework.org/)  - Kotlin testing framework.

Sample controller can look like this:
```kotlin
class SampleController : KotlinRouterFunction({
    GET("/super") { req ->
        ServerResponse.ok().body(fromObject("super!!!"))
    }
})
```

### DIFF
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/RouterFunctionExtensions.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/RouterFunctionExtensions.kt
@@ -9,7 +9,7 @@ import reactor.core.publisher.Mono
  * Provide a routing DSL for [RouterFunctions] and [RouterFunction] in order to be able to
  * write idiomatic Kotlin code as below:
  *
- * * ```kotlin
+ * ```kotlin
  * fun route(request: ServerRequest) = route(request) {
  * 		accept(TEXT_HTML).apply {
  * 			(GET("/user/") or GET("/users/")) { findAllView() }
@@ -46,7 +46,7 @@ class RouterDsl {
 	}
 
 	fun GET(pattern: String, f: (ServerRequest) -> Mono<ServerResponse>) {
-		routes += RouterFunctions.route(RequestPredicates.GET(pattern), HandlerFunction { f(it) } )
+		routes += RouterFunctions.route(RequestPredicates.GET(pattern), HandlerFunction { f(it) })
 	}
 
 	fun HEAD(pattern: String, f: (ServerRequest) -> Mono<ServerResponse>) {
@@ -81,7 +81,7 @@ class RouterDsl {
 		routes += RouterFunctions.route(RequestPredicates.contentType(mediaType), HandlerFunction { f(it) })
 	}
 
-	fun headers(headerPredicate: (ServerRequest.Headers)->Boolean, f: (ServerRequest) -> Mono<ServerResponse>) {
+	fun headers(headerPredicate: (ServerRequest.Headers) -> Boolean, f: (ServerRequest) -> Mono<ServerResponse>) {
 		routes += RouterFunctions.route(RequestPredicates.headers(headerPredicate), HandlerFunction { f(it) })
 	}
 
@@ -94,13 +94,12 @@ class RouterDsl {
 	}
 
 	fun resources(path: String, location: Resource) {
-		routes +=  RouterFunctions.resources(path, location)
+		routes += RouterFunctions.resources(path, location)
 	}
 
 	fun resources(lookupFunction: (ServerRequest) -> Mono<Resource>) {
-		routes +=  RouterFunctions.resources(lookupFunction)
+		routes += RouterFunctions.resources(lookupFunction)
 	}
-
 
 	@Suppress("UNCHECKED_CAST")
 	fun router(): RouterFunction<ServerResponse> {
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@melgene Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1323)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1323) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@melgene Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1323)!

sdeleuze:
@melgene Thanks for this PR, I like the idea but I have concerns about how that will works when you need to call `SampleController` member functions. Could you show an example?

melgene:
@sdeleuze There is no possibility to access class members as well as field injection (only default constructor injection). But in practice controllers most likely call business services and define request mappings.

The problem I came up with while writing this message is unit testing of such controllers. And I don't have solution for this issue.

sdeleuze:
After discussing with @poutsma, I think we will have to move to constructor based routing because current implementation is suboptimal (route are evaluated for each request and not just at `RouterFunction` creation. I guess that's a mater of style, but the testing issue is something we should care about.

I guess that means such controller should be tested using [`WebTestClient`](https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/test/web/reactive/server/WebTestClient.java)  and/or [`MockServerHttpRequest`](https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java).

I plan to experiment a little bit on [MiXiT](https://github.com/mix-it/mixit/) before eventually committing something on Spring Framework.

rstoyanchev:
The WebTestClient is just executing WebClient against a RouterFunction with MockServerHttpRequest and Response. That's as transparent as it gets. I see no reason why that shouldn't be the preferred way to begin with.

melgene:
>  current implementation is suboptimal (route are evaluated for each request and not just at RouterFunction creation

@sdeleuze, I noticed such behaviour. This issue will raise in Java as well as in Kotlin. Possible solution is to implemet lazy router function. The following can be done using Kotlin, but should be fixed for java too.
```kotlin
abstract class LazyRouterFunction : RouterFunction<ServerResponse> {
	val routerFunction: RouterFunction<ServerResponse> by lazy {
		RouterDsl().apply(routes).router()
	}
	abstract val routes: RouterDsl.() -> Unit
	override fun route(request: ServerRequest): Mono<HandlerFunction<ServerResponse>> {
		return routerFunction.route(request)
	}
}
```
And the usage is like the following: 
```kotlin
@Controller
class LazyController : LazyRouterFunction() {
	override val routes: RouterDsl.() -> Unit = {
		(GET("/foo/") or GET("/foos/")) { req -> handle(req) }
		accept(MediaType.APPLICATION_JSON).apply {
			DELETE("/api/foo/", this@LazyController::handle)
		}
	}
	fun handle(req: ServerRequest) = ServerResponse.ok().build()
}
```

@sdeleuze @rstoyanchev There is a question I am concerned about : complexity of request handler 
search. It seems that the `RouterFunction` approach in worst case can reach `O(n)` if the last `n`th function is the one that applies to request. Does `RequestMapping` makes things better, closer to hashmap complexity `O(1)`?

poutsma:
>  It seems that the RouterFunction approach in worst case can reach O(n) if the last nth function is the one that applies to request. Does RequestMapping makes things better, closer to hashmap complexity O(1)?

That really depends on the specific scenario. Though we do use a hash map for path-based lookups in the `@RequestMapping` processing, the only way a such a lookup can match if the request path is an exact match of the pattern defined in the `@RequestMapping` annotation. If the pattern uses a wildcard (`/*`) or path variable (`/{foo}`), then the hasmap-based lookup will not come up with a match, so we have no choice but to iterate over all existing mappings, and see which match. Next, we must also consider the other elements of the annotation, such as the `headers`, `produces` and `consumes` elements. We do this by sorting all potential hits according to their specificity: more specific matches come before more generic ones. This can even result in ambiguous matches, which result in an exception. Note that this process described above is a simplified version of the real algorithm, look at `org.springframework.web.servlet.mvc.method.RequestMappingInfoHandlerMapping` and its superclasses for more details on this process.

Router Functions, on the other hand, work quite differently. For instance,`RouterFunction` does not consider all potential matches: the first one that "hits" is the one that is picked. Furthermore, with nested routes, it is possible to optimise the searching space by placing request predicates under a specific constraint (such as a path prefix), so that only the candidates under that constraint will be considered (instead of all candidates, like with `@RequestMapping`).

sdeleuze:
@melgene The Kotlin router available in Spring Framework 5.0 M5 now supports path prefix + nested routes, see https://github.com/mix-it/mixit/blob/master/src/main/kotlin/mixit/controller/UserController.kt for an example.

About lazy route evaluation, both your proposed solution with `lazy` Kotlin keyword and the [RouterFunctionProvider](https://github.com/mix-it/mixit/blob/master/src/main/kotlin/mixit/support/RouterFunctionProvider.kt) in introduced in MiXiT app works. @poutsma is think about something usable both in Java and Kotlin to provide this behaviour by default, we will create a dedicated issue shortly on JIRA about that.

Feel free to create a JIRA issue or another PR if you think we can improve the syntax. Being able to call a member function is a must have.

PR #1330, opened 2017-02-18
Add tests for AbstractRoutingDataSource

kazuki43zoo:
I've added test case class for the `AbstractRoutingDataSource`.


snicoll:
Thanks again @kazuki43zoo and sorry it took so long to process this.

PR #1331, opened 2017-02-20
Make PathPatternParser multi-threaded

aclement:
With this change the original PathPatternParser is renamed
InternalPathPatternParser and a new PathPatternParser class is added.
This new PathPatternParser class is a very simple thread-safe wrapper for
the InternalPathPatternParser. It achieves this by creating a new
InternalPathPatternParser for each new parse request.  This follows the
model used for SpEL parsing.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/patterns/InternalPathPatternParser.java
+++ b/spring-web/src/main/java/org/springframework/web/util/patterns/InternalPathPatternParser.java
@@ -1,0 +1,408 @@
+/*
+ * Copyright 2002-2017 the original author or authors.
+ *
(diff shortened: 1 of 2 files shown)


aclement:
Hey @bclozel - this implements what we talked about to make it multi threaded. I submitted it as a PR for you to look over. Don't feel you have to take it if you don't like the approach! Feedback welcome.

bclozel:
Thanks @aclement ! I've merged this with a0505bf and removed the concurrency checks in `ParsingPathMatcher` which are not necessary anymore.
Right on time for the next milestone.

PR #1335, opened 2017-02-21
Refactor with String#isEmpty()

stonio:
Polish few classes with `.isEmpty()` instead of `.length() == 0`.


snicoll:
@stonio thanks for the PR, we prefer a single commit rather than small atomic commits. Can you please squash all that and push force your branch?

stonio:
@snicoll All my commits squashed into a single commit.
May be backported to 4.3.x branch.

lqjack (commented):
does the String (empty()) better then String (.length())?

snicoll:
Thanks, we'll backport it to `4.3.x` later today.

PR #1337, opened 2017-02-24
Update AnnotationUtils.java

jonathanborenstein:
https://jira.spring.io/browse/SPR-15287

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationUtils.java
@@ -803,10 +803,8 @@ public abstract class AnnotationUtils {
 		Assert.notNull(annotationType, "Annotation type must not be null");
 		Assert.notNull(clazz, "Class must not be null");
 		try {
-			for (Annotation ann : clazz.getDeclaredAnnotations()) {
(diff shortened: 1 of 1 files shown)


jhoeller:
See my comment on JIRA: I've adopted a different variant, using `getDeclaredAnnotation` instead.

jonathanborenstein:
Did you do something like this instead?

if (clazz.getDeclaredAnnotation(annotationType) != null){
				return true;
			}

jhoeller:
Indeed, and similar `getDeclaredAnnotation` calls returning the annotation found (if any) in other places. This seems to be the most efficient variant for our purposes, in particular when called against a `Class`.

I'll push this later tonight, along with a few other commits.

jonathanborenstein:
https://github.com/spring-projects/spring-framework/pull/1338 - I updated the patch2 with the getDeclaredAnnotation method for the findAnnotation methods where the first parameter is an AnnotatedElement and for when the first parameter is a class. Also for the isAnnotationDeclaredLocally method.

I should have made this one pull-request/JIRA issue to begin with.

jonathanborenstein:
https://github.com/spring-projects/spring-framework/pull/1338/files

That's the change I just made. It appears an Annotation array does not even need to be instantiated for both the findAnnotation methods I've updated.

PR #1344, opened 2017-03-03
Fix JRubyException message access

wltjr:
I do not think I created this PR properly but maybe. Newer JRuby changed access on message in JRubyException to private. This causes an error when attempting to use a private variable in public scope. I changed it from .message to ex.getMessage(). Also removed null check, as it is only needed on the JRubyException object.

If the PR is in fact improperly created, let me know and I will attempt to recreate/resubmit properly. This pertains to 4.3.7 release/branch.


pivotal-issuemaster:
@wltjr Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1344)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1344) to manually synchronize the status of this Pull Request.

snicoll:
Thanks for the PR. please resubmit with master, we are taking care of backports anyway.

wltjr:
The file has been removed from master. I resubmitted against the 4.3.x branch

jhoeller:
I've addressed this through a slightly different revision, retaining compatibility with JRuby 1.7.x while also allowing `JRubyScriptFactory` to be used with JRuby 9.1.7+ still.

Note that we primarily support JRuby 9000 through JSR-223, i.e. `StandardScriptFactory`, and that native JRuby support is deprecated since Spring Framework 4.2 already and therefore gone in 5.0.

wltjr:
Ok sounds good

PR #1345, opened 2017-03-03
SPR-15683 Avoid NullPointerException if cron expression is null

tonymanou:
When you pass a null cron expression to the constructor of ```CronTrigger```, it creates a ```CronSequenceGenerator``` instance which calls the ```parse()``` method from its constructor.
Inside this method the cron expression is first tokenized by ```StringUtils.tokenizeToStringArray()``` (which will return null if the cron expression is null) and put into the ```fields``` variable. Then, if the content of ```fields``` is considered invalid (null or not having the right number of elements), an ```IllegalArgumentException``` is thrown.
But as we are using ```fields.count```'s value to create the message of the exception, we get a ```NullPointerException``` (with no message) instead when ```fields``` is null.

These changes let the ```parse()``` method throw an ```IllegalArgumentException``` with a message when ```fields``` is null.

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/support/CronSequenceGenerator.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/support/CronSequenceGenerator.java
@@ -261,7 +261,9 @@ public class CronSequenceGenerator {
 	 */
 	private void parse(String expression) throws IllegalArgumentException {
 		String[] fields = StringUtils.tokenizeToStringArray(expression, " ");
-		if (!areValidCronFields(fields)) {
+		if (fields == null) {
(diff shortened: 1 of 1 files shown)


snicoll:
@tonymanou could you please create an issue in the tracker for this?

tonymanou:
Sorry for the delay, I just created the JIRA issue [SPR-15683](https://jira.spring.io/browse/SPR-15683)

tonymanou:
Oh well... Seems like PR #1442 provide a fix for this issue as well.

snicoll:
Indeed :) - Thanks for the PR anyway

PR #1349, opened 2017-03-05
Doc fix for example of GenericApplicationContext & representation of ConditionalGenericConverter

aviskarbasnet:
Making easier for understanding example of GenericApplicationContext by changing undeclared object `ctx` to `context`. And, also fix for representation of ConditionalGenericConverter interface. I think these changes are essential.


snicoll (approved):

snicoll:
Thanks for the pr!

PR #1352, opened 2017-03-07
[Documentation] Updated URI Template reference links

pgpx:
Updated link to URI Template reference - the existing link no longer works/exists.  Instead I've linked to the RFC and related GitHub repository.

### DIFF
--- a/src/asciidoc/web-mvc.adoc
+++ b/src/asciidoc/web-mvc.adoc
@@ -771,9 +771,8 @@ __URI templates__ can be used for convenient access to selected parts of a URL i
 `@RequestMapping` method.
 
 A URI Template is a URI-like string, containing one or more variable names. When you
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Thanks for the pull request.

The problem is that RFC 6570 provides very rich syntax that goes way beyond what `UriTemplate`  supports and takes a different approach to expanding and encoding. It's comparable to what is described as level 1 in the RFC.

I think we should just drop the link and explain in simple terms what a URI template with an example or two and optionally provide a note on RFC 6570.

A little [more background here](https://jira.spring.io/browse/SPR-12942?focusedCommentId=115649&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-115649).

rstoyanchev:
I've updated the docs as mentioned. Thanks for the suggestion and pull request.

PR #1353, opened 2017-03-08
spring-jdbc meta data prcoessing bugfix

seanmmills:
ResultSet.getBoolean doesn't describe how to handle anything other than
NULL, 0, or 1. Since the 'NULLABLE' column can return 2, meaning
unknown, the code should change to process the ResultSet with the
explicitly defined types.

Issue: SPR-15333


seanmmills:
If approved, can this be cherry-picked to the 4.3.x branch as well?

snicoll:
@seanmmills thanks for the PR. Sorry we failed to process it, we usually add a link from the Jira issue but we forgot that for some reason.

PR #1355, opened 2017-03-13
SPR-11587: Support multiple TilesContainer per ServletContext

tkrah:
Why i did this - because Spring fails to properly support multiple TilesContainer per ServletContext because its using a "null" key mapping, multiple container in different FrameworkServlets will override each others factory. See ticket for details.

Question to be answered:

1. Should i include visibility/extension changes according to the ticket in this one or do another changeset - still tbd?
2. Are the test enhancements / naming made sufficient?

Discussion in general:

Still need to polish it to the coding guidelines - but want to get early feedback on the general refactoring 
stuf, if it fits your expectations or how todo it better.
Package visibility access on many properties is not my favourite - should i refactor those in getters or pass the needed references to each instance via construction parameters? Any other ideas?

ICLA  submitted.

Issue: SPR-11587


rstoyanchev:
Closing since Tiles support has been dropped for 6.0 with #27423.

PR #1356, opened 2017-03-14
MediaTypeFactory should come with its own complete mime.types file

poutsma:
For your consideration, @jhoeller and @rstoyanchev, here is my PR for resolving https://jira.spring.io/browse/SPR-14908. I've opted to split the PR in three commits: the first showing the `MediaTypeFactory` changes, the second fixes the main code base to conform to these changes, and the third does the same for the test code base.

As per discussion with Juergen, I've dropped all JAF usage in favour of the new `MediaTypeFactory`, including that in pre 5.0 code, and deprecate the various `useJaf` properties we had in that code, with no direct replacement.


sbrannen:
What about the use of `javax.activation.FileTypeMap` in our `MockServletContext` implementations?

poutsma:
> What about the use of javax.activation.FileTypeMap in our MockServletContext implementations?

I decided to keep that as is, since the MediaTypeFactory does not have a manual override mechanism, and the Javadoc comment on that method seems to suggest that overriding is vital.

sbrannen:
`MockServletContext.getMimeType()` doesn't mention anything about overriding the method. 

In any case, it can't be overridden if the _Spring TestContext Framework_ or parts of Spring MVC Test are creating the mock for the user transparently within a test. So, in such situations, if JAF is not available (e.g., on Java 9), then the tests would no longer work.

Or am I missing something?

poutsma:
I wasn't talking about method overriding; I was talking about overriding or customising the mime types returned by `MockServletContext.getMimeType`. JAF has such a customisation mechanism, and as you know there is a [use case for it](https://github.com/spring-projects/spring-framework/blob/199aa776c96d9a50ff4fcbf18673eb59960c6d48/spring-test/src/test/java/org/springframework/mock/web/MockServletContextTests.java#L90-L90).

Except for replacing the `mime.types` file on the classpath, the new MediaTypeFactory does not have such a customisation mechanism. Its goal is to provide default mimes types, when none are provided through `PathExtensionContentNegotiationStrategy` or similar mechanisms (which are customisable).

So, in summary: JAF has a customisation mechanism; the new MediaTypeFactory does not. `MockServletRequest.getMimeType` needs such a mechanism, so it does not use MediaTypeFactory.

sbrannen:
OK, I now understand where you're coming from.

So... I guess we are then going to make JAF a requirement for the testing frameworks. I can live with that; we'll just have to document it.

My real concern was that production code and test code would demonstrate different behavior with regard to available mime types _out of the box_.

Do you not foresee any issues with that?

sbrannen:
By the way, nice investigative work on finding that _use case_. 😉

sbrannen:
The changes in commit 689f0d716a827e42db297ad41006c1a97486c213 look *much* more robust than what we had before.

Thanks, @poutsma!

rstoyanchev:
I've processed the PR to avoid further merge conflicts. 

That said we do need an alternative to the deprecated `usefJaf` property starting with `ContentNegotiationManagerFactoryBean` and `RequestedContentTypeResolverBuilder` and then all the places it was propagated to. This property serves an important purpose in allowing automatic lookups of extensions to be suppressed in favor of using only explicitly registered extensions. This can be a source of security issues.

We should also consider some improvement to `MediaTypeFactory#getMediaType(Resource)` which in all places where it's used it's then followed up with fallback on application/octet-stream. Perhaps `getMediaTypeOrDefault` or return an `Optional`.

poutsma:
`getMediaType` now returns an `Optional<MediaType>`.

poutsma:
I've pushed a branch [here](https://github.com/poutsma/spring-framework/tree/SPR-14908) that introduces a `useDefaults` property, taking the role of the `useJaf` property. Let me know what you think, @rstoyanchev.

PR #1357, opened 2017-03-14
Extract CompositeHttpHandler functionality

philwebb:
Extract CompositeHttpHandler to a package private class and add direct
support via `HttpHandler.of(...)`. This removes the need for the
`HttpHandlerAdapterSupport` class.

See https://github.com/spring-projects/spring-boot/issues/8559 for background.


spring-web/src/main/java/org/springframework/http/server/reactive/HttpHandler.java
```diff
@@ -34,4 +37,14 @@
 	 */
 	Mono<Void> handle(ServerHttpRequest request, ServerHttpResponse response);
 
+	/**
+	 * Return a composite {@link HttpHandler} that maps multiple
+	 * {@link HttpHandler}s each mapped to a distinct context path.
+	 * @param handlerMap the source handler map
+	 * @return a composite {@link HttpHandler}
+	 */
+	static HttpHandler of(Map<String, ? extends HttpHandler> handlerMap) {
```
284831721: gr8 job

spring-web/src/main/java/org/springframework/http/server/reactive/ReactorHttpHandlerAdapter.java
```diff
@@ -16,43 +16,46 @@
...
+	private static final Log logger = LogFactory.getLog(ReactorHttpHandlerAdapter.class);
 
-	public ReactorHttpHandlerAdapter(HttpHandler httpHandler) {
-		super(httpHandler);
-	}
 
-	public ReactorHttpHandlerAdapter(Map<String, HttpHandler> handlerMap) {
-		super(handlerMap);
-	}
+	private final HttpHandler httpHandler;
```
284831721: nice!  is a ----> has a

rstoyanchev:
This is now in with a slight deviation promoting `CompositeHttpHandler` to a public class and also calling it `ContextPathCompositeHandler`. I prefer to make that more explicit and the slightly reduced convenience doesn't have much impact at this very low level of abstraction. Instead I've overhauled the Javadoc on `HttpHandler` and `ContextPathCompositeHandler` which should server better for discoverability and orientation.

PR #1360, opened 2017-03-21
ServerHttpMessage[Reader|Writer] refactoring

rstoyanchev:
The refactoring removes the separate class hierarchy under `ServerHttpMessage[Reader|Writer]`, including `AbstractServerHttpMessage[Reader|Writer]` and `Jackson2ServerHttpMessage[Reader|Writer]`, which used to wrap and delegate to the "regular" `HttpMessage[Reader|Writer]`.

Instead `[Encoder|Decoder]HttpMessage[Writer|Reader]` now have an expanded role covering the server-side contract too. There is also a new `ServerHttp[Encoder|Decoder]` that allows encoders and decoders to encapsulate the preparation of the very hints they also look for. The approach works well for the SSE writer which no longer needs to be wrapped with the `Jackson2ServerHttpMessageWriter` to get the encoding hints needed for the `Jackson2Encoder`.

See individual commits for more details.


sdeleuze:
Nice refactoring @rstoyanchev, I like the way it implements the array versus stream behavior we gradually come to, the better isolation of hint resolution and the fact we can indeed remove the separate class hierarchy for server reader/writer.

I have just a remark : when we are talking about `Server` (for reader, writer or codecs) it seems to me we are referring more to the annotation-based programming model since I think it is not suitable for the functional Web server API for example. So I am wondering if `Server`is the right terminology for these classes. I may be ok as a "Annotation-based server" shortcut, but since it could be confusing for the functional server API, I prefer to raise the point to see what you think about that.

rstoyanchev:
Okay thanks for the review!

I guess those names aren't new but I do see your point. Especially since the last commit switched to using `ServerHttpMessageWriter` but only for annotated controllers. The functional side uses code shared with the client (`BodyInserter` and `BodyExtractor`) and it's currently not feasible. Perhaps something to revisit, I've created https://jira.spring.io/browse/SPR-15370 for that.

PR #1362, opened 2017-03-21
Add preferred content type configurer method

romeara:
Applications which support end points with a variety of content types often want to specify a default set of content types to "prefer" when a client does not request a specific content type. Such a list would replicate a set of default content types with "quality" parameters (such as those used in accept headers) for requests which accept all (*/*) content types.

The existing functionality to set the default content type will cause any end point which does not support the specified default to return a 406 status code when "*/*" (or nothing) is provided in an Accept header - however, based on the documentation, it appears this is intended. For this reason, this PR adds a new method to the configurer, as opposed to enhancing the defaultContentType method.

I proposed this addition/enhancement in JIRA issue [SPR-15367](https://jira.spring.io/browse/SPR-15367)
I have also submitted ICLA per the contribution guidelines

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/accept/ContentNegotiationManagerFactoryBean.java
+++ b/spring-web/src/main/java/org/springframework/web/accept/ContentNegotiationManagerFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


rstoyanchev:
Generally I like the idea but for the implementation, first `PreferredContentNegotiationStrategy` should not extend `HeaderContentNegotiationStrategy` only to check if anything has been requested. The strategies are meant to not be aware of each other. The `ContentNegotiationManager` delegates to each at a time and checks if each strategy returned anything other than `*/*`. 

Second I'd prefer for `defaultContentType` and `FixedContentNegotiationStrategy` to be enhanced rather than introducing a new kind of strategy. The former would become `defaultContentType(MediaType...)` and latter would accept a `List<MediaType>`. You would have to insert "*/*" at the end of the list yourself but I think that's a trivial requirement without adding more options.

romeara:
@rstoyanchev - I'd be more than happy to update this PR per your recommendations. The first point makes sense, and looking more into the documentation, it seems the strategies are meant to assume that the client did not use an accept header, so I'll just remove that check.

The second was my first thought on this enhancement, but I wasn't sure if the group would be comfortable with the change in existing behavior it will cause. I agree having people add "*/*"themselves (and documenting why they would want to) is a good alternative which prevents adding additional APIs.

I did my best to follow the other contribution guidelines - if anything else needs fixing, please let know and I'll be more than happy to adjust it as well

romeara:
@rstoyanchev - I have made the requested changed and squashed/force-pushed to update the branch. Let me know if any further changes are desired. Thanks!

rstoyanchev:
I've applied this to master with some minor updates to preserve existing methods accepting a single MediaType. Thanks for the PR and update!

PR #1363, opened 2017-03-22
Use a WebClient instead of a RestTemplate

jonathanborenstein:
Replace the RestTemplate with a WebClient.

### DIFF
--- a/spring-webflux/src/test/java/org/springframework/web/reactive/function/server/PublisherHandlerFunctionIntegrationTests.java
+++ b/spring-webflux/src/test/java/org/springframework/web/reactive/function/server/PublisherHandlerFunctionIntegrationTests.java
@@ -16,32 +16,32 @@
 
 package org.springframework.web.reactive.function.server;
 
-import java.net.URI;
+import static org.junit.Assert.assertEquals;
+import static org.springframework.web.reactive.function.BodyExtractors.toMono;
+import static org.springframework.web.reactive.function.BodyInserters.fromPublisher;
+import static org.springframework.web.reactive.function.server.RequestPredicates.GET;
+import static org.springframework.web.reactive.function.server.RequestPredicates.POST;
+import static org.springframework.web.reactive.function.server.RouterFunctions.route;
+
 import java.util.List;
 
 import org.junit.Test;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.reactive.function.BodyInserters;
+import org.springframework.web.reactive.function.client.ClientResponse;
+import org.springframework.web.reactive.function.client.WebClient;
+
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import org.springframework.core.ParameterizedTypeReference;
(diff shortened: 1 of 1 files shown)


pretizy (approved):

rstoyanchev:
hi what's the rational for the change? Having the RestTemplate used in some WebFlux tests at least is actually a good test. What client is used should not matter.

jonathanborenstein:
I thought that since the WebClient is part of the WebFlux module, it should be used in those tests. I think it also might be simpler to understand than the RestTemplate.

What are the proper use cases for certain classes in the WebFlux module? Why even use a RouterFunction when you could just use the RequestMapping and return a Flux?

jonathanborenstein:
Also, there are very few examples out of all the tests showing how to actually use the post method for the WebClient.

rstoyanchev:
Okay I see. Yes there is `WebClientIntegrationTests` for example. If you're missing anything in particular as an example perhaps you can suggest such a test there?

As for the `RestTemplate` many tests use it simply because we started out with server support. After `WebClient` was added we considered re-writing tests but did not do that since it is a good idea to have a mix. The RestTemplate uses different HTTP library underneath and we want to know if that works too.

jonathanborenstein:
Hmm. I definitely think some sort of side by side comparison should be used with a RestTemplate and a WebClient, just to see how they both work and how they can be used.

As far as the WebClientIntegrationTests, it's not using a RouterFunction, I don't think.

There's just very little in terms of showing how a WebClient interacts with a RouterFunction. And there is nothing showing a WebClient posting to a RouterFunction.

That's why I thought this commit might be worthwhile, at least you can see a post request to a RouterFunction. Maybe it needs to be cleaned up a bit.

rstoyanchev:
@poutsma do you have any opinion either way since this is in `PublisherHandlerFunctionIntegrationTests`? Perhaps we should have some dedicated tests with `WebClient` and `RouterFunction` on the server side.

poutsma:
The reason I used RestTemplate is that it is currently more robust: we can assume that it has less bugs than WebClient, because it has been around longer. A failing test should indicate the smallest possible number of components that have a potential failure. Using WebClient would not help in this respect: a test failure could indicate either a RouterFunction or WebClient bug, which is not as useful. If anything, we shouldn't  be using RestTemplate either, but a third party HTTP client.

jonathanborenstein:
For WebClientIntegrationTests, you can replace the get() methods with delete() and it still passes.

Also, the RouterFunction returns a ServerResponse, but would it be better to return an EntityReponse?

rstoyanchev:
@jonathanborenstein, thanks for trying out the new features but this is isn't a discussion forum. For further learning please take a look at the documentation on the functional web framework [in progress](https://github.com/spring-projects/spring-framework/blob/master/src/asciidoc/web-flux-functional.adoc)) and also the [list of examples](http://docs.spring.io/spring/docs/5.0.0.BUILD-SNAPSHOT/spring-framework-reference/htmlsingle/#web-reactive-getting-started-examples).

PR #1365, opened 2017-03-24
typo fix in docs

gavvvr:


jhoeller:
Good catch! This entire paragraph is suboptimal, so I'll revise it wholesale instead of just fixing that typo.

Thanks for the PR in any case...

PR #1375, opened 2017-04-04
SPR-15413: Add Kotlin coroutines for ListenableFuture

konrad-kaminski:
Prior to this commit there was no support for converting Kotlin
**coroutines** to/from **ListenableFuture**. Following that change a number of
functions will be available for such conversion.

Specifically these functions are introduced:

 - // for converting suspending lambda to ListenableFuture
   `fun <T> listenableFuture(context: CoroutineContext = CommonPool,
       block: suspend () -> T): ListenableFuture<T>`

 - // for converting Deffered into ListenableFuture
   `fun <T> Deferred<T>.asListenableFuture(): ListenableFuture<T>`

 - // for converting ListenableFuture into suspending function
   `suspend fun <T> ListenableFuture<T>.await(): T`

The code in this PR draws heavily on the code from https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-jdk8/src/main/kotlin/kotlinx/coroutines/experimental/future/Future.kt. This is why Roman Elizarov is also an author.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,6 +72,7 @@ configure(allprojects) { project ->
 	ext.junitVersion           = "4.12"
 	ext.junitJupiterVersion    = '5.0.0-M4'
 	ext.junitPlatformVersion   = '1.0.0-M4'
+	ext.kotlinCoroutinesVersion= "0.14.1"
 	ext.kotlinVersion          = "1.1.1"  // also change kotlin-gradle-plugin version when upgrading
 	ext.log4jVersion           = '2.8.1'
 	ext.nettyVersion           = "4.1.9.Final"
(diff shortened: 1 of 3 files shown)


enleur:
Just to have a reference to the issue https://jira.spring.io/browse/SPR-15413

snicoll:
@sdeleuze can you please check if this is still applicable?

poutsma:
After team discussion, we have decided not to invest in `ListenableFuture`. It can easily be converted into a `CompletableFuture, which has Kotlin support through `kotlinx-coroutines-jdk8`.

rstoyanchev:
Also see #27780.

sdeleuze:
+1 for that decision.

PR #1376, opened 2017-04-05
SPR-14925: Return 5xx HTTP status for invalid target types with Jackson

sdeleuze:
InvalidDefinitionException has been introduced in Jackson 2.9 to be
able to differentiate invalid data sent from the client (should still
generate a 4xx HTTP status code) from server side errors like beans with
no default constructor (should generate a 5xx HTTP status code).

Could you have a look and say me if you are ok @poutsma ?

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/HttpMessageConversionException.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/HttpMessageConversionException.java
@@ -17,15 +17,21 @@
 package org.springframework.http.converter;
 
 import org.springframework.core.NestedRuntimeException;
+import org.springframework.http.HttpStatus;
+
+import java.util.Optional;
 
 /**
  * Thrown by {@link HttpMessageConverter} implementations when a conversion attempt fails.
  *
  * @author Arjen Poutsma
+ * @author Sebastien Deleuze
  * @since 3.0
  */
 @SuppressWarnings("serial")
 public class HttpMessageConversionException extends NestedRuntimeException {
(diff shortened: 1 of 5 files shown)


spring-web/src/main/java/org/springframework/http/converter/HttpMessageConversionException.java
```diff
@@ -17,22 +17,29 @@
...
  * Thrown by {@link HttpMessageConverter} implementations when a conversion attempt fails.
  *
  * @author Arjen Poutsma
+ * @author Sebastien Deleuze
  * @since 3.0
  */
 @SuppressWarnings("serial")
 public class HttpMessageConversionException extends NestedRuntimeException {
 
+	private final Optional<HttpStatus> errorStatus;
```
poutsma: In general, `Optional`s should only be used as a return type, not a field value. So I would store this as a `HttpStatus` field with a potential null value. See http://blog.joda.org/2014/11/optional-in-java-se-8.html for reasons why.

sdeleuze: Make sense, fixed

jhoeller: We should generally remain defensive against overuse of `Optional`. So agreed, let's avoid it as field type and only expose it for method signatures where it nicely aligns with typical use of the method.

sdeleuze:
Merged

PR #1377, opened 2017-04-06
webmvc: fix AbstractEmitterSubscriber

akarnokd:
@smaldini brought this class to my attention and asked me to verify it. There were a couple of problems with the `AbstractEmitterSubscriber`:

- Using a shared queue for items, error and completion signals. The problem with this is that if the user is in the business of processing `Exception`s as values, the logic may treat it as an error even if it came through `onNext`.
- Since `onError` and `onComplete` are at most once events, there is no need to queue them up but could be simply saved in a field and read out in the emission side.
- Now that the terminal events have their own fields, the queue can be simplified to an `AtomicReference` because this class requests items one-by-one.
- It is unclear where the `onTimeout` body would trigger in respect to the regular signals, thus calling `emitter` directly from it is likely violating the serialization requirement of the emitter. The code now pretends to call `onComplete` which gets properly serialized with respect to any other signal.
- In the handler method, if the `send` method crashed, the exception was not forwarded to the emitter for some reason. I assume the error should be reported since the source is otherwise cancelled and won't send any new item/signal.
- Changed the `executing` to a linearizing atomic counter because the original CAS+Set based trampolining is generally hard to verify on not losing request for scheduling.

I also kept the one-by-one nature and the rescheduling logic assuming the intent was not to "cannibalize" the executor if there was plenty of source items available and behave "fairly" with other tasks running on the same executor.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ReactiveTypeHandler.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ReactiveTypeHandler.java
@@ -19,16 +19,14 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Optional;
-import java.util.Queue;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
-
 import org.springframework.core.MethodParameter;
 import org.springframework.core.ReactiveAdapter;
 import org.springframework.core.ReactiveAdapterRegistry;
@@ -178,10 +176,7 @@ class ReactiveTypeHandler {
 	}
 
 
-	private static abstract class AbstractEmitterSubscriber implements Subscriber<Object> {
-
-		private static final Object COMPLETE_SIGNAL = new Object();
-
+	private static abstract class AbstractEmitterSubscriber implements Subscriber<Object>, Runnable {
 
 		private final ResponseBodyEmitter emitter;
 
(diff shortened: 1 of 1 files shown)


rstoyanchev:
@akarnokd, first of all thanks for taking the time!

The simplifications around signal related fields make sense.

For `onTimeout` it comes from the underlying runtime (Tomcat, Jetty, etc) after a period of inactivity, completely independent thread from the `Publisher`. I did have have a "pretend onComplete" at one point but if we don't handle the timeout in the callback the Servlet container takes action and it's not the desired outcome. [ResponseBodyEmitter](https://github.com/spring-projects/spring-framework/blob/702e01ad52c308a2a3197a589b4b83db01ed86fb/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java#L189) is synchronized so it should be fine to call `terminate()` I think.

For the `send` method crashing [it cleans up](https://github.com/spring-projects/spring-framework/blob/702e01ad52c308a2a3197a589b4b83db01ed86fb/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java#L166-L176) internally so we don't have to.

Any pointers on why the CAS+set may not be enough? Curious to learn more..

Couple more questions on the `run` method. It orders error signal checking ahead of getting the item but complete signal is after. It seems like this can re-order error and item signals if onError arrives just when run starts processing an item signal? Why re-order? Also at the end it calls the `taskExecutor` directly, wouldn't it make sense to call `trySchedule` after decrementing since that has the added try-catch protection around successful task submission?

rstoyanchev:
Another thought on `onTimeout`, since pretty much anything outcome of `run` requires obtaining synchronization on `ResponseBodyEmitter` I could protect the entire `run` with `synchronized(emitter)` and that would properly guard against `onTimeout` completing at the same time without extra cost.

akarnokd:
> For onTimeout it comes from the underlying runtime ...

I'll restore the original direct emission

> For the send method crashing ...

Will restore the original pattern

> Any pointers on why the CAS+set may not be enough? Curious to learn more.

I'm not aware of any studies about this, but all of RxJava 2 and Reactor 3 are written via the getAndIncrement() approach. After thinking more about the code, it is likely correct in this case.

 > It orders error signal checking ahead of getting the item but complete signal is after.

I guess you wanted to delay errors after items. I'll update the code.

> Also at the end it calls the taskExecutor directly ...  added try-catch protection

I forgot about that. I'll update the code for that as well.

> ResponseBodyEmitter .. synchronized(emitter){ ... }

Isn't `ResponseBodyEmitter` already synchronized on its methods?

rstoyanchev:
Thanks! This is now merged. 

> Isn't ResponseBodyEmitter already synchronized on its methods?

Yes, `send` could raise an ISE if the emitter was completed but `run` would catch it and since `subscription.cancel()` should be idempotent, no harm done in calling `terminate()` again.

akarnokd:
https://github.com/spring-projects/spring-framework/commit/43eea41dd6b7260e6ee4c226ece23498e65eafbc#diff-8dc8847c9dbfc64ca96195331492e86eR294

`if (this.terminated) {`

This change is incorrect because the request() before may result in an item and an onComplete signal and this only picks up the complete signal. The correct order is: read terminated, read element, if terminated and element is null -> signal terminated.

PR #1379, opened 2017-04-06
Ensure the subscriber is initialized before switching the states

violetagg:
Fixed NPE thrown by WriteResultPublisher$State$2.publishComplete caused by a thread-safety issue

Issue: SPR-15418


rstoyanchev:
Merged, thanks!

PR #1384, opened 2017-04-10
Improve performance of StringUtils.replace() if pattern is not found

dreis2211:
Fixes https://jira.spring.io/browse/SPR-15430


sbrannen:
Implemented slightly differently in 3a1d431c7df30cffd70174c635879b0d9dd9f745 and ce4eff321cd976df4ba26fe3971642259af891d2.

PR #1389, opened 2017-04-13
remove unnecessary indent

alex-on-java:


lucasljj:
this pr change more than just indent...

alex-on-java:
@lucasljj I meant that we could move code left for better readability. This change doesn't affect logic of code.
But if you could propose better naming for this pr, I will be glad o hear it from you.

snicoll:
@alex-on-java thanks for the PR but it boils down to a stylistic preference and we prefer to keep the code as it is now.

PR #1390, opened 2017-04-14
SPR 15260 & 15419

aclement:
This fixes SPR-15260 and SPR-15419.

SPR-15260 relates to automatically matching trailing separators so you don't need to add paths with  and without them. The behaviour is configurable on the path pattern parser and defaults to ON.
It also better handles multiple separators, we don't silently match multiple separators in the path with a single one in the pattern.

SPR-15419 is related so i've included it in here. I've changed getPathRemaining() to return not only the remaining path but also any variables bound in the part of the path that matched.

I am submitting this as a PR because there are 'other tests' that are failing and I suspect some behaviours need altering to cope with the fact that trailing separators are auto matched now and I don't know 100% where all those places are. Feedback welcome.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/patterns/CaptureTheRestPathElement.java
+++ b/spring-web/src/main/java/org/springframework/web/util/patterns/CaptureTheRestPathElement.java
@@ -49,13 +49,8 @@ class CaptureTheRestPathElement extends PathElement {
 				matchingContext.candidate[candidateIndex] != separator) {
 			return false;
 		}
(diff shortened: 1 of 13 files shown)


aclement:
I removed the SPR-15419 commit and have applied that direct to master, this PR still contains the SPR-15260 change and only that.

aclement:
I am pulling the PR. I know Brian is out but I wanted to get this in so what I've done is integrate the changes but turned OFF the auto trailing separator matching. This makes it behave exactly as before so doesn't disturb any other code.

PR #1400, opened 2017-04-24
Clear the buffer provided by Undertow before reading the request

violetagg:
When reading more that once for a given request, the position/limit of
the buffer provided by Undertow should be reset in order to use the
full capacity of the buffer.


rstoyanchev:
@violetagg same here. We need to merge this still right?

violetagg:
Hi,
This is still relevant.
Regards,
Violeta

rstoyanchev:
Merged, thanks!

PR #1407, opened 2017-04-26
Add more static builder methods to ResponseEntity

ntakouris:
I added a couple more static helper methods that either produce a BodyBuilder or a ResponseEntity with specified body. I find myself typing .build() or 'manually' constructing such response entities frequently, thus I made these changes. Some unit tests were added as well as changed (prefixed with 'NoBody') in order to fit the new changes & be understood better by just reading the method name.

* I've signed the CLA

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ResponseEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/ResponseEntity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


rstoyanchev:
The ResponseEntity static factory methods are for common responses. This PR adds many error codes that should be processed in a cross-cutting fashion and should not appear in many places. If you're producing a 500 error code from many different places I'd say you're probably doing something wrong.

Zarkopafilis:
It's true that some are not such common responses and that 500 errors are
something that needs to be handled a different way sometimes but
nevertheless, wouldn't it be handy to have these extra methods?

rstoyanchev:
It's a but subjective at this point but be it as it may those static methods were designed to provide shortcuts for the most common scenarios and avoiding a long list of every possible option. 

There is no good reason I can see for having static methods for any error code since those should be handled from centralized places and hence the cost of an extra `.build()` doesn't justify seeing them all the time in ResponseEntity where what you do most often is something else.

In any case thanks for preparing a PR. In the future consider creating a ticket first to discuss.

PR #1408, opened 2017-04-28
SPR-15492 - Improve servlet request parameter name handling

HomoEfficio:
Parameters whose names are including square brackets
such as "items[0][id]" and "items[0][tags][]" produce
org.springframework.beans.InvalidPropertyException
when binding to a model object

This commit improves the WebUtils class by adding simple
regular expressions, so those parameters can be bound
to a model object without unnecessary exception

Issue: SPR-15492

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/WebUtils.java
+++ b/spring-web/src/main/java/org/springframework/web/util/WebUtils.java
@@ -579,6 +579,12 @@ public abstract class WebUtils {
 	 * Maps single values to String and multiple values to String array.
 	 * <p>For example, with a prefix of "spring_", "spring_param1" and
 	 * "spring_param2" result in a Map with "param1" and "param2" as keys.
+	 * <p>Parameter names using square brackets such as, "items[0][id]"
+	 * and "items[0]['tags'][]" will be converted to "items[0].id" and
+	 * "items[0].tags" respectively, so those parameters can be bound to
+	 * a model attribute object through
+	 * {@link org.springframework.web.bind.ServletRequestParameterPropertyValues}
+	 * and {@link org.springframework.web.bind.ServletRequestDataBinder}.
 	 * @param request HTTP request in which to look for parameters
 	 * @param prefix the beginning of parameter names
 	 * (if this is null or the empty string, all parameters will match)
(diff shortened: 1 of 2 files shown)


rstoyanchev:
The problem with this solution is that syntax such as `foo[bar]` or even `foo[1][bar]` indicates a map index in JavaBeans conventions with `"bar"` is the map key. So this fix effectively replaces that with a different interpretation. I have not run the tests to see but this will certainly cause regressions.

HomoEfficio:
I don't understand what the first sentence exactly means:

> The problem with this solution is that syntax such as `foo[bar]` or even `foo[1][bar]` indicates a map index in JavaBeans conventions where "bar" is the map key.

But I guess you mean, 
>`foo[bar]` will be ok, but `foo['bar']` or `foo["bar"]` will not be ok. 

Is that right?

HomoEfficio:
Oh, I don't have to close this pull request.
I just amended my commit which resolved the problems above.
Please review it again.

rstoyanchev:
> foo[bar] will be ok, but foo['bar'] or foo["bar"] will not be ok.

No, all of the above will not be ok:

````java
request.addParameter("myMap[foo1]", "bar1");
request.addParameter("myMap['foo2']", "bar2");
request.addParameter("myMap[\"foo3\"]", "bar3");

MyBean target = new MyBean();
WebRequestDataBinder binder = new WebRequestDataBinder(target);
binder.bind(new ServletWebRequest(request));

assertEquals("bar1", target.getMyMap().get("foo1"));
assertEquals("bar2", target.getMyMap().get("foo2"));
assertEquals("bar3", target.getMyMap().get("foo3"));
````

**Comment by HomoEfficio** [comment]
When MVC controller resolves the arguments, `WebRequestDataBinder` is not used.
Instead, `ServletRequestDataBinder` is used.

And `ServletRequestDataBinder.bind()` executes `MutablePropertyValues mpvs = new ServletRequestParameterPropertyValues(request);`, which invokes `WebUtils.getParametersStartingWith()` which returns a new Map.

I added some logic inside `WebUtils.getParametersStartingWith()`.

So the test code of the `WebUtils` should be like below:
```java
Map<String, Object> params = new HashMap<>();
params.put("myMap[foo1]", "bar1");
params.put("myMap['foo2']", "bar2");
params.put("myMap[\"foo3\"]", "bar3")

request.setParameters(params);
Map<String, Object> paramMap = WebUtils.getParametersStartingWith(request, "");

assertEquals("bar1", paramMap.get("myMap.foo1"));
assertEquals("bar2", paramMap.get("myMap.foo2"));
assertEquals("bar3", paramMap.get("myMap.foo3"));
```

This new `paramMap` is used to produce a `MutablePropertyValues` which is eventually bound to a model object by the `DataBinder`.

rstoyanchev:
Sorry but I think you're missing the point. The syntax `foo[bar]` is used to bind to a field named "foo" of type Map which has a key called "bar". You cannot blindly regex that into `foo.bar`.

PR #1412, opened 2017-05-01
Replace lambdas with method references

eddumelendez:


rstoyanchev:
I believe some of this intentional, i.e. lambda used only for functional interfaces.

rstoyanchev:
Also note that multipart support is undergoing review with polish and refactoring throughout. It might be best to withdraw those parts from this PR.

eddumelendez:
@rstoyanchev PR updated

rstoyanchev:
Thanks

rstoyanchev:
It looks like these were already resolved since the PR was raised.

PR #1413, opened 2017-05-02
SPR-15504 - UriComponentsBuilder's fromHttpRequest uses server port as host port when handling the Forwarded header

gvdenbro:
The UriComponentsBuilder's adaptFromForwardedHeaders method doesn't handle the host port of the Forwarded header correctly, which means that the URI generated through the fromHttpRequest method will append the server port to the host (of the Forwarded header), instead of just using the port (if any) defined in the host part of the Forwarded header.
This pull request will try to parse the port of the host of the Forwarded header, and configure the builder accordingly.
According to the RFC https://tools.ietf.org/html/rfc7239#section-5.3 and https://tools.ietf.org/html/rfc7230#section-5.4 the host part of the Forwarded header "provides the host and port information from the target URI", and as such should be handled in the adaptFromForwardedHeaders method.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -717,7 +717,14 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 			String forwardedToUse = StringUtils.tokenizeToStringArray(forwardedHeader, ",")[0];
 			Matcher matcher = FORWARDED_HOST_PATTERN.matcher(forwardedToUse);
 			if (matcher.find()) {
-				host(matcher.group(1).trim());
+				String host = matcher.group(1).trim();
+				if (host.matches(".*:\\d+$")) {
+					host(host.substring(0, host.lastIndexOf(':')));
+					port(host.substring(host.lastIndexOf(':') + 1));
+				} else {
+					host(host);
+					port(null);
+				}
 			}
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@gvdenbro Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1413)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1413) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gvdenbro Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1413)!

snicoll:
@gvdenbro we need a Jira issue for this change. Can you please [create one](https://jira.spring.io/browse/SPR)?

gvdenbro:
@snicoll I've created the following Jira ticket: https://jira.spring.io/browse/SPR-15504

rstoyanchev:
Thanks for the pull request. How about not using another regex here and instead handle it as we do `"X-Forwarded-Host"` just a few lines below?

gvdenbro:
@rstoyanchev I've adapted the code in order to remove the regex and it uses the same style as for the x-forwarded-host. it introduces a bit of code duplication though.

rstoyanchev:
This is now merged with a shared method for the duplicated code. Thanks!

PR #1417, opened 2017-05-07
SPR-15525-update HandlerExecutionChain toString()

QBNemo:
#SPR-15525


jhoeller:
I ended up resolving this differently: through a `getInterceptors()` call in `toString()`, analogous to the handling of the original chain in the copy constructor. Thanks for the PR, in any case!

PR #1419, opened 2017-05-07
Fix sample code for bean definition using functional style on web-flux.adoc

kazuki43zoo:
I've fixed wrong sample code for bean definition using functional style.

* The `registerBean` does not exist  on the `ApplicationContext`
* `AnnotationConfigApplicationContext` instance generated by default constructor is need calling the `refresh` method

Please review this.


rstoyanchev:
Merged, thanks.

PR #1422, opened 2017-05-08
SPR-15505 - Use original query string of forwarded request

svorcmar:
This pull request solves the problem happening when FlashMaps are used in conjunction with forwarded requests with differing query parameters. The AbstractFlashMapManager has used the originating URI but the query string of the forwarded request. That resulted to FlashMap not being matched even when both originating URI and query string matched the FlashMap attributes. The originating query string is now used to match the forwarded request.

This is my first pull request to Spring project so I hope I followed all the guidelines correctly. I've submitted the ICLA. 

Issue: SPR-15505


rstoyanchev:
This is now merged. Thanks for the pull request!

PR #1423, opened 2017-05-09
Remove redundant slashes in URI-string before it's URI will be resolve.

daggerok:
See more about issue here: https://github.com/spring-projects/spring-boot/issues/9133

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpRequest.java
@@ -34,6 +34,8 @@ import org.springframework.util.Assert;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 
+import static java.util.Objects.isNull;
+
 /**
  * Adapt {@link ServerHttpRequest} to the Reactor {@link HttpServerRequest}.
  *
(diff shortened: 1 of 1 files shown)


snicoll:
@daggerok as I've asked you on the issue you've opened in Spring Boot, we need a Jira issue in our tracker for this change. Thanks!

daggerok:
this one is ok? 
https://jira.spring.io/browse/SPR-15529

daggerok:
I reviewed little bit and understand, that this PR wont fix an issue

spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpRequest.java
```diff
@@ -58,10 +60,21 @@ public ReactorServerHttpRequest(HttpServerRequest request, NettyDataBufferFactor
...
 
+	/**
+	 * Remove redundant slashes from request URI-string
+	 *
+	 * <p>
+	 *   http://example.com/// -> http://example.com/
+	 * </p>
+	 */
+	private static String squashSlashes(String uri) {
+		return isNull(uri) ? uri : uri.replaceAll("/{2,}", "/");
```
Shredder121: Wouldn't this collapse `http://example.com///` to `http:/example.com/`?

daggerok: you are talking about URL, but npt URI, with actually i'm replacing, so answering on your question: no, it wouldn't collapse `http://x.y.z///` to `http:/x.y.z/`, instead, it should collapse into `http://x.y.z/`

snicoll:
@daggerok is that PR still relevant?

rstoyanchev:
This evolved from https://jira.spring.io/browse/SPR-15529 to https://jira.spring.io/browse/SPR-15560 and where the actual issue was not the duplicate slash but a bug in how `ReactorServerHttpRequest` created the URI. So this can be closed.

spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpRequest.java
luyuanwan: I think function squashSlashes should in some tool utility class. :)

PR #1425, opened 2017-05-12
Specify the HandlerMapping#order in the websocket example

violetagg:
The order value has to be specified for the HandlerMapping
so that it will be ordered before the one that serves the static resources.


rstoyanchev:
Perhaps a code comment to explain where 10 comes from? In other words > than annotated controller methods (0) but < than static resource (lowest priority - 1).

PR #1427, opened 2017-05-14
SPR-15543: Fix duplicate subscriptionId for destination-sessionId key

baot:
DefaultSubscriptionRegistry's destinationCache accessCache contains duplicate subscriptionId for destination - sessionId key if findSubscriptions is called after addSubscription and before updateAfterNewSubscription.

The change includes checking for existing subId in updateAfterNewSubscription method before adding subId value to sessionId key. (Set implementation to avoid duplication can be considered or other better solutions are appreciated).

Issue: SPR-15543

I have submitted CLA.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
@@ -292,8 +292,10 @@ public class DefaultSubscriptionRegistry extends AbstractSubscriptionRegistry {
 					String cachedDestination = entry.getKey();
 					if (getPathMatcher().match(destination, cachedDestination)) {
 						LinkedMultiValueMap<String, String> subs = entry.getValue();
-						subs.add(sessionId, subsId);
-						this.accessCache.put(cachedDestination, subs.deepCopy());
+						if (!subs.containsKey(sessionId) || !subs.get(sessionId).contains(subsId)) {
+							subs.add(sessionId, subsId);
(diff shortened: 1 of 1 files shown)


rstoyanchev:
This is now merged.

steviemo:
Is there not possible NPE if subs.get(sessionId) returns null? How about:

final LinkedMultiValueMap<String, String> subs = entry.getValue();
final List<String> subscriptionsForSession = subs.get(sessionId);
if (null == subscriptionsForSession || !subscriptionsForSession.contains(subsId)) {
    subs.add(sessionId, subsId);
}
this.accessCache.put(cachedDestination, subs.deepCopy());

spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
```diff
@@ -292,8 +292,10 @@ public void updateAfterNewSubscription(String destination, String sessionId, Str
 					String cachedDestination = entry.getKey();
 					if (getPathMatcher().match(destination, cachedDestination)) {
 						LinkedMultiValueMap<String, String> subs = entry.getValue();
-						subs.add(sessionId, subsId);
-						this.accessCache.put(cachedDestination, subs.deepCopy());
+						if (!subs.containsKey(sessionId) || !subs.get(sessionId).contains(subsId)) {
```
luyuanwan: 在这里加了一个判断，表示没看懂为什么要加这个判断

rstoyanchev: Indeed, thanks for catching this. Fixed!

PR #1431, opened 2017-05-23
add value predicate support in BeanUtils.copyProperties()

khannedy:
In my project, I have some ugly code to merge 2 properties object like this 

```java
Configuration global = ...
Configuration config = ...

Configuration mergeConfig = new Configuration()
BeanUtils.copyProperties(mergeConfig, global);

if(config.getFoo() != null){
   mergeConfig.setFoo(config.getFoo());
}

if(config.getBar() != null){
   mergeConfig.setBar(config.getBar());
}

if(config.getXxx() != null){
   mergeConfig.setXxx(config.getXxx());
}
```

My code only make sure that I only override non null values. This is very ugly because I need to check null in all of attributes.

This pull request is to add Predicate support for BeanUtils.copyProperties, so I can merge properties easily.

```java

Configuration global= ...
Configuration config = ...

Configuration mergeConfig = new Configuration();
BeanProperties.copyProperties(mergeConfig, global);
BeanProperties.copyProperties(mergeConfig, config, Object::nonNull)

```

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -30,6 +30,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.function.Predicate;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


snicoll:
@khannedy that sounds interesting. Could you please create a Jira issue for that change?

khannedy:
@snicoll  this is the jira https://jira.spring.io/browse/SPR-15581

snicoll:
Thanks for the PR and sorry it took so long but given that we don't have a use case for this ourselves and `BeanUtils` is mainly for internal use within the framework, I am going to close this.

PR #1432, opened 2017-05-24
Polish RequestMappingHandlerMapping.java

QBNemo:
type-level RequestMapping annotation is functionaly equal to Controller annotation


rstoyanchev:
Merged, thanks.

PR #1435, opened 2017-05-26
Timeout Overrides in Messaging Templates

garyrussell:
Issue: SPR-15591

Provide a mechanism to override the configured send and receive timeouts in the
GenericMessagingTemplate.

- overload `doReceive()` to take a receive timeout argument
- for `sendAndReceive()` methods examine message headers for these timeout values
  - remove headers to avoid propagation
  - avoid the unconditional rebuild of the reply message if the original headers weren't present
- also remove headers from simple `send()` operations
- change javadocs for the setters to indicate they are now defaults
- add properties to allow the user to override the header names used
- change `TemporaryReplyChannel` to use the `send` argument and change it to static
  - add package-protected ctor to avoid the compiler creating a synthetic ctor for access

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/core/GenericMessagingTemplate.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/core/GenericMessagingTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


artembilan:
> JUnit tests should include assert() or fail()

Looks like `ArgumentCaptor` and `assert...` afterwards should be used instead of asserts in the `Answer`

artembilan:
> JUnit tests should include assert() or fail()

Looks like `ArgumentCaptor` and `assert...` afterwards should be used instead of asserts in the `Answer`

spring-messaging/src/main/java/org/springframework/messaging/core/GenericMessagingTemplate.java
```diff
@@ -82,6 +96,47 @@ public long getReceiveTimeout() {
...
+		Assert.notNull(sendTimeoutHeader, "'sendTimeoutHeader' cannot be null");
+		this.sendTimeoutHeader = sendTimeoutHeader;
+	}
+
+	/**
+	 * @return the configured sendTimeoutHeader.
+	 * @since 5.0
+	 */
+	public String getSendTimeoutHeader() {
+		return sendTimeoutHeader;
```
artembilan: ` this.` ?

spring-messaging/src/main/java/org/springframework/messaging/core/GenericMessagingTemplate.java
```diff
@@ -82,6 +96,47 @@ public long getReceiveTimeout() {
...
+		Assert.notNull(receiveTimeoutHeader, "'receiveTimeoutHeader' cannot be null");
+		this.receiveTimeoutHeader = receiveTimeoutHeader;
+	}
+
+	/**
+	 * @return the configured receiveTimeoutHeader
+	 * @since 5.0
+	 */
+	public String getReceiveTimeoutHeader() {
+		return receiveTimeoutHeader;
```
artembilan: ` this.`, too

artembilan (changes requested):
Just a couple concerns from me.

rstoyanchev:
This is merged into master as a single commit. Thanks Gary.

spring-messaging/src/main/java/org/springframework/messaging/core/GenericMessagingTemplate.java
```diff
@@ -142,20 +212,25 @@ protected final void doSend(MessageChannel channel, Message<?> message) {
...
 		}
 		catch (RuntimeException ex) {
 			tempReplyChannel.setSendFailed(true);
 			throw ex;
 		}
 
-		Message<?> replyMessage = this.doReceive(tempReplyChannel);
-		if (replyMessage != null) {
+		Message<?> replyMessage = this.doReceive(tempReplyChannel, receiveTimeout);
+		if (replyMessage != null && (originalReplyChannelHeader!= null || originalErrorChannelHeader != null)) {
```
garyrussell: @rstoyanchev It turns out this change was incorrect; we need to null-out the channel headers if they were originally null.

Do you want me to submit another PR, or would you mind just reverting this line?

Thanks (and my apologies).

rstoyanchev: If you could push another commit on the same branch, so I can see the exact changes you have in mind.

garyrussell: Sorry - I just issued a new PR on that branch (having first reset it to master).

I issued another PR: https://github.com/spring-projects/spring-framework/pull/1440

Thanks

PR #1438, opened 2017-05-29
[javadoc]: fixed duplicate words from javadoc comments

ameyjadiye:


snicoll:
Thanks for the PR but your copy of master must be outdated as those 3 typos have already been fixed.

PR #1440, opened 2017-05-31
Revert Incorrect MessagingTemplate Change

garyrussell:
Issue: SPR-15991

The change to "optimize" the template by not rebuilding the reply message
when the original header channels was null was incorrect.

We need to null out those headers if they were originally null.


rstoyanchev:
The commit is in master.

PR #1441, opened 2017-05-31
Doc update

Buzzardo:
Changed a note to a warning and added more detail.

### DIFF
--- a/src/asciidoc/core-beans.adoc
+++ b/src/asciidoc/core-beans.adoc
@@ -4562,14 +4562,16 @@ automatically resolved, with no special setup necessary.
 	}
 ----
 
-[NOTE]
+[WARNING]
 ====
-`@Autowired`, `@Inject`, `@Resource`, and `@Value` annotations are handled by Spring
+`@Autowired`, `@Inject`, `@Resource`, and `@Value`  annotations are handled by Spring
 `BeanPostProcessor` implementations which in turn means that you __cannot__ apply these
 annotations within your own `BeanPostProcessor` or `BeanFactoryPostProcessor` types (if
 any). These types must be 'wired up' explicitly via XML or using a Spring `@Bean` method.
-====
 
+Also, do not use `@Autowired`, `@Inject`, and `@Resource` annotations within a BeanFactory class nor any of its dependencies. Doing so can result in a `BeanCurrentlyInCreationException`. The `@Value` annotation may also cause the problem, d
+epending on whether it causes a bean to be instantiated.
+====
 
 [[beans-autowired-annotation-primary]]
 === Fine-tuning annotation-based autowiring with @Primary


src/asciidoc/core-beans.adoc
```diff
@@ -4562,14 +4562,16 @@ automatically resolved, with no special setup necessary.
 	}
 ----
 
-[NOTE]
+[WARNING]
 ====
-`@Autowired`, `@Inject`, `@Resource`, and `@Value` annotations are handled by Spring
+`@Autowired`, `@Inject`, `@Resource`, and `@Value`  annotations are handled by Spring
```
snicoll: That extra space isn't necessary, right?

src/asciidoc/core-beans.adoc
```diff
@@ -4562,14 +4562,16 @@ automatically resolved, with no special setup necessary.
...
 `BeanPostProcessor` implementations which in turn means that you __cannot__ apply these
 annotations within your own `BeanPostProcessor` or `BeanFactoryPostProcessor` types (if
 any). These types must be 'wired up' explicitly via XML or using a Spring `@Bean` method.
-====
 
+Also, do not use `@Autowired`, `@Inject`, and `@Resource` annotations within a BeanFactory class nor any of its dependencies. Doing so can result in a `BeanCurrentlyInCreationException`. The `@Value` annotation may also cause the problem, d
+epending on whether it causes a bean to be instantiated.
+====
```
snicoll: that looks like a complete different conversation to me. What do you mean by "within a `BeanFactory`". I wouldn't expect anyone to use those annotations there (and that's not very typical in use code anyway).

The `@Value` is yet another thing.

snicoll (changes requested):
Thanks for the PR! I've added a couple of comments.

Rather than documenting the result of your investigation, can you please describe the problem you faced and how that change in the doc would have helped you?

src/asciidoc/core-beans.adoc
Buzzardo: I thought about making it into a separate warning. That would have resulted in two adjacent warnings that addressed the same set of annotations, so I combined them. I can split them, though, if that works better.

I didn't have the problem. Someone else did. Then the topic came up on Slack, where various folks mentioned that they had tried this and had trouble. I thought it rated a documentation update.

Several people have tried to use those annotations within a BeanFactory (or the dependencies of a BeanFactory) and have had trouble as a result. You can see the evidence of it in the following issues:
https://jira.spring.io/browse/SPR-9255
https://jira.spring.io/browse/SPR-8704
https://jira.spring.io/browse/SPR-10349

If you can tell me additional details around using @Value in this context, please do. I'll be happy to add them to the document.

src/asciidoc/core-beans.adoc
Buzzardo: Right.

src/asciidoc/core-beans.adoc
rstoyanchev: Good find digging up those tickets. 

One of tickets SPR-8704 shows that it was turned into a documentation task and documentation updates were applied on this topic already including both master https://github.com/spring-projects/spring-framework/commit/fd41f63ec065b890a539451532392fa9b5bb0b28 and 4.3 https://github.com/spring-projects/spring-framework/commit/28e114b569d0e3e19ea0e0160704b6726810f428.

I think the explanation here is not very good because explaining things in terms of the exception doesn't help much and made me go lookup the javadoc for that exception which didn't bring a more satisfying answer either:
> Doing so can result in a `BeanCurrentlyInCreationException`

I find this explanation under the ticket much more useful because it tells me why I shouldn't do it (interface vs annotation based dependency injection):
>  we tend to generally recommend against using annotation-driven injection in FactoryBean implementations. Since such classes implement an interface-based contract to begin with, I would rather design them to use programmatic BeanFactory lookups for the target beans needed.

@jhoeller from the above linked tickets it seems that quite a few people are running into this and documentation alone is not effective in preventing time wasted debugging. If there is anything that could be done to detect or proactively reject..

Buzzardo: I find the explanation at the bottom of https://github.com/spring-projects/spring-framework/commit/28e114b569d0e3e19ea0e0160704b6726810f428 to be insufficient. It should be a warning. It doesn't mention the consequences: getting an exception.

It's a comment in code, and I'm working on the reference guide. Also, it’s about FactoryBean, and I’m dealing with BeanFactory.  If we need to call out the connection between the two to address the issue of these annotations causing a synchronization problem that sometimes raises an exception, please let me know.

As for this particular content, suppose we borrow a bit of content from that code comment and make a warning that reads as follows: 

[WARNING]
Adding any of the `@Autowired`, `@Inject`, or `@Resource` annotations to a BeanFactory object or any of its dependencies can result in an `BeanCurrentlyInCreationException`, as the injections performed by those annotations may happen before the BeanFactory object has been completed.

I still need to know whether to include @Value in the list of annotations that can cause the problem.

Thanks.

gregturn: If FactoryBean and BeanFactory are in the same text, perhaps an admonition like "Don't confuse FactoryBean, a factory-method built bean with BeanFactory, an internal blah blah for creating beans."

rstoyanchev:
This seems outdated.

PR #1442, opened 2017-05-31
SPR-15604 Fixed cron expression validation method in CronSequenceGenerator

Enigo:
This method validated only fields length of the given expression, not the validity of cron expression. It seems logical that the method with name 'isValidExpression' would validate cron as well.
Plus, possible NPE fixed.


snicoll:
@Enigo please create a Jira issue for this change.

Enigo:
@snicoll, hey, [issue](https://jira.spring.io/browse/GREENHOUSE-658) raised, but I accidentally miss-clicked the project :( Is it possible to move the issue to the Spring Framework (SPR)?

snicoll:
nope, I have no right on that project, sorry.

Enigo:
Great, thanks! [Issue](https://jira.spring.io/browse/SPR-15604) raised)

spring-context/src/main/java/org/springframework/scheduling/support/CronSequenceGenerator.java
```diff
@@ -260,6 +260,9 @@ private void reset(Calendar calendar, List<Integer> fields) {
 	 * Parse the given pattern expression.
 	 */
 	private void parse(String expression) throws IllegalArgumentException {
+		if (expression == null) {
```
deodeveloper: Using standard way of spring exceptions being thrown. 
How about ```Assert.isNull(expression, "Cron expression cannot be null!")``` ?

Enigo: I'm sure you meant `Assert.notNull` :) fixed.

jhoeller:
Implemented in a slightly different fashion, as mentioned on the JIRA issue.

PR #1443, opened 2017-06-02
Support date properties in Content-Disposition HTTP header

sdeleuze:
As discussed with @rstoyanchev, before merging this PR we should choose a strategy regarding to date/time that we previously exposed with `long` like in `long HttpHeaders.getAccessControlMaxAge()`.

Should we continue to use `long` for consistency?
Should we add `Instant` based variants and deprecate `long` based ones?

Pros and cons should be discussed and the solution should be applied globally to Spring Framework API like we did for `Optional`.

cc @jhoeller 

Issue: SPR-15555


sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/97909f225819b2297c1036a8a9108c0582edcf90.

PR #1448, opened 2017-06-04
add ordered support in interceptor registration

khannedy:
Currently there is no guarantee for ordering of `HandlerInterceptor` to be invoked. So if we want to create `HandlerInterceptor` that invoked before or after third party library's `HandlerInterceptor`, we can not guarantee that.

The idea of this feature is add order attribute in `InterceptorRegistration`, and when `InterceptorRegistry.getInterceptors()` being invoked, it will sorted first, so we can customize  

```java
registry.addInterceptor(interceptor1).order(Ordered.LOWEST_PRECEDENCE);
registry.addInterceptor(interceptor2).order(Ordered.HIGHEST_PRECEDENCE);
```

This is the JIRA
https://jira.spring.io/browse/SPR-15620


rstoyanchev:
This is now merged, thanks! Note that I removed the use of `Ordered` since the registration is meant for internal use and shouldn't be public.

khannedy:
cool, thanks @rstoyanchev

PR #1449, opened 2017-06-05
Use Map.forEach instead of Map.entrySet().forEach

diguage:


diguage:
I create a issue at JIRA : [Use Map.forEach instead of Map.entrySet().forEach](https://jira.spring.io/browse/SPR-15627)

PR #1450, opened 2017-06-05
Use the diamond syntax

diguage:


diguage:
I find some test class file do not use the diamond syntax. 

May I submit the change of the test class files?

snicoll (approved):

PR #1451, opened 2017-06-05
Refact iterator of Map with Java 8 forEach

diguage:


diguage:
I create a issue at JIRA: [[SPR-15631] Refact iterator of Map with Java8's Map.forEach - Spring JIRA](https://jira.spring.io/browse/SPR-15631)

snicoll (approved):

PR #1453, opened 2017-06-07
use lambda to refact the code

diguage:


diguage:
It has some error. Close and resubmit.

PR #1455, opened 2017-06-09
add EditorConfig file, help to unified code style

diguage:


diguage:
I create an issue at JIRA: [[[SPR-15644] add EditorConfig file, help to unified code style - Spring JIRA]](https://jira.spring.io/browse/SPR-15644)

.editorconfig
```diff
@@ -0,0 +1,31 @@
...
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+max_line_length = 120
```
snicoll: I think this one is more of a recommendation than a hard requirement and enforcing it will lead to unpleasant results.

.editorconfig
```diff
@@ -0,0 +1,31 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
```
snicoll: In Spring Boot, [we've restrict ourselves to XML and Java](https://github.com/spring-projects/spring-boot/blob/master/.editorconfig). We could add groovy and kt files here but I am not keen to add more, at least for a first revision.

snicoll (commented):
Thanks for the PR, I've added a couple of comments.

.editorconfig
diguage: Ok. First, I restrict to XML and Java.

.editorconfig
diguage: `end_of_line = lf` and `max_line_length = 120` are found at [Code Style](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style). But I also find some lines are more than 120 characters. 

The other two are found in codes.

How do I change it as follows:

```
end_of_line = lf
insert_final_newline = true
trim_trailing_whitespace = true
```

diguage:
I looked at the EditorConfig at Spring Boot. I will change the config as the file in SpringBoot.

snicoll (approved):

PR #1457, opened 2017-06-10
add @FunctionalInterface annotation

diguage:


spring-aop/src/main/java/org/springframework/aop/TargetClassAware.java
```diff
@@ -29,6 +29,7 @@
  * @since 2.0.3
  * @see org.springframework.aop.support.AopUtils#getTargetClass(Object)
  */
+@FunctionalInterface
```
snicoll: I don't think this is required really. Just because an interface has a single method doesn't mean it's a functional interface. These callbacks are not meant to be used with lambdas so I don't think this should be done.

spring-beans/src/main/java/org/springframework/beans/factory/BeanNameAware.java
```diff
@@ -33,6 +33,7 @@
  * @see BeanFactoryAware
  * @see InitializingBean
  */
+@FunctionalInterface
```
snicoll: Ditto

spring-aop/src/main/java/org/springframework/aop/framework/AdvisorChainFactory.java
```diff
@@ -27,6 +27,7 @@
  * @author Rod Johnson
  * @author Juergen Hoeller
  */
+@FunctionalInterface
```
snicoll: I don't know, perhaps @jhoeller has an opinion?

snicoll (changes requested):

spring-aop/src/main/java/org/springframework/aop/TargetClassAware.java
diguage: I am also thinking about this problem. It meets the requirements, but it will not be used with labmda. Should it be added the `@FunctionalInterface` annotation?

spring-aop/src/main/java/org/springframework/aop/framework/AdvisorChainFactory.java
jhoeller: In my initial pass adding `@FunctionalInterface` a while ago, I did indeed skip interfaces which just happen to declare a single method for other reasons. From my perspective, a single abstract method is a required but not sufficient condition for calling an interface "functional".

snicoll:
Thanks for the PR but those 3 interfaces are fine as they are now, see comments for more details.

PR #1458, opened 2017-06-12
Introduce LocaleContextResolver in WebFlux

sdeleuze:
This commit introduces LocaleContextResolver interface, which is used
at ServerWebExchange level to resolve Locale, TimeZone and other i18n
related informations.

It follows Spring MVC locale resolution patterns with a few differences:
 - Only LocaleContextResolver is supported since LocaleResolver is less
   flexible
 - Support is implemented in the org.springframework.web.server
   package of spring-web module rather than in spring-webflux in order
   to be able to leverage it at ServerWebExchange level
 - The LocaleContextResolver to use is an optional DefaultServerWebExchange
   constructor parameter rather than a bean with well known name

2 implementations are provided:
 - FixedLocaleContextResolver
 - AcceptHeaderLocaleContextResolver

It can be configured with both functional or annotation-based APIs.

Issue: SPR-15036

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/i18n/SimpleTimeZoneAwareLocaleContext.java
+++ b/spring-context/src/main/java/org/springframework/context/i18n/SimpleTimeZoneAwareLocaleContext.java
@@ -52,6 +52,7 @@ public class SimpleTimeZoneAwareLocaleContext extends SimpleLocaleContext implem
 	}
 
 
+	@Override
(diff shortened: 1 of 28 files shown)


sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/e0e6736bc58101af13a352d4aa2a6c2dc0994d72.

PR #1463, opened 2017-06-23
Add onError callback to DeferredResult

violetagg:
Issue: SPR-15614

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitterReturnValueHandler.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitterReturnValueHandler.java
@@ -19,6 +19,8 @@ package org.springframework.web.servlet.mvc.method.annotation;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.List;
+import java.util.function.Consumer;
+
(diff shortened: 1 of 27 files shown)


spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitterReturnValueHandler.java
```diff
@@ -209,7 +211,7 @@ public void complete() {
 
 		@Override
 		public void completeWithError(Throwable failure) {
-			this.deferredResult.setErrorResult(failure);
+			// no-op
 		}
```
rstoyanchev: Couldn't we leave this as it was before? It is true that the PR removed the internal use of the method within the try-catch of `ResponseBodyEmitter#sendInternal` and that any errors from the container side should work as expected but a controller could also call `completeWithError` and the change eliminates that option (and would cause a regression).

If a container and an application thread both try to `setErrorResult` I suppose only one could win?

rstoyanchev (commented):
Everything looks good, thanks for the extensive PR! I'm only wondering about that one change to "no-op" calls to `ResponseBodyEmitter#completeWithError`.

violetagg:
I agree. I updated the change.

rstoyanchev:
Merged, thanks!

PR #1468, opened 2017-07-04
Adapt to the changes in Reactor Netty ClientOptions

violetagg:
ClientOptions/HttpClientOptions are now created via Builder


bclozel:
Already merged with fd1859c
Thanks!

PR #1472, opened 2017-07-10
Make getters and setters null-safety consistent

sdeleuze:
This commit ensures that null-safety is consistent between
getters and setters in order to be able to provide beans
with properties with a common type when type safety is
taken in account like with Kotlin.

It also adds a few missing property-level @Nullable
annotations.

Issue: SPR-15540

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/ApplicationEventPublisherAware.java
+++ b/spring-context/src/main/java/org/springframework/context/ApplicationEventPublisherAware.java
@@ -17,6 +17,7 @@
 package org.springframework.context;
 
 import org.springframework.beans.factory.Aware;
+import org.springframework.lang.Nullable;
 
 /**
  * Interface to be implemented by any object that wishes to be notified
@@ -37,6 +38,6 @@ public interface ApplicationEventPublisherAware extends Aware {
 	 * Invoked before ApplicationContextAware's setApplicationContext.
 	 * @param applicationEventPublisher event publisher to be used by this object
 	 */
-	void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher);
+	void setApplicationEventPublisher(@Nullable ApplicationEventPublisher applicationEventPublisher);
 
 }
--- a/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
+++ b/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
@@ -262,9 +262,8 @@ public abstract class AbstractApplicationContext extends DefaultResourceLoader
 	 * Typically done during initialization of concrete context implementations.
 	 * <p>Default is the object id of the context instance.
 	 */
-	public void setDisplayName(String displayName) {
-		Assert.hasLength(displayName, "Display name must not be empty");
-		this.displayName = displayName;
+	public void setDisplayName(@Nullable String displayName) {
+		this.displayName = (displayName != null ? displayName : ObjectUtils.identityToString(this));
 	}
 
 	/**
--- a/spring-context/src/main/java/org/springframework/scheduling/config/ScheduledTaskRegistrar.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/config/ScheduledTaskRegistrar.java
@@ -92,9 +92,8 @@ public class ScheduledTaskRegistrar implements InitializingBean, DisposableBean 
 	 * {@link java.util.concurrent.ScheduledExecutorService} to be wrapped as a
 	 * {@code TaskScheduler}.
 	 */
-	public void setScheduler(Object scheduler) {
-		Assert.notNull(scheduler, "Scheduler object must not be null");
-		if (scheduler instanceof TaskScheduler) {
+	public void setScheduler(@Nullable Object scheduler) {
+		if ((scheduler instanceof TaskScheduler) || (scheduler == null)) {
 			this.taskScheduler = (TaskScheduler) scheduler;
 		}
 		else if (scheduler instanceof ScheduledExecutorService) {
--- a/spring-context/src/main/java/org/springframework/validation/DataBinder.java
+++ b/spring-context/src/main/java/org/springframework/validation/DataBinder.java
@@ -424,7 +424,7 @@ public class DataBinder implements PropertyEditorRegistry, TypeConverter {
 	 * @see #setDisallowedFields
 	 * @see #isAllowed(String)
 	 */
-	public void setAllowedFields(String... allowedFields) {
+	public void setAllowedFields(@Nullable String... allowedFields) {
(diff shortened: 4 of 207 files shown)


spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
```diff
@@ -262,9 +262,8 @@ public String getApplicationName() {
 	 * Typically done during initialization of concrete context implementations.
 	 * <p>Default is the object id of the context instance.
 	 */
-	public void setDisplayName(String displayName) {
-		Assert.hasLength(displayName, "Display name must not be empty");
-		this.displayName = displayName;
+	public void setDisplayName(@Nullable String displayName) {
```
jhoeller: `getDisplayName()` isn't actually nullable, so I guess we can leave this setter non-null as well?

spring-context/src/main/java/org/springframework/context/ApplicationEventPublisherAware.java
```diff
@@ -37,6 +38,6 @@
 	 * Invoked before ApplicationContextAware's setApplicationContext.
 	 * @param applicationEventPublisher event publisher to be used by this object
 	 */
-	void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher);
+	void setApplicationEventPublisher(@Nullable ApplicationEventPublisher applicationEventPublisher);
```
jhoeller: Let's leave the `Aware` interfaces as non-null since the container will never call them with a non-null reference. For this particular case, we could keep the `AbstractBrokerMessageHandler.setApplicationEventPublisher` implementation as nullable but leave the interface untouched.

spring-context/src/main/java/org/springframework/validation/DataBinder.java
```diff
@@ -546,9 +550,11 @@ private void assertValidators(Validator... validators) {
 	 * @see #setValidator(Validator)
 	 * @see #replaceValidators(Validator...)
 	 */
-	public void addValidators(Validator... validators) {
+	public void addValidators(@Nullable Validator... validators) {
```
jhoeller: Here it seems sufficient to mark `setValidator` as nullable, paired with `getValidator()`. The multi-validator case in `getValidators()` isn't nullable anyway, so `addValidators` doesn't have to be nullable either (aside from not being setter and differing in array vs list as well).

spring-context/src/main/java/org/springframework/scheduling/config/ScheduledTaskRegistrar.java
```diff
@@ -92,9 +92,8 @@ public void setTaskScheduler(TaskScheduler taskScheduler) {
 	 * {@link java.util.concurrent.ScheduledExecutorService} to be wrapped as a
 	 * {@code TaskScheduler}.
 	 */
-	public void setScheduler(Object scheduler) {
-		Assert.notNull(scheduler, "Scheduler object must not be null");
-		if (scheduler instanceof TaskScheduler) {
+	public void setScheduler(@Nullable Object scheduler) {
+		if ((scheduler instanceof TaskScheduler) || (scheduler == null)) {
```
jhoeller: We could code this as a dedicated `if` clause: `if (scheduler == null) { this.scheduler = null } else if (scheduler instanceof TaskScheduler) ...`

spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceTransactionManager.java
```diff
@@ -155,8 +155,8 @@ public DataSourceTransactionManager(DataSource dataSource) {
 	 * @see TransactionAwareDataSourceProxy
 	 * @see org.springframework.transaction.jta.JtaTransactionManager
 	 */
-	public void setDataSource(DataSource dataSource) {
-		if (dataSource instanceof TransactionAwareDataSourceProxy) {
+	public void setDataSource(@Nullable DataSource dataSource) {
+		if (dataSource != null && dataSource instanceof TransactionAwareDataSourceProxy) {
```
jhoeller: `instanceof` works fine with `null`, so the extra `!= null` check upfront isn't actually necessary.

spring-web/src/main/java/org/springframework/http/client/HttpComponentsAsyncClientHttpRequestFactory.java
```diff
@@ -117,8 +117,7 @@ public HttpComponentsAsyncClientHttpRequestFactory(
 	 * @since 4.3.10
 	 * @see #setHttpClient(HttpClient)
 	 */
-	public void setAsyncClient(HttpAsyncClient asyncClient) {
-		Assert.notNull(asyncClient, "HttpAsyncClient must not be null");
+	public void setAsyncClient(@Nullable HttpAsyncClient asyncClient) {
```
jhoeller: The field should change to not-null along with direct field population in the constructors here. I'll commit this in a bit.

spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.java
```diff
@@ -90,8 +90,7 @@ public HttpComponentsClientHttpRequestFactory(HttpClient httpClient) {
 	 * Set the {@code HttpClient} used for
 	 * {@linkplain #createRequest(URI, HttpMethod) synchronous execution}.
 	 */
-	public void setHttpClient(HttpClient httpClient) {
-		Assert.notNull(httpClient, "HttpClient must not be null");
+	public void setHttpClient(@Nullable HttpClient httpClient) {
```
jhoeller: The field should change to not-null along with direct field population in the constructors here. I'll commit this in a bit.

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerAdapter.java
```diff
@@ -524,8 +524,11 @@ public void setParameterNameDiscoverer(ParameterNameDiscoverer parameterNameDisc
 	 * in method argument default values.
 	 */
 	@Override
-	public void setBeanFactory(BeanFactory beanFactory) {
-		if (beanFactory instanceof ConfigurableBeanFactory) {
+	public void setBeanFactory(@Nullable BeanFactory beanFactory) {
```
jhoeller: Since `getBeanFactory` is protected and returns a narrowed type anyway, the setter's nullability doesn't have to be aligned here, so I'd rather keep the straightforward check.

sdeleuze:
Feedback taken in account in https://github.com/spring-projects/spring-framework/commit/fb4ddb0746a3f194091043de328875596736427a.

PR #1473, opened 2017-07-11
SPR-15755: Introduce Kotlin functional bean registration DSL

sdeleuze:
As a follow-up of the [ApplicationContext Kotlin extensions](https://spring.io/blog/2017/01/04/introducing-kotlin-support-in-spring-framework-5-0#functional-bean-registration-with-kotlin), close to
the [Kotlin functional WebFlux DSL](https://spring.io/blog/2017/01/04/introducing-kotlin-support-in-spring-framework-5-0#spring-web-functional-api-the-kotlin-way) and similar to the [Groovy bean
configuration DSL](https://spring.io/blog/2014/03/03/groovy-bean-configuration-in-spring-framework-4) (but very different in term of implementation since
no internal XML-based application context is involved), this commit is
about introducing a lightweight Kotlin DSL for functional bean
registration.

It allows declaring beans as following:

```kotlin
beans {
	bean<Foo>()
	profile("bar") {
		bean<Bar>("bar")
	}
	environment({it.activeProfiles.contains("baz")}) {
		bean { Baz(it.ref()) }
		bean { Baz(it.ref("bar")) }
	}
}

```

Advantages compared to Regular `ApplicationContext` API are:
 - No exposure of low-level `ApplicationContext` API
 - Focused DSL easier to read, but also easier to write with a fewer
   entries in the auto-complete
 - Declarative syntax instead of functions with verbs like `registerBeans`
   while still allowing programmatic registration of beans with if
   expressions or for loops
 - Such DSL is idiomatic in Kotlin
 - No need to have an `ApplicationContext` instance to write how you
   register your beans since `beans { }` DSL is conceptually a
   `Consumer<ApplicationContext>`

This DSL effectively replaces `ApplicationContext` Kotlin extensions as
the recommended way to register beans in a functional way with Kotlin.

Issue: [SPR-15755](https://jira.spring.io/browse/SPR-15755)

### DIFF
--- a/spring-context/src/main/kotlin/org/springframework/context/BeanRegistrationDsl.kt
+++ b/spring-context/src/main/kotlin/org/springframework/context/BeanRegistrationDsl.kt
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2002-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


Tradunsky:
:+1:

sdeleuze:
Merged.

PR #1477, opened 2017-07-16
SPR-7582 - Rebuild Advisor chain when interceptor list is updated in ProxyFactoryBean

marijnm:
When getObject() on ProxyFactoryBean is called it might not be
fully initialized yet, leading to an exception. This is fine, but
the empty Advisor chain is never rebuilt, not even after initialization
has finished. This leads to 'no-op' proxies which don't do anything.

Issue: SPR-7582

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
@@ -391,6 +391,11 @@ public class AdvisedSupport extends ProxyConfig implements Advised {
 		return this.advisors;
 	}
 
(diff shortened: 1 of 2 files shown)


iarroyo:
+1

adperezmorales:
I faced the same problem while using constructor injection of proxies in proxies. I sorted it out using targetName in ProxyFactoryBean instead of target referencing the bean, because changing bean order definition is not always possible. This fix can be helpful in many situations

jhoeller:
On review, we're going with a simpler change (just covering late initialization, not re-initialization) for the original report #12238. Thanks for the pull request, in any case, and sorry that it took so long...

PR #1481, opened 2017-07-19
MimeType should use last `+` to delimit suffix

drautb:
According to the [RFC][1], multiple `+` symbols are allowed in a media type. The _last_ one should always be used to delimit the structured syntax suffix.  

The `MimeType` class was using the first `+` rather than the last, so media types with multiple `+` symbols failed to be included in the correct type matchers.  

Issue: [SPR-15795][2] 
CLA: Signed.  

[1]: https://trac.tools.ietf.org/html/draft-ietf-appsawg-media-type-regs-14#section-4.2 
[2]: https://jira.spring.io/browse/SPR-15795


drautb:
https://github.com/spring-projects/spring-framework/commit/d7a7b08b0845da9a0ee6ebae2db2f853e77c5214.

PR #1482, opened 2017-07-19
Support instantiating Kotlin classes with optional parameters

sdeleuze:
This commit updates BeanUtils class in order to add Kotlin optional
parameters with default values support to the immutable data classes
support introduced by SPR-15199.

@jhoeller Could you please review and hopefully merge it with your changes if any?

Issue: SPR-15673

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -18,6 +18,7 @@ package org.springframework.beans;
 
 import java.beans.PropertyDescriptor;
 import java.beans.PropertyEditor;
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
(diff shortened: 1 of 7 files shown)


spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
```diff
@@ -102,12 +113,11 @@
 		if (clazz.isInterface()) {
 			throw new BeanInstantiationException(clazz, "Specified class is an interface");
 		}
-		try {
-			return instantiateClass(clazz.getDeclaredConstructor());
-		}
-		catch (NoSuchMethodException ex) {
-			throw new BeanInstantiationException(clazz, "No default constructor found", ex);
+		Constructor<T> ctor = findPrimaryConstructor(clazz);
```
jhoeller: I guess we'll have to differentiate here: For `instantiateClass(Class)`, we can only deal with the no-arg default constructor (public or non-public) in the regular Java case... not with a non-default unique public constructor that `findPrimaryConstructor` might return to us.

So maybe we'll do some straight `KotlinDelegate.findPrimaryConstructor` vs `clazz.getDeclaredConstructor()` right here in `instantiateClass(Class)`, and only use the public `findPrimaryConstructor` variant for data binding purposes.

sdeleuze: Indeed I am going to change that accordingly to your proposal.

sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/fa4d13968451bfe96eab122634c0a1eab6083aaa.

PR #1485, opened 2017-07-22
meaningful constant Ordered.LOWEST_PRECEDENCE

QBNemo:


jhoeller:
I've rolled this into a larger revision towards the use of `Ordered.LOWEST_PRECEDENCE` (and corresponding javadoc) across the codebase. Thanks for raising this!

PR #1486, opened 2017-07-25
SPR-15819 - URI variables with MockRestRequestMatchers requestTo

drumonii:
Adds ability to use URI variables with `MockRestRequestMatchers`'s `requestTo`

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
@@ -87,6 +87,18 @@ public abstract class MockRestRequestMatchers {
 	}
 
 	/**
+	 * Assert the request URI string.
+	 * @param expectedUri the expected URI
+	 * @param uriVars zero or more URI variables to populate the expected URI
+	 * @return the request matcher
+	 */
+	public static RequestMatcher requestToUriTemplate(final String expectedUri, final Object... uriVars) {
+		Assert.notNull(expectedUri, "'uri' must not be null");
+		String uri = UriComponentsBuilder.fromUriString(expectedUri).buildAndExpand(uriVars).encode().toString();
+		return request -> assertEquals("Request URI", uri, request.getURI().toString());
+	}
+
+	/**
 	 * Expect a request to the given URI.
 	 * @param uri the expected URI
 	 * @return the request matcher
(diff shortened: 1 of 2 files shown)


snicoll:
@drumonii please create a Jira issue for this change.

drumonii:
@snicoll Created! [SPR-15819](https://jira.spring.io/browse/SPR-15819)

I'll update the test to include the issue number as a comment later

spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
```diff
@@ -79,11 +79,13 @@ public static RequestMatcher requestTo(final Matcher<String> matcher) {
...
 	 * Assert the request URI string.
 	 * @param expectedUri the expected URI
+	 * @param uriVars zero or more URI variables to populate the expected URI
 	 * @return the request matcher
 	 */
-	public static RequestMatcher requestTo(final String expectedUri) {
+	public static RequestMatcher requestTo(final String expectedUri, final Object... uriVars) {
 		Assert.notNull(expectedUri, "'uri' must not be null");
-		return request -> assertEquals("Request URI", expectedUri, request.getURI().toString());
+		String uri = UriComponentsBuilder.fromUriString(expectedUri).buildAndExpand(uriVars).encode().toString();
```
rstoyanchev: This will cause expectedUri to always be encoded which wasn't the case until now. 

One option would be to do this only if uriVars are actually provided but then the method behaves inconsistently with regards to encoding based on whether uriVars are provided or not.

Another option would be to have separate methods -- e.g. `requestTo` for exact string match, `requestToUriTemplate` to expand + encode, and possibly also `requestToUri(String expectedUri, Function<UriBuilder, URI> uriFunction)` for full control (as we've done on the new `WebClient`).

drumonii: In regards to the encoded expectedUri, I looked at the recent changes to use URI vars in [MockMvcResultMatchers](https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/test/web/servlet/result/MockMvcResultMatchers.java#L90) and modeled this chage off that.

Which option do you prefer? The first option is certainly more convenient. The separate method option also works well because if you are going back to refactor `requestTo` to use URI vars, then you can change the method name while at it.

rstoyanchev: Yes I see now that a similar change was made not long ago. The ticket also shows a comment about a subsequent regression https://jira.spring.io/browse/SPR-14790 .. :)

The first option isn't really an option because of the inconsistency I pointed out. I shouldn't have called it an option in the first place. Let's go for these two for now:

`requestTo(String)`
`requestToUriTemplate(String, Object...)`

The `requestTouri(String, Function<UriBuilder, URI>)` can always be added later.

drumonii: Sounds good! Wasn't aware of the regression but I'll work on these changes later.

rstoyanchev:
This is merged, thanks,

PR #1488, opened 2017-07-30
Upgrade jackson to 2.9.0

eddumelendez:
See SPR-15548


snicoll:
This actually broke the build (though Travis isn't reporting it)

PR #1494, opened 2017-08-05
Update RestTemplate.java

ooooow:
ResponseEntity<InputStreamResource> responseEntity = restTemplate.exchange(
                url,
                HttpMethod.GET,
                new HttpEntity<byte[]>(headers),
                InputStreamResource.class);
        OutputStream out = null;
        try {
            InputStream inputStream = responseEntity.getBody().getInputStream();

If I need an original InputStream ,the response is closed here,the original InputStream is closed

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
@@ -676,6 +676,7 @@ public class RestTemplate extends InterceptingHttpAccessor implements RestOperat
 		Assert.notNull(url, "'url' must not be null");
 		Assert.notNull(method, "'method' must not be null");
 		ClientHttpResponse response = null;
+		boolean isClose = true;
 		try {
 			ClientHttpRequest request = createRequest(url, method);
 			if (requestCallback != null) {
@@ -684,8 +685,16 @@ public class RestTemplate extends InterceptingHttpAccessor implements RestOperat
 			response = request.execute();
 			handleResponse(url, method, response);
 			if (responseExtractor != null) {
-				return responseExtractor.extractData(response);
-			}
+                T entityData = responseExtractor.extractData(response);
+                if(entityData instanceof ResponseEntity){
+                    ResponseEntity responseEntity = (ResponseEntity)entityData;
+                    Object o = responseEntity.getBody();
+                    if(o instanceof InputStreamResource){
+                        isClose = false;
+                    }
+                }
+                return entityData;
+            }
 			else {
 				return null;
 			}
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ooooow Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1494)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1494) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ooooow Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1494)!

spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
```diff
@@ -684,8 +685,16 @@ public void delete(URI url) throws RestClientException {
 			response = request.execute();
 			handleResponse(url, method, response);
 			if (responseExtractor != null) {
-				return responseExtractor.extractData(response);
-			}
+                T entityData = responseExtractor.extractData(response);
+                if(entityData instanceof ResponseEntity){
+                    ResponseEntity responseEntity = (ResponseEntity)entityData;
+                    Object o = responseEntity.getBody();
+                    if(o instanceof InputStreamResource){
```
kazuki43zoo: I think better to use `InputStreamSource` instead of `InputStreamResource`.

spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
```diff
@@ -684,8 +685,16 @@ public void delete(URI url) throws RestClientException {
 			response = request.execute();
 			handleResponse(url, method, response);
 			if (responseExtractor != null) {
-				return responseExtractor.extractData(response);
-			}
+                T entityData = responseExtractor.extractData(response);
+                if(entityData instanceof ResponseEntity){
```
kazuki43zoo: - [ ] Why check `ResponseEntity` only? I think should be check the `InputStreamSource` too.

kazuki43zoo (commented):
WDUT?

spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
ooooow: org.springframework.web.client.RestClientException: Could not extract response: no suitable HttpMessageConverter found for response type [interface org.springframework.core.io.InputStreamSource] and content type [application/octet-stream]
I need custom HttpMessageConverter ?

bclozel:
This looks a lot like [SPR-14882](https://jira.spring.io/browse/SPR-14882).
By design, `RestTemplate` is not meant to stream the HTTP response body.

If you think SPR-14882 is not relevant here, please create a new jira issue on http://jira.spring.io and explains what the issue is, preferably with a repro project we can take a look at.

PR #1496, opened 2017-08-12
SPR-14555

NielsDoucet:
Resolve map of dependencies by keying on an enum value instead of bean name.
This is more a proof of concept than a complete implementation, as I believe there are many ways to go about this.

I chose the route to just leave resolving beans by type in place, and remap the keys by searching a method on the value type that returns the expected key type.
If multiple methods have the return type, this throws an exception.
If multiple beans resolve to the same key, this throws an exception.

Very open to suggestions and guidance towards a complete solution.


snicoll:
Closing this as a result of https://github.com/spring-projects/spring-framework/issues/19123#issuecomment-979839996. Thanks for the PR, in any case!

PR #1499, opened 2017-08-16
Jackson encoder and decoder should use provided mime types

rlindooren:
Instead of always using the default JSON mime types.

SPR-15866


rstoyanchev:
Merged, thanks!

PR #1501, opened 2017-08-21
Delete extra characters

vincenthsin:
Underline is superfluous


snicoll:
Thanks for the PR!

Actually those underlines are required. The problem is that the code snippet is processed verbatim and it should include the `quotes` substitutions. I just did that in 3d28e80

PR #1506, opened 2017-08-24
Allow configuring separate CaffeineSpec properties per Cache name

stepio:
Hi,

I've implemented a minor improvement which we already use in our project with Spring and Guava Cache (have not migrated to Caffeine yet).

With this extension we're able to use a single simple `GuavaCacheManager`, just specifying the needed configuration through properties.

In my opinion this suites well for both Spring Framework and Spring Boot projects as allows to achieve more without implementing new code (just setting properties).

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCache.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCache.java
@@ -31,9 +31,9 @@ import org.springframework.util.Assert;
  *
  * <p>Requires Caffeine 2.1 or higher.
  *
- * @author Ben Manes
  * @author Juergen Hoeller
  * @author Stephane Nicoll
+ * @author Ben Manes
  * @since 4.3
  */
 public class CaffeineCache extends AbstractValueAdaptingCache {
--- a/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
 
 import com.github.benmanes.caffeine.cache.CacheLoader;
 import com.github.benmanes.caffeine.cache.Caffeine;
@@ -47,9 +48,10 @@ import org.springframework.util.ObjectUtils;
  *
  * <p>Requires Caffeine 2.1 or higher.
  *
- * @author Ben Manes
  * @author Juergen Hoeller
  * @author Stephane Nicoll
+ * @author Ben Manes
+ * @author Igor Stepanov
  * @since 4.3
  * @see CaffeineCache
  */
@@ -60,6 +62,9 @@ public class CaffeineCacheManager implements CacheManager {
 	private boolean dynamic = true;
 
 	private Caffeine<Object, Object> cacheBuilder = Caffeine.newBuilder();
+
+	@Nullable
+	private Function<String, Caffeine<Object, Object>> cacheBuilderSupplier;
 
 	@Nullable
 	private CacheLoader<Object, Object> cacheLoader;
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@stepio Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1506)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1506) to manually synchronize the status of this Pull Request.

snicoll:
Thanks for the PR!

`CaffeineSupplier` is definitely not framework material. Such feature has already been requested on the Spring Boot tracker and declined fyi.

As  for the change to the cache manager itself, I am not sure. This  current way would be a specialised function that would return the same builder regardless of the name. So perhaps we could change the implementation to do that instead.

What do you think @jhoeller ?

pivotal-issuemaster:
@stepio Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1506)!

stepio:
Duplicates https://github.com/spring-projects/spring-boot/issues/9301 and https://github.com/spring-projects/spring-boot/issues/7235.

stepio:
Sorry for wasting your time, I did searched in Spring Framework PRs, but forgot to check Spring Boot issues. Anyway I don't think that this functionality is replaceable by Spring `Configuration` class with multiple `Cache` beans.
My use case is:
1. Implementing new integration with external service, I often cache the received data. But I usually don't want to waste much time on defining each new `Cache` too precisely (with all the appropriate options as separate keys from properties) - I just mark a new business method with `Cacheable` annotation with specific label.
2. At the same time we usually need to do fine-tuning in Production to prevent unnecessary load on the appropriate service while keeping the data up-to-date. So I can easily achieve this with Caffeine's well-documented specification - just leave the appropriate note in JIRA for DevOps team with public link to get more information from.

ben-manes:
These request sounds like a nice usability improvement to me, so naively I'd be for them. Unfortunately since I don't have projects that use Spring Cache, I don't have the experience to advocate one way or the other and trust the expert's opinion. I forked the Guava integration and made minimal changes to assist the adoption.

If this was to move forward I would make the following asks,
 - Please move me to the 3rd author, as if I made myself first it was accidental.
 - Is `getCache` if called on every AOP operation, then `computeIfAbsent` should be avoided. It locks [pessimistically](https://github.com/ben-manes/caffeine/wiki/Benchmarks#compute) for subtle reasons, though JDK9 does add partial prescreening by checking the first entry in the bucket. A get-then-compute should be preferred in contended code paths. You could do this or use an unbounded Caffeine cache since it implements this idiom.

stepio:
Hi @ben-manes ,

Thank you very much for the Caffeine itself and your valuable comments above. Hope that I fulfilled them properly with the latest change.

Hi @snicoll ,

In response to your comment, I've created `getNativeCacheBuilder` method. Also I made `refreshKnownCaches` protected to allow invoking it in extended implementations.

P.S.: I fully understand that the PR in general most probably would be declined, but still I'll keep it up-to-date while it's open.

snicoll:
You didn't waste my time at all! Thanks a lot for your effort. The PR is still open and under consideration. 

Thanks for the feedback Ben!

stepio:
Thanks @snicoll !

stepio:
Hi @jhoeller ! Any thoughts on this?

jasonbritton-wk:
We could certainly make use of this improvement as well.  Just because a set of caches are managed by the same manager doesn't mean they should be forced to use the same configuration.

zi-yang-zhang:
👍

gaganbasra:
👍

brunocl:
Any progress with this PR? We wanted to avoid creating customized code on our side because we also have cases we want different cache settings using same cache manager.

davidgoate:
I'm also keen to know about progress on this: https://stackoverflow.com/questions/49885064/is-it-possible-to-set-a-different-specification-per-cache-using-caffeine-in-spri

stepio:
Gentleman,

I appreciate the work of Spring team and fully understand that they have much work with more important updates.
So I decided to manage this feature myself - now it's in a [separate tiny project](https://github.com/stepio/coffee-boots).

Detailed information is in project's README, however here is short info:
```xml
<dependency>
    <groupId>io.github.stepio.coffee-boots</groupId>
    <artifactId>coffee-boots</artifactId>
    <version>1.0.1</version>
</dependency>
```
Example of configuration:
```properties
coffee-boots.cache.spec.myCache=maximumSize=100000,expireAfterWrite=1m
```
where `myCache` is a name of the appropriate cache.

So if the case is still relevant for you, enjoy!

snicoll:
That works for me, let's close this PR in favor of https://github.com/stepio/coffee-boots

PR #1520, opened 2017-09-08
Remove "chin" in admonition blocks of reference documentation.

odrotbohm:
Without the change proposed, admonition blocks carry quite a "chin" that looks heavy.

Before:

bildschirmfoto 2017-09-08 um 15 50 29

After:

bildschirmfoto 2017-09-08 um 15 50 39


sdeleuze:
Fixed via https://github.com/spring-projects/spring-framework/commit/2d6364eb834791bf5a35ec274a71d8dbccab8b67.

@olivergierke I have backported the proposed changed to the original SASS source via https://github.com/bclozel/asciidoctor-stylesheet-factory/pull/1, thanks for contributing.

PR #1523, opened 2017-09-11
Spr 15950 PathMatchingResourcePatternResolver.java cannot find any resources in jar file in weblogic and websphere

kk580kk:
In my project, I used this code to find resources in jar file. It works nice in tomcat and spring-boot:run but cannot find any resources in jar file in weblogic and websphere . So by read this stackoverflow https://stackoverflow.com/questions/17429736/spring-configuration-files-loading-not-working-on-websphere-7 I think rewrite PathMatchingResourcePatternResolver() maybe the best answer. Issue: SPR-15950 huangjie 2017/9/11 09:29

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java
+++ b/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java
@@ -40,9 +40,9 @@ import java.util.zip.ZipException;
  * internal Ant-style regular expressions (matched using Spring's
  * {@link org.springframework.util.AntPathMatcher} utility).
  * Both of the latter are effectively wildcards.
- * <p>
+ *
  * <p><b>No Wildcards:</b>
- * <p>
+ *
  * <p>In the simple case, if the specified location path does not start with the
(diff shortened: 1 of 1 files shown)


kk580kk:
https://jira.spring.io/browse/SPR-15950

rstoyanchev:
The entire source file is modified to use spaces. Can you fix your pull request to use tabs (as it was before) so it is possible to see what actually changed?

kk580kk:
@rstoyanchev 
I'll try  to use tabs

kk580kk:
@rstoyanchev I have formatted the code now

lqjack (commented):
the exception should be throw and not swallow .

file scan can reuse the function framework provided

kk580kk:
In this situation catch the MalformedURLException  and fall back to normal function is the best choice. This change only for websphere and weblogic. @lqjack 
Reuse the function framework provided maybe better answer. But could you tell me which function?

snicoll:
Closing as discussed in the related issue, https://github.com/spring-projects/spring-framework/issues/20502#issuecomment-981779101

PR #1528, opened 2017-09-15
Allow to override a validation message using bean validation message code

kazuki43zoo:
I've fixed [SPR-15967](https://jira.spring.io/browse/SPR-15967).
Please review this.


snicoll:
Thanks for the PR @kazuki43zoo. I did rebase it on the latest change and I believe we need to take another look to how this can be solved.

PR #1530, opened 2017-09-18
fix class name in javadoc

qwazer:
appeared after renaming in commit 760cab8fea4710dc6bf207a3d7d38244518894cf


snicoll:
Thank you so much for your first contribution to Spring Framework. This is now merged in `master`.

PR #1533, opened 2017-09-25
Remove jackson-datatype-jdk7 auto registration

eddumelendez:
`jackson-datatype-jdk7` is not provided anymore. It's handlers are
incorporated in core databind.


bclozel:
Thanks @eddumelendez !

PR #1534, opened 2017-09-25
SPR-15993: Revisit meaning of response.close() in HTTP client

bclozel:
Prior to this issue, SPR-15920 added this new `close()` method which was
supposed to be called to clean resources after response processing.

This commit changes the meaning of that method: calling `close()` will
close the underlying HTTP connection. This has to be called if the
response body is not consumed by the application, since at that point
the underlying connection might be in an inconsistent state if shared in
a connection pool.

Issue: SPR-15993

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/ClientHttpResponse.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/ClientHttpResponse.java
@@ -43,12 +43,13 @@ public interface ClientHttpResponse extends ReactiveHttpInputMessage, Closeable 
 	MultiValueMap<String, ResponseCookie> getCookies();
 
 	/**
-	 * Close this response, freeing any resources created.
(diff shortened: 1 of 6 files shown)


bclozel:
Something we discussed today - WDYT @poutsma @rstoyanchev @sdeleuze ?

rstoyanchev:
I think it would be a good idea to explicitly mention in all those places in the javadoc that closing the connection is not needed if you use any methods that read the response body and that doing so cancels the benefits of connection pooling.

Also please add the same details in `WebClient#exchange` so we're covered in all 3 places.

oleg-alexeyev (approved):
While working around SPR-16011 we did exactly same change: `response.bodyToFlux()` -> `response.body(BodyExtractors.toFlux())`, effectively removing `close()` call from the chain, and everything started to work OK.

PR #1539, opened 2017-09-28
Fix SPR-16024 broken links to Message Converters

arend-von-reinersdorff:
Replace wrong link #rest-message-conversion with correct link 
#mvc-config-message-converters.


snicoll:
@arend-von-reinersdorff thank you for the PR. Actually this section already exists but has moved to `integration.adoc`. I fixed the broken link in f2d040c

PR #1542, opened 2017-09-29
SPR-16027: Remove possible Servlet instance leak

ArloL:
This happens for example from JolokiaMvcEndpoint in spring-boot-actuator 1.5.7 where ``@ConfigurationProperties`` triggers two calls to ``afterPropertiesSet()`` thus resulting in a duplicate AgentServlet that is initialized but never destroyed.


snicoll:
Thanks for the PR but I am not keen to merge it. It is a very broad solution to a very specific problem that includes many other layers and I'd rather focus on the issue at hand rather.

PR #1543, opened 2017-09-29
Proofreading pass

Buzzardo:
I proofread the WebFlux documentation. I had a couple suggestions and questions, which I put in as TODOs.

### DIFF
--- a/src/docs/asciidoc/web/webflux-functional.adoc
+++ b/src/docs/asciidoc/web/webflux-functional.adoc
@@ -4,63 +4,65 @@
 Spring WebFlux provides a lightweight, functional programming model where functions
 are used to route and handle requests and where contracts are designed for immutability.
 It is an alternative to the annotated-based programming model but runs on the same
-<<web-reactive.adoc#webflux-reactive-spring-web>> foundation
-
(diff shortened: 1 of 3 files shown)


rstoyanchev:
@Buzzardo I have to reject this PR in full due to unresolved merge conflicts in it. 

Looking at the history, both you and I worked on the same documentation extensively (since the 27th) and you ended up with lots of conflicts (i.e. changes to the exact same parts) by the time you sent the PR. Please refrain from making changes while the documentation is actively being worked on. It is not ready for proofreading and making changes simultaneously leads to this situation.

Also keep in mind you should never submit a PR with merges to begin with. Either rebase on top of the latest before submitting your PR, or let the reviewer do that. That gives the reviewer maximum control, the ability squash, rebase, and otherwise see the exact changes. In this case you have done a merge and I can't even squash the two commits into one. So I'm forced to review individual commits one by one.

For such extensive changes please refrain from making formatting changes (like the length of lines) in the same commit as other changes. It creates a lot of noise for the reviewer to try and separate what is real change and what is just a line length adjustment.

Buzzardo:
Sorry. I'm still getting used to git.

PR #1544, opened 2017-09-29
SPR-9019 - Clarify docs for usage of @Configurable objects

bedla:
Hi,
I had similar issue as #13658 but with LocalContainerEntityManagerFactoryBean and Hibernate creating instances of cutom UserType classes with injected members.
I was digging around and found, that it is all about depend-on as described in docs. But problem was that there is no bean with proper name created as noted in docs. I am sending PR with added alias for aspect bean and old bean name kept, because I want to be backward compatible. I have written note into docs, to be more specific and be able to be found by google search.
Hope PR is ok for you.
Thx,
Ivos


jhoeller:
Closing this PR based on https://github.com/spring-projects/spring-framework/issues/13658#issuecomment-1026785289 since we're trying to address the problem from a different angle, and for the depends-on workaround, it doesn't really matter whether the aspect bean name is "internal" for the time being.

PR #1548, opened 2017-10-05
SPR-9989 Using multiple PropertyPlaceholderConfigurer breaks @Value default value behaviour

NathanQingyangXu:
This is a difficult ticket requiring nontrivial refactoring (see comments on https://jira.spring.io/browse/SPR-9989 for details). I explored quite some alternatives and found the following plan most ideal:

- tweak PropertyPlaceholderHelper's implementation to accept a new boolean parameter (ignoreDefault) to control whether default value will be used for a specific StringValueResolver;

- change AbstractBeanFactory's internal implementation of 'resolveStringValue(String)' to dictate that only the last StringValueResolver will go about default value return (if needed) and ignore default value for all the other ones, by passing appropriate boolean parameter to the above tweaked PropertyPlaceHolderHelper's public method;

A a new unit testing method named 'resolveEmbeddedDefaultValue' was added to DefaultListableBeanFactoryTests.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractBeanFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractBeanFactory.java
@@ -71,11 +71,7 @@ import org.springframework.core.NamedThreadLocal;
 import org.springframework.core.ResolvableType;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.lang.Nullable;
-import org.springframework.util.Assert;
-import org.springframework.util.ClassUtils;
-import org.springframework.util.ObjectUtils;
-import org.springframework.util.StringUtils;
-import org.springframework.util.StringValueResolver;
+import org.springframework.util.*;
 
 /**
  * Abstract base class for {@link org.springframework.beans.factory.BeanFactory}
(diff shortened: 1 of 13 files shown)


spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractBeanFactory.java
```diff
@@ -830,8 +826,10 @@ public String resolveEmbeddedValue(@Nullable String value) {
 			return null;
 		}
 		String result = value;
-		for (StringValueResolver resolver : this.embeddedValueResolvers) {
-			result = resolver.resolveStringValue(result);
+		Iterator<DefaultIgnorableStringValueResolver> resolverIterator = this.embeddedValueResolvers.iterator();
+		while (resolverIterator.hasNext()) {
+			DefaultIgnorableStringValueResolver resolver = resolverIterator.next();
+			result = (resolverIterator.hasNext() ? resolver.resolveStringValueIgnoringDefault(result) : resolver.resolveStringValue(result));
```
NathanQingyangXu: The gist of this PR. Only the last DefaultIgnorableStringValueResolver will enable 'default value' processing.

PR #1549, opened 2017-10-06
Fix markdown

wilder:
The list of resources for working with Asciidoctor were being displayed in the same line because of the missing line break.


snicoll:
@WilderPereira thank you so much for your first contribution to Spring Framework. This is now merged in `master`.

PR #1558, opened 2017-10-12
Integrity violation exceptions are being hidden for EXASOL.

YLesko:
I have discovered that when java.sql.Connection#setAutoCommit set to 'false' PK/FK violations ignored. After debugging I noticed that Exasol does these checks on  java.sql.Statement#close rather than on java.sql.PreparedStatement#executeUpdate. Checked   JdbcUtils class I found that it is ignoring all "on close" exceptions however this specific case should be handled because otherwise  transactions  management  won't work with EXASOL.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/JdbcUtils.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/JdbcUtils.java
@@ -34,6 +34,7 @@ import javax.sql.DataSource;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.jdbc.CannotGetJdbcConnectionException;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@YLesko Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1558)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1558) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@YLesko Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1558)!

snicoll:
Thanks for the PR but we can't hardcode a vendor-specific handling like that. If you're still experiencing this issue, please create a new issue with a small sample and we can look what, if any, we can do in the core framework about it.

PR #1562, opened 2017-10-13
Rest based cache manager.

pablocloud:
A rest cache manager build on top of Unirest.

Expects an base endpoint and a class mapping for deserialization.

PD: first pull request ever, note me anything pls

### DIFF
--- a/spring-context-support/spring-context-support.gradle
+++ b/spring-context-support/spring-context-support.gradle
@@ -14,6 +14,7 @@ dependencies {
 	optional("org.quartz-scheduler:quartz:2.3.0")
 	optional("org.codehaus.fabric3.api:commonj:1.1.0")
 	optional("org.freemarker:freemarker:${freemarkerVersion}")
+	compile("com.mashape.unirest:unirest-java:1.4.9")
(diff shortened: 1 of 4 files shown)


snicoll:
@pablocloud thanks for the PR! We usually tend to have support for caching libraries outside of Spring Framework these days. That said, I am not sure I understand the use case yet. Can you please describe it in more details?

pablocloud:
A real case in my work place is : 

We have a really big hazelcast instance that's shared among all internal apps, and it is exposed via rest.
This instance doesnt know wich types the other apps use or wich one he is going to serialize, it just receives "objects" and creates a cache for it.

This way we have a single cache system and one App can use data from another.

Why rest and not to use the hazelcast connector? Simple, the app doesnt even know whats on the other said, it could be a Hazelcast, Coherence, JCache, EhCache...

Just thought this could be usefull for some people.

Have a good day

snicoll:
Thanks for the feedback. That confirms this change is quite specific and not a material for the framework. You can put the code in a separate jar that you release to Maven central for instance, I don't think there is a specific place where that code should land.

PR #1565, opened 2017-10-16
SPR-7925 Introduce Assert.noNullElements for collections

rweisleder:
As proposed in [SPR-7925](https://jira.spring.io/browse/SPR-7925).

The issue also proposes a `noNullKeys(Map)` and `noNullValues(Map)`. If this makes sense I would also implement it with this or a separate PR.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/Assert.java
+++ b/spring-core/src/main/java/org/springframework/util/Assert.java
@@ -480,6 +480,49 @@ public abstract class Assert {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


rweisleder:
PMD complains about missing asserts in the new tests. I just copied existing tests. Is there a better way to check that no exception was raised or is it okay for these tests to leave them without assert?

sbrannen:
Superseded by work performed in conjunction with #23151.

See 4000b244ff7b1db1f4eba9c4dbf2fc97162da254.

PR #1566, opened 2017-10-17
Missing some double backticks with apostrophes

drumonii:
Fixes incorrect monospace formatting of unwanted text until next backtick is found. I can also do for the 4.3.x branch if ok.


snicoll:
@drumonii thanks for the PR. I've backported the changes to `4.3.x` already.

PR #1571, opened 2017-10-21
SPR-7961 Allow MailMessage to have multiple Reply-To addresses

rweisleder:
As proposed in [SPR-7961](https://jira.spring.io/browse/SPR-7961).

The JavaMail API states that "normally only a single address will be specified", but `MailMessage` should be consistent with `MimeMessage`.


snicoll:
Thanks for the PR but the issue has been declined.

PR #1573, opened 2017-10-22
Remove redundant lower cased strings creation for their comparison in startsWithIgnoreCase method.

ruslan5t:
Remove redundant lower cased strings creation for their comparison in startsWithIgnoreCase method.


ruslan5t:
Probably it could be closed because of same reason as https://github.com/spring-projects/spring-framework/pull/1574

jhoeller:
Indeed, closing this as per the resolution of https://jira.spring.io/browse/SPR-16095.

PR #1574, opened 2017-10-22
Remove redundant lower cased strings creation for their comparision in endsWithIgnoreCase method.

ruslan5t:
Remove redundant lower cased strings creation for their comparision in endsWithIgnoreCase method.


jhoeller:
Thanks for the PR - again, a good catch there! It turns out that this can be implemented even more efficiently through direct use of `String.regionMatches', for 'endsWithIgnoreCase' as well as 'startsWithIgnoreCase', so I went with that even more streamlined version right away.

PR #1577, opened 2017-10-25
Allow empty usernames for BasicAuth

BattleRattle:
As described in [SPR-16116](https://jira.spring.io/browse/SPR-16116):

The RFCs around basic authentication don't explicitly disallow empty
usernames. On the other hand usernames containing colons are, as colons
are used to separate the username from the password.

This PR will allow empty usernames (empty string or NULL) to be used, but will disallow to use usernames containing a colon.


rwinch:
Thank you for the PR @BattleRattle! This is now merged into master

PR #1578, opened 2017-10-26
Handle parameters of Kotlin extension methods correctly.

diesieben07:
As described in [SPR-16119](https://jira.spring.io/browse/SPR-16119):

The `EXTENSION_RECEIVER` parameter of Kotlin's extension methods appear as
normal method parameters to Java and thus require a name. The synthetic
name `$receiver` is used here, as it is not a valid Kotlin identifier,
but valid in Java.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/KotlinReflectionParameterNameDiscoverer.java
+++ b/spring-core/src/main/java/org/springframework/core/KotlinReflectionParameterNameDiscoverer.java
@@ -75,11 +75,16 @@ public class KotlinReflectionParameterNameDiscoverer implements ParameterNameDis
 	private String[] getParameterNames(List<KParameter> parameters) {
 		List<KParameter> filteredParameters = parameters
 				.stream()
-				.filter(p -> KParameter.Kind.VALUE.equals(p.getKind()))
+				// SPR-16119: Extension receivers of extension methods must be included as they appear as normal method
+				//            parameters in Java
+				.filter(p -> KParameter.Kind.VALUE.equals(p.getKind()) || KParameter.Kind.EXTENSION_RECEIVER.equals(p.getKind()))
 				.collect(Collectors.toList());
 		String[] parameterNames = new String[filteredParameters.size()];
 		for (int i = 0; i < filteredParameters.size(); i++) {
-			String name = filteredParameters.get(i).getName();
+			KParameter parameter = filteredParameters.get(i);
+			// extension receivers are not explicitly named, but require a name for Java interoperability
+			// $receiver is not a valid Kotlin identifier, but valid in Java, so it can be used here
+			String name = KParameter.Kind.EXTENSION_RECEIVER.equals(parameter.getKind())  ? "$receiver" : parameter.getName();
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/core/KotlinReflectionParameterNameDiscoverer.java
```diff
@@ -75,11 +75,16 @@
...
+				//            parameters in Java
+				.filter(p -> KParameter.Kind.VALUE.equals(p.getKind()) || KParameter.Kind.EXTENSION_RECEIVER.equals(p.getKind()))
 				.collect(Collectors.toList());
 		String[] parameterNames = new String[filteredParameters.size()];
 		for (int i = 0; i < filteredParameters.size(); i++) {
-			String name = filteredParameters.get(i).getName();
+			KParameter parameter = filteredParameters.get(i);
+			// extension receivers are not explicitly named, but require a name for Java interoperability
+			// $receiver is not a valid Kotlin identifier, but valid in Java, so it can be used here
+			String name = KParameter.Kind.EXTENSION_RECEIVER.equals(parameter.getKind())  ? "$receiver" : parameter.getName();
```
sdeleuze: As described in my JIRA issue comment, I think we should get the actual parameter name from `Method` instead of introducing `$receiver`.

spring-core/src/test/kotlin/org/springframework/core/KotlinReflectionParameterNameDiscovererTests.kt
```diff
@@ -42,11 +42,22 @@ class KotlinReflectionParameterNameDiscovererTests {
 		assertThat(actualParams, `is`(arrayOf("message")))
 	}
 
+	@Test
+	fun getParameterNamesOnExtensionMethod() {
+		val method = ReflectionUtils.findMethod(UtilityClass::class.java, "identity", String::class.java)!!
+		val actualParams = parameterNameDiscoverer.getParameterNames(method)!!
+		assertThat(actualParams, `is`(arrayOf("\$receiver")))
```
sdeleuze: As described in my JIRA issue comment, I think we should check the actual parameter name against the one retrieved from `Method`.

sdeleuze (changes requested):

spring-core/src/main/java/org/springframework/core/KotlinReflectionParameterNameDiscoverer.java
diesieben07: Except that there is no actual parameter name. We either will always get `arg0` (which is pretty useless) or `$receiver` if Kotlin's `javaParameters` flag is set.

sdeleuze: Good point, let's keep the meaningful parameter name.

sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/0e49e321889d23828c5307842ac25d444ce7ec88, thanks for your contribution.

PR #1585, opened 2017-11-02
Fixed a broken link

Buzzardo:
John Blum reported that the link whose text reads "Request and session scoped beans" points to https://docs.spring.io/spring/docs/current/spring-framework-reference/testing.html#beans-factory-scopes-other but should point to https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#beans-factory-scopes.

I fixed the asciidoc file accordingly. My text editor (Atom) strips spaces at the ends of lines, which accounts for the other changes.


rstoyanchev:
Merged.

PR #1589, opened 2017-11-07
SPR-16171 MediaType, add application/schema+json

simplicii:
Implementation of SPR-16171 -  adding MediaTypes
- application/schema+json
- application/schema+json;charset=UTF-8
- application/problem+json;charset=UTF-8


bclozel:
Thanks for your contribution, but we’re going to decline it due to the reasons listed in the associated isssue.

PR #1591, opened 2017-11-10
SPR-16162 make JpaVendorAdapter configuration properties more transaction type aware

apixandru:
Autoconfigure based on the transaction type if detected. If the transaction type cannot be detected for whatever reason, it falls back to the current behavior.

This is basically to prevent the Vendor Adapter to set an invalid connection release mode for JTA transactions but can be used for further configuration based on the detected transaction type.

Running the application with embedded tomcat:
image

Running with weblogic
image


jhoeller:
As discussed on JIRA, a slightly different revision with the same intent is now in master. I'll backport it to 4.3.x over the weekend; it'd be great if you give it a try before the 5.0.2 / 4.3.13 release (mid next week).

PR #1592, opened 2017-11-13
SPR-16022: avoid implicit autowiring with Kotlin secondary ctors

sdeleuze:
This one is tricky, but after some time to think and experiment on it, I tend to think we went maybe a little bit to far with implicit autowiring of Kotlin classes with secondary constructors, which has impact on `ConstructorResolver`.

In order to fix [SPR-16022](https://jira.spring.io/browse/SPR-16022), my proposal is to avoid implicit autowiring when the Kotlin class has secondary constructors.

Notice that Kotlin compiler can generate (in additional to the synthetic constructor generated for primary constructor) a non-synthetic default constructor when using no-arg compiler plugin or when a class have primary constructor with optional parameters that have all default values. This proposed change try to handle all these use cases.


sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/edf82325557f05f482b0b9c0333d99908c8f1378

PR #1593, opened 2017-11-14
SPR-16098: Fix Reactive JsonView + HttpEntity handling

sdeleuze:
This commit adds an `AbstractMessageWriterResultHandler#writeBody` variant which allows to pass the actual `MethodParameter` in order to perform proper annotation-based hint resolution with nested generics like with `ResponseEntityResultHandler`.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/AbstractMessageWriterResultHandler.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/AbstractMessageWriterResultHandler.java
@@ -43,6 +43,7 @@ import org.springframework.web.server.ServerWebExchange;
  * to the response with {@link HttpMessageWriter}.
  *
  * @author Rossen Stoyanchev
+ * @author Sebastien Deleuze
  * @since 5.0
  */
 public abstract class AbstractMessageWriterResultHandler extends HandlerResultHandlerSupport {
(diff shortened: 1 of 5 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/AbstractMessageWriterResultHandler.java
```diff
@@ -86,9 +87,33 @@ protected AbstractMessageWriterResultHandler(List<HttpMessageWriter<?>> messageW
...
+	 * @param bodyParameter the {@link MethodParameter} of the body to write.
+	 * @param actualParameter the actual return type of the method that returned the
+	 * value; could be different from {@code bodyParameter} when processing {@code ResponseEntity}
+	 * for example.
+	 * @param exchange the current exchange.
+	 * @return indicates completion or error
+	 * @since 5.0.2
+	 */
+	@SuppressWarnings("unchecked")
+	protected Mono<Void> writeBody(@Nullable Object body, MethodParameter bodyParameter, ServerWebExchange exchange, @Nullable MethodParameter actualParameter) {
```
rstoyanchev: Let's put ServerWebExchange as the last argument, i.e. `(body, bodyParameter, actualParameter, exchange)`.

rstoyanchev (commented):
Looks fine. Do we have the same issue with @RequestBody and HttpEntity?

spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/AbstractMessageWriterResultHandler.java
sdeleuze: Indeed we have the same issue with `HttpEntity` I will fix that as well with a similar strategy. Notice that `@RequestBody HttpEntity<Foo> foo entity` does not seems to be currently supported, only `HttpEntity<Foo> foo entity`. Likely an `ArgumentResolver` ordering issue, I will create a JIRA issue for that.

sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/c530745015d5c7031cfcb6f633f3f459ed97935d.

PR #1594, opened 2017-11-14
Updating WebClient's javadocs

marcingrzejszczak:
without this change the code in the Javadocs of `WebClient` is invalid. It doesn't compile.
with this change, I've updated the Javadocs to be consistent with the current implementation

related issue - https://jira.spring.io/browse/SPR-16197

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClient.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClient.java
@@ -51,8 +51,8 @@ import org.springframework.web.util.UriBuilderFactory;
  * // Perform requests...
  * Mono&#060;String&#062; result = client.get()
  *     .uri("/foo")
- *     .exchange()
- *     .then(response -> response.bodyToMono(String.class));
+ *     .retrieve()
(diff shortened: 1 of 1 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClient.java
```diff
@@ -51,8 +51,8 @@
  * // Perform requests...
  * Mono&#060;String&#062; result = client.get()
  *     .uri("/foo")
- *     .exchange()
- *     .then(response -> response.bodyToMono(String.class));
+ *     .retrieve()
```
simonbasle: @marcingrzejszczak note the `exchange()` is still valid but then the example should use `flatMap(response -> ...)` rather than `then`

marcingrzejszczak: yeah, it's actually written later in the javadocs. I guess in this example we want the simplest case so I've picked the `retrieve` one. Does it make sense or should I change it?

rstoyanchev:
I've done a more comprehensive update. Thanks for the PR.

PR #1596, opened 2017-11-15
UndertowServerHttpRequest: suspend reading when no demand

violetagg:
- When there is no demand for reading, the implementation
should suspend reading otherwise useless events will be
send by Undertow to the registered read listener.
- There is not need to wait for an event for reading/writing after calling
resumeReads/resumeWrites

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractListenerReadPublisher.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractListenerReadPublisher.java
@@ -113,6 +113,12 @@ public abstract class AbstractListenerReadPublisher<T> implements Publisher<T> {
 	@Nullable
 	protected abstract T read() throws IOException;
 
+	/**
+	 * Suspend reading. Defaults to no-op.
+	 */
+	protected void suspendReading() {
+	}
+
 
 	/**
 	 * Read and publish data from the input. Continue till there is no more
@@ -256,6 +262,7 @@ public abstract class AbstractListenerReadPublisher<T> implements Publisher<T> {
 						}
 						else {
 							publisher.changeState(READING, NO_DEMAND);
+							publisher.suspendReading();
 						}
 					}
 					catch (IOException ex) {
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/UndertowServerHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/UndertowServerHttpRequest.java
@@ -152,7 +152,28 @@ class UndertowServerHttpRequest extends AbstractServerHttpRequest {
 		protected void checkOnDataAvailable() {
 			// TODO: The onDataAvailable() call below can cause a StackOverflowError
 			// since this method is being called from onDataAvailable() itself.
-			onDataAvailable();
+			if (isReadPossible()) {
+				onDataAvailable();
+			}
+		}
+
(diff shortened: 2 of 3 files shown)


wilkinsona:
I've built this locally and run the previously failing Spring Boot test against it. The test now passes.

rstoyanchev:
@violetagg the `WebSocketReceiverPublisher` sub-class has a `suspendReceiving` (and also `resumeReceiving` methods. The introduction of a `suspendReading` in the base class raises questions about how it relates to `suspendReceiving`? Can those be unified with a suspend/resume read protected methods in the base class?

violetagg:
> Can those be unified with a suspend/resume read protected methods in the base class?

I would suggest to keep these two changes separate.
We might need some refactoring.

sdeleuze:
@rstoyanchev I need to merge this PR to be able to upgrade to Reactor `Bismuth-SR4` otherwise a Spring Framework Websocket test fails. Could I merge it in its current form ?

sdeleuze:
The flacky Websocket test was unrelated, but as discussed in the Reactive meeting, I will merge this PR and @violetagg will commit the refactoring asked by @rstoyanchev after `5.0.2` release, that will be safer.

sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/0fbfa643856b07eb0d952905e527765fbd5c274f.

@violetagg Please create a dedicated new PR for the 5.0.3 design update requested by @rstoyanchev .

rstoyanchev:
I guess since the current change promotes a suspend concept to the base class, we need to decide if that should be shared with the WebSocket implementation or does it need to be in sub-classes only, and not in the base class. Let's come back to this soon after the 5.0.2. I've created a ticket to keep track of this https://jira.spring.io/browse/SPR-16207.

mp911de:
Chiming in here as Lettuce has a similar concept of suspending reads along a Pub/Sub (shared subscriptions) notion. Command parsing is suspended if subscriber demand is satisfied. Pub/Sub can be shared across multiple subscribers and they need to specify their backpressure behavior when obtaining a publisher to not prevent other subscribers from message processing and to not drop messages.

violetagg:
Hi,

A PR #1598 was created to meet the requirements for the web socket implementation and https://jira.spring.io/browse/SPR-16207 was updated with details for the approach that I chose to use.

Regards,
Violeta

spring-web/src/main/java/org/springframework/http/server/reactive/AbstractListenerReadPublisher.java
```diff
@@ -256,6 +262,7 @@ private State state() {
 						}
 						else {
 							publisher.changeState(READING, NO_DEMAND);
+							publisher.suspendReading();
```
rstoyanchev: @violetagg if we switch to NO_DEMAND first and then suspend, isn't it possible that if a request comes in concurrently, after NO_DEMAND but before suspend, it will checkOnDataAvailable, but then the suspend after may end up cancelling the request for data? It seems safer to suspend first and then change the state.

if we switch to NO_DEMAND first and then suspend reading, isn't there a possibility that in between a request may come which would call check

rstoyanchev: Also shouldn't suspend be called in UNSUBSCRIBED too before there is demand? It should apply every time we're about to switch to NO_DEMAND.

PR #1598, opened 2017-11-17
AbstractListenerWebSocketSession: suspend the channel when there is no demand

violetagg:
Issues: SPR-16207

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/socket/adapter/UndertowWebSocketSession.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/socket/adapter/UndertowWebSocketSession.java
@@ -61,12 +61,19 @@ public class UndertowWebSocketSession extends AbstractListenerWebSocketSession<W
 		return true;
 	}
 
+	@Override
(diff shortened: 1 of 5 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/socket/adapter/UndertowWebSocketSession.java
```diff
@@ -61,14 +61,21 @@ protected boolean canSuspendReceiving() {
...
 
+	@Override
 	protected void resumeReceiving() {
 		getDelegate().resumeReceives();
 	}
 
+	@Override
+	protected boolean isReceivesResumed() {
+		return getDelegate().isReceivesResumed();
+	}
```
rstoyanchev: Could we call this `isSuspended`? In both places where it's called it's negated..

violetagg: done

rstoyanchev:
Merged, thanks.

PR #1599, opened 2017-11-19
[Obvious Fix] Fix typos, formatting and escaping in reference documentation

bianjp:
Fix various typos, formatting and escaping issues that I find while reading through the reference documentation.

This is an obvious fix.

Note: Some symbol combinations (including "<=") serve as textual symbol replacements in AsciiDoc, and require escaping when replacement isn't expected. See http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#text-replacement


src/docs/asciidoc/core/core-aop.adoc
```diff
@@ -545,14 +545,14 @@ this purpose. A typical such aspect would look as follows:
 		 *
 		 * If you group service interfaces by functional area (for example,
 		 * in packages com.xyz.someapp.abc.service and com.xyz.someapp.def.service) then
-		 * the pointcut expression "execution(* com.xyz.someapp..service.*.*(..))"
+		 * the pointcut expression "execution(* com.xyz.someapp.service.*.*(..))"
```
dreis2211: Without looking too deep into it, I believe the "fix" here is wrong as using two-dots is actually intended in order to match both the `abc` and `def` package.

src/docs/asciidoc/core/core-aop.adoc
```diff
@@ -545,14 +545,14 @@ this purpose. A typical such aspect would look as follows:
...
 		 * could be used instead.
 		 *
 		 * Alternatively, you can write the expression using the 'bean'
 		 * PCD, like so "bean(*Service)". (This assumes that you have
 		 * named your Spring service beans in a consistent fashion.)
 		 */
-		@Pointcut("execution(* com.xyz.someapp..service.*.*(..))")
+		@Pointcut("execution(* com.xyz.someapp.service.*.*(..))")
```
dreis2211: Same as above, I believe

src/docs/asciidoc/core/core-aop.adoc
bianjp: You're right. I just amended the commit.

rstoyanchev:
Thanks for the updates! The changes are now in master.

PR #1602, opened 2017-11-21
SPR-16166: Handle top level json value in Jackson2JsonTokenizer

madgnome:
This commit introduces support for top level Json value
`String`, `true`, `false`, `null`, `number`) to `Jackson2JsonTokenizer`.

Before this change `Jackson2JsonTokenizer` would only consider token in an object or array and this scenario will fail:

```
@RestController
public class JsonValueResource {
    @GetMapping("/fail")
    public Mono<Boolean> fail() {
        return WebClient.builder().baseUrl("http://localhost:8000")
                .build().get().uri("/")
                .accept(MediaType.APPLICATION_JSON)
                .retrieve()
                .bodyToMono(Boolean.class)
                .switchIfEmpty(Mono.error(new RuntimeException("No response")));
    }

    @GetMapping
    Mono<Boolean> jsonValueBoolean() {
        return Mono.just(true);
    }

}
```

Issue: SPR-16166

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/json/AbstractJackson2Decoder.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/json/AbstractJackson2Decoder.java
@@ -44,6 +44,7 @@ import org.springframework.http.server.reactive.ServerHttpResponse;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.MimeType;
+import reactor.util.function.Tuples;
 
 /**
  * Abstract base class for Jackson JSON 2.9 decoding.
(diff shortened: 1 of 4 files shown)


madgnome:
Codacy/PMD complains about missing asserts in the new tests but I followed the existing pattern in the test. is there a way to ignore?

poutsma:
Thank you for providing a PR for this issue! However, I decided to fix it in an alternative, arguably cleaner way. See https://github.com/spring-projects/spring-framework/commit/8e253a316d7f9f6d76fc343b96d747e1d056a0ae

madgnome:
Thanks a lot for fixing it @poutsma, your solution is cleaner indeed!

Could you double check that this use case is working?
```
testTokenizeAsJsonToken(
				singletonList("true"),
 				singletonList(JsonToken.VALUE_TRUE));
```

poutsma:
Unfortunately, the JSON assert framework we use does not support null, true, or false as top-level elements (see [this issue](https://github.com/skyscreamer/JSONassert/issues/83)). I did test those values in a test project, though, and they worked for me.

PR #1603, opened 2017-11-22
removed incorrect @Nullable annotations from queryForObject methods

isopov:
In both javadocs and implementations it is clearly stated that these methods cannot return null and will throw IncorrectResultSizeDataAccessException instead. It seems that they were marked as `@Nullable` by mistake.


jhoeller:
This is indeed worth discussing; however, the matter is more complicated: It all starts with the `RowMapper` callback interface here which is allowed to map a given row to `null` (as a valid return value for a row that has been found) which bubbles through to those `queryForObject` methods. For clean non-nullability here, we'd have to enforce a non-null result from `RowMapper.mapRow`... which is a backwards-compatibility issue.

isopov:
I see. Sorry. Really nulls are possible in that case.

jhoeller:
That said, `DataAccessUtils` seems to be inconsistent in its `requiredSingleResult` vs `requiredUniqueResult` nullability declarations. I'll dig into this for 5.0.2 still.

PR #1620, opened 2017-12-09
SPR-16282 Deprecate classes removed in 5

marschall:
This commit deprecates the following things

 * the entire NativeJdbcExtractor hierarchy
 * all types in the org.springframework.beans.factory.access package
 * all types in the org.springframework.context.access package
 * all types in the org.springframework.ejb.interceptor package

Issue: SPR-16282


marschall:
The code quality review seems to be hanging because branch analysis doesn't seem to be enabled.

jhoeller:
As per SPR-16282, I'm afraid those candidates have been declared to remain non-deprecated in 4.x on purpose, for the infrastructure support range in that generation of the framework.

PR #1624, opened 2017-12-12
SPR-16295 - Programmatic creation of caching proxies using CacheProxy…

jxblum:
…FactoryBean does not work.

This changes the o.s.cache.interceptor.CacheProxyFactoryBean to implement the SmartInitializingSingleton and BeanFactoryAware interfaces, which delegate to the internal CacheInterceptor in order to appropraiately configure the interceptor to enable the creation of cache proxies programmatically.

Resolves SPR-16295


jhoeller:
As per my comment on SPR-16295, I went with a streamlined variant instead which I also intend to backport to 4.3.14.

PR #1626, opened 2017-12-21
SPR-16319 - limit the scope of the synchronized block

mathieu-pousse:


pivotal-issuemaster:
@mathieu-pousse Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1626)!

rstoyanchev:
It is a little more complicated. Request matching, verifying and counting must be done inside the synchronized block. I've done that with 0c289283ff4fab171b527b7f195639e7a950f470.

PR #1630, opened 2017-12-22
Refactor AnnotatedElementUtils to not cause NoSuchMethodExceptions

philwebb:
Refactor `AnnotatedElementUtils` to loop over methods rather than
cause a `NoSuchMethodException`. Since this method is called often,
this change can help improve application startup times.

Issue:

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotatedElementUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotatedElementUtils.java
@@ -31,6 +31,7 @@ import java.util.Set;
 import org.springframework.core.BridgeMethodResolver;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/core/annotation/AnnotatedElementUtils.java
```diff
@@ -1243,23 +1297,32 @@ else if (element instanceof Class) {
...
 		}
 
 		return null;
 	}
 
+	private static Method findEquivalentMethod(Class<?> clazz, Method method) {
+		Method[] candidates = clazz.getDeclaredMethods();
+		for (Method candidate : candidates) {
+			if (candidate.getName().equals(method.getName()) && Arrays.equals(
+					candidate.getParameterTypes(), method.getParameterTypes())) {
```
dreis2211: May I suggest a pre-check `candidate.getParameterCount() == method.getParameterCount())` in order to avoid possible allocations coming from `getParameterTypes()`?

philwebb: Good suggestions. I've updated the PR. Thanks!

philwebb:
Closing due to https://github.com/spring-projects/spring-framework/pull/1628#issuecomment-354589149

PR #1632, opened 2017-12-23
SPR-16288

Skelotron:
fix BridgeMethodResolver#findGenericDeclaration method work for classes which implement more than one interface


jhoeller:
I'm sorry, I did notice this pull request before... I've fixed this through a commit of my own with a rather extensive MVC integration tests in the meantime. In any case, thanks for the pull request, you were right on track there!

PR #1648, opened 2018-01-20
SPR-16404 - Removed unnecessary cast to int

mrmaus:
Since Java7 HttpURLConnection offers setFixedLengthStreamingMode method with long parameter which should be prefered over version with int argument, therefore casting ContentLength to int is no longer needed. Moreover it makes impossible to stream payload larger than Integer.MAX_VALUE

The issue was actually caught within one of our production applications where RestTemplate is used to stream large binaries. I understand that this is not the intended use of RestTemplate, but since the Spring Framework is baselined on Java8 there is no harm of applying this fix. Moreover according to HttpURLConnection javadoc using the method with long argument must be preferred and it takes precedence over the int version

Excerpt from HttpURLConnection.fixedContentLength javadoc: 

     * <P> <B>NOTE:</B> {@link #fixedContentLengthLong} is recommended instead
     * of this field, as it allows larger content lengths to be set.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/SimpleStreamingClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/SimpleStreamingClientHttpRequest.java
@@ -73,7 +73,7 @@ final class SimpleStreamingClientHttpRequest extends AbstractClientHttpRequest {
 	protected OutputStream getBodyInternal(HttpHeaders headers) throws IOException {
 		if (this.body == null) {
 			if (this.outputStreaming) {
-				int contentLength = (int) headers.getContentLength();
+				long contentLength = headers.getContentLength();
 				if (contentLength >= 0) {
 					this.connection.setFixedLengthStreamingMode(contentLength);
 				}


pivotal-issuemaster:
@mrmaus Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1648)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1648) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mrmaus Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1648)!

snicoll:
@mrmaus we need a Jira issue to track this change. Can you please [create  one](https://jira.spring.io/browse/SPR)?

mrmaus:
Ticket created: SPR-16404
Sorry, I thought the change is quite trivial and followed https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md : 
... 
2. For all but the most trivial of contributions, please create a ticket. ...

PR #1654, opened 2018-01-26
SPR-16372 The FieldError class does not provide access to the exception that triggered the binding error.

marknorkin:
[SPR-16372](https://jira.spring.io/browse/SPR-16372) Provide access to the actual exception that triggered the binding error


marknorkin:
Hello, @jhoeller,  could you please take a look ?

jhoeller:
As mentioned on JIRA, I've addressed this is a more general way. Thanks for the pull request, in any case!

PR #1655, opened 2018-01-27
SPR-16425 - Remove square brackets around stringified collections

logart:


pivotal-issuemaster:
@logart Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1655)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1655) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@logart Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1655)!

spring-web/src/main/java/org/springframework/web/util/UriComponents.java
```diff
@@ -273,7 +271,15 @@ private static String getVariableName(String match) {
 	}
 
 	private static String getVariableValueAsString(@Nullable Object variableValue) {
-		return (variableValue != null ? variableValue.toString() : "");
+		if (variableValue != null) {
```
fmcejudo: might we have this as `if(variableValue == null) {  return ""; }`? and afterward keep evaluating the variable... so, I had in mind something like:
` if(Collection.class.isAssignableFrom(variableValue.getClass())){
      return ((Collection) variableValue).stream().map(Object::toString).collect(Collectors.joining(","));
} else {
   return variableValue.toString();
}`

so, we are safe from a NullPointerException and avoiding nested blocks, what makes the code cleaner, and using a little bit more streams collections features. 

https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#joining-java.lang.CharSequence

What do you think?

fmcejudo:
Also, because there are two commits for only one file and purpose, you might want to squash them into one with rebase -i HEAD~2

https://stackoverflow.com/questions/2563632/how-can-i-merge-two-commits-into-one

logart:
Hello, sorry for long response.
Sure, I will do it today.

rstoyanchev:
Before you do that, please check my comment under SPR-16425.

rstoyanchev:
Closing as the related issue has also been closed.

PR #1659, opened 2018-02-01
MockHttpServletRequestBuilder decodes pathInfo

rwinch:
Previously MockHttpServletRequestBuilder calculated the pathInfo from the
provided URL without decoding the value. This meant that the pathInfo
incorrectly included URL encoded values.

Now MockHttpServletRequestBuilder properly decodes the pathInfo.

Fixes: SPR-16453

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
@@ -59,6 +59,7 @@ import org.springframework.web.servlet.FlashMapManager;
 import org.springframework.web.servlet.support.SessionFlashMapManager;
 import org.springframework.web.util.UriComponentsBuilder;
 import org.springframework.web.util.UriUtils;
+import org.springframework.web.util.UrlPathHelper;
 
 /**
  * Default builder for {@link MockHttpServletRequest} required as input to perform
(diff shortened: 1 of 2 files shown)


spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
```diff
@@ -696,11 +704,49 @@ private void updatePathRequestProperties(MockHttpServletRequest request, String
...
+	 * @see javax.servlet.ServletRequest#getCharacterEncoding()
+	 */
+	private String determineEncoding(HttpServletRequest request) {
+		String enc = request.getCharacterEncoding();
+		if (enc == null) {
+			enc = WebUtils.DEFAULT_CHARACTER_ENCODING;
+		}
+		return enc;
+	}
+
```
rstoyanchev: `decodeRequestString` and `determineEncoding` appear identical to the same methods in `UrlPathHelper`. Why not just use `UrlPathHelper#decodeRequestString` instead?

rwinch: You are right. I copied it from there and somehow missed that was a public method.

rwinch: I updated the Pull Request with your suggestion.

rstoyanchev:
Merged, thanks.

PR #1660, opened 2018-02-01
SPR-16468 - Default to MySQL55Dialect based on InnoDB

meltsufin:
`MySQL5Dialect` set currently as default is based on the MyISAM storage engine, which is not recommended.
MySQL 5.5 has been out since 2010. So, it's pretty safe to go with it as the default dialect.

Related to https://github.com/spring-cloud/spring-cloud-gcp/issues/392.

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaVendorAdapter.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaVendorAdapter.java
@@ -30,7 +30,7 @@ import org.hibernate.dialect.DerbyTenSevenDialect;
 import org.hibernate.dialect.H2Dialect;
 import org.hibernate.dialect.HSQLDialect;
 import org.hibernate.dialect.InformixDialect;
-import org.hibernate.dialect.MySQL5Dialect;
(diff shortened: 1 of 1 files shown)


snicoll:
@meltsufin we need a Jira issue to discuss this change. Can you please [create  one](https://jira.spring.io/browse/SPR)?

meltsufin:
@snicoll [Done](https://jira.spring.io/browse/SPR-16468).

meltsufin:
@snicoll Do you have any updates on this change? I see a number of discussions around it like https://github.com/spring-projects/spring-boot/issues/15342, but it's not clear what is the solution going forward. Thanks!
FYI @saturnism

snicoll:
At the framework level, we won't change the default anytime soon (see [the issue you've created](https://github.com/spring-projects/spring-framework/issues/21013#issuecomment-453467143)). There is a discussion in Spring Boot to stop setting the database so that Hibernate can detect it. 

Thanks for the reminder, I've tried to clarify the situation by creating a dedicated issue for it https://github.com/spring-projects/spring-boot/issues/16172. I suggest to close this PR for now as we don't intend to upgrade the default at the moment.

PR #1679, opened 2018-02-12
Polish: unused method parameters should be removed.

igor-suhorukov:
fix potential issue with parameter ClassWriter.addType(final Item item)

### DIFF
--- a/spring-context/src/main/java/org/springframework/format/datetime/DateTimeFormatAnnotationFormatterFactory.java
+++ b/spring-context/src/main/java/org/springframework/format/datetime/DateTimeFormatAnnotationFormatterFactory.java
@@ -58,15 +58,15 @@ public class DateTimeFormatAnnotationFormatterFactory  extends EmbeddedValueReso
 
 	@Override
 	public Printer<?> getPrinter(DateTimeFormat annotation, Class<?> fieldType) {
-		return getFormatter(annotation, fieldType);
+		return getFormatter(annotation);
 	}
 
 	@Override
 	public Parser<?> getParser(DateTimeFormat annotation, Class<?> fieldType) {
-		return getFormatter(annotation, fieldType);
+		return getFormatter(annotation);
 	}
 
-	protected Formatter<Date> getFormatter(DateTimeFormat annotation, Class<?> fieldType) {
+	protected Formatter<Date> getFormatter(DateTimeFormat annotation) {
 		DateFormatter formatter = new DateFormatter();
 		String style = resolveEmbeddedValue(annotation.style());
 		if (StringUtils.hasLength(style)) {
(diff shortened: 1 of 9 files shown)


spring-context/src/main/java/org/springframework/format/datetime/DateTimeFormatAnnotationFormatterFactory.java
```diff
@@ -58,15 +58,15 @@
...
 	}
 
 	@Override
 	public Parser<?> getParser(DateTimeFormat annotation, Class<?> fieldType) {
-		return getFormatter(annotation, fieldType);
+		return getFormatter(annotation);
 	}
 
-	protected Formatter<Date> getFormatter(DateTimeFormat annotation, Class<?> fieldType) {
+	protected Formatter<Date> getFormatter(DateTimeFormat annotation) {
```
snicoll: This is a potential hookpoint where a subclass may use that information. Please revert this.

spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectivePropertyAccessor.java
```diff
@@ -458,7 +458,7 @@ protected Field findField(String name, Class<?> clazz, boolean mustBeStatic) {
 	 * This method will just return the ReflectivePropertyAccessor instance if it is unable to build
 	 * something more optimal.
 	 */
-	public PropertyAccessor createOptimalAccessor(EvaluationContext evalContext, @Nullable Object target, String name) {
+	public PropertyAccessor createOptimalAccessor(@Nullable Object target, String name) {
```
snicoll: Idem

spring-core/src/main/java/org/springframework/asm/ClassWriter.java
```diff
@@ -1699,7 +1699,7 @@ int addUninitializedType(final String type, final int offset) {
      */
     private Item addType(final Item item) {
         ++typeCount;
-        Item result = new Item(typeCount, key);
+        Item result = new Item(typeCount, item);
```
snicoll: Looks unrelated to the intention of this PR. Can you please clarify?

snicoll (changes requested):
Thanks for the PR. I've made a few suggestions

jhoeller:
Well spotted... Some of those cases turn out to be by design: in particular for the `protected` methods (allowing custom variants to make use of the extra context in a subclass) but also as a design reminder for some internal delegates. Let's leave those as-is for the time being.

That said, I've manually applied two pieces: The internal constructor cross-calling in `DeclareParentsAdvisor` can definitely be simplified as suggested, and ASM's `addItem(Item)` should be coded to pick up its argument indeed (even if it's always been called with the key field as argument). Since the latter originates from ASM, I've checked its source but ASM got heavily refactored in the meantime, so there's no point in reporting this to them. BTW we'll pick up that refactored ASM 6.1 in Spring Framework 5.1 soon.

PR #1684, opened 2018-02-14
SPR-16496 - Exception handler class lookup fix

encircled:
Fixes SPR-16496


jhoeller:
As per SPR-16496, I've addressed this differently (through an AOP target class lookup).

PR #1685, opened 2018-02-14
Align spring-webflux's Webjars Locator dependency with spring-webmvc

wilkinsona:
The intention is to align with [this change](https://github.com/spring-projects/spring-framework/commit/8d42888476e892dfa1b099684c9a381bbd26f916#diff-6b23f872b29c97049a56354504169781L29) to spring-webmvc.

### DIFF
--- a/spring-webflux/spring-webflux.gradle
+++ b/spring-webflux/spring-webflux.gradle
@@ -22,7 +22,7 @@ dependencies {
 	optional(project(":spring-context-support"))  // for FreeMarker support
 	optional("javax.servlet:javax.servlet-api:3.1.0")
 	optional("javax.websocket:javax.websocket-api:1.1")
-	optional("org.webjars:webjars-locator:0.32-1")
+	optional("org.webjars:webjars-locator-core:0.35")
 	optional("org.freemarker:freemarker:${freemarkerVersion}")
 	optional("com.fasterxml.jackson.core:jackson-databind:${jackson2Version}")
 	optional("com.fasterxml.jackson.dataformat:jackson-dataformat-smile:${jackson2Version}")


jhoeller:
Good catch, Andy!

xak2000:
Why `org.webjars:webjars-locator` is replaced with `org.webjars:webjars-locator-core`? What right dependency I should include in the application to support webjars without version in the url?

Prevoiusly, it was `webjars-locator` and it was placed like this in my `bulid.gradle`:

```
compile('org.webjars:webjars-locator')
```

Now after upgrading spring-boot (and spring framework as a side effect), this doesn't work (without explicit dependency version), because version of this dependency is no more managed by Spring's BOM.

If I change this to

```
compile('org.webjars:webjars-locator-core')
```

then it starts to work.

But the documentation says:

> WebJars is also supported through `WebJarsResourceResolver` and is automatically registered when `org.webjars:webjars-locator` is present on the classpath.

So I just wonder if there is some incosistency here or both dependencies should work and what dependency is really "right"? Maybe it is worth to update the documentation, if the latter is "right" dependency? Or it is worth to add `webjars-locator` too into the BOM?

Also, the remove of `webjars-locator` from version management is breaking change (the build is no more works without explicit version declaration) so maybe it is worth to mention it in changelog.

sbrannen:
> Why `org.webjars:webjars-locator` is replaced with `org.webjars:webjars-locator-core`?

I believe that was because Spring only needs the "core" locator support from WebJars.

> Edit: if it is worth to open a new issue for this, please let me know (and issue about what: update docs, update bom or both).

Yes, please open a new issue requesting that the documentation be updated for the following to reflect the status quo.

- webmvc.adoc
- webflux.adoc
- org.springframework.web.reactive.resource.WebJarsResourceResolver
- org.springframework.web.servlet.resource.WebJarsResourceResolver

Thanks

snicoll:
@xak2000 mixing concerns of two different projects make it hard for us to give you a precise answer.

> Now after upgrading spring-boot (and spring framework as a side effect), this doesn't work (without explicit dependency version), because version of this dependency is no more managed by Spring's BOM.

This has nothing to do with Spring Framework, dependency management is a Spring Boot feature.

> Also, the remove of webjars-locator from version management is breaking change (the build is no more works without explicit version declaration) so maybe it is worth to mention it in changelog.

I think you probably missed [this section of the migration guide](
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#webjars-locator).

jhoeller:
FWIW, `webjars-locator-core` is the actual asset locator support, whereas `webjars-locator` is just RequireJS support that wraps the `webjars-locator-core` project (quoting their own GitHub summary). The core project also has up-to-date artifacts on Maven Central whereas `webjars-locator` usually lags behind, often by several months.

xak2000:
@sbrannen Thanks! I opened #22613.

@snicoll You are right! I must have missed it, sorry for that. I also was confused by the fact that this project has version declared for this dependency and mentions it as "optional", so I thought that it has something to do with spring-boot's bom. But now I understand that it is just for compilation to work.

@jhoeller Thanks, it clarifies the situation. :-)

PR #1687, opened 2018-02-15
SPR-16501: add caching support for cache2k

cruftex:
[cache2k](https://cache2k.org) is one of the [best performing Java caches](https://cruftex.net/2017/09/01/Java-Caching-Benchmarks-Part-3.html) with a hit rate efficiency similar to Caffeine.

The jar size of cache2k is smaller than the currently available alternatives like EHCache and Caffeine. Although its small jar size it comes with XML configuration and JMX support.

The direct support of cache2k makes use of the build in `null` value support which leads to less overhead than Caffeine or using the JCache API.


snicoll:
Thanks for the PR and the support on [SPR-16501](https://jira.spring.io/browse/SPR-16501). As we're suggesting to move this integration to `cache2k` itself, I am closing this PR.

PR #1689, opened 2018-02-15
SPR-16505 - getReader/getInputStream retaining previous result

av-pinzur:
SPR-16505


jhoeller:
Addressed in a revised fashion along with SPR-16499. Thanks for the pull request, in any case!

PR #1690, opened 2018-02-16
Polish: anonymous inner classes containing only one method should become lambdas,

igor-suhorukov:
Polish: anonymous inner classes containing only one method should become lambdas,
use getOrDefault instead of ternary operator


rstoyanchev (commented):
@igor-suhorukov thanks for your contributions. One small comment to check our guidelines for formatting commit messages on the wiki https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#submit-a-pull-request.

PR #1694, opened 2018-02-19
Return correct return type for Kotlin suspending functions in MethodParameter.

konrad-kaminski:
Return type for Kotlin suspending functions (as returned by `MethodParameter#getParameterType` and `MethodParameter#getgenericReturnType` methods) is incorrect. The true return type is the generic parameter of the last parameter of the method.

This change modifies the behaviour of the aforementioned methods so that they work correctly for all cases.

Issue: [SPR-16515](https://jira.spring.io/browse/SPR-16515)

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -25,6 +25,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -396,7 +397,9 @@ public class MethodParameter {
 		if (paramType == null) {
 			if (this.parameterIndex < 0) {
 				Method method = getMethod();
-				paramType = (method != null ? method.getReturnType() : void.class);
+				paramType = (method != null ?
+                        (KotlinDetector.isKotlinType(getContainingClass()) ?
+								KotlinDelegate.getReturnType(method) : method.getReturnType()) : void.class);
 			}
 			else {
 				paramType = this.executable.getParameterTypes()[this.parameterIndex];
@@ -416,7 +419,9 @@ public class MethodParameter {
 		if (paramType == null) {
 			if (this.parameterIndex < 0) {
 				Method method = getMethod();
-				paramType = (method != null ? method.getGenericReturnType() : void.class);
+				paramType = (method != null ?
+						(KotlinDetector.isKotlinType(getContainingClass()) ?
+								KotlinDelegate.getGenericReturnType(method) : method.getGenericReturnType()) : void.class);
 			}
 			else {
 				Type[] genericParameterTypes = this.executable.getGenericParameterTypes();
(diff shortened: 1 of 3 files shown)


spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -759,7 +765,47 @@ else if (ctor != null) {
...
+                return method.getGenericReturnType();
+            }
+		}
+
+		static private boolean isSuspend(Method method) {
+            KFunction<?> function = ReflectJvmMapping.getKotlinFunction(method);
+            return (function != null && function.isSuspend());
+        }
+
+        @NonNull
```
sdeleuze: Not necessary, already the default via package level `@NonNullApi`.

spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -759,7 +765,47 @@ else if (ctor != null) {
...
 			return false;
 		}
-		
-	}
 
+		/**
+		 * Returns a return type of a method using Kotlin Reflection API.
+		 * Introduced to support suspending functions.
+		 */
+		@NonNull
```
sdeleuze: Not necessary, already the default via package level `@NonNullApi`.

spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -759,7 +765,47 @@ else if (ctor != null) {
...
+		static private Class<?> getReturnType(Method method) {
+		    if (isSuspend(method)) {
+		        return getSuspendReturnType(method).resolve();
```
sdeleuze: `resolve()` can return `null` but `Class<?> getReturnType(Method method)` is expected to return non null values. Maybe we should use `resolve(Class<?> fallback)` variant or assert the returned class is not null before returning it if we are sure that will be the case here?

sdeleuze (changes requested):

spring-core/src/main/java/org/springframework/core/MethodParameter.java
konrad-kaminski: Fixed.

spring-core/src/main/java/org/springframework/core/MethodParameter.java
konrad-kaminski: Fixed.

spring-core/src/main/java/org/springframework/core/MethodParameter.java
konrad-kaminski: Assertion added. This should never be `null`.

sdeleuze:
After reviewing more in detail this PR, I am not sure yet Coroutines should be managed at `MethodParameter` level. I was initially tempted to refactor it to manage a generalized version of what we support for `Optional`, partially because we have [SPR-16530](https://jira.spring.io/browse/SPR-16530) which requires a more generalized support of wrapper types like `Option` / `Try` / `Either` / `Lazy`. We could for example imagine a `MethodParameter#nestedIfWrapped` that would unwrap these wrappers including Coroutine `Continuation` by reusing the logic you implemented @konrad-kaminski.

But I am not sure this is the right way to support that, because Coroutines are more high-level than just a wrapper. As described [here](https://github.com/Kotlin/kotlin-coroutines/blob/master/kotlin-coroutines-informal.md#continuation-passing-style), `suspend fun <T> foo(): T` becomes `fun <T> foo(continuation: Continuation<T>): Any?`. The type information moves from the return values to the last parameter generic nested type so IMO hardcoding a clever recognition of suspending function return type is not necessarily what we always want and could lead to some unexpected behavior.

So what are the alternatives? 

Coroutines are more related to `Mono` than to `Optional` so I thought it could maybe be a good idea to use `ReactiveAdapter` to deal with Coroutines types but an issue with current API is that `ReactiveTypeDescriptor` deals with `Class`, not with `ResolvableType` which could allow us to find the potentially suspending function via `ResolvableType#getSource`. So one option could maybe be to update `ReactiveTypeDescriptor` to deal with `ResolvableType` instead of `Class` (with deprecation in 5.1 of `Class` based API). This could allow us to find the right return type of suspending functions. Another important pro for using `ReactiveAdapter` here is that it could be maybe used to implement Coroutines -> `Mono` conversion in a more cleanly designed way than what is done in [CustomControllerMethodResolver](https://github.com/konrad-kaminski/spring-kotlin-coroutine/blob/f3af119cd0cc19d03b210b5e1bd727e75c0a2c56/spring-webflux-kotlin-coroutine/src/main/kotlin/org/springframework/web/reactive/result/method/annotation/CustomControllerMethodResolver.kt). This could maybe be a fix for [SPR-15413](https://jira.spring.io/browse/SPR-15413), reused in Spring Data, etc.

Any thoughts @jhoeller @rstoyanchev @konrad-kaminski ?

sdeleuze:
Merged via [this revised commit](https://github.com/spring-projects/spring-framework/commit/9302cb2f85fbde03f85da841d694edb10064fc27) which leverages Kotlin reflection that now supports Coroutines. Thanks @konrad-kaminski for this contribution.

PR #1696, opened 2018-02-20
Handle end of stream when reading the request body

violetagg:
When reading request body -1 means all data read.
Instead of waiting for an event from the server, call
onAllDataRead immediately.


rstoyanchev:
https://jira.spring.io/browse/SPR-16521

rstoyanchev:
I've handled this slightly differently, using a constant instead of raising an exception.

PR #1698, opened 2018-02-23
Polish: Replace this "Map.get()" and condition with a call to "Map.computeIfAbsent()"

andrey916:
Polish: Replace this "Map.get()" and condition with a call to "Map.computeIfAbsent()"


jhoeller:
I'm keeping this one on hold for the time being: While the `computeIfAbsent` syntax is certainly appealing, it comes with different locking behavior, in particular regarding our double-checked locking approach with a quick `ConcurrentHashMap` check first and a second attempt within full synchronization. 

On review, I did uncover two minor oddities in the affected code which I've revised and prepared for backporting to 4.3.x. Even if we're not merging such a pull request immediately, I really appreciate the opportunity to review those code spots! I'll revisit the `computeIfAbsent` part for 5.1 a bit later.

andrey916:
@jhoeller, thank you for taking your time for the review

jhoeller:
A quick update: I've applied the use of computeIfAbsent for the lazy creation of nested collection structures in places where this fits with our locking assumptions, therefore selectively addressing a part of your pull request. This will make it into 5.0.5 for a start; thanks for helping to identify those spots!

andrey916:
Always ready to help :-)

PR #1699, opened 2018-02-23
Take brotli encoding into account for caching

burka:
We use a BrotliResourceResolver to deliver pre compressed content. If we use the `CachingResourceResolver` including `ResourceHandlerRegistry.resourceChain(true)` we cache brotli compressed content and deliver it to only gzip aware clients.  This fix resolves that by taking the brotli encoding into account. As all brotli aware browsers most likely supports gzip I think those two ifs are fine :-)


rstoyanchev:
https://jira.spring.io/browse/SPR-16381

PR #1705, opened 2018-02-27
Polish: combine catches block with same body

igor-suhorukov:


jhoeller:
In a few of those cases, there were separate comments applying per exception type. Since those are somewhat outdated anyway, I'll revise them in a separate pass: within the combined catch block, simply differentiating the exception types in the wording of the comment.

Oh, and once again thanks for all your polish contributions :-) Perfect timing there since we're just a few weeks away from our 5.1 vs 5.0 split, so any contributions coming in at this point still fine-tune the 5.0 codebase before branching... and they'll make it into the Boot 2.0.1 release.

igor-suhorukov:
@jhoeller thank you! I hope that all this changes are help spring-framework project and reduce support costs.

Also please look at [PR#1698](https://github.com/spring-projects/spring-framework/pull/1698) - it will make old part of code more java8 friendly

PR #1706, opened 2018-02-27
SPR-16519 - Adding new acknowledge method version

mariusz-jasinski:
The change according to Rossen Stoyanchev' comment

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/DefaultStompSession.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/DefaultStompSession.java
@@ -328,12 +328,19 @@ public class DefaultStompSession implements ConnectionHandlingStompSession {
 			stompHeaders.setId(messageId);
 		}
 
-		String receiptId = checkOrAddReceipt(stompHeaders);
+		Receiptable receiptable = acknowledge(stompHeaders, consumed);
+
+		return receiptable;
+	}
+
+	@Override
+	public Receiptable acknowledge(StompHeaders headers, boolean consumed) {
+		String receiptId = checkOrAddReceipt(headers);
 		Receiptable receiptable = new ReceiptHandler(receiptId);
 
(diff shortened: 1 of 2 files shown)


spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/DefaultStompSession.java
```diff
@@ -328,12 +328,19 @@ public Receiptable acknowledge(String messageId, boolean consumed) {
 			stompHeaders.setId(messageId);
 		}
 
-		String receiptId = checkOrAddReceipt(stompHeaders);
+		Receiptable receiptable = acknowledge(stompHeaders, consumed);
```
jhoeller: This looks a case for a straight `return acknowledge(stompHeaders, consumed)` call.

On a related note, @rstoyanchev, there seems to be inconsistent naming of `StompHeaders` arguments, primarily between `StompSession` and implementing methods in `DefaultStompSession`, but also within the `StompSession` interface itself regarding the new `unsubscribe(StompHeaders)` variant. Let's consistently name those arguments `headers`, replacing all current `stompHeaders` names.

spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompSession.java
```diff
@@ -101,6 +101,19 @@
...
+	/**
+	 * Send an acknowledgement whether a message was consumed or not resulting
+	 * in an ACK or NACK frame respectively.
+	 * <p><strong>Note:</strong> to use this when subscribing you must set the
+	 * {@link StompHeaders#setAck(String) ack} header to "client" or
+	 * "client-individual" in order ot use this.
+	 * @param headers the headers for the ACK or NACK frame
+	 * @param consumed whether the message was consumed or not
+	 * @return a Receiptable for tracking receipts
+	 * @since 5.1
```
jhoeller: Let's roll this into 5.0.5, changing this `since` tag accordingly.

rstoyanchev:
This is now merged, also taking the comments into account.

PR #1708, opened 2018-02-28
Polish: extract nested ternary operation into an independent statement

igor-suhorukov:


jhoeller:
I actually prefer it the way we have it within inline expressions: It's more concise and retains the lazy evaluation characteristics of the nested right-hand side. If we'd untangle this, then probably rather to a mix of `if` and ternary with the same lazy evaluation behavior. That said, even with that in mind, I prefer the current variant for conciseness.

PR #1711, opened 2018-03-01
SPR-12756: Local visibility for writes to caches with transactional decorators

williamhoyle:
This patch adds an alternative to the TransactionAwareCacheDecorator that addresses issues raised in SPR-12756 - it makes writes to decorated caches visible within the scope of the transaction.

It is intended to be minimally obtrusive. The current interface is not changed beyond adding a `setCacheDecoratorFactory` method to the `TransactionAwareCacheManagerProxy` to allow the strategy to be selected. The behaviour defaults to use the existing `TransactionAwareCacheDecorator`.

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/transaction/BufferedCache.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/transaction/BufferedCache.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


githcheng:
嗯嗯

发自我的 iPhone

neiser:
@williamhoyle I found this PR while starting from this [StackOverflow question](https://stackoverflow.com/q/63034892) asking why no CacheErrorHandler is called when transaction support is enabled for caching. This is IMHO a shortcoming if not a bug in Spring's way of "transaction-aware caching". 

Now I found also this little project: https://github.com/ethlo/spring-tx-cache-decorator by @ethlo. Have you been aware of this project when you implemented this PR?

Do you think this PR could be extended to cover also the missing CacheErrorHandler call?

snicoll:
Looking at the history of this issue. It looks like that https://github.com/ethlo/spring-tx-cache-decorator could supersede this unless we decide we want to support that in the core framework and handle the additional complexity.

neiser:
@snicoll I don't see directly how #16574 is related to this issue, but if transaction-aware  caches are part of the Spring framework or not should really be decided as right now it is not consistently implemented. So either fix or remove 😀

snicoll:
Reviewing the history of this issue, I am not keen to bring that extra complexity and layering in the cache abstraction, so I am going to close this now. The discussion in #17353 provides additional context and justification. Thanks for the PR, in any case!

PR #1712, opened 2018-03-01
Polish: deprecated elements should have both the annotation and the Javadoc tag

igor-suhorukov:


jhoeller:
Good catch. I've addressed those few cases individually through explicit deprecation messages and in one case through removing the deprecation marker (`HibernateTemplate.prepareQuery`).

PR #1713, opened 2018-03-01
Polish: Overriding methods should do more than simply call the same method in the super class

igor-suhorukov:
Remove methods with the same visibility and without comments about workaround.


snicoll:
@jhoeller @igor-suhorukov that change to `MediaType` is a breaking change as those two methods are actually overriding the parameter type. I've reverted this in 38ebb6c

jhoeller:
Well spotted. I only checked for covariant return types but totally forgot about the parameter type narrowing which effectively makes it an overloaded method from a binary compatibility perspective.

PR #1716, opened 2018-03-04
Faster search for characters in string for deletion

akikrasic:
Using HashSet  time for searching every character in String charsToDelete  in StringUtils.deleteAny method, can be saved.


dreis2211:
@akikrasic I hope you don't mind my comment, but could you share a benchmark showing the improvement? Keep in mind that I'm not associated in any way to Pivotal, so my comment has no weight whatsoever.

I'm just wondering under which circumstances this will improve performance as you create a new HashSet that comes with its own allocation cost.  E.g. with the following input values:
```

snicoll:
@akikrasic I've reviewed the code and was wondering the same (thank you so much Christoph for the benchmark!). Unless you have a competitive argument (or something to say about that benchmark), I am inclined to decline this PR. 

Thoughts?

PR #1717, opened 2018-03-06
SPR-16556 - Incorrect isGetGeneratedKeysSimulated() return value logic in PostgresTableMetaDataProvider.java

anguslee:
Do a comparison by major and minor numbers instead of lexically


jhoeller:
As mentioned on JIRA, we can streamline this towards always returning true from isGetGeneratedKeysSimulated since we don't support any Postgres version older than 8.2 to begin with...

PR #1718, opened 2018-03-06
Polish: local variable and method parameter names should comply with a naming convention

igor-suhorukov:


jhoeller:
We generally leave our repackaged ASM fork in line with the original ASM source, accepting mismatches with our conventions there. I need to look at that `PropertyMatches` algorithm, this could be more clearly named in general (i.e. the main problem is not the underscore). Thanks for pointing at those, in any case!

igor-suhorukov:
@jhoeller thank you for details

PR #1719, opened 2018-03-06
Polish: constant names should comply with a naming convention

igor-suhorukov:


jhoeller:
Good point in general... but those are actually intentional: We differentiate between classic constants (upper case) and shared startup-computed state like classpath presence checks (lower case like instance fields). In other words, a `static final` field may be a constant with upper-case naming but doesn't have to be.

igor-suhorukov:
Ok, I understand this approach

PR #1720, opened 2018-03-07
Make MockRestRequestMatchers independent of JUnit

yankee42:
The MockRestRequestMatchers was using `org.junit.Assert.assertNotNull` and thus it could not be used in projects that use e.g. TestNG instead of JUnit, because this causes a `java.lang.NoClassDefFoundError: org/junit/Assert` at runtime.

This PR removes that dependency.


sbrannen:
Thanks for raising this issue!

I'm looking into it right now.

sbrannen:

Thanks!

sbrannen:
Note, however, that your change broke the build: the affected methods were expected to throw an `AssertionError` instead of an `IllegalArgumentException`.

I fixed that in commit fef0c7b15193e61a19302554ec3c421b64af7b7f.

PR #1722, opened 2018-03-07
Invert the check as ServletInputStream.isReady() is the active method

violetagg:
When checking whether there is still request body the first method
that should be checked is ServletInputStream.isReady() and then
ServletInputStream.isFinished(). ServletInputStream.isReady() is the active
method whereas the ServletInputStream.isFinished() is not.
It is important to call ServletInputStream.isReady() because if it returns
false it will schedule a dispatch and if the request body is already read it will
send onAllDataRead event.

Issue SPR-16521


rstoyanchev:
This is now merged, thanks!

PR #1724, opened 2018-03-08
Polish: collapsible "if" statements should be merged

igor-suhorukov:


rstoyanchev:
hi @igor-suhorukov, we have guidelines for line wrapping [described here](https://github.com/spring-projects/spring-framework/wiki/Code-Style#line-wrapping) and some of the changes here violate those rules. Certainly the lines over 120. 

As for the overall change, personally I am not sure all of the changes in this commit make the code more readable. In some cases I prefer a top-level condition that makes it easier to understand some pre-condition and skip. That helps when reading or debugging. Then there are some changes here that do make the look more readable. So it might be a case by case basis. I'll defer to @jhoeller.

jhoeller:
Well spotted, and certainly worth reviewing... However, in quite a few of those cases, the nested `if` statements were rather intentional: following the same code structure as other blocks in the same method, consistently just checking a primary condition in the uppermost statement, allowing for comments on nested conditions, or always keeping logger enable checks within the block. For that reason, I went through all those spots and manually collapsed the checks in about half of them, also applying some further refinements to the code around them. Thanks for the pull request, in any case!

igor-suhorukov:
@jhoeller thank you for feedback

PR #1725, opened 2018-03-08
Fix issue: getters and setters should be synchronized in pairs

igor-suhorukov:
When one part of a getter/setter pair is synchronized the other part should be too. Failure to synchronize both sides of a pair may result in inconsistent behaviour at runtime as callers access an inconsistent method state.


jhoeller:
Again well spotted! Thanks for pointing the finger at all of those... You're keeping me quite busy there :-)

I've addressed this one differently in avoiding full synchronization as far as possible: The underlying fields are volatile now, and the getters use a double-check lock, only entering full synchronization when no existing instance has been found. All we want to prevent there is entering the lazy initialization algorithm concurrently; once initialized, we're happy to go with whatever has been set, so a volatile field is sufficient there. We're using such a pattern in quite a few other places since it provides a reasonable tradeoff for all kinds of lazy default initialization.

igor-suhorukov:
@jhoeller thanks, I understand

PR #1727, opened 2018-03-08
Polish: inner class calls to super class methods should be unambiguous

igor-suhorukov:
When an inner class extends another class, and both its outer class and its parent class have a method with the same name, calls to that method can be confusing. The compiler will resolve the call to the superclass method, but maintainers may be confused, so the superclass method should be called explicitly, using super.


jhoeller:
It's a fair point, but frankly, I don't find that extra qualification important. I'm more concerned about static imports messing up the understanding of method calls than such inner vs super cases. Also, an IDE will point maintainers to the super method there anyway. All in all, I'd rather go with streamlined calls without the rather noisy prefix.

PR #1728, opened 2018-03-08
Polish: to get rid of "magic" time constants

igor-suhorukov:


jhoeller:
I like the idea of using `TimeUnit` that way...

spring-context-support/src/main/java/org/springframework/scheduling/quartz/SchedulerFactoryBean.java
```diff
@@ -697,7 +698,7 @@ protected void startScheduler(final Scheduler scheduler, final int startupDelay)
 				@Override
 				public void run() {
 					try {
-						Thread.sleep(startupDelay * 1000);
+						Thread.sleep(TimeUnit.SECONDS.toMillis(startupDelay));
```
rajadilipkolli: I would prefer to use TimeUnit.MILLISECONDS.sleep(startupDelay); If you are ok with this suggestion I will submit PR

igor-suhorukov: @rajadilipkolli startupDelay in seconds, sorry!

rajadilipkolli: @igor-suhorukov Updated comment and this should be fine now.

PR #1729, opened 2018-03-08
Polish: use StringBuilder for complex string concatenation

igor-suhorukov:


jhoeller:
While I'm generally concerned about overuse of `StringBuilder` since the Java compiler is pretty good at the code it generates for regular String concatenation, I agree with those two use cases.

PR #1732, opened 2018-03-08
Polish: parentheses should be removed from a single lambda input parameter when its type is inferred

igor-suhorukov:


snicoll:
@jhoeller for the record we always put the parenthesis there in Spring Boot. The rationale is that when there is more than one argument, the parentheses are mandatory so we always add them for consistency.

PR #1737, opened 2018-03-09
Polish: remove this unused private method parameters

igor-suhorukov:


igor-suhorukov:
Hello @jhoeller 
What do you think about this code cleanup?

jhoeller:
Those are rather specific cases where we retain currently unused parameters for consistency with related signatures or as a target for a particular functional interface which happens to declare more input parameters than needed for the case at hand. Since it doesn't hurt to declare those extra parameters for guidance purposes (in case a future revision of that code might benefit from that extra context) and there's no negative performance effect as far as I'm aware, I'm inclined to leave those spots as-is.

igor-suhorukov:
@jhoeller thank you for review

PR #1738, opened 2018-03-09
 Polish "Use Java 8 forEach method on Map"

Patrick0308:


jcs090218 (commented):
Would this be necessary? Consider adding parameter infront of the parameter is a bit wordy and I do not think is good to have a longer argument name in this case.

snicoll:
Thanks for the PR but there is nothing wrong with the way those variables are named at the moment IMO.

PR #1739, opened 2018-03-10
SPR-16578 - queryForMap returns last value

marschall:
queryForMap returns last value returns the last value of equally named
columns. The JDBC specification says that in such cases the first value
should be returned instead.

 - change ColumnMapRowMapper to not overwrite existing values

Issue: SPR-16578


jhoeller:
As mentioned on SPR-16578, this is in master as a slightly different implementation now. Thanks for the pull request, in any case!

PR #1740, opened 2018-03-14
SPR-16590: getParameterValues would work incorrectly

jasonjoo2010:
Affected method:

**DefaultMultipartHttpServletRequest::getParameterValues**
**DefaultMultipartHttpServletRequest::getParameterMap**

getParameterValues is used to return multi values of a same key in request, including query string, post contents.
And DefaultMultipartHttpServletRequest would be used instead of default servlet wrapper when using "**multipart/form-data**" post method to realize including vars in multipart body. And there is a problem in its getter in dealing values of same keys. It will not merge them to values in  query string.

It can be reproduced by following steps:
```
curl -F "key1=test" "http://test.com/hello?key1=other"
```
You will get a parameter map like:
```
key1:
	test
```

```
curl -d "key1=test" "http://test.com/hello?key1=other"
```
And you will get:
```
key1:
	test
	other
```

Let's sit down to discuss whether fix it?

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/multipart/support/DefaultMultipartHttpServletRequest.java
+++ b/spring-web/src/main/java/org/springframework/web/multipart/support/DefaultMultipartHttpServletRequest.java
@@ -16,11 +16,13 @@
 
 package org.springframework.web.multipart.support;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@jasonjoo2010 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1740)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1740) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jasonjoo2010 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1740)!

spring-web/src/main/java/org/springframework/web/multipart/support/DefaultMultipartHttpServletRequest.java
```diff
@@ -87,14 +89,28 @@ public String getParameter(String name) {
 		}
 		return super.getParameter(name);
 	}
+	
+	public String[] mergeTwoParameterValues(String[] val1, String[] val2) {
```
rstoyanchev: Does not need to be public, nor instance method for that matter.

rstoyanchev: Need not be a public method.

spring-web/src/main/java/org/springframework/web/multipart/support/DefaultMultipartHttpServletRequest.java
```diff
@@ -87,14 +89,28 @@ public String getParameter(String name) {
...
+		if (val1 == null) {
+			return val2;
+		}
+		if (val2 == null) {
+			return val1;
+		}
+		//merge them together before return
+		String[] merged = Arrays.copyOf(val2, val1.length + val2.length);
```
rstoyanchev: It's counter-intuitive for a method that merges to arrays, to put the second one first.

rstoyanchev (commented):
The PR is incomplete without any tests. Just mentioning it in case you want to make it complete.

spring-web/src/main/java/org/springframework/web/multipart/support/DefaultMultipartHttpServletRequest.java
jasonjoo2010: So i fix them and add getParameterValues test in CommonsMultipartResolverTests.

But i used forced pushing to avoid unnecessary commits and found i cannot reopen this.

Should i open another PR?

rstoyanchev: Force push is fine. Just keep the branch updated, no need to re-create the PR.

PR #1741, opened 2018-03-16
Support Kotlin parameter default values in handler methods

diesieben07:
Allows using default parameter values for any method parameters resolved
via `AbstractNamedValueMethodArgumentResolver`, similar to `defaultValue`
parameter of e.g. `@RequestParam` annotation, but skipping the converter
chain.

Issues: [SPR-16598](https://jira.spring.io/browse/SPR-16598).

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -343,6 +343,14 @@ public class MethodParameter {
 	}
 
 	/**
+	 * Return whether this parameter declares a language-level default value,
+	 * such as a default parameter in Kotlin.
+	 */
+	public boolean hasDefaultValue() {
+		return KotlinDetector.isKotlinType(getContainingClass()) && KotlinDelegate.hasDefaultValue(this);
+	}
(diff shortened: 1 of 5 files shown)


raderio:
An ETA when it can be merged?

diesieben07:
Rebased onto current master and resolved merge conflicts.

raderio:
An ETA when it can be merged?

raderio:
is something wrong with this pull request?

snicoll:
@raderio please be patient, we'll get to it as soon as we can.

sdeleuze:
Current Kotlin API forces us to have a rather involved implementation which is not a good fit with our infra. We would need a reflection API that allows Spring to know an optional parameter default value without calling the function. I will discuss with Kotlin team to see what is possible here in Kotlin 1.3.

diesieben07:
I am not sure if what you want is possible.
You can write something like this:
`fun foo(a: String, b: String = a)` or, even worse:
`fun foo(a: String, b: String = someFunctionWithSideEffects())`

This is impossible to represent in the way you propose.

PR #1749, opened 2018-03-22
FactoryBeanRegistrySupport atomicity issues

lmagyar89:
Issue: SPR-16625


jhoeller:
Addressed in a slightly revised and extended fashion, as mentioned on JIRA. Thanks for the PR, in any case!

PR #1751, opened 2018-03-23
SPR-16628 - Improve NoSuchBeanDefinitionException to include a reference to the dependent bean

andersonkyle:
[SPR-16628](https://jira.spring.io/browse/SPR-16628)

This PR allows a `NoSuchBeanDefinitionException` to refer to the dependent bean which improves clarity and traceability.

Additionally this information will be useful in a Spring Boot Failure Analyzer.


jhoeller:
Resolved in a different way in a style consistent with other missing dependency scenarios, as mentioned in JIRA. Thanks for the PR, in any case!

PR #1753, opened 2018-03-26
SPR-16642 - DefaultMessageListenerContainer JMS Pool leak with CACHE_CONSUMER and Bitronix in error path

drewtul:
f the error handling path in the DefaultMessageListenerContainer the JMS connection is closed and re-opened.

Combining this with CACHE_CONSUMER with bitronix results in an error from DualSessionWrapper due to the Connection being involved in multiple XA transactions during the close attempt, which bitronix rejects and DMLC logs, swallows and continues.

This results in the Connection not being closed and a new Connection obtained from the bitronix pool resulting in a gradual leak from the pool.

Utilising dedicated connections alleviates this and enables CACHE_CONSUMER to function without issue. So making this a configuration option to not use the shared connection whilst CACHE_* is enabled means we can avoid this path and only have a single thread using a dedicated connection avoiding the multiple active transaction close issue.


drewtul:
Closing for more local testing.

PR #1756, opened 2018-03-28
Polish: "equals(Object obj)" and "hashCode()" should be overridden in pairs

igor-suhorukov:
According to the Java Language Specification, there is a contract between equals(Object) and hashCode()


jhoeller:
In those three cases, we inherit a custom hashCode implementation from the base class while the overridden equals method just narrows the notion of equality within its superclass domain a bit. Since hashCode is only required to return the same value if equals is true but not to return distinct values for non-equal scenarios, this is ultimately ok... analogous to a local equals/hashCode pair where the hashCode only calculates against a key part of the state whereas equals itself compares every single field.

igor-suhorukov:
@jhoeller thank you for the clarification

PR #1759, opened 2018-03-29
Polish: nested "enum"s should not be declared static

igor-suhorukov:
According to [the docs](https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.9):  Nested enum types are implicitly static.


nkjackzhang:
According to the jls, `It is permissible to explicitly declare a nested enum type to be static. ` It's just a code style issue.

snicoll:
@nkjackzhang that is indeed a code style issue that this PR has fixed.

PR #1760, opened 2018-03-29
Polish: remove unreachable statement

igor-suhorukov:
Branch can not be reached because the condition duplicates a previous condition in the same sequence of "if/else if" statements from [line 109](https://github.com/igor-suhorukov/spring-framework/blob/d2f9090766a3f4f92266de7a0552b23ae1a05f60/spring-expression/src/main/java/org/springframework/expression/spel/ast/TypeReference.java#L109)


jhoeller:
Well spotted. I've revised several primitive type checks in SpEL for more consistency, removing this duplicate clause along the way.

PR #1761, opened 2018-03-29
Optimize ClassUtils.forName() for more cases

dreis2211:
Issue: https://jira.spring.io/browse/SPR-16667

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ClassUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ClassUtils.java
@@ -32,7 +32,9 @@ import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/util/ClassUtils.java
```diff
@@ -138,6 +140,9 @@
 				Object.class, Object[].class, Class.class, Class[].class);
 		registerCommonClasses(Throwable.class, Exception.class, RuntimeException.class,
 				Error.class, StackTraceElement.class, StackTraceElement[].class);
+		registerCommonClasses(Enum.class, Optional.class);
+		registerCommonClasses(Collection.class, List.class, Map.class, Set.class);
```
jhoeller: I would go even further here and register Iterable and Iterator as well (collapsed and reordered):

```
registerCommonClasses(Collection.class, List.class, Set.class, Map.class,
        Iterable.class, Iterator.class, Enum.class, Optional.class);
```

jhoeller:
I've rolled in a slightly revised variant, co-initializing the `javaLanguageInterfaces` set. Thanks for the PR!

PR #1769, opened 2018-03-30
Fix the wrong link typo

linzhaoming:
From `mvc-special-bean-types` to `mvc-servlet-special-bean-types`


jhoeller:
Well spotted! Being at it, I've fixed that link along with a wider revision of the surrounding chapters.

PR #1771, opened 2018-04-01
The bracket is dissymmetry, fix it

linzhaoming:


jhoeller:
Well spotted. There is even more at odds in that javadoc, actually: for example, `ManagedExecutorService` has only been introduced in Java EE *7*, not 6. I've removed that bracket in a corresponding wider revision. Thanks for the PR, in any case!

PR #1772, opened 2018-04-02
Add cpp to mime.types

nkjackzhang:
In most cases, the extend name of "c++" file is`cpp`, not `c++`.


jhoeller:
Good catch! I've added the cpp extension as part of a slighly wider revision of that definition file. Thanks for the PR, in any case!

PR #1774, opened 2018-04-02
MockHttpServletRequest return the requested Content-Type.

ishiis:
This PR updates MockHttpServletRequest does not overwrite Content-Type.

Issues: [SPR-16679](https://jira.spring.io/browse/SPR-16679)


nkjackzhang:
If you create `mockMvc` with `defaultRequest(get("/").characterEncoding("UTF-8"))`, the `defaultRequest` will be merged into the `MockHttpServletRequest` created in the `@Test` method, of course, the header will contain `charset=UTF-8`. May be that's why your `@Test` method print wrong message, you can check the `mockMvc` in your test class.

PR #1775, opened 2018-04-04
Fix '<=' symbol in SpEL documentation

fabriziocucci:
This fix is already on the `master`.


snicoll:
Thanks for the PR @fabriziocucci 

@rstoyanchev perhaps we should consider backporting https://github.com/spring-projects/spring-framework/commit/8f525f07e9dd6a34463d43546c22cb95afd0429b instead?

snicoll:
@fabriziocucci oops, this one felt trough the cracks, sorry. The doc in supported versions doesn't have this issue.

PR #1778, opened 2018-04-05
SPR-16642 - DefaultMessageListenerContainer JMS Pool leak with CACHE_CONSUMER and Bitronix in error path

drewtul:
Add a boolean flag to control the use of a dedicated connection for caching.

This enables using a dedicated connection per task which may be required
for some XA Transaction managers to avoid connection closing issues
in the error handling path resulting in a leak from a connection pool.

This has been seen with CACHE_CONSUMER and btm at least.


snicoll:
@drewtul we need a Jira issue to discuss this change. Can you please [create  one](https://jira.spring.io/browse/SPR)?

snicoll:
The related issue (#21183) is in the 6.x backlog but we haven't really figure out how to approach this. I don't know if this PR is still accurate.

snicoll:
With Bitronix being EOL now, I think we need to continue this convo on the issue.

PR #1783, opened 2018-04-10
Add support for Jetty Reactive Streams HTTP client

sdeleuze:
Ready to be reviewed before merging in `master` since I would like to begin having feedback with real applications using Spring Framework snapshots.

Notice that this PR is implemented with buffer copy instead of optimized buffer wrapping because the latter hangs since `Callback#succeeded` doesn't allow releasing the buffer and requesting more data at different times (required for `Mono<DataBuffer>` for example).

Simone is aware of that issue and will see if it can be changed on Jetty side (not sure yet). That could be changed with an additional commit if that becomes possible in the future.

Issue: [SPR-15092](https://jira.spring.io/browse/SPR-15092)

### DIFF
--- a/spring-web/spring-web.gradle
+++ b/spring-web/spring-web.gradle
@@ -29,6 +29,7 @@ dependencies {
 	optional("io.netty:netty-all")
 	optional("io.projectreactor.netty:reactor-netty")
 	optional("org.apache.tomcat.embed:tomcat-embed-core:${tomcatVersion}")
+	optional("org.eclipse.jetty:jetty-reactive-httpclient:1.0.0")
 	optional("org.eclipse.jetty:jetty-server:${jettyVersion}") {
 		exclude group: "javax.servlet", module: "javax.servlet-api"
 	}
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpConnector.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpConnector.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.http.client.reactive;
+
+import java.net.URI;
+import java.util.function.Function;
+
+import org.eclipse.jetty.client.HttpClient;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import org.springframework.context.SmartLifecycle;
+import org.springframework.core.io.buffer.DataBuffer;
+import org.springframework.core.io.buffer.DataBufferFactory;
+import org.springframework.core.io.buffer.DefaultDataBufferFactory;
+import org.springframework.http.HttpMethod;
+
+/**
+ * Jetty ReactiveStreams HttpClient implementation of {@link ClientHttpConnector}.
+ *
+ * @author Sebastien Deleuze
+ * @since 5.1
+ */
+public class JettyClientHttpConnector implements ClientHttpConnector, SmartLifecycle {
+
+	private final HttpClient httpClient;
+
+	private DataBufferFactory bufferFactory = new DefaultDataBufferFactory();
+
+
+	/**
+	 * Create a Jetty {@link ClientHttpConnector} with the default {@link HttpClient}.
+	 */
+	public JettyClientHttpConnector() {
+		this(new HttpClient());
+	}
+
+	/**
+	 * Create a Jetty {@link ClientHttpConnector} with the given {@link HttpClient}.
+	 */
+	public JettyClientHttpConnector(HttpClient httpClient) {
+		this.httpClient = httpClient;
+	}
+
+
+	public void setBufferFactory(DataBufferFactory bufferFactory) {
+		this.bufferFactory = bufferFactory;
+	}
+
+	@Override
+	public int getPhase() {
+		return Integer.MAX_VALUE;
+	}
+
+	@Override
+	public boolean isAutoStartup() {
+		return true;
+	}
+
+	@Override
+	public boolean isRunning() {
+		return this.httpClient.isRunning();
+	}
+
+	@Override
+	public void start() {
+		try {
+			// HttpClient is internally synchronized and protected with state checks
+			this.httpClient.start();
+		}
+		catch (Exception ex) {
+			throw new RuntimeException(ex);
+		}
+	}
+
+	@Override
+	public void stop() {
+		try {
+			this.httpClient.stop();}
+		catch (Exception ex) {
+			throw new RuntimeException(ex);
+		}
+	}
+
+	@Override
+	public void stop(Runnable callback) {
+		stop();
+		callback.run();
+	}
+
+	@Override
+	public Mono<ClientHttpResponse> connect(HttpMethod method, URI uri,
+			Function<? super ClientHttpRequest, Mono<Void>> requestCallback) {
+
+		if (!uri.isAbsolute()) {
+			return Mono.error(new IllegalArgumentException("URI is not absolute: " + uri));
+		}
+
+		if (!httpClient.isStarted()) {
+			try {
+				this.httpClient.start();
+			}
+			catch (Exception ex) {
+				return Mono.error(ex);
+			}
+		}
+
+		JettyClientHttpRequest clientHttpRequest = new JettyClientHttpRequest(
+				httpClient.newRequest(uri).method(method.toString()), bufferFactory);
+		return requestCallback.apply(clientHttpRequest).then(
+		Mono.from(clientHttpRequest.getReactiveRequest().response((reactiveResponse, contentChunks) -> {
+					// Implementation with buffer copy instead of optimized buffer wrapping
+					// because the latter hangs since Callback#succeeded doesn't allow
+					// releasing the buffer and requesting more data at different times
+					// (required for Mono<DataBuffer> for example).
+					// See https://github.com/eclipse/jetty.project/issues/2429
+					Flux<DataBuffer> content = Flux.from(contentChunks).map(chunk -> {
+						DataBuffer buffer = this.bufferFactory.allocateBuffer(chunk.buffer.capacity());
+						buffer.write(chunk.buffer);
+						chunk.callback.succeeded();
+						return buffer;
+					});
+
+					return Mono.just(new JettyClientHttpResponse(reactiveResponse, content));
+				})));
+	}
+
+}
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.http.client.reactive;
+
+import java.net.HttpCookie;
+import java.net.URI;
+import java.util.Collection;
+import java.util.function.Function;
+
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.reactive.client.ContentChunk;
+import org.eclipse.jetty.reactive.client.ReactiveRequest;
+import org.eclipse.jetty.util.Callback;
+import org.reactivestreams.Publisher;
+import reactor.core.Exceptions;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import org.springframework.core.io.buffer.DataBuffer;
+import org.springframework.core.io.buffer.DataBufferFactory;
+import org.springframework.core.io.buffer.DataBufferUtils;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.MediaType;
+import org.springframework.util.Assert;
+
+/**
+ * {@link ClientHttpRequest} implementation for the Jetty ReactiveStreams HTTP client.
+ *
+ * @author Sebastien Deleuze
+ * @since 5.1
+ */
+class JettyClientHttpRequest extends AbstractClientHttpRequest {
+
+	private final Request jettyRequest;
+
+	private final DataBufferFactory bufferFactory;
+
+	private ReactiveRequest reactiveRequest;
+
+
+	public JettyClientHttpRequest(Request jettyRequest, DataBufferFactory bufferFactory) {
+		this.jettyRequest = jettyRequest;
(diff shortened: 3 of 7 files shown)


spring-web/spring-web.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
 	optional("io.netty:netty-all")
 	optional("io.projectreactor.ipc:reactor-netty")
 	optional("org.apache.tomcat.embed:tomcat-embed-core:${tomcatVersion}")
+	optional("org.eclipse.jetty:jetty-reactive-httpclient:0.9.2")
```
sbordet: Version 1.0.0 with RS TCK implemented is available in Maven Central.

sdeleuze: Updated

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+	public DataBufferFactory bufferFactory() {
+		return this.bufferFactory;
+	}
+
+	@Override
+	public Mono<Void> writeWith(Publisher<? extends DataBuffer> publisher) {
+		String contentType = jettyRequest.getHeaders().contains(HttpHeader.CONTENT_TYPE) ?
+				jettyRequest.getHeaders().getField(HttpHeader.CONTENT_TYPE).getValue():
+				MediaType.APPLICATION_OCTET_STREAM_VALUE;
+		Publisher<ContentChunk> chunks = Flux.from(publisher).map(DataBuffer::asByteBuffer).map(ContentChunk::new);
```
poutsma: Even though this class is using `DefaultDataBuffer`, which does not require releasing, I still think it would be a good idea to release them here anyway, especially if we make the data buffer factory configurable.

I think the release can be implemented as the `ContentChunk` callback (I am assuming that the callback would be called after the data has been written, not sure about that). I see that below you said that releasing in the callback does not work for the response, but perhaps it does for the request?

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+		return doCommit(this::completes);
+	}
+
+	@Override
+	public Mono<Void> writeAndFlushWith(Publisher<? extends Publisher<? extends DataBuffer>> body) {
+		String contentType = this.jettyRequest.getHeaders().getField(HttpHeader.CONTENT_TYPE).getValue();
+		ReactiveRequest.Content content = ReactiveRequest.Content.fromPublisher(Flux
+				.from(body)
+				.flatMap(publisher -> publisher)
+				.map(buffer -> new ContentChunk(buffer.asByteBuffer())), contentType);
```
poutsma: Release needs to happen here as well.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+			this.reactiveRequest = ReactiveRequest.newBuilder(this.jettyRequest).build();
+		}
+		return doCommit(this::completes);
+	}
+
+	private Mono<Void> completes() {
+		this.response = Mono.from(
+				this.reactiveRequest.response((reactiveResponse, contentChunks) -> {
+
+					// Implementation with buffer copy instead of optimized buffer wrapping
```
poutsma: Not exactly sure what is meant by this, let's discuss it during our meeting.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+
+
+	public JettyClientHttpRequest(Request jettyRequest) {
+		this.jettyRequest = jettyRequest;
+	}
+
+
+	@Override
+	protected void applyHeaders() {
+		getHeaders().entrySet().forEach(e -> e.getValue().forEach(v -> this.jettyRequest.header(e.getKey(), v)));
```
poutsma: Use a variable instead of calling getHeaders() twice.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+	protected void applyCookies() {
+		getCookies().values().stream().flatMap(Collection::stream)
+				.map(cookie -> new HttpCookie(cookie.getName(), cookie.getValue()))
+				.forEach(this.jettyRequest::cookie);
+	}
+
+	@Override
+	public HttpMethod getMethod() {
+		HttpMethod method = HttpMethod.resolve(this.jettyRequest.getMethod());
+		Assert.notNull(method, "Method must not be null");
```
poutsma: Seems like `Assert.state` is more suitable here, because `notNull` results in an `IllegalArgumentException`.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.util.Assert;
+
+/**
+ * {@link ClientHttpRequest} implementation for the Jetty HTTP client.
+ *
+ * @author Sebastien Deleuze
+ * @since 5.1
+ */
+public class JettyClientHttpRequest extends AbstractClientHttpRequest {
```
poutsma: Could be package-protected instead, no need to expose this class.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpResponse.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+
+/**
+ * {@link ClientHttpResponse} implementation for the Jetty HTTP client.
+ *
+ * @author Sebastien Deleuze
+ * @since 5.1
+ */
+public class JettyClientHttpResponse implements ClientHttpResponse {
```
poutsma: Could be package-protected instead, no need to expose it.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpResponse.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private final ReactiveResponse reactiveResponse;
+
+	private final Flux<DataBuffer> content;
+
+
+	public JettyClientHttpResponse(ReactiveResponse reactiveResponse, Publisher<DataBuffer> content) {
```
poutsma: The `Publisher<DataBuffer>` parameter should be `Flux<DataBuffer>`, since the only place you're creating them is internally, from `JettyClientHttpRequest`.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpResponse.java
```diff
@@ -0,0 +1,88 @@
...
+		Assert.notNull(reactiveResponse, "reactiveResponse should not be null");
+		Assert.notNull(content, "content should not be null");
+		this.reactiveResponse = reactiveResponse;
+		this.content = Flux.from(content);
```
poutsma: See above, drop the `Flux.from`.

poutsma (commented):
Looks good overall. Some things that need to change and some things I need more info on during our meeting.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
sdeleuze: With current Jetty client implementation, we can never transform a `Flux<DataBuffer>` to `Mono<DataBuffer>` because the callback calls `request(1)` + release the buffer. In our case we want to `request(1)` ( to get next buffer) during buffer composition but do the release only when `PooledDataBuffer#release` is called, which is impossible with current API. See the related https://github.com/eclipse/jetty.project/issues/2429 Jetty issue.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpConnector.java
```diff
@@ -0,0 +1,122 @@
...
+
+	public void setBufferFactory(DataBufferFactory bufferFactory) {
+		this.bufferFactory = bufferFactory;
+	}
+
+	@Override
+	public Mono<ClientHttpResponse> connect(HttpMethod method, URI uri,
+			Function<? super ClientHttpRequest, Mono<Void>> requestCallback) {
+
+		if (!httpClient.isStarted()) {
```
rstoyanchev: Really minor, but perhaps delegate to `isRunning()`?

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpConnector.java
```diff
@@ -0,0 +1,122 @@
...
+	@Override
+	public boolean isRunning() {
+		return this.httpClient.isRunning();
+	}
+
+	@Override
+	public int getPhase() {
+		return Integer.MAX_VALUE;
+	}
+}
```
rstoyanchev: For ordering of `SmartLifecycle` methods, in the very least:
- stop(Runnable) should be after stop()
- getPhase and isAutoStartup up ahead among accessors

As lifecycle methods, personally I like to see start/stop/isRunning ahead of runtime methods like `connect`, and it's something I've been following, but I'm not sure if there a broader precedent.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpConnector.java
```diff
@@ -0,0 +1,122 @@
...
+	@Override
+	public void stop(Runnable callback) {
+		stop();
+		callback.run();
+	}
+
+	@Override
+	public void start() {
+		try {
+			this.httpClient.start();
```
rstoyanchev: A code level comment that explains HttpClient is internally synchronized and protected with state checks around here could be useful.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpConnector.java
```diff
@@ -0,0 +1,122 @@
...
+				this.httpClient.start();
+			}
+			catch (Exception ex) {
+				return Mono.error(ex);
+			}
+		}
+
+		if (!uri.isAbsolute()) {
+			return Mono.error(new IllegalArgumentException("URI is not absolute: " + uri));
+		}
```
rstoyanchev: No practical difference but maybe this check first (before starting the client above) ?

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+
+	@Override
+	protected void applyHeaders() {
+		HttpHeaders headers = getHeaders();
+		headers.entrySet().forEach(e -> e.getValue().forEach(v -> this.jettyRequest.header(e.getKey(), v)));
+		if (!headers.containsKey(HttpHeaders.ACCEPT)) {
+			this.jettyRequest.header(HttpHeaders.ACCEPT, "*/*");
+		}
```
rstoyanchev: Is it necessary to add `"*/*"` explicitly? It should be the same as not having it.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,174 @@
...
+
+					return Mono.just(new JettyClientHttpResponse(reactiveResponse, content));
+			}));
+		return Mono.empty();
+	}
+
+	public Mono<ClientHttpResponse> getResponse() {
+		return this.response;
+	}
+
```
rstoyanchev: The overall method ordering should be: accessors, public contract methods, protected contract (base class) methods (e.g. see `ReactorHttpRequest`).

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,174 @@
...
+	@Override
+	public DataBufferFactory bufferFactory() {
+		return this.bufferFactory;
+	}
+
+	@Override
+	public Mono<Void> writeWith(Publisher<? extends DataBuffer> publisher) {
+		String contentType = jettyRequest.getHeaders().contains(HttpHeader.CONTENT_TYPE) ?
+				jettyRequest.getHeaders().getField(HttpHeader.CONTENT_TYPE).getValue():
+				MediaType.APPLICATION_OCTET_STREAM_VALUE;
```
rstoyanchev: Is it not enough to look up via `getHeaders().getContentType()` ?

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,174 @@
...
+			public void succeeded() {
+				DataBufferUtils.release(buffer);
+			}
+
+			@Override
+			public void failed(Throwable x) {
+				DataBufferUtils.release(buffer);
+				throw Exceptions.propagate(x);
+			}
+		}));
```
rstoyanchev: This could be extracted:
```
private ContentChunk toContentChunk(DataBuffer buffer) {
	return new ContentChunk(buffer.asByteBuffer(), new Callback() {
		@Override
		public void succeeded() {
			DataBufferUtils.release(buffer);
		}

		@Override
		public void failed(Throwable x) {
			DataBufferUtils.release(buffer);
			throw Exceptions.propagate(x);
		}
	});
}
```

Then it becomes:
```
Publisher<ContentChunk> chunks = Flux.from(publisher).map(this::toContentChunk);
```
And the writeAndFlushWith method below does the same AFAICS.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,174 @@
...
+		ReactiveRequest.Content requestContent = ReactiveRequest.Content.fromPublisher(chunks, contentType);
+		this.reactiveRequest = ReactiveRequest.newBuilder(jettyRequest).content(requestContent).build();
```
rstoyanchev: Does `reactiveRequest` really need to be a field? It seems like it should be possible to pass it directly into the method called `completes()`. It takes more effort to understand how it's used when it is a field.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpRequest.java
```diff
@@ -0,0 +1,174 @@
...
+					});
+
+					return Mono.just(new JettyClientHttpResponse(reactiveResponse, content));
+			}));
+		return Mono.empty();
+	}
+
+	public Mono<ClientHttpResponse> getResponse() {
+		return this.response;
+	}
```
rstoyanchev: It's not obvious at first why a request object should have a method that returns the response, and a public method for that. It must be called at the right time too, or else the response field may not have been initialized, but I can see why the method is here, based on the way the Jetty API works.

I think it would help clarity if the method called `completes` simply stored `ReactiveRequest`, returned `Mono.empty()`, and nothing more. The method already returns `Mono.empty()` anyway, so the part about creating the response is completely independent and could be moved out of the request. In other words, the connector would access the `ReactiveRequest` (via package private method) and then create the response. Then would make it a little more instantly obvious what's going on.

The other point is that the `Mono<Void>` from writeWith is supposed to reflect the completion of the writing of the request body, per contract, but probably not a practical issue at this time. If we keep all this package private we have the option to evolve it in the future.

rstoyanchev:
Sorry for the delayed response, but my comments above. Let's get this ready for master, and continue there as needed.

spring-web/src/main/java/org/springframework/http/client/reactive/JettyClientHttpConnector.java
sdeleuze: I don't think so, `isStarted()` is different than `isRunning()`.

sdeleuze:
@rstoyanchev  Thanks for your feedback, I think I have taken your remarks in account. I have not yet merged it in `master` because I have a strange remaining issue: `SseIntegrationTests` and `WebClientIntegrationTests` fail on command line with `./gradlew clean test` (`AssertionError`, likely timeouts) but succeed in IntelliJ IDEA. Any idea?

sdeleuze:
Merged via https://github.com/spring-projects/spring-framework/commit/a87764f1fd218356690e05cf2aa964e92202eed6.

PR #1784, opened 2018-04-10
 Throw exception from user code in SpringFailOnTimeout even if a timeout occurs

igor-suhorukov:


spring-test/src/main/java/org/springframework/test/context/junit4/statements/SpringFailOnTimeout.java
```diff
@@ -88,14 +88,21 @@ public void evaluate() throws Throwable {
...
+			} catch (Exception e){
+				failure = e;
 			}
 			finally {
 				long elapsed = System.currentTimeMillis() - startTime;
 				if (elapsed > this.timeout) {
-					throw new TimeoutException(
+					TimeoutException timeoutException = new TimeoutException(
 							String.format("Test took %s ms; limit was %s ms.", elapsed, this.timeout));
+					if(failure != null && timeoutException.getCause() == null){
```
sbrannen: When could the cause ever not be null?

sbrannen: Please add missing leading and trailing spaces for if-block.

spring-test/src/main/java/org/springframework/test/context/junit4/statements/SpringFailOnTimeout.java
```diff
@@ -88,14 +88,21 @@ public void evaluate() throws Throwable {
 		}
 		else {
 			long startTime = System.currentTimeMillis();
+			Exception failure = null;
 			try {
 				this.next.evaluate();
+			} catch (Exception e){
```
sbrannen: We generally name exceptions `ex` instead of `e`.

sbrannen (changes requested):
Regarding the changes to `SpringFailOnTimeout`, I would prefer to address this in a separate PR as a bug fix for a new JIRA issue.

Could you please create a JIRA issue describing the bug?

Please add corresponding tests for the fix as well.

Thanks

igor-suhorukov:
@sbrannen thank you for review. I will write tests [like...](https://gist.github.com/igor-suhorukov/9af4520a8db964b914b2ae49909c07b6)

jhoeller:
I've raised https://jira.spring.io/browse/SPR-16717 for the `JmsMessageEndpoint` part, revisiting exception propagation within the JCA callback model overall. It's not just the `finally` block there, it's the entire notion of what to propagate from `onMessage`. Feel free to repurpose this PR for `SpringFailOnTimeout` only.

igor-suhorukov:
@jhoeller thanks! I revert JmsMessageEndpointFactory changes from this PR

Also thanks a lot for your [keynote speech on Jpoint](https://www.youtube.com/watch?v=GDmsPA4hZ48&feature=youtu.be&t=4266)

sbrannen:
Thanks for fixing the bug in `SpringFailOnTimeout`.

I've merged this into `master` and will backport accordingly.

spring-test/src/test/java/org/springframework/test/context/junit4/spr16716/SpringFailOnTimeoutExceptionTest.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class SpringFailOnTimeoutExceptionTest {
+
+	@Mock
+	private Statement statement;
+
+	@Test(expected = IllegalArgumentException.class)
+	public void validateOriginalExceptionFromEvaluateMethod() throws Throwable {
+		IllegalArgumentException expectedException = new IllegalArgumentException();
```
sbrannen: FYI: testing that an `IllegalArgumentException` is thrown is insufficient since the constructor throws that for invalid input. I revised the tests accordingly.

igor-suhorukov: Thanks for clarification!

igor-suhorukov:
@sbrannen thank you for help!

PR #1785, opened 2018-04-10
Remove unnecessary iterator allocation in type handlers

carterkozak:
HandlerMethodReturnValueHandlerComposite and
AbstractMethodMessageHandler iterate using index over collections
implementing RandomAccess to avoid unnecessary iterators.


rstoyanchev:
This is now merged, thanks.

PR #1788, opened 2018-04-12
Polishing for SPR-16710

nkjackzhang:
Remove unnecessary modifier `public` for `BookingController`.
Using `WebMvcConfigurerAdapter` instead of `WebMvcConfigurer` to avoid unimplemented methods error.


nkjackzhang:
Just create another pr.

snicoll:
@nkjackzhang no need to, you can force push to your branch to update this pr.

PR #1789, opened 2018-04-12
Polishing for SPR-16710

nkjackzhang:
Remove unnecessary modifier public for BookingController.
Using WebMvcConfigurerAdapter instead of WebMvcConfigurer to avoid unimplemented methods error.
Close context after test.


jhoeller:
The use of `WebMvcConfigurer` should be alright there since it has default methods as of Spring Framework 5.0. Only the 4.3.x variant uses `WebMvcConfigurerAdapter` there.

Closing the context is generally sensible but for the sake of brevity we're often omitting it in test cases when there is no actual physical resource to close (i.e. when garbage collection can do the job for us).

As for the `public` modifier, I tried to mimic the sample from the reference documentation. I'm in the process of revising this already, aligning it with an expanded discussion in that ref doc section, so I'll also change the modifier for the test case while I'm at it. Thanks for pointing this out, in any case!

nkjackzhang:
Thanks, I see it.

PR #1792, opened 2018-04-17
SPR-16738 Introduce new interface CacheDecorator.

ptahchiev:
The new interface allows users to customize the cache decorator of a given CacheManager. This new interface has 2 methods: one for decorating
the cacheManager, and another one for checking whether or not the cache decoration should happen.

We now have a TransactionAwareCacheDecorator that implements the interface and the AbstractTransactionSupportingCacheManager
uses it to configure the cache decoration.

All cache managers extending from AbstractTransactionSupportingCacheManager need new constructors and this is why the changes in EhCacheCacheManager
and JCacheCacheManager are needed.

Finally TransactionAwareCacheDecorator has been renamed to TransactionAwareCache to correctly correspond with the implemented interfaces.


spring-context-support/src/main/java/org/springframework/cache/ehcache/EhCacheCacheManager.java
```diff
@@ -49,10 +49,21 @@ public EhCacheCacheManager() {
 	}
 
 	/**
-	 * Create a new EhCacheCacheManager for the given backing EhCache CacheManager.
+	 * Create a new non-transaction-aware EhCacheCacheManager for the given backing EhCache CacheManager.
 	 * @param cacheManager the backing EhCache {@link net.sf.ehcache.CacheManager}
 	 */
 	public EhCacheCacheManager(net.sf.ehcache.CacheManager cacheManager) {
+		this(false, cacheManager);
```
snicoll: While I am not denying the setter is confusing, this is not the scope of this PR to change that.

spring-context-support/src/main/java/org/springframework/cache/transaction/TransactionAwareCache.java
```diff
@@ -0,0 +1,142 @@
...
+ * successful transaction. If no transaction is active, {@link #put}, {@link #evict} and
+ * {@link #clear} operations will be performed immediately, as usual.
+ *
+ * <p>Use of more aggressive operations such as {@link #putIfAbsent} cannot be deferred
+ * to the after-commit phase of a running transaction. Use these with care.
+ *
+ * @author Juergen Hoeller
+ * @author Stephane Nicoll
+ * @author Stas Volsky
+ * @since 3.2
```
snicoll: That's actually incorrect. This class has a new name and a new scope so it must be flagged 5.1

spring-context/src/main/java/org/springframework/cache/support/AbstractCacheManager.java
```diff
@@ -170,7 +177,7 @@ private void updateCacheNames(String name) {
 	 * or simply the passed-in Cache object by default
 	 */
 	protected Cache decorateCache(Cache cache) {
-		return cache;
+		return this.cacheDecorator != null ? this.cacheDecorator.decorateCache(cache) : cache;
```
snicoll: Moving that bits above and allowing the user to change the decorator via a public setter means you can completely shadow what `AbstractTransactionSupportingCacheManager` used to du using public API.

Also, if `CacheDecorator` is a first-class concept, you should probably expect that more than one decorator can apply at a given time. This wrapping shouldn't be the responsibility of the caller.

spring-context/src/main/java/org/springframework/cache/support/CacheDecorator.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.springframework.cache.Cache;
+
+/**
+ * A functonal interface to allow us to decorate the {@code Cache} instances before they are added to the {@code org.springframework.cache.CacheManager}.
+ *
+ * @author Petar Tahchiev
+ * @since 5.1.0
+ */
+public interface CacheDecorator {
```
snicoll: Should be flagged with `@FunctionalInterface`. Not sure I would add that in the description of the interface.

snicoll (commented):
Thanks for the PR.

I am not sure about this one. On one hand, offering a way to decorate a `Cache` is an interesting idea (we do it for caching). On the other, this looks like a lot of added complexity.

There are things that shouldn't have been changed in this PR and that makes it more verbose than it should be. One way of moving forward would be rework the PR to limit changes only to what is strictly necessary.

I've added a few comments.

spring-context-support/src/main/java/org/springframework/cache/transaction/TransactionAwareCache.java
```diff
@@ -0,0 +1,142 @@
...
+ * <p>Use of more aggressive operations such as {@link #putIfAbsent} cannot be deferred
+ * to the after-commit phase of a running transaction. Use these with care.
+ *
+ * @author Juergen Hoeller
+ * @author Stephane Nicoll
+ * @author Stas Volsky
+ * @since 5.1
+ * @see TransactionAwareCacheManagerProxy
+ */
+public class TransactionAwareCache implements Cache {
```
snicoll: Looking at the history, I think it would be more logical that `TransactionAwareCacheDecorator` was renamed to `TransactionAwareCache` (still with the 5.1 since given the rename. That would make the git history much more clear.

spring-context-support/src/main/java/org/springframework/cache/transaction/AbstractTransactionSupportingCacheManager.java
```diff
@@ -16,46 +16,36 @@
...
 	/**
-	 * Return whether this CacheManager has been configured to be transaction-aware.
+	 * Constructor that creates CacheManager with the given transactionAware flag.
+	 *
+	 * @param transactionAware Set whether this CacheManager should expose transaction-aware Cache objects.
+	 *                         <p>Default is "false". Set this to "true" to synchronize cache put/evict
+	 *                         operations with ongoing Spring-managed transactions, performing the actual cache
+	 *                         put/evict operation only in the after-commit phase of a successful transaction.
 	 */
-	public boolean isTransactionAware() {
```
snicoll: I don't see a reason why introducing a concept of `CacheDecorator` should remove those methods. IMO, that should be left untouched (same remark as my first review). This also means that the new `TransactionAwareCacheDecorator` in your proposal should probably be an private concern of this class (rather than a public class).

spring-context-support/src/test/java/org/springframework/cache/jcache/JCacheCacheManagerTests.java
```diff
@@ -45,11 +45,9 @@ public void setupOnce() {
 		cacheManagerMock.addCache(CACHE_NAME);
 
 		cacheManager = new JCacheCacheManager(cacheManagerMock.getCacheManager());
-		cacheManager.setTransactionAware(false);
 		cacheManager.afterPropertiesSet();
 
-		transactionalCacheManager = new JCacheCacheManager(cacheManagerMock.getCacheManager());
-		transactionalCacheManager.setTransactionAware(true);
+		transactionalCacheManager = new JCacheCacheManager(true, cacheManagerMock.getCacheManager());
```
snicoll: That can't compile anymore.

spring-context-support/src/test/java/org/springframework/cache/ehcache/EhCacheCacheManagerTests.java
```diff
@@ -43,11 +43,9 @@ public void setup() {
 		addNativeCache(CACHE_NAME);
 
 		cacheManager = new EhCacheCacheManager(nativeCacheManager);
-		cacheManager.setTransactionAware(false);
 		cacheManager.afterPropertiesSet();
 
-		transactionalCacheManager = new EhCacheCacheManager(nativeCacheManager);
-		transactionalCacheManager.setTransactionAware(true);
+		transactionalCacheManager = new EhCacheCacheManager(true, nativeCacheManager);
```
snicoll: That can't compile anymore.

spring-context/src/main/java/org/springframework/cache/support/AbstractCacheManager.java
```diff
@@ -191,4 +196,11 @@ protected Cache getMissingCache(String name) {
 		return null;
 	}
 
+	public Set<CacheDecorator> getCacheDecorators() {
+		return cacheDecorators;
+	}
+
+	public void addCacheDecorator(CacheDecorator cacheDecorator) {
```
snicoll: That's better, but it's not really changing the base problem. If we want to introduce a general purpose cache decorator, then we need to provide a contract that allows a user to manipulate this properly while making sure the base class (the transaction aware one) can enforce the behaviour that its implementation brings.

snicoll (changes requested):
Thanks, it's better but not  there quite yet. I've added a few comments.

I think there is a problem with the API that allows you to configure a `CacheDecorator`. I need to give it some thoughts as nothing obvious came to mind.

ptahchiev:
Hi @snicoll 
 - I have fixed the compilation issues
 - brought back the `isTransactionAware` and `setTransactionAware` methods
 - use lambda in the `AbstractTransactionSupportingCacheManager` because as you said it must be a private concern of this class only - this also makes it possible to delete the transaction cache decorator class. 
 - I'm not sure how to deal with the ordering issues in the `AbstractCacheManager` - MHO I think it's OK to not rely on ordering and just execute the cache decorators one after the other.

ptahchiev:
BTW I think now the only signature change is `TransactionAwareCacheDecorator` -> `TransactionAwareCache`. If we rename it to be `TransactionAwareCacheDecorator` do you think it would be possible to push this into the upcoming 5.0.6?

snicoll:
Thanks for the feedback. It is a new feature so we need some runway. If we decide to merge this, that'll be 5.1 only.

snicoll:
Thanks for the PR but I am not convinced about a first class concept to decorate caches, especially as the use case you've described isn't the right fit for it. See [this comment](https://jira.spring.io/browse/SPR-16738?focusedCommentId=159248&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-159248) for more details and an alternative proposal.

PR #1794, opened 2018-04-17
SPR-16729 - Add generic matches assertion to JsonPathAssertions

edeandrea:
Adds a generic matches assertion to JsonPathAssertions that
can use a generic Hamcrest Matcher.

Issue: [SPR-16729](https://jira.spring.io/browse/SPR-16729)


rstoyanchev:
I've added hamcrest matcher methods throughout.

PR #1797, opened 2018-04-18
SPR-16741 - Provide XML assertions in WebTestClient

edeandrea:
Adds XML-based assertions in `org.springframework.test.web.reactive.server.WebTestClient.BodyContentSpec` for asserting against the XML body as well as xPath, similar to those assertions offered when using `MockMvcResultMatchers.xpath()`.

Issue: [SPR-16741](https://jira.spring.io/browse/SPR-16741)


rstoyanchev:
This has been merged, thanks for the PR!

PR #1798, opened 2018-04-18
Handle null as null in Yaml

candrews:
https://jira.spring.io/browse/SPR-15425

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
@@ -300,7 +300,7 @@ public abstract class YamlProcessor {
 				}
 			}
 			else {
-				result.put(key, (value != null ? value : ""));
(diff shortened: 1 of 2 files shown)


spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
```diff
@@ -300,7 +300,7 @@ else if (value instanceof Collection) {
 				}
 			}
 			else {
-				result.put(key, (value != null ? value : ""));
+				result.put(key, value);
```
snicoll: This is a too simplistic view of the problem I am afraid. `Properties` does not support `null` value (per javadoc).

snicoll (changes requested):
Thanks for the PR but I think it needs to be reworked. There is some more context in the related issue about the proposed change.

snicoll:
Closing due to lack of feedback. Let's follow-up on the Jira issue.

PR #1806, opened 2018-04-25
Fix typos.

nkjackzhang:
Consist with the "xml code" example in next part.


snicoll:
@nkjackzhang thank you, this is now merged in `master` and in `5.0.x`.

PR #1808, opened 2018-04-25
SPR-16771 - prevent useless init in DefaultWebClientBuilder

joshiste:
Move init of exchangeStrategies to default ctor in DefaultWebClientBuilder

When the copy constructor is used the exchangeStrategies need not to be
initialized as they are set in the copyconstructor

fixes SPR-16771


poutsma:
Merged, thanks!

PR #1809, opened 2018-04-26
Fix typos.

nkjackzhang:
1.Consist with "xml code" examples.
2."xml()" is a static method and will use default builder config, so use createXmlMapper(true) instead.
3.fix mvc namespace tag tpyo.


sdeleuze:
@nkjackzhang Thanks, this is merged in `master` and in `5.0.x`.

PR #1810, opened 2018-04-26
SPR-16773 fix NPE in error-buffered HttpUrlConnection

MarkusKull:
See https://jira.spring.io/browse/SPR-16773


jhoeller:
I went with an extended `catch` block in the end, as mentioned on JIRA. Thanks for the PR, in any case!

PR #1825, opened 2018-05-10
SPR-16809 - Use "internalSet" instead of normal setter.

nkjackzhang:
Using "internalSet" instead of normal setter can avoid checking int param, because if `constants.asNumber()` works as expected and does't throw exception, the param checking in normal setter will never fail.
Issue: SPR-16809


jcs090218 (commented):
I think having another setter is a little bit odd. Even though this does not cause the exception, setter's task is to prevent failure and in this case will be easier for programmer to track down the error if something bad happens.

PR #1826, opened 2018-05-10
SPR-16797 - Update CollectionFactory to support Immutables

dancerjohn:
Update CollectionFactory to support Guava's ImmutableCollections (List, Set, Multiset) and ImmutableMaps (Map, BiMap, SortedMap).


snicoll:
Thanks for the PR but Spring Framework intentionally dropped support with Guava as of Spring Framework 5 and we don't want to reintroduce this library, even in an optional fashion.

PR #1827, opened 2018-05-11
SPR-16812 extract createRunnable() in ScheduledAnnotationBeanPostProcessor

panchenko:
I am interested in overriding this method and having some custom logic in it.
At the moment this project uses 4.3, so it would be nice it this change is backported.

~Please let me know if this change is trivial enough or if I should create a ticket.~

https://jira.spring.io/browse/SPR-16812


snicoll:
@panchenko creating a new hook point certainly requires an issue to discuss the requirement. Please share it in the issue.

jhoeller:
I've addressed this along with SPR-16834 (use of a custom ScheduledTaskRegistrar) in a slightly revised fashion. Thanks for the pull request, in any case!

PR #1828, opened 2018-05-13
[SPR-16816] Catching quartz fails to initialize, shutdown scheduler and rethrows

ScienJus:
jira: https://jira.spring.io/browse/SPR-16816

For some reason, there is an occasional exception arise from database queries when using quartz's cluster mode. when this exception occurs, the Spring container will close and exit, leaving Java Process suspending.

If an exception is thrown in `afterPropertiesSet`, the `destroy` method will not be called, subsequently, these threads will not be closed in the meanwhile.

So I prepare to catch the invocation of `registerJobsAndTriggers` and manually call the destroy method when it throwing an exception.


jhoeller:
As noted on JIRA, I ended up with some more extensive refactoring. Thanks for the pull request in any case!

PR #1830, opened 2018-05-17
SPR-16831 Allow null value for second attribute for ConcurrentModel.a…

sunriize:
…ddAttribute(String, Object) according to the documentation

### DIFF
--- a/spring-context/src/main/java/org/springframework/ui/ConcurrentModel.java
+++ b/spring-context/src/main/java/org/springframework/ui/ConcurrentModel.java
@@ -72,7 +72,6 @@ public class ConcurrentModel extends ConcurrentHashMap<String, Object> implement
 	 */
 	public ConcurrentModel addAttribute(String attributeName, @Nullable Object attributeValue) {
 		Assert.notNull(attributeName, "Model attribute name must not be null");
-		Assert.notNull(attributeValue, "ConcurrentModel does not support null attribute value");
 		put(attributeName, attributeValue);
 		return this;
 	}


spring-context/src/main/java/org/springframework/ui/ConcurrentModel.java
```diff
@@ -16,14 +16,14 @@
...
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-
 import org.springframework.core.Conventions;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
```
snicoll: Please refrain from making formatting changes like these.

snicoll (commented):
Thanks for the PR! This assertion is indeed out of sync with the `Nullable` annotation. I've added a comment, can you please take a look?

spring-context/src/main/java/org/springframework/ui/ConcurrentModel.java
ddelamotte: Sure, sorry i'll change it right now

ddelamotte:
@snicoll I removed the extra file formating

jhoeller:
The assertion is actually right; it's just the javadoc that is wrong. Overriding the `addAttribute` method from `Model` means that we formally have to declare it as `@Nullable` while this particular implementation doesn't actually support `null` values (it would end up as an NPE in `ConcurrentHashMap.put`).

PR #1832, opened 2018-05-18
SPR-16838 PathMatchingResourcePatternResolver should enforce consistent alphabetical sorting of directory content

benbenw:
A sort was added in SPR-14085 but the default File#compareTo depends
upon the underlying system.
The ordering of the ressources returned by
PathMatchingResourcePatternResolver#retrieveMatchingFiles may differ
between OS (Windows, Unix).
This leads to subtle bugs hard to track.
This PR replace the default sort with one based on the file name.


jhoeller:
As per my JIRA comments, resolved in a more extensive fashion. Thanks for the pull request, in any case!

PR #1833, opened 2018-05-21
SPR-16855 - Make FormHttpMessageWriter easier to extend

XYUU:
The serializeForm method of the FormHttpMessageWriter class is not easy to extend functionality, making a simple modification to make it easier to extend.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageWriter.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageWriter.java
@@ -132,7 +132,7 @@ public class FormHttpMessageWriter implements HttpMessageWriter<MultiValueMap<St
 		message.getHeaders().setContentType(mediaType);
 
 		return Mono.from(inputStream).flatMap(form -> {
-					String value = serializeForm(form, charset);
+					String value = serializeForm(form, charset).toString();
 					ByteBuffer byteBuffer = charset.encode(value);
 					DataBuffer buffer = message.bufferFactory().wrap(byteBuffer);
 					message.getHeaders().setContentLength(byteBuffer.remaining());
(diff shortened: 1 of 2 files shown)


XYUU:
@rstoyanchev 
Yes, you are very keen to guess my needs, I intend to use FormHttpMessageWriter on WebClient, mainly because I think the core function of FormHttpMessageWriter is serializeForm, if you declare it as private, you completely lose the possibility of inheritance. For example, if I need a function similar to the FormHttpMessageWriter, just want to add a verification signature after the Writer, then I need to completely copy the entire class, and then simply modify a line or two in the inside, these practices need to endure the exact same Code, IDEA repeat code warning. In fact, if serializeForm is turned into protected, it will not expand the scope too much, and at the same time, it will simply avoid copying the content and make the code look more beautiful.

BodyInserters does not meet our needs, but in most cases it is sufficient, for example:
```java
		WebClient client = WebClient.builder().build();
		MultiValueMap<String, String> data = new LinkedMultiValueMap<>();
		FormHttpMessageWriter writer = new SignFormHttpMessageWriter();
		client.post().body((o, c) ->
			writer.write(Mono.just(data),
					ResolvableType.forClassWithGenerics(MultiValueMap.class, String.class, String.class),
					MediaType.APPLICATION_FORM_URLENCODED,
					o,c.hints())
		);
```

rstoyanchev:
Yes, it's my job to anticipate your needs and those of others beyond your pull request. If you could please create a ticket in JIRA as per [our requirements](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#submit-a-pull-request).

XYUU:
I already created a ticket：
https://jira.spring.io/browse/SPR-16855

rstoyanchev:
Oops, didn't see that. Thank you.

rstoyanchev:
The serializeForm method is now protected.

PR #1834, opened 2018-05-21
SPR-16856 - Allows the BodyExtractor and HttpMessageConverterExtractor to specify a correct MediaType to extract

XYUU:
Some server-side responses to ContentType are incorrect, but we expect to use the right way to extract data. Now I can specify a correct MediaType to extract.


rstoyanchev:
Please create a ticket in JIRA. If you need to know why it matters, it's explained [in the wiki](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#submit-a-pull-request).

XYUU:
Thank you for your guidance.
https://jira.spring.io/browse/SPR-16856

PR #1842, opened 2018-05-30
Use correct access for execute(Map parameters)

nkjackzhang:
Because `execute(Map parameters)` actually has `public` access.


jhoeller:
Good catch! I've revised this a bit differently though since the `public` notion is not really relevant at that point, so doesn't even have to be mentioned. Being at it, I've also fine-tuned references to "Spring Framework JDBC" towards "Spring Framework's JDBC support". Thanks for the pull request, in any case!

PR #1844, opened 2018-05-31
Fixes in the code examples for WebFlux functional endpoints

violetagg:
- Add import to o.s.w.reactive.function.server.RouterFunctions.route
- When creating response body with a Publisher, one need to provide the type


snicoll:
Thanks @violetagg! FYI, we merge in `master` and then backport if necessary so we prefer the base branch to be `master`.

violetagg:
@snicoll OK good to know. Thanks a lot.

PR #1847, opened 2018-06-03
Introduce Stream support in PropertySources

dreis2211:
Issue: https://jira.spring.io/browse/SPR-16894


jhoeller:
As per my JIRA comment, rolled into master as part of a broader revision. Thanks for the PR, in any case!

PR #1850, opened 2018-06-05
Add JDBC SqlValue classes for arrays

marschall:
Introduce two SqlValue classes for more convenient use of arrays from
Spring JDBC.

- AbstractSqlArrayValue abstract base class for array values
- SqlArrayValue concrete implementation class that uses JDBC 4.0 to
  create an array

Issue: SPR-16904


jhoeller:
I'm bringing this into 6.1 as a single streamlined `SqlArrayValue` class based on JDBC 4, also serving as a template for custom `SqlValue` implementations (since we only provide internal implementations within `Jdbc4SqlXmlHandler` up until this point).

Thanks for the PR in any case, after all those years! I've reused your tests and credited your as a co-author.

marschall:
Thanks a lot, looking forward to the 6.1 release.

PR #1851, opened 2018-06-05
SPR-16858 @BeforeSchedule annotation definition and processing

nikolaikopernik:


quaff:
I think you can add some condition check before your actual logic code in the same method, or you can write a aspect to intercept all @Scheduled method.

snicoll:
@nikolaikopernik thanks for the PR but I am not keen to add annotation support like this. We use method-level annotations when the method signature isn't too constrained. 

We need a more  fine grained contract (for instance what happens if you have 3 `@Scheduled` method invocations and you want to only disable one of them?). If the hook point has to return something very specific (i.e. `boolean` in your proposal) it should be defined as a contract the user has to implement. 

We can iterate on the Jira issue you've created.

nikolaikopernik:
> Actually I would argue that the component responsible to determine if the job has to run should be different than the job itself.

I would expect it in the same component and is very easy to use.     
For supporting multiple `@Scheduled` jobs we can introduce the job name (```@Scheduled(name="A")```) and use this name in the startup method: 

```
boolean shouldStartTheJob(String name);
``` 

But the case with multiple Scheduled jobs in one class is rare, I think. Does it worth to complicate the framework for all if 90% users wont use it?

PR #1875, opened 2018-07-08
Support background init tasks via `@Scheduled(cron="@start")`

panchenko:
I would like to use `@Scheduled` to perform some initialization tasks in background.
The syntax is inspired by `@reboot` in `crontab`.

#21557

WDYT?


simonbasle:
This PR has been around for a long time, sorry for coming back to it only now.
We think that the [application event listeners](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#context-functionality-events) feature makes more sense than `@Scheduled` here, especially given the fact that `@start` is a new concept (not a standard cron expression) that is "made up" to shoehorn that usecase into `@Scheduled`.

`@EventListener` with the `@Async` annotation would be a good candidate, with the `ContextRefreshedEvent` event.
In Spring Boot, you also have the `ApplicationStartupEvent` special application event.

PR #1877, opened 2018-07-09
Fix typo

maximAtanasov:
`read read `was mentioned 2 times in the `New in 3.2` Chapter.
Issue was opened here: https://jira.spring.io/browse/GREENHOUSE-557
and is very old at this point.


snicoll:
Thanks for the PR but `3.x` is EOLed now and we won't be making any further change.

PR #1889, opened 2018-07-20
Set SameSite default to Lax

vpavic:
See [SPR-16418 (comment)](https://jira.spring.io/browse/SPR-16418?focusedCommentId=160945&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-160945):

> Isn't `Strict` as a default value for `SameSite` attribute bit too aggressive? If site A issues a session cookie with `SameSite` set to `Strict` that would effectively mean that if users follows a link to site A from any other site, they wouldn't be logged in even though they have a valid session. This can be a bit confusing.
>
> See [this article](https://scotthelme.co.uk/csrf-is-dead/) for some considerations around the topic. The [owasp.org](https://www.owasp.org/index.php/SameSite) also supports Lax as default since it _provides a reasonable balance between security and usability_.

/cc @rwinch

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/session/CookieWebSessionIdResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/server/session/CookieWebSessionIdResolver.java
@@ -125,7 +125,7 @@ public class CookieWebSessionIdResolver implements WebSessionIdResolver {
 				.maxAge(maxAge)
 				.httpOnly(true)
 				.secure("https".equalsIgnoreCase(exchange.getRequest().getURI().getScheme()))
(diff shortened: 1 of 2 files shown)


vpavic:
This was apparently merged to `master` via 82194f4.

BoomManPro:
now , cookie   attritube Samesite default Lax,But i think the cookie attritube "Samesite" canbe update or remove,because add the attritube  "Samesite" will occur some problem . now mandatory usage  it, cannot update or remove , should be configurable.

BoomManPro:
@rwinch  thanks

PR #1892, opened 2018-07-21
Polishing

plx927:
I think AspectJProxyFactory is the extension of ProxyFactory,AspectJProxyFactory can reuse the functionality of ProxyFactory; but i found both class has such method:
```
  public <T> T getProxy()
  public <T> getProxy(ClassLoader classLoader)
```
so i make AspectJProxyFactory extends ProxyFactory directly, eliminate duplicate code.
In addition, i add a convenience method that create Proxy in ProxyFactory, though ProxyFactory has some convenience method by TargetSource, its not useful, because common developer may not what is TargetSource，TargetSource is inner API, but they know target object。


jhoeller:
This arrangement is intentional since we initially wanted to constrain the on-the-fly type binding to `AspectJProxyFactory`. The `ProxyFactory` class itself is more general in purpose and may have subclasses in extensions and user code out there, so we can't easily narrow its `getProxy` delegate signatures for backwards compatibility reasons. Instead, we have the static `getProxy` convenience methods there which provide type binding based on the given proxy interface parameter.

Thanks for raising this, in any case!

PR #1893, opened 2018-07-21
Suggest use of ConfigurableApplicationContext

vincenthsin:
Why it's not ConfigurableApplicationContext


jhoeller:
That section specifically talks about methods on `DefaultListableBeanFactory` and access to them from an `ApplicationContext`. A `ConfigurableApplicationContext` can only really be the starting point for this, with the actual registration methods to be found on the inner `BeanFactory`.

freeman0432:
@jhoeller,
               Thank you very much for answering my doubts. I have found the actual registration method inside the BeanFactory, which is indeed the DefaultListableBeanFactory. I am very sorry to bother you.

PR #1894, opened 2018-07-24
polishing

plx927:
Double check cachedAdvisorBeanNames, to prevent use synchronized each time.


jhoeller:
In fact, there is no hard reason to use synchronization there at all (since the bean factory can handle re-retrieval of getBeanNamesForType just fine), so I went with a plain volatile field and no synchronization at all there. Thanks for raising this!

plx927:
This class always working in single thread environment, so it does not have the problem of memory consistency.If you remove synchronized, volatile can also remove. Another case is `BeanFactoryAspectJAdvisorsBuilder` .

jhoeller:
It's certainly true that this is mostly used during singleton creation on startup. However, there are also the cases of lazy-init singletons and proxied non-singleton beans, so a degree of defensiveness is appropriate there. The use of volatile is a better tradeoff than full synchronization in such scenarios, in particular if the underlying call (getBeanNamesForType in this case) is fully thread-safe itself and no lazily initialized instances are to be guarded there.

PR #1895, opened 2018-07-25
Eliminate ambiguity

vincenthsin:
This is a minor modification, but it can eliminate the original ambiguity.


freeman0432:
I am very sorry, because my English is a bit poor, so it is ambiguous. I already understand what you mean. Thank you very much for your comments.

PR #1899, opened 2018-07-26
Modify a parameter name in document

vincenthsin:
The “text” is more meaningful than "test" in context.

### DIFF
--- a/src/docs/asciidoc/core/core-beans.adoc
+++ b/src/docs/asciidoc/core/core-beans.adoc
@@ -8441,9 +8441,9 @@ simple class that extends Spring's `ApplicationEvent` base class:
 	public class BlackListEvent extends ApplicationEvent {
 
 		private final String address;
-		private final String test;
+		private final String text;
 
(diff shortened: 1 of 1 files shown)


jhoeller:
Well spotted! I've renamed that field to 'content' right away, making its purpose clearer.

freeman0432:
@jhoeller ，
Yes, the 'content' is more appropriate.
However, please note that there are other places that reference the old "test".I found two places：
1. “the test attribute” In the fragment below.
2."#blEvent.test == 'foo'" In the fragment below.

> For instance, our notifier can be rewritten to be only invoked if the test attribute of the event is equal to foo:
> 
> @EventListener(condition = "#blEvent.test == 'foo'")
> public void processBlackListEvent(BlackListEvent blEvent) {
> 	// notify appropriate parties via notificationAddress...
> }

jhoeller:
Indeed, I'll fix those ASAP. Thanks for the careful review!

freeman0432:
It's nothing.I hope to make a little contribution.

PR #1901, opened 2018-07-26
Make "PathVariable reflects value up to 1st semicolon" consistent

izeye:
When I was reading 66d73017d5406e46c176a44e5204922d167bef93, I noticed that assertions for "reactive" and ones for "servlet" are different. This PR tries to make "servlet" codes align with "reactive" ones to be consistent with an assumption that "reactive" ones are correct.


rstoyanchev:
Thanks for noticing. They're different and that's intentional. The idea is to support APIs like `"/foo/{ids}"` and URL `"/foo/id=1;id=2;id=3"` where the whole path segment is a sequence of name-value pairs. WebFlux currently does not support that but it might in the future.

izeye:
Ah, sorry for the noise. I failed to read that intention. Adding some note for it might help for anyone who will misread like me in the future :sweat_smile:

rstoyanchev:
Good idea I've done that.

PR #1903, opened 2018-07-26
SPR-17074 pass argument of addAll/putAll into constructor

stsypanov:
Pass argument of `Collection::addAll`/`Map::putAll` directly into constructor. This is shorter and allocates collection/map of exact size

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/ast/PropertyOrFieldReference.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/ast/PropertyOrFieldReference.java
@@ -322,8 +322,7 @@ public class PropertyOrFieldReference extends SpelNodeImpl {
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


jhoeller:
In those three places, we intentionally did not allocate through the copy constructor since the affected collection is being populated from several sources. Taking the first source into the copy constructor would force the capacity of the target collection to exactly that size of the first source, not accounting for further elements to immediately follow from other sources. Am I missing something?

stsypanov:
All you wrote is correct, basically assuming my changes are applied here we have two cases:
- items are added into collection/map after construction
- no items added after construction

In the first case indeed there could be collection's map internal storage inflation resulting in slight performance degradation.
In the second case, however we have slight performance gain due to precise allocation at construction time.

In both cases difference is not much significant, I prefer immediate population due to its brevity.

jhoeller:
Good point, such subsequent population is always conditional there, so might not even happen...

PR #1907, opened 2018-07-28
Specify param nullability in ModelAndView and ModelMap

sorra:
I think no need to file JIRA ticket for this?

And a question:
The javadoc of `ModelMap#containsAttribute` says `attributeName` should be never null, but it's safe to pass in a null value (no exception thrown).
So should it have `@NonNull` ?

### DIFF
--- a/spring-context/src/main/java/org/springframework/ui/ModelMap.java
+++ b/spring-context/src/main/java/org/springframework/ui/ModelMap.java
@@ -21,6 +21,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 
 import org.springframework.core.Conventions;
+import org.springframework.lang.NonNull;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@sorra Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1907)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1907) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sorra Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=1907)!

jhoeller:
Indeed, the nullable declaration for values is inconsistent: just on `ModelMap.addAttribute` but not on the corresponding `ModelMap` constructor and not on `ModelAndView` either. However, we declare the packages to be non-null by default anyway, so there are only really two `@Nullable` declarations missing... which I am fixing in the course of a larger revision in the `spring-webmvc` module. Thanks for the pull request, in any case!

jhoeller:
You've got fine points there! It's just that we declare our packages as `@NonNullApi` by default which the Kotlin compiler understands, so we only really need to annotate exceptions to this rule: i.e. just `@Nullable` parameters, all others are non-null by formal convention.

I've rolled this into a larger revision large night, addressing this according to our conventions (along with other places where nullability declarations were missing). I hope it works for your purposes now.

PR #1908, opened 2018-07-30
SPR-17105 HttpMethod improvement: stop resizing inner map during initialization.

stefankiv:
Issue: https://jira.spring.io/browse/SPR-17105

Putting all 8 values in the map triggers map's resizing (when it's filled with 6 values), which could be avoided, changing load factor to 1.


jhoeller:
As mentioned on JIRA, I'll rather adapt the capacity since we don't customize the load factor anywhere else either. I'll also do a corresponding review in related places. Thanks for the pull request, in any case!

PR #1909, opened 2018-07-30
SPR-16989 Response body limit filter

fred84:
Implementation for https://jira.spring.io/browse/SPR-16989 .This is default implementation without throwing exception on response size limit exceed.

@rstoyanchev please take a look.


rstoyanchev:
As per comment under ticket, closing this PR.

PR #1911, opened 2018-07-31
SPR-17074 Use lambda for lazy instantiation

stsypanov:
Use lambdas to prevent unnecessary String concatenation / objects instantiation.


sbrannen:

Thanks

PR #1915, opened 2018-08-02
[DOC] Delete MockServer when using RestTemplate with MockMvc

jasssonpet:
I think the `mockServer.verify()` call shouldn't be in this example (as I don't see it defined) but please ignore if I'm wrong.


rstoyanchev:
Yes indeed. That's showing using MockMvc as the server, so not using any of the client-side support.

PR #1917, opened 2018-08-06
SPR-17123 - Revise StringUitls.collectionToCommaDelimitedString

rinx:
Revised: StringUtils.collectionToCommaDelimitedString function.

issue: SPR-17123


jhoeller:
As mentioned on JIRA, I've addressed this as part of a larger revision. Thanks for the pull request, in any case!

PR #1918, opened 2018-08-06
Polish build script

vpavic:
This simplifies build script by removing the following unnecessary properties from `wrapper` task configuration:

- `description`: has no effect as no such property exists in Gradle's [`Wrapper`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.wrapper.Wrapper.html#N2855B) which can be verified with:

```shell
$ ./gradlew tasks | grep wrapper
wrapper - Generates Gradle wrapper files.
```

- `gradleVersion`: not needed as `wrapper` task should preferably be [invoked with `--gradle-version`](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper) option which also eliminates the need to modify the build script on each upgrade of Gradle:

```shell
$ ./gradlew wrapper --gradle-version 4.8.1
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -289,9 +289,6 @@ configure(rootProject) {
 	}
 
 	wrapper {
-		description = "Generates gradlew[.bat] scripts"
-		gradleVersion = '4.8.1'
-
 		doLast() {
 			def gradleOpts = "-XX:MaxMetaspaceSize=1024m -Xmx1024m"
 			def gradleBatOpts = "$gradleOpts -XX:MaxHeapSize=256m"


snicoll (approved):

sbrannen:
> - `description`: has no effect as no such property exists in Gradle's [`Wrapper`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.wrapper.Wrapper.html#N2855B) which can be verified with:
> 
> ```shell
> $ ./gradlew tasks | grep wrapper
> wrapper - Generates Gradle wrapper files.
> ```

It appears to me that you executed that _after_ having removed the custom `description`.

FYI: the `description` property in fact exists in all Gradle tasks that extend from `org.gradle.api.internal.AbstractTask`.

With our custom description place, we had the following output (just verified against `master` with the description re-added).

```
wrapper - Generates gradlew[.bat] scripts
```

After the removal:

```
wrapper - Generates Gradle wrapper files.
```

There's not much of a difference between our custom description and the default. So I guess we can leave it with the default, but I wanted to point out that the `description` was not superfluous.

vpavic:
> It appears to me that you executed that after having removed the custom `description`.

Yes, that was on purpose to demonstrate that `wrapper` task already has a description attached to it (I probably didn't express myself in best way in opening comment). I believe the presence of explicit `description` is a remnant from the time when `wrapper` task was explicitly declared using `task wrapper(type: Wrapper)`.

sbrannen:
> I probably didn't express myself in best way in opening comment.

No worries.

> I believe the presence of explicit `description` is a remnant from the time when `wrapper` task was explicitly declared using `task wrapper(type: Wrapper)`.

Yep, that's definitely the case.

PR #1921, opened 2018-08-08
SPR-17144 - Fix ContextPathCompositeHandler response completion

maubergine:
Updated ContextPathCompositeHandler so that it does not discard the
result of setting the response as complete. This was causing a 200
status code to be returned instead of the intended 404 status code.

Issues: SPR-17144


rstoyanchev:
I've addressed this in a slightly different way, simply returning the result of calling `setComplete`. In other there is no need for `Mono.defer` in this case since the whole block is nested inside a `Supplier`. Thanks for the PR nevertheless!

PR #1929, opened 2018-08-16
SPR-17188 - Declare SmartLifecycle with default methods

quaff:


jhoeller:
As per my comment on JIRA, I've rolled this as part of a broader revision. Thanks for the pull request, in any case!

PR #1931, opened 2018-08-17
SPR-17194 - Change the type of parameters for expires and lastModified to long

making:
This commit changes the type of parameters so that HeaderAssertions
can assert expires and lastModified properly.

Issue: SPR-17194


rstoyanchev:
Merged, thanks!

PR #1932, opened 2018-08-18
WIP: CaffeineCacheManager: Allow configuration of custom caches 

deveth0:
Hi,

This pull request is a work in progress suggestion how the CaffeineCacheManager could be extended to allow the configuration of individual caches.
Currently it's only possible to provide a global configuration for the CacheManager (or use a workaround and utilize multiple CacheManager Instances).
In one of our projects we need to provide a way that we can add / change caches through configuration-files (using spring-boot). 
I started a discussion on [Stackoverflow](https://stackoverflow.com/questions/51890588/configuration-for-specific-caffeine-caches-in-spring) and there is already at least one other [pull-request](https://github.com/spring-projects/spring-framework/pull/1506) which was not updated in the last months. Although #1506 does also deal with the issue, I tried to keep and extend the existing functionality. 

This PR is currently WIP which requires some 'external' input on the idea and I would be grateful for any feedback. 

Cheers

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
@@ -32,6 +32,7 @@ import org.springframework.cache.CacheManager;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
+import org.springframework.util.StringUtils;
 
 /**
  * {@link CacheManager} implementation that lazily builds {@link CaffeineCache}
@@ -59,7 +60,9 @@ public class CaffeineCacheManager implements CacheManager {
 
 	private boolean dynamic = true;
 
-	private Caffeine<Object, Object> cacheBuilder = Caffeine.newBuilder();
+	private final ConcurrentMap<String, Caffeine> cacheBuilder = new ConcurrentHashMap<>(16);
+
+	private Caffeine<Object, Object> defaultCacheBuilder = Caffeine.newBuilder();
 
 	@Nullable
 	private CacheLoader<Object, Object> cacheLoader;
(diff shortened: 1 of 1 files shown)


deveth0:
Any feedback?

spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
```diff
@@ -59,7 +60,9 @@
 
 	private boolean dynamic = true;
 
-	private Caffeine<Object, Object> cacheBuilder = Caffeine.newBuilder();
+	private final ConcurrentMap<String, Caffeine> cacheBuilder = new ConcurrentHashMap<>(16);
```
snicoll: I wonder if we shouldn't put that outside of the `CacheManager` business. After all if you want to configure your infrastructure, that could/should happen upfront. I think this would also vastly simplify the current arrangement where we have to manage the state if the setter is called once some caches are already being built.

snicoll (commented):
I'd be tempted to look at a `Builder` approach where things are configured upfront. and the builder can be used to create a new `CaffeineCacheManager` instance. 

This may be a really bad idea once I actually try to implement it though.

snicoll:
> Currently it's only possible to provide a global configuration for the CacheManager (or use a workaround and utilize multiple CacheManager Instances).

That's not the full story. You can create individual caches and wrap them in a `GenericCacheManager` if you wanted to. There is nothing in the `CacheManager` contract that forces you to use the one that matches the cache library you're using. 

As a matter of a fact, if you expose `Cache` instances as beans, Spring Boot will detect them and wrap them automatically in a `CacheManager`. That would be my solution of choice for the time being. This PR is really about making it centralized but there's really no need to.

snicoll:
@deveth0 I am not keen to bake that configuration business in `CaffeineCacheManager` in that form. I feel this is a configuration concern that should happen before the cache manager is created, so something like a builder would be a better fit IMO but that's a breaking change. There is a separate initiative to reach the same goal in https://github.com/stepio/coffee-boots

PR #1933, opened 2018-08-19
SPR-17195 Fix unusable assertions in HeaderAssertions

alto-000:
Two header assertions (expires, lastModified) accept int parameter
whereas HttpHeaders class stores those headers as long.
It makes them always fail as Object::equals is used for comparing.

The change casts incoming values to long.


snicoll:
@konrad-czajka, thank you for the PR but this was already raised in [SPR-17194](https://jira.spring.io/browse/SPR-17194) and PR #1931

konrad-czajka:
@snicoll I checked for existing issues before SPR-17194 had been reported. I admit I haven't checked again before creating my issue/PR. Sorry for the duplicate.

PR #1936, opened 2018-08-22
More defensive about bad URLs

dsyer:
We can safely catch Throwable here without affecting the logic

Issue: SPR-17198


dsyer:
Superseded by 6ef093

PR #1938, opened 2018-08-22
Polish

izeye:
This PR fixes some typos and polishes trivial stuff.

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
@@ -28,7 +28,7 @@ import org.springframework.util.Assert;
 
 /**
  * {@link org.springframework.cache.Cache} implementation on top of a
- * {@link javax.cache.Cache} instance.
+ * {@link Cache javax.cache.Cache} instance.
  *
  * <p>Note: This class has been updated for JCache 1.0, as of Spring 4.0.
  *
@@ -42,7 +42,7 @@ public class JCacheCache extends AbstractValueAdaptingCache {
 
 
 	/**
-	 * Create an {@link org.springframework.cache.jcache.JCacheCache} instance.
+	 * Create a {@link org.springframework.cache.jcache.JCacheCache} instance.
 	 * @param jcache backing JCache Cache instance
 	 */
 	public JCacheCache(Cache<Object, Object> jcache) {
@@ -50,7 +50,7 @@ public class JCacheCache extends AbstractValueAdaptingCache {
 	}
 
 	/**
-	 * Create an {@link org.springframework.cache.jcache.JCacheCache} instance.
+	 * Create a {@link org.springframework.cache.jcache.JCacheCache} instance.
 	 * @param jcache backing JCache Cache instance
 	 * @param allowNullValues whether to accept and convert null values for this cache
 	 */
@@ -128,7 +128,7 @@ public class JCacheCache extends AbstractValueAdaptingCache {
 				}
 				catch (Exception ex) {
 					throw new EntryProcessorException("Value loader '" + valueLoader + "' failed " +
-							"to compute  value for key '" + entry.getKey() + "'", ex);
+							"to compute value for key '" + entry.getKey() + "'", ex);
 				}
 				entry.setValue(toStoreValue(value));
 				return value;
--- a/spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCacheManager.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCacheManager.java
@@ -28,7 +28,7 @@ import org.springframework.util.Assert;
 
 /**
  * {@link org.springframework.cache.CacheManager} implementation
- * backed by a JCache {@link javax.cache.CacheManager}.
+ * backed by a JCache {@link CacheManager javax.cache.CacheManager}.
  *
  * <p>Note: This class has been updated for JCache 1.0, as of Spring 4.0.
  *
@@ -53,7 +53,7 @@ public class JCacheCacheManager extends AbstractTransactionSupportingCacheManage
 
 	/**
 	 * Create a new JCacheCacheManager for the given backing JCache.
-	 * @param cacheManager the backing JCache {@link javax.cache.CacheManager}
+	 * @param cacheManager the backing JCache {@link CacheManager javax.cache.CacheManager}
 	 */
 	public JCacheCacheManager(CacheManager cacheManager) {
 		this.cacheManager = cacheManager;
@@ -61,14 +61,14 @@ public class JCacheCacheManager extends AbstractTransactionSupportingCacheManage
 
 
 	/**
-	 * Set the backing JCache {@link javax.cache.CacheManager}.
+	 * Set the backing JCache {@link CacheManager javax.cache.CacheManager}.
 	 */
 	public void setCacheManager(@Nullable CacheManager cacheManager) {
 		this.cacheManager = cacheManager;
 	}
 
 	/**
-	 * Return the backing JCache {@link javax.cache.CacheManager}.
+	 * Return the backing JCache {@link CacheManager javax.cache.CacheManager}.
 	 */
 	@Nullable
 	public CacheManager getCacheManager() {


spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
```diff
@@ -28,7 +28,7 @@
 
 /**
  * {@link org.springframework.cache.Cache} implementation on top of a
- * {@link javax.cache.Cache} instance.
```
sbrannen: Without the fully qualified package name, we'd want to have "JCache" there in addition.

So we might as well just leave it as is for clarity, as was originally intended.

Thus, please either undo this change or prefix that with "JCache" as is done elsewhere in the Javadoc for this class.

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
```diff
@@ -42,15 +42,15 @@
 
 
 	/**
-	 * Create an {@link org.springframework.cache.jcache.JCacheCache} instance.
+	 * Create a {@link org.springframework.cache.jcache.JCacheCache} instance.
```
sbrannen: The package name starts with an "o". Thus this is grammatically correct.

Please undo this change.

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
```diff
@@ -28,7 +28,7 @@
 
 /**
  * {@link org.springframework.cache.Cache} implementation on top of a
- * {@link javax.cache.Cache} instance.
+ * {@link Cache} instance.
```
snicoll: That full package name is actually on purpose as the Spring Cache abstraction has also a concept of `Cache`.

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
```diff
@@ -42,15 +42,15 @@
...
+	 * Create a {@link org.springframework.cache.jcache.JCacheCache} instance.
 	 * @param jcache backing JCache Cache instance
 	 */
 	public JCacheCache(Cache<Object, Object> jcache) {
 		this(jcache, true);
 	}
 
 	/**
-	 * Create an {@link org.springframework.cache.jcache.JCacheCache} instance.
+	 * Create a {@link org.springframework.cache.jcache.JCacheCache} instance.
```
sbrannen: The package name starts with an "o". Thus this is grammatically correct.

Please undo this change.

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCacheManager.java
```diff
@@ -28,7 +28,7 @@
 
 /**
  * {@link org.springframework.cache.CacheManager} implementation
- * backed by a JCache {@link javax.cache.CacheManager}.
+ * backed by a JCache {@link CacheManager}.
```
snicoll: Ditto

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCacheManager.java
```diff
@@ -53,22 +53,22 @@ public JCacheCacheManager() {
 
 	/**
 	 * Create a new JCacheCacheManager for the given backing JCache.
-	 * @param cacheManager the backing JCache {@link javax.cache.CacheManager}
+	 * @param cacheManager the backing JCache {@link CacheManager}
```
snicoll: And here

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCacheManager.java
```diff
@@ -53,22 +53,22 @@ public JCacheCacheManager() {
...
 	 */
 	public JCacheCacheManager(CacheManager cacheManager) {
 		this.cacheManager = cacheManager;
 	}
 
 
 	/**
-	 * Set the backing JCache {@link javax.cache.CacheManager}.
+	 * Set the backing JCache {@link CacheManager}.
```
snicoll: etc etc

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCacheManager.java
```diff
@@ -53,22 +53,22 @@ public JCacheCacheManager() {
...
 	 */
 	public void setCacheManager(@Nullable CacheManager cacheManager) {
 		this.cacheManager = cacheManager;
 	}
 
 	/**
-	 * Return the backing JCache {@link javax.cache.CacheManager}.
+	 * Return the backing JCache {@link CacheManager}.
```
snicoll: That too

snicoll (commented):
Thanks for the PR @izeye! I've added a few suggestions to better explain the rationale. Does that make sense?

sbrannen (dismissed):

sbrannen:
@snicoll, it looks like our reviews crossed paths in the either. Oops! 😉

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
izeye: @snicoll If so, it should be `{@link Cache javax.cache.Cache}`, not `{@link javax.cache.Cache}` as the latter will be rendered as just `Cache` in Javadoc. Is it okay to change to `{@link Cache javax.cache.Cache}`?

See https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/jcache/JCacheCache.html

sbrannen:
@snicoll, after our combined reviews, it appears that the only remaining polishing would be the removal of the extra space in the exception message.

Thus, I've dismissed my review and will let you take it from here.

Cheers,

Sam

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
izeye: @sbrannen It looks the same concern as @snicoll 's https://github.com/spring-projects/spring-framework/pull/1938#discussion_r212008591. I answered there and let me know what you think.

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
izeye: @sbrannen "a" is intended for the rendered link in Javadoc, "JCacheCache", not "org.springframework.cache.jcache.JCacheCache". Am I missing something?

See https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/jcache/JCacheCache.html

sbrannen: Hmmm.... good question!

I believe we (i.e., the collective core Spring Team) were thinking that the fully qualified class name (FQCN) shows up in the generated HTML when linked via FQCN.

So, if that's not the case, that would mean there is a larger issue across the entire code base in the core Spring Framework, since we routinely use "an" instead of "a" in such cases.

Perhaps it's better to address that nuanced feature of Javadoc in a dedicated JIRA issue.

@snicoll, thoughts?

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
snicoll: Oh dear, good catch. Yes please.

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
snicoll: @sbrannen thanks. Could you please raise a SPR issue for this? I got caught as well and I am wondering if that's not a new feature of javadoc or something (or I've been wrong this all time).

izeye:
@snicoll @sbrannen Thanks for the feedback! I've updated this PR as suggested.

snicoll (approved):

snicoll:
Thanks a lot @izeye, this is now merged in `master`.

spring-context-support/src/main/java/org/springframework/cache/jcache/JCacheCache.java
sbrannen: Indeed! I am also wondering if this was a change to how Javadoc works in a recent release (perhaps in JDK 8?).

And well... if you've been wrong this whole time, so have I and many others. 😉 

I'll raise a JIRA issue to track this.

sbrannen: Done: https://jira.spring.io/browse/SPR-17208

PR #1943, opened 2018-08-26
SPR-17215 - prevent instantiation of utility classes

sjh836:
https://jira.spring.io/browse/SPR-17215

I think the utility class should not create objects.
Most utility classes in the Spring Framework are therefore abstract classes.
But SocketUtils, NamedParameterBatchUpdateUtils and ReflectUtils are not abstract classes.
I added the abstract keyword to these classes, to prevent instantiation and to adapt coding conventions.

### DIFF
--- a/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
+++ b/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
@@ -45,7 +45,7 @@ import org.springframework.asm.Type;
  * @version $Id: ReflectUtils.java,v 1.30 2009/01/11 19:47:49 herbyderby Exp $
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
-public abstract class ReflectUtils {
+public class ReflectUtils {
 
 	private ReflectUtils() {
 	}
--- a/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
+++ b/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
@@ -45,7 +45,7 @@ import org.springframework.asm.Type;
  * @version $Id: ReflectUtils.java,v 1.30 2009/01/11 19:47:49 herbyderby Exp $
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
-public class ReflectUtils {
+public abstract class ReflectUtils {
 
 	private ReflectUtils() {
 	}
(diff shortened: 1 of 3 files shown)


spring-core/src/main/java/org/springframework/util/SocketUtils.java
```diff
@@ -38,7 +38,7 @@
  * @author Gary Russell
  * @since 4.0
  */
-public class SocketUtils {
+public abstract class SocketUtils {
```
sbrannen: `SocketUtils` is intentionally *not* abstract. 

See the JavaDoc for its constructor. 

Please undo this change accordingly.

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/util/SocketUtils.java
sjh836: Thank you for your review. I modified it.

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
```diff
@@ -45,7 +45,7 @@
  * @version $Id: ReflectUtils.java,v 1.30 2009/01/11 19:47:49 herbyderby Exp $
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
-public class ReflectUtils {
+public abstract class ReflectUtils {
```
sbrannen: `ReflectUtils` is a type from CGLIB which is _repackaged_ into `spring-core` using [Jar Jar](https://github.com/shevek/jarjar).

As such, it is not from Spring, and we therefore do not make such modifications to CBLIB classes.

Thus, please remove this change.

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/util/SocketUtils.java
sbrannen: You're welcome, and thanks for making the change.

sbrannen: I also just noticed that the `ReflectUtils` class you modified is from CGLIB. So I requested additional changes in a separate review.

Sorry for not catching that the first time: I noticed the issue with `SocketUtils` immediately on my phone (b/c I'm an author of that class), but I neglected to see the `cglib` package at that time.

sbrannen (approved):

PR #1948, opened 2018-09-01
SPR-17233 - Fix `ListBasedXMLEventReader.getElementText()`

aholowko:
Fix `ListBasedXMLEventReader.getElementText()`


jhoeller:
I've addressed this in a slightly different fashion through a refactoring between AbstractXMLEventReader and ListBasedXMLEventReader (since both of them are internal classes of ours anyway), with all concrete stream processing code moving to the subclass. Thanks for the PR, in any case!

PR #1950, opened 2018-09-06
SPR-17223 - Support for methods that are declared in interfaces and implemented in super class

s-mop:
https://jira.spring.io/browse/SPR-17223


rstoyanchev:
Closing as per https://github.com/spring-projects/spring-framework/issues/21756#issuecomment-985435644.

PR #1951, opened 2018-09-06
Add Kotlin DSL for working with MockMvc

checketts:
This DSL make working with `MockMvc` much cleaner.

Example:

```
	@Test
	fun json() {
		mockMvc.performGet("/person/{name}", "Lee") {    // 1
			builder { accept(MediaType.APPLICATION_JSON) } // 2
			printRequestAndResponse() // 3
			expect { // 4
				status { isOk }
				content { contentType("application/json;charset=UTF-8") }
				jsonPath("$.name") { value("Lee") }
				json("""{"someBoolean": false}""", strict = false)
			}
		}
	}
```

Highlights:
1. Extension methods to mockMvc to start dsl
1. Builder block for configuring request
1. Useful equivalent to `andDo(print())` that avoids static imports
1. Expect block that highlights response expectations

This code is based on the work of @petrbalat and myself in https://github.com/petrbalat/kd4smt

### DIFF
--- a/spring-test/src/test/kotlin/org/springframework/test/web/servlet/MockMvcDslTests.kt
+++ b/spring-test/src/test/kotlin/org/springframework/test/web/servlet/MockMvcDslTests.kt
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2002-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.test.web.servlet
+
+import org.hamcrest.CoreMatchers
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+import org.junit.jupiter.api.Assertions.assertNotNull
+import org.junit.jupiter.api.assertThrows
+import org.springframework.http.HttpMethod
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import org.springframework.stereotype.Controller
+import org.springframework.test.util.AssertionErrors
+import org.springframework.test.web.Person
+import org.springframework.test.web.servlet.result.MockMvcResultHandlers
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+
+import org.springframework.test.web.servlet.setup.MockMvcBuilders.standaloneSetup
+
+/**
+ * MockMvcDsl tests that verify builder, actions, and expect blocks
+ *
+ * @author Clint Checketts
+ * @author Petr Balat
+ */
+class MockMvcDslTests {
+	private lateinit var mockMvc: MockMvc
+
+	@Before
+	fun setup() {
+		mockMvc = standaloneSetup(PersonController()).build()
+	}
+
+	@Test
+	fun json() {
+		mockMvc.perform(HttpMethod.GET,"/person/{name}", "Lee") {
+			builder { accept(MediaType.APPLICATION_JSON) }
+			print()
+			expect {
+				status { isOk }
+				content { contentType("application/json;charset=UTF-8") }
+				jsonPath("$.name") { value("Lee") }
+				json("""{"someBoolean": false}""", strict = false)
+			}
+		}
+	}
+
+	@Test
+	fun `negative assertion tests to verify the matchers throw errors when expected`() {
+		val name = "Petr"
+		mockMvc.perform(HttpMethod.GET,"/person/$name") {
+			builder { accept(MediaType.APPLICATION_JSON) }
+			print()
+			expect {
+				assertThrows<AssertionError> { content { contentType(MediaType.APPLICATION_ATOM_XML) } }
+				assertThrows<AssertionError> { contentString("Wrong") }
+				assertThrows<AssertionError> { jsonPath("name", CoreMatchers.`is`("Wrong")) }
+				assertThrows<AssertionError> { json("""{"name":"wrong"}""") }
+				assertThrows<AssertionError> { jsonPath("name") { value("wrong") } }
+				assertThrows<AssertionError> { cookie { value("name", "wrong") } }
+				assertThrows<AssertionError> { flash { attribute<String>("name", "wrong") } }
+				assertThrows<AssertionError> { header { stringValues("name", "wrong") } }
+				assertThrows<AssertionError> { model { attributeExists("name", "wrong") } }
+				assertThrows<AssertionError> { model<String>("name") { Assert.assertThat(this, CoreMatchers.`is`("wrong")) } }
+				assertThrows<AssertionError> { redirectedUrl("wrong/Url") }
+				assertThrows<AssertionError> { redirectedUrlPattern("wrong/Url") }
+				assertThrows<AssertionError> { redirectedUrlPattern("wrong/Url") }
+				assertThrows<AssertionError> { status { isAccepted } }
+				assertThrows<AssertionError> { viewName("wrongName") }
+				assertThrows<AssertionError> { HttpStatus.ACCEPTED.isStatus() }
+				assertThrows<AssertionError> { "$.name" jsonPathIs "wrong" }
+				assertThrows<AssertionError> { "$.name" jsonPathMatcher CoreMatchers.`is`("wrong") }
+				assertThrows<AssertionError> { "name" jsonPath { value("wrong") } }
+				assertThrows<AssertionError> { jsonPath("name") { value("wrong") } }
+				assertThrows<AssertionError> { +HandlerMethod("helloJsonWrong") }
+			}
+		}
+	}
+
+	@Test
+	fun `negative assertion tests for xpath`() {
+		mockMvc.perform(HttpMethod.GET,"/person/Clint") {
+			builder { accept(MediaType.APPLICATION_XML) }
+
+			andDo(MockMvcResultHandlers.print())
+			andExpect(status().isOk)
+
+			print()
+			expect {
+				assertThrows<AssertionError> { xpath("//wrong") { nodeCount(1) } }
+			}
+		}
+	}
+
+	@Test
+	fun `demonstrate actions block for raw ResultAction calls as MvcResult is returned by default`() {
+		val result = mockMvc.perform(HttpMethod.GET,"/person/{name}", "Lee") {
+			builder { accept(MediaType.APPLICATION_JSON) }
+			actions {
+				andExpect(status().isOk)
+			}
+			expect{ status { isOk } }
+		}
+
+		assertNotNull(result)
+	}
+
+
+	@Controller
+	private inner class PersonController {
+
+		@RequestMapping("/person/{name}")
+		@ResponseBody
+		operator fun get(@PathVariable name: String): Person {
+			return Person(name)
+		}
+	}
+
+	/** Silly example matcher to demonstrate how to add custom matchers that aren't in the DSL using unary operator */
+	class HandlerMethod(private val name: String) : ResultMatcher {
+
+		override fun match(result: MvcResult) {
+			val handler = result.handler
+			if (handler is org.springframework.web.method.HandlerMethod) {
+				AssertionErrors.assertEquals("Handler name", name, handler.method.name)
+			}
+		}
+	}
+
+
+}
--- a/spring-test/src/main/kotlin/org/springframework/test/web/servlet/MockMvcExtensions.kt
+++ b/spring-test/src/main/kotlin/org/springframework/test/web/servlet/MockMvcExtensions.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.test.web.servlet
+
+import org.springframework.http.HttpMethod
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders
+import java.net.URI
+
+/**
+ *
+ * @author Clint Checketts
+ * @author Petr Balat
+ */
+
+/**
+ * Create a [MockMvcRequestDsl] with the given HTTP verb.
+ * @param method the HTTP method (GET, POST, etc)
+ * @param urlTemplate a URL template; the resulting URL will be encoded
+ * @param uriVars zero or more URI variables
+ * @param block lambda for configuring [MockMvcRequestDsl]
+ * @return the [MvcResult] (never `null`)
+ * @since 5.1.0
+ *
+ */
+fun MockMvc.perform(method: HttpMethod, urlTemplate:String, vararg uriVars: String,  block: MockMvcRequestDsl.() -> Unit = {}): MvcResult {
+    return performDsl(this, MockMvcRequestBuilders.request(method, urlTemplate, *uriVars), block)
+}
+
+/**
+ * Create a [MockMvcRequestDsl] with the given HTTP verb.
+ * @param method the HTTP method (GET, POST, etc)
+ * @param uri the URL
+ * @param block lambda for configuring [MockMvcRequestDsl]
+ * @return the [MvcResult] (never `null`)
+ * @since 5.1.0
+ */
+fun MockMvc.perform(method: HttpMethod, uri: URI, block: MockMvcRequestDsl.() -> Unit = {}): MvcResult {
+    return performDsl(this, MockMvcRequestBuilders.request(method, uri), block)
+}
+
+private fun performDsl(mockMvc: MockMvc,
+					   requestBuilder: MockHttpServletRequestBuilder,
+					   block: MockMvcRequestDsl.() -> Unit = {}): MvcResult {
+    val request = MockMvcRequestDsl(requestBuilder).apply(block)
+    val result = mockMvc.perform(request.buildRequest())
+    return request.applyResult(result).andReturn()
+}
--- a/spring-test/src/main/kotlin/org/springframework/test/web/servlet/MockMvcRequestDsl.kt
+++ b/spring-test/src/main/kotlin/org/springframework/test/web/servlet/MockMvcRequestDsl.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.test.web.servlet
+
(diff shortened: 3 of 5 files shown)


checketts:
Can I get any feedback in regards to this PR? No rush, just curious if I should improve parts.

sdeleuze:
Thanks for this pull request, please find bellow my proposal inspired from yours:

```kotlin
@Test
fun json() {
	mockMvc.perform(HttpMethod.GET, "/person/{name}", "Lee") {
		accept = MediaType.APPLICATION_JSON
		print()
		expect {
			status.isOk()
			content.contentType("application/json;charset=UTF-8")
			jsonPath("\$.name").value("Lee")
			json("""{"someBoolean": false}""", strict = false)
		}
	}
}
```

Any thoughts?

checketts:
Thanks for the reply! 

I like it. The main differences I see:

1. No `builder` block: this mean all configuration will need to happen in the root. This should be fine, I think there was a reason we needed the builder block, but if we can code around them, removing an extra layer would be great.
2. Using `print()` instead of `printRequestAndResponse()`:  Kotlin has a native `print()` function that writes to the console so without the Java `andDo` surrounding syntax it looks like a print to console out. Also as new developers have come on board and they had wonders 'How can I print the request and response in this test, the longer name has jumped out at them and made it simple for them to recognize. I change it if you want, but it does feel like it makes the code less clear.
3. How to support `andDo` blocks like as needed for Spring RestDocs? Just leave them as a top level method as they currently are? (I'm glad discussion is happening on https://github.com/spring-projects/spring-restdocs/issues/547 at the same time)
4. Avoiding blocks in `expect`clause: I like the idea of removing the blocks however it will involve a fair amount of code. For example `status { isOk() }` simply makes the receiver a `StatusResultMatchers` that already exists in Spring. I'll do a spike to see if that change will be very difficult.

So I'll begin researching how difficult points 1 and 4 are to accomplish. I want to say there were technical limitations that made the original approach needed, but I'll investigate to be certain.

checketts:
@sdeleuze One discoverability caveat we ran into is when using `perfom(HttpMethod.GET)` versus `performGet()` is that the one with the `HttpMethod.GET` relies on the user to do an additional static import instead of being able to auto-complete the method. Since there are only 6 verbs, moving them to the top level helps with the auto-complete.

checketts:
`jsonPath("\$.name").value("Lee")`  The escape (`\$`) is not needed when follwed immediately by a dot.

sdeleuze:
2. I think I would keep `print()`
3. `andDo` does not feel very idiomatic in Kotlin DSL, `do` is a reserved keyword, so I would maybe suggest `handle { ... }`
4. Good point, we can keep `status { isOk() }`

sdeleuze:
For `perfom(HttpMethod.GET)` versus `performGet()`, I tend to think the IDE will do a pretty good job proposing the various `HttpMethod` methods and that the static import is needed only if you want `perfom(GET)`, so I am not sure adding the `performXxx()` is really needed.

checketts:
I've pushed updates to incorporate your comments. Here is its current state:

```kotlin
@Test
fun json() {
    mockMvc.perform(HttpMethod.GET,"/person/{name}", "Lee") {
      builder { accept(MediaType.APPLICATION_JSON) }
      print()
      expect {
        status { isOk }
        content { contentType("application/json;charset=UTF-8") }
        jsonPath("$.name") { value("Lee") }
        json("""{"someBoolean": false}""", strict = false)
      }
    }
 }
```

I've had to keep the `builder` since it creates a receiver block for the `MockHttpServletRequestBuilder`. Any recommendations or a different name? `request` or `requestBuilder` perhaps?

spring-test/src/test/kotlin/org/springframework/test/web/servlet/MockMvcDslTests.kt
```diff
@@ -0,0 +1,147 @@
...
+class MockMvcDslTests {
+	private lateinit var mockMvc: MockMvc
+
+	@Before
+	fun setup() {
+		mockMvc = standaloneSetup(PersonController()).build()
+	}
+
+	@Test
+	fun json() {
```
checketts: Here is the passing test that demonstrates the DSL

spring-test/src/test/kotlin/org/springframework/test/web/servlet/MockMvcDslTests.kt
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	@Test
+	fun `negative assertion tests for xpath`() {
+		mockMvc.perform(HttpMethod.GET,"/person/Clint") {
+			builder { accept(MediaType.APPLICATION_XML) }
+
+			andDo(MockMvcResultHandlers.print())
+			andExpect(status().isOk)
```
checketts: I left `andDo` and `andExpect` as methods that allow the user to transition to the new syntax. They can use the syntax they are familiar with as well as include their existing matchers.

By changing to `print()` for printing the request and response then it forces the user to qualify `MockMvcResultHandlers.print()`, but I think that is fine.

I haven't switched `andDo` to `handle` yet. I think we need some better examples before changing that. `handle(document())` for RestDocs for example feels weird.

sdeleuze:
Thanks I will take it from here.

sdeleuze:
After more thoughts, I have rewritten the DSL entirely and ended-up with that more idiomatic DSL which is still pretty close to `MockMvc` spirit with a `router { }` style DSL:

```kotlin
mockMvc {
	GET("/person/{name}", "Lee") {
		print()
		request {
			secure = true
			headers {
				accept = listOf(MediaType.APPLICATION_JSON)
			}
			principal = Principal { "foo" }
		}
		expect {
			status { isOk }
			content { contentType("application/json;charset=UTF-8") }
			jsonPath("$.name") { value("Lee") }
			content { json("""{"someBoolean": false}""", false) }
		}
	}
	POST("/person") {
		request {
			content = """{ "name": "foo" }"""
			headers {
				accept = listOf(MediaType.APPLICATION_JSON)
				contentType = MediaType.APPLICATION_JSON
			}
		}
		expect {
			status {
				isCreated
			}
		}
	}
}
```
See [this commit](https://github.com/sdeleuze/spring-framework/commit/kotlin-mockmvc) for more details. Any thoughts @checketts @jnizet @gzoritchak?

sdeleuze:
Note it is possible to remove the `request { }` block if we want. Not sure.

checketts:
It looks good! The only questions I have are around extensiblility.

1. What is the equivalent of `with(user(myUser))` in the request builder? I use that extensively to test code that relies on logged in user details
2. In `MockHttpServletRequestDsl(private val builder: MockHttpServletRequestBuilder)` could we make the `builder` public so I can add my own extension functions (or expose the builder some way)? 

    In my own code I've added a couple helpers like `json` where I set the content type to json and pass in a string (setting the content) all at once. I realize that is a specialized usecase, so that is why I had left it out of the DSL, but allowed for the extensiblity in my own tests.
3. We also might want to consider making `MockMvcPerformDsl` more extensible as well. For example, how would RestDocs' `andDo(document())` hook in?  Methods that allow
4. `MockMvcDsl` do we want to add a `perform` method so a user can pass in the method as a variable?
5. In one test (https://github.com/spring-projects/spring-framework/compare/master...sdeleuze:kotlin-mockmvc#diff-1950a1703b6b3d8d2b5110324ebbe4e2R124) you nested the `print` and `expect`, was that intentional? Do we want to add `@DslMarker` to discourage that?

jnizet:
Here are my thoughts on this example alone. I'll comment on the code after, as comments to the commit.

 - `accept = listOf(MediaType.APPLICATION_JSON)`. I find this a bit too verbose. In most of the cases, in my experience, we pass a single value for this header. So maybe a `fun accept(vararg mediaType: MediaType)` would be nicer: `accept(MediaType.APPLICATION_JSON)` in the common case, and if we want more: `accept(MediaType.APPLICATION_JSON, MediaType.TEXT_HTML)`
- I find it strange to have `jsonPath("$.name") { value("Lee") }` outside of the `content { }` block, since the `jsonPath` expectation is about the content.  
- The contentType() expectation, on the other hand, is not about the content, but about the `Content-Type` header. So I wouldn't expect it inside a `content { }` block. If it must be in a block, I would expect it inside a `headers` block.
- I'm not sure I like the all-uppercase GET and POST methods. They violate the standard naming conventions, and are inconsistent with the Java DSL which uses get and post.

sdeleuze:
@checketts I have updated my branch in order to take in account your feedback:
1. I added `MockHttpServletRequestDsl#with`
2. I made the DSL classes `open` instead
3. I added `MockMvcPerformDsl#handler` + DSL class is `open` now so you can use extensions on them
4. I added `MockMvcDsl#request` and `MockMvcDsl#multipart`
5. Food catch, the DSL is now using `@DslMarker`

sdeleuze:
@jnizet Please find my feedback:
 - I have bring back `accept` and `contentType` shortcuts.
 - For `jsonPath` expectation I think I just follows the Java DSL
 - For `contentType` expectation I think I just follows the Java DSL
 - Maybe better to stay consistent with the Java DSL indead.

I have also removed the `request { }` that is a kind of artificial one.

Latest version:

```kotlin
mockMvc {
	get("/person/{name}", "Lee") {
		print()
		secure = true
		accept = MediaType.APPLICATION_JSON
		headers {
			contentLanguage = Locale.FRANCE
		}
		principal = Principal { "foo" }
		expect {
			status { isOk }
			content { contentType("application/json;charset=UTF-8") }
			jsonPath("$.name") { value("Lee") }
			content { json("""{"someBoolean": false}""", false) }
		}
	}
	post("/person") {
		content = """{ "name": "foo" }"""
		headers {
			accept = listOf(MediaType.APPLICATION_JSON)
			contentType = MediaType.APPLICATION_JSON
		}
		expect {
			status {
				isCreated
			}
		}
	}
}
```

I like it :-)

checketts:
Yeah it looks great! The only question I have left is: 'How can a user add a custom matcher?' (extending the expect block)

sdeleuze:
You can use `expect { match(customMatch) }` or create a `MockMvcResultMatchersDsl` extension that will invoke it.

jnizet:
@sdeleuze I have added a few comments on your commit (https://github.com/sdeleuze/spring-framework/commit/kotlin-mockmvc)

jnizet:
Another concern. Suppose I want to plug a Spring Rest Docs Kotlin DSL into this DSL. I'd like something like this:
```
mockMvc {
	get("/person/{name}", "Lee") {
		expect {
			status { isOk }
		}
		document {
			pathParameters {
				add("name", "the name identifying the person to get")
			}
		}
	}
}
```

This means that I need to be able to create the following extension function:
```
fun MockMvcPerformDsl.document(configure: DocumentationScope.() -> Unit) {
    // here I need to add a result handler somehow, but that doesn't seem possible
}
```

I think MockMvcPerformDsl should either expose the underlying ResultActions, or expose a low-level do() method that allows passing an existing ResultHandler. It should probably be the same for expect() that would accept an existing ResultMatcher. Or it should expose a method taking a `ResultActions.() -> Unit` as argument, to let the caller do whatever it wants with the ResultActions.

sdeleuze:
`MockMvcPerformDsl#handle(ResultHandler)` and `MockMvcResultMatchersDsl#match(ResultMatcher)` are designed for that, is it ok for your need?

sdeleuze:
I need to rework the DSL, the lifecycle between preparing the request and applying the result matchers/handlers is not correct as pointed out by @jnizet. That's a tricky one, I will maybe see if I can use optional named parameters to configuration the request (with an optional lambda block for advanced use cases).

checketts:
I addressed that by unifying them into a list of `ResultActions` see https://github.com/spring-projects/spring-framework/pull/1951/files#diff-231a1c3cd6f63c3dd461e70d7002830cR58

jnizet:
> `MockMvcPerformDsl#handle(ResultHandler)` and `MockMvcResultMatchersDsl#match(ResultMatcher)` are designed for that, is it ok for your need?

I think it is, yes.

sdeleuze:
@checketts Thanks, but I think there we need to stay closer to `MockMvc` lifecycle, I think I have been too far in mixing various parts of the DSL. 

Here is my latest proposal which seems to me more consistent. The number of optional parameter is important but with named parameter it seems doable and this is more close to `MockMvc` spirit and lifecycle.

```kotlin
mockMvc {
	request(GET, "/person/Lee",
			secure = true,
			accept = MediaType.APPLICATION_JSON,
			headers = { contentLanguage = Locale.FRANCE },
			principal = Principal { "foo" }) {
		status { isOk }
		print()
		content { contentType("application/json;charset=UTF-8") }
		jsonPath("$.name") { value("Lee") }
		content { json("""{"someBoolean": false}""", false) }
	}

	request(POST, "/person",
			content = """{ "name": "foo" }""",
			headers = {
				accept = listOf(MediaType.APPLICATION_JSON)
				contentType = MediaType.APPLICATION_JSON
			}) {
		status { isCreated }
	}
}
```
I have pushed the updated implementation on [my branch](https://github.com/sdeleuze/spring-framework/tree/kotlin-mockmvc). Any thoughts?

jnizet:
Hmm, I'm not a big fan, especially since the varargs for url variables is missing here, and is quite important, IMHO.

Why not just chain call, just the way we do it with `list.map { ... }.filter { ... }`?

The DSL could look like the following, which is actually closer to your original proposal:

mockMvc {
	get("/persons/{name}", "Lee") { // urlTemplate, vararg variables
		secure = true,
		accept(MediaType.APPLICATION_JSON) // vararg, since we might want to pass several media types
		headers {
			contentLanguage = Locale.FRANCE
		}
		principal = Principal { "foo" }
	}.then {
		status { isOk }
		print()
		content { 
			contentType("application/json;charset=UTF-8") 
		}
		jsonPath("$.name") { value("Lee") }
		content {
			json("""{"someBoolean": false}""", false) 
		}
	}
}

This is actually closer to the Java DSL, while still being very Kotlinesque, IMHO. It allows doing everything you want to create the request. Since the get()/post()/put(), etc. functions take some `RequestBuilder.() -> Unit` extension function, we can extend the DSL by creating helper extension functions to specify several properties at once. And the `ActionsBuilder.() -> Unit` passed to `then()` has the MockMvc ResultActions available to simply execute all the result matchers and result handlers sequentially.

What do you think?

sdeleuze:
Yeah, I was hesitating between both approaches, I will update my branch accordingly. Not sure about `then`, we can maybe keep `expect` even if we integrate some handlers to the mix ?

jnizet:
Yes, I chose `then() because it was more neutral. Maybe `and()`? expect() would be a bit strange indeed, since print(), document(), log(), etc. are not expectations.

checketts:
I'm not sure I understand the need to separate the request/response. Is it just to keep the 2 separate? I thought the reworking was to preserve the call/expect ordering that @jnizet had noted.

If we stayed with the `request` block @sdeleuze had first mentioned, we would have the separation:

```
request {
			secure = true
			headers {
				accept = listOf(MediaType.APPLICATION_JSON)
			}
			principal = Principal { "foo" }
		}
```

sdeleuze:
Latest iteration closer to original `MockMvc` Java DSL but still idiomatic Kotlin:

```kotlin
mockMvc.get("/person/{name}", "Lee") {
	secure = true
	accept = MediaType.APPLICATION_JSON
	headers {
		contentLanguage = Locale.FRANCE
	}
	principal = Principal { "foo" }
} andExpect {
	status { isOk }
	content { contentType("application/json;charset=UTF-8") }
	jsonPath("$.name") { value("Lee") }
	content { json("""{"someBoolean": false}""", false) }
} andDo {
	print()
}

mockMvc.post("/person") {
	content = """{ "name": "foo" }"""
	headers {
		accept = listOf(MediaType.APPLICATION_JSON)
		contentType = MediaType.APPLICATION_JSON
	}
} andExpect {
	status {
		isCreated
	}
}
```

checketts:
This is much better than the one with all the optionalarguments.

If I don't need to customize the request then a minimal one would be:

```kotlin
mockMvc.get("/person/{name}", "Lee") andExpect {
	status { isOk }
} 
```

jnizet:
I like this last one a lot (except for `accept = MediaType.APPLICATION_JSON`which I would still replace by accept() with a vararg). 👏 
andExpect and andDo both return the same type, right? So we can chain them in any order?

jnizet:
I had a look at the code, and it confirms that this last version is the right one to me: it looks very Kotlinesque and the code makes it clear that it's actually a thin wrapper around the Java DSL, making the transition between both intuitive and natural. 👍 

I'm not sure I understand why the classes are open, though.

sdeleuze:
Glad you like it @checketts @jnizet :-)

Yes `andExpect` and `andDo` return the same type. I will have a look to `accept` but I kind like the property syntax for a single one which is pretty common, while `headers { accept(listOf(...)) }`  allows to set multiple one if needed.

Classes are open to allow users to add extensions.

jnizet:
> Classes are open to allow users to add extensions.

A class doesn't need to be open to add extensions to it: you can add extensions to String, Int, whatever. It only needs to be open if you want to create a subclass. But the DSL (`ResultActionsWrapper`, for example) won't ever create an instance of a subclass (it calls `MockMvcResultMatchersDsl(actions)`) , so I don't see why it would be necessary.

sdeleuze:
What you describe is true only for Java classes, Kotlin classes need to be open to allow users to add extensions on them.

checketts:
Please add default values for the `dsl: MockHttpServletRequestDsl.() -> Unit` parameter.

Otherwise the minimal usecase requires you to customize the response, even when it isn't needed:

```kotlin
mockMvc.get("/person/{name}", "Lee") {} andExpect { //<- note the extra curlies before the andExpect 
	status { isOk }
}
```

checketts:
The latest is definitely the best. Some of the extensions I had tried before weren't possible, and are possible again.

Here is the comparison I've been working against: https://github.com/sdeleuze/spring-framework/pull/1/files

checketts:
@jnizet Here is an example of extending the expects DSL: https://github.com/sdeleuze/spring-framework/pull/1/files#diff-525cfe15474d3e20b0902649bd2e8695R31 which in this case required it to be `open`

And here is where it is used: https://github.com/sdeleuze/spring-framework/pull/1/files#diff-525cfe15474d3e20b0902649bd2e8695R53

```kotlin
	@Test
	fun `hello json`() {
		val name = "Petr"
		mockMvc.get("/hello/$name").andExpectCustom(::ClintMatchers) {
			"$.surname" jsonPathIs name //JsonPath
			"surname" jsonPath { value("Petr") }
			model<HelloPostDto>("helloPostDto") {
				assertEquals("Balat", surname)
			}
		}
	}
```

I'm open to feedback on if this approach is even useful. I was mainly exploring if it was open enough to support these goals.

jnizet:
@sdeleuze please check [this link](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS4zLVJDIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJmb2xkZWRCdXR0b24iOnRydWUsInJlYWRPbmx5IjpmYWxzZSwiY29kZSI6Ii8qKlxuICogWW91IGNhbiBlZGl0LCBydW4sIGFuZCBzaGFyZSB0aGlzIGNvZGUuIFxuICogcGxheS5rb3RsaW5sYW5nLm9yZyBcbiAqL1xuXG5jbGFzcyBGb28gLy8gbm90IG9wZW5cblxuZnVuIEZvby5iYXIoKSA9IFwiaGVsbG9cIlxuXG5mdW4gbWFpbigpIHtcbiAgICBwcmludGxuKEZvbygpLmJhcigpKVxufSJ9). It shows that a class doesn't need to be open to accept extension functions. They're just disguised static helper method taking the receiver as argument.

checketts:
Here is a better approach: if we make `actions` public in `ResultActionsWrapper` then it allows the custom expect block as well as `andDocument` which restdocs would want:

See: https://github.com/sdeleuze/spring-framework/pull/1/files#diff-525cfe15474d3e20b0902649bd2e8695R54

```kotlin
	fun ResultActionsWrapper.andDocument(identifier: String, configure: DocumentationScope.() -> Unit): ResultActionsWrapper {
		actions.andDo(DocumentationScope(identifier).apply(configure).document())
		return this
	}
```

sdeleuze:
@jnizet My mistake, I thought it was required.

Thanks for your help, I will be off during my week of vacation. I will check extensibility with you, document and add more tests when I will be back.

jnizet:
Meanwile, I've polished a few things in the restdocs DSL, which is now fully documented, tested with 100% coverage, and with a DSL for preprocessors too. You can check it out at https://github.com/Ninja-Squad/spring-rest-docs-kotlin. I'm pretty happy with the result.

With the handle() method, adding an `andDocument()` extension function which will call `andDo` and delegate to `handle()` to add the documentation handler shouldn't be a problem.

sdeleuze:
I have the possibility to avoid `ResultActionsWrapper` by using `infix` extensions like `infix fun ResultActions.andExpect(dsl: MockMvcResultMatchersDsl.() -> Unit): ResultActions`.

On one side, I like the fact we avoid an artificial wrapper that can limit extensibility and make the DSL available in every `ResultActions`. But it may confuse people since `mockMvc.get(...) {
... }.andExpect { ... }` won't provide the Kotlin DSL, only `mockMvc.get(...) {
... } andExpect { ... }` will.

Any thoughts @jnizet @checketts?

jnizet:
The confusion between `.andExpect` and `andExpect` is a deal-breaker to me. I don't really mind the ResultActionsWrapper class existence, but a better name would be nice.

checketts:
Another strike I noticed is that you can't chain a non-infix after an infix.  For example if I created an `andDocument("iHaveAParameterSoICan'tBeInfix") { }` block then the compiler wouldn't let me add it after calling `andDo` or `andExpect` using their infix forms.

jnizet:
:tada:

PR #1958, opened 2018-09-16
Refactor String#replaceAll

durigon:
If we repeatedly call String#replaceAll, we internally repeatedly call the regular expression pattern compilation every time as following:
```java
    public String replaceAll(String regex, String replacement) {
        return Pattern.compile(regex).matcher(this).replaceAll(replacement);
    }
```
The modifications are to keep the compiled pattern.
Therefore, compiling a relatively expensive regular expression pattern does not have to be done every time.

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/ast/StringLiteral.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/ast/StringLiteral.java
@@ -16,6 +16,8 @@
 
 package org.springframework.expression.spel.ast;
 
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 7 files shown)


jhoeller:
Well spotted! I've addressed those under https://jira.spring.io/browse/SPR-17279 now, in a somewhat different fashion: using our own `StringUtils.replace` for simple replacement steps, and pre-stripping the Weak ETag prefix with `String.substring` before we start ETag header value comparisons.

durigon:
@jhoeller I think it is better to use the proper function(StringUtils#replace, String#substring) depending on the situation.

PR #1960, opened 2018-09-17
Fix DefaultHandlerExceptionResolver Javadoc comments to match behaviour

simonbowring:
An "obvious fix". 

Most notably, a `MissingServletRequestParameterException` actually results in a 400 response, not 500.


simonbowring:
You're welcome!

PR #1963, opened 2018-09-19
Allow configuration of Jetty WebSocket's max receive packet size

botorabi:
Currently, Jetty's WebSocket packet size for receive direction is limited by a constant (DEFAULT_FRAME_MAX_SIZE = 64Kb). This causes a problem while receiving e.g. multimedia contents such as jpg images. The pull requests tries to provide a configuration of this max packet size.

Honesty, I am not sure if it is the best idea using @Value annotation. But at least, this pull request shows the demand for a proper configuration.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/socket/WebSocketMessage.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/socket/WebSocketMessage.java
@@ -158,7 +158,7 @@ public class WebSocketMessage {
 		/**
 		 * WebSocket pong.
 		 */
-		PONG;
+		PONG
 	}
 
 }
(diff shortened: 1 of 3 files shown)


rstoyanchev:
Did you mean Netty (it says Jetty in the PR heading)? We're waiting on change in Reactor Netty however, see https://jira.spring.io/browse/SPR-16228 and https://github.com/reactor/reactor-netty/issues/223. The change you have made does not actually impact Netty, and only impacts the aggregation of fragments within a WebSocket message.

botorabi:
Indeed, it's about Netty and its receiving fragment aggregation size. In our application we use spring cloud gateway to setup a reverse proxy and run into the mentioned size limit -- the service behind the proxy serves a WebSocket for delivering jpeg images.

GreenNun:
> This is now fixed.

Could you provide configuration for spring-cloud-gateway users please?

PR #1965, opened 2018-09-22
Polish Aware Javadoc

rupertw:
Refer to method initializeBean of AbstractAutowireCapableBeanFactory is more appropriate than the class itself in context.


jhoeller:
On close review, I've removed the reference to `AbstractAutowireCapableBeanFactory` completely since `ApplicationContextAwareProcessor` is a much better example here that be used as a template for custom extensions. Thanks for the pull request, in any case!

PR #1966, opened 2018-09-23
SPR-17299 - Fix javadoc error

rupertw:
Comments on applying the annotation Autowired to constructor are error, we can refer to the javadoc of class AutowiredAnnotationBeanPostProcessor for correct description.

Issue: SPR-17299


jhoeller:
Addressed as part of a wider javadoc revision now. Thanks for the pull request, in any case!

PR #1975, opened 2018-10-02
Updated referenced dependency on project reactor to working version

kylec32:
Tried to use the referenced `io.projectreactor.netty:reactor-netty` however the dependency was not found. Changing it to `io.projectreactor.netty:reactor-netty:0.8.0.RELEASE` leads to the dependency being found but getting a `ClassNotFoundException` at runtime.

After switching to: `io.projectreactor.ipc:reactor-netty` everything started working.

### DIFF
--- a/src/docs/asciidoc/web/websocket.adoc
+++ b/src/docs/asciidoc/web/websocket.adoc
@@ -1595,7 +1595,7 @@ messages to it, and then forwards all messages received
 from the broker to clients through their WebSocket sessions. Essentially,
 it acts as a "`relay`" that forwards messages in both directions.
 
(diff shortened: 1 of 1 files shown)


snicoll:
This is the documentation for `master` (Spring 5.1) and [this looks legit to me](https://github.com/spring-projects/spring-framework/commit/82a8e42ff937a0064d20f378debafe80a5b4c4d5#diff-a5481d1f7fabe9df2f53476093ce9309). The doc for `5.0.x` is different and consistent with what the build uses.

If you believe you've found an issue with that setup, please create an issue in Jira and we can have a look.

PR #1976, opened 2018-10-03
Leverage ObjectProvider instead of autowired containers

sdeleuze:
In order to be able to leverage WebFlux configuration in a functional way, `WebHttpHandlerBuilder` and `RouterFunctionMapping` should leverage new `ObjectProvider` capabilities to get a sorted list of beans by type instead of using autowired containers.

@bclozel @jhoeller After a second look, it is not required to refactor [`DelegatingWebFluxConfiguration`](https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/config/DelegatingWebFluxConfiguration.java) since in functional mode the autowired setter can be called externally using `ObjectProvider` capabilities + I don't think we could reach consistent design via a constructor approach.

@poutsma Are you ok with these changes?

Issue: [SPR-17327](https://jira.spring.io/browse/SPR-17327)

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/adapter/WebHttpHandlerBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/server/adapter/WebHttpHandlerBuilder.java
@@ -18,13 +18,11 @@ package org.springframework.web.server.adapter;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


poutsma (approved):
Looks good to me!

sdeleuze:
Merged via 3ff573142954a88625bf23d590582b225f9795ca.

PR #1977, opened 2018-10-03
Optimize HTTP headers parsing

bclozel:
Several benchmarks underlined a few hotspots for CPU and GC pressure in
the Spring Framework codebase:

* `org.springframework.util.MimeType.<init>(String, String, Map)`
* `org.springframework.util.LinkedCaseInsensitiveMap.convertKey(String)`

Both are linked with HTTP request headers parsing and manipulation
during the request processing phase.

The first one is linked to repeated calls to
`HttpHeaders.getContentType` within a single request handling. The media
type parsing operation is expensive and the result doesn't change
between calls, since the request headers are immutable at that point.

This commit improves this by caching the parsed `MediaType` for the
`"Content-Type"` request header in the `ReadOnlyHttpHeaders` class. This change
is available for both Spring MVC and Spring WebFlux.

The second one is linked to insertions/lookups in the
`LinkedCaseInsensitiveMap`, which is the data structure behind
`HttpHeaders`. Those operations are creating a lot of garbage (including
a lot of `String` created by `toLowerCase`). We could choose a more
efficient data structure for storing HTTP headers data.

As a first step, this commit is focusing on Spring WebFlux and
introduces `MultiValueMap` implementations mapped by native HTTP headers
for the following servers: Tomcat, Jetty, Reactor Netty and Undertow.
Such implementations avoid unnecessary copying of the request headers
and leverages as much as possible optimized operations provided by the
native implementations.

This change has a few consequences:

* `HttpHeaders` can now wrap a `MultiValueMap` directly
* The default constructor of `HttpHeaders` is still backed by a
`LinkedCaseInsensitiveMap`
* The HTTP request headers for the websocket HTTP handshake now need to
be cloned, because native headers are likely to be pooled/recycled by
the server implementation, hence gone when the initial HTTP exchange is
done

Issue: SPR-17250

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -1617,4 +1617,17 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 		}
 	}
 
+	/**
+	 * Return a {@code HttpHeaders} object that can read and written to.
+	 */
+	public static HttpHeaders writableHttpHeaders(HttpHeaders headers) {
+		Assert.notNull(headers, "HttpHeaders must not be null");
+		if (headers instanceof ReadOnlyHttpHeaders) {
+			return new HttpHeaders(headers.headers);
+		}
+		else {
+			return headers;
+		}
+	}
+
 }
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -35,8 +35,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -47,7 +45,9 @@ import java.util.stream.Collectors;
 
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedCaseInsensitiveMap;
+import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 import org.springframework.util.StringUtils;
 
@@ -78,7 +78,8 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 	/**
 	 * The empty {@code HttpHeaders} instance (immutable).
 	 */
-	public static final HttpHeaders EMPTY = new HttpHeaders(new LinkedHashMap<>(), true);
+	public static final HttpHeaders EMPTY =
+			new ReadOnlyHttpHeaders(new HttpHeaders(new LinkedMultiValueMap<>(0)));
 	/**
 	 * The HTTP {@code Accept} header field name.
 	 * @see <a href="http://tools.ietf.org/html/rfc7231#section-5.3.2">Section 5.3.2 of RFC 7231</a>
@@ -397,35 +398,27 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 	private static final DateTimeFormatter[] DATE_FORMATTERS = new DateTimeFormatter[] {
 			DateTimeFormatter.RFC_1123_DATE_TIME,
 			DateTimeFormatter.ofPattern("EEEE, dd-MMM-yy HH:mm:ss zz", Locale.US),
-			DateTimeFormatter.ofPattern("EEE MMM dd HH:mm:ss yyyy",Locale.US).withZone(GMT)
+			DateTimeFormatter.ofPattern("EEE MMM dd HH:mm:ss yyyy", Locale.US).withZone(GMT)
 	};
 
 
-	private final Map<String, List<String>> headers;
-
-	private final boolean readOnly;
+	final MultiValueMap<String, String> headers;
 
 
 	/**
-	 * Constructs a new, empty instance of the {@code HttpHeaders} object.
+	 * Construct a new, empty instance of the {@code HttpHeaders} object.
 	 */
 	public HttpHeaders() {
-		this(new LinkedCaseInsensitiveMap<>(8, Locale.ENGLISH), false);
+		this(CollectionUtils.toMultiValueMap(
+				new LinkedCaseInsensitiveMap<>(8, Locale.ENGLISH)));
 	}
 
 	/**
-	 * Private constructor that can create read-only {@code HttpHeader} instances.
+	 * Construct a new {@code HttpHeaders} instance backed by an existing map.
 	 */
-	private HttpHeaders(Map<String, List<String>> headers, boolean readOnly) {
-		if (readOnly) {
-			Map<String, List<String>> map = new LinkedCaseInsensitiveMap<>(headers.size(), Locale.ENGLISH);
-			headers.forEach((key, valueList) -> map.put(key, Collections.unmodifiableList(valueList)));
-			this.headers = Collections.unmodifiableMap(map);
-		}
-		else {
-			this.headers = headers;
-		}
-		this.readOnly = readOnly;
+	public HttpHeaders(MultiValueMap<String, String> headers) {
+		Assert.notNull(headers, "headers must not be null");
+		this.headers = headers;
 	}
 
 
@@ -1474,8 +1467,7 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 	@Override
 	@Nullable
 	public String getFirst(String headerName) {
-		List<String> headerValues = this.headers.get(headerName);
-		return (headerValues != null ? headerValues.get(0) : null);
+		return this.headers.getFirst(headerName);
 	}
 
 	/**
@@ -1488,19 +1480,17 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 	 */
 	@Override
 	public void add(String headerName, @Nullable String headerValue) {
-		List<String> headerValues = this.headers.computeIfAbsent(headerName, k -> new LinkedList<>());
-		headerValues.add(headerValue);
+		this.headers.add(headerName, headerValue);
 	}
 
 	@Override
 	public void addAll(String key, List<? extends String> values) {
-		List<String> currentValues = this.headers.computeIfAbsent(key, k -> new LinkedList<>());
-		currentValues.addAll(values);
+		this.headers.addAll(key, values);
 	}
 
 	@Override
 	public void addAll(MultiValueMap<String, String> values) {
-		values.forEach(this::addAll);
+		this.headers.addAll(values);
 	}
 
 	/**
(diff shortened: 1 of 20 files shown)


spring-web/src/main/java/org/springframework/http/server/ServletServerHttpResponse.java
```diff
@@ -153,6 +153,9 @@ public String getFirst(String headerName) {
 			Assert.isInstanceOf(String.class, key, "Key must be a String-based header name");
 
 			Collection<String> values1 = servletResponse.getHeaders((String) key);
+			if (headersWritten) {
+				return new ArrayList<>(values1);
```
bclozel: without that change, I'm seeing duplicate response headers because both the native servlet response ones and the ones held by Spring's servlet response are added. Once the response is committed and headers written, reading headers from both yields duplicates. I think this is a bug uncovered by other changes in this PR.

spring-web/src/main/java/org/springframework/http/server/reactive/TomcatServerHttpRequest.java
```diff
@@ -0,0 +1,102 @@
...
+			throws IOException, URISyntaxException {
+
+		super(createTomcatHttpHeaders(request), request, context, servletPath, factory, bufferSize);
+		this.factory = factory;
+		this.bufferSize = bufferSize;
+	}
+
+	private static HttpHeaders createTomcatHttpHeaders(HttpServletRequest request) {
+		Request tomcatRequest = ((org.apache.catalina.connector.Request) ReflectionUtils
+				.getField(COYOTE_REQUEST_FIELD, request)).getCoyoteRequest();
```
bclozel: Tomcat is the only server implementation that hides everything behind a facade; without reflection, I can't access the native Coyote request and its headers map.

spring-webflux/src/main/java/org/springframework/web/reactive/socket/server/support/HandshakeWebSocketService.java
```diff
@@ -272,7 +272,8 @@ private HandshakeInfo createHandshakeInfo(ServerWebExchange exchange, ServerHttp
 			@Nullable String protocol, Map<String, Object> attributes) {
 
 		URI uri = request.getURI();
-		HttpHeaders headers = request.getHeaders();
+		HttpHeaders headers = new HttpHeaders();
+		headers.addAll(request.getHeaders());
```
bclozel: As mentioned in the commit message, this is the only place where we actually need to clone the request headers.
Because native headers are often pooled by the server, those values are recycled by the time the HTTP handshake is done and those aren't available anymore during the websocket session.

bclozel:
Hey @jhoeller , @rstoyanchev 

As promised, here's a first review cycle for this PR.

Caching the parsed content-type is done for both Spring WebFlux and Spring MVC, but the native headers optimization is not applied to Spring MVC because 1) this might not be the main performance issue in Spring MVC right now and 2) there's only so much we can do with the `Servlet` API as a common bassis. We can revisit that later if we need to.

For now, this PR is trying to stick as much as possible to the previous behavior. I've added `MultiValueMap` implementations that wrap the native headers. Those are often then wrapped by an immutable `MultiValueMap` implementation when the request is declared as read only.

## Possible improvements

We could go one step further and make those wrappers immutable from the start, but this clashes with the general architecture and might not work in certain places.

The content-type optimization only kicks in for HTTP PUT/POST requests where we actually look at the request content-type. The second optimization only applies to Spring WebFlux. We could improve the Spring MVC situation a bit by using a different implementation than `LinkedCaseInsensitiveMap`. My benchmarks show that a simple `List` + `String.equalsIgnoreCase` lookups or even a `TreeMap` + case insensitive `hashcode` yield better results, since `toLowerCase` is the expensive part here (creating `String` and the complex logic with `Locale`).

Let me know what you think!

spring-core/src/main/java/org/springframework/util/UnmodifiableMultiValueMap.java
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Immutable implementation of {@link MultiValueMap}.
+ *
+ * @author Brian Clozel
+ * @since 5.1
+ * @param <K> the key type
+ * @param <V> the value element type
+ */
+public class UnmodifiableMultiValueMap<K, V> implements MultiValueMap<K, V> {
```
poutsma: I don't think this needs to be a public type. It could simply be package friendly, and returned from org.springframework.util.CollectionUtils#unmodifiableMultiValueMap.

But I'm not even sure we need it at all, as I'd rather see a `ReadOnlyHttpHeaders`, see below.

spring-web/src/main/java/org/springframework/http/DefaultHttpHeadersMap.java
```diff
@@ -0,0 +1,166 @@
...
+import org.springframework.util.MultiValueMap;
+
+/**
+ * Default {@code MultiValueMap} implementation for {@link HttpHeaders}.
+ *
+ * @author Brian Clozel
+ * @since 5.1
+ */
+@SuppressWarnings("serial")
+class DefaultHttpHeadersMap implements MultiValueMap<String, String> {
```
poutsma: I think this type is not needed, as it can simply be replaced with `CollectionUtils.toMultiValueMap( new LinkedCaseInsensitiveMap<>(8, Locale.ENGLISH));`

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -927,8 +927,17 @@ public void setContentType(@Nullable MediaType mediaType) {
...
+			return this.cachedContentType;
+		}
+		else {
+			String value = getFirst(CONTENT_TYPE);
+			MediaType contentType = (StringUtils.hasLength(value) ? MediaType.parseMediaType(value) : null);
+			if (this.readOnly) {
+				this.cachedContentType = contentType;
+			}
+			return contentType;
+		}
```
poutsma: This can be improved. The `this.readOnly` check shows that this method is mixing two concerns: map lookup and caching. I would prefer if caching was done in a separate wrapper class(`ReadOnlyHttpHeaders`), that is returned when `HttpHeaders.readOnlyHttpHeaders` is called (instead of backing the map with an unmodifiable `MultiValueMap`).

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -397,30 +396,31 @@
...
-	 * Private constructor that can create read-only {@code HttpHeader} instances.
+	 * Construct a new {@code HttpHeaders} instance backed by an existing map.
 	 */
-	private HttpHeaders(Map<String, List<String>> headers, boolean readOnly) {
+	private HttpHeaders(MultiValueMap<String, String> headers, boolean readOnly) {
 		if (readOnly) {
-			Map<String, List<String>> map = new LinkedCaseInsensitiveMap<>(headers.size(), Locale.ENGLISH);
-			headers.forEach((key, valueList) -> map.put(key, Collections.unmodifiableList(valueList)));
-			this.headers = Collections.unmodifiableMap(map);
+			this.headers = new UnmodifiableMultiValueMap<>(headers);
```
poutsma: I think we should create a `ReadOnlyHttpHeades` instead of using a `UnmodifiableMultiValueMap` here, see below.

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -1626,4 +1631,12 @@ public static HttpHeaders readOnlyHttpHeaders(HttpHeaders headers) {
 		return (headers.readOnly ? headers : new HttpHeaders(headers, true));
 	}
 
+	/**
+	 * Return a {@code HttpHeaders} object that wraps an native headers map.
+	 */
+	public static HttpHeaders wrapNativeHeaders(MultiValueMap<String, String> nativeHeaders) {
```
poutsma: Not really sure this needs a static method. Personally, I think having a public `HttpHeaders(MultiValueMap<String, String> nativeHeaders`) constructor is good enough.

spring-web/src/main/java/org/springframework/http/server/reactive/TomcatServerHttpRequest.java
poutsma: Can't we ask Tomcat to open their headers API for this use case, and offer a getter? I think that using reflection for this is taking things a bit too far, even if that means going back to the plain servlet approach.

poutsma (changes requested):

bclozel:
I've applied @poutsma 's comments (and we fixed some more things as well).
I've just applied the same reasoning to the HTTP response headers and made sure to avoid HTTP headers copying in `DefaultServerHttpRequestBuilder` - those changes were applied in two new commits.

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -1623,7 +1609,25 @@ public String toString() {
...
+		}
+	}
+
+	/**
+	 * Return a {@code HttpHeaders} object that can read and written to.
+	 */
+	public static HttpHeaders writableHttpHeaders(HttpHeaders headers) {
+		Assert.notNull(headers, "HttpHeaders must not be null");
+		if (headers instanceof ReadOnlyHttpHeaders) {
+			return new HttpHeaders(headers.headers);
```
rstoyanchev: If `ReadOnlyHttpHeaders` is a wrapper around `HttpHeaders`, couldn't we check here if `headers.headers` is an instance of `HttpHeaders` and if so return it directly?

spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpHeaders.java
```diff
@@ -0,0 +1,213 @@
...
+import org.springframework.lang.Nullable;
+import org.springframework.util.MultiValueMap;
+
+/**
+ * {@code MultiValueMap} implementation for wrapping Reactor Netty HTTP headers.
+ *
+ * @author Brian Clozel
+ * @since 5.1
+ */
+class ReactorServerHttpHeaders implements MultiValueMap<String, String> {
```
rstoyanchev: A few questions on the class name.

1. `HttpHeaders` is from Netty so probably shouldn't be called Reactor? It could be used with other Netty based runtimes too besides Reactor.
2. The Netty `HttpHeaders` are for client and server use, so "Server" is a misnomer.
3. Both Netty and Spring have "HttpHeaders", and this class is an instance of neither, so calling it `HttpHeaders` is not ideal. It is a MultiValueMap but that would be long. It is an Adapter so perhaps `NettyHeadersAdapter`?
4. 

For Jetty the adapted `HttpFields` is also both client and server. Perhaps `JettyHeadersAdapter`. 

For Tomcat and Undertow they're server specific but no strong reason to use "Server" (what else could they be?) so `TomcatHeadersAdapter` and `UndertowHeadersAdapter` for consistency.

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
bclozel: In that case, `headers.headers` is itself a `MultiValueMap`. @poutsma and I changed `ReadOnlyHttpHeaders` so that it extends HttpHeaders (and calls `super(MultiValueMap)`) to avoid issues on the `equals` contract.

rstoyanchev: Ok I see.

rstoyanchev:
One more comment. In `AbstractServerHttpResponse` the `getHeaders` method returns a new read-only wrapper on every call after committed. Perhaps we can save the read-only wrapper and create it once?

PR #1978, opened 2018-10-04
SPR-17338. Do not throw ServerWebInputException when mandatory request parameter is present

DmytoMrachkovskyi:
ServerWebInputException is throw each time Http request with mandatory request parameters is handled in AbstractNamedValueArgumentResolver. Even though this does not brake request handling flow, exception (and stack trace)  is generated each time and it has performance impact. 

The solution: Default value should be evaluated only if request parameter is empty: 

[JIAR ticket](https://jira.spring.io/browse/SPR-17338)


rstoyanchev:
This has been merged. Thanks for the report and fix.

PR #1980, opened 2018-10-05
Use custom set of beans rather than all beans for EventListener candidates

rahulsh1:
Allow applications to specify a custom set of beans rather than the default implementation for EventListener candidates


rahulsh1:
Jira [SPR-17343](https://jira.spring.io/browse/SPR-17343)

snicoll:
@rahulsh1 sorry it took so long but I don't think we should expose such low-level method in the core framework. The context indexer can be used if you have a lot of beans and the work in the next generation of the framework is going to provide to pre-process those at build time. Thanks anyway.

PR #1981, opened 2018-10-05
Memoized implementation of PathMatchingResourcePatternResolver

rahulsh1:
Allow applications that use uber jar to benefit from cached entries rather than using the slow jar lookup  for component scanning.


snicoll:
There [wasn't a reply](https://github.com/spring-projects/spring-framework/issues/21878#issuecomment-614024370) to the feedback we requested so I am going to close this as well. We can reconsider the original issue if the feedback is provided.

PR #1986, opened 2018-10-09
Refactor the handleNullValue method

arnoldmartinez:
Instead of nesting all the logic inside an if when the variable value is null, that statement is eliminated, improving the readability, eliminating the indentation.


antonianemi:
i agree with the refactoring in order to avoid the nesting if, however i think you should consider keep the else if as "else if (paramType.isPrimitive())" so it can specify that rule is going to be consider in order to throw the exception versus only the else.

snicoll:
Thanks for the PR but checking if the `value` is `null` is a prerequisite of running that logic and your suggestion breaks that.

PR #1987, opened 2018-10-10
Refactor the handle null value method

arnoldmartinez:
Instead of nesting all the logic inside an if when the value is null, now a guard against not null values will cause an early return, improving the readability by removing that indentation.


snicoll:
@arnold-martinez thanks for the PR but I am keen to keep the current version that I find more readable.

arnold-martinez:
To @snicoll 

Friend with all respect, What happens if the value does not meet the first condition?. Obviously null value, we are dealing with developers is not something that is not understood. Now, with all due respect, I am not making random changes, if you have read the book of 'Refactoring' by Martin Fowler the use of if nested is a smell.

I do not share the same opinion with you.

Thanks.

rstoyanchev:
Most often such statements don't apply universally. I don't think there is anything demonstrably wrong with the present code, and it is more concise and readable.

PR #1994, opened 2018-10-16
SPR-17390  - Increase the preprocessing when executing asynchronous calls

hqch0708:
When the method is marked @async, it allows developers to customize the pre-processing to meet functional requirements when performing asynchronous calls through a thread pool.

For example, you can customize ThreadFactory to pass information from ThreadLocal to asynchronous threads as ThreadName when you create a thread, and then perform pre-processing at the asynchronous call.

jira: [https://jira.spring.io/browse/SPR-17390](https://jira.spring.io/browse/SPR-17390)


hqch0708:
I found other solutions. You can use Aspect. This provides another solution. You can also close this PR.

PR #2005, opened 2018-11-02
SPR-17456 ApplicationListenerMethodAdapter can't be removed after targetBean destroy

Pramy:
[SPR-17456](https://jira.spring.io/projects/SPR/issues/SPR-17456?filter=allopenissues)
ApplicationListenerMethodAdapter can't be removed after targetBean destroy
[Demo](https://github.com/Pramy/spring-listener-bug)


snicoll:
Duplicate of #2006

PR #2006, opened 2018-11-02
SPR-17456 ApplicationListenerMethodAdapter can't be removed after targetBean destroy

Pramy:
[SPR-17456](https://jira.spring.io/projects/SPR/issues/SPR-17456?filter=allopenissues)
[Demo](https://github.com/Pramy/spring-listener-bug)


snicoll:
Thanks for the PR but there are many other use of `addApplicationListener`. It is a public API so anybody can call it really and not calling it and replacing that by a bean definition is not the right way of fixing the issue.

I am going to close this now and we can continue on #21988.

PR #2009, opened 2018-11-04
SPR-17458 Documentation explicit Executor usage on @Async section

dimitrisli:
Making more explicit the reference to the Executor usage for `@Async`.

### DIFF
--- a/src/docs/asciidoc/integration.adoc
+++ b/src/docs/asciidoc/integration.adoc
@@ -3413,7 +3413,7 @@ The following table summarizes these registration behaviors:
   previous instance).
 |===
 
-The values in the preceding table are defined as enums on the `RegistrationPolicy` class. 
+The values in the preceding table are defined as enums on the `RegistrationPolicy` class.
 If you want to change the default registration behavior, you need to set the value of the
 `registrationPolicy` property on your `MBeanExporter` definition to one of those
 values.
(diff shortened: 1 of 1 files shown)


src/docs/asciidoc/integration.adoc
```diff
@@ -3413,7 +3413,7 @@ The following table summarizes these registration behaviors:
   previous instance).
 |===
 
-The values in the preceding table are defined as enums on the `RegistrationPolicy` class. 
+The values in the preceding table are defined as enums on the `RegistrationPolicy` class.
```
snicoll: Thanks but that's unrelated to the purpose of the PR. I've fixed that in https://github.com/spring-projects/spring-framework/commit/86846507e2196ec8dbd7ea14bd9ee0f218761ddc

src/docs/asciidoc/integration.adoc
```diff
@@ -6680,8 +6680,35 @@ in combination with a custom pointcut.
...
 
-By default, when specifying `@Async` on a method, the executor that is used is the
-one supplied to the "`annotation-driven`" element, as described earlier. However, you can use the `value`
+As seen previously, you can enable the `@Async` annotation by adding `@EnableAsync` to any of your
+`@Configuration` classes:
+
+====
+[source,java,indent=0]
+[subs="verbatim,quotes"]
+----
```
snicoll: Explaining how to configure `@EnableAsync` in a section named "Executor Qualification with `@Async`" looks odd to me. This is also a copy/paste of the javadoc of `@EnableAsync` so I am not too keen to duplicate that.

Can we take a step back and can you describe why you'd want this to be added here?

snicoll (commented):
Thanks for the PR. I am not sure we should merge this at this point, I've added a question.

src/docs/asciidoc/integration.adoc
dimitrisli: Sure. The main purpose of this PR is as described on https://jira.spring.io/projects/SPR/issues/SPR-17458 to make more clear where the _XML-based 'annotation-driven' element_ default considered executor is coming from. [As also commented here](https://jira.spring.io/browse/SPR-17458?focusedCommentId=163124&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-163124) unless there is a way to link directly to the referred `task:annotation-driven `XML definition (which in my opinion is a better approach), duplicating it in this section is the only other way. 

In particular the initial section you are quoting here is indeed not related to the executor usage but I've included it because that's how I've interpreted the last section of the last phrase of the description of the ticket: _'and make sure it refers to annotations'_. Unless I got it wrong of course - maybe it means an `@Bean` Java Config definition of an Executor?

snicoll:
Thanks for the feedback. I am not keen to copy/paste the setup instruction there so I've added a link and restructured that section a bit  3232825. Thanks for the PR in any case.

PR #2015, opened 2018-11-13
5.0.x

595979101:
I knew that springboot2 could integrate Prometheus.Can spring provide integrated standard jar packages


bclozel:
Please use StackOverflow for questions.

PR #2016, opened 2018-11-13
SPR-17492: FastByteArrayOutputStream.read byte-to-int conversion

janotav:
While evaluating FastByteArrayOutputStream I have noticed that it doesn't behave correctly when negative bytes are encountered in the read() method.

Quoting from InputStream javadoc:

> The value byte is returned as an int in the range 0 to 255

To make the situation more obvious in the test I have deliberately used another InputStream implementation to make the situation clear and avoided using constant that could be confusing.


pivotal-issuemaster:
@janotav Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2016)!

PR #2026, opened 2018-11-24
The name of the parameter is misleading

liuhuagui:
This method sends a message to the application through the  inboundChannel instead of outChannel. This parameter is confusing to developers who read the source code. Thank you for your contribution.

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/messaging/SubProtocolHandler.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/messaging/SubProtocolHandler.java
@@ -51,9 +51,9 @@ public interface SubProtocolHandler {
 	 * Handle the given {@link WebSocketMessage} received from a client.
 	 * @param session the client session
 	 * @param message the client message
-	 * @param outputChannel an output channel to send messages to
+	 * @param inboundChannel an inbound channel to send messages to this application.
 	 */
-	void handleMessageFromClient(WebSocketSession session, WebSocketMessage<?> message, MessageChannel outputChannel)
+	void handleMessageFromClient(WebSocketSession session, WebSocketMessage<?> message, MessageChannel inboundChannel)
 			throws Exception;
 
 	/**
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@51Bigod Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2026)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2026) to manually synchronize the status of this Pull Request.

spring-websocket/src/main/java/org/springframework/web/socket/messaging/SubProtocolHandler.java
```diff
@@ -51,9 +51,9 @@
 	 * Handle the given {@link WebSocketMessage} received from a client.
 	 * @param session the client session
 	 * @param message the client message
-	 * @param outputChannel an output channel to send messages to
+	 * @param inboundChannel an inbound channel to send messages to this application.
```
neVERberleRfellerER: "this application" part looks confusing to me, since while inboundChannel is used by remote side to send messages to "this application", it is not used for same purpose by "this application" itself, but to send messages to remote side instead. And this effectively make inboundChannel and outboundChannel from "this application" perspective which is probably reason for original name.

51Bigod: So this is always very confusing.

neVERberleRfellerER: I apologize. I was wrong and your changes are not confusing (not even that part with application). I have mistaken this with something unrelated and confusing naming together with
https://github.com/spring-projects/spring-framework/blob/5d7fb1a1c2e4925d23a6e4aed37362a174184154/spring-websocket/src/main/java/org/springframework/web/socket/messaging/SubProtocolWebSocketHandler.java#L324
just reinforce my confusion. The channel here is not used for the purpose I said.
I hope that I am not confused this time but I checked the sources and tests.

51Bigod: Thank you very much.

PR #2027, opened 2018-11-25
AbstractBeanFactory.isTypeMatch fix for generic type

it3r:
Issue: SPR-17524


jhoeller:
As mentioned on JIRA, I've resolved this in a slightly different way. Thanks for the pull request, in any case!

PR #2034, opened 2018-11-30
Update StreamUtils.java

liaohang122:
why many InputStream in this file function, needn't close?


snicoll:
@liaohang122 this is on purpose as [indicated on the class Javadoc](https://github.com/spring-projects/spring-framework/blob/bc3b95db92c8775e6405d3510d15d16fbb0a2499/spring-core/src/main/java/org/springframework/util/StreamUtils.java#L35-L36).

PR #2039, opened 2018-12-06
SPR-17571 Allow Instant and ZonedDateTime as Last-Modified header

larsgrefer:
https://jira.spring.io/browse/SPR-17571


jhoeller:
Well spotted! We have overloaded methods for `java.time` types in other APIs already, even in `HttpHeaders` itself, but not for the last-modified timestamp yet. I'm going to merge this and add a few similar methods to other places in a follow-up commit.

PR #2040, opened 2018-12-06
indexOf() expressions can be replaced with a call to the contains()

igor-suhorukov:
indexOf() expressions can be replaced with a call to the contains() method available in Java 5 and newer.


igor-suhorukov:
Hello @jhoeller 
What do you think about this polishing?

PR #2042, opened 2018-12-07
Fix typo in core-beans.adoc

PascalSchumacher:
(obvious fix)


snicoll:
@PascalSchumacher good catch, thank you.

PR #2043, opened 2018-12-08
Fix typo in webflux.adoc

marchpig:
Obvious Fix.


snicoll:
Good catch, thank you!

PR #2050, opened 2018-12-13
Fix sample in EnableAsync Javadoc

diguage:
I think there should be a `new` word here.


snicoll:
@diguage good catch, thank you!

PR #2052, opened 2018-12-13
Join together identical catch branches

igor-suhorukov:


spring-aop/src/main/java/org/springframework/aop/aspectj/AspectJAdviceParameterNameDiscoverer.java
```diff
@@ -272,22 +272,14 @@ public void setThrowingName(@Nullable String throwingName) {
 				}
 			}
 		}
-		catch (AmbiguousBindingException ambigEx) {
+		catch (AmbiguousBindingException | IllegalArgumentException ambigEx) {
```
snicoll: That's nice. Could you please consider polishing this suggestion a bit more by renaming the variable to `ex`? The only reason we're not using it here is because we have two catch blocks

snicoll (commented):
Thanks for the PR, I've made a suggestion that applies to pretty much all change. Could you please have a look?

spring-aop/src/main/java/org/springframework/aop/aspectj/AspectJAdviceParameterNameDiscoverer.java
igor-suhorukov: Sure, thanks!

snicoll:
@igor-suhorukov thanks for the quick feedback but that suggestion applies for all branches.

snicoll:
I did and forgot to act on it, Sorry.

snicoll (approved):

PR #2053, opened 2018-12-14
SPR-17600 Vararg based methods in MessageSource

ankurpathak:
It adds a vararg based getMessage methods
in MessageSource. The methods are added using
Java 8 Interface Default Methods so doesn't
need any implementation in any concrete
MessageSource.


ankurpathak:
Pull request for Jira issue:
https://jira.spring.io/browse/SPR-17600

ankurpathak:
@jhoeller Can we merge it as it will make useing MessageSource much more elegant.

ankurpathak:
@jhoeller Can you please review this request, as its a long pending issue.

bclozel:
I don't think we should add those variants as it will make IDE auto-completion noisy.
Thanks!

PR #2054, opened 2018-12-14
Optimize `CharSequence` writing to `DataBuffer` instances

bclozel:
As explained in [SPR-17558](https://jira.spring.io/browse/SPR-17558), our current implementation doesn't use optimized variants provided by `DataBuffer` implementations, like Netty's.

This PR aims at introducing a new feature in `DataBuffer` to do that.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
@@ -19,6 +19,8 @@ package org.springframework.core.codec;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
@@ -48,6 +50,9 @@ public final class CharSequenceEncoder extends AbstractEncoder<CharSequence> {
 	 */
 	public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
 
+	private final ConcurrentMap<Charset, Float> charsetToMaxBytesPerChar =
+			new ConcurrentHashMap<>(3);
+
 
 	private CharSequenceEncoder(MimeType... mimeTypes) {
 		super(mimeTypes);
@@ -75,7 +80,8 @@ public final class CharSequenceEncoder extends AbstractEncoder<CharSequence> {
 				});
 			}
 			boolean release = true;
-			DataBuffer dataBuffer = bufferFactory.allocateBuffer();
+			int capacity = calculateCapacity(charSequence, charset);
+			DataBuffer dataBuffer = bufferFactory.allocateBuffer(capacity);
 			try {
 				dataBuffer.write(charSequence, charset);
 				release = false;
@@ -90,6 +96,13 @@ public final class CharSequenceEncoder extends AbstractEncoder<CharSequence> {
 			}
 			return dataBuffer;
 		});
+	}
+
+	private int calculateCapacity(CharSequence sequence, Charset charset) {
+		float maxBytesPerChar = this.charsetToMaxBytesPerChar
+				.computeIfAbsent(charset, cs -> cs.newEncoder().maxBytesPerChar());
+		float maxBytesForSequence = sequence.length() * maxBytesPerChar;
+		return (int) Math.ceil(maxBytesForSequence);
 	}
 
 	private Charset getCharset(@Nullable MimeType mimeType) {
--- a/spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
@@ -16,8 +16,6 @@
 
 package org.springframework.core.codec;
 
-import java.nio.ByteBuffer;
-import java.nio.CharBuffer;
 import java.nio.charset.Charset;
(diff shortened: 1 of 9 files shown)


spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
```diff
@@ -75,9 +74,21 @@ public boolean canEncode(ResolvableType elementType, @Nullable MimeType mimeType
 					return Hints.getLogPrefix(hints) + "Writing " + formatted;
 				});
 			}
-			CharBuffer charBuffer = CharBuffer.wrap(charSequence);
-			ByteBuffer byteBuffer = charset.encode(charBuffer);
-			return bufferFactory.wrap(byteBuffer);
+			boolean release = true;
+			DataBuffer dataBuffer = bufferFactory.allocateBuffer();
```
bclozel: We're allocating here a buffer with the default size. The problem is, there's no consensus about what capacity should be allocated. Some implementations favor "string length x average byte count per char with that charset", others (like Netty) directly go for "string length x max byte count per char with that charset".

spring-core/src/main/java/org/springframework/core/io/buffer/DefaultDataBuffer.java
```diff
@@ -317,6 +323,43 @@ private void write(ByteBuffer source) {
 		this.writePosition += length;
 	}
 
+	@Override
+	public DataBuffer write(CharSequence charSequence, Charset charset) {
+		Assert.notNull(charSequence, "'charSequence' must not be null");
+		Assert.notNull(charset, "'charset' must not be null");
+		CharsetEncoder charsetEncoder = charset.newEncoder()
```
bclozel: Calling `charset.encode` allocates a `ByteBuffer` (inefficient?) but uses `ThreadLocalCoders.encoderFor` to get a `CharsetEncoder` from a local cache. Because `ThreadLocalCoders` is an internal API, we can't use it here.

spring-core/src/main/java/org/springframework/core/io/buffer/DefaultDataBuffer.java
```diff
@@ -317,6 +323,43 @@ private void write(ByteBuffer source) {
...
+				.onMalformedInput(CodingErrorAction.REPLACE)
+				.onUnmappableCharacter(CodingErrorAction.REPLACE);
+		CharBuffer inBuffer = CharBuffer.wrap(charSequence);
+		int estimatedSize = (int) (inBuffer.remaining() * charsetEncoder.averageBytesPerChar());
+		ensureCapacity(estimatedSize);
+		charsetEncoder.reset();
+		ByteBuffer outBuffer = asByteBuffer(this.writePosition, writableByteCount());
+		try {
+			for (; ; ) {
```
poutsma: I would take a look at `io.netty.buffer.ByteBufUtil#encodeString0`, and see what Netty does there. Their algorithm looks somewhat simpler.

spring-core/src/main/java/org/springframework/core/io/buffer/NettyDataBuffer.java
```diff
@@ -229,6 +237,51 @@ public NettyDataBuffer write(ByteBuf... byteBufs) {
...
+						outBuffer = asByteBuffer(this.byteBuf.writerIndex(), writableByteCount());
+					}
+				}
+			}
+			catch (CoderMalfunctionError ex) {
+				throw new IllegalStateException(ex);
+			}
+			writePosition(outBuffer.position());
+		}
+		return this;
```
poutsma: The non-UTF-8, non-ASCII code should really live in a base class, since it's shared between `DefaultDataBuffer` and `NettyDataBuffer`.  Or perhaps better yet: give the method a default implementation in `DataBuffer`.

spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
```diff
@@ -75,12 +79,32 @@ public boolean canEncode(ResolvableType elementType, @Nullable MimeType mimeType
...
-			ByteBuffer byteBuffer = charset.encode(charBuffer);
-			return bufferFactory.wrap(byteBuffer);
+			boolean release = true;
+			int capacity = calculateCapacity(charSequence, charset);
+			DataBuffer dataBuffer = bufferFactory.allocateBuffer(capacity);
+			try {
+				dataBuffer.write(charSequence, charset);
+				release = false;
+			}
+			catch (IllegalStateException ex) {
```
poutsma: Shouldn't this be changed to `CoderMalfunctionError`, if I remember correctly?

spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
```diff
@@ -75,12 +79,32 @@ public boolean canEncode(ResolvableType elementType, @Nullable MimeType mimeType
...
+			finally {
+				if (release) {
+					DataBufferUtils.release(dataBuffer);
+				}
+			}
+			return dataBuffer;
 		});
 	}
 
+	private int calculateCapacity(CharSequence sequence, Charset charset) {
```
poutsma: I think I would like to see a separate unit test for this method, but I can add that test myself as well.

poutsma (approved):

spring-core/src/main/java/org/springframework/core/codec/CharSequenceEncoder.java
bclozel: Ah right, missed that part. I'll force push the branch.

PR #2058, opened 2018-12-18
No need to assert `session!=null`, since `request.getSession()` alway…

AsinRay:
No need to assert `session!=null`, since `request.getSession()` always return a no-null HttpSession


pivotal-issuemaster:
@AsinRay Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2058)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2058) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@AsinRay Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2058)!

sbrannen:
> No need to assert `session!=null`, since `request.getSession()` always return a no-null HttpSession

There is unfortunately no guarantee that an invocation of `getSession()` never returns `null`.

`HttpServletRequest` is an interface which specifies the intended contract, but it does not _enforce_ that contract.

Furthermore, if the non-null check is removed the code would then throw an `IllegalStateException` for a `null` session which was previously not the case.

Since that would be a potentially breaking change for various implementations of `HttpServletRequest`, I am therefore closing this PR.

PR #2059, opened 2018-12-20
2018.12.20拉取

gaoyining:


pivotal-issuemaster:
@gaoyining Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2059)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2059) to manually synchronize the status of this Pull Request.

wsy:
I don't think you should commit localized translations of comments into master branch.

If everyone does so, every line of comment will be dozens of lines long, English, French, Italian, Germany, Chinese, Korean, Japanese…
Then each line of code is flooded by hundreds lines of comment. Good code should be self explaining ones, and comments should be supplemental.

Also, currently, I don't think modern Java IDEs are capable of handling multi-language java doc. This means if everyone comes to translate these comments, then spring-xxx-javadoc.jar will contain java doc for every language. When we use this in IDEs, we will have to turn hundreds of pages to find java doc of our own language.

On the other hand, I do encourage such translation behavior.
So I suggest you open a NEW branch for Chinese-translated java doc, and then translate English comments into Chinese comments and delete English comments on THAT BRANCH.
So that someone can use that branch to build spring-xxx-javadoc-zh-cn.jar.
Then, instead of having one spring-xxx-javadoc.jar with comments written in hundreds of languages, we can have spring-xxx-javadoc.jar with English comments and spring-xxx-javadoc-zh-cn.jar with Chinese comments and spring-xxx-javadoc-fr-fr.jar with French comments, etc.
I think that should be a good approach.

snicoll:
@gaoyining there's a lot of different things here and it is not entirely obvious what you're trying to contribute. I'd ask you to split each topic in its dedicated PR but please take into account we will not accept non english comment in the codebase. Thanks for the PR, in any case.

PR #2063, opened 2018-12-28
Use try-with-resources language construct where feasible

igor-suhorukov:


snicoll (approved):

snicoll:
On close inspection, this PR brings a significant of build warning we need to handle

```/Users/snicoll/workspace/pivotal/spring-framework/spring-core/src/test/java/org/springframework/util/SocketUtilsTests.java:83: warning: [try] auto-closeable resource socket is never referenced in body of corresponding try statement```

I've had a look at some of those and there is a way to avoid them with a template method but it requires more code than it should. 

Let's see what the rest of the team thinks.

sbrannen:
> `/Users/snicoll/workspace/pivotal/spring-framework/spring-core/src/test/java/org/springframework/util/SocketUtilsTests.java:83: warning: [try] auto-closeable resource socket is never referenced in body of corresponding try statement`

For the "unused" sockets in `SocketUtilsTests`, I'd just suppress the warning about the unused variable.

For each unclosed `ApplicationContext` in a test, we often just suppress the warnings as well. Though, we do sometimes actually close the context. In the end, I don't suppose it makes much of a difference for the tests unless someone suspects that a particular test `ApplicationContext` should clean up its resources via an explicit invocation of `close()`.

> I've had a look at some of those and there is a way to avoid them with a template method but it requires more code than it should.

If we have to resort to the introduction of template methods in production code just to _benefit_ from the try-with-resources construct, I would generally be against that. For such cases, I think we should decide on a case-by-case basis.

sbrannen:
This has been incorporated into `master` in a slightly modified form in 8099fc8178d73375c709e795370ad1767f3f5b01.

Thanks

PR #2065, opened 2018-12-31
SPR-16886 Provide two mechanisms for SQL Server sequence incrementers.

tkvangorder:
This PR is to address a deadlocking issue with MS SQL Server and is related to https://jira.spring.io/browse/SPR-16886. Please see that issue for details.

I tried to ensure that this fix does not break downstream projects that leverage the incrementer on SQL Server.

This means that all versions of SQL Server are still going to use a sequence table with the difference being that the incrementer uses a reaper interval to delete records out of the sequence table.

I have also included an incrementer that can leverage native database sequences in 2012 and higher. It will be at the discretion of each downstream project as to when/if they want to leverage the new incrementer.

I am open to any feedback or changes you might want to make. I had toyed with requiring a platform transaction manager being injected into the incrementer (and creating a new transaction when the reaping interval was reached) as an alternative to using a thread, but I felt that was a more invasive change to downstream projects.

I have also updated my little test harness to use "copies" of the incrementers I have submitted in this pull request: https://github.com/tkvangorder/jdbc-sqlserver-incrementer

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/SqlServerMaxValueIncrementer.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/SqlServerMaxValueIncrementer.java
@@ -16,17 +16,27 @@
 
 package org.springframework.jdbc.support.incrementer;
 
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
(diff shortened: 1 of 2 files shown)


sbrannen:
Closing as superseded by the following issues which have been closed.

- https://github.com/spring-projects/spring-batch/issues/1448
- #29447 

If you feel that certain issues still need to be addressed, please open a new issue.

PR #2069, opened 2019-01-03
Extract Method::getParameterTypes out of the loop where possible

stsypanov:
Method::getParameterTypes returns copy of Method.parameterTypes.
Often this can be extracted out of the loop


jhoeller:
I've only selectively refined `ReplaceOverride` and `BridgeMethodResolver` since the shortcut evaluation in the other clauses seems more important to me here, avoiding many calls to `Method.getParameterTypes()` to begin with: e.g. if no interfaces are declared (or if there is no method of the given name). To some degree, I find readability more important than avoiding the array copy, so I'm also not keen on upfront checks or lazy resolution of a local array copy when actually needed.

Thanks for pointing these out, in any case!

PR #2071, opened 2019-01-06
SPR-8335: fixed default-lazy-init processing when xsd validation is disabled

MichelSchudel:
The xml parser does not fill in defaults provided in the xsd when validation is disabled. 

As a result, attributes like default-lazy-init will not receive the value "default" but an empty
string value. This also applies for other default attributes in the root or parent elements.

The BeanDefinitionParserDelegate did not take this into account. So, we should also check on an empty string value, which is semantically the same as the value "default".

Issues: SPR-8335

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
@@ -98,6 +98,8 @@ public class BeanDefinitionParserDelegate {
 
 	public static final String DEFAULT_VALUE = "default";
 
+	private static final String EMPTY_STRING = "";
+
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@MichelSchudel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2071)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2071) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@MichelSchudel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2071)!

sdeleuze:
I have merged a polished version of your commit, thanks.

PR #2072, opened 2019-01-07
Custom resolver should be first add in 

linxingwu:
Custom resolver should be first add in since they may want overwrite the default behavior,especially map resolver.
I want my @RequestParam Map<String,Object> contains some Integer value instead of all String value. I wrote my own resolver extends RequestParamMapMethodArgumentResolver  , but RequestParamMapMethodArgumentResolver will be the first resolver be detected, so my resolver won't work. So I add my own annotation and overwrite the  supportsParameter method , then MapMethodProcessor will be the first . I have no way to deal this solution. So I think it's better to add custom resolver first, then the default resolver.


pivotal-issuemaster:
@linxingwu Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2072)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=2072) to manually synchronize the status of this Pull Request.

PR #2078, opened 2019-01-08
SPR-17645: Fix ClassCastException in FormHttpMessageConverter

wsy:
We should not cast MultiValueMap<String, ?> to MultiValueMap<String, String>


wsy:
[JIRA SPR-17645](https://jira.spring.io/browse/SPR-17645)

PR #2080, opened 2019-01-08
toString for AbstractSqlParameterSource

schauder:
Issue: SPR-9753


jhoeller:
Finally getting to this, I took the opportunity to do a wider revision and eventually went with a different format with Map-style brackets and reverse lookup of SQL type constant names. To be committed soon.

PR #18152, opened 2019-01-11
ResourceLoaderAware is not a marker interface

mhewedy:


jhoeller:
Well spotted! I've revised that section and the corresponding discussion in the beans chapter a bit further, making it clear that those 'aware' interfaces are callback interfaces. Thanks for the pull request, in any case!

PR #21855, opened 2019-01-11
Correctly parse property name in path "map[key[foo]]"

mkurz:
We have a `Bean` that contains a `Map`. In this `Map` we want to save properties with property names which contain an index that is relevant for our application only, like `key5[foo]`. For spring that index doesn't matter, it should just treat `key5[foo]` as property name.
However spring-bean is not able to correctly retrieve the property name, it tries to retrieve `key5[foo` (missing square bracket at the end). That's because spring just looks for the first occurrence of `]` and thinks it's the closing square bracket of the property name, however in our case that square bracket is part of the property name, and the relevant square bracket for spring (to actually close the property name) comes afterwards.

- [x] Added tests
- [x] Signed the CLA.

Would be nice if this could be backported to `5.1.x`.
Thanks!

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
@@ -933,7 +933,7 @@ public abstract class AbstractNestablePropertyAccessor extends AbstractPropertyA
 			int keyStart = propertyName.indexOf(PROPERTY_KEY_PREFIX, searchIndex);
 			searchIndex = -1;
 			if (keyStart != -1) {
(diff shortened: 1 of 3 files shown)


mkurz:
The Codacy check fails because the `switch` statement I added does not have a `default`. However that isn't necessary for the logic I added.
Do I really have to add an empty/no-op `default` to that `switch` statement?

mkurz:
@jhoeller any chance you will assign this issue to the `5.1.5` milestone? I think the bug here - as well as the fix - is obvious. Thanks!

jhoeller:
On review, I'll rather make this a 5.2 topic since the bean property binding is a high-risk area for regressions. Even if some such paths are not meant to be valid from our side, people may encounter e.g. property paths with incomplete nested brackets which we would possibly parse differently then.

PR #22246, opened 2019-01-12
References of CountingBeforeAdvice target its previous location

diguage:
I have some questions about AOP. So I research the source codes.

I want to debug the test source codes of AOP. I found the references of `CountingBeforeAdvice` was missed, and I found `org.springframework.tests.aop.advice.CountingBeforeAdvice` is in the repo.

Maybe the reference of `CountingBeforeAdvice` is incorrect. Please check it. Thanks a lot.


diguage:
@snicoll 
What's wrong with the PR? 

Happy Chinese New Year! 😆

snicoll:
@diguage sorry it took so long to process it.

PR #22275, opened 2019-01-17
ResponseEntityExceptionHandler add to  generic type 

cheese10yun:
If the ResponseEntityExceptionHandler class is extedns, then the return type is `ResponseEntity <Object>`

This part was inconvenient. Thanks to Spring Open Source


rstoyanchev:
I see no good reason for `ResponseEntityExceptionHandler` to be parameterized. This class is invoked by framework infrastructure and never by the application directly. 

Also please, do not submit unrelated commits in a single pull request.

atomcat1978:
THere is nothing that holds people back form (mis)usingit, and I just saw a project, which is implementing this class.

In addition, if U consider, making it generics also makes the framework infrastructure code safer....

PR #22287, opened 2019-01-21
Fix truncation of response body in AbstractMessageConverterMethodProcessor

lifenjoy51:
It seems to be minor mistakes from SPR-17254.

Second parameter of LogFormatUtils.formatValue is (boolean limitLength). It should be false at TRACE. So !traceOn is correct. Other code that call formatValue method are all !traceOn

related issue : [SPR-17254] #21787


rstoyanchev:
Thanks for the fix, and apologies for losing the author on it. I was trying something new and din't quite work out as expected.

PR #22291, opened 2019-01-22
Do concatenation with String::substring instead of StringBuilder::append in short append chain

stsypanov:
Changed code is not only shorter and more simple, but also faster and consumes less memory, see measurements here: https://youtrack.jetbrains.com/issue/IDEA-205676

### DIFF
--- a/spring-core/src/main/java/org/springframework/asm/Type.java
+++ b/spring-core/src/main/java/org/springframework/asm/Type.java
@@ -505,11 +505,7 @@ public final class Type {
     if (sort == OBJECT) {
       return valueBuffer.substring(valueBegin - 1, valueEnd + 1);
     } else if (sort == INTERNAL) {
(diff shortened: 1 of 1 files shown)


jhoeller:
You have a point there. However, this code is a repackaged fork of https://gitlab.ow2.org/asm/asm/commits/master, so any such refactorings should really start there... and we'll automatically pick it up for our fork at our next sync point then.

stsypanov:
oh, I wasn't aware of that. Should I then make similar pull-requests into that repo?

jhoeller:
Indeed, ASM itself would certainly benefit from such fine-grained optimizations. They applied quite a few such refactorings last year already, and I'm sure they'll appreciate some further love for the detail there.

stsypanov:
Done for both PRs.

PR #22314, opened 2019-01-28
Use for-each instead of counting loop when iterating arrays

stsypanov:
subj


snicoll:
Duplicate of #22315

stsypanov:
yes, but this one mistakenly includes unrelated commit. This is why I closed this one.

snicoll:
You can push force on an existing branch and it will update an existing PR, you shouldn't have to create another one.

stsypanov:
Good, I'll keep this in mind for future!

PR #22315, opened 2019-01-28
Use for-each instead of counted loop when iterating over arrays

stsypanov:
Just some cosmetic changes


jhoeller:
Good catch, I've partially rolled those into a polishing commit of mine.

Note that there's one CGLIB class in there (forked from https://github.com/cglib/cglib) which we won't modify on our end beyond the necessary patches. I've also kept `MBeanClientInterceptor` as-is largely for code alignment between both of its delegate methods there.

PR #22316, opened 2019-01-28
Replace c-style array declaration with Java-style

stsypanov:
subj


jhoeller:
I'm afraid both of those source files are forks from CGLIB classes where we intend to strictly follow the original source code except for our necessary patches. You could report those to the CGLIB project at https://github.com/cglib/cglib/issues...

PR #22320, opened 2019-01-29
RestTemplate adds duplicate accept header information 

geminiKim:
Hi :)

In 4.3.x version, I found a bug that caused the `Accept` header to be duplicated. 
I have added a validation test and fixed the bug, if you could review this PR then I appreciate it.

It seems to have been fixed in 5.0.x and above. It appears to be occurring in versions below 4.3.x.

If this PR is not in the correct format, please let me know. I will fix it.
Thank you.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jhoeller:
This got addressed for 5.1 through #21231. @bclozel, if we decide to backport a variant of that change, let's backport it to 5.0.13 as well as 4.3.23.

spring-issuemaster:
Fixed via 3e1b3c32e3dfae52c2537fe9249ec3e509417d60

bclozel:
I've picked up your commits in a single change with 3e1b3c3.
I didn't backport the original fix since it was using `java.util.stream.Stream` and rewriting a lot of tests as a more general revision. The 5.0.x backport will be taken care of in #22401.

Thanks for your contribution!

PR #22324, opened 2019-01-30
modify-for-loop-mode

ZhZGod:
It modified the way of the for-loop in the method parseMetaElements 
of the class BeanDefinitionParserDelegate 
form for (int i = 0; i < nl.getLength(); i++) 
to for (int i = 0, j = nl.getLength(); i < j; i++).
It makes you only call the method nl.getLength() once no matter how many times the for-loop used.There are a lot of similar places in this code.
I think which should be optimised is:
1.The depth of VM stack can be reduced.
2.Some complicated methods would be waste time to calculate the length.
It also can be optimeised.I will modify the for-loop in the project if you identified the optimization

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
@@ -645,7 +645,7 @@ public class BeanDefinitionParserDelegate {
 
 	public void parseMetaElements(Element ele, BeanMetadataAttributeAccessor attributeAccessor) {
 		NodeList nl = ele.getChildNodes();
-		for (int i = 0; i < nl.getLength(); i++) {
+		for (int i = 0, j = nl.getLength(); i < j; i++) {
 			Node node = nl.item(i);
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ZhZGod Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22324)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22324) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ZhZGod Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22324)!

ZhZGod:
WHY？The first optimization is wrong. But the second is feasible. Does each loop call the nl. getLength () method once?

ZhZGod:
How is this suggestion inappropriate? Can you give some reasons? @jhoeller

snicoll:
Thanks for the PR and sorry it took so long to give you some feedback. We're not keen to process with this, in this particular case, or in the core framework at large. `getLength()` is cached effectively.

PR #22327, opened 2019-01-31
Simplification in constructor of MvcUriComponentsBuilder.MethodArgumentBuilder

stsypanov:
Accodring to JLS newly created array is populated with default values for array's type. In case of reference type it is `null` so manual population of array is pointeless in this case.


rstoyanchev:
This has been merged.

Issue #22268, opened 2019-01-16
Ability to set max frame length in ReactorNettyWebSocketClient


benitocm:
Hi,

We are obtaining this exception  _Max frame length of 65536 has been exceeded_  in a ReactorNettyWebSocketClient that receive a big payload. Having a look to the code, it seems that all is ready for having this feature in the client side but the websocket  method without the additional params is being called instead of the one that provides a parameter for the max frame length is provided. In summary, currently this is being used  https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java#L84
but this should be used:
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClient.java#L888

The versions that we are using are the following:
Spring Boot 2.1.1 with Netty 
Reactor Netty 0.8.3

sdeleuze:
@rstoyanchev What do you think of adding `default int getMaxFramePayloadLength()` to `WebSocketHandler` that would return `65536` and allow customization if needed?

rstoyanchev:
This is already supported on the server side:
https://github.com/spring-projects/spring-framework/blob/4fdfa4246ff24872e2e17d86cc86dd8471f0da17/spring-webflux/src/main/java/org/springframework/web/reactive/socket/server/upgrade/ReactorNettyRequestUpgradeStrategy.java#L78-L85

A similar property on `ReactorNettyWebSocketClient` makes sense. Notice that it's also passed into the session which uses it for frame aggregation, so it should be passed there too, in addition to the `sendWebSocket` method call.

rstoyanchev:
This is superseded by #22268.

philsttr:
#22367

PR #22328, opened 2019-01-31
Adds protected method to createWebSocketSession

spencergibb:
https://github.com/spring-projects/spring-framework/issues/20776 added the ability to set max frame size for netty websocket support. 

This PR adds the ability to override `createWebSocketSession()` to the new constructor may be called. Users want to be able to do this in Spring Cloud Gateway. Rather than overriding all of execute, we only have to override creation of the class.

See https://github.com/spring-cloud/spring-cloud-gateway/issues/163


rstoyanchev:
@spencergibb if the end goal is to be able to set the max frame size, then maybe #22268 is what you are looking for?

spencergibb:
@rstoyanchev yes. Users want to set it. Feel free to close this if that will be taken care of another way.

rstoyanchev:
Okay I'll leave it for #22268 then since it addressed the same issue. I just wanted to make sure that was the primary reason.

benitocm:
Hi,

Please, could you tell me when will this be available? 

Thanks in advance

PR #22330, opened 2019-02-01

stsypanov:
Employment of `StringJoiner` greatly simplifies code by dropping `if` statement and subsequence simplification of `fori` to `foreach`.

As of performance currently (JDK 8) `StringJoiner` is splightly slower than `StringBuilder`, but much faster (and meory-saving) on JDK 11.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/propertyeditors/ClassArrayEditor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/propertyeditors/ClassArrayEditor.java
@@ -17,6 +17,7 @@
 package org.springframework.beans.propertyeditors;
 
 import java.beans.PropertyEditorSupport;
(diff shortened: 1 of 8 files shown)


darsey:
Hi @stsypanov, i see you left all the StringJoiner variables named as "sb" (probably because previously was used a StringBuilder), maybe it would be more intuitive to change them to "sj".

stsypanov:
Hello @darsey you are right, previously it was `StringBuilder`. And indeed it would be nice to have renaming here. For the first commit I've intentionally left variables' name intact to highlight only the desired changes. If there'll be no other concerns regarding migration from `StringBuider` to `StringJoiner` all push one more commit here with requested renaming.

jhoeller:
Good point, we should start using `StringJoiner` across the codebase, also for new code in 5.2.

PR #22334, opened 2019-02-02
query(PSC, PSS RSE) missing from JdbcOperations

marschall:
JdbcTemplate.query(PSC, PSS RSE) is the only public query method on
JdbcTemplate that is not on JdbcOperations.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcOperations.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcOperations.java
@@ -300,6 +300,23 @@ public interface JdbcOperations {
 	<T> T execute(String sql, PreparedStatementCallback<T> action) throws DataAccessException;
 
 	/**
+	 * Query using a prepared statement, allowing for a PreparedStatementCreator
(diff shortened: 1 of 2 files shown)


jhoeller:
This is intentional: The query methods on `JdbcOperations` take a `PreparedStatementCreator` or a `PreparedStatementSetter`... but not both at the same time since a user is only meant to specify one or the other. The combined method on `JdbcTemplate` is primarily an internal delegation convenience that happens to be publicly exposed on the template class as well.

marschall:
I have a use case that would benefit from having both a `PreparedStatementCreator` and a `PreparedStatementSetter`. If you want to use [Oracle explicit statement caching](https://docs.oracle.com/en/database/oracle/oracle-database/18/jjdbc/statement-and-resultset-caching.html#GUID-DFBC7F09-5F27-42E1-8044-24733A6AE5F8) you need to use different methods for creating and closing a `PreparedStatement`, this can be achieved with a `PreparedStatementCreator`.
I would be nice if the binding of parameters would not also have to be implemented in this class and could instead be moved to a dedicated `PreparedStatementSetter`.

PR #22335, opened 2019-02-02
Add CompositeTaskDecorator

ttddyy:
Hi,

Currently, task executors only accept single `TaskDecorator` instance.
When there are multiple `TaskDecorator` beans to apply, I need to write such composite decorator for now. It would be helpful if it comes from spring.

Thanks,

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/task/CompositeTaskDecorator.java
+++ b/spring-core/src/main/java/org/springframework/core/task/CompositeTaskDecorator.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.task;
+
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


snicoll:
> TaskExecutionAutoConfiguration can accept multiple of TaskDecorator beans 

Thanks for the PR. I am not entirely convinced by that argument yet. If you believe this is something Spring Boot should do, it is not mandatory that such implementation should be available at the framework level so I suggest we restrict the scope to a custom multi decorator arrangement.

ttddyy:
Hi @snicoll
so, you mean instead of adding `CompositeTaskDecorator` in spring-framework, putting it in spring-boot and accommodate multiple task decorators there?

I think it's a question of how much support do you provide for multiple task decorators.
Spring-boot only or make it in spring and allow non spring-boot environment to get benefit.

Let me know, I can update the PR if it is better in spring-boot only.

Thanks,

snicoll:
Unfortunately, that's not what I meant. What I meant is that there is no guarantee that Spring Boot would add such feature once/if this PR is merged and I suggest to keep the discussion focused on the composite itself.

ttddyy:
Ah, ok, thanks for pointing out.

Yeah, I'm aware of that making multiple decorator support in spring-boot is a separate topic.
It was just a reason behind why I needed the composite decorator.

So, to clarify, purpose of this PR is to add `CompositeTaskDecorator` in spring-framework, that's all about it.

ttddyy:
I have updated the PR comment, accordingly.

Thanks,

jhoeller:
Design-wise, with the language options that we have now, we would rather introduce a static `asComposite(TaskDecorator...)` method on the `TaskDecorator` interface itself... or an `andThen` arrangement along the lines of `java.util.function.Function`. So we would not rely on a public composite implementation class here, at least not for our core framework purposes.

However, since we don't ship a library of decorator implementations out of the box, I'm not convinced that this is necessary. I rather see `TaskDecorator` as a general purpose extension hook, not as a programming model component. If it turns out to become an arrangement of provided components e.g. in Boot, I'm open to introducing methods along the lines above.

PR #22337, opened 2019-02-03
Small fix in DeferredResultReturnValueHandlerTests regarding CompletableFuture tests

mhewedy:
While I was investigating regarding A problem I am facing (asked as [SO](https://stackoverflow.com/q/54507134/171950) question here), I noticed that there's some small issue that regarding using `SettableListenableFuture` in `CompletableFuture` test cases.


snicoll (approved):

snicoll:
Thank you @mhewedy

Issue #22338, opened 2019-02-04
spring-context-indexer seems to do nothing


the4yeast:
I've included the dependency as per https://docs.spring.io/spring/docs/5.1.2.RELEASE/spring-framework-reference/core.html#beans-scanning-index into my Boot app. 

The app is fairly trivial, with one controller and one service.

When I build my fat jar and unpack it, I do not see the `spring.components` file in `META-INF`. 
Is there a way for me to get more insight into the process? How do I know the indexer is invoked?

vpavic:
Are you using Gradle? If so, `annotationProcessor` configuration should be used to add `spring-context-indexer` dependency - I've opened #22339 to address this in reference manual.

snicoll:
Thanks @vpavic, it could be related though I am quite surprised that Gradle stopped handling `compileOnly` in the `4.x` line. 

@the4yeast regardless of whether this is your problem or not, as mentioned in [the guidelines for contributing](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#discuss), we prefer to use the issue tracker only for bugs and enhancements. Feel free to update this issue with a link to the re-posted StackOverflow question (so that other people can find it) or add some more details if you feel this is a genuine bug.

the4yeast:
thanks @vpavic 

sorry @snicoll, I didn't realise that was the process. If/when I have the next question, I'll ask on SO.

This is the question in SO: https://stackoverflow.com/questions/54524942/spring-context-indexer-seems-to-do-nothing-on-gradle-4-6. I'll follow up there.

PR #22339, opened 2019-02-04
Improve spring-context-indexer documentation

vpavic:
The reference manual currently has the following snippet as an example of how to enable `spring-context-indexer` in Gradle based build:

```groovy
dependencies {
    compileOnly("org.springframework:spring-context-indexer:5.1.4.RELEASE")
}
```

This doesn't produce `spring.components` file using recent Gradle releases (I've verified with `4.10.3` and `5.1.1`), since `annotationProcessor` configuration should be used instead.

Possibly related to #22338.

### DIFF
--- a/src/docs/asciidoc/core/core-beans.adoc
+++ b/src/docs/asciidoc/core/core-beans.adoc
@@ -5958,7 +5958,7 @@ The following example shows how to do so with Gradle:
 [subs="verbatim,quotes,attributes"]
 ----
 	dependencies {
-		compileOnly("org.springframework:spring-context-indexer:{spring-version}")
+		annotationProcessor("org.springframework:spring-context-indexer:{spring-version}")
 	}
 ----
 


snicoll:
@vpavic I've created a project with start.spring.io and the following build

```
buildscript {
	ext {
		springBootVersion = '2.1.2.RELEASE'
	}
	repositories {
		mavenCentral()
	}
	dependencies {
		classpath("org.springframework.boot:spring-boot-gradle-plugin:${springBootVersion}")
	}
}

apply plugin: 'java'
apply plugin: 'org.springframework.boot'
apply plugin: 'io.spring.dependency-management'

group = 'com.example'
version = '0.0.1-SNAPSHOT'
sourceCompatibility = '1.8'

repositories {
	mavenCentral()
}

dependencies {
	implementation 'org.springframework.boot:spring-boot-starter'
	compileOnly 'org.springframework.boot:spring-boot-configuration-processor'
	compileOnly 'org.springframework:spring-context-indexer'
}
```

I've added a `@ConfigurationProperties` (flagged as `@Component`) to test both annotation processor and both generates content just fine. The `compileOnly` flag has been deprecated though and we missed the update here so thanks for the PR, in any case.

I am just curious to understand why it doesn't work for you with Gradle 4.x. In any case, that's not related to Spring in any way I think as those are regular annotation processors.

vpavic:
You're right @snicoll, it does indeed work with Gradle `4.10.3` and `compileOnly`. I must've messed something up while configuring Gradle during initial testing.

So in summary:
- declaring `spring-context-indexer` as `compileOnly` dependency works in Gradle `4.10.3` but doesn't in Gradle `5.2`
- declaring `spring-context-indexer` as `annotationProcessor` dependency works both in Gradle `4.10.3` and Gradle `5.2`

> The `compileOnly` flag has been deprecated though

I don't know where did you get this from - `compileOnly` dependency configuration isn't deprecated.

izeye:
@vpavic I guess @snicoll was referring to the following message which will be printed when using `--warning-mode all` option:

```
Detecting annotation processors on the compile classpath has been deprecated. Gradle 5.0 will ignore annotation processors on the compile classpath. The following annotation processors were detected on the compile classpath: 'org.springframework.boot.configurationprocessor.ConfigurationMetadataAnnotationProcessor' and 'org.springframework.context.index.processor.CandidateComponentsIndexer'.  Please add them to the annotation processor path instead. If you did not intend to use annotation processors, you can use the '-proc:none' compiler argument to ignore them.
```

PR #22367, opened 2019-02-06
Added configurable maxFramePayloadLength to ReactorNettyWebSocketClient

envious:
Currently the ReactorNettyWebSocketClient (RNWSC) doesn't allow for a configurable maxFramePayloadLength. The HttpClient used by the RNWSC allows for a configurable maxFramePayloadLength however, the method .websocket() used by the RNWSC ignores whatever is set by the HttpClient and uses a default value of 65536. As shown below:

**ReactorNettyWebSocketClient invokes .websocket()**

    .websocket(StringUtils.collectionToCommaDelimitedString(handler.getSubProtocols()))

**The method invoked on the HttpClient ignores the value you may have set in your HttpClient**

	public final WebsocketSender websocket(String subprotocols) {
		return websocket(subprotocols, 65536);
	}

I've solved the issue by using the overloaded .websocket() method on the HttpClient which takes a maxFramePayloadLength.

    .websocket(StringUtils.collectionToCommaDelimitedString(handler.getSubProtocols()), 
    getMaxFramePayloadLength())

The maxFramePayloadLength is now configurable in the RNWSC.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java
@@ -16,14 +16,8 @@
 
 package org.springframework.web.reactive.socket.client;
 
-import java.net.URI;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import reactor.core.publisher.Mono;
-import reactor.netty.http.client.HttpClient;
-import reactor.netty.http.websocket.WebsocketInbound;
-
 import org.springframework.core.io.buffer.NettyDataBufferFactory;
 import org.springframework.http.HttpHeaders;
 import org.springframework.util.Assert;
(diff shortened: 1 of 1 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java
```diff
@@ -102,6 +102,14 @@ public HttpClient getHttpClient() {
...
 	}
 
+	public int getMaxFramePayloadLength() {
+		return maxFramePayloadLength;
+	}
+
+	public void setMaxFramePayloadLength(int maxFramePayloadLength) {
+		this.maxFramePayloadLength = maxFramePayloadLength;
+	}
+
```
rstoyanchev: In case you want to update your PR, these would need to be ordered earlier where getters an setters are, i.e. before implementation methods. Probably right ahead after getHttpClient(). Setters come first and some Javadoc on such a property is required.

envious: I'll go through the contribution guidelines and update this PR. Thanks for pointing that out.

envious: @rstoyanchev PR comments have been addressed.

philsttr:
I noticed that this PR did not change [`ReactorNettyWebSocketClient`](https://github.com/spring-projects/spring-framework/blob/0fc0849c0a13eb504cd2f308d3824148f8f36806/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java#L116) to call the constructor for [`ReactorNettyWebSocketSession` that takes the `maxFramePayloadLength`](https://github.com/spring-projects/spring-framework/blob/0fc0849c0a13eb504cd2f308d3824148f8f36806/spring-webflux/src/main/java/org/springframework/web/reactive/socket/adapter/ReactorNettyWebSocketSession.java#L61-L63)

It seems odd that the `ReactorNettyWebSocketSession.maxFramePayloadLength` is not set on the client side, but is set on the server-side.  i.e. `ReactorNettyWebSocketSession.maxFramePayloadLength` will keep the default value (64k) on the client side regardless of how `ReactorNettyWebSocketClient.maxFramePayloadLength` is configured.

Is this intended?

Will something bad happen if `ReactorNettyWebSocketClient.maxFramePayloadLength` has a different value than `ReactorNettyWebSocketSession.maxFramePayloadLength`  ?

rstoyanchev:
Good catch @philsttr, thanks.

eliquinox:
Is it possible to use this feature now or do we have to wait for the milestone?

benitocm:
Please, how can I check when this ticket will be available?

snicoll:
@benitocm perhaps you're looking at this from your phone? Please use a desktop, this issue is closed and assigned to `5.2 M1` as indicated by the milestone attribute. If you click on it you can see it is assigned to April 5th at the moment

PR #22382, opened 2019-02-07
Polish warning message

artembilan:
The second part of the warning message in the `TemporaryReplyChannel` is stuck to the colon
making it hard to read:
```
[org.springframework.messaging.core.GenericMessagingTemplate$TemporaryReplyChannel] - Reply message received but the receiving thread has exited due to an exception while sending the request message:ErrorMessage [payload=org.springframework.integration.transformer.MessageTransformationException: 
```


snicoll:
Good catch, thanks Artem.

PR #22385, opened 2019-02-07
Ensure indexer gracefully handle missing meta annotations

vpavic:
If `spring-context-indexer` is used in an application that uses `@javax.transaction.Transactional`, compilation will fail due to `javax.interceptor.InterceptorBinding` not being available.

This can be easily reproduced with a project as simple as:

```gradle
plugins {
    id 'java'
    id 'io.spring.dependency-management' version '1.0.6.RELEASE'
    id 'org.springframework.boot' version '2.1.2.RELEASE'
}

sourceCompatibility = '1.8'

repositories {
    mavenCentral()
}

dependencies {
    implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
    implementation 'com.h2database:h2'
    annotationProcessor 'org.springframework:spring-context-indexer'
}
```

```java
@SpringBootApplication
@javax.transaction.Transactional
public class DemoApplication {

        SpringApplication.run(DemoApplication.class, args);
    }

}
```

```bash
$ ./gradlew compileJava 

> Task :compileJava FAILED
error: cannot access InterceptorBinding
  class file for javax.interceptor.InterceptorBinding not found
  Consult the following stack trace for details.
  com.sun.tools.javac.code.Symbol$CompletionFailure: class file for javax.interceptor.InterceptorBinding not found
1 error

* What went wrong:
Execution failed for task ':compileJava'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
1 actionable task: 1 executed
```

This PR addresses the problem by adding a classpath presence check for annotation before proceeding to collect its stereotypes.

### DIFF
--- a/spring-context-indexer/spring-context-indexer.gradle
+++ b/spring-context-indexer/spring-context-indexer.gradle
@@ -4,5 +4,6 @@ dependencies {
 	testCompile(project(":spring-context"))
 	testCompile("javax.inject:javax.inject:1")
 	testCompile("javax.annotation:javax.annotation-api:1.3.2")
+	testCompile("javax.transaction:javax.transaction-api:1.3")
 	testCompile("org.eclipse.persistence:javax.persistence:2.2.0")
 }
--- a/spring-context-indexer/src/main/java/org/springframework/context/index/processor/TypeHelper.java
+++ b/spring-context-indexer/src/main/java/org/springframework/context/index/processor/TypeHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.context.index.processor;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.AnnotationMirror;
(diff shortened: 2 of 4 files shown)


spring-context-indexer/src/main/java/org/springframework/context/index/processor/IndexedStereotypesProvider.java
```diff
@@ -60,12 +61,22 @@ public IndexedStereotypesProvider(TypeHelper typeHelper) {
...
-			if (next != null) {
+			if (next != null && isPresent(next)) {
 				collectStereotypesOnAnnotations(seen, stereotypes, next);
 			}
 		}
 	}
 
+	private boolean isPresent(Element element) {
+		try {
+			Class.forName(element.toString(), false, getClass().getClassLoader());
```
snicoll: I am not keen to do `Class.forName` in here, that's quite unusual.

snicoll (commented):
Thanks for the PR. 

I've actually noticed that issue working on a prototype for another feature on `master`. I don't like the idea of using reflection and perhaps we could catch an exception when collection meta annotations on an annotation that is not present? If we can narrow the catch to that specific use case, I think that is better than using `Class.forName`.

Thoughts?

vpavic:
> If we can narrow the catch to that specific use case, I think that is better than using `Class.forName`.
>
> Thoughts?

Sounds good to me - I'll try to update the PR accordingly at some point today.

vpavic:
I've updated the PR so that we now catch an exception when collecting meta annotations.

snicoll:
Thanks @vpavic your fix was almost identical to the one I had in my spike for another feature.

PR #22394, opened 2019-02-10
fix typo in reference WebSocket documentation

boojongmin:
changed interface type from WebSocketConfigurer to WebSocketMessageBrokerConfigurer.

`WebSocketConfigurer` have not  ` registerStompEndpoints(StompEndpointRegistry registry)`

```java
public interface WebSocketConfigurer {
    void registerWebSocketHandlers(WebSocketHandlerRegistry var1);
}
```

`WebSocketMessageBrokerConfigurer` have ` registerStompEndpoints(StompEndpointRegistry registry)`

```java
public interface WebSocketMessageBrokerConfigurer {
    default void registerStompEndpoints(StompEndpointRegistry registry) {
    }
    // ....
```

### DIFF
--- a/src/docs/asciidoc/web/websocket.adoc
+++ b/src/docs/asciidoc/web/websocket.adoc
@@ -663,7 +663,7 @@ The following example shows how to do so in Java configuration:
 ----
 	@Configuration
 	@EnableWebSocket
-	public class WebSocketConfig implements WebSocketConfigurer {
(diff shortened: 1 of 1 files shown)


rstoyanchev:
The `@Enable` annotation is also wrong. I've fixed both in a separate commit. Thanks for reporting.

PR #22404, opened 2019-02-11
Extract duplicated code into a separate method

stsypanov:
subj


jhoeller:
Looks worthwhile indeed.

Quick side note: There's no need to declare `@NonNull` explicitly, our core packages are all declared as `@NonNullApi` at the package level for a non-null default rule.

stsypanov:
`@NonNull` removed.

Issue #22475, opened 2019-02-26
Unwrap CompletionException for MVC handlers that return CompletableFuture


sviperll:
CompletableFuture exceptions are always wrapped into CompletionException.
We need to unwrap CompletionException to properly handle exceptions in MVC handlers that return CompletableFuture.
Otherwise it's impossible to implement `@ExceptionHandler` in `@ControllerAdvice` to handle underlying exception.
See https://stackoverflow.com/questions/49676889/spring-controller-advice-does-not-correctly-handle-a-completablefuture-completed

snicoll:
Duplicate of #22475

@sviperll if you intend to submit a PR, an issue is no necessary anymore.

PR #22476, opened 2019-02-26
Fix CompletableFuture exception handling

sviperll:
CompletableFuture exceptions are always wrapped into CompletionException.
We need to unwrap CompletionException to properly handle exceptions in MVC handlers that return CompletableFuture.
See https://stackoverflow.com/questions/49676889/spring-controller-advice-does-not-correctly-handle-a-completablefuture-completed

#22475

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/DeferredResultMethodReturnValueHandler.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/DeferredResultMethodReturnValueHandler.java
@@ -16,6 +16,7 @@
 
 package org.springframework.web.servlet.mvc.method.annotation;
 
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.function.BiFunction;
 
@@ -93,6 +94,9 @@ public class DeferredResultMethodReturnValueHandler implements HandlerMethodRetu
 		DeferredResult<Object> result = new DeferredResult<>();
 		future.handle((BiFunction<Object, Throwable, Object>) (value, ex) -> {
 			if (ex != null) {
+				if (ex instanceof CompletionException) {
+					ex = ex.getCause();
+				}
 				result.setErrorResult(ex);
 			}
 			else {


pivotal-issuemaster:
@sviperll Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22476)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22476) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sviperll Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22476)!

jazdw:
Can we get this merged? FWIW, it looks good to me.

rstoyanchev:
I've resolved this in a separate commit also adding a check to ensure the cause is not null. Thanks for submitting a PR and for pointing this out!

whiskeysierra:
Is this fix already part of a release?

bclozel:
Yes, Spring Framework 5.1.7

Blackdread:
@bclozel weird, I am using Spring Boot v2.1.9.RELEASE but still have this error (dependency of spring is 5.1.10)

maverick1601:
I have the same problem, when returning e.g. `CompletableFuture<ResponseEntity<Void>>`from a controller method. The debugger revealed, that the `DeferredResultMethodReturnValueHandler`isn't used in that case, since `ResponseBodyEmitterReturnValueHandler`is kicking in instead.

rstoyanchev:
@maverick1601, `ResponseBodyEmitterReturnValueHandleris` delegates to `ReactiveTypeHandler` which handles adapts the future via a Reactor-based adapter registered through the `ReactiveAdapterRegistry`. So this comes from Project Rector and I see a fix there https://github.com/reactor/reactor-core/issues/1652 related to this for version 3.3. Spring Framework 5.2 is based on that version of Reactor but you can try independently upgrading Reactor first.

maverick1601:
Never mind, I figured, that it works as expected, just not via the `DeferredResultMethodReturnValueHandler`, as you've mentioned. The problem was on my side, testing with a mis-configured `ExceptionHandler`.

Issue #22480, opened 2019-02-27
Repeat calls to Databuffer#write(CharSequence, Charset) has unexpected contents in the DataBuffer


drewtul:
Expected: repeated calls to `DataBuffer#write(CharSequence, Charset)` should be appending to the buffer at the current write position, so the effect of multiple calls is to join these `CharSequence` together inside the buffer.

Actual: I seem to see varied outputs depending on the length of the `CharSequence`, but any chained `write()` calls result in unexpected contents in the buffer.

Also if I encode the strings myself and call `write(ByteBuffer)` multiple times this works as expected.

Test:
```Java
import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Test;
import org.springframework.core.io.buffer.DataBuffer;
import org.springframework.core.io.buffer.DataBufferFactory;
import org.springframework.core.io.buffer.DefaultDataBufferFactory;

import java.nio.CharBuffer;
import java.nio.charset.Charset;
import java.nio.charset.CharsetDecoder;
import java.nio.charset.CharsetEncoder;
import java.nio.charset.StandardCharsets;

public class TestDataBuffer {

    private DataBufferFactory dataBufferFactory = new DefaultDataBufferFactory();

    private Charset charset = StandardCharsets.UTF_8;

    private CharsetDecoder decoder;

    private CharsetEncoder encoder;

    @BeforeEach
    public void decoder() {
        decoder = charset.newDecoder();
        encoder = charset.newEncoder();
    }

    @Test
    public void testSingleWrite() throws Exception {

        String payload = "ababab";
        DataBuffer buffer = dataBufferFactory.allocateBuffer();
        buffer.write(payload, charset);

        String output = decoder.decode(buffer.asByteBuffer()).toString();
        Assertions.assertEquals(payload, output);
    }

    @Test
    public void testDoubleWrite() throws Exception {

        String payload = "ababab";
        String payloadTwo = "cdcdcdcd";
        DataBuffer buffer = dataBufferFactory.allocateBuffer();
        buffer.write(payload, charset)
              .write(payloadTwo, charset);

        String output = decoder.decode(buffer.asByteBuffer()).toString();
        Assertions.assertEquals(payload + payloadTwo, output);
    }
    @Test
    public void testTripleWrite() throws Exception {

        String payload = "ababab";
        String payloadTwo = "cdcdcdcd";
        String payloadThree = "efefefefef";
[...]

PR #22484, opened 2019-02-27
Fix repeat DataBuffer#write(CharSequence, Charset) calls

drewtul:
Fixes issue described in #22480


bclozel:
Thanks @drewtul !
I've applied this change to our 5.1.x branch so this will ship with Spring Framework 5.1.6.

drewtul:
No worries! Thanks for the prompt review, had me confused at my first use of `DataBuffer`, thought I was doing something wrong for a bit.

PR #22497, opened 2019-03-01
add TODOs to be done after migration to JDK 11

stsypanov:
In JDK 11 `String::repeat` becomes available and can be used to simplify code.


snicoll:
Thanks for the PR but I don't think it is very healthy for the project to keep a PR around for a long time. We're not going to switch to JDK 11 any time soon and the chances are quite high this PR will be outdated by the time we get to it.

Thanks anyway.

PR #22542, opened 2019-03-07
Polish: do simple clean-up

stsypanov:
- try-with-resources
- map.keySet().contains() -> map.containsKey()
- Character -> char
- unncessary cast

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
@@ -156,8 +156,7 @@ public abstract class YamlProcessor {
 			if (logger.isDebugEnabled()) {
 				logger.debug("Loading from YAML: " + resource);
 			}
-			Reader reader = new UnicodeReader(resource.getInputStream());
-			try {
+			try (Reader reader = new UnicodeReader(resource.getInputStream())) {
 				for (Object object : yaml.loadAll(reader)) {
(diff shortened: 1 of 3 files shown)


jhoeller:
@stsypanov this looks good in general, however, the cast to `Buffer` was intentional there: `ByteBuffer` has a covariant return type on JDK 9, so any code compiled against hard-codes the specific return type signature in its own bytecode. The only way around this - allowing the framework codebase to be compiled on JDK 8 or 9+ while also being able to run on JDK 8 or 9+ - is a cast to `Buffer` which has a consistent return type declaration across both JDK generations. We use such a cast in several places.

jhoeller:
If you could revert that change to the Buffer cast, ideally forcing it into a single commit for the PR, I'd be happy to merge this right away.

stsypanov:
@jhoeller should I then put `@SuppressWarning` over the method with cast to `ByteBuffer` to prevent other users from doing the same mistake?

jhoeller:
Is there a well-defined `@SuppressWarnings` key for 'unnecessary' casts? I'm not seeing any warnings in javac or IntelliJ that can be suppressed really... So maybe the best we can keep doing is a code comment, as present in the line above the cast already.

stsypanov:
Yes, we can use `@SuppressWarnings("RedundantCast")` - this prevents IDEA from reporting casts otherwise being subject of 'Redundant type cast` inspection.

I've put the cast back.

PR #22545, opened 2019-03-08
Migrate CacheControl to use java.time.Duration for durations

larsgrefer:


jhoeller:
Thanks for raising this! Other classes such as `ResponseCookie` and `TaskScheduler` are `Duration-based` already, so `CacheControl` was really an odd one out.

PR #22546, opened 2019-03-08
Enhance support for java.time types in spring-web

larsgrefer:
see also #22545 and #22103


jhoeller:
I went with a custom commit that implements these accessors a bit differently, adapting `Duration` values to the existing storage format rather than going via `Duration` even for long values. Thanks for the PR and for raising this gap to begin with!

Issue #22548, opened 2019-03-08
FormHttpMessageConverter should use ISO-8859-1 charset as default for application/x-www-form-urlencoded


candrews:
FormHttpMessageConverter currently uses utf-8 as the default encoding for reading and writing the `application/x-www-form-urlencoded` content type when no `charset` parameter is specified.

However, the default really should be ISO-8859-1 (aka ASCII aka Windows-1252) as that's what specified in the IANA registration: https://www.w3.org/TR/2016/WD-html51-20160310/iana.html#application-x-www-form-urlencoded

It's also what browsers and servers expect when no charset parameter is specified.

I'm reporting this issue as FormHttpMessageConverter always writes forms as `application/x-www-form-urlencoded;charset=utf-8` and the application receiving that data doesn't expect that - it requires `application/x-www-form-urlencoded` (notably without the charset), and it (correctly) expects that since there is no charset specified it will receive ISO-8859-1 encoded data.

Also, if the charset is ISO-8859-1 when writing `application/x-www-form-urlencoded`, then the content-type written should be `application/x-www-form-urlencoded` without the charset set (since the charset is the default, it should not be specified).

rstoyanchev:
The switch from "ISO-8859-1" to "UTF-8" as the default dates back to version 3.0.6, issue #12445. Switching to the previous default 8+ years later would be odd to say the least and more importantly it would be a step back to switch to ISO-8859-1 as a default for anything in 2019! 

The recommendation from `URLEncoder#encode` which we use for encoding:

> Note: The World Wide Web Consortium Recommendation states that UTF-8 should be used. Not doing so may introduce incompatibilities.

HTML 5.2 points to UTF-8 consistently as the default encoding in the [form submission](https://www.w3.org/TR/html52/sec-forms.html#selecting-a-form-submission-encoding) section. 

I don't see what evidence you have. The link you provided is from a 5.1 draft which doesn't contradict and simply says it's octets (i.e. encoded).

PR #22549, opened 2019-03-08
ISO-8859-1 as default for application/x-www-form-urlencoded

candrews:
The IANA registration at https://www.w3.org/TR/2016/WD-html51-20160310/iana.html#application-x-www-form-urlencoded specifies that a US-ASCII encoding should be used by default. In practice, it's also what browsers and servers expect when no charset parameter is specified.

When writing with content type application/x-www-form-urlencoded, if the charset is the default, leave the charset off of the content-type. Since it's the default, it need not be specified, and some (particularly) servers don't work correctly when a charset is specified on application/x-www-form-urlencoded. See https://bugzilla.mozilla.org/show_bug.cgi?id=241540 for background.

Fixes https://github.com/spring-projects/spring-framework/issues/22548

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/FormHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/FormHttpMessageConverter.java
@@ -98,9 +98,10 @@ public class FormHttpMessageConverter implements HttpMessageConverter<MultiValue
 	 * The default charset used by the converter.
 	 */
 	public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
+	public static final Charset DEFAULT_APPLICATION_FORM_URLENCODED_CHARSET = StandardCharsets.ISO_8859_1;
 
 	private static final MediaType DEFAULT_FORM_DATA_MEDIA_TYPE =
-			new MediaType(MediaType.APPLICATION_FORM_URLENCODED, DEFAULT_CHARSET);
+			new MediaType(MediaType.APPLICATION_FORM_URLENCODED, DEFAULT_APPLICATION_FORM_URLENCODED_CHARSET);
 
 
 	private List<MediaType> supportedMediaTypes = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


spring-web/src/main/java/org/springframework/http/converter/FormHttpMessageConverter.java
```diff
@@ -312,14 +312,24 @@ private void writeForm(MultiValueMap<String, Object> formData, @Nullable MediaTy
 		else {
 			StreamUtils.copy(bytes, outputMessage.getBody());
 		}
+		
+		if (MediaType.APPLICATION_FORM_URLENCODED.includes(contentType) && charset.equals(DEFAULT_APPLICATION_FORM_URLENCODED_CHARSET) ) {
+			contentType = (MediaType.APPLICATION_FORM_URLENCODED;
```
andrei-ivanov: does this line compile?

XYUU:
This change will result in many historical projects not working after the update, please be careful to merge.

candrews:
> This change will result in many historical projects not working after the update, please be careful to merge.

I'm not so sure that's true.

This form data is supposed to be url encoded to only US ASCII, and that's what browsers seem to do. If the form does something else, it should be setting the charset (ex, if it's sending url encoded data but not encoding utf-8 characters, it should set the charset to utf-8, and at least in the case of Spring, that is what it does).

This change shouldn't break anything, brings Spring in line with what is done in practice and what is specified by the rfc, and fixes currently broken use cases (such as where Spring is used to make form data for a server that doesn't accept the charset parameter, which according to the iana registration for this content type, not accepting a charset parameter is a valid behavior, which is something Mozilla discovered over a decade ago which can be seen from the bug report I linked to).

If you believe this would break something, can you please more information?

snicoll:
@candrews in the future there is no need to open an issue and PR now we've moved the issue tracker to Github, one is enough. Also, I'd expect such a change to contain unit tests to exercice the behaviour.

XYUU:
For programs that do not currently have a specified character set, UTF-8 will be used to decode and encode. If it is currently working, it will not work after this merge. The problem caused by this historical reason has brought other problems, so I think it should be carefully considered.

XYUU:
Moreover, the code of your change is not elegant, and it is not uniform with webflux. For example, [FormHttpMessageReader ](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageReader.java)and [FormHttpMessageWriter](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageWriter.java) still use UTF-8.

candrews:
> I'd expect such a change to contain unit tests to exercice the behaviour.

Point taken - I've updated this PR with tests proving the behavior.

> Moreover, the code of your change is not elegant, 

Feedback welcome! If you have something specific, please let me know and I'll do my best to address it.

> and it is not uniform with webflux. For example, [FormHttpMessageReader ](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageReader.java)and [FormHttpMessageWriter](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageWriter.java) still use UTF-8.

I'm not a webflux user at the moment so I didn't notice those classes. I've updated this PR with corresponding changes to those as well - thank you for pointing it out.

rstoyanchev:
See my response under #22548.

PR #22573, opened 2019-03-12
Add missing break statements to stop looping when result is known

stsypanov:
Assuming 
- raise-the-flag loop raises the flag 
- `if` statement has not side effects requiring iteration up to the end
We can add break right after the flag is raised.


sbrannen:

Thanks

PR #22574, opened 2019-03-12
Introduce @SingletonScope and @PrototypeScope composed annotations

bojanv55:
Since there are already: @ApplicationScope, @RequestScope and @SessionScope ([here](https://github.com/spring-projects/spring-framework/tree/master/spring-web/src/main/java/org/springframework/web/context/annotation)). In order to be consistent, I guess it would be nice to include this 2 core annotations?

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/PrototypeScope.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/PrototypeScope.java
@@ -1,0 +1,34 @@
+package org.springframework.context.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
+
+/**
+ * {@code @PrototypeScope} is a specialization of {@link Scope @Scope} for a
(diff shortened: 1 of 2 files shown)


sbrannen:
I initially proposed those two composed annotations in #18022, but the Spring team decided against inclusion of them at the time.

Labeling with "for: team-attention" to discuss if the team would like to consider introducing such scope annotations now.

sbrannen:
After some internal discussions, we have decided that we do not wish to introduce composed annotations for singleton and prototype scopes, since we feel that they do not add much value over  `@Scope("singleton")` and `@Scope("prototype")` declarations, especially since singleton is the default.

You are of course free to implement such composed annotations and use them in your own projects if you feel that they add value for your team.

Thanks anyway for the PR!

bojanv55:
> After some internal discussions, we have decided that we do not wish to introduce composed annotations for singleton and prototype scopes, since we feel that they do not add much value over `@Scope("singleton")` and `@Scope("prototype")` declarations, especially since singleton is the default.
> 
> You are of course free to implement such composed annotations and use them in your own projects if you feel that they add value for your team.
> 
> Thanks anyway for the PR!

For singleton, you can say that it is default so there - it makes sense to leave it out (the only reason for inclusion would be consistency). But for prototype, I guess you would get compile time safety since otherwise you can always write something like `@Scope("prototipe")` and have program start without any problem until actual injection..

PR #22593, opened 2019-03-14
Simplify String concatenation

stsypanov:
In some cases we can skip creation of `StringBuilder`  or use `invokedynamic`-based concatenation.


sbrannen:
@jhoeller, I've assigned this to you for consideration.

izeye:
> use `invokedynamic`-based concatenation

@stsypanov This doesn’t seem to be true unless the Spring Framework itself is compiled against Java 9+. Am I missing something?

stsypanov:
Correct, as mentioned in https://openjdk.java.net/jeps/280 the byte code with indified concatenation is produced by javac of JDK 9. However as of current version (8) javac will produce `new StringBuilder().append()` implicitly, so explicit declaration of `StringBuilder` is not necessary.

Issue #22461, opened 2019-02-24
Clarify configuration lite support and offer an opt-in option


snicoll:
The Spring Boot team has been investigating startup time improvements by not relying on the proxy capabilities of configuration classes. Turns out that not using that stuff (basically switching all auto-configurations and related imports to "lite mode") makes quite a difference https://github.com/spring-projects/spring-boot/issues/9068#issuecomment-461520814

I am not a big fan that we were looking for `@Bean` methods in any component, treating them in lite mode if they don't have the `@Configuration` stereotype. I was wondering if we would be open to revisit this in the 5.2 line.

Being able to opt-out from the proxy mode using the `@Configuration` stereotype would be ideal. We are not keen to remove `@Configuration` and use the lite mode as it stands as:

* That's not representative of what a good citizen of the framework should do (using the right stereotype)
* We have an annotation processor that identifies configuration classes to extract conditions so that we can back off (without loading the ASM metadata or the class) when a static condition does not match (i.e. class not on the classpath)

Having a way to opt-out would be also a nice opportunity to clarify the behaviour and stop scanning components for bean factory methods.

wilkinsona:
I'd like to echo Stephane's dislike of using `@Bean` on any component. I find the behaviour surprising and it's caused confusion in the past. I too would prefer that you have to opt in to `@Bean` method support via a stereotype annotation that explicitly indicates that a class is a configuration class. Once you've opted in to being a configuration class and `@Bean` methods defining beans, we'd then need a way to opt out of proxying. Given that the use of proxies is more than an implementation detail and is already something that a user needs to think about (`@Configuration` classes cannot be `final`, for example), I wonder if a `proxy` attribute on `@Configuration` that defaults to `true` would suffice here.

In terms of the benefits of not proxying `@Configuration` classes, the table in the linked Spring Boot issue shows some of the performance benefits in terms of startup time. There are also other benefits with both metaspace and code cache usage being reduced due to fewer classes being loaded and subsequently processed by the JIT.

@philwebb raised another benefit of being able to identify a configuration class where `@Bean` methods should be supported but where proxying should not be used. It would allow an application to be started in a mode where the configuration classes were proxied anyway, but purely for the purpose of catching and warning about accidental interdependencies between `@Bean` methods. I modified Framework to use this technique to identify `@Configuration` classes in Spring Boot that currently rely on proxying and it proved very useful.

sbrannen:
Historically related issues:

- gh-13690
[...]

PR #22596, opened 2019-03-14
Avoid calling other bean methods in web config

bclozel:
This commit changes the main configuration classes for Spring MVC and
Spring WebFlux to not call other bean methods when setting up the web
infrastructure. This allows the `DelegatingWebFluxConfiguration` and
`DelegatingWebMvcConfiguration` to opt-in the lite-mode, as introduced
in gh-22461.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/setup/StandaloneMockMvcBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/setup/StandaloneMockMvcBuilder.java
@@ -62,6 +62,7 @@ import org.springframework.web.servlet.handler.MappedInterceptor;
 import org.springframework.web.servlet.i18n.AcceptHeaderLocaleResolver;
 import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter;
 import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerMapping;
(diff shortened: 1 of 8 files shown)


bclozel:
In this PR, I'm trying to check what it would take to leverage lite-mode in the web configurations.
This changes a few public methods, but arguably those are `@Bean` methods and not meant to be called directly. This shows the exact dependencies required by each bean, which was not necessarily obvious in the previous version.

@jhoeller , @rstoyanchev I'd like to get your opinion on that.
Thanks!

rstoyanchev (commented):
The configuration support classes are documented as the "advanced option" for configuring Spring MVC or Spring WebFlux, so this will disrupt some applications using or overriding the protected methods. Fortunately those are few. 

What I'm more concerned is that `proxyBeanMethods` is `false` by default. Applications extending these classes for advanced config would need to be careful not to call bean methods directly which is easy to overlook, and would be surprising. Classes may even run into that by simply upgrading to 5.2 if they already do that. It could also be an optimization that's not actually needed.

Would it make more sense to go with proxied bean methods by default and only set it to false when you're ready and willing to opt in. Boot can certainly turn off the proxying since it is in control of this configuration and guarantee it's done right.

bclozel:
@rstoyanchev you're right - I've changed my PR accordingly. We can still add that `proxyBeanMethods ` attribute in Spring Boot.

sdeleuze:
:+1: for that change!

PR #22643, opened 2019-03-22
Introduce copyProperties() in BeanUtils for copying specific properties

matoujun:
When a bean has a lot of properties, copyProperties() by ignore ones is not convenient if we want to copy a few of fields.


yurizp:
I think it's a good idea to create tests to validate this functionality.

snicoll:
Thanks for the PR but `BeanUtils` is mostly for our own use within the framework and we prefer not to extend its surface API.

Issue #22590, opened 2019-03-13
Reactive transaction management support


jhoeller:
Since we have two cases in Spring Data already where transactions are supported with reactive drivers - namely for MongoDB and R2DBC -, let's revisit our core transaction management abstraction and our `@Transactional` support for reactive scenarios.

The transaction interceptor could automatically detect reactive scenarios through a `Publisher` return type from the method it is supposed to proxy and automatically go through a reactive transaction management SPI in that case. Such an SPI has been prototyped for MongoDB and R2DBC already, so it's mostly a matter of rolling it into Spring Framework proper.

Configuration-wise, the existing transaction configuration elements would largely still apply, just pointing to a reactive transaction manager implementation instead of a classic `PlatformTransactionManager`.

jhoeller:
I've baked this into `TransactionAspectSupport` directly, as inner classes that are conditionally used when a `Publisher`-based method is discovered. A transaction manager can now be configured as a `TransactionManager` marker interface, automatically cast to `PlatformTransactionManager` or `ReactiveTransactionManager` as needed.

bdeneuter:
We are trying to upgrade to spring-boot 2.2/Spring 5.2. We are using the reactive WebClient in combination with plain JDBC (we are using Aurora MySQL). This means we have reactor on the classpath. After upgrading we receive the following error: java.lang.IllegalStateException: Specified transaction manager is not a ReactiveTransactionManager. What is the path we need to follow to upgrade in this case? Or should I create a bug ticket?

sbrannen:
@bdeneuter, your question sounds like it may have already been covered by #23832.

Can you please try out Spring Framework 5.2.1?

Also, please refrain from asking questions on closed issues. If you encounter a regression, please open a new issue.

PR #22646, opened 2019-03-23
Add reactive transaction support SPI

mp911de:
This commit adds SPI interfaces to support reactive transactions through spring-tx with optional dependencies to Project Reactor and supportive implementations for TransactionalOperator and AbstractReactiveTransactionManager.

See #22590 for more background. The implementation is mainly inspired by `spring-tx` leveraging Reactor's Subscriber Context.

### DIFF
--- a/spring-tx/spring-tx.gradle
+++ b/spring-tx/spring-tx.gradle
@@ -1,5 +1,11 @@
 description = "Spring Transaction"
 
+dependencyManagement {
+	imports {
+		mavenBom "io.projectreactor:reactor-bom:${reactorVersion}"
+	}
+}
+
 dependencies {
 	compile(project(":spring-beans"))
 	compile(project(":spring-core"))
@@ -10,8 +16,10 @@ dependencies {
 	optional("javax.resource:javax.resource-api:1.7.1")
 	optional("javax.transaction:javax.transaction-api:1.3")
 	optional("com.ibm.websphere:uow:6.0.2.17")
+	optional("io.projectreactor:reactor-core")
 	optional("io.vavr:vavr:0.10.0")
 	testCompile("org.aspectj:aspectjweaver:${aspectjVersion}")
 	testCompile("org.codehaus.groovy:groovy:${groovyVersion}")
 	testCompile("org.eclipse.persistence:javax.persistence:2.2.0")
+	testCompile("io.projectreactor:reactor-test")
 }
--- a/spring-tx/src/main/java/org/springframework/transaction/ReactiveTransactionManager.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/ReactiveTransactionManager.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.transaction;
+
+import reactor.core.publisher.Mono;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * This is the central interface in Spring's reactive transaction infrastructure.
+ * Applications can use this directly, but it is not primarily meant as API:
+ * Typically, applications will work with either transactional operators or
+ * declarative transaction demarcation through AOP.
+ *
+ * @author Mark Paluch
+ * @since 5.2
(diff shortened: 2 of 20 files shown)


jhoeller:
@mp911de Thanks for turning it into a PR, looks like a fine start for integration with `@Transactional`!

Just one immediate thing: Could you please move the `transaction.reactive.support` contents to `transaction.reactive` itself since the latter seems to be empty at this point? The `reactive` subpackage would then effectively be the support-like package for reactive transactions anyway... Would be great to sort this out within the PR still.

mp911de:
Resolved `reactive.support` into `reactive` now.

jhoeller:
Running out of time for 5.2 M1, rather expectedly... Let's make this a 5.2 M2 topic, first thing once M1 is out. I hope that's also early enough for Spring Data's purposes.

mp911de:
That should work for us.

jhoeller:
This is in master now, including a first pass with a revision of mine that removes all legacy configuration options from `AbstractReactiveTransactionManager` plus a few further things. The intent is to keep the reactive transaction abstraction rather streamlined, not burdening it with JTA/JPA-oriented options that `AbstractPlatformTransactionManager` historically had to deal with. Let me know if anything important is missing now for the Mongo/R2DBC implementations; we can of course consider reintroducing relevant options through template methods or the like (ideally not through boolean setters again).

A second pass is coming in just a bit, renaming some of the types in `transaction.reactive` through dropping the "Reactive" prefix: `TransactionSynchronizationManager`, `TransactionCallback` etc read much nicer in the corresponding API declarations and are locally prefixed by "reactive" through their containing package anyway. Also, they don't immediately conflict with the same-named types in `transaction.support`, whereas we have to use specific prefixes for new types in the `transaction` base package (e.g. `ReactiveTransactionStatus` vs `TransactionStatus`). We applied the same reasoning to our reactive HTTP abstractions, using same-named types in quite a few cases there as well.

@mp911de Please wait for that renaming commit before integrating the Spring Data bits with it.

jhoeller:
I also consider renaming `ReactiveTransactionStatus` to simply `ReactiveTransaction`, making it extend a common `TransactionExecution` interface that has the common methods with `TransactionStatus`... and shortening its implementation name to `GenericReactiveTransaction`, making the template methods in `AbstractReactiveTransactionManager` nicer.

jhoeller:
@mp911de Are you ok with the shortened type names suggested above? It's ready for committing from my side.

jhoeller:
Being at it, I also intend to drop `ReactiveResourceHolderSynchronization`, making the `transaction.reactive` package completely decoupled from `transaction.support`. I wouldn't recommend building any reactive resource support with the `ResourceHolder` type which is really quite oriented towards thread binding (`isVoid` etc don't make much sense otherwise); if needed, we can introduce a dedicated reactive resource holder variant.

mp911de:
Thanks a lot for merging this one. I'm fine with shorter names. I'm always torn having multiple types with the same name.

Right now, when using multiple transactional resources, it's sufficient to keep a single transaction manager and additional resources can join by registering `ReactiveResourceHolderSynchronization`.

By removing `ReactiveResourceHolderSynchronization`, we no longer have this possibility and only a single transactional resource can be used in a reactive transaction.

jhoeller:
So who would use `ReactiveResourceHolderSynchronization` - is it primarily integration code of ours (which could include a custom variant of that class for its purposes) or general user code? If we keep shipping such a class out of the box, we should at least decouple it from `transaction.support.ResourceHolder`...

jhoeller:
Hmm it seems that `ReactiveResourceHolderSynchronization` is just generically typed to the `ResourceHolder` class anyway... We could simply remove that generic bound and decouple it from `ResourceHolder` that way.

mp911de:
It would make sense to keep `static TransactionalOperator create(ReactiveTransactionManager transactionManager)` as an additional factory method, without requiring `DefaultTransactionDefinition`.

jhoeller:
The main reason for dropping it was the use of `DefaultTransactionDefinition` which lives in `transaction.support` and introduces an unwanted dependency that the `transaction.reactive` package overall doesn't have anymore. If we find this important enough to have, we could reintroduce it and use some local dummy `TransactionDefinition` instead; after all, we just need an instance which returns default values, not a modifiable instance like `DefaultTransactionDefinition`. Thinking about it, we could even have a static defaults accessor on the `TransactionDefinition` interface itself, returning an unmodifiable instance... that would be convenient for users as well, not having to use `DefaultTransactionDefinition` and do several configuration steps with it.

jhoeller:
That seems to work nicely with default methods on the `TransactionDefinition` interface itself. I'll reintroduce `static TransactionalOperator create(ReactiveTransactionManager transactionManager)` with an implementation along those lines.

mp911de:
I found two bugs during the integration in R2DBC and MongoDB (`doCleanupAfterCompletion` does not get subscribed to and supposedly `TransactionalOperator` sends a cancellation signal after a few emitted elements). I'll provide either today or Monday a PR to fix these.

jhoeller:
Alright, good to catch those bugs early enough still...

The `TransactionalOperator` revision with the overloaded `create` method is in master now. I really like the new `TransactionDefinition.withDefaults()` method, reads nicely in several of our own code spots...

PR #22656, opened 2019-03-25
Handle Optional<?> @RequestParam() when generating URL using MvcUriComponentsBuilder

fprochazka:
Example of an action:

```java
@GetMapping("/{userId}")
public ModelAndView productGroups(
    @PathVariable @AssertUuid final String userId,
    @RequestParam("parentId") final Optional<@AssertUuid String> rawParentId,
    @PageableDefault(size = 100) final Pageable pageable
) throws ControllerException
```

this action validates, that the `parentId` is UUID, if it's present (which works great), but it also rejects the request completely, if it's not (with `400 BAD REQUEST`). Which means any of the following ends with `400` - `?parentId`, `?parentId=`, `?parentId=garbage`

-----

- Generating URL using `MvcUriComponentsBuilder.fromMethodCall` doesn't work without the `OptionalUnwrappingConverter` I've provided
- Without the fix in `RequestParamMethodArgumentResolver`, it generates empty query args - resulting in url with `?parentId` (without `=` and without value) - which is not valid and ends with `400`

I'm not sure if this is the correct way to fix this, so if you'll please give me some feedback on this, I'd be happy to add some tests.

-----

Related issue #16785  - it states that spring handles `Optional<?>` in handler arguments

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/DefaultConversionService.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/DefaultConversionService.java
@@ -97,6 +97,7 @@ public class DefaultConversionService extends GenericConversionService {
 		converterRegistry.addConverter(new IdToEntityConverter((ConversionService) converterRegistry));
 		converterRegistry.addConverter(new FallbackObjectToStringConverter());
 		converterRegistry.addConverter(new ObjectToOptionalConverter((ConversionService) converterRegistry));
+		converterRegistry.addConverter(new OptionalUnwrappingConverter((ConversionService) converterRegistry));
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


rstoyanchev:
@fprochazka, couldn't this all be handled in `RequestParamMethodArgumentResolver`, i.e. unpacking the value from the optional and then formatting it as usual? That would be a simpler solution, minus the converter it seems.

fprochazka:
I'm not sure, you tell me :) I know Spring well enough to bend it, but not yet to solve this kind of problems systematically.

FYI, I've just copied already existing convertor and just reversed it with few tweaks. Imho the convertor should be available anyway.

rstoyanchev:
Could you revise the PR to change only the argument resolver please? Also a test for it.

Issue #22720, opened 2019-04-01
o.s.h.server.reactive.ChannelSendOperator may leak the memory on cancel/error


violetagg:
**Affects:** 5.1.6.RELEASE

---
`org.springframework.http.server.reactive.ChannelSendOperator` does prefetch

https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/server/reactive/ChannelSendOperator.java#L132

When cancel/error signal is received by `org.springframework.http.server.reactive.ChannelSendOperator.WriteCompletionBarrier` the item that was prefetched should be freed.

This is related to https://github.com/spring-cloud/spring-cloud-gateway/issues/810

violetagg:
@rstoyanchev Additional fix is needed

In Reactor Netty we first send the headers and then start to consume the Publisher.
It might happen that while sending the headers we receive an exception which we will propagate (for example `javax.net.ssl.SSLException: SSLEngine closed already`). At that point Spring Framework should release the cached item.

PR #22728, opened 2019-04-02
Release cached item in ChannelSendOperator also in case of error from writeFunction

violetagg:
A follow-up fix for gh-22720.


rstoyanchev:
The use case is an error from the writeFunction, e.g. if Reactor Netty fails to flush the response status and the headers before consuming the cached item.

rstoyanchev:
This is merged in https://github.com/spring-projects/spring-framework/commit/9bd0ec33f808cd2aa55048e535a9e8775f524dc6.

PR #22752, opened 2019-04-05
Add paging and sorting mechanism with the JdbcTemplate

davidmalis:
Based on the Spring Data Pageable API, this extension adds a
queryForPage(...) method to the JdbcTemplate which through SQL
subselect wrapping techniques enables the user to query for paged
(and sorted) results.

 Changes to be committed:
	modified:   spring-jdbc/spring-jdbc.gradle
	new file:   spring-jdbc/src/main/java/org/springframework/jdbc/core/PagingAndSortingJdbcTemplate.java

### DIFF
--- a/spring-jdbc/spring-jdbc.gradle
+++ b/spring-jdbc/spring-jdbc.gradle
@@ -5,6 +5,7 @@ dependencies {
 	compile(project(":spring-core"))
 	compile(project(":spring-tx"))
 	optional(project(":spring-context"))  // for JndiDataSourceLookup
+	optional("org.springframework.data:spring-data-commons:2.1.5.RELEASE")
 	optional("javax.transaction:javax.transaction-api:1.3")
 	optional("org.hsqldb:hsqldb:${hsqldbVersion}")
 	optional("com.h2database:h2:1.4.199")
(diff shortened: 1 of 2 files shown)


spring-jdbc/spring-jdbc.gradle
```diff
@@ -5,6 +5,7 @@ dependencies {
 	compile(project(":spring-core"))
 	compile(project(":spring-tx"))
 	optional(project(":spring-context"))  // for JndiDataSourceLookup
+	optional("org.springframework.data:spring-data-commons:2.1.5.RELEASE")
```
mp911de: This isn't going to work as we end up with a dependency cycle. Spring Data depends on Spring Framework, so Spring Framework cannot depend on Spring Data.

davidmalis: Hi, @mp911de, thanks for the review. I wasn't aware of the dependency problem, as Data-Commons isn't dependent on spring-jdbc directly. 

Could you please review the main class of this PR, and tell if it is even worth considering? Would this work if we transfer it to one of the Spring Data modules?

mp911de:
Spring Data JDBC is dialect-aware with an extensible dialect mechanism. Data JDBC builds on top of JdbcTemplate (NamedParameterJdbcTemplate in particular). It typically uses repositories and not so much SQL augmentation. 

I’d suggest filing a ticket at https://jira.spring.io/projects/DATAJDBC and continue the discussion over there.

snicoll:
It indeed creates a cycle so please follow-up with Spring Data JDBC if you haven't done so already.

PR #22758, opened 2019-04-07
Add Embedded Postgresql

juliano:
I have been using [Postgres Embedded](https://github.com/yandex-qatools/postgresql-embedded) for testing, this PR makes it available via Spring APIs for the same purpose.


snicoll:
Thanks for the PR but using flapdoodle isn't the same than the database we support at the moment (which are "natively embedded").

PR #22769, opened 2019-04-08
@CachePut evaluates key before cache condition

### LINKED ISSUE #22294: @CachePut evaluates key before cache condition
**Affects:** Spring Context 4.3.4 and higher

---

Trying to use `@CachePut` on a method that might return `null`.

```java
@CachePut(cacheNames = Constants.Caches.USER, key = "#result.id", unless = "#result == null")
public MyObject doSomething();
```

You would expect the key expression not to be evaluated if the `unless` condition is `true` but this is not happening. The key is being evaluated before the `unless` condition is evaluated which leads to a 

> org.springframework.expression.spel.SpelEvaluationException: EL1007E:(pos 0): Property or field 'id' cannot be found on null

```java
                if (cacheHit != null && cachePutRequests.isEmpty() && !hasCachePut(contexts)) {
			// If there are no put requests, just use the cache hit
			cacheValue = cacheHit.get();
			returnValue = wrapCacheValue(method, cacheValue);
		}
		else {
			// Invoke the method if we don't have a cache hit
			returnValue = invokeOperation(invoker);
			cacheValue = unwrapReturnValue(returnValue);
		}

		// Collect any explicit @CachePuts
		collectPutRequests(contexts.get(CachePutOperation.class), cacheValue, cachePutRequests);

		// Process any collected put requests, either from @CachePut or a @Cacheable miss
		for (CachePutRequest cachePutRequest : cachePutRequests) {
			cachePutRequest.apply(cacheValue);
		}
```

The `collectPutRequests` tries to evaluate the key if the `condition` passes (there is no condition). I think the keys need to be evaluated only if `unless` is `false`?

lgxbslgx:
At present, `@Cacheable`, `@CacheEvict` and `@CachePut` evaluate their keys after `condition` but before  `unless`. I think the keys should be evaluated after all the cache condition which contains the `condition` and `unless`.

I change the code, add the corresponding test cases and fix the JavaDoc for @Cacheable, @CacheEvict and @CachePut.

Fixes #22294.


capatcapworld:
When will this fix be merged in to master ? We have this problem in our code.

lgxbslgx:
I merged the main branch and solved the conflicts just now.

snicoll:
It took me a bit to understand why the evaluation of cacheable was moved as it felt unrelated but evaluating the unless expression earlier means that we need the result for cacheable as well.

snicoll:
@lgxbslgx thank you for making your first contribution to Spring Framework.

PR #22781, opened 2019-04-10
Polish BeanDefinitionVisitor

rupertw:
Remove unnecessary 'String.equals()' call.
Remove unnecessary local variables.


jhoeller:
I'm afraid the current code is that way by design: We are trying to avoid re-populating the properties on the `BeanDefinition` since those setter methods are defined on an interface, with the actual implementation allowing for custom behavior. Even if our standard `BeanDefinition` implementations mostly just store the given values as-is, a setter call could have side effects such as indicating a user-set value over a default value. Even if the particular properties here do not seem to be affected at this point, we do such stuff for other settings and might do it to our own `BeanDefinition` impls in the future, even aside from user-provided `BeanDefinition` impls/subclasses doing it at this point already.

That's also the reason why we're not defensively coding the latter case: modifying a `TypedStringValue` instance which is an implementation class that we control, by design not meant to have side effects.

Thanks for the PR, in any case! Such code spots are always worth another look.

PR #22784, opened 2019-04-10
Get all the aliases at getPropertyNames

### LINKED ISSUE #22168: JOptCommandLinePropertySource selects property names in an inconsistent way [SPR-17639]
**[chludwig-haufe](https://jira.spring.io/secure/ViewProfile.jspa?name=chludwig-haufe)** opened **[SPR-17639](https://jira.spring.io/browse/SPR-17639?redirect=false)** and commented

I recently had to build a Spring Boot command line application whose application context depends on command line options. I specified the qualified property names (i.e., `"<prefix>.<prop-name>"` as option alias names using the `OptionParser#addAll(List<String>)` method of JOpt's fluent interface. I injected a customized `StandardEnvironment` with a `JOptCommandLinePropertySource` based on the parsed command line arguments into the `SpringApplication`. To my surprise, however, some of the properties were never bound to the corresponding `@ConfigurationProperties` object.

I found that this was caused by the implementation of `JOptCommandLinePropertySource#getPropertyNames()`: A comment in the source code claims _"only the longest name is used for enumerating"._ However, the implementation always returns the _last_ element of `OptionSpec#options()`. This is, in general, _not_ the longest option name, because `OptionSpec#options()`
* first clusters the option names into short and long options and
* then sorts these groups lexicographically, respectively.

For some options, `JOptCommandLinePropertySource#getPropertyNames()` therefore contained the qualified alias name required for binding the property to the `@ConfigurationProperties` object, while for other options it contained the unqualified option alias. Admittedly, the documentation of `JOptCommandLinePropertySource` never specifies which property names are enumerated; but I can't see a scenario where the current behavior is useful.

This ticket's reference URL points to a JUnit class that demonstrates this behavior. The test class also includes alternative property name selection strategies that would work better in my scenario. (I wonder why `JOptCommandLinePropertySource#getPropertyNames()` needs to filter the option alias names in the first place.)

---

**Affects:** 5.0.11, 5.1.3, 5.1.4

**Reference URL:** https://github.com/chludwig-haufe/JOptCommandLinePropertySource-Issues/blob/master/src/test/java/com/haufe/bugreports/spring/wrongpropertynamedemo/JOptCommandLinePropertySourceTests.java

lgxbslgx:
At present, the `JOptCommandLinePropertySource.getPropertyNames` get the last element in the result of `spec.options()`. But the comment says `Only the longest name is used for enumerating`. It is a contradiction and one of the two(code or comment) has to be changed.

I change the code according the comment because I think the longest name is more useful. And if more than one name have the same longest length, the first one(alphabetically) will be used in my solution.

Fix #22168.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/env/JOptCommandLinePropertySource.java
+++ b/spring-core/src/main/java/org/springframework/core/env/JOptCommandLinePropertySource.java
@@ -88,10 +88,19 @@ public class JOptCommandLinePropertySource extends CommandLinePropertySource<Opt
 	public String[] getPropertyNames() {
 		List<String> names = new ArrayList<>();
 		for (OptionSpec<?> spec : this.source.specs()) {
-			String lastOption = CollectionUtils.lastElement(spec.options());
-			if (lastOption != null) {
-				// Only the longest name is used for enumerating
-				names.add(lastOption);
+			List<String> aliases = new ArrayList<>(spec.options());
+			String longestAliase = "";
+			for (String aliase : aliases) {
+				if (aliase.length() > longestAliase.length()) {
+					// Only the longest name is used for enumerating.
+					// If more than one name have the same longest length,
+					// the first one(alphabetically) will be used.
+					longestAliase = aliase;
+				}
+			}
+			if (!"".equals(longestAliase) && !names.contains(longestAliase)) {
+				// The names are not added repeatedly.
+				names.add(longestAliase);
 			}
 		}
 		return StringUtils.toStringArray(names);
--- a/spring-core/src/test/java/org/springframework/core/env/JOptCommandLinePropertySourceTests.java
+++ b/spring-core/src/test/java/org/springframework/core/env/JOptCommandLinePropertySourceTests.java
@@ -88,6 +88,7 @@ public class JOptCommandLinePropertySourceTests {
 		CommandLinePropertySource<?> ps = new JOptCommandLinePropertySource(options);
 		assertEquals(Arrays.asList("bar","baz","biz"), ps.getOptionValues("foo"));
 		assertThat(ps.getProperty("foo"), equalTo("bar,baz,biz"));
+		assertEquals(1, ps.getPropertyNames().length);
 	}
 
 	@Test
@@ -158,6 +159,19 @@ public class JOptCommandLinePropertySourceTests {
 		assertThat(ps.containsProperty("o2"), is(true));
 		String nonOptionArgs = ps.getProperty("NOA");
 		assertThat(nonOptionArgs, equalTo("noa1,noa2"));
+	}
+
+	@Test
+	public void withMultipleOptionAliases() {
+		OptionParser parser = new OptionParser();
+		parser.acceptsAll(Arrays.asList("u", "username", "user"));
+		parser.acceptsAll(Arrays.asList( "MN", "NM"));
+		parser.acceptsAll(Arrays.asList("A", "CBA", "B", "ABC", "BCA", "AB", "BA"));
+
+		OptionSet optionSet = parser.parse("-u", "-ABC", "-MN");
+		EnumerablePropertySource<?> ps = new JOptCommandLinePropertySource(optionSet);
+
+		assertArrayEquals(new String[]{"username", "ABC", "MN"},  ps.getPropertyNames());
 	}
 
 	@Test


jhoeller:
I'm not quite sure why the comment says "longest" there to begin with... I'm not sure that's a good rule to follow, it should rather be the last (or first - if we were to choose this freshly) entry for an intuitive convention. From that perspective, I'd rather fix the comment.

lgxbslgx:
As @chludwig-haufe commented on the issue #22168, at present the code of `JOptCommandLinePropertySource.getPropertyNames`  may cause some of the properties were never bound to the corresponding @ConfigurationProperties object.

You can test it using the following code.

```Java
@ConfigurationProperties(prefix = "guess")
@Component
public class Guess {
	private String name;

	public String getName() {
		return name;
	}

	public void setName(String name) {
		this.name = name;
	}
}
```

```Java
@EnableConfigurationProperties(Guess.class)
@SpringBootApplication
public class IssueTest implements ApplicationRunner {

	@Autowired
	private Environment springEnvironment;

	@Autowired
	private Guess guess;

		OptionParser optionParser = new OptionParser(false);
		optionParser.acceptsAll(Arrays.asList("name", "guess.name")).withRequiredArg();
		OptionSet optionSet = optionParser.parse("-name", "apple");
		JOptCommandLinePropertySource configurationProperties = new JOptCommandLinePropertySource(optionSet);

		StandardEnvironment standardEnvironment = new StandardEnvironment();
		standardEnvironment.getPropertySources().addFirst(configurationProperties);
		SpringApplication springApplication = new SpringApplication(IssueTest.class);
		springApplication.setEnvironment(standardEnvironment);
		springApplication.run();
	}

	@Override
	public void run(ApplicationArguments args) throws Exception {
		System.out.println("guess.name = " + guess.getName());
	}

}
```

When I execute the main method using the `spring-boot 2.1.4.RELEASE` , the `name` in object `guess` is `null` . It is a bug caused by both `spring-boot` and `spring-framework`.

1. In `spring-framework`, the `JOptCommandLinePropertySource.getPropertyNames` only returns a name of a property but not all the aliases of the property.

2. In `spring-boot`, the `Binder.bindObject` uses `findProperty` to find a property and use `containsNoDescendantOf` to determine whether  the property name has a descendant. If property has a descendant, it would continue to find the descendant property to bind the `target`.

For the code I show above , firstly the `findProperty` in `Binder.bindObject` will return `null` using the parameter `ConfigurationPropertyName` which includes `guess` . Then the `containsNoDescendantOf(context.getSources(), name)` will return `true`. Because `JOptCommandLinePropertySource.getPropertyNames` return array `["name"]` (don't contain `guess.name`) and the `name` is not a descendant of `guess`. As a result, the expression `property == null && containsNoDescendantOf(context.getSources(), name)`  is `true` and `Binder.bindObject`  will return `null` so that the binding fail.

To fix the bug, we can change the code in either `spring-framework` or `spring-boot`.

1. In `spring-framework`, if `JOptCommandLinePropertySource.getPropertyNames` return array `[ "guess.name"]` or  `["name", "guess.name"]`, the `containsNoDescendantOf(context.getSources(), name)` will return false. Because the `guess.name` is a descendant of `guess`. As a result, the expression `property == null && containsNoDescendantOf(context.getSources(), name)`  is `false` and `Binder.bindObject`  will continue so that it can bind data successfully.

2. In `spring-boot`, @snicoll added a commit [430571](https://github.com/spring-projects/spring-boot/commit/430571b37b808c42b950c14778695a541adc45e9) which can fix the bug. It changes the expression  `property == null && containsNoDescendantOf(context.getSources(), name)` to  `property == null && containsNoDescendantOf(context.getSources(), name) && context.depth != 0`. So if I use `spring-boot 2.2.0.BUILD-SNAPSHOT`  instead of `spring-boot 2.1.4.RELEASE`, the bug doesn't occur.

We could simply consider that `spring-boot` uses `JOptCommandLinePropertySource.getPropertyNames` by mistake. But the name called `getPropertyNames` means the users, such as `spring-boot` , can use it to get property names without needing to know how it implements. So I want to change code at method `getPropertyNames` to make it more robust to use.

It may be a problem that we didn't discuss before. What does the `getPropertyNames`  return if the property has aliases? We should make a convention about this situation after discussing.

In my opinion, I don't think it is good to get the first, last, or longest name because it even can't solve the problem above. I would like to return all the aliases .

@jhoeller what is your opinion? Maybe you have a better solution.

PR #22785, opened 2019-04-10
Add ServerRequestTransformer before routing

### LINKED ISSUE #22408: Add filter before RouterHandler in WebFlux
I have my app deployed behind load balancer and I take advantage of `ForwardedHeaderTransformer` to be able to have proper redirect urls (mostly for Spring Security OAuth) 

Then I have my router functions which need to have conext path prefix hardcoded (since `ForwardedHeaderTransformer` changed conextPath and uri). 

I would like not being forced to hard code context path into my router functions. Instead I would like to have such context path stripped out by filter before going to `RouterHandler`. Therefore I suggest adding a filter mechanism that would be used just before `RouterHandler`. I cannot use general purpose `WebFilter` because it will work also for requests that will be base for redirections.

kkocel:
Fixes #22408

### DIFF
--- a/spring-core/src/main/kotlin/org/springframework/core/env/PropertyResolverExtensions.kt
+++ b/spring-core/src/main/kotlin/org/springframework/core/env/PropertyResolverExtensions.kt
@@ -34,13 +34,23 @@ operator fun PropertyResolver.get(key: String) : String? = getProperty(key)
 
 /**
  * Extension for [PropertyResolver.getProperty] providing a `getProperty<Foo>(...)`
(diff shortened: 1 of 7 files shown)


kkocel:
@poutsma IHMO router fn code is mostly authored by you. Would you mind taking a look? :)

rstoyanchev:
@kkocel thanks for the PR, and apologies for not responding on the original issue, but the description there does not make the motivation clear. I am going to close this PR for now, and will respond on the original issue. The PR will still be here if we need it.

PR #22792, opened 2019-04-14
Some trivial optimization in method populateBean()

hecenjie:
I don't think it's necessary to call the same get method so many times, maybe can do some trivial optimization here.


pivotal-issuemaster:
@cenjieHo Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22792)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22792) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@cenjieHo Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22792)!

jhoeller:
Thanks for raising this! Even if such field access through method calls is usually very efficient at runtime, this is worth changing for clarification purposes in any case: namely evaluating the same state of the field, never getting a different autowire mode returned from that getter. A one-time method call with a locally cached result makes this clear.

PR #22801, opened 2019-04-15
fix #22800 Optimize mono bodies

### LINKED ISSUE #22800: Optimize writing with Mono Publisher
`AbstractServerResponse` is currently masking the response body type information with the custom write barrier from `ChannelSendOperator`. In fact if the body is a Mono, we shouldn't require the barrier at all.

If we preserve the Mono information, the reactor netty response adapter can then preserve it too and call the underlying reactor-netty `send(Publisher)` with a Mono which should provide room for further optimization at that level.

Same thing on the client request side, where `ReactorClientRequest#writeWith` can be optimized to preserve the mono too.

Suggestions are implemented in :
https://github.com/spring-projects/spring-framework/tree/optimize-mono-bodies

smaldini:


rstoyanchev:
I've applied a similar fix except for the doOnDiscard which has to be inside the flatMap and addressed some issues related to that. For the future, if you submit a PR after the issue, please cross-reference the two. I'm afraid the reference in the title doesn't do it.

rstoyanchev:
Also for PRs please use your own fork rather than pushing branches to this repo.

PR #22804, opened 2019-04-16
Make inner classes static where possible

stsypanov:
Rid a reference to outer class from where it's not required.


sbrannen:

Thanks

PR #22810, opened 2019-04-17
Reduce unnecessary memory waste in PostProcessorRegistrationDelegate's registerBeanPostProcessors()

hecenjie:
We can delete list orderedPostProcessorNames AND nonOrderedPostProcessorNames which are unnecessary here, cause we can put these operations into the same for loop, both lists waste time and space, and make the code hard to understand.


jhoeller:
That part of the code is about `BeanPostProcessor` instantiation order, enforcing the creation of priority-ordered processor instances before attempting to instantiate regular ones. This allows priority-ordered processors to get applied to regular processor instances, whereas immediate creation of all processor instances would prevent such meta-processing of processor instances themselves. For that reason, the name-based separation of processor determination versus instantiation needs to stay as-is. It's very internal code that only runs once on startup, so the extra complexity is acceptable from my perspective.

pivotal-issuemaster:
@hecenjie Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22810)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=22810) to manually synchronize the status of this Pull Request.

PR #22816, opened 2019-04-18
Temporal should be simple value type like Date

quaff:


sbrannen:
Hi @quaff,

Can you please explain why you are suggesting this change?

quaff:
Temporal is the substitute of Date, and immutable, so it's more reasonable than Date being a simple value type.

PR #22829, opened 2019-04-23
Avoid possible memory leak in ResolvableType

dreis2211:
Hi,

in recent profiling sessions of Spring-Boot apps, I noticed that instances of ResolvableType are retained by their `DefaultVariableResolver`:
grafik
Until now I've not seen any critical leaks but I thought better be safe than sorry.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/ResolvableType.java
+++ b/spring-core/src/main/java/org/springframework/core/ResolvableType.java
@@ -935,7 +935,7 @@ public class ResolvableType implements Serializable {
 		if (this == NONE) {
 			return null;
 		}
-		return new DefaultVariableResolver();
+		return new DefaultVariableResolver(this);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jhoeller:
Isn't this just moving the `ResolvableType` instance reference in the inner class from implicit to explicit? What difference does this make in terms of garbage collection?

dreis2211:
The explicitness comes from the change to a static class. But you're right. I missed that `asVariableResolver()` is not public. As the `VariableResolver` can't really be exposed to the outside there is probably no immediate risk of a memory leak (unless someone does some weird reflection).

Having that said, I still think it is a worthwhile change. There is the consistency argument as `TypeVariablesVariableResolver` is static as well. Additionally, some IDEs (or tools) warn about non-static inner classes as they **can** cause memory leaks. In fact, the tool I used warned me about retained objects by their inner class - I've just missed to dig deeper (sorry for that).

Of course, feel free to decline the PR.

jhoeller:
Good point about consistency there, and it's an easy enough change to make indeed.

PR #22836, opened 2019-04-25
Add missing fail() calls in MockRestServiceServerTests

izeye:
This PR adds missing `fail()` calls in `MockRestServiceServerTests`.


snicoll:
Thanks again @izeye!

PR #22839, opened 2019-04-25
Add Nohttp Checks

rwinch:


src/nohttp/suppressions.xml
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE suppressions PUBLIC
+	"-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+	"https://checkstyle.org/dtds/suppressions_1_2.dtd">
+
+<suppressions>
+	<suppress checks="io.spring.nohttp.checkstyle.check.NoHttpCheck" files="BeanDefinitionParserDelegate.java" lines="1409"/>
```
snicoll: Do I understand this is referring to an hard-coded line? Couldn't we add something in the code instead?

build.gradle
```diff
@@ -1,9 +1,11 @@
 buildscript {
 	repositories {
 		maven { url "https://repo.spring.io/plugins-release" }
+		maven { url "https://repo.spring.io/plugins-snapshot" }
 	}
 	dependencies {
 		classpath("io.spring.gradle:propdeps-plugin:0.0.9.RELEASE")
+		classpath("io.spring.nohttp:nohttp-gradle:0.0.1.BUILD-SNAPSHOT")
```
snicoll: Is there a timeline for a release?

snicoll (commented):
Thanks for the PR @rwinch 

Looks good to me, I've added a few comments. Did you have a target branch in mind for this? Looks like `5.1.x` would be a good candidate.

build.gradle
jhoeller: I'm a bit concerned about baking such a very early-state plugin into existing builds, so I'd rather only have it in master for the 5.2 generation. As for the PR overall, the missing Apache license headers all affect new files in 5.2, so quite a bit seems to be master-specific to begin with. The only part I see as a 5.1 backport candidate are the URL examples in the reference docs.

rwinch: I expect to release it in the next week, but am wanting to get feedback from projects before that release is done.

src/nohttp/suppressions.xml
rwinch: I'm not sure I understand. Can you expand on what you mean by this?

rwinch:
> Did you have a target branch in mind for this? Looks like 5.1.x would be a good candidate.

@snicoll Thanks for the comments. I was not sure where the Framework team wanted this merged. It appears that @jhoeller is ok with the general changes going into 5.1.x except for the nohttp being added to the build. I'll let him verify what his thoughts are.

As per the discussion on Slack, I pushed changes to make it so that BeanDefinitionParserDelegate now uses http://www.springframework/schema/ so it is automatically ignored by nohttp. This means there is no need for suppressions.xml at all anymore.

src/nohttp/suppressions.xml
jhoeller: I suppose this was meant towards what we discussed earlier today: a marker in the affected Java source file instead of an external reference to a line in a .java file. Even better that we don't need that suppression anymore now.

jhoeller:
Indeed, the documentation changes in particular could easily go into 5.1.x.
The license URL updates in the new source files are clearly master only, as are the tool-related changes.

rwinch:
@snicoll @jhoeller Thanks for all the feedback. Is there anything else you need from me on this?

snicoll:
@rwinch a build based on that branch does not work for me. Is that to be expected?

spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
```diff
@@ -1406,7 +1406,7 @@ public BeanDefinitionHolder decorateIfRequired(
 					return decorated;
 				}
 			}
-			else if (namespaceUri.startsWith("http://www.springframework.org/")) {
+			else if (namespaceUri.startsWith("http://www.springframework.org/schema/")) {
```
snicoll: Can you explain why that change is part of the nohttp check? Not denying that's useful but I expect something like this in a dedicated issue.

spring-test/src/test/java/org/springframework/test/web/servlet/htmlunit/HostRequestMatcherTests.java
```diff
@@ -31,14 +31,14 @@
 	public void localhost() throws Exception {
 		WebRequestMatcher matcher = new HostRequestMatcher("localhost");
 		assertMatches(matcher, "http://localhost/jquery-1.11.0.min.js");
-		assertDoesNotMatch(matcher, "http://example.com/jquery-1.11.0.min.js");
+		assertDoesNotMatch(matcher, "http://company.example/jquery-1.11.0.min.js");
```
snicoll: Shouldn't that use `https`?

spring-test/src/test/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilderTests.java
```diff
@@ -178,7 +178,7 @@ public void buildRequestContextPathUsesFirstSegmentByDefault() {
 
 	@Test
 	public void buildRequestContextPathUsesNoFirstSegmentWithDefault() throws MalformedURLException {
-		webRequest.setUrl(new URL("http://example.com/"));
+		webRequest.setUrl(new URL("https://example.com/"));
```
snicoll: I thought we shouldn't be using `example.com` anymore?

snicoll (commented):
Thanks for the PR, I've added a few comments. I must share I am bit confused as when to add `https` and when `http` is ok in example URLs. I wonder how the tool is supposed to catch that.

Also, 18 commits sounds a bit invasive for such a change. I had started to polish it to 3: one to add the plugin to the build, one to change the header to https and one with all the rest (with the note that one commit should be a separate PR IMO).

Would you mind rebase and squash the URLs changes in one commit?

spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
jhoeller: It seems the tool ignores "http://www.springframework.org/schema/" prefixes in general since that's used by XML namespace declarations, so it automatically also ignores that Java source bit if its String contains that full prefix.

jhoeller:
@rwinch As Stephane suggested, can we decompose this PR a bit? Separating the nohttp tool bit from the doc changes seems worthwhile indeed. Also, it's not entirely clear to me either which example URLs we should really be using now.

jhoeller:
FWIW, I'll make the license headers in those two new test sources consistent with the rest in a separate revision. That's really a plain oversight even aside from the nohttp tool, in contrast to the headers in the gradlew scripts which are generated... So those remaining changes are really primarily motivated by the tool, and we need to take into account that those gradlew scripts get regenerated on a Gradle upgrade.

rwinch:
> @rwinch a build based on that branch does not work for me. Is that to be expected?
> 
> ```
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> Execution failed for task ':checkstyleNohttp'.
> > Checkstyle rule violations were found. See the report at: file:///Users/snicoll/workspace/pivotal/spring-framework/build/reports/checkstyle/nohttp.html
>   Checkstyle files with violations: 1024
>   Checkstyle violations by severity: [error:6153]
> ```

It does not fail for me. Can you post more details (i.e. files and the failures)? Perhaps you have stale files from the build or IDE still? Did you try a fresh clone or using `git clean -dfxn` (remove the n and run again if you are fine with the files being removed).

> Thanks for the PR, I've added a few comments. I must share I am bit confused as when to add https and when http is ok in example URLs. I wonder how the tool is supposed to catch that.

See https://github.com/spring-io/nohttp/tree/master/nohttp#thought-process

 
> Also, 18 commits sounds a bit invasive for such a change. I had started to polish it to 3: one to add the plugin to the build, one to change the header to https and one with all the rest (with the note that one commit should be a separate PR IMO).
> 
> Would you mind rebase and squash the URLs changes in one commit?

No problem. Done.

spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
rwinch: Yes. The tooling ignores well known XML names. For Spring that is it starts with http://www.springframework.org/schema/ and does not end with `.xsd`

spring-test/src/test/java/org/springframework/test/web/servlet/htmlunit/HostRequestMatcherTests.java
rwinch: The test is validating another http URL does not match, so we need http here. In that case we can use See https://github.com/spring-io/nohttp/tree/master/nohttp#thought-process

spring-test/src/test/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilderTests.java
rwinch: Please see See https://github.com/spring-io/nohttp/tree/master/nohttp#thought-process

rwinch:
It seems there might be [differing](https://github.com/spring-projects/spring-framework/pull/22839#pullrequestreview-234351701) [opinions](https://github.com/spring-projects/spring-framework/pull/22839#issuecomment-490017215) on how to split up the commits. I updated to switch to how @snicoll suggested the commits be split up (mostly because it was the first comment I saw). If you prefer, please feel free to split up the commits in another way.

src/nohttp/suppressions.xml
sbrannen: > Even better that we don't need that suppression anymore now.

Was support added for a marker in the source code?

sbrannen: Never mind: I just read https://github.com/spring-projects/spring-framework/pull/22839#issuecomment-487636513.

sbrannen (approved):
I reviewed the changes within the GitHub UI, and I didn't notice anything that seemed out of place.

snicoll:
I've rebased this work but I am not keen to merge at this point:

* If your working copy is using IntelliJ IDEA, the IDE puts files in a separate location (`out` in each modules). If you happen to fix this outside the IDE it may not have catch up and will fail the build
* Running the plugin on each build is significantly increasing the total build time. Running `./gradlew nohttp` on the command line took almost 6 minutes, that's roughly, a third of our current build time
* I had one build failure with an OOM

Running `./gradlew nohttp` with my rebase leads to

```
> Task :nohttp FAILED

* What went wrong:
Execution failed for task ':nohttp'.
> Process 'command '/Library/Java/JavaVirtualMachines/zulu-8.jdk/Contents/Home/bin/java'' finished with non-zero exit value 60

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

@rwinch can you please have a look to it? I am particularly concerned at the build time and I don't think we should enable this for each build and increasing our CI feedback loop isn't great either.

rwinch:
@snicoll The `nohttp` task isn't really intended for anything other that fixing an existing codebase. At the moment, it [doesn't use settings](https://github.com/spring-io/nohttp/issues/16) from the nohttp extension. To use it, [follow the examples](https://github.com/spring-io/nohttp/tree/master/nohttp-gradle#nohttp-samples) in the documentation.

Please run `checkstyleNohttp` to verify the build. For me this runs in 3 seconds and passes. Note that the plugin adds `checkstyleNohttp` to a dependency to `check` task, so you running `check` executes `checkstyleNohttp`.

snicoll:
Yes, I've edited my comment in the meantime. It does not run in 3s for me though. What about the `out` directories? Excluding them would also allow duplicate checks.

rwinch:
> Yes, I've edited my comment in the meantime. 

Hmm...not sure I understand this. I don't [see an edit](https://github.com/spring-projects/spring-framework/pull/22839#issuecomment-501611083) since I commented?

image

> It does not run in 3s for me though.

How long does it take?

> What about the out directories? Excluding them would also allow duplicate checks.

You could add an exclusion on them in nohttp configuration if you want to. I'm not sure how much this is going to gain since `*.class` files are already excluded.

snicoll:
In my edit 5h ago, I wrote the following

>  I got confused initially as there is a separate task for checkstyle that took 20s.

rwinch:
Ok thanks that clarifies the edit and the amount of time it is taking. I'm still wondering:

> You could add an exclusion on them in nohttp configuration if you want to. I'm not sure how much this is going to gain since *.class files are already excluded.

snicoll:
@rwinch yeah probably my use case was odd. I hadn't rebuild your PR in my IDE so it blew up but I guess once we're good it's not really necessary. The only use case I see is someone making a mistake in an xml file, fixing it, then the build would fail again until a build is done at the IDE level. I still think excluding those directories would be better (especially as it represents a duplication)

rwinch:
I pushed an update to my branch that:
* merges your changes for rebasing off master
* excludes the out directory of each project

src/nohttp/whitelist.lines
```diff
@@ -0,0 +1,17 @@
+^http://[^/]*caucho.com.*
+^http://[^/]*gebish.org.*
+^http://[^/]*groovy-lang.org.*
+^http://[^/]*hsqldb.org.*
+^http://[^/]*sourceforge.net.*
```
JLLeitschuh: Why all of source forge?

sbrannen: Several projects that we link to from the documentation are published on Source Forge without HTTPS access.

I just updated two of them in beb454a29952bc4f2fff533424c88105012039d6, but we still have to link via `http://` for libraries such as JiBX and HtmlUnitSource Forge.

sbrannen: Or are you perhaps suggesting that we only (now) have explicit exclusions for `htmlunit.sourceforge.net` and `jibx.sourceforge.net`?

JLLeitschuh: Only having explicit exclusions sounds reasonable.

sbrannen: Thanks for the feedback.

@rwinch, can you make this change to the PR?

rwinch: @sbrannen Done

sbrannen: Thanks, Rob!

snicoll:
Running the build breaks for me as it's looking at `spring-framework/spring-websocket/bin/spring-websocket.log`. This may be an old artifact in my local repo but ignoring `.log` file looks sensible to me.

snicoll:
@rwinch did you test this branch against a fresh checkout or something (i.e. without building the framework first). I thought I had some outdated artifacts but it looks like the exclusions are not properly applied.

Here is a gist that shows the error I currently have: https://gist.github.com/snicoll/ac6b5ba8c469bd79a22183704b5df852

I am happy to revisit those and polish the PR but I'd like to first understand if I am missing something and if it's ok to exclude those resources.

rwinch:
The build works for me even though I have built other branches.

It seems like the failures you are getting are due to the build directory of modules that no longer exist in source control. For example, `spring-instrument-tomcat` and `spring-struts` do not exist in master. We should only exclude build directories of actual projects. You should make sure to clean that up before checking.

snicoll:
Sorry @rwinch I was focused on the `spring-websocket/bin` and I thought that one was legit. I should have reviewed the list more carefully.

It's now merged, I also upgraded to `0.0.3.RELEASE`. Thanks!

sbrannen:
The updates to `whitelist.lines` somehow got lost in the merge, but I added them in 5e9a22d118778487255a1e1d87c2368f69f2e329.

So hopefully that takes care of everything for this PR.

snicoll:
Ooops, for some reason I had a stale version of the PR. Thanks for noticing Sam!

PR #22870, opened 2019-05-02
Abort early in PropertyPlaceholderHelper if no placeholder exists

dreis2211:
Hi,

I am currently working on https://github.com/spring-projects/spring-boot/issues/16401 and investigate performance issues in the binding context. I noticed a small optimization opportunity in PropertyPlaceholderHelper if no placeholders exist. We can early out in those cases and thus save the allocations coming from the StringBuilder (and the HashSet).
To be fair: this is not a major performance bottleneck at all (there are ~5000 StringBuilder/HashSet allocations that we could save), but every little helps and who knows what other people might be doing with that class.

Cheers,
Christoph


jhoeller:
I've addressed this slightly differently, only doing the `contains`/`indexOf` check once in `parseStringValue` and rather passing in `null` for the `HashSet,` lazily initializing it within the recursion.

Thanks for raising this, in any case!

Issue #22896, opened 2019-05-06
DefaultListableBeanFactory.destroySingleton(s) not thread-safe for manual singletons


lmagyar89:
Within class `org.springframework.beans.factory.support.DefaultListableBeanFactory` field `manualSingletonNames` almost correctly synchronized.
But in method `destroySingleton` and `destroySingletons` directly modify the Set, there should be an immutable-like reference override as in case other modifications e.g.: `registerBeanDefinition`.

lmagyar89:
My PR at https://github.com/spring-projects/spring-framework/pull/22898

PR #22898, opened 2019-05-06
Fix direct modification to manualSingletonNames

lmagyar89:


jhoeller:
I'll deal with this under the original issue #22896 as part of a slightly wider refactoring. Thanks for the PR, in any case!

In general, please create a PR right away if there's a PR coming. No need for a separate GitHub issue in such a case, and even if the PR is not used as-is, the GitHub PR entry can be treated like a regular issue.

PR #22918, opened 2019-05-07
Fix suspend/resume in AbstractReactiveTransactionManager

mp911de:
We now correctly unwrap suspended resources instead of capturing the `Mono` emitting suspended resources.

We also properly continue resume by chaining resume Mono's instead of terminating eagerly.


mp911de:
That one is be the last one that we found during R2DBC and MongoDB transaction integration.

PR #22921, opened 2019-05-07
Extend MultipartBodyBuilder methods so they accept a filename parameter

### LINKED ISSUE #22616: Content-Disposition for multipart formdata is not correctly set when using Mono<DataBuffer> source
I tried for hours to upload data via multipart post request with WebClient. As async part I use a mono from a databuffer:

```
    final MultipartBodyBuilder builder = new MultipartBodyBuilder();
    builder.asyncPart("file", Mono.just(someDataBuffer), DataBuffer.class);
    MultiValueMap<String, ?> data = builder.build();
    
    return webClient
        .post()
        .uri("http://myurl/upload")
        .body(BodyInserters.fromMultipartData(data))
        .retrieve()
        .bodyToMono(String.class);
```

https://github.com/spring-projects/spring-framework/blob/44df98c82da75114081ba02364e467004252bb77/spring-web/src/main/java/org/springframework/http/codec/multipart/MultipartHttpMessageWriter.java#L291

That always failed because there is not filename known in the generated content disposition header. A good way to solve this would be to add a filename parameter to the asyncPart builder.

koziolk:
This fix allows passing a filename to MultipartBodyBuilder methods and
sets the proper value of the content disposition header based on it.

Closes gh-22616

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/MultipartBodyBuilder.java
+++ b/spring-web/src/main/java/org/springframework/http/client/MultipartBodyBuilder.java
@@ -32,6 +32,7 @@ import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
+import org.springframework.util.StringUtils;
 
 /**
  * A mutable builder for multipart form bodies. For example:
@@ -124,7 +125,7 @@ public final class MultipartBodyBuilder {
 		Object partBody;
 		HttpHeaders partHeaders = new HttpHeaders();
 
-		if (filename != null) {
+		if (StringUtils.hasLength(filename)) {
 			partHeaders.setContentDispositionFormData(name, filename);
 		}
 
--- a/spring-web/src/main/java/org/springframework/http/client/MultipartBodyBuilder.java
+++ b/spring-web/src/main/java/org/springframework/http/client/MultipartBodyBuilder.java
@@ -78,7 +78,7 @@ public final class MultipartBodyBuilder {
 	 * Add a part from an Object.
 	 * @param name the name of the part to add
 	 * @param part the part data
-	 * @param contentType MediaType which is used to determine how to encode the part
+	 * @param contentType used to determine how to encode the part
 	 * @return builder that allows for further customization of part headers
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/client/MultipartBodyBuilder.java
```diff
@@ -100,6 +124,10 @@ public PartBuilder part(String name, Object part, @Nullable MediaType contentTyp
 		Object partBody;
 		HttpHeaders partHeaders = new HttpHeaders();
 
+		if (filename != null) {
```
ielatif: You should also check that filename has length.
if (StringUtils.hasLength(filename))

spring-web/src/main/java/org/springframework/http/client/MultipartBodyBuilder.java
```diff
@@ -137,6 +165,30 @@ public PartBuilder part(String name, Object part, @Nullable MediaType contentTyp
...
+	/**
+	 * Add an asynchronous part with {@link Publisher}-based content.
+	 * @param name the name of the part to add
+	 * @param filename the filename of the part to add
+	 * @param publisher the part contents
+	 * @param elementClass the type of elements contained in the publisher
+	 * @return builder that allows for further customization of part headers
+	 */
+	public <T, P extends Publisher<T>> PartBuilder asyncPart(
+			String name, String filename, P publisher, Class<T> elementClass) {
```
ielatif: filename should be optional.

spring-web/src/main/java/org/springframework/http/client/MultipartBodyBuilder.java
koziolk: Right, fixing in my next change.

rstoyanchev:
I'm not sure this is headed in the right direction, it doubles the number of methods. I would add a single dedicated method for file parts, something like:

```java
public PartBuilder filePart(String name, String filename, Publishder<DataBuffer> content);
```

Also more generally, please organize into logical commits via squash. In this PR I would expect one commit only.

rstoyanchev:
I'll resolve this differently in #23083 through a filename parameter on `PartBuilder`.

Issue #22821, opened 2019-04-19
LinkedCaseInsensitiveMap track removals via keySet, entrySet or values


Deadpool1111:
I'm learning spring boot (using Spring-boot 2.1.3) while playing with REST API calls i found something interesting with `HttpHeaders` object

Suppose if i have an HttpHeaders object with some values

```java
      HttpHeaders headers = new HttpHeaders();
      headers.add("Connection", "keep-alive");
```

But now if i try to delete and add them again it throws `NullPointerException`

```java
      headers.entrySet().removeIf(entry -> entry.getKey().equals("Connection"));
      headers.add("Connection", "keep-alive");
```

Error:

```
       Caused by: java.lang.NullPointerException: null
	at org.springframework.util.CollectionUtils$MultiValueMapAdapter.add(CollectionUtils.java:460) ~[spring-core-5.1.5.RELEASE.jar:5.1.5.RELEASE]
	at org.springframework.http.HttpHeaders.add(HttpHeaders.java:1559) ~[spring-web-5.1.5.RELEASE.jar:5.1.5.RELEASE]
	at com.demo.DemoMain.run(DemoMain.java:27) ~[main/:na]
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:813) ~[spring-boot-2.1.3.RELEASE.jar:2.1.3.RELEASE]
	... 5 common frames omitted
```

rhamedy:
@philwebb should this be a `bug` to be fixed, I offer to work on it 👍

Deadpool1111:
@philwebb @rhamedy How this works? any one can contribute to this? i thought only spring people are allowed to fix these and outsiders are just to report bugs, correct me please if i'm wrong

philwebb:
@Deadpool1111 @rhamedy Thanks for the offer of help. Anyone is free to contribute to Spring as long as they are happy to sign a CLA. Submissions are usually handled via pull-requests which are reviewed by a member of the team.

For this specific issue, I think we should wait until someone from the Framework team has reviewed it. It looks like the fix might be quite involved as we'll probably need to return specialized types from `getValues()` `getKeys()` and `getKeySet()` that can track removals and update the `caseInsensitiveKeys`.

sbrannen:
FYI: I introduced failing tests that reproduce this bug in e187a42bfca79b75548cdf6dd5c46298ccfa8134.

nkjackzhang:
> 
> 
[...]

PR #22926, opened 2019-05-08
Polish LinkedCaseInsensitiveMap

izeye:
This PR polishes `LinkedCaseInsensitiveMap`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
+++ b/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
@@ -239,32 +239,26 @@ public class LinkedCaseInsensitiveMap<V> implements Map<String, V>, Serializable
 
 	@Override
 	public Set<String> keySet() {
-		Set<String> keySet = this.keySet;
-		if (keySet == null) {
-			keySet = new KeySet(this.targetMap.keySet());
-			this.keySet = keySet;
+		if (this.keySet == null) {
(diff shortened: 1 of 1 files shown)


jhoeller:
Following up on #22821, this would indeed benefit from some further polishing. However, those lazy key/value/entry are really meant to be published in a thread-safe fashion, preserving the notion of multi-threaded exposure for read-only purposes... so those fields should be declared volatile, in which case the current initialization pattern makes sense for thread-safe publication purposes.

I'll do a custom commit for the volatile declarations, also picking up your EntryIterator refactoring in it. Thanks for the PR, in any case!

izeye:
@jhoeller Thanks for the quick feedback!

> in which case the current initialization pattern makes sense for thread-safe publication purposes.

If "the current initialization pattern" means [having a local variable](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java#L240-L268), I'd like to understand that removing the local variable causes any problem. Could you please explain on this?

jhoeller:
It's just a defensive coding pattern where the local copy of a volatile field can be reliably checked against (in case of multiple checks, it's a single consistent value which is not going to reflect concurrent changes from other threads), and where we're optimizing access to the volatile field to begin with (one read attempt only). We're using a similar pattern in other places.

izeye:
@jhoeller Thanks for clarification :+1:

PR #22927, opened 2019-05-08
Getters for LocalSessionFactoryBeans

djechelon:
Implement public getter methods for LocalSessionFactoryBean

Motivation: we have extended LSFB in order to perform additional actions to compute the `annotatedClasses`.
In general, it would be advisable that a class extending the LSFB can read the properties other than setting them.

For example, if you want to _add_ an annotated class using your own logic, you can invoke both `getAnnotatedClasses`, edit the array, and then `setAnnotatedClasses` with the new array.

In this scenario, it could also have been possible to add a `protected addAnnotatedClasses` or `protected addXXX` method

### DIFF
--- a/spring-orm-hibernate4/src/main/java/org/springframework/orm/hibernate4/LocalSessionFactoryBean.java
+++ b/spring-orm-hibernate4/src/main/java/org/springframework/orm/hibernate4/LocalSessionFactoryBean.java
@@ -135,6 +135,17 @@ public class LocalSessionFactoryBean extends HibernateExceptionTranslator
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


djechelon:
As a practical example, we currently (due to poor old design choices) choose the annotated classes that compose the SessionFactory by using both the inclusion filters (e.g. `@Entity`, `@MappedSuperclass`, `@Embeddable`) **and** a pluggable exclusion filter.

So in our application we do not currently rely on `setPackagesToScan` out of the box.

snicoll:
Thanks for the PR but this hasn't get a lot of traction and I am not keen to open up the API more as it limits our ability to refactor it if we want to.

PR #22938, opened 2019-05-09
Spelling fix in `core-expressions.adoc`

crewmanmud:
Obvious Fix


snicoll:
@crewmanmud good catch, thank you!

PR #22982, opened 2019-05-16
Simplify iteration over a map

stsypanov:
subj


jhoeller:
I've included this in a revision on the 5.1.x branch, just using `values` instead of `objects` as the local variable name there. Thanks for the PR, in any case!

PR #22992, opened 2019-05-18
Cannot enhance @Configuration bean definition

### LINKED ISSUE #22990: When using Configuration annotation BeanDefinitionRegistryPostProcessor, the instance type of dependency injection is inconsistent with getBeanClassName in BeanDefinition
When using Configuration annotation BeanDefinitionRegistryPostProcessor, the instance type of dependency injection is inconsistent with getBeanClassName in BeanDefinition. The code is as follows

```java
@Configuration
public class MyBeanDefinitionRegistryPostProcessor implements BeanDefinitionRegistryPostProcessor {

    @Override
    public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry beanDefinitionRegistry) throws BeansException {
        System.out.println();
    }

    @Override
    public void postProcessBeanFactory(ConfigurableListableBeanFactory configurableListableBeanFactory) throws BeansException {
        System.out.println();
    } 
}
```

```java
@RunWith(SpringRunner.class)
@SpringBootTest(classes = {DemoApplication.class})
public class DemoApplicationTests {

    @Autowired
    private ConfigurableApplicationContext context;

    @Autowired
    private MyBeanDefinitionRegistryPostProcessor myBeanDefinitionRegistryPostProcessor;

    @Test
    public void beanClassTest() {

        String beanClass = myBeanDefinitionRegistryPostProcessor.getClass().getName();

        String beanDefinitionBeanClass = context.getBeanFactory()
                .getBeanDefinition("myBeanDefinitionRegistryPostProcessor")
                .getBeanClassName();

        System.out.println(beanClass);
        System.out.println(beanDefinitionBeanClass);

        Assert.assertEquals("beanClass should equals beanDefinitionBeanClass", beanClass, beanDefinitionBeanClass);
    }
}
```
 
image

The reasons are as follows:

The method enhanceConfigurationClasses in org.springframework.context.annotation.ConfigurationClassPostProcessor ,as long as beanDef is a full configuration, it will be added to configBeanDefs and then generated by cglib into the new class, resulting in updating the beanClassName in BeanDefinition. The code should be changed to the following

image

lgxbslgx:
Cannot enhance @Configuration bean definition if its singleton instance has been created too early.
Add the corresponding testcase.
Fix #22990.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassPostProcessor.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassPostProcessor.java
@@ -405,13 +405,17 @@ public class ConfigurationClassPostProcessor implements BeanDefinitionRegistryPo
 					throw new BeanDefinitionStoreException("Cannot enhance @Configuration bean definition '" +
 							beanName + "' since it is not stored in an AbstractBeanDefinition subclass");
 				}
-				else if (logger.isInfoEnabled() && beanFactory.containsSingleton(beanName)) {
-					logger.info("Cannot enhance @Configuration bean definition '" + beanName +
-							"' since its singleton instance has been created too early. The typical cause " +
-							"is a non-static @Bean method with a BeanDefinitionRegistryPostProcessor " +
-							"return type: Consider declaring such methods as 'static'.");
+				else if (beanFactory.containsSingleton(beanName)) {
+					if (logger.isInfoEnabled()) {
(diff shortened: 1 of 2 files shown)


mentallurg (approved):

snicoll:
Thanks for the PR but skipping the  enhancement step for circular reference scenarios where the enhanced class won't be used is not going to provide anything. This scenario should be fixed at the user level anyway and we're not keen to add a change for that scenario. Thanks anyway for the PR.

PR #23002, opened 2019-05-20
Add FreeMarker macro support in spring-webflux

### LINKED ISSUE #22950: spring.ftl is missing in spring-webflux
**Affects:** `spring-webflux` 5.1.6

In `org.springframework.web.reactive.result.view.freemarker.FreeMarkerConfigurer` the `spring.ftl` is mentioned, but actually does not exist.

I tried to use the one shipped with `spring-webmc`, but that led to further problems.

Please provide a webflux ready version of the ftl. 

Thank you!

ielatif:
See #22950 for details.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/view/freemarker/FreeMarkerConfigurer.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/view/freemarker/FreeMarkerConfigurer.java
@@ -46,8 +46,6 @@ import org.springframework.util.Assert;
  * instance via the "configuration" property. This allows to share a FreeMarker
  * Configuration for web and email usage for example.
  *
- * <p>TODO: macros
- *
  * <p>This configurer registers a template loader for this package, allowing to
  * reference the "spring.ftl" macro library contained in this package:
  *
@@ -59,6 +57,7 @@ import org.springframework.util.Assert;
  * Note: Spring's FreeMarker support requires FreeMarker 2.3 or higher.
  *
  * @author Rossen Stoyanchev
+ * @author Issam El-atif
  * @since 5.0
  */
 public class FreeMarkerConfigurer extends FreeMarkerConfigurationFactory
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/view/freemarker/FreeMarkerConfigurer.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/view/freemarker/FreeMarkerConfigurer.java
@@ -57,7 +57,6 @@ import org.springframework.util.Assert;
  * Note: Spring's FreeMarker support requires FreeMarker 2.3 or higher.
  *
  * @author Rossen Stoyanchev
- * @author Issam El-atif
  * @since 5.0
  */
 public class FreeMarkerConfigurer extends FreeMarkerConfigurationFactory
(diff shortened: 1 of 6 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/result/view/freemarker/FreeMarkerConfigurer.java
```diff
@@ -59,6 +57,7 @@
  * Note: Spring's FreeMarker support requires FreeMarker 2.3 or higher.
  *
  * @author Rossen Stoyanchev
+ * @author Issam El-atif
```
sbrannen:
```suggestion
```

We don't add ourselves as authors of a class if the only change we make is deletion of a TODO. So please revert this change.

spring-webflux/src/main/resources/org/springframework/web/reactive/result/view/freemarker/spring.ftl
```diff
@@ -0,0 +1,353 @@
...
+ * Spring's FreeMarker support will automatically make this file and therefore
+ * all macros within it available to any application using Spring's
+ * FreeMarkerConfigurer.
+ *
+ * To take advantage of these macros, the "exposeSpringMacroHelpers" property
+ * of the FreeMarker class needs to be set to "true". This will expose a
+ * RequestContext under the name "springMacroRequestContext", as needed by
+ * the macros in this library.
+ *
+ * @author Issam El-atif
```
sbrannen:
```suggestion
 * @author Darren Davison
 * @author Juergen Hoeller
 * @author Issam El-atif
```

Please include the original authors.

sbrannen (changes requested):
I've requested two minor changes.

Please introduce tests for this functionality as well.

Thanks in advance!

spring-webflux/src/main/java/org/springframework/web/reactive/result/view/freemarker/FreeMarkerConfigurer.java
ielatif: Done. Got it :)

spring-webflux/src/main/resources/org/springframework/web/reactive/result/view/freemarker/spring.ftl
ielatif: Done.

ielatif:
> Please introduce tests for this functionality as well

Done 👍

sbrannen (approved):

sbrannen (changes requested):
Thanks for making the recent changes and adding the tests!

Unfortunately, we have recently switched to AssertJ for all assertions. So the test classes will fail the build due to Checkstyle violations.

Please update the tests to use AssertJ instead of Hamcrest and JUnit 4.

You can use the latest versions (in `master`) of the test classes you copied as a basis.

Thanks in advance.

ielatif:
@sbrannen Thanks.
Migrated JUnit 4 assertions to AssertJ

sbrannen (approved):

sbrannen:
This has been merged into `master` in b58f34d802f4503789bbf94576e1cb9c8eb92eb8 and refined in 2f640fe20569a6e08794934c579e3ad00550325a.

Thanks again for the PR!

PR #23013, opened 2019-05-22
Polish ImportAwareTests

izeye:
This PR polishes `ImportAwareTests` by removing unused code.


jhoeller:
It seems that those types were meant to used in a mix with the `ConfigurationOne`/`ConfigurationTwo` override tests. I'll revise this accordingly.

PR #23033, opened 2019-05-25
Enforce the use of BDDMockito

sbrannen:
This commit modifies the existing Checkstyle rule for Mockito to
enforce the use of BDDMockito for all interactions with the core
Mockito APIs.

In addition, this commit migrates to consistent use of the MockitoRule
for JUnit 4.


sbrannen:
@spring-projects/spring-framework, I have submitted this PR for team discussion.

I feel that it's cleaner to work only with `BDDMockito` in imports; however, that is somewhat debatable since IDEs (such as Eclipse) may automatically convert static imports for methods in `Mockito` (e.g., `mock()`, `verify()`, etc.) from `BDDMockito` to `Mockito` (which actually declares the static methods) which results in a Checkstyle violation based on the Checkstyle change in this PR.

In light of that, I would like feedback from the team before merging this into `master`.

If the team does not want the change to the Checkstyle rule, I will revert those changes and likely retain the other changes in this commit (i.e., use of the `MockitoRule` and polishing).

philwebb:
I don't think we should do this, it will cause quite a bit of pain in the IDE (for example, I can't run a full organize imports run). It also feel incorrect to me because the `BDDMockito` class does not actually declare those static methods.

sbrannen:
> I don't think we should do this, it will cause quite a bit of pain in the IDE (for example, I can't run a full organize imports run). It also feel incorrect to me because the `BDDMockito` class does not actually declare those static methods.

Understood, and I agree. I think I'll just extract the polishing from this PR.

Thanks for the feedback.

sbrannen:
Superseded by 141ef9082fed900c97bc294c95cc48bf7108fb73

PR #23036, opened 2019-05-27
Allocate objects only when it's necessary

stsypanov:


sbrannen:

Thanks

PR #23042, opened 2019-05-28
Add constants for application/cbor to MediaType

izeye:
This PR adds constants for `application/cbor` to `MediaType` and changes to use them where possible.


sdeleuze:
Since CBOR is now also used on RSocket side in a first class way, maybe it deserves its constants in `MediaType`. Any thoughts @bclozel @rstoyanchev?

PR #23051, opened 2019-05-29
Fix Javadoc for HttpHeaders.setContentLanguage()

izeye:
This PR changes `set()` to `addAll()` in Javadoc for `HttpHeaders.setContentLanguage()` as `set()` for the parameter types doesn't seem to exist and `addAll()` seems most suitable to the following description around available methods.


sdeleuze:
Thanks for spotting this, I fixed it by using `put` instead as suggested by @poutsma since `HttpHeaders` is also a `Map<String, List<String>>`.

PR #23061, opened 2019-05-30
Add support for escaped delimiter in StringUtils when splitting strings

fmcejudo:
Let `StringUtils` read the values with delimiter inside, but escaping such delimiter with backward slash.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -1219,16 +1219,59 @@ public abstract class StringUtils {
 		else {
 			int pos = 0;
 			int delPos;
-			while ((delPos = str.indexOf(delimiter, pos)) != -1) {
-				result.add(deleteAny(str.substring(pos, delPos), charsToDelete));
+			while ((delPos =  nextDelimiterPosition(str, delimiter, pos)) != -1) {
+				result.add(extractEffectiveValue(str.substring(pos, delPos), delimiter, charsToDelete));
 				pos = delPos + delimiter.length();
 			}
 			if (str.length() > 0 && pos <= str.length()) {
 				// Add rest of String, but not in case of empty input.
-				result.add(deleteAny(str.substring(pos), charsToDelete));
+				result.add(extractEffectiveValue(str.substring(pos), delimiter, charsToDelete));
 			}
 		}
 		return toStringArray(result);
+	}
+
+	/**
+	 * It finds the next delimiter position not escaped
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@fmcejudo Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23061)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23061) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@fmcejudo Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23061)!

fmcejudo:
@jhoeller @snicoll do you think this can be helpful? the idea is including the delimiter as part of the string values.

Sorry for disturbing you

snicoll:
No need to apologize @fmcejudo and sorry for the late handling of this PR. I can see how having a way to ignore certain delimiters can be useful but I don't think we could change the semantic of this method without breaking existing use cases (that may rely on `\` to be present before the delimiter). 

I haven't seen a lot of demand for that particular feature either so perhaps it's best if such dedicated helper would stay in your own code. Or perhaps we could extend the existing API to offer an option but we probably want to be consistent with related helper methods.

Flagging for team attention to get more feedback.

sbrannen:
> but I don't think we could change the semantic of this method without breaking existing use cases (that may rely on `\` to be present before the delimiter).

I agree.

> Or perhaps we could extend the existing API to offer an option but we probably want to be consistent with related helper methods.

If we add support for escaped delimiters, I think that would be the way to go: new overloaded variants that accept a `boolean ignoreEscapedDelimiters` argument.

sbrannen:
**Team Decision**: Since existing clients of `StringUtils.delimitedListToStringArray(...)` within the Spring Framework cannot be modified to honor escaped delimiters (for backwards compatibility reasons), the addition of such proposed overloaded variants would be of little benefit to Spring itself. In light of that and the fact that `StringUtils` is primarily intended for use within the framework (see Javadoc quote below), the team has decided not to accept this PR.

_Javadoc_:
> Mainly for internal use within the framework; consider [Apache's Commons Lang](https://commons.apache.org/proper/commons-lang/) for a more comprehensive suite of String utilities.

Our recommendation is therefore to use a third-party library or implement the feature yourself.

fmcejudo:
Thanks

PR #23070, opened 2019-05-31
Fixes issue with optional @RequestPart Mono argument being resolved to null instead of Mono.empty

### LINKED ISSUE #23060: Optional Mono parameter in webflux mapping should not be null
**Affects:** 5.1.7

---
In this call
```java
  @PostMapping(value = "/{callId}", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
  public Mono<Void> store(
      @PathVariable final String callId,
      @RequestPart(value = "fieldOne", required = false) final Mono<String> fieldOne) 
```

If `fieldOne` is not provided then it should be empty Mono instead of null.

L00kian:
Fixes #23060


rstoyanchev:
@L00kian thank you for the pull request! I've applied it to 5.1.x and forward merged to master.

PR #23098, opened 2019-06-07
Simplify comparison of primitives

stsypanov:
Use utility method `compare()` to compare numbers.


sbrannen (changes requested):
Although the changes to `AspectJPrecedenceComparator` do in fact remove quite a bit of code, I don't think the code is any clearer now. In fact, I think it's now harder to follow, and the inline comments are now incorrect.

So please revert the changes to `AspectJPrecedenceComparator`, and I'll merge in the changes to the other two classes which look fine.

Thanks

stsypanov:
@sbrannen done

sbrannen (approved):

sbrannen:

Thanks

PR #23103, opened 2019-06-09
Use try with resources

ielatif:


spring-core/src/main/java/org/springframework/core/io/AbstractResource.java
```diff
@@ -153,13 +152,6 @@ public long contentLength() throws IOException {
 			}
 			return size;
 		}
-		finally {
-			try {
-				is.close();
-			}
-			catch (IOException ex) {
```
sbrannen: Although the method declares that it may throw an `IOException`, this is potentially a breaking change: with the removal of the catch block this can now result in an `IOException` if an error occurs while closing the `InputStream`; whereas, such an error would have been previously ignored.

@jhoeller What are your thoughts on this change?

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/core/io/AbstractResource.java
stsypanov: I think this is a good change, otherwise framework ignores exceptions resulting in probable leaks and hanging resources. When this inspection is thrown up it may be visible in logs giving a developer a message about somthing is wrong.

ielatif: Swallowing exceptions hides information that could be useful for debugging

> whereas, such an error would have been previously ignored

@sbrannen @jhoeller I am curious to know if the intention here was only to satisfy the compiler 🤔.

sbrannen: The rationale is that exceptions are rarely thrown while closing the input stream, and Spring often silently ignores such exceptions when closing an input stream, a database connection, etc.

Furthermore, this code is in the abstract base class. Many (most?) concrete subclasses actually override this method with something more meaningful. Take a look at FileSystemResource, FileSystemResource, ByteArrayResource, etc. for examples.

Having said that, I'll update `AbstractResource.contentLength()` so that it logs the exception at DEBUG level in case developers wish to know about it while debugging an application.

sbrannen:
superseded by gh-23116

Issue #23114, opened 2019-06-12
provide `.queryParam(String name, Collection<?> items)` to provide multiValued query parameters


bjconlan-fc:
When using the `UrlBuilder` (via the `WebClient`) you can pass a collection as the query parameter value like:

```java
webClient.get().uri(uriBuilder -> uriBuilder.path("/find")
  .queryParam("items", List.of("dogs", "cats"))                   // <!-- this
  .build()
```

This doesn't generate the expected 'multivalued' query param (ie `?items=dogs&items=cats` or `?items=dogs,cats`).

It would be good if it were possible to do something like this.

NOTE: `.queryParam(String name, Object... varargs)` does solve this problem, but having 1st class collection support would be much nicer than shimming things to work with this (or using `StringUtils.collectionToDelimitedString(List.of("dogs", "cats"), ",")`).

nosan:
May I tackle this issue?

rstoyanchev:
Superceded by pull request #23119.

PR #23119, opened 2019-06-12
Add queryParams and replaceQueryParams with Collection to UriBuilder

nosan:
Add `UriBuilder.queryParams(name, Collection<?> values)` and `UriBuilder.replaceQueryParams(name, Collection<?> values)` methods to provide multiValued query parameters.

gh-23114


rstoyanchev:
Thanks for the pull request. I've made a couple of adjustments. 
* dropped the "s" from the method names as they both deal with a single parameter. 
* switched the implementations around so that the new List variants delegate to the existing varag variant and not other way around, to avoid the extra List creation for all existing methods.

nosan:
Thank you very much @rstoyanchev

Issue #23123, opened 2019-06-13
method postForObject in RestTemplate causes NPE


TRockis:
**Affects:** \<4.3.13-RELEASE>
---
Hi, 
  When I was using postForObject method of RestTemplate, it may occur NPE. after debuggging, I found the exception was caused by 772 line in org.springframework.web.client.RestTemplate.java
```java
public void doWithRequest(ClientHttpRequest request) throws IOException {
    if (this.responseType != null) {
        Class<?> responseClass = null;
        if (this.responseType instanceof Class) {
            responseClass = (Class<?>) this.responseType;
        }
        List<MediaType> allSupportedMediaTypes = new ArrayList<MediaType>();
        for (HttpMessageConverter<?> converter : getMessageConverters()) {
            if (responseClass != null) {
                if (converter.canRead(responseClass, null)) {        //converter may be null, so NPE may occur. and I don't know why it may be null, and why not do something to prevent this exception occuring
                    allSupportedMediaTypes.addAll(getSupportedMediaTypes(converter));
                }
            }
            else if (converter instanceof GenericHttpMessageConverter) {
                GenericHttpMessageConverter<?> genericConverter = (GenericHttpMessageConverter<?>) converter;
                if (genericConverter.canRead(this.responseType, null, null)) {
                    allSupportedMediaTypes.addAll(getSupportedMediaTypes(converter));
                }
            }
        }
        if (!allSupportedMediaTypes.isEmpty()) {
            MediaType.sortBySpecificity(allSupportedMediaTypes);
            if (logger.isDebugEnabled()) {
                logger.debug("Setting request Accept header to " + allSupportedMediaTypes);
            }
            request.getHeaders().setAccept(allSupportedMediaTypes);
        }
    }
}
```
I don't know whether the exception occers because I used such expression such as 
```java
restTemplate.getMessageConverters().add(0,new StringHttpMessageConverter(Charset.forName("UTF-8")));
```
or
```java
rest = new RestTemplate();
rest.getMessageConverters().set(1, new StringHttpMessageConverter(StandardCharsets.UTF_8));
```
After the exception occured, I need restart my project, however, it may occur again after sometimes.
I do hope someone can give me some hints, thanks!

rustyTheClone:
Hello,
I tried to reproduct this problem.
This NPE occurs if you have more than one message converter, plus `null` coming before.

One way to fix this is to check for the use of `null`-converters.

[...]

PR #23132, opened 2019-06-13
Ignore null HttpMessageConverter in RestTemplate and HttpMessageConverterExtractor

fodil-a:
Prior to this commit, if a `null` `HttpMessageConverter` was configured
in the `RestTemplate`, this would lead to a `NullPointerException` once
the list of converters was accessed.

This commit avoids such exceptions by ignoring `null` converters.

See gh-23123

PR #23140, opened 2019-06-15
Ignore null HttpMessageConverter in RestTemplate and HttpMessageConverterExtractor

fodil-a:
Prior to this commit, if a `null` `HttpMessageConverter` was configured
in the `RestTemplate`, this would lead to a `NullPointerException` once
the list of converters was accessed.

This commit avoids such exceptions by ignoring `null` converters.

See gh-23123

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/HttpMessageConverterExtractor.java
+++ b/spring-web/src/main/java/org/springframework/web/client/HttpMessageConverterExtractor.java
@@ -103,7 +103,7 @@ public class HttpMessageConverterExtractor<T> implements ResponseExtractor<T> {
 					}
 				}
 				if (this.responseClass != null) {
-					if (messageConverter.canRead(this.responseClass, contentType)) {
+					if (messageConverter != null && messageConverter.canRead(this.responseClass, contentType)) {
 						if (logger.isDebugEnabled()) {
 							String className = this.responseClass.getName();
 							logger.debug("Reading to [" + className + "] as \"" + contentType + "\"");
--- a/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
@@ -25,6 +25,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Objects;
 import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


rustyTheClone:
This is an edit from pull request: https://github.com/spring-projects/spring-framework/pull/23132
**I needed to change my fork's branch.**

sbrannen:
Tentatively slated for 5.2 RC1 and considering whether to backport to 5.1.x.

sbrannen:
@rstoyanchev and @jhoeller, shall we consider this a bug instead of an enhancement and assign it to `5.1.x`?

rstoyanchev:
Having `null` converter is something we don't even expect and seeing the proposed checks makes it even stranger. Why would there be a null converter in the first place? I think the solution should be to reject these earlier where converters are configured.

rstoyanchev:
@sbrannen I don't consider it a bug. It's simply a missing validation.

sbrannen:
> @sbrannen I don't consider it a bug. It's simply a missing validation.

Thanks for the feedback.

I actually came to the same conclusion myself.

> Having null converter is something we don't even expect and seeing the proposed checks makes it even stranger. Why would there be a null converter in the first place?

I agree that it is a user error if a `null` converter is present.

> I think the solution should be to reject these earlier where converters are configured.

The challenge is that `RestTemplate#getMessageConverters()` is documented as follows.

> Return the list of message body converters.
> 
> The returned List is active and may get appended to.

Thus, users can modify the list (e.g., by appending/inserting `null`) at any time. In other words, `RestTemplate` returns a modifiable `ArrayList` of converters, and we probably shouldn't change that at this point.

I would rather perform the non-null validation every time the converters are accessed in `RestTemplate` and `HttpMessageConverterExtractor` and throw an `IllegalStateException` if we encounter a `null` reference, as opposed to simply ignoring them as in this PR.

@rstoyanchev, what do you think about that?

rstoyanchev:
> users can modify the list (e.g., by appending/inserting null) at any time.

I would definitely add checks in the constructor and setter since that would provide the benefit of an early warning. Beyond that I think it's not worth repeatedly checking for something that should not be happening in the first place. At that point it almost doesn't matter how it fails, NPE or IAE. Either way it fails it needs to be fixed in the application.

sbrannen:
Superseded by #23151

PR #23169, opened 2019-06-20
Added application/graphql media type

marceloverdijk:


sbrannen:
Tentatively slated for 5.2 RC1 for consideration

sbrannen:
### Team Decision

Spring Framework should only introduce `MediaType` constants related to features supported directly by the framework.

Spring Framework does not support GraphQL directly. Thus third parties adding that support to Spring or users relying on the media type for a particular project should provide that media type themselves.

In the end, it is as easy as `new MediaType("application", "graphql")`.

In light of that, we are closing this issue.

marceloverdijk:
Fair enough I understand, np.

marceloverdijk:
Maybe this one can be re-evaluated as https://github.com/spring-projects-experimental/spring-graphql started (or we just add a constant to that experimental project off course).

bclozel:
Postponing that decision was actually a good call, and it still is.
The "GraphQL over HTTP" spec group is not considering that as high priority right now and might even submit a different MIME type to the IANA registry, `"application/graphql+json"`. We'll probably keep `"application/json"` for now.

Thanks!

marceloverdijk:
Thx for joining @bclozel , maybe good to continue the discussion here https://github.com/spring-projects-experimental/spring-graphql/issues/11
Currently I only have https://graphql.org/learn/serving-over-http/ but if there is more information to share please do.

PR #23195, opened 2019-06-25
Handle absolute path on Windows in PathEditor

### LINKED ISSUE #22511: PathEditor does not handle absolute path on Windows
Consider using a windows absolute path in a property file:
`repository.directory= C:some_path`
We want to inject directly a Path using @Value:
```
    public MyRepository(@Value("${repository.directory}") Path repositoryDirectory) {
        this.repositoryDirectory = repositoryDirectory;
    }
```

1.  If we are using the original Windows path, with back-slash:
`repository.directory=C:\Users\lacasoub\myUser\Local\Temp`
Some class  (Properties.load() i guess) trims the slash and we end up with the following string: C:UsersmyUserAppDataLocalTemp.
Which is then concat with the Sprint boot server relative path, ending with a String like this: **C:\Users\myUser\AppData\Local\Temp\undertow-docbase.2615281017312972350.8080\C:UsersmyUserAppDataLocalTemp**

Its a normal behavior for properties file.

2. If we are using double back-slash, it works.

3. If we are using slash instead:
`repository.directory=C:/Users/myUser/AppData/Local/Temp`

PathEditor has the following behavior:
- Converting this string to an URI, which is strange because URI and Windows Path are not compliant. 
- Adding a "/" because it don't understand that's it's a Windows path
- Concat the relative path **/C:/Users/myUser/AppData/Local/Temp** with the Sprint boot server relative path, ending with a String like this: **C:\Users\myUser\AppData\Local\Temp\undertow-docbase.2615281017312972350.8080\C:\Users\myUser\AppData\Local\Temp**
- Java path API throws an exception because the path creating by Sprint Boot PathEditor class is invalid:
`Caused by: java.nio.file.InvalidPathException: Illegal char <:> at index`

The solution 2 works but the solution 3 should also works as the Java Path class allows Windows absolute path with back-slash (Linux style) and not the PathEditor.

Antoniossss:
Closes https://github.com/spring-projects/spring-framework/issues/22511

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/propertyeditors/PathEditor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/propertyeditors/PathEditor.java
@@ -23,6 +23,7 @@ import java.net.URISyntaxException;
 import java.nio.file.FileSystemNotFoundException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.regex.Pattern;
 
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceEditor;
@@ -42,7 +43,6 @@ import org.springframework.util.Assert;
  * if no existing context-relative resource could be found.
  *
  * @author Juergen Hoeller
- * @since 4.3.2
  * @see java.nio.file.Path
  * @see Paths#get(URI)
  * @see ResourceEditor
@@ -53,6 +53,7 @@ import org.springframework.util.Assert;
 public class PathEditor extends PropertyEditorSupport {
 
 	private final ResourceEditor resourceEditor;
+	private final Pattern WINDOWS_OS_ABSOLUTE_PATH_PREFIX_PATTERN = Pattern.compile("^[a-zA-Z]:.*$");
 
 
 	/**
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@Antoniossss Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23195)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23195) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Antoniossss Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23195)!

mentallurg (changes requested):
You don't address at least following case:
`\\computername\C$\some\path`

It can start with "\\".

It can contain "$".

Also you don't address pipes. Their names start on Windows with "\\". Java can perfectly write to and read from pipes on Windows.

yurkom:
Hello, @jhoeller 

Is there a chance to merge this pull request? Or what should be done to fix #22511 issue in the next minor release?

Thanks,
Yury

snicoll:
#22511 has been resolved so I am going to close this as a dupe. Sorry we didn't get to this sooner @Antoniossss.

PR #23198, opened 2019-06-26
Add no-op TransactionOperations implementation

vpavic:
A no-op implementation can be useful for:

- having `TransactionOperations` consuming code switch to non-transactional behavior by injecting `NoOpTransactionOperations` instead of `TransactionTemplate`
- testing purposes

For first point we have a use case in Spring Session's `JdbcOperationsSessionRepository`. I've used similar approach in a number of other projects so having Framework provide it OOTB would be nice. Also note that `TransactionOperations` isn't parameterized, but `TransactionOperations#execute` is which makes it impossible to have no-op implementation as lambda expression.

### DIFF
--- a/spring-tx/src/main/java/org/springframework/transaction/support/NoOpTransactionOperations.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/support/NoOpTransactionOperations.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


sbrannen:
Thanks for the PR, @vpavic.

We'll consider it.

> Also note that `TransactionOperations` isn't parameterized, but `TransactionOperations#execute` is which makes it impossible to have no-op implementation as lambda expression.

Yes, that is a bit unfortunate.

jhoeller:
Good idea in general! I've implemented this slightly differently as a static `withoutTransaction()` method on `TransactionOperations,` along the lines of the recently introduced `withDefaults()` method on `TransactionDefinition`.

vpavic:
I don't see anything pushed to the `master` yet so I'm not sure if I understood the described changes correctly, but if non-transactional behavior is implemented using a new operation on `TransactionOperations` I'm afraid that won't allow changing transactional behavior by injecting a different `TransactionOperations` implementation which was the motivation for this PR.

Ideally, `TransactionOperations` consuming code shouldn't be aware of the fact that there's no transaction involved. This is similar like to, for instance, changing propagation behavior by injecting a differently configured `TransactionTemplate`.

jhoeller:
I'm about to commit this along with a number of other changes now, so it'll be in master in a moment. Thanks for your patience :-)

As for the implementation, `withoutTransaction()` is a new static factory method on `TransactionOperations,` returning a `TransactionOperations` instance with a regular `execute` method which executes the callback without an actual transaction. This should allow for substitution underneath any `TransactionOperations` caller as intended.

Issue #23199, opened 2019-06-27
what is recommend log level for org.springframework packages


boojongmin:
when I destroy bean, I can not check error log.

*my logback configuration is*
```xml
<logger name="com.company" level="DEBUG"/>
 <root level="ERROR">
     <appender-ref ref="STDOUT" />
 </root>
```

*bean destroy method*
```java
@PreDestroy
void destory() {
  throw new RuntimeException("message");
}
```
*console log*
```shell
09:27:45.932 [Thread-8] INFO  o.s.b.f.s.DisposableBeanAdapter - Destroy method 'destroy' on bean with name 'sqsReceiver' threw an exception: java.lang.RuntimeException: message
```

I couldn't see logs because  my  configuration of 'org.springframework' is 'ERROR'.
I waste lots of time on debugging.

would you explain spring logging rules?

and I think better than log level **info** that change log level is **warn** or **error**.(**DisposableBeanAdapter** invokeCustomDestroyMethod)

snicoll:
The related PR was merged. The framework tries to be a good citizen so that WARN is a reasonable tradeoff. The level you apply depends certainly on the environment and the lifecycle of the application.

PR #23200, opened 2019-06-27
Raise log level when bean destruction fails

boojongmin:
related to  this issue https://github.com/spring-projects/spring-framework/issues/23199

logger runned on `org.springframework.beans.factory.support`  package.
but InvocationTargetException was makes by my code.
I want to check my error log.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/DisposableBeanAdapter.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/DisposableBeanAdapter.java
@@ -342,15 +342,10 @@ class DisposableBeanAdapter implements DisposableBean, Runnable, Serializable {
 		catch (InvocationTargetException ex) {
 			String msg = "Destroy method '" + this.destroyMethodName + "' on bean with name '" +
 					this.beanName + "' threw an exception";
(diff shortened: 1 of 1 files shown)


jhoeller:
On review, this is a regression from 5.0/4.3 where we mostly logged bean destruction exceptions at warn level. Restoring this in 5.1.9 seems feasible, even within the revised log level usage that we introduced in 5.1 (for less noise at info and warn level). That said, such exceptions are technically not a hard error but rather just a partial effect in a multi-step destruction process, so logging at error level would arguably create unnecessary (follow-up) noise for scenarios with many related beans that need to be destroyed. This is particularly relevant for production scenarios where error level may trigger operator escalations.

Generally speaking, we recommend running with info or at least warn log level in development. Also, destroy method ideally should not propagate exceptions but rather handle them locally (possibly even log them locally with a descriptive log message).

boojongmin:
@jhoeller 

Thank you for your kind reply.
Kotlin handles errors with an unchecked exception.
My mistake is that when I close a resource like db, I miss the try-catch block related to checked exception like Java. I think kotlin programmer might have experience like me when I close resources.
Thank you.

PR #23212, opened 2019-06-28
Generalize ServerResponse and WebClient body methods

sdeleuze:
The commit is a follow-up of d6b5c2005849e0676781ec1086bf8fc994e1a2c4
which applies the same principle to ServerResponse,
WebClient.RequestBodySpec and WebTestClient.RequestBodySpec body
methods.

It introduces a body(Object) method where the parameter can be:
 - Concrete value
 - Publisher of value(s)
 - Any other producer of value(s) that can be adapted to a
   Publisher via ReactiveAdapterRegistry

Variants with Class and ParameterizedTypeReference parameters are also
provided and syncBody(Object) is deprecated.

With this change, in Kotlin it is now mandatory to specify explicitly
the generic type (for example body<List<Foo>>(body)) in order to
invoke the ParameterizedTypeReference shortcut. If it is not specified,
the extension is shadowed by the Java body(Object) method which is not
able to use Kotlin type inference to resolve the generic type.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,8 @@ import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
 import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.core.ReactiveAdapter;
+import org.springframework.core.ReactiveAdapterRegistry;
 import org.springframework.core.ResolvableType;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.buffer.DataBuffer;
@@ -44,6 +46,7 @@ import org.springframework.util.MultiValueMap;
  *
  * @author Arjen Poutsma
  * @author Rossen Stoyanchev
+ * @author Sebastien Deleuze
  * @since 5.0
  */
 public abstract class BodyInserters {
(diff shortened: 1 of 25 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClient.java
```diff
@@ -585,8 +622,46 @@ static WebClient create(String baseUrl) {
...
+		 * the body of the part, or an
+		 * {@link org.springframework.http.HttpEntity HttpEntity} representing
+		 * a part with body and headers. The {@code MultiValueMap} can be built
+		 * with {@link org.springframework.http.client.MultipartBodyBuilder
+		 * MultipartBodyBuilder}.
+		 * @param body the body to write to the request
+		 * @return this builder
+		 * @since 5.2
+		 */
+		RequestHeadersSpec<?> body(Object body);
```
rstoyanchev: How about ordering this ahead of the other body methods? I would expect this one to be used most often.

spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
```diff
@@ -72,19 +75,86 @@
...
+	 * @param body the body to write
 	 * @param <T> the type of the body
 	 * @return the inserter to write a single object
 	 */
 	public static <T> BodyInserter<T, ReactiveHttpOutputMessage> fromObject(T body) {
+		Assert.notNull(body, "'body' must not be null");
+		ReactiveAdapter adapter = ReactiveAdapterRegistry.getSharedInstance().getAdapter(body.getClass());
+		ResolvableType bodyType;
+		if (body instanceof Publisher || adapter != null) {
+			bodyType = ResolvableType.forClass(Object.class);
```
rstoyanchev: So if it is async we use Object? Maybe I'm missing something but this doesn't seem right. In RSocketRequester we wait for the actual value in order to determine the type and then encode.

spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
```diff
@@ -72,19 +75,86 @@
...
 	 * {@link org.springframework.web.reactive.function.client.WebClient WebClient} and
 	 * {@link org.springframework.web.reactive.function.server.ServerResponse ServerResponse}.
-	 * @param body the body to write to the response
+	 * @param body the body to write
 	 * @param <T> the type of the body
 	 * @return the inserter to write a single object
 	 */
 	public static <T> BodyInserter<T, ReactiveHttpOutputMessage> fromObject(T body) {
+		Assert.notNull(body, "'body' must not be null");
+		ReactiveAdapter adapter = ReactiveAdapterRegistry.getSharedInstance().getAdapter(body.getClass());
```
rstoyanchev: Using the shared instance means if someone tries to register additional adapters, they don't have a way to influence this.

poutsma: Can't we use the hints mechanism to resolve the registry?

sdeleuze: @rstoyanchev Could you please elaborate? I am not sure to understand.

spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
sdeleuze: As discussed, I don't think this is possible given current `writer.write(publisher)` API, and also as pointed out by Rossen RSocket and WebFlux are different: Payload oriented (one message at a time) vs HTTP producing a complete stream. So I will provide only the variant with explicit types for now.

PR #23239, opened 2019-07-04
Rename PathPatternRouteMatcherTest to PathPatternRouteMatcherTests

izeye:
This PR renames `PathPatternRouteMatcherTest` to `PathPatternRouteMatcherTests` to align its name with the names of the other test classes.


sbrannen:
Good catch!

Thanks

Issue #23254, opened 2019-07-09
Simple method for converting lists


firatkucuk:
It might be good a simple method in Conversion Service to easily convert generic lists to some other target list. Like:

```java
var List<MyTargetPojo> newItems = conversionService.convertList(items, MyTargetPojo.class);
```

snicoll:
Thanks for the suggestion but we'd rather not add overloads for collections in `ConversionService`.

PR #23255, opened 2019-07-09
convertList method added

firatkucuk:
Implementation for #23254

A simple method for converting lists in an iterative way. Considered to get converter outside of the loop but getConverter method uses a cache mechanism and fetches from a HashMap this means. it's already O(1).

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
@@ -18,6 +18,7 @@ package org.springframework.core.convert.support;
 
 import java.lang.reflect.Array;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -28,6 +29,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.springframework.core.DecoratingProxy;
 import org.springframework.core.ResolvableType;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@firatkucuk Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23255)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23255) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@firatkucuk Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23255)!

spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
```diff
@@ -166,6 +167,23 @@ public boolean canBypassConvert(@Nullable TypeDescriptor sourceType, TypeDescrip
...
+	@SuppressWarnings("unchecked")
+	@Nullable
+	public <S, T> List<T> convertList(final @Nullable List<S> sourceList, final Class<T> targetClass) {
+
+		if (sourceList == null) {
+			return null;
+		}
+
+		if (sourceList.isEmpty()) {
+			return new ArrayList<>();
```
ivancea: Collections.emptyList() here, as it may return an already existing list

spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
```diff
@@ -166,6 +167,23 @@ public boolean canBypassConvert(@Nullable TypeDescriptor sourceType, TypeDescrip
 		return (converter == NO_OP_CONVERTER);
 	}
 
+	@SuppressWarnings("unchecked")
+	@Nullable
+	public <S, T> List<T> convertList(final @Nullable List<S> sourceList, final Class<T> targetClass) {
+
```
ivancea: I'm not sure if this line break is allowed by the code style. At least, I can't find any references to it. [Code Style](https://github.com/spring-projects/spring-framework/wiki/Code-Style)

ivancea (commented):
What about a generic convertCollection method? Maybe with a Supplier<Collection<>> parameter. convertList() could then use it. That way, we allow users to convert any implementation of Collection, even mutable ones.

Also, should it be overloaded with a TypeDescriptor parameter?

spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
firatkucuk: Gonna change. That makes sense.

spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
firatkucuk: Removed final modifier.

spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
```diff
@@ -166,6 +169,44 @@ public boolean canBypassConvert(@Nullable TypeDescriptor sourceType, TypeDescrip
...
+		return (List<T>) convertCollection(sourceList, TypeDescriptor.valueOf(targetClass), ArrayList::new);
+	}
+
+	@SuppressWarnings("unchecked")
+	@Nullable
+	public <S, T> List<T> convertList(@Nullable List<S> sourceList, TypeDescriptor targetType) {
+		return (List<T>) convertCollection(sourceList, targetType, ArrayList::new);
+	}
+
+	@SuppressWarnings("unchecked")
```
ivancea: Also, after the "C extends Collection<T>", check if the  @SupressWarnings("unchecked") are still required

ivancea (approved):

snicoll:
Thanks for the PR but we've decided not to pursue with this change, see https://github.com/spring-projects/spring-framework/issues/23254#issuecomment-970273051

PR #23265, opened 2019-07-10
Polish DisposableBeanAdapter

izeye:
This PR polishes `DisposableBeanAdapter`.


mentallurg (approved):

snicoll:
Thanks for the PR @izeye but I feel that `warnWithConditionalStackTrace` is a bit confusing and the copy/paste of the logic doesn't bother me as much as it is more straightforward when browsing the code currently.

PR #23267, opened 2019-07-10
Avoid unnecessary boxing where primitives can be used

stsypanov:
We don't need to box primitives with `valueOf` when `parseXxx` can be used.


sbrannen:

Thanks

PR #23290, opened 2019-07-14
Fix the bug for method mergeProperties().

hanbernate:
Using CollectionFactory.createStringAdaptingProperties() instead of new Properties().Using method getProperty() will get null after mergeProperties() when the value is non-string type, .

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/support/PropertiesLoaderSupport.java
+++ b/spring-core/src/main/java/org/springframework/core/io/support/PropertiesLoaderSupport.java
@@ -25,6 +25,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.core.io.Resource;
+import org.springframework.core.CollectionFactory;
 import org.springframework.lang.Nullable;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.DefaultPropertiesPersister;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@LiuMenghan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23290)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23290) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@LiuMenghan Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23290)!

sbrannen (changes requested):
Can you please provide some background on your use case?

Also, can you please introduce tests that failed before the change and pass after the change?

LiuMenghan:
> Can you please provide some background on your use case?
> 
> Also, can you please introduce tests that failed before the change and pass after the change?

I use YamlPropertiesFactoryBean and PropertyPlaceholderConfigurer together when trying to upgrade spring and add yaml support in an old project. Some value with the type of integer in yaml file cannot be injected correctly. One alternative way to solve my problem is to use PropertySourcesPlaceholderConfigurer instead. But I still think it is a bug if someone forgetting not to use PropertyPlaceholderConfigurer like me.

snicoll:
Sorry but I don't understand the problem you're describing and, without a failing test case that shows what it is trying to solve, we can't accept the change. I am going to close this now but we can reconsider if you provide the additional context.

PR #23300, opened 2019-07-16

stsypanov:
I think this is the last change of this kind in Spring's code base

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/style/DefaultValueStyler.java
+++ b/spring-core/src/main/java/org/springframework/core/style/DefaultValueStyler.java
@@ -18,10 +18,10 @@ package org.springframework.core.style;
 
 import java.lang.reflect.Method;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringJoiner;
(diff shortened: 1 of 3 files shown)


sbrannen (changes requested):
Before we introduce these changes, we really should have some regression tests in place for the refactored functionality.

Can you please introduce unit tests for `DefaultValueStyler` that pass against the status quo and after the changes?

stsypanov:
@sbrannen sure, I'll do it in a couple of days.

stsypanov:
@sbrannen I've added the test

sbrannen (approved):

sbrannen:
> @sbrannen I've added the test

Thanks! It's great to have those in place.

A few tips...

- test classes must end with `*Tests`
- test classes need copyright headers as well
- updated classes need to have their copyright dates updated
- import order is important
- execute `./gradlew check` to see Checkstyle violations, etc.

I've already made these changes locally and will merge soon.

sbrannen:
This has been merged into `master` in 1d6e44ab04cc007e34c09191cd77517795cbd8ca and refined in a3c7ae2c58dc932be4f576ebbad007442aff2cff.

Thanks

Issue #23307, opened 2019-07-17
@ConditionalOn behaves unintuitively for @Bean methods with same name (depends on method declaration order)


tobisinghania:
**Affects:** 4.3.x - 5.1.x

Probably affects older versions as well.

---

When `@ConditionalOn..` annotations are used on `@Bean` methods with the same name  within one configuration class, the outcome depends on the declaration order of the methods.

Consider the simple example, in which application startup fails, because the bean name 'conditionalBean' becomes blacklisted on the upper definition, thus the second one is ignored.

```java
    @SpringBootApplication
    @Configuration
    public class DemoApplication {

	public static void main(String[] args) {
		SpringApplication.run(DemoApplication.class, args);
	}

	@Bean
	@ConditionalOnExpression("false")
	public ConditionalBean conditionalBean(ApplicationContext ctx) {
		return new ConditionalBean();
	}

	@Bean
	public ConditionalBean conditionalBean() {
		return new ConditionalBean();
	}

	@Bean
	public SomeOtherBean someOtherBean(ConditionalBean conditionalBean) {
		return new SomeOtherBean();
	}

	public class ConditionalBean {}
	public class SomeOtherBean {}

    }
```

In this example the application starts up properly:

```java
	@Bean
	public ConditionalBean conditionalBean() {
		return new ConditionalBean();
	}

	@Bean
	@ConditionalOnExpression("false")
	public ConditionalBean conditionalBean(ApplicationContext ctx) {
		return new ConditionalBean();
	}
```

The reason for this behavior lies in  `ConfigurationClassBeanDefinitionReader::loadBeanDefinitionsForBeanMethod`, where the method name is internally stored as skipped: 

```java
    // Do we need to mark the bean as skipped by its condition?
	if (this.conditionEvaluator.shouldSkip(metadata, ConfigurationPhase.REGISTER_BEAN)) {
		configClass.skippedBeanMethods.add(methodName);
		return;
	}
	if (configClass.skippedBeanMethods.contains(methodName)) {
		return;
	}
```

A sample project to reproduce this error can be found here: https://github.com/tobisinghania/spring-conditionaOn-order

redhead:
Hi. Just ran into this and was quite surprised by this behavior. 

[...]

PR #23315, opened 2019-07-18
Prevent @Conditional from ignoring subsequent bean definition with same name

tobisinghania:
See #23307 for reason of this PR.

I removed the check, whether a bean is already in the skippedBeanMethods collection. 
This collection only stores the method name, which leads to the undesired behavior as mentioned in #23307.

After this change, bean methods are only skipped, if they are annotated with a condition which evaluates to false.


snicoll:
Thanks for the PR but I don't think that we should be merging this. This would ignore the original problem that the same bean definition is registered twice. See https://github.com/spring-projects/spring-framework/issues/23307#issuecomment-1257761033

Issue #23056, opened 2019-05-29
Provide functionality similar to Spring Boot's BeanTypeRegistry


philwebb:
Spring Boot currently has an internal class called [`BeanTypeRegistry`](https://github.com/spring-projects/spring-boot/blob/e481eccd577fd7dcee9c02c20fd10c60b762bf7d/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/BeanTypeRegistry.java#L68) that is used by `@ConditionalOnBean` and `@ConditionalOnMissingBean`. We've have a [recent issue](https://github.com/spring-projects/spring-boot/pull/16615#discussion_r278441383) where we need something similar, but rather than make it public in Spring Boot it might be better if Spring Framework could offer methods that we could use.

What we specifically need is a way to predict the bean type in a way that's very fast to call during condition evaluation and doesn't cause any early initialization of any beans (including `FactoryBeans`).

---

After some analysis, I'm going to split this issue up into a number of smaller parts:
- [x] Add missing ResolvableType variants of getBeanNamesForType (#23335)
- [x] Retain merged bean definition caches during post processing (#23336)
- [x]  Cache isFactoryBean detection (#23337)
- [x] Consider generics when predicting FactoryBean types (#23338)
- [x] Add attribute support for FactoryBean bean definitions (#23339)
- [x] Use static final constants for common zero length arrays (#23340)

tkvangorder:
@philwebb Any movement on this issue? I could take a stab at PR for this, but I would need some guidance on which package to move the code and/or what to rename the `BeanTypeRegistry` such that it conforms to how things are organized in the core Spring framework.

philwebb:
Thanks for the offer @tkvangorder but I'm already in the middle of trying various changes. We won't be able to migrate `BeanTypeRegistry` as it is, instead we need to look at changing some of the core `BeanFactory` implementations. Unfortunately that's quite a complex part of the codebase so it's taking some time.

tkvangorder:
yep, not for the faint of heart. 8) Make sense, thank you!

PR #23341, opened 2019-07-23
Provide functionality similar to Spring Boot's BeanTypeRegistry

philwebb:
Initial attempt at fixing the issues raised from #23056. Since these changes are quite core, careful review is required.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/BeanFactoryUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/BeanFactoryUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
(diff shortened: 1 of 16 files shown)


tkvangorder:
@philwebb Your changes look good. In context to https://github.com/spring-projects/spring-boot/pull/16615, I have one (probably silly) question: Is there a way to ask the BeanFactory for a Bean Definition's type without forcing the initialization of that bean? Essentially a `BeanFactory.getType(String name, boolean allowEagerInit)`. The code in that PR iterates over all the bean definitions, gets the type for each definition and then consults the predicates: https://github.com/spring-projects/spring-boot/pull/16615/files#diff-66d2d216c30cb3f8db7e7e3f09a66249R61-R65 I am likely missing something obvious. thanks!

philwebb:
@tkvangorder I'm not sure there is such an API yet. I've raised #23374 to look into it after this one is merged.

PR #23344, opened 2019-07-23
Support arrays of `AnnotatedElement` with `MergedAnnotations.from`

### LINKED ISSUE #23327: Allow to create a MergedAnnotations instance composed of multiple MergedAnnotations instance
I find myself having to lookup annotations from an ordered set of different locations and use the information from the first location a particular annotation is present. The particular use case is to find some object mapping annotations on some property accessor methods first but falling back to the field backing a property.

I currently keep [a `List<MergedAnnotations>` with the accessor method derived instances first, the one created from the field last](https://github.com/spring-projects/spring-hateoas/blob/a71aeeb4d12650ee82302b01cde54b891a44ef6b/src/main/java/org/springframework/hateoas/mediatype/PropertyUtils.java#L301-L304) and stream over them, check for whether the annotation is present, if so use that and continue looking for it otherwise (see [this](https://github.com/spring-projects/spring-hateoas/blob/a71aeeb4d12650ee82302b01cde54b891a44ef6b/src/main/java/org/springframework/hateoas/mediatype/PropertyUtils.java#L371-L378)). It would be cool if that external iteration could be hidden behind a `MergedAnnotations` instance in the first place.

/cc @philwebb

philwebb:
Provide an overloaded version of `MergedAnnotations.from` that takes
an array of elements. This shortcut is particularly useful in property
situations where an annotation may be used on a getter, setter or field.

Closes gh-23327

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
@@ -78,6 +78,30 @@ abstract class AnnotationsScanner {
 	}
 
 	/**
+	 * Scan the the specified elements using {@link SearchStrategy#DIRECT} for
+	 * relevant annotations and call the processor as required.
+	 * @param context an optional context object that will be passed back to the
+	 * processor
+	 * @param sources the element sources to scan
+	 * @param processor the processor that receives the annotations
+	 * @return the result of {@link AnnotationsProcessor#finish(Object)}
+	 */
+	@Nullable
+	static <C, R> R scan(C context, AnnotatedElement[] sources,
+			AnnotationsProcessor<C, R> processor) {
+
+		for (AnnotatedElement source : sources) {
+			if (source != null) {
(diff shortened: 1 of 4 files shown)


philwebb:
I'm happy to just merge this one but I'd quite like another pair of eyes on it first. The one thing I'm not too sure about is having both a `from(@Nullable AnnotatedElement element)` and and overloaded `from(AnnotatedElement... elements)`. There's some precedent in `LogMessage` for that pattern.

sbrannen:
@philwebb and @odrotbohm, is this feature request still relevant?

odrotbohm:
It is, yes. Not very urgent, but the sample code linked to from my original post still looks the same and would benefit from fix here.

snicoll:
@sbrannen would it be possible to get this triaged?

sbrannen:
**Team Decision**: we have decided to decline this PR on the basis that we do not want to introduce additional complexity in the `MergedAnnotations` infrastructure to support this feature.

PR #23353, opened 2019-07-25
Add Javadoc since for Jaxb2XmlDecoder(MimeType...)

izeye:
This PR adds Javadoc `@since` tag for the `Jaxb2XmlDecoder(MimeType...)` constructor.


snicoll:
Good catch, thanks again @izeye!

Issue #20759, opened 2017-11-18
Expose parameter names from ParsedSql [SPR-16211]


spring-projects-issues:
**[Yanming Zhou](https://jira.spring.io/secure/ViewProfile.jspa?name=quaff)** opened **[SPR-16211](https://jira.spring.io/browse/SPR-16211?redirect=false)** and commented

I'm developing a query console using NamedParameterJdbcTemplate, I need get parameter names from sql to construct UI form let user input parameters

```java
ParsedSql parsedSql = NamedParameterUtils.parseSqlStatement(sql);
List<String> params = parsedSql.getParameterNames(); //NOT public
```

It's better if NamedParameterJdbcTemplate could provide a method to get parameter name and parameter type, using PreparedStatement.getParameterMetaData().getParameterType()

---

**Reference URL:** https://github.com/spring-projects/spring-framework/pull/1597

**Issue Links:**
- #15442 make org.springframework.jdbc.core.namedparam.ParsedSql public accessible  (_**"is duplicated by"**_)
- #20887 NamedParameterJdbcTemplate fails on PostgreSQL array
- #20908 SqlParameterSource needs method for extracting parameter names

rengy-github:
Namedparameterutils is a tool class. I also want to use this class instead of copying one

PR #23371, opened 2019-07-26
Make getters public in NamedParameterJdbcTemplate and ParsedSql

garydgregory:
Make getters public in two public classes.


sbrannen:
Can you please provide the rationale and a concrete use case for the proposed changes?

garydgregory:
Hi @sbrannen ,

I work on a proprietary product that uses Spring Batch as an ETL component in a suite. This ETL component has a custom copy of `org.springframework.batch.item.database.JdbcBatchItemWriter` as well as a subclass. This does all sorts of magic specific to our product. This code needs access to various bits and pieces of the Spring object tree which is currently achieved via reflection (nasty) which would be better served via public and protected APIs (nice.) I hope this helps clarify things.

Thank you,
Gary

quaff:
+1 https://github.com/spring-projects/spring-framework/issues/20759

snicoll:
Thanks for the PR but we're not keen to expose APIs like that, see https://github.com/spring-projects/spring-framework/pull/23766#issuecomment-1257728214

Issue #13568, opened 2011-12-14
PropertySourcesPlaceholderConfigurer never calls "convertProperties" on merged properties [SPR-8928]


spring-issuemaster:
**[Daniel Fernández](https://jira.spring.io/secure/ViewProfile.jspa?name=dfernandez)** opened **[SPR-8928](https://jira.spring.io/browse/SPR-8928?redirect=false)** and commented

org.springframework.context.support.PropertySourcesPlaceholderConfigurer is overriding the "postProcessBeanFactory" method defined in org.springframework.beans.factory.config.PropertyResourceConfigurer in order to register the required chain of PropertySource objects that will be use for resolving properties. This makes sense and is OK.

But the problem is that the PropertySource object for the local properties (those coming from the "location" attribute and also from the "properties" property) is being created like this:

PropertySource<?> localPropertySource =
new PropertiesPropertySource(LOCAL_PROPERTIES_PROPERTY_SOURCE_NAME, this.mergeProperties());

...whereas in the original "postProcessBeanFactory" method in PropertyResourceConfigurer these merged properties are post-processed by executing the "convertProperties" method:

Properties mergedProps = mergeProperties();
// Convert the merged properties, if necessary.
convertProperties(mergedProps);
// Let the subclass process the properties.
processProperties(beanFactory, mergedProps);

This means that the new PropertySourcesPlaceholderConfigurer class never calls "convertProperties", and so disables one of the extension mechanisms of the old pre-3.1 PropertyPlaceholderConfigurer class: overriding the "convertProperty(...)" and "convertPropertyValue(...)" methods.

I am jasypt's author http://www.jasypt.org and I am creating a Spring3.1-compatible EncryptedPropertySourcesPlaceholderConfigurer, but this bug doesn't allow me to transparently apply property decryption at the "convertProperty" method :-(

---

**Affects:** 3.1 GA

**Issue Links:**
- #18574 convertPropertyValue (for reading encrypted values) not working (_**"is duplicated by"**_)
- #13603 Allow the use of custom PropertySource annotations in @Configuration classes
- [SEC-3123](https://jira.spring.io/browse/SEC-3123) Encrypted property value support
- #15294 Add encryption support for PropertyPlaceholderConfigurer
- #17236 Backport encrypted property functionality from spring-cloud-config environment work

8 votes, 18 watchers

spring-issuemaster:
Bulk closing outdated, unresolved issues. Please, reopen if still relevant.

[...]

PR #23377, opened 2019-07-27
Issue-13568: Make PropertySourcesPlaceholderConfigurer call a convert…

srdo:
…perty method when resolving properties

See https://github.com/spring-projects/spring-framework/issues/13568.

The idea here is to provide a new `convertProperty` method for people to implement as a replacement for the ones in `PropertyResourceConfigurer`. The methods in that class are a poor fit for `PropertySourcesPlaceholderConfigurer`, as they only support String-type values.

This PR implements and deprecates the methods that should not be used, and provides a new `convertProperty` that the configurer will actually call.

I think a cleaner way to do this would be to make `PropertySourcesPlaceholderConfigurer` not subclass `PropertyResourceConfigurer`, but I did it this way to avoid breaking the API.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java
+++ b/spring-context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java
@@ -17,7 +17,9 @@
 package org.springframework.context.support;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


srdo:
Is there anything I should do to get this reviewed? It would be nice to get this fixed, as it's blocking Apache Camel from upgrading their placeholder resolution code to the `PropertySourcesPlaceholderConfigurer`.

srdo:
Closing, out of date

PR #23379, opened 2019-07-29
Introduce Converter.andThen(...)

### LINKED ISSUE #22381: Consider adding support for chaining converters
It would be nice to have some support for chaining converters:

```java
public interface Converter<S, T> {
	// ...

	default <U> Converter<S, U> andThen(Converter<T, U> andThen) {
		return source -> andThen.convert(this.convert(source));
	}
}
```

Then, a user could more easily compose them:

```java
Converter<A, B> aToB = ...;
Converter<B, C> bToC = ...;
Converter<A, C> aToC = aToB.andThen(bToC);
```

This is a modest improvement on:

```java
Converter<A, C> aToC = source -> bToC.convert(aToB.convert(source));
```

in that code using `andThen` will read in the correct order, e.g. "do A then B" instead of "B.do(A.do)".

I'd be happy to submit a PR, if it's agreed that it's a reasonable improvement.

jzheaux:
Fixes: gh-22381

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/converter/Converter.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/converter/Converter.java
@@ -17,6 +17,7 @@
 package org.springframework.core.convert.converter;
 
 import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
 
 /**
  * A converter converts a source object of type {@code S} to a target of type {@code T}.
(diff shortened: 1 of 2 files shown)


sbrannen:
This has been merged into `master` in a0c00362c3ba279494fec86ed6e403d829e10bee and revised in a6bede22c113e69206c13735dbf820c7cb043e2e.

Thanks

PR #23380, opened 2019-07-29
@ExceptionHandler to keep nesting beyond the first cause

Syquel:
spring-web exception handling considers only the first cause of an exception to determine the applicable `@ExceptionHandler`.
This patch makes the resolver search through the full exception hierarchy.


rstoyanchev:
Tentatively scheduling for 5.3 M1.

rstoyanchev:
**Team Decision:** The original support for only the first cause was to keep the resolution simpler as well out of caution (it was introduced in 4.3). Trying to resolve causes one level at a time is still relatively simple, e.g. vs checking the entire chain, and with lots of remote calls from a controller this, nesting more than one level may be outside an application's control. We'll go ahead and do this in 5.3 M1.

rstoyanchev:
This has been merged.

PR #23393, opened 2019-07-31
Use Arrays.copyOf and Arrays.copyOfRange where possible

stsypanov:


mentallurg (approved):

sbrannen:
@stsypanov, can you please provide the rationale for the proposed changes?

mentallurg:
> @stsypanov, can you please provide the rationale for the proposed changes?

@sbrannen: The suggested changed makes the code simpler: Now are only 2 parameters required instead of 5. Now there is 1 line of code instead of 2. Easier to understand. Less error prone.

sbrannen:
> > @stsypanov, can you please provide the rationale for the proposed changes?
> 
> @sbrannen: The suggested change makes the code simpler: Now are only 2 parameters used instead of 5. Now there is 1 line of code instead of 2. Easier to understand. Less error prone.

Yes, I am aware of the improvements in readability.

But since @stsypanov often suggests improvements with an eye on performance, I was actually asking him whether he proposed the changes due to performance reasons or merely for the sake of readability.

In addition, we typically ask for the rationale for proposed changes unless it's something as simple as a typo.

@stsypanov ?

stsypanov:
Hello @sbrannen this PR is about readability and making code less error-prone. Also in some cases, e.g. in method `subarray` of class `Tokenizer` reducing code to one line makes it possible to inline the method.

As of performance both code snippets have same performance, however I think `Arrays.copyOf` might be intrinsified in future.

sbrannen:
Thanks for the feedback, @stsypanov. Much appreciated.

After investigating a bit further, it appears to me that there are several more places in the code base that could benefit from a switch to `Arrays.copyOf`.

Is there a reason you limited this change to the particular classes in this PR?

sbrannen:
> Is there a reason you limited this change to the particular classes in this PR?

In any case, I've gone ahead and merged this into `master` in 78d56dc61b05a1ee0b4674ba3408476a30415c80.

Depending on your answer to the above question, feel free to submit an additional PR with similar changes to other classes in the code base (or let us know that we should do it if desired).

Thanks!

PR #23408, opened 2019-08-02
Add a warning to AnnotationReadingVisitorUtils

### LINKED ISSUE #23406: AnnotationReadingVisitorUtils does not log a warning when it encounters an exception
**Affects:** 5.1.8

The following section causes the problem:
`
catch (Throwable ex) {
				// Class not found - can't resolve class reference in annotation attribute.
				result.put(entry.getKey(), ex);
			}
`
Catching any Throwable seems to be excessive.

marekkrk:
Closes gh-23406

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/type/classreading/AnnotationReadingVisitorUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/type/classreading/AnnotationReadingVisitorUtils.java
@@ -23,6 +23,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.asm.Type;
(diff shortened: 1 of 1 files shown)


sbrannen:
Catching `Throwable` is necessary since such code might throw a `LinkageError`.

As for logging the exception, that is effectively unnecessary since any attempt to access the value for that attribute later will result in an exception. For example, see `AnnotationAttributes#assertNotException(String, Object)`.

sbrannen:
Having said that, now that I've looked at `AnnotationReadingVisitorUtils` and `AnnotationAttributes#assertNotException(String, Object)` again, I noticed that the latter should actually check `if (attributeValue instanceof Throwable)` instead of `if (attributeValue instanceof Exception)`.

@jhoeller, what do you think about making that change in `AnnotationAttributes`?

Reopening in order to make a decision regarding this.

sbrannen:
Superseded by gh-23424

PR #23412, opened 2019-08-03
【infra 38】

xjhxjhxjhxjh:
When using org.springframework.util.CollectionUtils#hasUniqueObject
such as 

public void test() {
    List<Integer> list = new ArrayList<>();
    list.add(1);
    list.add(2);
    System.out.println(CollectionUtils.hasUniqueObject(list));
}
From the function name, the result should be true,because there are different objects, but it is actually false.

another example

public void test() {
    List<Integer> list = new ArrayList<>();
    list.add(1);
    list.add(1);
    System.out.println(CollectionUtils.hasUniqueObject(list));
}
From the function name, the result should be false,because there are same*objects, but it is actually true.

I believe the author may reversed the return.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/CollectionUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/CollectionUtils.java
@@ -282,10 +282,10 @@ public abstract class CollectionUtils {
 				candidate = elem;
 			}
 			else if (candidate != elem) {
-				return false;
+				return true;
 			}
 		}
-		return true;
+		return false;
 	}
 
 	/**
--- a/spring-core/src/test/java/org/springframework/util/CollectionUtilsTests.java
+++ b/spring-core/src/test/java/org/springframework/util/CollectionUtilsTests.java
@@ -210,6 +210,18 @@ public class CollectionUtilsTests {
 		assertThat(CollectionUtils.hasUniqueObject(list)).isFalse();
 	}
 
+	@Test
+    public void testHasUniqueObject2() {
+        List<Integer> list1 = new ArrayList<>();
+        list1.add(1);
+        list1.add(1);
+        System.out.println(CollectionUtils.hasUniqueObject(list1));
+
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@xjhxjhxjhxjh Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23412)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23412) to manually synchronize the status of this Pull Request.

xjhxjhxjhxjh:
When using org.springframework.util.CollectionUtils#hasUniqueObject
such as

public void test() {
List list = new ArrayList<>();
list.add(1);
list.add(2);
System.out.println(CollectionUtils.hasUniqueObject(list));
}
From the function name, the result should be true,because there are different objects, but it is actually false.

another example

public void test() {
List list = new ArrayList<>();
list.add(1);
list.add(1);
System.out.println(CollectionUtils.hasUniqueObject(list));
}
From the function name, the result should be false,because there are same*objects, but it is actually true.

I believe the author may reversed the return.

pivotal-issuemaster:
@xjhxjhxjhxjh Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23412)!

mdeinum:
If you check the java doc it works as designed

> Determine whether the given Collection only contains a single unique object.

Your first sample contains 2 and not a single unique element, whereas the second contains a single unique element. Hence it works as designed.

xjhxjhxjhxjh:
By careful inspection, I found that you are right, this is my negligence.

snicoll:
Thanks both for the feedback, closing as it is working as designed.

xjhxjhxjhxjh:
By careful inspection, I found that you are right, this is my negligence.

PR #23414, opened 2019-08-03
Add queryParam method to MockHttpServletRequestBuilder

### LINKED ISSUE #23296: Add queryParam method to MockHttpServletRequestBuilder
`ServletUriComponentsBuilder` uses `HttpServletRequest#getQueryString` to build on the current request.
However, in a GET or HEAD request, `MockHttpServletRequestBuilder#param` only adds to the combined parameter map and not the query string, causing it to fail.

The workaround is to provide the parameters directly in the URI string, but this is inferior to the fluent API. The different behaviour is probably surprising to many users.

See also https://github.com/spring-projects/spring-restdocs/issues/26

jo2:
fix #23296

add query params to query string in builder method in MockHttpServletRequestBuilder

add tests

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
@@ -337,6 +337,12 @@ public class MockHttpServletRequestBuilder
 		return this;
 	}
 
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Thanks for the PR but it does not match what was agreed upon in the related ticket:

> we can add a queryParam method as an extra convenience

Also we expect commits to be organized into logical commits, and squashed if necessary. The idea is that we don't want extra commits that reflect various drafts.

jo2:
new pr #23980

PR #23415, opened 2019-08-04
A dead loop found in SerializableTypeWrapper.java

enkao:
a dead loop found

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/SerializableTypeWrapper.java
+++ b/spring-core/src/main/java/org/springframework/core/SerializableTypeWrapper.java
@@ -91,7 +91,7 @@ final class SerializableTypeWrapper {
 	public static <T extends Type> T unwrap(T type) {
 		Type unwrapped = type;
 		while (unwrapped instanceof SerializableTypeProxy) {
-			unwrapped = ((SerializableTypeProxy) type).getTypeProvider().getType();
+			unwrapped = ((SerializableTypeProxy) unwrapped).getTypeProvider().getType();
 		}
 		return (unwrapped != null ? (T) unwrapped : type);
 	}


pivotal-issuemaster:
@enkao Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23415)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23415) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@enkao Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23415)!

yanglin1501804006:
> 找到一个死循环
HEllo:
I want to ask a question：

in this class :SerializableTypeWrapper.java  

public static <T extends Type> T unwrap(T type) {
		Type unwrapped = type;
		while (unwrapped instanceof SerializableTypeProxy) {
			unwrapped = ((SerializableTypeProxy) type).getTypeProvider().getType();
		}
		return (unwrapped != null ? (T) unwrapped : type);
	}

this code ：Under what circumstances will cause an dead loop ,Tell me about it ，thanks. 
 Syntactically the same

sbrannen:
> Under what circumstances will cause an dead loop?

As written, the loop is unnecessary and could be replaced with an `if` statement.

The reason is that `unwrapped` is not updated within the loop based on the current nested `unwrapped` value but rather on the `type` (which never changes). The unwrapping is therefore effectively limited to a single level of proxies.

It would appear that this is not a problem for use cases relying on the `unwrap()` method. In other words, it does not appear that there are ever any nested proxies in use.

@philwebb and @jhoeller, can you shed some light on why this method was written with the intention of unwrapping arbitrary levels of proxies?

sbrannen:
Related commit: https://github.com/spring-projects/spring-framework/commit/59604b1cd5a392501aa9c43fdb63b7bcfae617ce

philwebb:
I think it was just a defensive measure on my part. Looking at `SerializableTypeWrapper.forTypeProvider` there's already an `instanceof Serializable` check so I don't think it's possible to double wrap.

sbrannen:
Thanks for the feedback, Phil.

> ... so I don't think it's possible to double wrap.

That's what I assumed as well after a cursory look at the code in that class.

So I'll just switch to an `if` statement for Spring Framework 5.2 and leave the code "as is" in previous branches, since the single-pass loop doesn't actually do any harm.

sbrannen:
Thanks for the PR and for bringing this to our attention, @enkao!

I've taken a slightly different approach in a43ba052e94eb3b07ec9e407500e713d24e02b16.

PR #23417, opened 2019-08-05
Preserve the originalMessage when creating an ErrorMessage in MessageBuilder

L00kian:
Hello,

I was creating a small POC using spring integration and for some cases I noticed that Gateways that use messagingTemplate underneath do not preserve the original message when sending a related ErrorMessage to the error channel.

I didn't find any better place to address that. 

In my opinion original message worth preserving in any case. However there is another possibility to pass the MessageBuilder.originalMessage link without unwrapping it

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/support/MessageBuilder.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/support/MessageBuilder.java
@@ -99,6 +99,7 @@ public final class MessageBuilder<T> {
 		this.headerAccessor.removeHeaders(headerPatterns);
 		return this;
 	}
+
 	/**
 	 * Remove the value for the given header name.
 	 */
@@ -153,7 +154,11 @@ public final class MessageBuilder<T> {
 		}
 		MessageHeaders headersToUse = this.headerAccessor.toMessageHeaders();
 		if (this.payload instanceof Throwable) {
-			return (Message<T>) new ErrorMessage((Throwable) this.payload, headersToUse);
+			Message<?> originalMessage = null;
+			if (this.originalMessage != null && this.originalMessage instanceof ErrorMessage) {
+				originalMessage = ((ErrorMessage) this.originalMessage).getOriginalMessage();
+			}
+			return (Message<T>) new ErrorMessage((Throwable) this.payload, headersToUse, originalMessage);
 		}
 		else {
 			return new GenericMessage<>(this.payload, headersToUse);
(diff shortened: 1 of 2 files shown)


L00kian:
Also if the aforementioned is the case - it might be worthwhile to apply the same flow for org.springframework.integration.support.MessageBuilder#build implementation from integration-core package

spring-messaging/src/main/java/org/springframework/messaging/support/MessageBuilder.java
```diff
@@ -153,7 +154,11 @@ private MessageBuilder(T payload, MessageHeaderAccessor accessor) {
 		}
 		MessageHeaders headersToUse = this.headerAccessor.toMessageHeaders();
 		if (this.payload instanceof Throwable) {
-			return (Message<T>) new ErrorMessage((Throwable) this.payload, headersToUse);
+			Message<?> originalMessage = null;
+			if (this.originalMessage != null && this.originalMessage instanceof ErrorMessage) {
+				originalMessage = ((ErrorMessage) this.originalMessage).getOriginalMessage();
+			}
+			return (Message<T>) new ErrorMessage((Throwable) this.payload, headersToUse, originalMessage);
```
rstoyanchev: Shouldn't we simply pass the current "originalMessage", whatever it may be (error or not)? In other words:
```java
@SuppressWarnings("unchecked")
public Message<T> build() {
	if (this.originalMessage != null && !this.headerAccessor.isModified()) {
		return this.originalMessage;
	}
	MessageHeaders headersToUse = this.headerAccessor.toMessageHeaders();
	if (this.payload instanceof Throwable) {
		return (this.originalMessage != null ?
				(Message<T>) new ErrorMessage((Throwable) this.payload, headersToUse, this.originalMessage) :
				(Message<T>) new ErrorMessage((Throwable) this.payload, headersToUse));
	}
	else {
		return new GenericMessage<>(this.payload, headersToUse);
	}
}
```

/cc @artembilan @garyrussell

garyrussell: There is some confusion here `ErrorMessage.originalMessage` is not the same as `MessageBuilder.originalMessage`. The fields, unfortunately, have the same name.

Consider a new message entering at some source endpoint (a gateway in this case). The message goes through several transformations and an exception is thrown. The `MessagingException` has a `failedMessage` field, containing the message at the point of the failure.

We added the `originalMessage` field to `ErrorMessage` so we can capture the "original" message at the point where the exception is caught (in this case an inbound gateway), so the error handling flow can have access to the state before any transformation was done (e.g. to post it to a dead letter repository).

I don't think it's a common use case to build an `ErrorMessage` from another but since the existing code supports that, it seems reasonable that we should propagate the `originalmessage.originalMessage` field.

rstoyanchev: Thanks for the clarification. I figured as much that they were different but wasn't sure if a chain of messages would be preferable or not.

rstoyanchev:
This has been merged with some minor polishing.

PR #23418, opened 2019-08-05
Resolve generic type in more complex scenarios

Syquel:
The spring-core GenericTypeResolver does not resolve generic type variables correctly in more complex scenarios like renamed transitive type variables and partial specialization of classes.


snicoll:
@Syquel thank you for making your first contribution to Spring Framework.

Issue #23404, opened 2019-08-02
Add a new RequestEntity$HeadersBuilder#headers(HttpHeaders) method


grimsa:
**Affects:** 5.1.9 and previous versions (spring-web)

---

`RequestEntity` builders are very handy, however adding multiple headers can currently only be done like this:

```java
        return RequestEntity.get(url)
            .header("first", "value")
            .header("second", "other value")
            .build();
```

This is cumbersome when a common set of headers needs to be added to various requests.

Adding a `B headers(HttpHeaders headers)` method to `RequestEntity$HeadersBuilder` would solve it by making this possible:

```java
        return RequestEntity.get(url)
            .headers(new CommonHttpHeaders())        // implements HttpHeaders
            .build();
```

The same method was added to `ResponseEntity$HeadersBuilder` with #16929, so this would also improve consistency between request and response builders.

rstoyanchev:
In other places we have `headers(Consumer<HttpHeaders>)`. For consistency I would add since it is useful in any case to be able to modify the underlying headers. In your case it would look:

```java
return RequestEntity.get(url)
        .headers(headers -> headers.putAll(new CommonHttpHeaders()))
        .build();
```

poutsma:
I added `headers(HttpHeaders)` to `RequestEntity` and `headers(Consumer<HttpHeaders>)` to both `RequestEntity` and `ResponseEntity`.

PR #23422, opened 2019-08-06
Add a new RequestEntity$HeadersBuilder method

andluu:
New `headers(MultiValueMap headers)` in RequestEntity$HeadersBuilder.
Just like in ResponseEntity$HeadersBuilder.

In addition, made wider argument type: `headers(MultiValueMap headers)`
instead of `headers(HttpHeaders headers)` on both Builders, for better polymorphism.

Issue: #23404


poutsma:
Thank you for leaving a PR for #23404. In the end, I decided to take a different route and go with @rstoyanchev suggestion to also introduce a `Consumer`-based version, which allows for even greater flexibility and does not require to change method signatures.

PR #23439, opened 2019-08-09
Fix */* in Javadoc

GreenRecycleBin:
> {@code}
> 
>     {@code text }
> 
> Equivalent to <code>{@literal text }</code>.
> 
> Displays text in code font without interpreting the text as HTML markup or nested Javadoc tags.

https://docs.oracle.com/en/java/javase/12/docs/specs/doc-comment-spec.html

This is why `{@code &#42;&#47;&#42;}` isn't rendered as <code>&#42;/&#42;</code> as expected.

@pivotal-issuemaster This is an Obvious Fix.


sbrannen:
Good catch.

Thanks

PR #23441, opened 2019-08-09
Do not override BeanPostProcessor default methods unnecessarily

sbrannen:
This commit also makes `InstantiationAwareBeanPostProcessorAdapter` somewhat obsolete.


jhoeller:
As per our discussion, let's keep the redeclared methods as-is for the time being, mostly for efficiency reasons: Default methods are rather specific in their handling within the JVM, and generally speaking redeclared methods within a class hierarchy are more efficient to dispatch to... and properly allow for inlining. A positive inlining effect alone is worth it, in particular for empty methods in multi-callback SPI interfaces such as `BeanPostProcessor`.

sbrannen:
For related discussions, see [this Stack Overflow thread](https://stackoverflow.com/questions/30312096/java-default-methods-is-slower-than-the-same-code-but-in-an-abstract-class
) that Juergen shared with me.

PR #23445, opened 2019-08-09
Simplify the if statements and replace 'try finally' with 'try' with resources

zhuzhuman978:
1.Simplify the if statements, making it easier to read.
2.Replace 'try finally' with 'try' with resources in DataSourceTransactionManager


sbrannen:

Thanks

PR #23447, opened 2019-08-10
Simplify the code

NB216ers:
Unnecessary 'null' check before 'instanceof' expression


sbrannen:
Thanks for the PR.

Since the `Matchers` class no longer exists in `master` and `ConcurrentReferenceHashMap` has already been modified in that regard in `master`, I am closing this PR.

PR #23455, opened 2019-08-13
Avoid storing unnecessary beanName in ApplicationListenerDetector

GungnirLaevatain:
it would store name that implementation of the **ApplicationListener** rather than store all names

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/ApplicationListenerDetector.java
+++ b/spring-context/src/main/java/org/springframework/context/support/ApplicationListenerDetector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,7 +59,9 @@ class ApplicationListenerDetector implements DestructionAwareBeanPostProcessor, 
 
 	@Override
 	public void postProcessMergedBeanDefinition(RootBeanDefinition beanDefinition, Class<?> beanType, String beanName) {
-		this.singletonNames.put(beanName, beanDefinition.isSingleton());
+		if (ApplicationListener.class.isAssignableFrom(beanType)) {
+			this.singletonNames.put(beanName, beanDefinition.isSingleton());
+		}
 	}
 
 	@Override


spring-context/src/main/java/org/springframework/context/support/ApplicationListenerDetector.java
```diff
@@ -59,7 +59,9 @@ public ApplicationListenerDetector(AbstractApplicationContext applicationContext
 
 	@Override
 	public void postProcessMergedBeanDefinition(RootBeanDefinition beanDefinition, Class<?> beanType, String beanName) {
-		this.singletonNames.put(beanName, beanDefinition.isSingleton());
+		if (ApplicationListener.class.isAssignableFrom(beanType)) {
+			this.singletonNames.put(beanName, beanDefinition.isSingleton());
+		}
```
mentallurg: I find this change bad.

Method should either do what the caller expects it to do or should fail. If you silently ignore the passed bean as in your suggested change, then the method doesn't do what is expected, neither throws an exception. This can lead to errors that are very hard to detect.

I would suggest that you consider one of following approaches:

1. If you believe that the passed bean should have not been passed to this method, then throw an exception. But before doing that please analyse thoroughly all usage scenarios and provide corresponding tests.

2. Change method declaration: Instead of `Class<?> beanType` use `Class<? extends ApplicationListener> beanType`. Then you will be sure that only proper beans will be passed to this method. Then `isAssignableFrom()` in this method will not be needed. Then you need to change also all usages of this method. This will be a good indicator if your change is good. Also it can indicate some places where this method was incorrectly used. Thus you will also fix such bugs.

mentallurg (changes requested):

spring-context/src/main/java/org/springframework/context/support/ApplicationListenerDetector.java
GungnirLaevatain: Thank you for your suggestion.
But MergedBeanDefinitionPostProcessor declares this method, and many classes implement this interface, so I think it is not good to change method declaration.
In addition, other method that declared in this class process the bean that implement the ApplicationListener, and ignore other bean, so I think this method can do the same.

sbrannen: > But MergedBeanDefinitionPostProcessor declares this method, and many classes implement this interface, so I think it is not good to change method declaration.

That's correct: `MergedBeanDefinitionPostProcessor` is a public API since Spring Framework 2.5 and cannot be modified.

Issue #23452, opened 2019-08-12
@EnableWebFlux Javadoc references nonexistent WebFluxConfigurer API


niksergey:
There is no configureMessageWriters method in WebFluxConfigurer, 
but javadoc for EnableWebFlux annotation suggests use it.

Sauhardstark:
Hi, I'm relatively new to contributing to open source, though I've been benefitting from spring and other open source libraries for quite some time now. I took a look at the issue and I feel that I would be able to fix said issue. Any help or even slight guidance towards that would be deeply appreciated. I have already forked the repository and would fix the issue by tomorrow.

sbrannen:
Good catch, @niksergey.

I assume the example refers to an earlier (potentially unpublished) version of the API.

@Sauhardstark, if you want to submit a fix for the example in the documentation, you might want to demonstrate how to use `org.springframework.web.reactive.config.WebFluxConfigurer.configureHttpMessageCodecs(ServerCodecConfigurer)`, since that seems in line with the current example. However, you could also demonstrate something simpler than that. The goal of the documentation is not necessarily to demonstrate something involved but rather just to demonstrate that custom configuration is possible via the `WebFluxConfigurer` API.

sbrannen:
Superseded by gh-23457

PR #23457, opened 2019-08-13
Modify the javadoc of @EnableWebFlux annotation

Sauhardstark:
Previously the Javadoc of EnableWebFlux referred to `configureMessageWriters()` method in `WebFluxConfigurer` which wasn't there. Now it has been modified to give an example of the `configureHttpMessageCodecs()` method which is present

This fixes the [issue 23452](https://github.com/spring-projects/spring-framework/issues/23452)


sbrannen:

Thanks for the PR!

PR #23459, opened 2019-08-14
Allow use of DataBufferUtils from ServerHttpResponseDecorator#writeWith

ilgrosso:
`ServerHttpResponseDecorator` accepts `Publisher<? extends DataBuffer>` while `DataBufferUtils` expects `Publisher<DataBuffer>`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
@@ -522,11 +522,11 @@ public abstract class DataBufferUtils {
 	 * @return a buffer that is composed from the {@code dataBuffers} argument
 	 * @since 5.0.3
 	 */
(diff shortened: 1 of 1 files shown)


rstoyanchev:
It looks like `ReactiveHttpOutputMessage` was [updated](https://github.com/spring-projects/spring-framework/commit/8d26c738a016d9aa57fc3b080fe58f0c44fb341d ) at some point before 5.0 to accept `Publisher<? extends DataBuffer>`. I think it would make sense to do the same in `DataBufferUtils` (and not only the `join` method). @poutsma what do you think?

rstoyanchev:
@ilgrosso I added the description a little to make it more clear.

poutsma:
Sure, though I suspect we might need a `@SuppressWarnings("unchecked")` on the method as a consequence.

ilgrosso:
@rstoyanchev how to move fwd?

Issue #23234, opened 2019-07-04
spring-core-5.2.0.M3.pom missing netty dependencies


suztomo:
**Affects:** 5.2.0.M3

Would you declare dependencies in pom.xml when installing spring artifacts to a repository?

---

- [Spring-core 5.2.0.M3's pom.xml](https://repo.spring.io/libs-milestone-local/org/springframework/spring-core/5.2.0.M3/spring-core-5.2.0.M3.pom) in Spring milestone repository does not have Netty dependency.
  ```
  <dependencies>
    <dependency>
      <groupId>org.springframework</groupId>
      <artifactId>spring-jcl</artifactId>
      <version>5.2.0.M3</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
  ```
- [Spring-core 5.1.8-RELEASE's pom.xml](https://repo1.maven.org/maven2/org/springframework/spring-core/5.1.8.RELEASE/spring-core-5.1.8.RELEASE.pom) published in Maven Central has netty dependency with optional: true.
  ```
  ...
  <dependency>
    <groupId>io.netty</groupId>
    <artifactId>netty-buffer</artifactId>
    <version>4.1.36.Final</version>
    <scope>compile</scope>
    <optional>true</optional>
  </dependency>
  ...
  ```

# Context

Although these optional dependencies of a transitive dependency are only used in compilation time of spring-core and not automatically retrieved by users' build tools, declaring optional dependencies in pom.xml is helpful for the following reasons:

## For static analysis tools

I develop [a tool](https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/enforcer-rules) to check missing dependency. The tool reads pom.xml of Maven artifacts. The tool detected spring-core 5.2.0.M3 does not have proper dependencies declared in pom.xml.

## For human

Declaring these dependencies serves as human-readable documentation. For example some of the classes in spring-core use `io.netty.buffer.ByteBufAllocator`. Declaring optional dependency in spring-core's pom.xml helps its users to pick up correct Maven artifact to make the classes work.

-------

With the reasons above, would you declare dependencies in pom.xml when installing to a repository?

[...]

PR #23465, opened 2019-08-15
Replacing netty-all with specific dependencies

suztomo:
As per @bclozel 's comment on the below:

> Spring Framework is sometimes depending on "-all.jar" variants for historical reasons, and doing so in an application or library is not advised. Arguably we should not do that in our build and this is something we should fix. But in the meantime this is incorrect/misleading information we're publishing

https://github.com/spring-projects/spring-framework/issues/23234#issuecomment-521153626

spring-framework/spring-web/src contains references to following netty packages covered by the artifacts:

- artifact netty-buffer
  - io.netty.buffer
- artifact netty-transport
  - io.netty.channel
  - io.netty.bootstrap
  - io.netty.channel.nio
- artifact netty-handler
  - io.netty.handler.ssl
  - io.netty.handler.ssl.util
  - io.netty.handler.timeout
- artifact netty-codec-http
  - io.netty.handler.codec.http


mentallurg (commented):
I support the idea to use more precise libraries. But have you tested it? How about `netty-common`? Anything else?

suztomo:
> But have you tested it?

It compiles successfully. I added the list of packages and artifacts relationship. I didn't find netty-common needed.

rstoyanchev:
This has been rebased and merged 21d390a01860fc89856750dd6c836ea2205bb023.

suztomo:
Thank you.

PR #23469, opened 2019-08-16
Use enum instance instead of instantiating AutowiredArgumentMarker

stsypanov:
Currently AutowiredArgumentMarker is instantiated for each autowired argument in ConstructorResolver. This can be avoided with use of `enum`.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
@@ -771,7 +771,7 @@ class ConstructorResolver {
 							methodParam, beanName, autowiredBeanNames, converter, fallback);
 					args.rawArguments[paramIndex] = autowiredArgument;
 					args.arguments[paramIndex] = autowiredArgument;
-					args.preparedArguments[paramIndex] = new AutowiredArgumentMarker();
+					args.preparedArguments[paramIndex] = AutowiredArgumentMarker.INST;
 					args.resolveNecessary = true;
 				}
 				catch (BeansException ex) {
@@ -809,7 +809,7 @@ class ConstructorResolver {
 		for (int argIndex = 0; argIndex < argsToResolve.length; argIndex++) {
 			Object argValue = argsToResolve[argIndex];
 			MethodParameter methodParam = MethodParameter.forExecutable(executable, argIndex);
-			if (argValue instanceof AutowiredArgumentMarker) {
+			if (argValue == AutowiredArgumentMarker.INST) {
 				argValue = resolveAutowiredArgument(methodParam, beanName, null, converter, fallback);
 			}
 			else if (argValue instanceof BeanMetadataElement) {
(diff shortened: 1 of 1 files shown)


spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
```diff
@@ -809,7 +809,7 @@ private ArgumentsHolder createArgumentArray(
 		for (int argIndex = 0; argIndex < argsToResolve.length; argIndex++) {
 			Object argValue = argsToResolve[argIndex];
 			MethodParameter methodParam = MethodParameter.forExecutable(executable, argIndex);
-			if (argValue instanceof AutowiredArgumentMarker) {
+			if (argValue == AutowiredArgumentMarker.INST) {
```
mentallurg: 1. This may save several CPU steps. But it makes the code less understandable. Many developers will not understand, why some values are compared by value, the other by type.

2. Besides, in the future this may be extended to handle special cases like autowired on setter or autowired on constructor or autowired with parameters like `@Autowired(required = true)`. That's why I would prefer keeping  the type, not the value.

spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
```diff
@@ -966,7 +966,8 @@ public void storeCache(RootBeanDefinition mbd, Executable constructorOrFactoryMe
 	/**
 	 * Marker for autowired arguments in a cached argument array.
  	 */
-	private static class AutowiredArgumentMarker {
+	private enum AutowiredArgumentMarker {
+		INST
```
mentallurg: I'd suggest to use a singleton instance on this place, not enumeration.

spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
```diff
@@ -771,7 +771,7 @@ private ArgumentsHolder createArgumentArray(
 							methodParam, beanName, autowiredBeanNames, converter, fallback);
 					args.rawArguments[paramIndex] = autowiredArgument;
 					args.arguments[paramIndex] = autowiredArgument;
-					args.preparedArguments[paramIndex] = new AutowiredArgumentMarker();
+					args.preparedArguments[paramIndex] = AutowiredArgumentMarker.INST;
```
mentallurg: If you use the singleton here, that would be OK, considering my other comments below.

mentallurg (changes requested):

sbrannen:
Thanks for the PR and for bringing this to our attention.

I have taken a slightly different approach in 4cd635e9c87d58b990a7b19125fdfe53007ee648.

spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
sbrannen: Thanks for the review, @mentallurg.

Regarding point 1, I don't think many developers will actually be reading those lines of code, and those that do can now read the improved Javadoc in 4cd635e9c87d58b990a7b19125fdfe53007ee648.

Regarding point 2, this marker technique has not been changed in over 11 years (potentially longer). So I don't foresee it changing in the near future, but even if it does change, we can revisit this section of code at that time.

PR #23470, opened 2019-08-16
Simplify String concatenation

stsypanov:


spring-core/src/main/java/org/springframework/util/backoff/FixedBackOff.java
```diff
@@ -109,14 +109,12 @@ public long nextBackOff() {
...
 			String attemptValue = (FixedBackOff.this.maxAttempts == Long.MAX_VALUE ?
 					"unlimited" : String.valueOf(FixedBackOff.this.maxAttempts));
-			sb.append(", currentAttempts=").append(this.currentAttempts);
-			sb.append(", maxAttempts=").append(attemptValue);
-			sb.append('}');
-			return sb.toString();
+			return "FixedBackOff{interval=" + FixedBackOff.this.interval +
+					", currentAttempts=" + this.currentAttempts +
+					", maxAttempts=" + attemptValue +
+					'}';
```
mentallurg: The code is already pretty simple. The reason `StringBuilder` is used is performance. Have you checked what byte code is generated for your code? Have you tested performance? On what JREs have you tested it? Without such information I reject your pull request.

mentallurg (changes requested):
The code is already pretty simple. The reason `StringBuilder` is used is performance. Have you checked what byte code is generated for your code? Have you tested performance? On what JREs have you tested it? Without such information I reject your pull request.

spring-core/src/main/java/org/springframework/util/backoff/FixedBackOff.java
stsypanov: > The reason StringBuilder is used is performance.

Before claiming this you should

1) read more about *usage* of StringBuilder, mainly about usage of `append` method, e.g. here http://alblue.bandlem.com/2016/04/jmh-stringbuffer-stringbuilder.html

2) read how javac and JVM handle expressions like `String s = s1 + s3 + s3;`

You'll be surprised.

sbrannen:

Thanks

PR #23478, opened 2019-08-17
Use Collections.addAll where feasible

OLPMO:
Simplifed some if statement and use Collection.addAll instead of iterator to improve the  efficiency

### DIFF
--- a/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
+++ b/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
@@ -382,7 +382,7 @@ public class ReflectUtils {
 			}
 		}
 		methods.remove(null);
-		return (Method[]) methods.toArray(new Method[methods.size()]);
+		return (Method[]) methods.toArray(new Method[0]);
 	}
 
 	public static PropertyDescriptor[] getBeanProperties(Class type) {
@@ -412,7 +412,7 @@ public class ReflectUtils {
 					properties.add(pd);
 				}
 			}
-			return (PropertyDescriptor[]) properties.toArray(new PropertyDescriptor[properties.size()]);
+			return (PropertyDescriptor[]) properties.toArray(new PropertyDescriptor[0]);
 		}
 		catch (IntrospectionException e) {
 			throw new CodeGenerationException(e);
--- a/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
+++ b/spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
@@ -382,7 +382,7 @@ public class ReflectUtils {
 			}
 		}
 		methods.remove(null);
-		return (Method[]) methods.toArray(new Method[0]);
+		return (Method[]) methods.toArray(new Method[methods.size()]);
 	}
 
 	public static PropertyDescriptor[] getBeanProperties(Class type) {
@@ -412,7 +412,7 @@ public class ReflectUtils {
 					properties.add(pd);
 				}
 			}
-			return (PropertyDescriptor[]) properties.toArray(new PropertyDescriptor[0]);
+			return (PropertyDescriptor[]) properties.toArray(new PropertyDescriptor[properties.size()]);
 		}
 		catch (IntrospectionException e) {
 			throw new CodeGenerationException(e);
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
@@ -234,9 +234,7 @@ abstract class AnnotationsScanner {
 		if (includeEnclosing && enclosingClass != null) {
 			R enclosingResult = processClassHierarchy(context, aggregateIndex,
 					enclosingClass, processor, classFilter, includeInterfaces, true);
-			if (enclosingResult != null) {
-				return enclosingResult;
-			}
+			return enclosingResult;
 		}
 		return null;
 	}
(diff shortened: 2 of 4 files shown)


pivotal-issuemaster:
@OLPMO Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23478)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23478) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@OLPMO Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23478)!

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
```diff
@@ -382,7 +382,7 @@ public static Method findNewInstance(Class iface) {
 			}
 		}
 		methods.remove(null);
-		return (Method[]) methods.toArray(new Method[methods.size()]);
+		return (Method[]) methods.toArray(new Method[0]);
```
mentallurg: This is bad, because it it creates **2 arrays**: first with `new Method[0]`, then `toArray()` will see that the capacity of the given array is not sufficient and will create **one more array**.

mentallurg (changes requested):
These changes actually lead to extra calls for memory allocation (which means worse performance) and make the code **worse**. Besides, this code uses **more** memory than the existing code, because even an empty array needs memory for meta. I suggest to **reject** this change.

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
sbrannen: I agree with @mentallurg here.

Please revert this change.

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
```diff
@@ -412,7 +412,7 @@ public static Method findNewInstance(Class iface) {
 					properties.add(pd);
 				}
 			}
-			return (PropertyDescriptor[]) properties.toArray(new PropertyDescriptor[properties.size()]);
+			return (PropertyDescriptor[]) properties.toArray(new PropertyDescriptor[0]);
```
sbrannen: Please revert this change.

spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
```diff
@@ -234,9 +234,7 @@ private AnnotationsScanner() {
 		if (includeEnclosing && enclosingClass != null) {
 			R enclosingResult = processClassHierarchy(context, aggregateIndex,
 					enclosingClass, processor, classFilter, includeInterfaces, true);
-			if (enclosingResult != null) {
-				return enclosingResult;
-			}
+			return enclosingResult;
```
sbrannen: Although this may be technically equivalent, it breaks with the structure of the rest of the method.

As such, I do not think the change is warranted.

So, please revert this change.

spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
```diff
@@ -323,9 +321,7 @@ private AnnotationsScanner() {
 		if (superclass != Object.class && superclass != null) {
 			R superclassResult = processMethodHierarchy(context, aggregateIndex,
 					superclass, processor, classFilter, rootMethod, includeInterfaces);
-			if (superclassResult != null) {
-				return superclassResult;
-			}
+			return superclassResult;
```
sbrannen: Although this may be technically equivalent, it breaks with the structure of the rest of the method.

As such, I do not think the change is warranted.

So, please revert this change.

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
andrei-ivanov: What about https://shipilev.net/blog/2016/arrays-wisdom-ancients/ ?

OLPMO: @andrei-ivanov The information you provided is much more detailed than mine. I have learned a lot from it, thank you.

sbrannen: Indeed. I had momentarily forgotten about those findings from Shipilёv.

Thanks for the links.

We'll keep the proposed change! 👍

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
sbrannen: See previous comments.

sbrannen (approved):

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
sbrannen: Please revert this change.

Rationale:

_Please refrain from modifying classes under `org.springframework.asm`, `org.springframework.cglib`, and `org.springframework.objenesis`. Those include repackaged forks of the third-party libraries ASM, CGLIB, and Objenesis. Any refactoring to those classes should take place upstream in the originating repository. The Spring Framework will then pick up the changes when syncing with official updates of the forked third-party libraries._

sbrannen (changes requested):
The changes look good now.

I have only asked for the change to the repackaged CGLIB code be reverted due to our policy for repackaged third-party libraries.

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
OLPMO: OK, thank you for your analysis. @sbrannen

sbrannen (approved):

sbrannen:

Thanks

OLPMO:
It's my pleasure.

Issue #23461, opened 2019-08-14
Consider addressing alerts reported by lgtm.com


jbduncan:
I learned about a static analysis website called [lgtm.com](https://lgtm.com) some time ago, and I noticed that the Spring Framework has an entry on the website where currently [115 alerts](https://lgtm.com/projects/g/spring-projects/spring-framework/alerts) are reported against it, a few of which seem to be security-related.

I personally think these alerts should be looked into and ideally fixed. What's the Spring team's thoughts on this? :thinking:

jbduncan:
(I'm unsure whether I should disclose the security-related alerts to https://pivotal.io/security, so I'd love some guidance on this.)

jbduncan:
I raised a similar issue at https://github.com/spring-projects/spring-boot/issues/17868, where I was told that I should have indeed responsibly disclosed the security-related alerts after verifying them. So I intend to do so when I next have free time.

snicoll:
@jbduncan indeed. Closing as per https://github.com/spring-projects/spring-boot/issues/17868#issuecomment-521375630

PR #23485, opened 2019-08-19
Fix "array index out of bounds" problem reported by LGTM.com

jbduncan:
This PR is a follow-on to https://github.com/spring-projects/spring-framework/issues/23461.

Specifically the problem that this PR fixes is https://lgtm.com/projects/g/spring-projects/spring-framework/snapshot/fb09647cbd34d6a540dc6e9b7e86aa7bcbe8e223/files/spring-web/src/main/java/org/springframework/http/ContentDisposition.java#xcfd962975fb7b533:1.

Feedback and other thoughts are welcome!

### DIFF
--- a/spring-web/src/test/java/org/springframework/http/ContentDispositionTests.java
+++ b/spring-web/src/test/java/org/springframework/http/ContentDispositionTests.java
@@ -28,6 +28,7 @@ import org.springframework.util.ReflectionUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * Unit tests for {@link ContentDisposition}
@@ -35,7 +36,6 @@ import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException
  * @author Sebastien Deleuze
  */
 public class ContentDispositionTests {
-
 
 	@Test
 	public void parseTest() {
@@ -196,6 +196,24 @@ public class ContentDispositionTests {
 		ReflectionUtils.makeAccessible(decode);
 		assertThatIllegalArgumentException().isThrownBy(() ->
 				ReflectionUtils.invokeMethod(decode, null, "UTF-16''test"));
+	}
+
+	@Test
+	public void decodeHeaderFieldParamShortInvalidEncodedFilename() {
+		Method decode = ReflectionUtils.findMethod(ContentDisposition.class,
+				"decodeHeaderFieldParam", String.class);
+		ReflectionUtils.makeAccessible(decode);
+		assertThatIllegalArgumentException().isThrownBy(() ->
+				ReflectionUtils.invokeMethod(decode, null, "UTF-8''%A"));
+	}
+
+	@Test
+	public void decodeHeaderFieldParamLongerInvalidEncodedFilename() {
+		Method decode = ReflectionUtils.findMethod(ContentDisposition.class,
+				"decodeHeaderFieldParam", String.class);
+		ReflectionUtils.makeAccessible(decode);
+		assertThatIllegalArgumentException().isThrownBy(() ->
+				ReflectionUtils.invokeMethod(decode, null, "UTF-8''%A.txt"));
 	}
 
 }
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@jbduncan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23485)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23485) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jbduncan Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23485)!

rstoyanchev:
Could you re-work the tests to avoid the use reflection?

mentallurg:
@rstoyanchev: The only reason for reflection is to test private method. Then one should implement a test for public methods. I find it better. But the tests already contained 2 tests for exactly this same private method and they of course use reflection. Be **consistent**. Either accept new reflection based tests or rework other _already existing_ reflection based tests.

mentallurg (approved):
I find the reflection based tests OK in this case. If one prefers not to use reflection in the tests, then it should be handled in a separate issue, and that would be a huge change, because there are already many reflection based tests in Spring.

jbduncan:
Hi @rstoyanchev @mentallurg, sorry for the delay in responding! I was on holiday and then catching up with work and IRL things.

I wasn't sure about introducing a reflection-based test either, but I used the same reasoning as @mentallurg did and I decided to use reflection for consistency with the other tests.

@rstoyanchev Would you prefer that I switch all the reflection-based tests in `ContentDispositionTests` to use the public API instead?

I'm not sure if either of you are Spring maintainers, but is there anything else you'd like me to do before this PR is merged in? :)

sbrannen:
> I'm not sure if either of you are Spring maintainers

@jbduncan, you can look at comments from people to see if there is a `Member` button, like the one you see for Rossen here: https://github.com/spring-projects/spring-framework/pull/23485#issuecomment-522908909

The _People_ page can also be of assistance for the entire Spring portfolio: https://github.com/orgs/spring-projects/people

spring-web/src/test/java/org/springframework/http/ContentDispositionTests.java
```diff
@@ -28,6 +28,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
sbrannen:
```suggestion
```

Delete unused import

sbrannen:
> I wasn't sure about introducing a reflection-based test either, but I used the same reasoning that @mentallurg did and I decided to use reflection for consistency with the other tests.

That's valid reasoning.

However, I don't see any reason that `encodeHeaderFieldParam()` and `decodeHeaderFieldParam()` should be `private`. Changing their visibility to package-private would remove the need for reflection in the tests since `ContentDispositionTests` resides in the same package as `ContentDisposition`, and making that change shouldn't hurt anything with regard to encapsulation.

@rstoyanchev, WDYT?

spring-web/src/test/java/org/springframework/http/ContentDispositionTests.java
jbduncan: Thanks @sbrannen! Done.

jbduncan:
Hey @sbrannen, fancy meeting you here. :)

> @jbduncan, you can look at comments from people to see if there is a Member button, like the one you see for Rossen here: #23485 (comment)
>
> The People page can also be of assistance for the entire Spring portfolio: https://github.com/orgs/spring-projects/people

Oh cool, thank you for reminding me about that button and the existence of GitHub's "People" tab. Cheers! :+1:

rstoyanchev:
This PR has been merged. I removed the use of reflection by doing actual parsing and/or formatting with an encoded filename attribute. There were even some existing tests like that so I don't know why we had reflection there in the first place. 

The refactoring helped to uncover a small issue with tolerating `"filename*"` without a charset (i.e. US_ASCII) which the decode method was dealing with but the duplicate code in parse wasn't.

jbduncan:
Thanks @rstoyanchev! :+1:

PR #23494, opened 2019-08-20
Fix groupId for spring-framework-bom

dreis2211:
Hi,

the recent changes to the Gradle builds seem to have caused the BOM to be published with a wrong groupId. E.g. it is `<groupId>spring</groupId>` now while it was (and probably should be) `<groupId>org.springframework</groupId>`

This PR is an attempt to fix this by applying the group as in [build.gradle#L53](https://github.com/spring-projects/spring-framework/blob/master/build.gradle#L53).

Cheers,
Christoph

### DIFF
--- a/spring-framework-bom/spring-framework-bom.gradle
+++ b/spring-framework-bom/spring-framework-bom.gradle
@@ -1,4 +1,5 @@
 description = "Spring Framework (Bill of Materials)"
+group = "org.springframework"
 
 apply plugin: "java"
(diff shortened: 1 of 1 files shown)


bclozel:
Sorry I missed your PR @dreis2211 ! You've probably noticed I've been struggling with our Gradle build, but everything should be fine now.
It's been fixed in 7ce1f5e652fee3f already.
Thanks!

PR #23521, opened 2019-08-26
Use dependency management in Framework build

bclozel:
Prior to this commit, the Spring Framework build would partially use the
dependency management plugin to import and enforce BOMs.

This commit applies the dependency management plugin to all Java
projects and regroups all version management declaration in the root
`build.gradle` file (versions and exclusions).

Some versions are overridden in specific modules for
backwards-compatibility reasons or extended support.


bclozel:
Note that while working on that PR, I've found a few discrepancies with dependency versions.

hibernate-core was (now all aligned on `5.4.4`):
* `5.1.17` in integration-tests
* `5.4.3` in spring-test
* `5.4.4` in spring-orm

joda-time (now all aligned on `2.10.2`):
* `2.10.1` spring-context
* `2.10.2` spring-webmvc

I've left some of those as I think they were intentional.

servlet is still:
* `3.1` in spring-orm and spring-web
* `4.0.1` everywhere else

validation dependencies:
* in in spring-context, hibernate-validator is `5.4.3` and validation-api is `1.1.0`
* in context-support, test, webmvc, webflux, hibernate-validator is `6.0.17` and validation-api is 2.0.1

com.sun.xml.bind dependencies are a mix of `2.3.0.1` and `2.3.1` (jaxb-core `2.3.1` is not released)

PR #23530, opened 2019-08-27
Remove redundant code and simplify the code of invokeBeanFactoryPostProcessors method

tswstarplanet:
Remove redundant code and simplify the code


sbrannen:
The proposed change actually changes the order in which the `orderedPostProcessors` and `nonOrderedPostProcessors` are initialized in the `BeanFactory`, so I'm not sure that we want to change the behavior in that regard.

tswstarplanet:
@sbrannen Oh, thank your comments! I think I would make a mistake.

PR #23557, opened 2019-08-31
Introduce MockMvc content() method to convert object to JSON

AlexJuca:
While using springs mock mcv class to handle tests in a spring boot API it is **necessary** to send JSON content with the request body.

Eg.
```java
mockMvc.perform(post("/api/v1/user_registration/phone_code/resend").content(jsonBody).contentType(MediaType.APPLICATION_JSON_UTF8))
                .andDo(print())
                .andExpect(MockMvcResultMatchers.status().isOk()); 
```
Unfortunately, the only way I can do this is by first converting the object that I wish to send in the request body to JSON and then pass it as an argument to the ```MockHttpServletRequestBuilder.content()``` method. 

This usually adds extra noise to tests to handle the object to JSON serialization.

It would be very handy for those of us writing tests for APIs to have a ```content(Object content)``` method that 
accepts a plain old java object and does the conversion to JSON for us as well as define the content-type to "application/js" since it is obvious by the context that we will be sending JSON in the request body.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,6 +122,7 @@ configure(allprojects.findAll { (it.name != "framework-bom") } ) { project ->
 	}
 
 	dependencies {
+        compile("com.fasterxml.jackson.core:jackson-databind:${jackson2Version}")
 		testCompile("org.junit.jupiter:junit-jupiter-api")
 		testCompile("org.junit.jupiter:junit-jupiter-params")
 		testCompile("org.mockito:mockito-core:3.0.0") {
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
@@ -269,10 +269,9 @@ public class MockHttpServletRequestBuilder
 	}
 
 	/**
-	 * Set the request body as a UTF-8 string and
-	 * Tries to convert the object to a JSON object.
+	 * Set the request body as a UTF-8 formatted JSON object.
 	 * Automatically sets the content type to application/json.
-	 * @param content the body content that will be converted to json
+	 * @param content the object that will be converted to json
 	 */
 	public MockHttpServletRequestBuilder content(Object content) {
 		ObjectMapper mapper = new ObjectMapper();
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java
@@ -27,12 +27,16 @@ import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.*;
 import java.util.Map;
 import javax.servlet.ServletContext;
 import javax.servlet.ServletRequest;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpSession;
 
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@AlexJuca Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23557)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23557) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@AlexJuca Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23557)!

sbrannen:
If we were to introduce a feature like this, I think it would be a bad idea to have the `content(Object)` implementation limited to JSON.

Rather, we would likely want to harness the support for various `HttpMessageConverter` implementations. In that regard, the content type would need to be supplied so that content negotiation could be leveraged to pick the appropriate `HttpMessageConverter`.

@rstoyanchev, what are your thoughts on the matter?

AlexJuca:
@sbrannen I agree that it would be limited. Having a single content(Object) method convert the object to JSON, XML etc would be a considerable improvement over this implementation. I will update my PR with your feedback.

AlexJuca:
@sbrannen I have a question. Where can I find the description of the labels used for the project? Example: in-test, waiting-for-triage...etc

sbrannen:
> @sbrannen I have a question. Where can I find the description of the labels used for the project? Example: in-test, waiting-for-triage...etc

You can hover your mouse over a label to see its description.

To see all descriptions, visit the [Labels](https://github.com/spring-projects/spring-framework/labels) page.

rstoyanchev:
MockMvc is built on the Servlet API mock objects (request, response, etc) and as such is a server side framework that helps you set the content of the request and inspect the content of the response but converting to and from higher level objects is a level above that. By contrast the `WebTestClient` is an actual client that is not aware of server-side request and response objects and naturally it provides higher level serialization and deserialization.

I don't think we should add this to MockMvc. Either enter the JSON or use a helper to produce it. For full integration tests you can use the WebTestClient against Spring MVC (or any server).

AlexJuca:
I still believe mockMVC should allow us to write more focused tests without having to worry about implementing code to do the conversion for us. In that way, our tests can remain focused and less noisy. 

If we can do this:

```java
(MockMvcResultMatchers.jsonPath("$['errors'][0]['title']").value("User or password incorrect")
```
To parse JSON and find matches as part of the entire test operation, I don't see why MockMvc can't do the conversion for us to keep out tests as focused as possible with as minimal noise as problem. 

In an alternative reality, the user could do the JSON parsing himself but that would lead to more noise in the tests. I mostly want to focus on testing my business code and have MockMVC deal with the rest.

rstoyanchev:
It's not that it isn't a valid or useful feature. There is a lot more involved in building a serialization/deserialization layer than simply invoking Jackson's `ObjectMapper` and that's just not in scope for MockMvc. 

For your own needs adding 10 lines of code is sufficient and I would argue it is a perfect solution for you. For us to add such support, it would have to be much more comprehensive from the start and not be tied to Jackson. Once we make this possible there would be further requests to support it on the server side, as well as to continue to evolve it over time would be legitimate too.

A much more feasible route for us would be to support testing of Spring MVC controllers without a running Servlet container via `WebTestClient`.

AlexJuca:
I have a question just so I can better understand your argument:

I believe Jackson is already being used internally, right? In what way should the feature not be tied with Jackson?

rstoyanchev:
I think @sbrannen already [expressed this well](https://github.com/spring-projects/spring-framework/pull/23557#issuecomment-526919839) above.

PR #23562, opened 2019-09-02
Omit cancellation of transactional Monos in TransactionOperator

### LINKED ISSUE #23304: No means to synchronize when cancelling a TransactionalOperator Subscription (Publisher)
Canceling a `Subscription` (in reactive transactions) leads to a state where transaction cleanup happens asynchronously and detached from completion signals.
Consider the following code:

```java
TransactionalOperator transactionalOperator = …;
DatabaseClient databaseClient = …;
JdbcTemplate jdbcTemplate = …;

Flux<Integer> integerFlux = transactionalOperator.execute(status -> {
	return databaseClient
		.execute("INSERT INTO legoset (idl) VALUES(42055)").fetch().rowsUpdated();
});

Mono<Integer> next = integerFlux.next();

next.as(StepVerifier::create).expectNext(1).verifyComplete();

assertThat(jdbcTemplate.queryForMap("SELECT id, name, manual FROM legoset")).hasEntrySatisfying("id",42055);
```

Initially, the table is empty and both, `DatabaseClient` and `JdbcTemplate` are configured to point to the same database.

The assertion with `queryForMap` typically fails with `EmptyResultDataAccessException`. This is, because calling `integerFlux.next()` cancels the upstream subscription while emitting completion as soon as an element was emitted. 

`TransactionalOperator` and its backing `ReactiveTransactionManager` implementations issue a commit to clean up the transaction that happens asynchronously, without an ability to await commit completion.

Not sure whether we can fix the problem at all or whether we can mitigate it. One approach could be with `TransactionalOperator.transactional(Mono)` to cancel the innermost `Publisher` and hand out a `Mono`. This change would return a properly bounded `Mono` and cancellation would happen on the innermost `Publisher` and preventing cancellation of the `Publisher` that is returned from `TransactionalOperator`.

This ticket is an opportunity to discuss that effect and its potential impact on cancellation of `Publishers` which are enhanced for transactions.

/cc @smaldini @simonbasle

mp911de:
`TransactionOperator.as(Mono)` now no longer short-cuts via a `Flux.next()` but provides an implementation via `Mono.usingWhen(…)`.
The short-cut previously issued a cancellation signal to the transactional `Mono` causing the transaction cleanup to happen without a handle for synchronization.

Using `Mono.usingWhen(…)` initiates transaction cleanup when the `Mono` completes eliminating the need for cancellation of the transactional `Publisher`.

This change does not fully fix #23304 but it softens its impact because `TransactionalOperator.transactional(Mono)` avoids cancellation.

/cc @michael-simons @simonbasle @smaldini 

---

- [x] We should get feedback from Michael or someone from the reactor team prior to merging this PR.

### DIFF
--- a/spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperatorImpl.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperatorImpl.java
@@ -71,6 +71,21 @@ final class TransactionalOperatorImpl implements TransactionalOperator {
 		return this.transactionManager;
 	}
 
+	@Override
+	public <T> Mono<T> transactional(Mono<T> mono) {
+		return TransactionContextManager.currentContext().flatMap(context -> {
+			Mono<ReactiveTransaction> status = this.transactionManager.getReactiveTransaction(this.transactionDefinition);
+			// This is an around advice: Invoke the next interceptor in the chain.
+			// This will normally result in a target object being invoked.
+			// Need re-wrapping of ReactiveTransaction until we get hold of the exception
+			// through usingWhen.
+			return status.flatMap(it -> Mono.usingWhen(Mono.just(it), ignore -> mono,
+					this.transactionManager::commit, s -> Mono.empty())
+					.onErrorResume(ex -> rollbackOnException(it, ex).then(Mono.error(ex))));
+		})
+		.subscriberContext(TransactionContextManager.getOrCreateContext())
+		.subscriberContext(TransactionContextManager.getOrCreateContextHolder());
+	}
 
 	@Override
 	public <T> Flux<T> execute(TransactionCallback<T> action) throws TransactionException {
--- a/spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperator.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperator.java
@@ -38,6 +38,10 @@ import org.springframework.transaction.TransactionException;
  * application services utilizing this class, making calls to the low-level
  * services via an inner-class callback object.
  *
+ * <p>Transactional Publishers should avoid Subscription cancellation.
+ * Cancelling initiates asynchronous transaction cleanup that does not allow for
+ * synchronization on completion.
+ *
  * @author Mark Paluch
  * @author Juergen Hoeller
  * @since 5.2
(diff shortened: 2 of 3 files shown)


spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperatorImpl.java
```diff
@@ -71,6 +71,21 @@ public ReactiveTransactionManager getTransactionManager() {
...
 
+	@Override
+	public <T> Mono<T> transactional(Mono<T> mono) {
+		return TransactionContextManager.currentContext().flatMap(context -> {
+			Mono<ReactiveTransaction> status = this.transactionManager.getReactiveTransaction(this.transactionDefinition);
+			// This is an around advice: Invoke the next interceptor in the chain.
+			// This will normally result in a target object being invoked.
+			// Need re-wrapping of ReactiveTransaction until we get hold of the exception
+			// through usingWhen.
+			return status.flatMap(it -> Mono.usingWhen(Mono.just(it), ignore -> mono,
```
simonbasle: wouldn't it be better to start from `Mono.usingWhen`?
```java
return Mono.usingWhen(status, ignore -> mono, this.transactionManager::commit, ...)
```

spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperatorImpl.java
```diff
@@ -71,6 +71,21 @@ public ReactiveTransactionManager getTransactionManager() {
...
+					this.transactionManager::commit, s -> Mono.empty())
```
simonbasle: note that in case of cancellation, since no `asyncCancel` is defined here, it defaults to `commit`. Also, it feels weird that the error case is basically ignored

simonbasle (changes requested):

spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperatorImpl.java
mp911de: We require the status value outside of `usingWhen(…)` (see `rollbackOnException`)

spring-tx/src/main/java/org/springframework/transaction/reactive/TransactionalOperatorImpl.java
mp911de: Commit on cancel is what we want if just a subset of data is processed. In imperative code this compares to processing a subset of a result list.

The error case is ignored because errors are handled after the `usingWhen` operator. In case a pre-commit action fails, we want to issue a rollback. Committing a transaction consists of several activities of which the actual commit is just one of them. So if a commit fails, we still need to rollback,

simonbasle: makes sense, thanks for the explanations. nothing to add 👍

simonbasle (approved):

mp911de:
This PR is good to be merged from my side. It would be good to have this fix available in 5.2 GA.

PR #23567, opened 2019-09-02
Log HTTP method in logging filters and revise log message format

schnapster:
Ahoy,

I was surprised to neither find the http method in the log messages of the `AbstractRequestLoggingFilter` nor the option to turn them on.

This looks like a fairly uncontroversial, fully optional and backwards compatible change to me so I did not open an issue up front. I'm happy to open one if more discussion is required.

### Two things to note for the reviewer:
~~- [ ] I copy-pasted the method javadocs, and I'm not 100% sure if anything else needs to be done to get~~
```
	 * <p>Should be configured using an {@code <init-param>} for parameter name
	 * "includeHttpMethod" in the filter definition in {@code web.xml}.
```
~~to work. I did some CTRL-Fing for the other values and nothing interesting came up.~~

~~- [ ] I did not add a `@since 5.2` annotation as I'm not sure which version this will hit. Just amend it in :)~~

Cheers,
Napster

Btw, setting up the local build and running tests was a breeze, great docs!

### DIFF
--- a/spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
+++ b/spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
@@ -24,7 +24,6 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.mock.web.test.MockHttpServletRequest;
@@ -44,31 +43,23 @@ import static org.assertj.core.api.Assertions.assertThat;
  */
 public class RequestLoggingFilterTests {
 
-	private MyRequestLoggingFilter filter = new MyRequestLoggingFilter();
-
-	@BeforeEach
-	public void reset() {
-		filter = new MyRequestLoggingFilter();
-	}
+	private final MyRequestLoggingFilter filter = new MyRequestLoggingFilter();
 
 	@Test
 	public void defaultPrefix() throws Exception {
-		final MockHttpServletRequest request = new MockHttpServletRequest("POST", "/hotels");
+		MockHttpServletRequest request = new MockHttpServletRequest("POST", "/hotels");
 		MockHttpServletResponse response = new MockHttpServletResponse();
 
 		FilterChain filterChain = new NoOpFilterChain();
 		filter.doFilter(request, response, filterChain);
 
-		assertThat(filter.beforeRequestMessage).isNotNull();
-		assertThat(filter.beforeRequestMessage.startsWith(AbstractRequestLoggingFilter.DEFAULT_BEFORE_MESSAGE_PREFIX)).isTrue();
-
-		assertThat(filter.afterRequestMessage).isNotNull();
-		assertThat(filter.afterRequestMessage.startsWith(AbstractRequestLoggingFilter.DEFAULT_AFTER_MESSAGE_PREFIX)).isTrue();
+		assertThat(filter.beforeRequestMessage).startsWith(AbstractRequestLoggingFilter.DEFAULT_BEFORE_MESSAGE_PREFIX);
+		assertThat(filter.afterRequestMessage).startsWith(AbstractRequestLoggingFilter.DEFAULT_AFTER_MESSAGE_PREFIX);
 	}
 
 	@Test
 	public void customPrefix() throws Exception {
-		final MockHttpServletRequest request = new MockHttpServletRequest("POST", "/hotels");
+		MockHttpServletRequest request = new MockHttpServletRequest("POST", "/hotels");
 		MockHttpServletResponse response = new MockHttpServletResponse();
 
 		filter.setBeforeMessagePrefix("Before prefix: ");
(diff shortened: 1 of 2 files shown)


sbrannen:
After a quick discussion with @rstoyanchev, we have decided that we should simply always include the HTTP method in those logs messages. In other words, there's no need to make it optional.

Can you please rework this PR accordingly?

sbrannen:
**Team Decision**:

After some further internal brainstorming, we have decided on the following.

- [ ] Always include the HTTP method.
- [ ] Change the format of the generated log message to the following, keeping the existing _optional_ behavior in tact for the query string, client info, headers, and payload.

Format: `<prefix><method> <uri + query string>, <client>, <session>, <user>, <headers>, <payload><suffix>`

Example: `Before request [GET /foo?x=2, client=..., session=..., user=..., headers=..., payload=...]`

@napstr, please let us know if you are willing to rework this PR along those lines.

Thanks

napstr:
Will do, I'll aim to get this done over the weekend :)

Just to clarify, did you also decide to change the existing `;`s to `, `s, @sbrannen or that a typo in the format/example?

sbrannen:
The Spring Team would also like to point out that the `logRecord()` method in the `DispatcherServlet` is capable of logging very detailed information at DEBUG and TRACE levels.

This detailed logging in the `DispatcherServlet` is available since Spring Framework 5.0 and may serve as an alternative to `AbstractRequestLoggingFilter` and its subclasses (e.g., `CommonsRequestLoggingFilter`, `ServletContextRequestLoggingFilter`, and custom implementations).

Consult the [Logging](https://docs.spring.io/spring/docs/5.1.9.RELEASE/spring-framework-reference/web.html#mvc-logging) section of the _Spring Web MVC_ documentation for further details.

sbrannen:
> Will do, I'll aim to get this done over the weekend :)

Great!

> Just to clarify, did you also decide to change the existing `;`s to `, `s, @sbrannen or that a typo in the format/example?

Yes, we decided to change the delimiter from `;` to `, ` (with the comma) in order to make it more human readable and to align with the format in `org.springframework.web.servlet.DispatcherServlet.logRequest(HttpServletRequest)`.

napstr:
How does this look?

spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
```diff
@@ -44,8 +44,96 @@
  */
 public class RequestLoggingFilterTests {
 
-	private final MyRequestLoggingFilter filter = new MyRequestLoggingFilter();
```
sbrannen: Please revert the `@BeforeEach` change. It is unnecessary and undoes the exact opposite change introduce in https://github.com/spring-projects/spring-framework/commit/08ddd1b3bc783d0dc0ef1ab5e361315cbba90e10#diff-8c84ed40b500108bc760d43f76bbdaf9R43.

spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
```diff
@@ -44,8 +44,96 @@
...
+
+	@Test
+	public void defaultPrefix() throws Exception {
+		final MockHttpServletRequest request = new MockHttpServletRequest("POST", "/hotels");
+		MockHttpServletResponse response = new MockHttpServletResponse();
+
+		FilterChain filterChain = new NoOpFilterChain();
+		filter.doFilter(request, response, filterChain);
+
+		assertThat(filter.beforeRequestMessage).isNotNull();
```
sbrannen:
```suggestion
```

spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
```diff
@@ -44,8 +44,96 @@
...
+		assertThat(filter.beforeRequestMessage.startsWith(AbstractRequestLoggingFilter.DEFAULT_BEFORE_MESSAGE_PREFIX)).isTrue();
```
sbrannen:
```suggestion
		assertThat(filter.beforeRequestMessage).startsWith(AbstractRequestLoggingFilter.DEFAULT_BEFORE_MESSAGE_PREFIX);
```

With AssertJ, there's no need to check for "not null" and "is true". Simply use `.startsWith(...)` which handles both of those use cases.

Please apply the above two changes to all affected test methods.

spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
```diff
@@ -44,8 +44,96 @@
...
+	private MyRequestLoggingFilter filter = new MyRequestLoggingFilter();
 
+	@BeforeEach
+	public void reset() {
+		filter = new MyRequestLoggingFilter();
+	}
+
+	@Test
+	public void defaultPrefix() throws Exception {
+		final MockHttpServletRequest request = new MockHttpServletRequest("POST", "/hotels");
```
sbrannen:
```suggestion
		MockHttpServletRequest request = new MockHttpServletRequest("POST", "/hotels");
```

Please make this change to all affected test methods.

spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
```diff
@@ -58,11 +146,11 @@ public void uri() throws Exception {
 		filter.doFilter(request, response, filterChain);
 
 		assertThat(filter.beforeRequestMessage).isNotNull();
-		assertThat(filter.beforeRequestMessage.contains("uri=/hotel")).isTrue();
+		assertThat(filter.beforeRequestMessage.contains("/hotel")).isTrue();
```
sbrannen:
```suggestion
		assertThat(filter.beforeRequestMessage).contains("/hotel");
```

This is analogous to the aforementioned `.startsWith()` change. Please apply to all affected methods.

spring-web/src/test/java/org/springframework/web/filter/RequestLoggingFilterTests.java
```diff
@@ -58,11 +146,11 @@ public void uri() throws Exception {
 		filter.doFilter(request, response, filterChain);
 
 		assertThat(filter.beforeRequestMessage).isNotNull();
```
sbrannen:
```suggestion
```

sbrannen (changes requested):
I think the changes look good.

I've requested simplifications to the tests. So please make those and perhaps add an additional test that's more inclusive -- for example, one that contains everything like the previously posted example: `Before request [GET /foo?x=2, client=..., session=..., user=..., headers=..., payload=...]`.

Thanks

napstr:
@sbrannen thank you for the hints on how to improve usage of the available AssertJ api. I believe I applied all your suggestions now - let me know if anything is left.

sbrannen (approved):

sbrannen:
> @sbrannen thank you for the hints on how to improve usage of the available AssertJ api.

Glad to be of service.

> I believe I applied all your suggestions now - let me know if anything is left.

I think that looks good now. I've accepted the changes and will merge the PR in the coming days.

Enjoy your weekend!

sbrannen:
This has been merged into `master` in 33cee066e55fea156f024bc4469fb056aba11f82, and the tests have been further polished in 52128fe86391b945a99b6611a47817a00d0b3eda.

Thanks for the PR

Issue #23578, opened 2019-09-04
PostProcessorRegistrationDelegate下的invokeBeanFactoryPostProcessors方法，性能问题


SuiSuiNian19:
```java
List<BeanDefinitionRegistryPostProcessor> currentRegistryProcessors = new ArrayList<BeanDefinitionRegistryPostProcessor>();

			// First, invoke the BeanDefinitionRegistryPostProcessors that implement PriorityOrdered.
			String[] postProcessorNames =
					beanFactory.getBeanNamesForType(BeanDefinitionRegistryPostProcessor.class, true, false);
			for (String ppName : postProcessorNames) {
				if (beanFactory.isTypeMatch(ppName, PriorityOrdered.class)) {
					currentRegistryProcessors.add(beanFactory.getBean(ppName, BeanDefinitionRegistryPostProcessor.class));
					processedBeans.add(ppName);
				}
			}
			sortPostProcessors(currentRegistryProcessors, beanFactory);
			registryProcessors.addAll(currentRegistryProcessors);
			invokeBeanDefinitionRegistryPostProcessors(currentRegistryProcessors, registry);
			currentRegistryProcessors.clear();

			// Next, invoke the BeanDefinitionRegistryPostProcessors that implement Ordered.
			postProcessorNames = beanFactory.getBeanNamesForType(BeanDefinitionRegistryPostProcessor.class, true, false);
			for (String ppName : postProcessorNames) {
				if (!processedBeans.contains(ppName) && beanFactory.isTypeMatch(ppName, Ordered.class)) {
					currentRegistryProcessors.add(beanFactory.getBean(ppName, BeanDefinitionRegistryPostProcessor.class));
					processedBeans.add(ppName);
				}
			}
			sortPostProcessors(currentRegistryProcessors, beanFactory);
			registryProcessors.addAll(currentRegistryProcessors);
			invokeBeanDefinitionRegistryPostProcessors(currentRegistryProcessors, registry);
			currentRegistryProcessors.clear();

			// Finally, invoke all other BeanDefinitionRegistryPostProcessors until no further ones appear.
```

lgxbslgx:
The issue description in English is shown below.

**Title**
A performance-related optimization recommendation in `PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors` 

**Content**
```Java
                        List<BeanDefinitionRegistryPostProcessor> currentRegistryProcessors = new ArrayList<>();

[...]

PR #23580, opened 2019-09-04
Optimize PostProcessorRegistrationDelegate to improve performance

lgxbslgx:
A performance-related optimization recommendation in `PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors`

See #23578.


sbrannen:
This is essentially a duplicate of #23530, which was withdrawn due to https://github.com/spring-projects/spring-framework/pull/23530#issuecomment-525803947.

sbrannen:
See also #23583

jhoeller:
This is unfortunately not equivalent, since `BeanDefinitionRegistryPostProcessors` may register further `BeanDefinitionRegistryPostProcessors` which we have to pick up still. A key reason to declared a post-processor as `PriorityOrdered` over `Ordered` is exactly the ability to register further post-processors. From that perspective, we have to retain our separate retrieval calls as distinct phases.

PR #23583, opened 2019-09-04
use lambda & code simplify

HollowTraces:
1.use lambda.
2.In my opinion , orderedPostProcessorNames and nonOrderedPostProcessorNames are unnecessary,beanPostProcessors can be added directly to the list insted of two additional cycles.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/PostProcessorRegistrationDelegate.java
+++ b/spring-context/src/main/java/org/springframework/context/support/PostProcessorRegistrationDelegate.java
@@ -63,7 +63,7 @@ final class PostProcessorRegistrationDelegate {
 			List<BeanFactoryPostProcessor> regularPostProcessors = new ArrayList<>();
 			List<BeanDefinitionRegistryPostProcessor> registryProcessors = new ArrayList<>();
 
-			for (BeanFactoryPostProcessor postProcessor : beanFactoryPostProcessors) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@fakefrog Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23583)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23583) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@fakefrog Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23583)!

sbrannen:
This is essentially a duplicate of #23530, which was withdrawn due to https://github.com/spring-projects/spring-framework/pull/23530#issuecomment-525803947.

jhoeller:
Immediately retrieving post-processor instances is unfortunately not an option since each such post-processor may affect subsequent post-processor bean definitions, so we need to instantiate each processor as lazily as possible, also keeping up the distinct phases for ordered and non-ordered ones.

As for the lambda expression, while such a change is generally feasible, we tend to retain traditional iteration in the core container, not least of it all since plain iteration leads to simpler stacktraces.

PR #23591, opened 2019-09-05
Add opt-in build scan integration with Gradle Enterprise

wilkinsona:
With thanks to @etiennestuder for his guidance, this pull request adds opt-in build scan integration with Gradle Enterprise. When the GRADLE_ENTERPRISE_URL environment variable is set on a developer's machine or on CI, a build scan will be automatically uploaded to Gradle Enterprise at the end of the build.

This initial integration will establish a baseline for Spring Framework builds. Once that has happened we can use the build scans to identify ways in which the build can be optimized and updated to make use of Gradle's build caching which should reduce build times, significantly so for changes that only affect tasks near the leaf nodes of the task graph.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,9 +12,19 @@ plugins {
 	id 'org.asciidoctor.convert' version '1.5.8'
 	id 'io.spring.nohttp' version '0.0.3.RELEASE'
 	id 'de.undercouch.download' version '4.0.0'
+	id 'com.gradle.build-scan' version '2.4.1'
 	id "com.jfrog.artifactory" version '4.9.8' apply false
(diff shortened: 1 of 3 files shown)


bclozel:
Thanks @wilkinsona !
This is now in master and our CI (for the master branch only) is publishing scans automatically.
The first build using this feature is: https://build.spring.io/browse/SPR-PUBM-26

sbrannen:
Awesome! Thanks, @wilkinsona. 👍

etiennestuder:
Awesome @bclozel. Next step is that @wilkinsona and I will look at build performance and possible optimizations.

PR #23597, opened 2019-09-06
Tweak detection of Bamboo

wilkinsona:
When exposed as an environment variable Bamboo's `bamboo.resultsUrl` is mapped to `bamboo_resultsUrl`. This PR updates the build scan user data script to look for the latter rather than the former.


sbrannen:

Thanks

PR #23607, opened 2019-09-07
DefaultWebSessionManager should not cancel the subscription to WebSessionStore when there is no need to.

### LINKED ISSUE #23606: spring-web: DefaultWebSessionManager cancels subscription to Mono from underlying ReactiveSessionRepository#findById
I am building a `ReactiveSessionRepository` and was surprised that after adding metrics I saw a lot of `cancellation` signals on `findById` which I did not expect.

The cause for this how `DefaultWebSessionManager#retrieveSession` deals with multiple `sessionIds` as it uses a [`MonoNext`](https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/MonoNext.java#L75) which causes a cancel `onNext`.

This feels a bit unexpected and I'm not really sure if this was an intended behaviour, anyway I created a PR to explain the issue including a fix.

AndreasKl:
After receiving the first signal from `WebSessionStore#retrieveSession` the subscription on the `Mono` is
cancelled which could confuse `WebSessionStore` implementations.

The flow currently is: request -> onNext -> cancel -> onComplete
After the change the flow is: request -> onNext -> onComplete

Fixes gh-23606.

Author: Andreas Kluth <mail@andreaskluth.net>

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/session/DefaultWebSessionManager.java
+++ b/spring-web/src/main/java/org/springframework/web/server/session/DefaultWebSessionManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,7 +18,6 @@ package org.springframework.web.server.session;
 
 import java.util.List;
 
-import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


rstoyanchev:
This is now taking the last session instead of the first session id. The lambda could return `first` but instead of `reduce` something like this would be more illustrative I think:

```java
.collectList()
.filter(list -> !list.isEmpty())
.map(list -> {
        // list.get(0)
        // ...
})
```

AndreasKl:
@rstoyanchev `DefaultWebSessionManagerTests#multipleSessionIds` validates that the last session id is returned. I broke the test when I just returned the first element with `Stream#findFirst` which was my obvious first choice. I could replace the `reduce` with some variation or most likely more readable with an imperative condition.

rstoyanchev:
Oh I see. Actually it isn't about the last session id, but rather the first session id for which a session is found. In the test `retrieveSession` happens to return empty for the first two.

Given this I'm not sure we could rewrite without cancelling after the first found session. The only alternative would be to try and retrieve all sessions, collectToList, and pick the first.

AndreasKl:
I think then we should close this PR, as I now understand what the purpose of the code was :-)

Issue #23434, opened 2019-08-08
Please publish PGP signing keys publicly


EricPetersonIBKR:
This is a follow-up to issue gh-20182.

Could Spring please publish the **public** PGP keys used to create artifact signatures so that they are available via https download from an official Spring site?

That way users have a way to reliably verify the authenticity of downloaded artifacts.

rstoyanchev:
Maven requires those to be published (see "Distributing Your Public Key" under https://central.sonatype.org/pages/working-with-pgp-signatures.html), and can be downloaded. For example:

```
$ gpg --keyserver pool.sks-keyservers.net --recv-key 9A2C7A98E457C53D
gpg: key 9A2C7A98E457C53D: public key "Spring Buildmaster <buildmaster@springframework.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```

EricPetersonIBKR:
@rstoyanchev I don't think you've understood the problem I'm facing here yet. You need to understand that _anyone_ can upload a key to the public keyservers, and it can be associated with any email address. You gave the example of searching for the key on pool.sks-keyservers.net:

![Screen Shot 2019-08-22 at 7 46 09 AM](https://user-images.githubusercontent.com/1138860/63512154-13aef780-c4b1-11e9-986c-e764fefd0e27.png)

You can see that the key has not been signed by anyone! There is no indication that this key is legitimate. If I wanted to carry out a supply chain attack, this is what can happen: replace the artifacts on Maven with malicious ones, and also upload signatures of the malicious artifacts using a PGP key that the attacker controls. Then upload the public key to the keyservers, and bingo! If users follow your policy, they will trust the malicious artifacts because the only requirement you put on trusting a key is that it exists on a public keyserver.

Spring needs to indicate the _correct_ key for signing Maven artifacts, so that users can trust those key(s) and no others. Another way would be to leverage the PGP Web of Trust by having other people sign your PGP key, but as I pointed out, that has not been done here.

EricPetersonIBKR:
Here is an example of someone doing the correct thing:

https://pivotal.io/security
(See the section "Reporting a vulnerability".)

Notice that the Pivotal security team publishes the fingerprint of their PGP key on a website they own and which is protected by https. As far as I know, Spring has not published fingerprints (or better yet the public keys themselves) anywhere for the Maven signing keys.

bclozel:
[...]

PR #23615, opened 2019-09-10
Add checksum-dependency-plugin for verification of plugin/dependency checksums

vlsi:
`checksum-dependency-plugin` is a superset of `gradle-witness`, and it enables to increase the level of security.

See https://github.com/vlsi/vlsi-release-plugins#checksum-dependency-plugin
See https://medium.com/@vladimirsitniko/dependency-verification-checksum-vs-pgp-582e76207019?sk=7485298b76eaf9f935b899b002f4c3b5

See https://github.com/spring-projects/spring-framework/issues/23434

### DIFF
--- a/buildSrc/checksum.xml
+++ b/buildSrc/checksum.xml
@@ -1,0 +1,17 @@
+<?xml version='1.0' encoding='utf-8'?>
+<dependency-verification version='1'>
+  <trust-requirement pgp='GROUP' checksum='NONE' />
+  <ignored-keys />
+  <trusted-keys>
+    <trusted-key id='beabcfbee059e4e5' group='com.github.siom79.japicmp' />
+    <trusted-key id='10066a9707090cf9' group='org.javassist' />
+  </trusted-keys>
+  <dependencies>
+    <dependency group='javax.activation' module='activation' version='1.1'>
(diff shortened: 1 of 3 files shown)


vlsi:
Hi, I see the PR has not been reviewed much yet :)

You might be interested that [Gradle 6.2 introduces in-core dependency verification](https://github.com/gradle/gradle/issues/10443#issuecomment-568741472)

The documentation can be reviewed here: https://github.com/gradle/gradle/pull/11755

From what I know Gradle would cover more cases when compared with `checksum-dependency-plugin`. For instance, it will be able to verify `pom.xml` which are implicitly fetched by Gradle when resolving transitive dependencies and probably other cases.

Some bits can be previewed in the current release candidates, release nightly builds and master nightly builds (see https://gradle.org/releases/ )

It would be nice if you could preview the feature and provide your feedback.

bclozel:
Spring Framework has many optional dependencies and maintaining a `verification-metadata.xml` file that's over 330K big is not really manageable. In the meantime we've seen that sigstore (and it seems you're contributing there) is trying a different approach.

We're generally interested in this theme but we don't think the current approach will fit our project. We'll keep an eye on alternate solutions. Thanks!

vlsi:
@bclozel , sigstore won't be much different by the way.

In a nutshell:
* PGP: "we maintain a list of PGP keys and checksums we trust. The PGP key id for each dependency is taken from the official website"
* Sigstore: "we trust logback if the artifact was built with GitHub Action release.yaml running in github.com/qos-ch/logback repository"

In both cases, the amount of metadata is pretty much the same, and the main difference would be that "PGP key ids are unreadable" while "GitHub Action release.yaml running in github.com/qos-ch/logback repository" would be pretty-much understandable by humans.

PR #23616, opened 2019-09-10
Disable caching of changing modules (snapshots) and dynamic versions

wilkinsona:
Previously, changing modules (snapshots) and dynamic versions were cached for Gradle's default period of 24 hours and `--refresh-dependencies` was used to pick up the latest artifacts. This approach has a notable downside. `--refresh-dependencies` causes all dependencies to be refreshed, irrespective of whether they are expected to change. At a minimum, this results in a HEAD request for every dependency in the build. Running an up-to-date build without `--refresh-dependencies` takes in the region of 6 seconds:

```
$ ./gradlew build

BUILD SUCCESSFUL in 6s
203 actionable tasks: 203 up-to-date
```

The same build with --refresh-dependencies takes almost ten times as long:

```
$ ./gradlew build --refresh-dependencies

BUILD SUCCESSFUL in 58s
203 actionable tasks: 203 up-to-date
```

This commit replaces the manual usage of `--refresh-dependencies` on the command line with a 0 second caching period for changing modules and dynamic versions. This should remove the need to use `--refresh-dependencies` both locally and on CI, saving almost 1 minute per full build.

/cc @etiennestuder

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -279,7 +279,7 @@ configure(allprojects) { project ->
 			dependency "javax.xml.ws:jaxws-api:2.3.1"
 
 			dependency "org.eclipse.persistence:javax.persistence:2.2.0"
-			
+
 			// Substitute for "javax.management:jmxremote_optional:1.0.1_04" which
 			// is not available on Maven Central
 			dependency "org.glassfish.external:opendmk_jmxremote_optional_jar:1.0-b01-ea"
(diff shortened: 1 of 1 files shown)


bclozel:
@wilkinsona Should we keep this other directive around or is it duplicating effort with the changes introduced by this PR?

https://github.com/spring-projects/spring-framework/blob/333711fd36f7774fd0ffeebf514e83899b90f6c7/build.gradle#L293-L295

wilkinsona:
We should keep that. It controls caching of changing boms, which are handled separately to normal dependencies.

bclozel:
Thanks! I've updated the CI configuration before merging this PR.
The [next CI run](https://build.spring.io/browse/SPR-PUBM-37) should reflect those changes.

PR #23620, opened 2019-09-11
Enable parallel builds by default

wilkinsona:
Building in parallel is now recommended by the Gradle team so this pull request enables it by default. On my MacBook Pro it reduces build time by roughly 50%. Machines with greater or fewer cores may see a greater or lesser reduction in build time. There may also be improvements that can be made to the build to improve its parallelisation. We can hopefully identify these in the future once we've got a number of parallel builds to examine.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,4 @@
 version=5.2.0.BUILD-SNAPSHOT
 org.gradle.caching=false
+org.gradle.parallel=true
+org.gradle.jvmargs=-Xmx1536M


snicoll:
This works fine for me and the difference is quite impressive. However, I think it's more likely to fail our performance build that expects a test to complete in a certain amount of time.

Is there a way to disable parallel execution for the performance tests? Right now we pass some switches to the build so we can update CI that add a few more.

sbrannen:
> This works fine for me and the difference is quite impressive. However, I think it's more likely to fail our performance build that expects a test to complete in a certain amount of time.

I've been using `--parallel` for years now (locally on my machine), and it usually works well; however, it does increase the likelihood that some of our _flaky_, asynchronous tests fail. Locally, it's not a problem since I just rerun the previous build command, and the tests often then pass.

But... I do have some concerns that executing our CI builds with `--parallel` will cause more frequent failures. I guess we can just try it out to see how it goes on the Bamboo CI server.

> Is there a way to disable parallel execution for the performance tests? Right now we pass a switch to the build (`-DtestGroups=ALL`) so we can update the configuration to add one more I guess.

The [Gradle documentation](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_performance) states that `--no-parallel` should cover that.

wilkinsona:
>  I guess we can just try it out to see how it goes on the Bamboo CI server.

That's what I would prefer. If some flaky tests do start failing more frequently, I'd prefer to use that as motivation to try and fix those tests (and I'm happy to help with that if I can) rather than motivation to go back to sequential builds.

sbrannen:
> That's what I would prefer. 

OK. Let's go for it!

@snicoll, I see you've assigned this to yourself. So I assume you'll take care of it, but you can pass it off to me if you like.

> If some flaky tests do start failing more frequently, I'd prefer to use that as motivation to try and fix those tests (and I'm happy to help with that if I can) rather than motivation to go back to sequential builds.

That would be an extremely well welcomed contribution!!!! 🎉 

The team has discussed "those tests" on numerous occasions, but unfortunately no one has found the time to _fix_ them. For concrete examples, just look at some of our failed builds (like [this one](https://build.spring.io/browse/SPR-PUBM-32)).

snicoll:
Thank you very much Andy, this is now merged in `master`. I've disabled parallel support on our performance build and will monitor the situation.

PR #23623, opened 2019-09-11
Move jmx.log beneath build/ to avoid up-to-date pollution

wilkinsona:
Previously, some tests in spring-context wrote a file named jmx.log into spring-context/. This led to the file being included in the sources of the Checkstyle NoHTTP tasks, breaking that task's up-to-date checks and causing it to execute unnecessarily.

This PR updates the tests to write the jmx.log file beneath spring-context/build/ so that it is not included in the sources of the Checkstyle NoHTTP task.


sbrannen:
Good catch!

PR #23626, opened 2019-09-12
Optimize test code with lambdas where feasible

OLPMO:
Optimize code with lambda where feasible


OLPMO:
I thought these commits were simple, so I didn't rebuild the project. Sorry, it is my carelessness. Thank you for your reminder.

sbrannen:
No problem!

We (the core committers) get caught by those Checkstyle rules all the time. So we all just have to remember to at least run `./gradlew check` before submitting. 😉

PR #23628, opened 2019-09-12
Fix OkHttp3ClientHttpRequestFactory shutdown flow

alexey-anufriev:
This fix now triggers closing and removing all idle connections inside the connection pool used by OkHttpClient.

The problem was discovered after switching from Java 8 to Java 11 and consequently to HTTP/2.

The recommendation from OkHttp developer was a proper shutdown:
https://github.com/square/okhttp/issues/5446

Problem can be reproduced with this snippet of code:
https://github.com/alexey-anufriev/okhttpclient-shutdown-problem/commit/29abbda198dc2ad9aa2714141c0b17a573490d34#diff-70a5fa50b52c03a98de3a02c98de30a1


alexey-anufriev:
Anyone? Is there a chance to get this merged?

PR #23642, opened 2019-09-16
Fix race condition in sendBlockingMessage

wilkinsona:
Previously, tests in `ConcurrentWebSocketSessionDecoratorTests` that use the BlockingSession would fail intermittently. This appears to have been due to a race condition in `sendBlockingMessage` where the call to `getSentMessageLatch()` that stores a latch in `nextMessageLatch` on the main thread may happen after the call to `sendMessage` that counts down the latch if it is non-null occurs on the executor's thread.

This pull request updates `sendBlockingMessage` to call `getSentMessageLatch()`
(and therefore store the latch) before it sumbmits the task to the executor. This ensures that the latch will be available when the exeuctor's thread attempts to retrieve and decrement it.
`BlockingSession`'s `AtomicReference` fields have also been made `final` to eliminate the possibility of any visibility problems across threads.


sbrannen:

Thanks, @wilkinsona! 👍

rstoyanchev:
Thanks indeed!

PR #23658, opened 2019-09-18
Avoid ArrayIndexOutOfBoundsException in SpEL's Indexer

JoyceZhan:
When `index == arrayLength`, the array index is also out of bounds.

For this use case, a `SpelEvaluationException` should be thrown instead of
`ArrayIndexOutOfBoundsException`.

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/ast/Indexer.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/ast/Indexer.java
@@ -439,7 +439,7 @@ public class Indexer extends SpelNodeImpl {
 	}
 
 	private void checkAccess(int arrayLength, int index) throws SpelEvaluationException {
-		if (index > arrayLength) {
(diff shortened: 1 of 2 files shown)


sbrannen:
Can you please provide a test case that fails before the change and passes after the change?

JoyceZhan:
> Can you please provide a test case that fails before the change and passes after the change?

Test case is checked in, see to IndexerTests.java

sbrannen:
> Test case is checked in, see to IndexerTests.java

Thanks for adding the test!

For future reference, please make sure you run `./gradlew check` before pushing commits, since this PR breaks the build due to a missing blank link in the import section of that test class.

I'll fix that before merging.

sbrannen:
This has been merged into `master` in fde7b1e5454ed169719562234799941d9e409135, and I revised the test and moved it to `PropertyAccessTests` in 734ceed3012162513d0e65b0b31edb87f78bd3ab.

Thanks!

Issue #20653, opened 2017-10-24
Add support for MockRestServiceServer to verify multipart request [SPR-16105]


spring-issuemaster:
**[Haruki Okada](https://jira.spring.io/secure/ViewProfile.jspa?name=ocadaruma)** opened **[SPR-16105](https://jira.spring.io/browse/SPR-16105?redirect=false)** and commented

It would be useful to add RequestMatcher to verify multipart/form-data request.

Such as:

```java
// in MockRestRequestMatchers
public static RequestMatcher multipart(final MultiValueMap<String, byte[]> items) {
	// assert equality
}
```

I would like to submit a pull request later.

---

1 votes, 5 watchers

sbrannen:
This related to gh-23671 and should be assessed in conjunction with that.

rstoyanchev:
This is now superseded by #23772.

PR #23671, opened 2019-09-19
Enhanced form data request support in MockRestServiceServer

spac-valentin:
Added FormDataRequestMatchers to be able to verify expected form parameters when using `MockRestServiceServer`

E.g. 

```java
MockRestServiceServer mockServiceServer = MockRestServiceServer.createServer(restTemplate);

mockServiceServer.expect(ExpectedCount.once(), requestTo("/foobar"))
                .andExpect(method(POST))
                .andExpect(formData().value("foo", "bar"))
                .andExpect(formData().value("baz", Matchers.containsInAnyOrder("fizz", "buzz")))
                .andRespond(withSuccess().location(URI.create("/foobar/123")));
```

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/FormDataRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/FormDataRequestMatchers.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.test.web.client.match;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.hamcrest.Matcher;
+import org.hamcrest.Matchers;
+
+import org.springframework.http.client.ClientHttpRequest;
+import org.springframework.http.converter.FormHttpMessageConverter;
+import org.springframework.mock.http.MockHttpInputMessage;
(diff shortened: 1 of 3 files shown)


sbrannen:
This related to gh-20653 and should be assessed in conjunction with that.

spac-valentin:
Opened PR for gh-20653 (https://github.com/spring-projects/spring-framework/pull/23772) which should address this too

spring-test/src/main/java/org/springframework/test/web/client/match/FormDataRequestMatchers.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.springframework.test.util.AssertionErrors.fail;
+
+/**
+ * Factory for assertions on the form data parameters.
+ *
+ * <p>An instance of this class is typically accessed via {@link MockRestRequestMatchers#formData()}
+ *
+ * @author Valentin Spac
+ * @since 5.2
```
rstoyanchev: This will be 5.3.

spring-test/src/main/java/org/springframework/test/web/client/match/FormDataRequestMatchers.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public class FormDataRequestMatchers {
+
+
+	public RequestMatcher value(String parameter, String... expectedValues) {
+		List<Matcher<? super String>> matcherList = Arrays.stream(expectedValues)
+				.map(Matchers::equalTo)
+				.collect(Collectors.toList());
+
+		return this.value(parameter, matcherList);
```
rstoyanchev: We don't use `this" for methods.

spring-test/src/main/java/org/springframework/test/web/client/match/FormDataRequestMatchers.java
```diff
@@ -0,0 +1,123 @@
...
+	public RequestMatcher value(String parameter, Matcher<Iterable<? extends String>> matchers) {
+		return new AbstractFormDataRequestMatcher() {
+			@Override
+			protected void matchInternal(MultiValueMap<String, String> requestParams) {
+				assertValueCount(parameter, requestParams, 1);
+
+				List<String> values = requestParams.get(parameter);
+				assertThat("Request parameter [" + parameter + "]", values, matchers);
+			}
+		};
```
rstoyanchev: I think these might look better with a lambda + a static method to convert the body to params.

rstoyanchev:
Closing, see https://github.com/spring-projects/spring-framework/pull/23772#issuecomment-645825832.

PR #23676, opened 2019-09-23
Reinitialize SQLErrorCodeSQLExceptionTranslator when initial JDBC connection retrieval failed

### LINKED ISSUE #23675: Incorrect exception type throw by spring jdbc
**Affects:** 5.1.9 and previous versions

Spring-jdbc uses SQLErrorCodeSQLExceptionTranslator with mapping file /spring-jdbc/src/main/resources/org/springframework/jdbc/support/sql-error-codes.xml to map between vendor error codes and relevant Exceptions. E.g.  If underlying database is mysql then vendor error code 1062 gets mapped to DuplicateKeyException. However to be able to identify the vendor itself spring-jdbc invokes JdbcUtils#extractDatabaseMetaData(dataSource, "getDatabaseProductName") which then calls JdbcUtils.extractDatabaseMetaData(DataSource, DatabaseMetaDataCallback). The call JdbcUtils#extractDatabaseMetaData creates a connection using the datasource and then retrieves the database vendor (although vendor name could be available from the driver itself and creating connection may not be necessary). If however creating the connection fails for whatever reason, then SQLErrorCodesFactory will never try to re-initialize SQLErrorCodes and then for the entire life cycle of the application spring will fall back to the SQLExceptionSubclassTranslator.  

This causes spring to throw inconsistent exceptions. E.g. if mysql throws 1062 and SQLErrorCodeSQLExceptionTranslator  is not properly initialized due to connection issue, then SQLExceptionSubclassTranslator will throw DataIntegrityViolationException instead of DuplicateKeyException and this will continue until the application is restarted.

Ideally spring should keep on retrying to discover vendor name if JdbcUtils#extractDatabaseMetaData fails.  This can be very annoying in a cluster where one node with properly initialized SQLErrorCodes  will throw a different exception from a node which is falling back to SQLExceptionSubclassTranslator.

samratdhillon:
closes gh-23675

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLErrorCodeSQLExceptionTranslator.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLErrorCodeSQLExceptionTranslator.java
@@ -78,6 +78,10 @@ public class SQLErrorCodeSQLExceptionTranslator extends AbstractFallbackSQLExcep
 	@Nullable
 	private SQLErrorCodes sqlErrorCodes;
 
+	/** Reference to the dataSource if SQLErrorCodeSQLExceptionTranslator is initialized via new SQLErrorCodeSQLExceptionTranslator(DataSource dataSource).**/
+	@Nullable
+	private DataSource dataSource;
+
 
 	/**
 	 * Constructor for use as a JavaBean.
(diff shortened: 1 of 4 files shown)


snicoll:
@samratdhillon thanks for the PR and sorry it took so long to process it. I believe this has been addressed in the meantime by #25681.

PR #23682, opened 2019-09-23
Fix DataSize class according to ISO/IEC 80000

ez-pzz:
This PR fixes DataSize class according to ISO/IEC 8000 spec (see [Binary prefix](https://en.wikipedia.org/wiki/Binary_prefix))

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/unit/DataSize.java
+++ b/spring-core/src/main/java/org/springframework/util/unit/DataSize.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-core/src/main/java/org/springframework/util/unit/DataSize.java
+++ b/spring-core/src/main/java/org/springframework/util/unit/DataSize.java
@@ -24,11 +24,13 @@ import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
- * A data size, such as '12MB'.
+ * A data size, such as '12MB' or '42MiB'.
  *
- * <p>This class models a size in terms of bytes and is immutable and thread-safe.
+ * <p>This class models a size in terms of bytes (ISO 80000) and is immutable and thread-safe.
+ * <p>Be careful: this class does not work with decimal values.
  *
  * @author Stephane Nicoll
+ * @author Evgeniy Zubakhin
  * @since 5.1
  */
 public final class DataSize implements Comparable<DataSize> {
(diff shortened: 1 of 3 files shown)


spring-core/src/main/java/org/springframework/util/unit/DataSize.java
```diff
@@ -112,6 +133,42 @@ public static DataSize ofTerabytes(long terabytes) {
 		return new DataSize(Math.multiplyExact(terabytes, BYTES_PER_TB));
 	}
 
+	/**
+	 * Obtain a {@link DataSize} representing the specified number of terabytes.
```
smedelyan: Fix doc-comments please :)

sbrannen:
At a glance, this looks like a series of breaking changes in the API that has existed since Spring Framework 5.1, and such breaking changes are not acceptable.

If we were to add ISO 8000 support, it would have to be "in addition" to the existing feature set.

newzubakhin:
Ok, I'll implement something like ISODataSize class with a corresponding converter and tests.
Should I close this PR?

snicoll:
@newzubakhin please hold on. As Sam indicated, the current proposal is a breaking change and we won't and can't change `DataSize` that way. Before you spent more time on the topic, I'd like to get a chance to review the suggestion and the impacts of having two separate `DataSize` types.

snicoll:
Thanks for the PR @newzubakhin. We've discussed this and decided to keep `DataSize` in its current form and clarify its Javadoc, see #23697

sstock:
I understand keeping the existing `DataSize` behavior for backward compatibility.  However I encourage adding a standards compliant variation as suggested by @newzubakhin.  The #23697 documentation certainly helps with direct consumers of the class.  But does not alleviate the need to replicate such clarification to values parsed by `DataSize`, e.g. Spring Boot externalized configuration.  The current behavior creates confusion as [The International System of Units](https://www.bipm.org/en/publications/si-brochure/) (9th edition, section 3 Decimal multiples and sub-multiples of SI units) states "The SI prefixes refer strictly to powers of 10. They should not be used to indicate powers of 2 (for example, one kilobit represents 1000 bits and not 1024 bits)."  This wording is mirrored by NIST's [SI prefixes summary](https://physics.nist.gov/cuu/Units/prefixes.html).

vpavic:
As this was declined due to being a breaking change back in 2019, could you reconsider it for `6.0`?

IMO it's very unfortunate to have a piece of javadoc that basically redefines something as well established as the SI prefixes. If there's a configuration property of type `DataSize` annotated with `@DataSizeUnit(DataUnit.KILOBYTES)` that should be quite self explanatory.

vpavic:
As someone who got burned by this in a project that surfaced `DataSize` through configuration properties, I'd kindly ask the team again to reconsider this. Not respecting (and redefining) the SI prefixes only has the potential to cause confusion, and I can only echo what @sstock already [pointed out](https://github.com/spring-projects/spring-framework/pull/23682#issuecomment-816857856):

> The https://github.com/spring-projects/spring-framework/issues/23697 documentation certainly helps with direct consumers of the class. But does not alleviate the need to replicate such clarification to values parsed by `DataSize`, e.g. Spring Boot externalized configuration.

PR #23690, opened 2019-09-24
Obfuscate IP addresses in build scans

wilkinsona:


sbrannen:
See https://docs.gradle.com/build-scan-plugin/#obfuscating_identifying_data

PR #23702, opened 2019-09-25
Improve exclusions of checkstyleNoHttp task

wilkinsona:
This pull request adds the following exclusions to the source of the `checkstyleNoHttp` task for each project in the build and for `buildSrc`:

- `bin/` directory (created by Eclipse Buildship)
- `.settings/` directory (created by Eclipse)
- `.classpath` file (created by Eclipse)
- `.project` file (created by Eclipse)

An exclusion for the `build/` directory is also configured as the NoHttp plugin does not exclude it by default for `buildSrc`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -406,11 +406,15 @@ configure(rootProject) {
 	nohttp {
 		source.exclude "**/test-output/**"
 		whitelistFile = project.file("src/nohttp/whitelist.lines")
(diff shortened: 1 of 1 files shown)


sbrannen:
This is exceedingly helpful when switching between `5.1.x` and `master`.

Thanks!

PR #23714, opened 2019-09-26
Add ability for StopWatch to aggregate results

Catchwa:
Add configuration option for the StopWatch to aggregate results of
timers with the same name, as opposed to replacing them. To preserve
existing behaviour, this functionality is disabled (set to false) by
default.

Changed the internals of the taskList to use a LinkedHashMap instead
of a LinkedList so as to enable fast lookups of previous timers whilst
also preserving the previous behaviour of ordering the list by
insertion order.

Closes gh-6655

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StopWatch.java
+++ b/spring-core/src/main/java/org/springframework/util/StopWatch.java
@@ -17,8 +17,8 @@
 package org.springframework.util;
 
 import java.text.NumberFormat;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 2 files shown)


Catchwa:
The reason why I think this is useful is for trying to find out bottlenecks in code that is invoked many times (e.g. in loops).

e.g.

```
for (int i = 0; i < LARGE_NUMBER; i++) {
  stopWatch.start("Task A");
  taskA();
  stopWatch.stop();
  stopWatch.start("Task B");
  taskB();
  stopWatch.stop();
}
```

Without this feature, the above code would print out 2 * LARGE_NUMBER lines in pretty-print mode, but with it you'd just get the total for Task A and the total for Task B. The only alternative I can think of is to have two separate StopWatch instances, then do something like:

```
stopWatchA.keepTaskList(false);
stopWatchB.keepTaskList(false);
for (int i = 0; i < LARGE_NUMBER; i++) {
  stopWatchA.start("Task A");
  taskA();
  stopWatchA.stop();
  stopWatchB.start("Task B");
  taskB();
  stopWatchB.stop();
}
```

...then print the total for each StopWatch instance (but then it's more annoying to work out the % duration of each timer as a % of the whole).

bclozel:
There hasn't been much demand for this, so I'm declining this PR.
On top of that, I don't think that using `StopWatch` for performance testing is the right call, as it's limited in many ways. Tools like JMH fixes all those issues and more.

PR #23729, opened 2019-09-30
Ensure spring-core jar task is UP-TO-DATE

facewindu:
- Use the [shadow plugin](https://github.com/johnrengelman/shadow) for repackaging.
  - Using this plugin allows to get rid of Ant JarJar task
  - ~Using `preserveFileTimestamps = false` and `reproducibleFileOrder = true` on all jars ensure consistent checksums whenever the jar are rebuilt.~
- Use a more dogmatic approach to embed the `kotlin-coroutines` classes into `spring-core` jar.
  - Instead of using a `zipTree` of the project default configuration, let the `kotlin-coroutines` project export an artifact containing the output of the kotlin compilation task.

~See [this Gradle Enterprise task inputs comparison](https://ge.spring.io/c/e6m3krl5jfwjk/viakdt3a6d5eu/task-inputs) between the same project checkout, in two different directories. The `spring-core:jar` is not present anymore, because its inputs were the same in both builds.~

~Regarding the new generated `spring-core.jar`.
I checked using `pkgdiff` that its content is byte-to-byte identical to the former one. The MD5 is different though, as the timestamps are not changed anymore when rebuilt (and that's the reason why I changed to the Shadow plugin)~

EDIT: After discussion with @lptr, we agreed that using `preserveFileTimestamps = false` and `reproducibleFileOrder = true` just to get the non-cacheable `:spring-core:jar` task up-to-date is not worth it.
A filter on the task input comparison to focus on _cacheable_ task will probably be added, as a way to filter out the noise produced by non-cacheable tasks

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -307,7 +307,6 @@ configure(allprojects) { project ->
 
 	tasks.withType(AbstractArchiveTask) {
 		preserveFileTimestamps = false
-		reproducibleFileOrder = true
 	}
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -304,10 +304,6 @@ configure(allprojects) { project ->
 			cacheDynamicVersionsFor 0, "seconds"
 		}
 	}
-
-	tasks.withType(AbstractArchiveTask) {
-		preserveFileTimestamps = false
-	}
 }
 
 configure([rootProject] + javaProjects) { project ->
--- a/build.gradle
+++ b/build.gradle
@@ -304,6 +304,11 @@ configure(allprojects) { project ->
 			cacheDynamicVersionsFor 0, "seconds"
 		}
 	}
+
+	tasks.withType(AbstractArchiveTask) {
+		preserveFileTimestamps = false
+		reproducibleFileOrder = true
+	}
 }
 
 configure([rootProject] + javaProjects) { project ->
--- a/spring-core/kotlin-coroutines/kotlin-coroutines.gradle
+++ b/spring-core/kotlin-coroutines/kotlin-coroutines.gradle
@@ -3,10 +3,10 @@ description = "Spring Core Coroutines support"
 apply plugin: "kotlin"
 
 configurations {
-    classesOnlyElements {
-        canBeConsumed = true
-        canBeResolved = false
-    }
+	classesOnlyElements {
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@facewindu Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23729)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23729) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@facewindu Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23729)!

spring-core/kotlin-coroutines/kotlin-coroutines.gradle
```diff
@@ -2,6 +2,17 @@ description = "Spring Core Coroutines support"
 
 apply plugin: "kotlin"
 
+configurations {
+    classesOnlyElements {
```
sbrannen: Please use tabs instead of spaces.

facewindu: fixed

sbrannen: Thanks

facewindu:
I removed the unneeded `reproducibleFileOrder` setting.

build.gradle
```diff
@@ -307,7 +307,6 @@ configure(allprojects) { project ->
 
 	tasks.withType(AbstractArchiveTask) {
 		preserveFileTimestamps = false
-		reproducibleFileOrder = true
 	}
```
lptr: This whole block should be unnecessary. With `@Classpath` and `@CompileClasspath` inputs Gradle normalizes away both the file order and the timestamps in JAR files.

facewindu: I removed the whole block. After discussion, it's not worth doing this just to get the `:spring-core:jar` up-to-date.

snicoll:
@facewindu thank you very much!

PR #23733, opened 2019-09-30
cannot readOnly, developer need add HttpHeader in org.springframework…

qinarmy-zoro:
org.springframework.web.reactive.function.client.DefaultClientRequestBuilder.BodyInserterRequest 
 cannot readOnly, developer need add HttpHeader in org.springframework.http.codec.HttpMessageWriter,eg: add a HttpHeader xxx-bodyOfSignature

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultClientRequestBuilder.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultClientRequestBuilder.java
@@ -17,11 +17,7 @@
 package org.springframework.web.reactive.function.client;
 
 import java.net.URI;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.Consumer;
 
 import org.reactivestreams.Publisher;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@kosmosimonyi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23733)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23733) to manually synchronize the status of this Pull Request.

rstoyanchev:
From the Javadoc of `ClientRequest`:
> Represents a typed, **immutable**, client-side HTTP request

To modify an already prepared request, use `ClientRequest`#from(ClientRequest)`.

kosmosimonyi:
@rstoyanchev WebClient's architecture make developer customize anything difficultly, ClientRequest don't need immutable ,because http message is send by developer, develop should update it . please consider

kosmosimonyi:
@rstoyanchev RestTemplate is very good ,if I want to customize RestTemplate,it easy;
but WebClient very very very difficultly, WebClient lag behind

rstoyanchev:
I don't see the difficulty in mutating the `ClientRequest`. Immutable objects are quite common in functional APIs which in this case is also asynchronous, so you could have filters that make remote calls (e.g. OAuth) before making the actual request. Making `ClientRequest` mutable would give the false impression that if the object can be modified at any time, say if some filter code held onto the request, and that would not be true.

PR #23737, opened 2019-10-01
Code improvement in `for` statement to use lambda expression

ze12augusto:
Code improvement in `for` statement to use lambda expression.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/filter/CompositeFilter.java
+++ b/spring-web/src/main/java/org/springframework/web/filter/CompositeFilter.java
@@ -80,10 +80,7 @@ public class CompositeFilter implements Filter {
 	 */
 	@Override
 	public void destroy() {
-		for (int i = this.filters.size(); i-- > 0;) {
-			Filter filter = this.filters.get(i);
-			filter.destroy();
-		}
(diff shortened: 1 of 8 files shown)


spring-web/src/main/java/org/springframework/web/filter/CompositeFilter.java
```diff
@@ -80,10 +80,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 	 */
 	@Override
 	public void destroy() {
-		for (int i = this.filters.size(); i-- > 0;) {
-			Filter filter = this.filters.get(i);
-			filter.destroy();
-		}
+		this.filters.forEach(Filter::destroy);
```
sbrannen: This is a breaking change, since it no longer invokes the filters in reverse order.

So please revert this.

ze12augusto: Done

sbrannen:
Thanks for the proposal.

After further consideration, we have decided to decline this PR.

Rationale: due to performance reasons, we prefer to use simple for-loops rather than streams in _core_ parts of the framework.

PR #23739, opened 2019-10-01
Fix typo in web-uris doc

drumonii:
In `web-uris.adoc`, there is a typo in a Java code snippet that builds a `URI`: `RI` when it should be `URI`.

This typo is only affecting master (5.2).


sbrannen:

Thanks

Issue #23752, opened 2019-10-02
Pre-flight requests return with error code: 401


rolaca11:
**Affects:** 5.2.0.RC2

---
When sending a POST request to the autoconfigured `/oauth/token` endpoint, most browsers send a preflight request to determine if the actual request has any chance of succeeding.

I have configured an application with [spring-secutiry-oauth2-autoconfigure](https://github.com/spring-projects/spring-security-oauth2-boot), [spring-security-oauth](https://github.com/spring-projects/spring-security-oauth) and the brand new [spring-security](https://github.com/spring-projects/spring-security).

I tried implementing the new implementation of OAuth2 authentication with JWTs, but that doesn't support an authorization server, which I need.

This works fine from _Postman_, if I send a POST request to the right endpoint, I get an access token. However, things get wrong, when I try the same from a browser, because of the above mentioned pre-flight request.

When the browser sends this request, the `/oauth/token` endpoint realises that there are no authentication data sent alongside the request, so it throws an error, Through some error handling path, the request gets to the cors processor, which realises that the original request was in fact a pre-flight one and sets the headers of the response accordingly, except for the status code, which remains 401. The browser, seeing the status code aborts the request with a `Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://localhost:8080/oauth/token. (Reason: CORS preflight channel did not succeed).` message.

I tried, from the debugger, set the status code to 200 manually and in this case, the browser sees it as a green light, to send the actual request, which in turn gives the access token in question.

I'm planning on creating a solution for this problem myself, since it's an easy one. This is the first problem that I'm trying to help solve for the spring framework and I'm really excited to start contributing.

Curious, if I'm getting something wrong. The problem originally occured with the latest stable release of spring boot, the _2.1.7.RELEASE_ version, that's why I'm trying out the milestone version.

sbrannen:
Superseded by #23754

PR #23754, opened 2019-10-02
Ensure pre-flight request does not return error code 401

rolaca11:
Based on issue: https://github.com/spring-projects/spring-framework/issues/23752

Provides a test, which shows exactly in what conditions does the issue arises and provides a fix to that issue.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/cors/DefaultCorsProcessor.java
+++ b/spring-web/src/main/java/org/springframework/web/cors/DefaultCorsProcessor.java
@@ -133,6 +133,7 @@ public class DefaultCorsProcessor implements CorsProcessor {
 		responseHeaders.setAccessControlAllowOrigin(allowOrigin);
 
 		if (preFlightRequest) {
+			response.setStatusCode(HttpStatus.OK);
 			responseHeaders.setAccessControlAllowMethods(allowMethods);
 		}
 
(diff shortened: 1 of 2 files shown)


rstoyanchev:
If I understand correctly from the description under #23752 there is a 401 error during the pre-flight request, and then as part of error handling (in an ERROR dispatch) you want the `DefaultCorsProcessor` to reset the status to 200? 

It doesn't seem right that `DefaultCorsProcessor` should be overriding the response status unconditionally but I'll defer for an opinion from the Spring Security team.

jgrandja:
@rolaca11 Did you see my [comment](https://github.com/spring-projects/spring-security-oauth/pull/1750#issuecomment-534095415)? The flow you are implementing has security issues. Please see my comment for alternative solution.

rstoyanchev:
I'm closing this as I don't think it makes sense for DefaultCorsProcessor to be setting the response status.

rolaca11:
I'm sorry, didn't really have time to respond, was busy. The reason why the
DefaultCorsProcessor should set the status code is because in some cases
preflight requests are sent to a resource that requires an authenticated
user to access. However, preflight requests are not sent with
Authentication headers.

In this case, the AffirmationVoter (or whichever is active) will of course
deny access and set the status code to 401. Understandable, since the
request is not authenticated.

But this status code is inherited, when the request gets to the cors
processor, so the sender of the preflight request sees that the given
resource is not accessible because of the status code, when in reality, the
client has a valid token, it just doesn't send it for a preflight request.

Maybe my solution is too cut in stone and would require a bit more
configurability tho. I'm down to helping you guys, I just have a busy week
behind and towards me.



Rossen Stoyanchev <notifications@github.com> ezt írta (időpont: 2019. nov.
8., P 16:37):

rstoyanchev:
I do not consider this something to be resolved on the Spring Framework side at this point. Please, see the comment and link by @jgrandja above and follow through on that discussion rather than opening additional tickets.

PR #23756, opened 2019-10-04
Introduce sessionAttributeDoesNotExist in RequestResultMatchers

drumonii:
Given the matcher `attributeDoesNotExist` of `ModelResultMatchers`, it makes sense for a session attribute `ResultMatcher` equivalent – `sessionAttributeDoesNotExist` which asserts that the given session attributes are `null` in the `HttpSession`.


sbrannen:
For anyone interested in achieving the same goal prior to Spring Framework 5.2.1, please note that this is already possible for a single attribute name using a Hamcrest `Matcher` as follows.

```java
this.mockMvc.perform(get("/"))
	.andExpect(request().sessionAttribute("bogus", is(nullValue())))
```

Issue #22788, opened 2019-04-12
Deprecate MediaType.APPLICATION_JSON_UTF8 in favor of APPLICATION_JSON


reschke:
In  spring-framework/spring-web/src/main/java/org/springframework/http/MediaType.java:

 	 * <p>This {@link MediaType#APPLICATION_JSON} variant should be used to set JSON
	 * content type because while
	 * <a href="https://tools.ietf.org/html/rfc7159#section-11">RFC7159</a>
	 * clearly states that "no charset parameter is defined for this registration", some
	 * browsers require it for interpreting correctly UTF-8 special characters

1. RFC 7159 has been obsoleted by RFC 8259
2. That states "No "charset" parameter is defined for this registration. Adding one really has no effect on compliant recipients." (at the end of Section 11)
3. ...and browsers have indeed fixed their code.

So it would be best to clarify the text and deprecate the constant.

sdeleuze:
Indeed, the related [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=438464) discussed in #18178 is fixed since September 2017, and I have been able to validate that UTF-8 characters are now rendered correctly with `Content-Type: application/json`, so I am ok to switch back to `MediaType.APPLICATION_JSON`, deprecate `MediaType.APPLICATION_JSON_UTF8` and update the documentation accordingly in Spring Framework 5.2.

sdeleuze:
It is a shame, but Safari still has this bug so we should decide what to do ...

sdeleuze:
I have created https://bugs.webkit.org/show_bug.cgi?id=197369 on Safari bugtracker, let's see what is their feedback.

sdeleuze:
The Safari bug is only impacting the display in the main frame, not via `XMLHttpRequest`, so I think we can move forward and merge that in `master`.

reschke:
> The Safari bug is only impacting the display in the main frame, not via XMLHttpRequest, ...

FWIW, I believe the same way true for the matching bugs in Chromium and Firefox.

PR #23764, opened 2019-10-07
Switch default to UTF-8 in RestClientResponseException

### LINKED ISSUE #23738: Consider defaulting to UTF-8 in RestClientResponseException
Spring version: 5.2.0.RELEASE

Related to https://github.com/spring-projects/spring-framework/issues/22954 and the ```UTF-8``` charset removal, the ```RestClientResponseException``` now defaults to ```ISO_8859_1``` here https://github.com/spring-projects/spring-framework/blob/3a0f309e2c9fdbbf7fb2d348be861528177f8555/spring-web/src/main/java/org/springframework/web/client/RestClientResponseException.java#L105
instead of ```UTF-8``` previously.

To mitigate this issue, we are now forced to create a custom ```ResponseErrorHandler``` and forcing ```UTF-8```

ask4gilles:
closes https://github.com/spring-projects/spring-framework/issues/23738

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestClientResponseException.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestClientResponseException.java
@@ -33,7 +33,7 @@ public class RestClientResponseException extends RestClientException {
 
 	private static final long serialVersionUID = -8803556342728481792L;
 
-	private static final Charset DEFAULT_CHARSET = StandardCharsets.ISO_8859_1;
+	private static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
 
 
 	private final int rawStatusCode;


ask4gilles:
Hi @sbrannen and @jhoeller ,
Sorry to ping but this one is a one liner which would prevent us writing a custom error handler, so if you would have a minute to have a look :) Thanks!

rstoyanchev:
Note that the same decision also applies to the WebClient, in `DefaultClientResponse#createException()` which then determines the encoding for resulting `WebClientResponseException`. 

@sdeleuze, @poutsma, following the change in #22788, JSON responses no longer specify a charset as per RFC 7159 defaulting to UTF-8. However client exceptions use ISO-8859-1 if no charset is specified. I suggest we switch the default to UTF-8 and also add  an overloaded `getResponseBodyAsString(Charset)` to `RestClientResponseException` and `WebClientResponseException` for additional control. WDYT?

rstoyanchev:
@ask4gilles please refrain from pings on issues like that. What may be a one-liner for you could become a time sink for others and for us if not carefully considered.

PR #23772, opened 2019-10-09
Multipart request support in MockRestServiceServer

spac-valentin:
Introduce `MultipartFormDataRequestMatchers` for `MockRestServiceServer`. Handles **ONLY** `multipart/form-data` requests.

E.g.
```
MockRestServiceServer mockServiceServer = MockRestServiceServer.createServer(restTemplate);
MultiValueMap<String, Resource> filesMultiMap = new LinkedMultiValueMap<>();
filesMultiMap.add(foo.getName(), foo.getResource());
filesMultiMap.add(bar.getName(), bar.getResource());

MultiValueMap<String, Object> paramsMultiMap = new LinkedMultiValueMap<>();
paramsMultiMap.add("fooParam", "foo value");
paramsMultiMap.add("barParam", "bar value");

mockServiceServer.expect(ExpectedCount.once(), requestTo("/foobar"))
                .andExpect(method(POST))
                .andExpect(content().multipart().param("foo", "bar"))
                .andExpect(content().multipart().param("foo", "bar", "baz"))
                .andExpect(content().multipart().params(paramsMultiMap))

                .andExpect(content().multipart().file("fooFile",  foo.getBytes()))
                .andExpect(content().multipart().file("fooBarFile",  fooResource, barResource))
                .andExpect(content().multipart().file("bazFile",  customResourceMatcher))
                .andExpect(content().multipart().files("fooBarFile",  filesMultiMap))

                .andRespond(withSuccess().location(URI.create("/foobar/123")));
```

Related to https://github.com/spring-projects/spring-framework/pull/23671 which handles `x-www-form-urlencoded` requests

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
@@ -151,13 +151,15 @@ public class MultipartFormDataRequestMatchers {
 	}
 
 
-	private void assertByteArrayMatch(String parameterName, List<byte[]> expectedFiles, List<MultipartFile> actualFiles) {
+	private void assertByteArrayMatch(String parameterName, List<byte[]> expectedFiles,
+									  List<MultipartFile> actualFiles) {
 		for (int index = 0; index < actualFiles.size(); index++) {
 			MultipartFile multiPartFile = actualFiles.get(index);
 			byte[] expectedContent = expectedFiles.get(index);
 
 			try {
-				assertEquals("Content mismatch for file " + parameterName, expectedContent, multiPartFile.getBytes());
+				assertEquals("Content mismatch for file " + parameterName, expectedContent,
+						multiPartFile.getBytes());
 			}
 			catch (IOException ex) {
 				throw new AssertionError("Could not get bytes from actual multipart files", ex);
@@ -173,7 +175,8 @@ public class MultipartFormDataRequestMatchers {
 			try {
 				byte[] fileContent = IOUtils.toByteArray(expectedResource.getInputStream());
 
-				assertEquals("Content mismatch for file " + parameterName, fileContent, multiPartFile.getBytes());
+				assertEquals("Content mismatch for file " + parameterName, fileContent,
+						multiPartFile.getBytes());
 				assertEquals("Filename ", expectedResource.getFilename(), multiPartFile.getOriginalFilename());
 			}
 			catch (IOException ex) {
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
@@ -89,7 +89,7 @@ public class MultipartFormDataRequestMatchers {
 
 				String[] values = requestParams.get(parameter);
 
-				Assert.state(values != null, "No request parameter values");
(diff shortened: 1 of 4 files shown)


spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
```diff
@@ -0,0 +1,252 @@
...
+import static org.springframework.test.util.AssertionErrors.assertEquals;
+import static org.springframework.test.util.AssertionErrors.fail;
+
+/**
+ * Factory for assertions on the multipart form data parameters. Handles only {@link MediaType#MULTIPART_FORM_DATA}
+ *
+ * <p>An instance of this class is typically accessed via {@link ContentRequestMatchers#multipart()}
+ *
+ * @author Valentin Spac
+ * @since 5.2
```
rstoyanchev: This will be 5.3.

spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
```diff
@@ -0,0 +1,252 @@
...
+			try {
+				assertEquals("Content mismatch for file " + parameterName, expectedContent, multiPartFile.getBytes());
+			}
+			catch (IOException ex) {
+				throw new AssertionError("Could not get bytes from actual multipart files", ex);
+			}
+		}
+	}
+
+	private void assertResourceMatch(String parameterName, List<Resource> expectedFiles, List<MultipartFile> actualFiles) {
```
rstoyanchev: This does not adhere to line length rules https://github.com/spring-projects/spring-framework/wiki/Code-Style#line-wrapping.

spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
```diff
@@ -0,0 +1,252 @@
...
+		return new AbstractFormDataRequestMatcher() {
+			@Override
+			protected void matchInternal(MultipartHttpServletRequest request) {
+				Map<String, String[]> requestParams = request.getParameterMap();
+				assertValueCount(parameter, requestParams, 1);
+
+				String[] values = requestParams.get(parameter);
+				assertThat("Request parameter [" + parameter + "]", Arrays.asList(values), matchers);
+			}
+		};
```
rstoyanchev: This might look better if the abstract class is replaced with a lambda that calls a static method to parse the multipart content.

anchaoc:
666666666666666666

spring-test/src/main/java/org/springframework/test/web/client/match/MultipartFormDataRequestMatchers.java
spac-valentin: Agree. Please see latest changes

rstoyanchev:
I have processed these changes and scaled back the solution a bit. I've kept 2 multipart methods, one with `MultiValueMap<String, ?>` for exact matching and one with `Map<String, ?>` for a contains match. This meets the requirements of the original request #20653. It is how the `RestTemplate` takes input for multipart requests, and is consistent the existing matcher for form data.

As for form data, i.e. the other PR # 23671, I've added a similar new method with a `Map<String, String>` for a contains match. It's next to the existing one in `ContentRequestMatchers` rather than at the top level `MockRestRequestMatchers`.

In any case thanks for the contributions.

Issue #22325, opened 2019-01-30
ApplicationContext.refresh() causes stale listeners to be added to ApplicationEventMulticaster 


eknizat:
The problem is caused by the change in Spring AbstractApplicationContext, in particular how it handles the addition of ApplicationListener objects.
In short, all of the registered ApplicationListener objects are retained even after application context is refreshed, 
at which point the stale listeners become unusable as AbstractLifecycleBean(s) would have had their destroy methods invoked.

The change that caused this problem:
https://github.com/spring-projects/spring-framework/issues/21858
Commit:
https://github.com/spring-projects/spring-framework/commit/c8c0737ce712569a7e03a97dfe7ef11cbbc33e39

With this commit all listeners are added to AbstractApplicationContext's list applicationListeners instead of just adding them to an instance of ApplicationEventMulticaster.
This causes stale listeners to be added to a new instance of ApplicationEventMulticaster when the AbstractApplicationContext is refreshed via its refresh() method.
The refresh() method creates a new ApplicationEventMulticaster but then it adds all of the stale listeners (by registerListeners()) from its list of listeners.
Subsequently, when the refresh finishes (AbstractApplicationContext.finishRefresh()), a new instance of ApplicationEventMulticaster publishes the refresh event too all registered listeners, including the stale listeners whose destroy methods were called as part of the refresh.

jhoeller:
@wilkinsona since this one addresses a side effect of #21858 which you requested, could you please double-check that the revision (in master and 5.1.x at this point) still works for your purposes, exposing all application listeners that you intend to see from `getApplicationListeners()`? I'll backport it to 5.0.x then where we backported the original change as well.

wilkinsona:
This looks good to me with 5.1.5 snapshots. The sample provided with https://github.com/spring-projects/spring-boot/issues/14490 stops as expected.

jhoeller:
Alright, good to hear. Thanks for the immediate feedback!

jhoeller:
@eknizat since this is available in `5.1.5.BUILD-SNAPSHOT` from https://repo.spring.io/snapshot now, do you have a chance to give it an early try in your scenario before the official 5.1.5 release on Wednesday?

eknizat:
The 5.1.5 snapshot works as expected. Thanks.

rom16rus:
[...]

PR #23782, opened 2019-10-11
Fixing 22325 issue application listeners logic

platonovr:
There was an issue #22325 and @jhoeller has fixed it, but now I have the following problem: on the application startup there is a call of prepareRefresh(), so it works like that
`this.earlyApplicationListeners = new LinkedHashSet<>(this.applicationListeners);`
`this.applicationListeners` is an empty collection.
Then it's needed to refresh the Application Context, app provokes the refresh() at AbstractApplicationContext, then Spring calls prepareRefresh and then checks
`this.earlyApplicationListeners == null,`
but it's an empty LinkedHashSet, not null, it's zero size set, so then it comes to
`this.applicationListeners.clear();`
and all the listeners are removed, but they were created, so I'm loosing all of applicationListeners there.
Application listeners collection was filled by CXF Servlet [code](https://github.com/apache/cxf/blob/ec4423437bd7d6b63e3464761394cfafb50cacfa/rt/transports/http/src/main/java/org/apache/cxf/transport/servlet/CXFServlet.java#L90)

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
+++ b/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
@@ -604,7 +604,7 @@ public abstract class AbstractApplicationContext extends DefaultResourceLoader
 		getEnvironment().validateRequiredProperties();
 
 		// Store pre-refresh ApplicationListeners...
-		if (this.earlyApplicationListeners == null) {
+		if (this.earlyApplicationListeners == null || this.earlyApplicationListeners.isEmpty()) {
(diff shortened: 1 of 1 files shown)


snicoll:
@platonovr unfortunately it's hard for us to judge the problem at hand without test cases that demonstrate the problem you've described. Looking at the code you've referenced, you should call `addApplicationListener` rather than the reflection-based code. 

Please amend your PR with some tests that are failing before your change.

snicoll:
I am going to close this now for the reason outlined above. If the feedback is provided, we can reopen this PR and continue the review.

PR #23784, opened 2019-10-11
ApplicationListenerMethodAdapter: gracefully handle beans which are actually NullBean

stevenschlansker:
Currently, if you have an optional event listener (via a @Bean method returning `null`)
this causes the event multicaster to explode violently.  Now, we just safely skip it.


stevenschlansker:
Any comments on this fix?  We are eagerly anticipating having our code work as expected, and it seems unlikely to cause a regression.

stevenschlansker:
Thank you!

PR #23793, opened 2019-10-14
Debug logging for WebSocket lifecycle in WebSocketStompClient

alexandrumm:
@pivotal-issuemaster This is an Obvious Fix

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/messaging/WebSocketStompClient.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/messaging/WebSocketStompClient.java
@@ -57,6 +57,7 @@ import org.springframework.web.socket.WebSocketHttpHeaders;
 import org.springframework.web.socket.WebSocketMessage;
 import org.springframework.web.socket.WebSocketSession;
 import org.springframework.web.socket.client.WebSocketClient;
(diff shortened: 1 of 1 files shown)


rstoyanchev:
I think this should be done by decorating the WebSocketHandler (i.e. `WebSocketTcpConnectionHandlerAdapter`) with `LoggingWebSocketHandlerDecorator` consistent with what we do in other places like `WebSocketConnectionManager` and `WebSocketHttpRequestHandler`.

alexandrumm:
Hello, @rstoyanchev. 
First of all thanks for responding and taking care of this PR. 
I have a question with regards to your suggestion, `WebSocketTcpConnectionHandlerAdapter` it is also implementing the interface `ListenableFutureCallback` and the `LoggingWebSocketHandlerDecorator` does not expect or suppose, I think, to cover this type of handler, so could you please give me a hint here on how to proceed?

Thanks in advance!

rstoyanchev:
This should be fine:
```java
WebSocketTcpConnectionHandlerAdapter adapter =  new WebSocketTcpConnectionHandlerAdapter(session);
getWebSocketClient()
        .doHandshake(new LoggingWebSocketHandlerDecorator(adapter), handshakeHeaders, url)
        .addCallback(adapter);
```

alexandrumm:
Right, I missed the obvious solution there. Thanks a lot! :)

PR #23798, opened 2019-10-15
Updates Tomcat to 9.0.27

mgmeiner:
Tomcat should be updated to version 9.0.27 as it cannot load a ssl key-store without exception. 

See: [Stackoverflow Question](https://stackoverflow.com/questions/58189855/spring-boot-2-embed-tomcat-9-0-26-can-not-load-jks-file-stream-closed)

Affects all spring-framework 5.x versions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,14 +137,14 @@ configure(allprojects) { project ->
 			dependency "org.webjars:webjars-locator-core:0.37"
 			dependency "org.webjars:underscorejs:1.8.3"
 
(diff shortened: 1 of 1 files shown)


kowsercse (approved):

bclozel:
Spring Framework is getting Tomcat as an optional dependency and does not expose it on its runtime classpath nor as a managed dependency.
This change won’t solve the issue for Spring Framework consumers.

snicoll:
Agreed and we prefer to do dependency updates in due course. Thanks for the PR anyway.

Issue #23792, opened 2019-10-14
Autowiring performance degradation due to 5.2's MethodParameter.getParameterType() implementation


e-hubert:
Our applications makes considerable use of programmatic autowiring of beans via org.springframework.beans.factory.config.AutowireCapableBeanFactory#autowire.

Therefore we have a couple of performance tests, which directly and indirectly measure the throughput of the underlying Spring implementation.
All those tests degraded by about 20 to 25 percent, after only updating the Spring dependencies from version 5.1.9 to version 5.2.0.

We did a bit of work to pinpoint the changes in Spring, which resulted in this performance drop and thought it would be a good idea to report it here so you can verify, whether you like to accept this or have an idea how to achieve both, improved code robustness while maintaining performance.

The root of the actual performance degradation seems to originate from changes to org.springframework.core.MethodParameter#getParameterType to use org.springframework.core.ResolvableType (in our case itself used from org.springframework.beans.factory.support.ConstructorResolver#resolveAutowiredArgument).

The relevant issue motivating this change seems to be https://github.com/spring-projects/spring-framework/issues/23385

I quickly verified that it is enough to use 5.1.9 as base and only change the above mentioned two core classes, to observe the performance drop. I have not yet checked the Spring project for own performance tests.

lgxbslgx:
I submit a PR to improve the performance of the method MethodParameter.getParameterType. Hope it could help.

quaff:
I have the same issue. my web application built with both struts2 and spring mvc, after upgrade to spring 5.2.0, throughput of struts2 decrease to about 80%, but throughput of spring mvc is increased. struts2 action is autowired by `AutowireCapableBeanFactory::createBean(clazz, AUTOWIRE_BY_NAME, false)` .

BTW: @lgxbslgx's PR not helps much.

quaff:
@e-hubert can you provide a simple project and try the PR https://github.com/spring-projects/spring-framework/pull/23811 ?

jhoeller:
According to my local benchmarks, it seems sufficient to restore a `containingClass != null` check as we effectively had it in the 5.1.x code path. Refining it with a check for `containingClass != declaringClass` does not hurt but seems like an optimization beyond 5.1.x already, nevertheless worth it. I'll double-check the effect and will proceed with that change for inclusion in 5.2.1.

jhoeller:
[...]

PR #23811, opened 2019-10-15
Improve the performance of the method MethodParameter.getParameterType

lgxbslgx:
The method ResolvableType.forMethodParameter can be unused in some cases.
See #23792.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -486,7 +486,13 @@ public class MethodParameter {
 		if (paramType != null) {
 			return paramType;
 		}
-		paramType = ResolvableType.forMethodParameter(this, null, 1).resolve();
+		paramType = computeParameterType();
+		// In order to improve performance, the method forMethodParameter should not be used in some cases.
+		if ("java.lang.Object".equals(paramType.getCanonicalName()) &&
+				!getDeclaringClass().equals(getContainingClass()) &&
+				!"java.lang.Object".equals(getGenericParameterType().getClass().getCanonicalName())) {
+			paramType = ResolvableType.forMethodParameter(this, null, 1).resolve();
+		}
 		if (paramType == null) {
 			paramType = computeParameterType();
 		}


spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -486,7 +486,13 @@ void setParameterType(@Nullable Class<?> parameterType) {
 		if (paramType != null) {
 			return paramType;
 		}
-		paramType = ResolvableType.forMethodParameter(this, null, 1).resolve();
+		paramType = computeParameterType();
+		// In order to improve performance, the method forMethodParameter should not be used in some cases.
+		if ("java.lang.Object".equals(paramType.getCanonicalName()) &&
```
quaff: It's better to use `Object.class.getName()`

spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -486,7 +486,13 @@ void setParameterType(@Nullable Class<?> parameterType) {
...
+				!getDeclaringClass().equals(getContainingClass()) &&
+				!"java.lang.Object".equals(getGenericParameterType().getClass().getCanonicalName())) {
```
kowsercse: Condition in the _if_ block is not readable. Extracting following code to a method with a meaningful name should be more readable.

> "java.lang.Object".equals(paramType.getCanonicalName()) &&
				!getDeclaringClass().equals(getContainingClass()) &&
				!"java.lang.Object".equals(getGenericParameterType().getClass().getCanonicalName())

kowsercse (changes requested):

jhoeller:
See my comment on the original issue: I intend to go with a simpler check along the lines of the effective behavior in 5.1.x, just checking for a containing class different from the declaring class. I've got this working locally, showing a significant enough performance difference in a microbenchmark.

As for this PR, a specific check for `java.lang.Object` doesn't seem to be ideal since an unresolved generic can expose its upper bound (which may be any type) as the plain parameter type from reflection. Also, resolving `getGenericParameterType` can be expensive on its own when just done to prevent some other expensive computation. Overall, restoring behavior along the lines of 5.1.x seems like a good enough compromise. Thanks for the PR, in any case!

PR #23815, opened 2019-10-16
Changed the default character encoding to UTF-8 in both the exceptions

### LINKED ISSUE #23803: Add option to use charset in RestClientResponseException
Partial backport of gh-23764. 

The default is not switched to UTF-8. However an additional convenience method is provided with a charset argument.

Sauhardstark:
Also added an overload method for clients to get the response body in a different character encoding. Closes [Issue 23803](https://github.com/spring-projects/spring-framework/issues/23803)


rstoyanchev:
Duplicate of #23764. I realize you updated both exceptions but it isn't helpful to have multiple PRs for the same thing.

PR #23835, opened 2019-10-19
Handle %2F in simp messaging usernames [cancelled to recreate based on CONTRIBUTING.md]

njlaw:
### Current State
Since messaging destinations are delimited by "/", the existing code performs a simple string replacement of "/" -> "%2F" in usernames.  This allows usernames which contain "/" to be used and have messaging destinations parsed correctly.  See [SimpMessagingTemplate.java](https://github.com/spring-projects/spring-framework/blob/243f2890ee9e98c97c8dc7278f033def3bf33f86/spring-messaging/src/main/java/org/springframework/messaging/simp/SimpMessagingTemplate.java#L227) and [DefaultUserDestinationResolver.java](https://github.com/spring-projects/spring-framework/blob/243f2890ee9e98c97c8dc7278f033def3bf33f86/spring-messaging/src/main/java/org/springframework/messaging/simp/user/DefaultUserDestinationResolver.java#L217)

### Issue with Current State
Usernames that already contain "%2F" end up with that character sequence converted to a "/" by DefaultUserDestinationResolver.  E.g., `abc%2Fabc` > `abc/abc`; however, `abc/abc` is not the username, so destination matching fails.

[Here is an example test that fails](https://github.com/njlaw/spring-framework/commit/6683c795465f6ec2a91443549357a05a50a910db), resulting in

```
org.springframework.messaging.simp.user.DefaultUserDestinationResolverTests > handleMessageEncodedUserNameWithPercentTwoEff() FAILED
--
org.opentest4j.AssertionFailedError at DefaultUserDestinationResolverTests.java:205
 
717 tests completed, 1 failed
```
([Gradle build scan](https://gradle.com/s/rbezilmjwrhhi))

### Possible Solutions
1. Use a custom encoding scheme that accounts both for "/" and whatever "/" encodes to in the scheme.
2. Use a known encoding scheme like application/x-www-form-urlencoded or URL-safe Base64 ([RFC 4648](https://tools.ietf.org/html/rfc4648#section-5)) and either:
  (a) Encode all usernames
  (b) Encode only usernames that contain characters that need to be encoded and flag the username as encoded so that only encoded usernames are decoded

### Sample Solution
A sample solution is included in this pull request using 2(b).

### Outstanding Questions
Should an explicit encoding be provided for String > byte[] and byte[] > String?  If so, what?  UTF-8?  Or just let the behavior be undefined if running Java in an environment where the default encoding does not support a particular character?

### Other Alternatives

- Document "%2F" as not supported in a username and add a runtime check for invalid user names instead of the current behavior of just not matching message destinations correctly

Any feedback is appreciated!

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/SimpMessagingTemplate.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/SimpMessagingTemplate.java
@@ -31,7 +31,7 @@ import org.springframework.messaging.support.MessageHeaderAccessor;
 import org.springframework.messaging.support.MessageHeaderInitializer;
 import org.springframework.messaging.support.NativeMessageHeaderAccessor;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
+import org.springframework.util.Base64Utils;
 
 /**
  * An implementation of
@@ -224,7 +224,7 @@ public class SimpMessagingTemplate extends AbstractMessageSendingTemplate<String
 			throws MessagingException {
 
 		Assert.notNull(user, "User must not be null");
-		user = StringUtils.replace(user, "/", "%2F");
+		user = Base64Utils.encodeToUrlSafeString(user.getBytes());
 		destination = destination.startsWith("/") ? destination : "/" + destination;
 		super.convertAndSend(this.destinationPrefix + user + destination, payload, headers, postProcessor);
 	}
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@njlaw Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23835)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23835) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@njlaw Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23835)!

PR #23836, opened 2019-10-19
Reject "%2F" as an invalid sequence in simp messaging usernames

njlaw:
### Current State
Since messaging destinations are delimited by "/", the existing code performs a simple string replacement of "/" -> "%2F" in usernames.  This allows usernames which contain "/" to be used and have messaging destinations parsed correctly.  See [SimpMessagingTemplate.java](https://github.com/spring-projects/spring-framework/blob/243f2890ee9e98c97c8dc7278f033def3bf33f86/spring-messaging/src/main/java/org/springframework/messaging/simp/SimpMessagingTemplate.java#L227) and [DefaultUserDestinationResolver.java](https://github.com/spring-projects/spring-framework/blob/243f2890ee9e98c97c8dc7278f033def3bf33f86/spring-messaging/src/main/java/org/springframework/messaging/simp/user/DefaultUserDestinationResolver.java#L217)

### Issue with Current State
Usernames that already contain "%2F" end up with that character sequence converted to a "/" by DefaultUserDestinationResolver.  E.g., `abc%2Fabc` > `abc/abc`; however, `abc/abc` is not the username, so destination matching fails.

[Here is an example test that fails](https://github.com/njlaw/spring-framework/commit/6683c795465f6ec2a91443549357a05a50a910db), resulting in

```
org.springframework.messaging.simp.user.DefaultUserDestinationResolverTests > handleMessageEncodedUserNameWithPercentTwoEff() FAILED
--
org.opentest4j.AssertionFailedError at DefaultUserDestinationResolverTests.java:205
 
717 tests completed, 1 failed

:spring-messaging:testorg.springframework.messaging.simp.user.DefaultUserDestinationResolverTests » handleMessageEncodedUserNameWithPercentTwoEff() (0.044s)
org.opentest4j.AssertionFailedError: 

Expecting:
 <0> (instead resolves to 0 destinations)
to be equal to:
 <1> (should resolve to a single destination)
but was not.
```
([Gradle build scan](https://gradle.com/s/rbezilmjwrhhi))

### Possible Solutions
1. Use a custom encoding scheme that accounts both for "/" and whatever "/" encodes to in the scheme.
2. Use a known encoding scheme like application/x-www-form-urlencoded or URL-safe Base64 ([RFC 4648](https://tools.ietf.org/html/rfc4648#section-5)) and either:
  (a) Encode all usernames
  (b) Encode only usernames that contain characters that need to be encoded and flag the username as encoded so that only encoded usernames are decoded

### Sample Solution
A sample solution is included in this pull request using 2(b).

### Outstanding Questions
Should an explicit encoding be provided for String > byte[] and byte[] > String?  If so, what?  UTF-8?  Or just let the behavior be undefined if running Java in an environment where the default encoding does not support a particular character?

### Other Alternatives

- Document "%2F" as not supported in a username and add a runtime check for invalid user names instead of the current behavior of just not matching message destinations correctly

Any feedback is appreciated!

As a note, I recreated this pull request since I didn't branch my fork of the code first or squash commits as recommended by the contribution guidelines.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/SimpMessagingTemplate.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/SimpMessagingTemplate.java
@@ -31,7 +31,7 @@ import org.springframework.messaging.support.MessageHeaderAccessor;
 import org.springframework.messaging.support.MessageHeaderInitializer;
 import org.springframework.messaging.support.NativeMessageHeaderAccessor;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
+import org.springframework.util.Base64Utils;
 
 /**
  * An implementation of
@@ -224,7 +224,8 @@ public class SimpMessagingTemplate extends AbstractMessageSendingTemplate<String
 			throws MessagingException {
 
 		Assert.notNull(user, "User must not be null");
-		user = StringUtils.replace(user, "/", "%2F");
+		if (user.startsWith("B64:") || user.contains("/"))
+			user = "B64:" + Base64Utils.encodeToUrlSafeString(user.getBytes());
 		destination = destination.startsWith("/") ? destination : "/" + destination;
 		super.convertAndSend(this.destinationPrefix + user + destination, payload, headers, postProcessor);
 	}
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/user/DefaultUserDestinationResolver.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/user/DefaultUserDestinationResolver.java
@@ -31,7 +31,7 @@ import org.springframework.messaging.simp.SimpMessageHeaderAccessor;
 import org.springframework.messaging.simp.SimpMessageType;
 import org.springframework.util.Assert;
 import org.springframework.util.PathMatcher;
-import org.springframework.util.StringUtils;
+import org.springframework.util.Base64Utils;
 
 /**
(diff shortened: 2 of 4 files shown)


rstoyanchev:
Hi, thanks for raising the issue. I would prefer to reject `%2F` as an invalid sequence. I presume you have such a case. Are those generated names or how did they end up with that?

njlaw:
I agree that rejecting `%2F` as invalid is a reasonable option since it is almost definitely a corner case.

In our case it came up because the site we are developing is multi-tenanted and we want the username to represent a unique user-tenant combination, so the username in Spring is a combination of the tenant’s OAuth2 issuer and their subject.  The issuer includes slashes and may contain our chosen delimiter, so we were URL encoding it before appending the delimiter and subject.

I’m not very familiar with the codebase here, but if you can point me in the right direction, I’m glad to add a rejection message or assertion for the character sequence `%2F`.

rstoyanchev:
Okay thanks for elaborating. 

> if you can point me in the right direction, I’m glad to add a rejection message or assertion for the character sequence %2F.

This could probably be rejected [at the point](https://github.com/spring-projects/spring-framework/blob/64dfa462a6354bb0dc3240ca9be68ce37e695c17/spring-messaging/src/main/java/org/springframework/messaging/simp/user/DefaultUserDestinationResolver.java#L192) of trying to subscribe to a user destination. It could also be checked and rejected when trying to send a message in `SimpMessagingTemplate` just before replacing "/" with "%2F".

PR #23840, opened 2019-10-21
Make Kotlin functions accessible in CoroutinesUtils

mgmeiner:
After updating to spring-webflux 5.2 i noticed that `@ExceptionHandler`'s in private Kotlin-Classes stopped working. 
After debbuging I found that this regression was introduced with kotlin-coroutines support for reactive Webflux and reactive Messaging handlers. 

The Problem is that when converting Java methods to Kotlin functions the information if a method is accessible or not will not be copied and the Kotlin function has to be made accessible again via its `isAccessible` setter.

For this I introduced a new very simple `KotlinReflectionUtils` class which both `InvocableHandlerMethod's` use when Kotlin is available. I'm not 100% sure if this is the best solution and if its worth to introduce a new utils class for this.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/handler/invocation/reactive/InvocableHandlerMethod.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/handler/invocation/reactive/InvocableHandlerMethod.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Stream;
 
+import kotlin.reflect.KFunction;
 import reactor.core.publisher.Mono;
 
 import org.springframework.core.CoroutinesUtils;
@@ -37,6 +38,7 @@ import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.handler.HandlerMethod;
 import org.springframework.messaging.handler.invocation.MethodArgumentResolutionException;
+import org.springframework.util.KotlinReflectionUtils;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.ReflectionUtils;
 
@@ -129,11 +131,13 @@ public class InvocableHandlerMethod extends HandlerMethod {
 			Object value;
 			try {
 				Method method = getBridgedMethod();
-				ReflectionUtils.makeAccessible(method);
 				if (KotlinDetector.isKotlinReflectPresent() && KotlinDetector.isKotlinType(method.getDeclaringClass())) {
-					value = CoroutinesUtils.invokeSuspendingFunction(method, getBean(), args);
+					KFunction<?> function = KotlinReflectionUtils.methodToFunction(method);
+					KotlinReflectionUtils.makeFunctionAccessible(function);
+					value = CoroutinesUtils.invokeSuspendingFunction(function, getBean(), args);
 				}
 				else {
+					ReflectionUtils.makeAccessible(method);
 					value = method.invoke(getBean(), args);
 				}
 			}
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/InvocableHandlerMethod.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/InvocableHandlerMethod.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Stream;
 
+import kotlin.reflect.KFunction;
 import reactor.core.publisher.Mono;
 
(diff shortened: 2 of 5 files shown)


spring-messaging/src/main/java/org/springframework/messaging/handler/invocation/reactive/InvocableHandlerMethod.java
```diff
@@ -129,11 +131,13 @@ public void setReactiveAdapterRegistry(ReactiveAdapterRegistry registry) {
...
 				Method method = getBridgedMethod();
-				ReflectionUtils.makeAccessible(method);
 				if (KotlinDetector.isKotlinReflectPresent() && KotlinDetector.isKotlinType(method.getDeclaringClass())) {
-					value = CoroutinesUtils.invokeSuspendingFunction(method, getBean(), args);
+					KFunction<?> function = KotlinReflectionUtils.methodToFunction(method);
+					KotlinReflectionUtils.makeFunctionAccessible(function);
+					value = CoroutinesUtils.invokeSuspendingFunction(function, getBean(), args);
 				}
 				else {
+					ReflectionUtils.makeAccessible(method);
```
rstoyanchev, line 140: The code for this has moved to the constructor of `HandlerMethod` so this needs to be removed from here.

spring-webflux/src/main/java/org/springframework/web/reactive/result/method/InvocableHandlerMethod.java
```diff
@@ -137,12 +139,14 @@ public void setReactiveAdapterRegistry(ReactiveAdapterRegistry registry) {
...
-				ReflectionUtils.makeAccessible(getBridgedMethod());
 				Method method = getBridgedMethod();
 				if (KotlinDetector.isKotlinReflectPresent() && KotlinDetector.isKotlinType(method.getDeclaringClass())) {
-					value = CoroutinesUtils.invokeSuspendingFunction(method, getBean(), args);
+					KFunction<?> function = KotlinReflectionUtils.methodToFunction(method);
+					KotlinReflectionUtils.makeFunctionAccessible(function);
+					value = CoroutinesUtils.invokeSuspendingFunction(function, getBean(), args);
 				}
 				else {
+					ReflectionUtils.makeAccessible(method);
```
rstoyanchev, line 149: Likewise, the code for this has moved to the constructor of `HandlerMethod` so this needs to be removed from here.

rstoyanchev:
I'm wondering if these classes need to remain private, i.e. is this something we want to work?

sdeleuze:
> I'm wondering if these classes need to remain private, i.e. is this something we want to work?

@rstoyanchev If you mean classe like `CoroutinesController` in the PR and if that works on Java, I would say yes.

That said, we can't modify `org.springframework.core.CoroutinesUtils#invokeSuspendingFunction` signature to use `KFunction` instead of `Method`, and `CoroutinesUtils` is now a Java class. So I think I would suggest to modify:

```
public static Publisher<?> invokeSuspendingFunction(Method method, Object target, Object... args) {
	KFunction<?> function = Objects.requireNonNull(ReflectJvmMapping.getKotlinFunction(method));
	// ...
}
```
To
```
public static Publisher<?> invokeSuspendingFunction(Method method, Object target, Object... args) {
	KFunction<?> function = Objects.requireNonNull(ReflectJvmMapping.getKotlinFunction(method));
	if (method.isAccessible() && !kotlin.reflect.jvm.KCallablesJvm.isAccessible(function)) {
		kotlin.reflect.jvm.KCallablesJvm.setAccessible(function, true);
	}
	// ...
}
```
This will impact other `invokeSuspendingFunction` invocations as well (not just those from `InvocableHandlerMethod`), but implemented like that I tend to think that's ok.

@rstoyanchev Do you agree? Is it ok to target 5.3.x for that fix?

sdeleuze:
We probably need to do the same thing on [`BeanUtils#instantiateClass`](https://github.com/spring-projects/spring-framework/blob/ec9ce5b4b0f948e510e5d958b5bcb60dc84476b3/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java#L195) and try to detect similar other patterns by a search on `ReflectionUtils#makeAccessible` usage where `kotlin-reflect` is involved.

rstoyanchev:
@sdeleuze I agree, as far as the changes to `CoroutinesUtils#invokeSuspendingFunction`.

sdeleuze:
I applied the discussed changes via 8eb618b480896092cae04d1c5e412f7a5dba0c9c on `5.3.x` and merged them to `main`.

PR #23848, opened 2019-10-22
Update Spring Boot references in testing documentation

pdeneve:
@pivotal-issuemaster This is an Obvious Fix

References to Spring Boot have been updated to point to the correct url.

Reference to old @WebIntegrationTest has been replaced with
@SpringBootTest.


snicoll:
@pdeneve good catch and thank you for making your first contribution to Spring Framework.

PR #23875, opened 2019-10-27
MockHttpServletResponse now checks for committed responses

### LINKED ISSUE #22304: MockHttpServletResponse setHeader() not checking if response is already committed
**Affects:** spring-test 4.3.16 (Maybe still the case in master)

Had a bug in my code where I was setting a header after the HttpResponse was committed. MockMvc tests were not failing but real-life app did not behave the same (with Tomcat)

https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletResponse.java#L556

compared to this

https://github.com/apache/tomcat/blob/trunk/java/org/apache/catalina/connector/ResponseFacade.java#L532

Is it intentionally missing the isCommitted() check?

bckrispi:
Fixes #22304

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletResponse.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletResponse.java
@@ -241,6 +241,9 @@ public class MockHttpServletResponse implements HttpServletResponse {
 
 	@Override
 	public void setContentLength(int contentLength) {
+		if(isCommitted()) {
+			return;
+		}
 		this.contentLength = contentLength;
 		doAddHeaderValue(HttpHeaders.CONTENT_LENGTH, contentLength, true);
 	}
@@ -251,6 +254,9 @@ public class MockHttpServletResponse implements HttpServletResponse {
 
 	@Override
 	public void setContentLengthLong(long contentLength) {
+		if(isCommitted()) {
+			return;
+		}
 		this.contentLength = contentLength;
 		doAddHeaderValue(HttpHeaders.CONTENT_LENGTH, contentLength, true);
 	}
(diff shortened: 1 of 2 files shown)


rstoyanchev:
There are some issues. First, the PR does not update all methods with side effects, e.g. setLocale, addCookie, setCookie, etc. Those would be easy to change. The bigger issue is that I've found some test failures that point to the fact Servlet containers are more sophisticated in terms of when exactly they commit the response.

To see the failures you have to also update the [MockHttpServletResponse copy](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/test/java/org/springframework/mock/web/test/MockHttpServletResponse.java) in the `spring-web/src/test` and then run tests in `spring-webmvc` and `spring-test`.

As one example, `DefaultHandlerExceptionResolverTests#handleHttpMediaTypeNotSupported()` fails because in [this code](https://github.com/spring-projects/spring-framework/blob/3b724c72b643caf795151496de7477a750a1db4a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/support/DefaultHandlerExceptionResolver.java#L281-L285) `sendError` commits the response and doesn't allow the subsequent `setHeader` but in a real Servlet container that does work.

There are other failures too that would need to be investigated further before this can be accepted. As it stands this does not look promising. The original goal was for MockMvc to behave just like a server with regards to committing the response, but replicating that exact behavior seems more nuanced and probably not feasible to align them completely.

juncevich (approved):

rstoyanchev:
I'm resetting to waiting-to-triage. Until there is a better answer to the questions above, I don't think this is workable.

rstoyanchev:
Closing for now as it does not seem possible or straight forward to align with actual Servlet container behavior. It would require a deeper look  at container vs MockHttpServletResponse behavior with regards to committing and even then there might remain differences by container.

amit-chauhan:
I know this is closed, but I would expect MockHttpServletRespnse to behave as per the specs, i.e. if committed, then don't allow changes to headers. If that's not possible because you are trying to mimic how other servlet contains handle this, then may be provide an option to enable/disable this feature as needed by the user, e.g. MockHttpServletRespnse.setStrictCommitted (boolean)

PR #23891, opened 2019-10-30
Modify the default value of ddl-auto to "none"

handong0123:
My project uses spring-data-jpa.

The company stipulates that the test environment and the generated environment database account have only read and write permissions.

I found that when ddl-auto is set to none, the database structure is also changed because there is no permission to cause the service to fail to start.

Official documentation says：
You can set spring.jpa.hibernate.ddl-auto explicitly and the standard Hibernate property values are none, validate, update, create, and create-drop. 

If I choose none, the program should not do anything.However, in spring-boot-autoconfigure, the none configuration is automatically removed, which is equivalent to not configuring ddl-auto.

I think there needs to be an option for him not to make any changes or verify the behavior of the database.

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaVendorAdapter.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaVendorAdapter.java
@@ -148,7 +148,7 @@ public class HibernateJpaVendorAdapter extends AbstractJpaVendorAdapter {
 		}
 
 		if (isGenerateDdl()) {
-			jpaProperties.put(AvailableSettings.HBM2DDL_AUTO, "update");
+			jpaProperties.put(AvailableSettings.HBM2DDL_AUTO, "none");
 		}
 		if (isShowSql()) {
 			jpaProperties.put(AvailableSettings.SHOW_SQL, "true");


snicoll:
> I found that when ddl-auto is set to none, the database structure is also changed because there is no permission to cause the service to fail to start.

Thanks for the PR but I am not sure I got that. Can you please share a small sample that reproduces the behaviour you've described? A zip attached to this issue we can run or a link to a github project would be ideal.

handong0123:
> > I found that when ddl-auto is set to none, the database structure is also changed because there is no permission to cause the service to fail to start.
> 
> Thanks for the PR but I am not sure I got that. Can you please share a small sample that reproduces the behaviour you've described? A zip attached to this issue we can run or a link to a github project would be ideal.



> > I found that when ddl-auto is set to none, the database structure is also changed because there is no permission to cause the service to fail to start.
> 
> Thanks for the PR but I am not sure I got that. Can you please share a small sample that reproduces the behaviour you've described? A zip attached to this issue we can run or a link to a github project would be ideal.

Hello, I have uploaded the sample to github：https://github.com/handong0123/jpa_test

handong0123:
Here is the method in spring-boot-autoconfigure:
`	
private Map<String, Object> getAdditionalProperties(Map<String, String> existing, HibernateSettings settings) {
		Map<String, Object> result = new HashMap<>(existing);
		applyNewIdGeneratorMappings(result);
		applyScanner(result);
		getNaming().applyNamingStrategies(result);
		String ddlAuto = determineDdlAuto(existing, settings::getDdlAuto);
		if (StringUtils.hasText(ddlAuto) && !"none".equals(ddlAuto)) {
			result.put(AvailableSettings.HBM2DDL_AUTO, ddlAuto);
		}
		else {
			result.remove(AvailableSettings.HBM2DDL_AUTO);
		}
		Collection<HibernatePropertiesCustomizer> customizers = settings.getHibernatePropertiesCustomizers();
		if (!ObjectUtils.isEmpty(customizers)) {
			customizers.forEach((customizer) -> customizer.customize(result));
		}
		return result;
	}
`If set spring.jpa.hibernate.ddl-auto=none ,it will do not work

handong0123:
So I should use spring.jpa.generate-ddl = false instead of spring.jpa.hibernate.ddl-auto=none, right?

I accidentally found that spring.jpa.hibernate.ddl-auto can write a string at will have the same effect.

handong0123:
Thank you for your answer

PR #23892, opened 2019-10-30
Only enable caching in main build when building a snapshot

wilkinsona:
Previously, caching was enabled for the main build irrespective of what type of version was being built. This meant that caching was used for snapshots, milestones, release candidates, and releases.

To ensure that tagged and released versions are built in isolation from previous builds, this PR only enables caching when building a snapshot.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -5,7 +5,9 @@ pluginManagement {
 	}
 }
 
-apply from: "$rootDir/gradle/build-cache-settings.gradle"
+if (version.endsWith("-SNAPSHOT")) {
+	apply from: "$rootDir/gradle/build-cache-settings.gradle"
+}
 
 include "spring-aop"
 include "spring-aspects"


wilkinsona:
I experimented a bit to see what effect simply changing the version number has on caching. It results in lots of misses, but there are still quite a large number of hits for compilation of main and test code. You can see the details in [this build scan](https://ge.spring.io/s/tyn2fcwp2kkic/performance/buildCache?cacheDetails=WyJsb2NhbC1IaXQiXQ). If you want release builds to be isolated from the output of earlier builds, the changes proposed in this PR seem to be necessary.

jbduncan:
@wilkinsona I'm not entirely sure why it's desirable to build the tagged and release versions without caching.

Is it because not all Gradle tasks in this project are cacheable yet, which has or may have an effect on the tagged/release versions of Spring? Or is there another reason?

snicoll:
@jbduncan Andy looked at this at my request as I was wondering if we shouldn't disable the cache for release builds. We discussed several potential problems and the outcome of that is that I don't see anything that should prevent us to keep that mechanism working.

As discussed Andy, I am closing this PR.

jbduncan:
Oh cool, thanks for enlightening me @snicoll. 👍

Issue #23907, opened 2019-11-01
WebMvcConfigurer.setUseSuffixPatternMatch(false) ignored if addPathPrefix() is also used


tpischke-bedag:
Affected version: 5.1.8

I am trying to disable suffix pattern matching so that PathVariables containing dots will not be truncated. The [documented solution](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-requestmapping-suffix-pattern-match) for this problem

`webMvcConfigurer.setUseSuffixPatternMatch(false)`

works fine, but it breaks if I also try to add a path prefix

`webMvcConfigurer.addPathPrefix("/api", HandlerTypePredicate.forAnnotation(RestController.class))`

The problem is in PatternsRequestCondition.combine(PatternRequestCondition other) at line 179. Here the useSuffixPatternMatch on my url is ignored in favor of the value on the '/api' Condition create at RequestMappingHandlerMapping.228

RequestMappingHandlerMapping.java
```java
@Override
@Nullable
protected RequestMappingInfo getMappingForMethod(Method method, Class<?> handlerType) {
		RequestMappingInfo info = createRequestMappingInfo(method);
	if (info != null) {
		RequestMappingInfo typeInfo = createRequestMappingInfo(handlerType);
		if (typeInfo != null) {
			info = typeInfo.combine(info);
		}
		String prefix = getPathPrefix(handlerType);
		if (prefix != null) {
                        // PROBLEM HERE
			info = RequestMappingInfo.paths(prefix).build().combine(info);**
		}
	}
	return info;
}
```
PatternsRequestCondition.java
```java
@Override
public PatternsRequestCondition combine(PatternsRequestCondition other) {
	Set<String> result = new LinkedHashSet<>();
	if (!this.patterns.isEmpty() && !other.patterns.isEmpty()) {
		for (String pattern1 : this.patterns) {
			for (String pattern2 : other.patterns) {
				result.add(this.pathMatcher.combine(pattern1, pattern2));
			}
		}
	}
	else if (!this.patterns.isEmpty()) {
		result.addAll(this.patterns);
	}
	else if (!other.patterns.isEmpty()) {
		result.addAll(other.patterns);
	}
	else {
		result.add("");
	}
        // PROBLEM HERE
	return new PatternsRequestCondition(result, this.pathHelper, this.pathMatcher,
			this.useSuffixPatternMatch, this.useTrailingSlashMatch, this.fileExtensions);
}
`

[...]

PR #23910, opened 2019-11-02
PatternsRequestCondition combine bug fix

kwahome:
When combining the current `PatternsRequestCondition` instance and another, a new instance with combined URL patterns from both is created and returned.
If the other instance sets other fields such as `useSuffixPatternMatch` and `useTrailingSlashMatch`, they are ignored as [the current instance (`this`) is supplied](https://github.com/spring-projects/spring-framework/blob/v5.2.0.RELEASE/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/condition/PatternsRequestCondition.java#L195) to the [constructor instead of the `other` instance](https://github.com/spring-projects/spring-framework/blob/v5.2.0.RELEASE/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/condition/PatternsRequestCondition.java#L128) presenting buggy behaviour while customizing path matching and URL treatment options.

see gh-23907

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/condition/PatternsRequestCondition.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/condition/PatternsRequestCondition.java
@@ -192,7 +192,7 @@ public final class PatternsRequestCondition extends AbstractRequestCondition<Pat
 		else {
 			result.add("");
 		}
-		return new PatternsRequestCondition(result, this);
+		return new PatternsRequestCondition(result, other);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


rstoyanchev:
@kwahome thanks for the pull request, but I'm not sure this is making things better. 

The challenge is in picking between `this` and the other instance, which typically correspond to type vs method-level annotations. It's just as random to pick the 1st or  the 2nd. The expectation actually is that both are created in the same way (in `RequestMappingHandlerMapping`) and that the build options are applied to both, so it doesn't matter which one is picked. 

I would rather keep using `this` but apply the builder options over here:
https://github.com/spring-projects/spring-framework/blob/8855056395a3603ea7c8e50684714f2caad63827/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerMapping.java#L235.

rstoyanchev:
Thanks for the PR again. This has been in https://github.com/spring-projects/spring-framework/commit/203977972b9c6e0a26cf99e30b620f9ab127bc5b.

PR #23914, opened 2019-11-03
Rename earlyProxyReferences to earlyBeanReferences in AbstractAutoProxyCreator

chenqimiao:
The original map named earlyProxyReferences seems to stores the target object, not the proxy object.


snicoll:
Thanks @chenqimiao. For consistency with the method that fills the cache, I've renamed that to `earlyBeanReferences`.

PR #23917, opened 2019-11-04
Improve BeanPropertyRowMapper.underscoreName()

stsypanov:
Currently the method allocates a substring of length = 1 for each character in field name. This can be improved by using `char` instead of `String`.

Also I doubt that we need name's length check: in fact there can be no property with empty name or one consisting of  exclusively whitespaces/tabs. I think this check can be removed, and even if the code throws after transformation this indicates a problem on the upper level.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/BeanPropertyRowMapper.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/BeanPropertyRowMapper.java
@@ -248,15 +248,14 @@ public class BeanPropertyRowMapper<T> implements RowMapper<T> {
 			return "";
 		}
 		StringBuilder result = new StringBuilder();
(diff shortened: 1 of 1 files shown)


jhoeller:
While this seems worthwhile for the common case, we'd lose our existing delegation to the `lowerCaseName` template method... so any customizations in subclasses wouldn't get invoked anymore. Or am I missing something?

stsypanov:
@jhoeller you mean one might override `lowerCaseName` to use their own locale?

jhoeller:
Indeed, from the perspective of potential customizations in user-level subclasses (through overriding those protected template methods), I'm afraid we need to stick with the current implementation.

stsypanov:
> I'm afraid we need to stick with the current implementation.

Then I close this PR.

PR #23923, opened 2019-11-04
Improve usage of String.substring()

stsypanov:
Proposed transformations make code more simple and faster due to reduction of allocations.


sbrannen:

Thanks

PR #23927, opened 2019-11-04
Add Kotlin extensions for Cache and CacheManager

sokomishalov:
kotlin extensions for Cache and CacheManager


snicoll:
Thanks @sokomishalov

PR #23931, opened 2019-11-05
Transform upper keys to coroutines context fix #23859

### LINKED ISSUE #23859: Provide transform from mono Context to coroutinesContext
Hello, we want to keep mono context in coroutine scope
For example:
we use webFilter and set something to Context(reactor.util.context)
```
...
filter(exchange: ServerWebExchange, chain: WebFilterChain): Mono<Void> {
      context.put(ourCoroutineContext, "test")
...
}
...
```
And when we use such methods in controller:
```
suspend fun test():Response<Entity> {
       val test = coroutineContext[ourCoroutineContext]
}
```
test will be null.
I think, it would be good to convert mono Context to coroutine Context, before invoke function.
I see a part of code in InvocableHandlerMethod
```
ReflectionUtils.makeAccessible(getBridgedMethod());
				Method method = getBridgedMethod();
				if (KotlinDetector.isKotlinReflectPresent() && KotlinDetector.isKotlinType(method.getDeclaringClass())) {
					value = CoroutinesUtils.invokeSuspendingFunction(method, getBean(), args);
				}
```
It would be good to extend invokeSuspend fun to take a mono Context, which will be transformed.
What do you think? Also I would like to submit a PR, but I can't find what CoroutineUtils is.

kostya05983:
@sdeleuze , I think it would be better if keys which are coroutinesContext will be transform in upper keys, but not only as keys to ReactorContext. ReactorContext is uncomfortable to use.

See #23859

### DIFF
--- a/spring-core/kotlin-coroutines/src/main/kotlin/org/springframework/core/CoroutinesUtils.kt
+++ b/spring-core/kotlin-coroutines/src/main/kotlin/org/springframework/core/CoroutinesUtils.kt
@@ -17,6 +17,7 @@
 @file:JvmName("CoroutinesUtils")
 package org.springframework.core
 
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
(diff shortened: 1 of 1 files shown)


sdeleuze:
Hi and sorry for the long delay for reviewing this PR. Unless I miss something, Reactor / Coroutines context interoperability  is expected to be performed at https://github.com/Kotlin/kotlinx.coroutines/ level (more specifically [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/reactive/kotlinx-coroutines-reactor/src/ReactorContext.kt)). If something does not behaves as expected, please report the issue with a reproducer in [their issue tracker](https://github.com/Kotlin/kotlinx.coroutines/issues).

If the related issue is Spring specific, please comment and provide a reproducer (JUnit test if possible, otherwise application).

PR #23945, opened 2019-11-07
Fix bug : sorting problem with beanPostProcessor annotated by @Order

chenqimiao:
The `beanPostProcessor` annotated with @Order will not be sorted, but the implementation of `PriorityOrdered` or `Ordered` will be sorted.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/PostProcessorRegistrationDelegate.java
+++ b/spring-context/src/main/java/org/springframework/context/support/PostProcessorRegistrationDelegate.java
@@ -243,7 +243,7 @@ final class PostProcessorRegistrationDelegate {
 				internalPostProcessors.add(pp);
 			}
 		}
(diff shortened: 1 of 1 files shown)


chenqimiao:
Looking forward to your reply .

sbrannen:
The [Ordering](https://docs.spring.io/spring/docs/5.2.1.RELEASE/javadoc-api/org/springframework/beans/factory/config/BeanPostProcessor.html) section of the class-level Javadoc for `BeanPostProcessor` states the following.

> `BeanPostProcessor` beans that are autodetected in an `ApplicationContext` will be ordered according to `PriorityOrdered` and `Ordered` semantics. In contrast, `BeanPostProcessor` beans that are registered programmatically with a `BeanFactory` will be applied in the order of registration; any ordering semantics expressed through implementing the `PriorityOrdered` or `Ordered` interface will be ignored for programmatically registered post-processors. Furthermore, the `@Order` annotation is not taken into account for `BeanPostProcessor` beans.

PR #23957, opened 2019-11-08
Update webflux.adoc

denisgalaybo:
Fixed error in examples.


denisgalaybo:
I'm not an expert in Kotlin, but I think ".build()" was missing as well.

PR #23966, opened 2019-11-11
Fix @Configuration class's setBeanFactory method is called multiple times

chenqimiao:
`@Configuration` class's `setBeanFactory` method is called multiple times.

- First point is `AbstractAutowireCapableBeanFactory#invokeAwareMethods`.
- Second point is `ImportAwareBeanPostProcessor#postProcessProperties`.

So, I think we can clean some code.

I don't know if I am right, I am looking forward to your answer.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassPostProcessor.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassPostProcessor.java
@@ -448,11 +448,6 @@ public class ConfigurationClassPostProcessor implements BeanDefinitionRegistryPo
 
 		@Override
 		public PropertyValues postProcessProperties(@Nullable PropertyValues pvs, Object bean, String beanName) {
-			// Inject the BeanFactory before AutowiredAnnotationBeanPostProcessor's
-			// postProcessProperties method attempts to autowire other configuration beans.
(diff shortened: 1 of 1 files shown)


chenqimiao:
Maybe I need to provide a sample code.
`
@Configuration

public class AppConfig implements BeanFactoryAware {

	@Override
	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
		System.out.println("Call AppConfig#beanFactory method ");
	}
}`


`
public class TestApplication {
    
    public static void main(String args[]) {
   	       AnnotationConfigApplicationContext ac = new AnnotationConfigApplicationContext(AppConfig.class);
    
          }
}`

Run main method of TestApplication class, the string of `Call AppConfig#beanFactory method` will be print two times

sbrannen:
I've edited your comments to improve the formatting. You might want to check out this [Mastering Markdown guide](https://guides.github.com/features/mastering-markdown/) for future reference.

chenqimiao:
@sbrannen Thanks, I will pay attention to my format in the future.

snicoll:
Thanks for the PR and the proposal. None of the Aware callback APIs have a contract that the callback is only called once. What can be seen as an optimization is more complex than this, with other code path where the code you've removed would be the only time where we get a chance to set the `BeanFactory`.

PR #23970, opened 2019-11-11
Tidy up classpath pollution caused by resource creation in the tests

wilkinsona:
Previously, spring-webmvc and spring-webflux both contained tests that would create gzipped files, write them to the filesystem alongside the project's compiled test classes, and configure them to be deleted on JVM exit. The output location placed the files on the classpath, polluting it for every subsequent test that used the same ClassLoader. The test-sources plugin combined with Gradle's use of worker JVMs, broadens the scope of this pollution to other, downstream
projects in the same build. For example, the tests for spring-websocket will have a different classpath depending on whether or not the tests for spring-webmvc have already been run on the same worker as part of the current build.

This PR updates the spring-webmvc and spring-webflux modules to introduce a new JUnit Jupiter extension, GzipSupport. This extension allows gzipped files to be created via an injectable GzippedFiles class and automatically deletes each created file in an after-each callback. This ensures that a gzipped file only exists on the classpath for the duration of the test that needs it, avoiding the pollution of the classpath of any subsequent tests.

### DIFF
--- a/spring-webflux/src/test/java/org/springframework/web/reactive/resource/CachingResourceResolverTests.java
+++ b/spring-webflux/src/test/java/org/springframework/web/reactive/resource/CachingResourceResolverTests.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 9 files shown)


sbrannen:

Nice `GzipSupport` extension for JUnit Jupiter. 👍 

Thanks

PR #23980, opened 2019-11-12
Add queryParam method to MockHttpServletRequestBuilder

### LINKED ISSUE #23296: Add queryParam method to MockHttpServletRequestBuilder

jo2:
add queryParam() which calls param() and updates queryString

Fixes #23296

@rstoyanchev opened a new pr because I had to force push to the branch and GitHub couldn't connect the branch to the old repo


rstoyanchev:
This is now merged with some extra, minor refactoring. Thanks for the pull request!

PR #23981, opened 2019-11-12
Polish AbstractAutowireCapableBeanFactory#populateBean

GungnirLaevatain:
1.  if **postProcessAfterInstantiation** return false,  it could return immediately
2.  **instanceof** could replace **isInstance**


jhoeller:
Wells spotted! Thanks for the PR.

PR #23994, opened 2019-11-14
Avoid indefinite wait to connect in JettyWebSocketClient

Kukosoft:
In some cases, `org.eclipse.jetty.websocket.client.WebSocketClient.connect(listener, uri, request)` call will return a future that is never completed. It is reasonable that our `future.get()` must have a timeout to avoid thread blocking.
I suggest something like:
```java
Callable<WebSocketSession> connectTask = () -> {
	Future<Session> future = this.client.connect(listener, uri, request);
	try {
		// TODO Configurable timeout
		future.get(2000, TimeUnit.MILLISECONDS);
	} catch (Exception ex){
		logger.error("Failed to connect to remote websocket endpoint", ex);
		future.cancel(true); // This method will stop the running underlying task
	}
	return wsSession;
};
```

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/client/jetty/JettyWebSocketClient.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/client/jetty/JettyWebSocketClient.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jetty.websocket.api.Session;
 import org.eclipse.jetty.websocket.client.ClientUpgradeRequest;
@@ -155,7 +156,13 @@ public class JettyWebSocketClient extends AbstractWebSocketClient implements Lif
 
 		Callable<WebSocketSession> connectTask = () -> {
 			Future<Session> future = this.client.connect(listener, uri, request);
-			future.get();
+			try {
+				// TODO Configurable timeout
+				future.get(2000, TimeUnit.MILLISECONDS);
+			} catch (Exception ex){
+				logger.error("Failed to connect to remote websocket endpoint", ex);
+				future.cancel(true); // This method will stop the running underlying task
+			}
 			return wsSession;
 		};
 


pivotal-issuemaster:
@Kukosoft Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23994)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23994) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Kukosoft Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=23994)!

rstoyanchev:

rstoyanchev:
A connect timeout is better set at the level of the underlying HttpClient. In this case, `org.eclipse.jetty.websocket.client.WebSocketClient` does have a `connectTimeout` property that is set to 15 seconds by default but you can change that.

Kukosoft:
> 
> 
> A connect timeout is better set at the level of the underlying HttpClient. In this case, `org.eclipse.jetty.websocket.client.WebSocketClient` does have a `connectTimeout` property that is set to 15 seconds by default but you can change that.

In practice, in a performance test I have set this connectTimeout to 500ms and I have few threads blocked in the get() method indefinitely...
I think we need to protect against any bug from third party libraries.
Could you check this out, please?
Thanks.

rstoyanchev:
Okay we could get the connectTimeout from the Jetty WebClient and wait for as long or a little longer. I don't think we should be exposing the same property for this though. It sounds like an issue that should be reported to Jetty because if the Jetty HttpClient is stuck and keep giving up, that is also an issue.

Kukosoft:
> 
> 
> Okay we could get the connectTimeout from the Jetty WebClient and wait for as long or a little longer. I don't think we should be exposing the same property for this though. It sounds like an issue that should be reported to Jetty because if the Jetty HttpClient is stuck and keep giving up, that is also an issue.

I'm totally agree with you. Jetty WebSocketClient is sometimes and somehow not completing its Future connection, so, Spring's side needs a protection like you said. 
Okay with connectTimeout+paddingTime, but what about throwing an exception inside the callable task? How to inform  the ListenableFuture<WebSocketSession> about this timeout?

rstoyanchev:
If the `Callable` fails, listeners of the `ListenableFuture` will be notified (see `ListenableFutureTask`).

Kukosoft:
Rossen, let me edit the file to change
future.get(2000, TimeUnit.MILLISECONDS);
to
future.get(this.client.getConnectTimeout(), TimeUnit.MILLISECONDS);
and also throw an exception.
```
import org.springframework.web.socket.server.HandshakeFailureException;
...
Callable<WebSocketSession> connectTask = () -> {
	Future<Session> future = this.client.connect(listener, uri, request);
	try {
		future.get(this.client.getConnectTimeout(), TimeUnit.MILLISECONDS);
	} catch (Exception ex){
		future.cancel(true); // This method will stop the running underlying task
                throw HandshakeFailureException("Failed to connect to remote websocket endpoint", ex);
	}
	return wsSession;
};
```
Thanks

PR #24002, opened 2019-11-14
Simplify AbstractAsyncConfiguration

michaldo:
1. There is simpler way to express that 0 or 1 bean only is allowed
2. This commit allows flexible AsyncConfigurer substitution with `@Primary`


spring-context/src/main/java/org/springframework/scheduling/annotation/AbstractAsyncConfiguration.java
```diff
@@ -62,20 +61,12 @@ public void setImportMetadata(AnnotationMetadata importMetadata) {
...
-		}
-		if (configurers.size() > 1) {
-			throw new IllegalStateException("Only one AsyncConfigurer may exist");
-		}
-		AsyncConfigurer configurer = configurers.iterator().next();
-		this.executor = configurer::getAsyncExecutor;
-		this.exceptionHandler = configurer::getAsyncUncaughtExceptionHandler;
+	@Autowired
+	void setConfigurers(Optional<AsyncConfigurer> configurerOptional) {
+		configurerOptional.ifPresent(configurer -> {
```
spac-valentin: This could throw `NullPointerException` as `configurerOptional` can be null. In addition, `Optional` is rather best used for showing that a computation might not return a result than an optional input parameter. For this case, static polymorphism (overloading) would be a better option

spring-context/src/main/java/org/springframework/scheduling/annotation/AbstractAsyncConfiguration.java
```diff
@@ -62,20 +61,12 @@ public void setImportMetadata(AnnotationMetadata importMetadata) {
...
-			return;
-		}
-		if (configurers.size() > 1) {
-			throw new IllegalStateException("Only one AsyncConfigurer may exist");
-		}
-		AsyncConfigurer configurer = configurers.iterator().next();
-		this.executor = configurer::getAsyncExecutor;
-		this.exceptionHandler = configurer::getAsyncUncaughtExceptionHandler;
+	@Autowired
+	void setConfigurers(Optional<AsyncConfigurer> configurerOptional) {
```
spac-valentin: In my opinion, a better approach would be to create a new function which clearly shows the intent (if exists, only one configurer should be used) and deprecate the old one. This would still allow configuration using `@Primary` 
```
@Autowired(required = false)
void setAsyncConfigurer(AsyncConfigurer asyncConfigurer)
```

michaldo: I agree, that looks better comparing to version with `Optional`

spring-context/src/main/java/org/springframework/scheduling/annotation/AbstractAsyncConfiguration.java
michaldo: `configurerOptional` can't be null. Spring does not allow to inject `null Optional` - Optional was invented to avoid null problems.

Anyway, I modified signature and removed `Optional` - but for simplicity, not for null fear

spac-valentin: That's correct, Spring doesn't allow `null` injection. But leaving injection aside, that is still a function that can be invoked by some other class from the `org.springframework.scheduling.annotation` package (although is not the case for now)

snicoll:
Thanks for the PR but I'd like to keep things as they are. One gotcha is that the supplier is kept as late as possible and only resolved when necessary whereas this change can trigger early initialization.

I also agree they are simpler ways to express the 0-1 beans use case but I favor readability of the code vs. optimization for configuration arrangements. Thanks anyway

PR #24007, opened 2019-11-15
Asssociate timeout to underlying Jetty HTTP client

Kukosoft:
The websocket handshake timeout is now the sum of connection timeout, idle tiemout and 1 margin second.

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/client/jetty/JettyWebSocketClient.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/client/jetty/JettyWebSocketClient.java
@@ -158,7 +158,7 @@ public class JettyWebSocketClient extends AbstractWebSocketClient implements Lif
 		Callable<WebSocketSession> connectTask = () -> {
 			Future<Session> future = this.client.connect(listener, uri, request);
 			try {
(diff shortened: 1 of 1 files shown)


rstoyanchev:
The timeout part has already been done (see the commit that closed #23994). As for cancelling the future, as far as I can see in Jetty's `WebSocketUpgradeRequest` the `CompletableFuture` is only notified on completion/erro and never checked for cancellation. I don't think the cancellation would make any difference.

Kukosoft:
Ok, Rossen. One note about the current timeout implementation. I think that this.client.getHttpClient().getIdleTimeout() must be added to HTTP client connect timeout because the websocket upgrade can take also the request "read time"

rstoyanchev:
@Kukosoft, I agree handshake time isn't quite the same as the connect time, but `idleTimeout` is a completely different property altogether that's not a good fit for connecting. How long you might be willing to leave the connection idle isn't necessarily how long you want to wait to connect. Also it doesn't any default value that we can use.

If you goal is to have configurable control, you already can change `connectTimeout` to anything and that gives enough control for this case I think. Just set it to however long you're willing to wait altogether.

Issue #23348, opened 2019-07-24
Support escape chars in PropertyPlaceholderHelper


philwebb:
We have [a Spring Boot pull request](https://github.com/spring-projects/spring-boot/pull/17530) open that uses `PropertyPlaceholderHelper` to resolve `{variable}`  items in a message source. We'd like to also support escaping so that user can write `\{notreplaced}` to bypass expansion.

jhoeller:
In terms of its design, `PropertyPlaceholderHelper` is not really meant to go to that degree of sophisticated parsing. We got `ignoreUnresolvablePlaceholders` there, as a sort-of alternative to escaping. That seems good enough for its original purpose of parsing configuration values. I wonder whether message subtitution isn't better off with some other parser, even a custom one...

philwebb:
Thanks @jhoeller, I'll close this one and consider at an alternative parser.

PR #24010, opened 2019-11-17
Support escape chars in PropertyPlaceholderHelper

nosan:
see gh-23348


nosan:
PR does not correctly handle `escape` chars hence I'm closing it for now.

Issue #13874, opened 2012-03-14
SimpleJdbcInsert does not quote table and column names [SPR-9236]


spring-projects-issues:
**[Michel Zanini](https://jira.spring.io/secure/ViewProfile.jspa?name=michelz)** opened **[SPR-9236](https://jira.spring.io/browse/SPR-9236?redirect=false)** and commented

If you have table or column names that are reserved SQL words, like "varchar", "values", "integer", etc... the generated insert SQL must have escaping.

For example, now `SimpleJdbcInsert` is generating this SQL:

```sql
INSERT INTO crud (text, integer, char, datetime, decimal, varchar) VALUES(?, ?, ?, ?, ?, ?)
```

but it should be generating this:

```sql
INSERT INTO `crud` (`text`, `integer`, `char`, `datetime`, `decimal`, `varchar`) VALUES(?, ?, ?, ?, ?, ?)
```

As a workaround I am able to make it work adding the escape char "`" to the column names like this:

```java
simpleJdbcInsert.withTableName("`" + tableName + "`");
simpleJdbcInsert.usingColumns("`" + column1 + "`", "`" + column2 + "`", "`" + column3 + "`");
```

but I believe `SimpleJdbcInsert` should support this out of the box. For example, it could have a boolean property like this:

```java
simpleJdbcInsert.withTableName(tableName);
simpleJdbcInsert.usingColumns(column1, column2, column3);
simpleJdbcInsert.usingEscaping(true);
```

and then `SimpleJdbcInsert` would do the escaping for me.

---

**Affects:** 3.1.1

4 votes, 6 watchers

vgrankina:
We need the same functionality for PostgreSQL because we use a mix of uppercase and lowercase letters for table names and we need to escape them with double quotes. When I set table name with double quotes like so ```
new SimpleJdbcInsert(dataSource)
                .withTableName("\"Table_Name\"");
```
I see a log message that metadata for table not found.

**Comment by sadafskolon** [comment]
It's 2023 and simple JDBC insert still does not support keyword columns.

**Comment by sadafzohra19** [comment]
It worked for me. What I did is,

```
String[] columns = {"name", "desc",  "code", "`order`"}

SimpleJdbcInsert insert = new SimpleJdbcInsert(jdbcTemplate.getJdbcTemplate());
insert.withTableName(tableName);
insert.usingColumns(columns);
insert.usingGeneratedKeyColumns("id");

[...]

PR #24013, opened 2019-11-18
Support quoted identifiers in `SimpleJdbcInsert`

kitaisreal:
Motivation https://github.com/spring-projects/spring-framework/issues/13874

**Current implementation steps**
1.  Getting identifierQuoteString to use for escaping from dataBaseMetadata in `GenericTableMetadataProvider` http://docs.oracle.com/javase/6/docs/api/java/sql/DatabaseMetaData.html#getIdentifierQuoteString%28%29.
2. Added property usingEscaping in `SimpleJdbcInsert` and logic related to handling it in `TableMetadataContext`.
3. Updated tests.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/GenericTableMetaDataProvider.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/GenericTableMetaDataProvider.java
@@ -76,6 +76,8 @@ public class GenericTableMetaDataProvider implements TableMetaDataProvider {
 	/** Collection of TableParameterMetaData objects. */
 	private List<TableParameterMetaData> tableParameterMetaData = new ArrayList<>();
 
+	/** the string used to quote SQL identifiers. */
+	private String identifierQuoteString = "";
 
 	/**
 	 * Constructor used to initialize with provided database meta-data.
@@ -212,6 +214,15 @@ public class GenericTableMetaDataProvider implements TableMetaDataProvider {
 				logger.warn("Error retrieving 'DatabaseMetaData.storesLowerCaseIdentifiers': " + ex.getMessage());
 			}
 		}
+
+		try {
+			this.identifierQuoteString = databaseMetaData.getIdentifierQuoteString();
+		}
+		catch (SQLException ex) {
+			if (logger.isWarnEnabled()) {
+				logger.warn("Error retrieving 'DatabaseMetaData.getIdentifierQuoteString': " + ex.getMessage());
+			}
+		}
 	}
 
 	@Override
@@ -302,6 +313,14 @@ public class GenericTableMetaDataProvider implements TableMetaDataProvider {
 	@Nullable
 	protected String getDatabaseVersion() {
 		return this.databaseVersion;
+	}
+
+	/**
+	 * Provide access to identifier quote string.
+	 */
+	@Override
+	public String getIdentifierQuoteString() {
+		return this.identifierQuoteString;
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


Dretch:
This is awesome, thank you very much!

foldvari:
As far as I see this solution does not support cases when the opening and the closing character used for quoting is different. (E.g. MS SQL [identifier] ).

spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/GenericTableMetaDataProvider.java
```diff
@@ -76,6 +76,8 @@
 	/** Collection of TableParameterMetaData objects. */
 	private List<TableParameterMetaData> tableParameterMetaData = new ArrayList<>();
 
+	/** the string used to quote SQL identifiers. */
+	private String identifierQuoteString = "";
```
foldvari: There is at least one RDBMS (MS SQL) which uses different characters for opening and closing. ( [identifier] )

kitaisreal: Could you please provide link to documentation where it is mentioned ? I could only find this https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/quoted-identifiers?view=sql-server-ver15. 
We are getting this identifier to quote SQL from database metadata https://docs.oracle.com/javase/6/docs/api/java/sql/DatabaseMetaData.html#getIdentifierQuoteString%28%29. If this class does not support case where there are 2 different characters to quote SQL identifiers, we also could not.

foldvari: Thanks for the links to the documentation. It seems that the driver takes care about switching on the support of quoted identifiers (QUOTED_IDENTIFIER) so the change will work with MS SQL as well.

manicmonkey:
@sbrannen you recently made changes in this area - can you comment on the viability of this PR?

sbrannen (changes requested):
@kitaisreal,

Thanks for the PR, and sorry for taking so long to review this.

Do you have time to make the requested changes (consistently referring to "quote identifiers" instead of "using escaping") and rebase on `main`?

If not, just let us know, and we'll take over the PR.

Cheers,

Sam

sbrannen:
> Do you have time to make the requested changes (consistently referring to "quote identifiers" instead of "using escaping") and rebase on `main`?

@kitaisreal, you can disregard that. I am taking care of that and other changes locally after merging your PR.

spring-jdbc/src/test/java/org/springframework/jdbc/core/simple/SimpleJdbcInsertTests.java
```diff
@@ -81,4 +82,36 @@ public void testNoSuchTable() throws Exception {
...
+	}
+
+	@Test
+	public void testSimpleJdbcInsertWithEscapingWithSchemaName() throws Exception {
+		SimpleJdbcInsert jdbcInsert = new SimpleJdbcInsert(dataSource).withSchemaName("S").withTableName("T").usingColumns("F", "S").usingEscaping(true);
+
+		given(databaseMetaData.getIdentifierQuoteString()).willReturn("`");
+
+		jdbcInsert.compile();
+		String expected = "INSERT INTO `S.T` (`F`, `S`) VALUES(?, ?)";
```
sbrannen, line 101:
```suggestion
		String expected = "INSERT INTO `S`.`T` (`F`, `S`) VALUES(?, ?)";
```

Just a side note: the schema and table names have to be quoted independently.

I've fixed this in my local branch and added integration tests with H2 to verify it.

sbrannen:
This has been merged into `main` in d39034754fc99862fc80eb8bd860004afd4b6d5a and revised in 7dc0653f38a0ac653bf29068dbc4bca632530eea. See also 070590cb1162ff5ef3cd2a3f7a64390caa58782c.

Thanks, and congratulations on getting your first PR for the Spring Framework merged! 👍

PR #24014, opened 2019-11-18
Simplify UriComponentsBuilder.FullPathComponentBuilder.build()

stsypanov:
String::replace is more simple and likely to perform better than repeating substring concatenation


jhoeller:
I'm afraid this is a somewhat subtle case: We are reducing any number of slashes there to single slashes, so for a "///" section we'll first cut it down to "//" then to "/". So it's a not simple replacement, unfortunately. For the actual purpose there, the original algorithm is probably better left as-is.

As a side note, if it were a simple replacement, we'd rather use `StringUtils.replace` in order to avoid internal regex pattern compilation in `String.replace`.

stsypanov:
Thanks for the review, I've completely missed the case when we could have "///"!

PR #24019, opened 2019-11-18
Add parameter name module to object mapper

jehrhardt:
The three modules for Java 8 feature support in Jackson have been merged into one (https://github.com/FasterXML/jackson-modules-java8). It makes sense to treat all three now as well known modules that are registered by default.

As the documentation from the new project says about Jackson 3:

> Because of this parameter-names and datatypes modules are merged into jackson-databind and need not be registered; datetime module (JavaTimeModule) remains separate module due to its size and configurability options.

Spring could provide this behavior already today.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java
@@ -82,6 +82,8 @@ import org.springframework.util.xml.StaxUtils;
  * <ul>
  * <li><a href="https://github.com/FasterXML/jackson-datatype-jdk8">jackson-datatype-jdk8</a>:
  * support for other Java 8 types like {@link java.util.Optional}</li>
+ * <li><a href="https://github.com/FasterXML/jackson-module-parameter-names">jackson-module-parameter-names</a>:
+ * support for constructor parameter names during deserialization</li>
  * <li><a href="https://github.com/FasterXML/jackson-datatype-jsr310">jackson-datatype-jsr310</a>:
  * support for Java 8 Date & Time API types</li>
  * <li><a href="https://github.com/FasterXML/jackson-datatype-joda">jackson-datatype-joda</a>:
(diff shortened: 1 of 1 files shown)


jhoeller:
For some earlier discussion about registering the parameter names module by default, see #17886. @sdeleuze what's your late-2019 opinion here?

mauromol:
I also find it confusing that the parameter name module is not automatically registered by Spring Framework alone, while Spring Boot does it. Although I also discovered that Jackson 2 ability to detect parameter name is not as effective as Spring's one, since you must use Java 8+ with `-parameter` compiler option, otherwise it won't work anyway.

mauromol:
There is also the `Paranamer` module which seems to work much better: https://github.com/FasterXML/jackson-modules-base/tree/master/paranamer
No need for Java 8+ (which is not so important nowadays anyway), but also no need for `-parameters` compiler flag. This flag is disabled by default, but while Spring (Framework, Boot, Data) is able to detect parameter names all the same, Jackson 2 is not and so, even if the `ParameterName` module is registered, you get an inconsistent experience (I lost a lot of time today to figure this out).

rstoyanchev:
It looks like this was added as a feature in Spring Boot after #17886 was closed. Perhaps we can move this into `Jackson2ObjectMapperBuilder` for 6.0, following a similar approach by declaring `new ParameterNamesModule(JsonCreator.Mode.PROPERTIES)`. 

I'm not sure if there are any further considerations around this with how the Jackson auto-config has evolved?

wilkinsona:
We switched to using `DEFAULT` in Boot for reasons described in [this issue](https://github.com/spring-projects/spring-boot/issues/4336). That was back in Boot 1.3 (2015) and things haven't changed since then.

One other consideration is that Boot, via its Gradle plugin and Maven parent, can enable compilation with `-parameters` automatically for users. It's hard to quantify how much that has helped people use the parameter names module successfully. We know that there are Maven users who don't use `spring-boot-starter-parent` but I don't recall many issues being reported by folks for whom things weren't working.

I think it would be a shame if rolling this into the `ObjectMapper` build made it harder for Boot users to configure a customised parameter names module. They can do so today by providing the module as a bean which is then registered via `Jackson2ObjectMapperBuilder.modulesToInstall(Module...)`. Hopefully that could still work and would override whatever the builder would do by default.

Having written the above, I've just noticed that the parameter names module has been folded into jackson-databind in Jackson 3.0. That makes me wonder if it's worth hold off here to see if the Jackson 3.0 release date becomes more clear.

rstoyanchev:
Given that there isn't a clear way to instantiate the module without some trade-offs, and the fact that it will no longer need to be registered in Jackson Databind 3.0, I don't think we should do this. We would only do so for 6.0 at this point, which would introduce a change, that would have then to be removed for Jackson 3.0, yet another change. That's more confusion that is worth for something that's already taken care of in Boot.

PR #24032, opened 2019-11-19
Support the feature that combine multiple SqlParameterSource

### LINKED ISSUE #23962: Support combining multiple SqlParameterSource
I want to support the feature that combine multiple `SqlParamterSource` to a one `SqlParameterSource`. For example, When save a domain object to the database, I want to bind retrieving values  from domain object and other object(e.g. login user object) to an SQL using `NamedParameterJdbcOperations`.

WDYT?

**Code sample:**

```java
Account newAccount = new Account();
// ...
KeyHolder keyHolder = new GeneratedKeyHolder();
jdbcOperations.update(
    "INSERT INTO accounts (name, email, created_at, created_by) " +
    "VALUES(:name, :email, current_timestamp, :operatorId)",
    SqlParameterSource.combine(
        new BeanPropertySqlParameterSource(newAccount), // retrieving 'name' and 'email'
        new BeanPropertySqlParameterSource(operator)), // retrieving 'operatorId'
    keyHolder);
// ...
```

```java
public interface SqlParameterSource {

  // …

  static SqlParameterSource combine(SqlParameterSource... sources) {
    return new CompositeSqlParameterSource(sources);
  }

  public static class CompositeSqlParameterSource implements SqlParameterSource {
    private final Map<String, SqlParameterSource> cache = new HashMap<>();
    private final SqlParameterSource[] sources;

    private CompositeSqlParameterSource(SqlParameterSource... sources) {
      this.sources = sources;
    }

    @Override
    public boolean hasValue(String paramName) {
      return cache.computeIfAbsent(paramName,
          x -> Stream.of(sources).filter(s -> s.hasValue(paramName)).findFirst().orElse(null)) != null;
    }

    @Override
    public Object getValue(String paramName) {
      return hasValue(paramName) ? cache.get(paramName).getValue(paramName) : null;
    }

    @Override
    public String getTypeName(String paramName) {
      return hasValue(paramName) ? cache.get(paramName).getTypeName(paramName) : null;
    }

    @Override
    public int getSqlType(String paramName) {
      return hasValue(paramName) ? cache.get(paramName).getSqlType(paramName) : TYPE_UNKNOWN;
    }

    @Override
    public String[] getParameterNames() {
      return Stream.of(sources)
          .map(SqlParameterSource::getParameterNames)
          .flatMap(Stream::of).distinct()
          .toArray(String[]::new);
    }
  }

}

```

kazuki43zoo:
Fixes gh-23962

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/CompositeSqlParameterSource.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/CompositeSqlParameterSource.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.jdbc.core.namedparam;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+import org.springframework.lang.NonNull;
+import org.springframework.lang.Nullable;
+
+/**
+ * A composite implementation of the {@link SqlParameterSource} interface.
+ *
+ * @author Kazuki Shimizu
+ * @since 5.2.2
+ * @see BeanPropertySqlParameterSource
+ * @see MapSqlParameterSource
+ */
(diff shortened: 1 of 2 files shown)


maheksaluja (commented):
Looks good to me..

virtual-machinist:
@maheksaluja , any ETA on when this PR is to be merged?

kazuki43zoo:
I would be happy if merging.

aegliv:
I would also love to see this PR merged

snicoll:
Thanks for the PR but we're not keen on having a composite that merges the parameter names behind the scenes. The same can be accomplished by `MapSqlParameterSource`, it even has a `addValues` for such a scenario. Please consider using that, and we can continue the discussion if it turns out additional changes might be required.

virtual-machinist:
@snicoll , can you elaborate on how is `MapSqlParameterSource#addValues` a replacement for multiple `SqlParameterSource`-s that are not backed by a map?

The way I see it the proposed PR makes no assumptions about the parameter sources used - one can be backed by a map, the other by a bean using reflection etc.

snicoll:
I understand that, but it's the merging of properties that we're not keen on. It can have subtle effect and increase the maintenance cost on our end. However, applying the merge in your local setup if that makes sense can be applied.

Issue #14182, opened 2012-06-27
Avoid repeated attempts to load a failing `ApplicationContext` in the TestContext framework [SPR-9548]


spring-projects-issues:
**[Mukarram Baig](https://jira.spring.io/secure/ViewProfile.jspa?name=kilokahn)** opened **[SPR-9548](https://jira.spring.io/browse/SPR-9548?redirect=false)** and commented

Lets say I have a test class called `ServiceTest` with three test methods `test1`, `test2`, and `test3`. All three methods use some resources which are provided by Spring. In the current state of affairs, if there is a problem with the loading of the spring context, the context load is retried for each test method.

Is there a way that I can have it aborted on the first failure itself with existing hook-ins?

There might be good reasons behind it - but I fail to appreciate it. Any pointers would be helpful.

The real problem is that the context load takes a few minutes, and it is useless to keep reattempting to load the context if it has already failed for the first time and only seeks to prolong the time that the CI server takes to report failures.

I was thinking of having an `attemptMap` in `ContextCache` that the `TestContext` modifies when it encounters a failed `ApplicationContext`. A repeated load will be (on an opt-in flag set) disallowed via an exception. I have attached a rudimentary patch to indicate what I meant. Sorry that I haven't tested it (don't have gradle installed or an IDE plugin that can do it for me - am still a maven boy :))

---

**Affects:** 3.1 GA

**Reference URL:** http://stackoverflow.com/questions/11202535/ability-to-avoid-repeated-load-of-application-context-in-spring-testcontext-via

**Attachments:**
- [spring-test.patch](https://jira.spring.io/secure/attachment/20002/spring-test.patch) (_3.72 kB_)

6 votes, 8 watchers

sbrannen:
Related Issues:

- https://github.com/spring-projects/spring-boot/issues/24888

HzjNeverStop:
Is there any progress on this issue? Will it be implemented in spring framework 6.1？

snicoll:
@sbrannen is there a reason this issue is moved from one milestone to the next? (see history). I think we've decided to use `x Backlog` for this.

sbrannen:
> I think we've decided to use `x Backlog` for this.

I moved it to `6.0.x` last October: https://github.com/spring-projects/spring-framework/issues/14182#event-7555188932.

sbrannen:
If you've voted for this issue, you'll be happy to know that...

[...]

PR #24047, opened 2019-11-20
Already failed contexts should not be started again for each test

michaldo:
Before this commit, Spring Test Framework caches contexts shared among different test methods. However, context failures are not cached and failing context is started again and again for each test method the context match.

Starting a context is often a time and space consuming operation. It should be avoided when not necessary. If context fails once, there rather not possible it run second time.

This commit remembers failed context and save resources by not starting them again

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/cache/DefaultCacheAwareContextLoaderDelegate.java
+++ b/spring-test/src/main/java/org/springframework/test/context/cache/DefaultCacheAwareContextLoaderDelegate.java
@@ -28,6 +28,9 @@ import org.springframework.test.context.MergedContextConfiguration;
 import org.springframework.test.context.SmartContextLoader;
 import org.springframework.util.Assert;
 
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
 /**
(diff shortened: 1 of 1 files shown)


sbrannen:
This appears to be a PR for issue #14182.

michaldo:
Indeed, that's quite old story.

The interesting point is that the more Spring test suite is optimized, the more it suffers repeated context start.

If suboptimal suite each test has own contexts. Starting all contexts consumes much resources but failure one context does not impact performance.

Optimal suite starts one context for all tests -  it saves a lot of time. But if the contest fails, not only all tests fail, but also all tests repeat context start - suite pay back its income.

michaldo:
Recently I found a question how to limit number of contexts used in integration test: https://stackoverflow.com/questions/62621248/how-to-count-the-number-of-times-the-spring-boot-loads-on-spring-tests/65286912#65286912

The only answer I know is log number of context used, without automatic break when number exceeds the limit.

I thinking about implementation of the limit. But this PR will be useful for  the implementation

michaldo:
I had to accept that retrying failed contexts in unavoidable

michaldo:
feature independently implemented:

> Spring Framework 6.1 will finally have "test context failure threshold" support! 🍃 🥳

PR #24048, opened 2019-11-21
Revise AbstractBeanDefinition equals implementation

chenqimiao:
The following way of writing seems a bit redundant
```java
rtn = rtn &= ObjectUtils.nullSafeEquals(this.scope, that.scope);
````
I think we can polish it by writing the following.

```java
rtn &= ObjectUtils.nullSafeEquals(this.scope, that.scope);
````


sbrannen:

Thanks

dreis2211:
I wonder if that makes a difference in performance 🤔

Issue #22474, opened 2019-02-26
Add utility to parse Duration from a String


mp911de:
It would make sense to have a common utility to parse `java.time.Duration` from a `CharSequence`.

Examples:
* `10ms` -> 10 milliseconds
* `10s` -> 10 seconds
* `10m` -> 10 minutes 
* `10h` -> 10 hours

`ChronoUnit`s to support:
* Nanos
* Micros
* Millis
* Seconds
* Minutes
* Hours
* Days

[Spring Boot](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/DurationStyle.java) provides this functionality and we'd like to reuse it in Spring Data.

ilyavy:
Relates to #22013

rstoyanchev:
**Team Decision:** We can do this in fallback mode, trying the built-in `Duration` formatting, and then falling back on alternative formats, in effect extending the built-in one.

simonbasle:
linked to #22013 and superseded by gh-30396

PR #24069, opened 2019-11-25
Add flexible Duration value parsing in scheduled tasks

### LINKED ISSUE #22013: Support Spring Boot style flexible Duration value parsing [SPR-17481]
**[manderson23](https://jira.spring.io/secure/ViewProfile.jspa?name=manderson23)** opened **[SPR-17481](https://jira.spring.io/browse/SPR-17481?redirect=false)** and commented

Spring Boot supports specifying duration values in property files using values such as 15ms, 10s etc. (See https://github.com/spring-projects/spring-boot/issues/11078).

However, this means that there are inconsistencies when trying to use the values of these properties with Spring Framework annotations. An example was provided in https://github.com/spring-projects/spring-boot/issues/15126 where I would like to use a property value in a Scheduled annotation e.g.

```java
@Scheduled(fixedRateString = "${my-prop.rate}")
```

In this case I have to specify the property value using java.time.Duration syntax instead of the Spring Boot flexible syntax as `ScheduledAnnotationBeanPostProcessor` is hardcoded to use java.time.Duration parsing.

It would be good if Spring Framework also supported the flexible syntax to allow consistent property values to be used.

---

**Affects:** 5.1.2

ilyavy:
Closes gh-22013

Add flexible Duration value parsing in scheduled tasks like in Spring Boot for configuration's consistency.

Reasoning about decisions:
- The solution is based on a [corresponding commit in Spring Boot](https://github.com/spring-projects/spring-boot/commit/99afc4bc1f3e5c57fcabf6ee9551034a266bb5ab)
- The logic for parsing delay is not extracted into a separate class following the decision made before by implementors of `ScheduledAnnotationBeanPostProcessor` class
- However, it is extracted into a separate method, because now it became more complex, and in my opinion, separating it in a different method simplifies the code
- Added method `parseDelayAsLong` is not documented using `JavaDoc` because as observed from the code private methods are not documented in Spring

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
@@ -18,10 +18,12 @@ package org.springframework.scheduling.annotation;
 
 import java.lang.reflect.Method;
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


mrpiggi:
Is there any news on this? Will this feature be integrated into the framework or not?

snicoll:
Thanks for the PR but we prefer to introduce parsing of a `Duration` as a core support that can be leveraged elsewhere, see #22474

PR #24081, opened 2019-11-26
PropertyResolver extension getProperty with default value

kkocel:
This convenient metod will return non null value with reified type

### DIFF
--- a/spring-core/src/main/kotlin/org/springframework/core/env/PropertyResolverExtensions.kt
+++ b/spring-core/src/main/kotlin/org/springframework/core/env/PropertyResolverExtensions.kt
@@ -34,13 +34,23 @@ operator fun PropertyResolver.get(key: String) : String? = getProperty(key)
 
 /**
  * Extension for [PropertyResolver.getProperty] providing a `getProperty<Foo>(...)`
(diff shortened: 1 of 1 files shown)


kkocel:
@sdeleuze I've created a tiny PR with PropertyResolver extension - can you take a look?

kkocel:
@snicoll Hey, I deduct that `PropertyResolverExtensions.kt` was (re)moved? Was it moved somewhere?

snicoll:
@kkocel apologizes for this. We used GitHub's branch renaming feature and it wasn't supposed to close this PR. It would be nice to get a review from @sdeleuze before asking you to submit that code again.

sdeleuze:
I intentionally did not provide the proposed extension because Kotlin provide an idiom for that use case : `propertyResolver.getProperty<String>("name") ?: "foo"`. So I don't think it is necessary to provide a new PR, and I will fix myself the documentation typo.

sdeleuze:
See d7f67b3dc3bd989940ced850f3375474e57560f3 for the typo fix.

kkocel:
@sdeleuze @snicoll  thanks! I'm glad that it got resolved

PR #24086, opened 2019-11-26
Fix typos

KangZhiDong:
fix typos


rstoyanchev:
These all begin with a consonant and therefore "a" is correct actually.

Issue #24022, opened 2019-11-19
Protobuf MessageConverter in spring-messaging for use with STOMP


onlybytes:
Hi,

I have created a ReactorNettyTcpStompClient to connect to RabbitMQ STOMP.

My requirement is to send **protobuf ByteString** to the destination queue. I tried with all the message converters that are available under _org.springframework.messaging.converter_ package. Either i get `MessageDeliveryException: nested exception is java.lang.ClassCastException` or `MessageConversionException: Unable to convert payload with.... `

I am thinking of overriding/overloading the send() method of `DefaultStompSession `class and use it.

Is there any other easier way?

How to use protocol buffers with Spring Messaging?

Thanks,
Strive

rstoyanchev:
Currently there isn't a [MessageConverter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jms/support/converter/MessageConverter.html) in `spring-messaging` for Protobuf. The `spring-web` module however has an `ProtobufHttpMessageConverter` and it wouldn't be too hard to copy the implementation of that in order create a `MessageConverter` equivalent.

onlybytes:
I created a converter that suits my needs (took a clue from ByteArrayMessageConverter).
Definitely will have look at ProtobufHttpMessageConverter.

Thanks for adding this to 5.x backlog

Regards,
Strive

parviz-93:
Could i take this task ?

rstoyanchev:
Yes, it is marked "ideal-for-contribution" and that means we've already identified it as something that should be straight-forward to address.

parviz-93:
Please. check my [PR](https://github.com/spring-projects/spring-framework/pull/24087).
I also added a question on which I need advice

rstoyanchev:
This is now superseded by #24087.

PR #24087, opened 2019-11-26
Add protobuf MessageConverter

parviz-93:
PR for #24022.

Should I add this converter to ...?

- [ ] `SimpAnnotationMethodMessageHandler`
- [ ] `AbstractMessageBrokerConfiguration`
- [ ] `MessageBrokerBeanDefinitionParser`

I found that other converters are used in these classes.


spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
```diff
@@ -0,0 +1,368 @@
...
+
+		if (formatSupport != null) {
+			this.protobufFormatSupport = formatSupport;
+		} else if (ClassUtils.isPresent("com.googlecode.protobuf.format.FormatFactory", getClass().getClassLoader())) {
+			this.protobufFormatSupport = new ProtobufJavaFormatSupport();
+		} else if (ClassUtils.isPresent("com.google.protobuf.util.JsonFormat", getClass().getClassLoader())) {
+			this.protobufFormatSupport = new ProtobufJavaUtilSupport(null, null);
+		} else {
+			this.protobufFormatSupport = null;
+		}
```
rstoyanchev: Please check [our guidelines](https://github.com/spring-projects/spring-framework/wiki/Code-Style#formatting) on source code style.

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
```diff
@@ -0,0 +1,368 @@
...
+			throw new MessageConversionException("Could not write proto message" + e.getMessage(), e);
+
+		}
+		return payload;
+	}
+
+	@Override
+	protected boolean supports(Class<?> clazz) {
+		return Message.class.isAssignableFrom(clazz);
+	}
```
rstoyanchev: This method should be up by `canConvertTo` as it will be invoked from it.

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
```diff
@@ -0,0 +1,368 @@
...
+			if (method == null) {
+				method = clazz.getMethod("newBuilder");
+				methodCache.put(clazz, method);
+			}
+			return (Message.Builder) method.invoke(clazz);
+		} catch (Exception ex) {
+			throw new MessageConversionException(
+					"Invalid Protobuf Message type: no invocable newBuilder() method on " + clazz, ex);
+		}
+	}
```
rstoyanchev: Private methods need to appear below the first method that's using them.

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
```diff
@@ -0,0 +1,368 @@
+package org.springframework.messaging.converter;
```
rstoyanchev: Needs copyright.

spring-messaging/src/test/java/org/springframework/messaging/converter/ProtobufMessageConverterTest.java
```diff
@@ -0,0 +1,183 @@
+package org.springframework.messaging.converter;
```
rstoyanchev: Needs copyright.

rstoyanchev (changes requested):
Protobuf 2 is more than 5 years old, so for this converter being added I think we should focus on Protobuf 3+ along with the official `protobuf-java-util`. In other words eliminate anything related to `protobuf-java-format`.

/cc @sdeleuze @jhoeller for comment.

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
parviz-93: .Could you describe in more detail here? This function is called when I call `super.canConvertTo(payload, headers)` in method `canConvertTo`

rstoyanchev: It's a minor comment about the order of methods. Logically this one should be right below `canConvert` but isn't. No big deal either way. I can polish that after you're done.

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
parviz-93: added

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
parviz-93: changed

spring-messaging/src/test/java/org/springframework/messaging/converter/ProtobufMessageConverterTest.java
parviz-93: added

parviz-93:
if  we want eliminate anything related to protobuf-java-format, do we need `ProtobufFormatSupport ` interface ?

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
parviz-93: thanks. got it

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
```diff
@@ -227,9 +225,25 @@ protected Object convertToInternal(Object payload, @Nullable MessageHeaders head
...
+		try {
+			assert supports(clazz);
+
+			Method method = methodCache.get(clazz);
+			if (method == null) {
+				method = clazz.getMethod("newBuilder");
+				methodCache.put(clazz, method);
+			}
+			return (Message.Builder) method.invoke(clazz);
+		} catch (Exception ex) {
```
rstoyanchev: Correct the braces throughout, please :)

rstoyanchev:
Indeed we could simplify a bit by removing `ProtobufFormatSupport`, but I would still keep `ProtobufJavaUtilSupport` as is so that it's easy to compare to `ProtobufHttpMessageConverter` which we have to maintain in parallel.

spring-messaging/src/main/java/org/springframework/messaging/converter/ProtobufMessageConverter.java
parviz-93: Sorry. it is my first contribution to spring. thanks for correcting me

parviz-93:
I’m doing new commits so that you can see what I fixed, at the end I’ll collect in one commit.

rstoyanchev:
This looks much better. Feel free to squash the commits, I'm only looking for the the end result. 

In terms of your other question about where to add the converter to, I'd say in `AbstractMessagteBorkerConfiguration` and `MessageBorkerBeanDefinitionParser` (i.e. the Java and XML config).

rstoyanchev:
@parviz-93 thanks so much for your contribution and following through with the changes! This is now merged in. I ended up making a few further changes, adding a sub-class of the protobuf converter (just like in spring-web) because if it all came in one converter, the JSON format support wouldn't be optional. As a consequence I also decided to pull out the automatic registration in Java and XML config.

parviz-93:
Thank you very much for your help and advice.

onlybytes:
This was quick. Thanks a lot

PR #24089, opened 2019-11-27
Fix consecutive-word duplications in documentation

ryenus:
This is a follow-up of the other PR https://github.com/spring-projects/spring-security/pull/7673, which checks for consecutive words for possible duplication(s). such as `the the`, `can can` etc.


snicoll:
@ryenus thank you for making your first contribution to Spring Framework.

PR #24104, opened 2019-11-29
Add fast-path for no-args constructor in BeanUtils.instantiateClass

stsypanov:
For the cases when constructor has no args we could have fast-paths which is likely to improve start-up time as BeanUtils.instantiateClass is called at least once for each bean instantiation in application context.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -48,6 +48,7 @@ import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.ConcurrentReferenceHashMap;
+import org.springframework.util.ObjectUtils;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 
@@ -182,11 +183,18 @@ public abstract class BeanUtils {
 		try {
 			ReflectionUtils.makeAccessible(ctor);
 			if (KotlinDetector.isKotlinReflectPresent() && KotlinDetector.isKotlinType(ctor.getDeclaringClass())) {
+				if (ObjectUtils.isEmpty(args)) {
+					return KotlinDelegate.instantiateClass(ctor);
+				}
 				return KotlinDelegate.instantiateClass(ctor, args);
 			}
 			else {
+				int constructorParamCount = ctor.getParameterCount();
+				Assert.isTrue(args.length <= constructorParamCount, "Can't specify more arguments than constructor parameters");
+				if (constructorParamCount == 0) {
+					return ctor.newInstance();
+				}
 				Class<?>[] parameterTypes = ctor.getParameterTypes();
-				Assert.isTrue(args.length <= parameterTypes.length, "Can't specify more arguments than constructor parameters");
 				Object[] argsWithDefaultValues = new Object[args.length];
 				for (int i = 0 ; i < args.length; i++) {
 					if (args[i] == null) {
@@ -787,6 +795,24 @@ public abstract class BeanUtils {
 			}
 			return kotlinConstructor.callBy(argParameters);
 		}
+
+		/**
+		 * Instantiate a Kotlin class using provided no-arg constructor.
+		 * @param ctor the constructor of the Kotlin class to instantiate
+		 */
+		public static <T> T instantiateClass(Constructor<T> ctor)
+				throws IllegalAccessException, InvocationTargetException, InstantiationException {
+
+			KFunction<T> kotlinConstructor = ReflectJvmMapping.getKotlinFunction(ctor);
+			if (kotlinConstructor == null) {
+				return ctor.newInstance();
+			}
+			List<KParameter> parameters = kotlinConstructor.getParameters();
+			Assert.isTrue(parameters.isEmpty(), "Default no-args constructor must have no params");
+			Map<KParameter, Object> argParameters = Collections.emptyMap();
+			return kotlinConstructor.callBy(argParameters);
+		}
+
 
 	}
 


stsypanov:
@sdeleuze Could you have a look into this?

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
```diff
@@ -182,11 +183,18 @@
 		try {
 			ReflectionUtils.makeAccessible(ctor);
 			if (KotlinDetector.isKotlinReflectPresent() && KotlinDetector.isKotlinType(ctor.getDeclaringClass())) {
+				if (ObjectUtils.isEmpty(args)) {
+					return KotlinDelegate.instantiateClass(ctor);
```
sdeleuze, line 187: Given the fact that https://github.com/JetBrains/kotlin/pull/2855 has been closed because in practice it brings no benefit, is that part still relevant from your POV and if yes, could you elaborate on why and provide a JMH benchmark to quantify the gain?

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
```diff
@@ -182,11 +183,18 @@
...
+				}
 				return KotlinDelegate.instantiateClass(ctor, args);
 			}
 			else {
+				int constructorParamCount = ctor.getParameterCount();
+				Assert.isTrue(args.length <= constructorParamCount, "Can't specify more arguments than constructor parameters");
+				if (constructorParamCount == 0) {
+					return ctor.newInstance();
```
sdeleuze, line 195: Could you please provide a JMH benchmark of the gains?

sdeleuze (commented):
I would like to have data points for Kotlin and Java via JMH to be able to decide if we integrate such change or not, could you please measure that and share the results?

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
stsypanov: Yeah, I'll try to. The immediate gain here is that we don't allocate a new array calling `ctor.getParameterTypes()` for zero-arg constructor case

sdeleuze: I am still interested by concrete data points with JMH benchmarks before and after the proposed change, could you please provide that?

stsypanov: Hi, here's the benchmark:
```kotlin
@State(Scope.Thread)
open class MyBenchmark {
    private val noArgConstructor = TestClass1::class.java.getDeclaredConstructor()
    private val constructor = TestClass2::class.java.getDeclaredConstructor(Int::class.java, String::class.java)

    @Benchmark
    fun emptyConstructor(): Any {
        return BeanUtils.instantiateClass(noArgConstructor)
    }

    @Benchmark
    fun nonEmptyConstructor(): Any {
        return BeanUtils.instantiateClass(constructor, 1, "str")
    }

    class TestClass1()

    class TestClass2(int: Int, string: String)
}
```
The results are a bit confusing to me as for current version (6.0.0-SNAPSHOT built ad-hoc from sources) on Java 11 I get
```
Benchmark                                                         Mode  Cnt     Score     Error   Units
MyBenchmark.emptyConstructor                                      avgt  100    10,796 ?   0,145   ns/op
MyBenchmark.emptyConstructor:?gc.alloc.rate                       avgt  100  2261,834 ?  29,102  MB/sec
MyBenchmark.emptyConstructor:?gc.alloc.rate.norm                  avgt  100    32,003 ?   0,001    B/op
MyBenchmark.emptyConstructor:?gc.churn.G1_Eden_Space              avgt  100  2264,374 ?  33,918  MB/sec
MyBenchmark.emptyConstructor:?gc.churn.G1_Eden_Space.norm         avgt  100    32,036 ?   0,205    B/op
MyBenchmark.emptyConstructor:?gc.churn.G1_Survivor_Space          avgt  100     0,011 ?   0,001  MB/sec
MyBenchmark.emptyConstructor:?gc.churn.G1_Survivor_Space.norm     avgt  100    ? 10??              B/op
MyBenchmark.emptyConstructor:?gc.count                            avgt  100  2060,000            counts
MyBenchmark.emptyConstructor:?gc.time                             avgt  100  1957,000                ms
MyBenchmark.nonEmptyConstructor                                   avgt  100    15,643 ?   1,347   ns/op
MyBenchmark.nonEmptyConstructor:?gc.alloc.rate                    avgt  100  2566,194 ? 128,004  MB/sec
MyBenchmark.nonEmptyConstructor:?gc.alloc.rate.norm               avgt  100    54,405 ?   7,498    B/op
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Eden_Space           avgt  100  2568,167 ? 129,691  MB/sec
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Eden_Space.norm      avgt  100    54,450 ?   7,522    B/op
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Survivor_Space       avgt  100     0,011 ?   0,001  MB/sec
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Survivor_Space.norm  avgt  100    ? 10??              B/op
MyBenchmark.nonEmptyConstructor:?gc.count                         avgt  100  2132,000            counts
MyBenchmark.nonEmptyConstructor:?gc.time                          avgt  100  1974,000                ms
```
for patched one:
```
Benchmark                                                         Mode  Cnt     Score    Error   Units
MyBenchmark.emptyConstructor                                      avgt  100     6,901 ?  0,108   ns/op
MyBenchmark.emptyConstructor:?gc.alloc.rate                       avgt  100  1770,067 ? 27,858  MB/sec
MyBenchmark.emptyConstructor:?gc.alloc.rate.norm                  avgt  100    16,002 ?  0,001    B/op
MyBenchmark.emptyConstructor:?gc.churn.G1_Eden_Space              avgt  100  1773,080 ? 30,429  MB/sec
MyBenchmark.emptyConstructor:?gc.churn.G1_Eden_Space.norm         avgt  100    16,030 ?  0,122    B/op
MyBenchmark.emptyConstructor:?gc.churn.G1_Survivor_Space          avgt  100     0,010 ?  0,001  MB/sec
MyBenchmark.emptyConstructor:?gc.churn.G1_Survivor_Space.norm     avgt  100    ? 10??             B/op
MyBenchmark.emptyConstructor:?gc.count                            avgt  100  1741,000           counts
MyBenchmark.emptyConstructor:?gc.time                             avgt  100  1715,000               ms
MyBenchmark.nonEmptyConstructor                                   avgt  100    21,224 ?  0,303   ns/op
MyBenchmark.nonEmptyConstructor:?gc.alloc.rate                    avgt  100  3160,788 ? 43,383  MB/sec
MyBenchmark.nonEmptyConstructor:?gc.alloc.rate.norm               avgt  100    88,006 ?  0,001    B/op
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Eden_Space           avgt  100  3161,967 ? 48,861  MB/sec
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Eden_Space.norm      avgt  100    88,032 ?  0,508    B/op
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Survivor_Space       avgt  100     0,012 ?  0,001  MB/sec
MyBenchmark.nonEmptyConstructor:?gc.churn.G1_Survivor_Space.norm  avgt  100    ? 10??             B/op
MyBenchmark.nonEmptyConstructor:?gc.count                         avgt  100  2463,000           counts
MyBenchmark.nonEmptyConstructor:?gc.time                          avgt  100  2440,000               ms
```
So the case with empty constructor seems to be significantly better (and it's expected) but the opposite case has regressed, which is unexpected.

stsypanov: Sorry, I forgot to include `kotlin-reflect` into dependencies so I've measured Java reflection, not Kotlin one. Here are updated results:
```
before

Benchmark                                                         Mode  Cnt     Score    Error   Units
MyBenchmark.emptyConstructor                                      avgt  100    52,279 ?  1,089   ns/op
MyBenchmark.emptyConstructor:?gc.alloc.rate.norm                  avgt  100   140,815 ?  2,499    B/op
MyBenchmark.nonEmptyConstructor                                   avgt  100   272,787 ?  3,971   ns/op
MyBenchmark.nonEmptyConstructor:?gc.alloc.rate.norm               avgt  100   985,684 ?  4,364    B/op

after

Benchmark                                                         Mode  Cnt     Score    Error   Units
MyBenchmark.emptyConstructor                                      avgt  100    50,635 ?  0,870   ns/op
MyBenchmark.emptyConstructor:?gc.alloc.rate.norm                  avgt  100    91,853 ?  2,329    B/op
MyBenchmark.nonEmptyConstructor                                   avgt  100   265,335 ?  5,555   ns/op
MyBenchmark.nonEmptyConstructor:?gc.alloc.rate.norm               avgt  100   988,885 ?  3,272    B/op
```

stsypanov: Here's the benchmark:
```java
@State(Scope.Thread)
public class MyBenchmark {
    private Constructor<TestClass1> noArgConstructor;
    private Constructor<TestClass2> constructor;

    @Setup
    public void setUp() throws NoSuchMethodException {
        noArgConstructor = TestClass1.class.getDeclaredConstructor();
        constructor = TestClass2.class.getDeclaredConstructor(int.class, String.class);
    }

    @Benchmark
    public Object emptyConstructor() {
        return BeanUtils.instantiateClass(noArgConstructor);
    }

    @Benchmark
    public Object nonEmptyConstructor() {
        return BeanUtils.instantiateClass(constructor, 1, "str");
    }

    static class TestClass1{
    };

    static class TestClass2 {
        private final int value1;
        private final String value2;

        TestClass2(int value1, String value2) {
            this.value1 = value1;
            this.value2 = value2;
        }
    }
}
```
And results:
```
before

Benchmark                                                         Mode  Cnt     Score     Error   Units
MyBenchmark.emptyConstructor                                      avgt   40     9,197 ±   0,245   ns/op
MyBenchmark.emptyConstructor:·gc.alloc.rate.norm                  avgt   40    48,003 ±   0,001    B/op
MyBenchmark.nonEmptyConstructor                                   avgt   40    13,813 ±   0,395   ns/op
MyBenchmark.nonEmptyConstructor:·gc.alloc.rate.norm               avgt   40    72,005 ±   0,001    B/op

after

Benchmark                                                         Mode  Cnt     Score     Error   Units
MyBenchmark.emptyConstructor                                      avgt   40     4,348 ±   0,205   ns/op
MyBenchmark.emptyConstructor:·gc.alloc.rate.norm                  avgt   40    32,002 ±   0,001    B/op
MyBenchmark.nonEmptyConstructor                                   avgt   40    19,687 ±   0,312   ns/op
MyBenchmark.nonEmptyConstructor:·gc.alloc.rate.norm               avgt   40    96,006 ±   0,001    B/op

sdeleuze (commented):
@stsypanov Thanks for your patience. I am ok with the Java changes, but the Kotlin ones break some tests and introduce too much duplication. I have refined the implementation with https://github.com/sdeleuze/spring-framework/commit/9d8a035d2024519842579500ed398a0db6a5ff4f, which also makes it slightly faster (the branch is available [here](https://github.com/sdeleuze/spring-framework/tree/gh-24104)). Could you please provide me your feedback on those changes?

PR #24105, opened 2019-11-29
Enable reading from the remote build cache for all

wilkinsona:
Previously, the remote build cache was only enabled if the `GRADLE_ENTERPRISE_URL` environment variable was configured. This meant that contributors would not benefit from the build time improvements of the caching without some additional setup.

This pull request updates the `buildCache` configuration so that reading from the remote build cache at https://ge.spring.io is enabled for all. Pushing to the cache continues to be disabled unless the required credentials are provided. Build scan configuration has also been updated in line with this change. While the server URL is now hardcoded, publishing is still opt-in via an environment variable. The exact mechanism by which someone can opt in will change in the future once some server-side changes have been made. At this point, only a change to `publishAlwaysIf` should be necessary.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,16 +18,14 @@ plugins {
 	id "com.github.ben-manes.versions" version "0.24.0"
 }
 
-if (System.getenv('GRADLE_ENTERPRISE_URL')) {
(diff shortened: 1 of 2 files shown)


sbrannen:

Thanks

PR #24120, opened 2019-12-03
Optimize WebClient BodyExtractors default contentType strategy

klboke:
The ContentType of APPLICATION_OCTET_STREAM is used by default, and an exception will be thrown when using the BodyToMono to transform the Entity. I think in addition to the default strategy here, users should be free to choose which default ContentType to use

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyExtractors.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyExtractors.java
@@ -59,6 +59,8 @@ public abstract class BodyExtractors {
 
 	private static final ResolvableType VOID_TYPE = ResolvableType.forClass(Void.class);
 
+	private static MediaType defaultContentType = MediaType.APPLICATION_OCTET_STREAM;
+
 
 	/**
 	 * Extractor to decode the input content into {@code Mono<T>}.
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@klboke Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24120)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24120) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@klboke Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24120)!

rstoyanchev:
We can't allow changing the default through a static variable without risking causing a surprise between server and client(s) running within the same JVM with no option to resolve their different opinions.

A fallback on default or fixed content type needs to be more local (e.g. per request). For example, while a client could use a filter like this:

```java
ExchangeFilterFunction filter = (request, next) ->
    next.exchange(request)
        .map(response -> {
          if (response.headers().contentType().isPresent()) {
            return response;
          }
          else {
            return ClientResponse.from(response)
                .header("Content-Type", "...")  // set default content type
                .body(response.bodyToFlux(DataBuffer.class))
                .build();
          }
        });

WebClient client = WebClient.builder().filter(filter).build();
```

A server can configure [content-type resolution](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html#webflux-config-content-negotiation) or use a `WebFilter`.

PR #24135, opened 2019-12-04
Add firstElement to CollectionUtils

vitalikrst:
I was wondering why CollectionUtils has lastElement but not firstElement. It's quite handy!


jhoeller:
Thanks for the PR, this is a worthwhile addition indeed!

PR #24157, opened 2019-12-07
Regression: prototype @ControllerAdvice beans should not be cached

yokotaso:
It is ignored that the scope of bean, which is annotated `@ControllerAdvice` .

Broken at this commit.
https://github.com/spring-projects/spring-framework/commit/21267e56b7b4645e145422ee918eb75fab250f56

Bean of ControllerAdvice is wrongly cached, I think.
If Scope of Bean which is annotated `@ControllerAdvice` and 
register as a not singleton component, `ControllerAdviceBean#resolveBean()` should 
request to BeanFactory.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/ControllerAdviceBean.java
+++ b/spring-web/src/main/java/org/springframework/web/method/ControllerAdviceBean.java
@@ -54,6 +54,7 @@ public class ControllerAdviceBean implements Ordered {
 	 */
 	private final Object beanOrName;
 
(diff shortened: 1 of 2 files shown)


sbrannen:
Good catch! Thanks for submitting the PR.

sbrannen:
This has been merged into `master` in d7d474f65890a7490e9d1de989ecd96660f70cf6 and refined in fc42ca286656d7d4a910b5ea737165baa0058f5e.

Thanks

PR #24172, opened 2019-12-10
Use hasSize() where possible

izeye:
This PR changes to use `hasSize()` where possible.


snicoll:
Thanks again @izeye!

sbrannen:
Why were these changes merged?

AFAIK, we do not have any policy stating that AssertJ's `size()` method is forbidden, and I intentionally implemented those tests using that feature, because I feel it reads better.

In other words, I don't think we should be accepting PRs that introduce changes based solely on a matter of taste if the team does not have corresponding guidelines in place.

sbrannen:
Regarding my previous question, please see #24178.

Issue #22789, opened 2019-04-12
Extract MimeTypeUtils.ConcurrentLruCache and make it public


andrei-ivanov:
**Affects:** 5.1.6

---
I'm using a custom `LruCache` based on `LinkedHashMap.removeEldestEntry` and I ran a search in this repository to see if such a thing is already provided to discover there is a nice `MimeTypeUtils.ConcurrentLruCache`, but it's private.

Is there a chance to extract it somewhere else and make it public?

sbrannen:
Depending on whether that implementation is made public and whether it gets an extension point for "on removal", it could potentially replace the `org.springframework.test.context.cache.DefaultContextCache.LruCache` as well.

bclozel:
Hi @andrei-ivanov 
The team discussed this issue and we came to the conclusion that we don't want to provide such a feature in our public API, for the following reasons:

1. We provided collection-related implementation in the past, and we missed quite a few subtleties in their implementations; that required maintenance work for something we're not really in the business of in the first place (a lot of libraries provide efficient, configurable, collection implementations).
2. Extracting it would mean making it more general, applicable to more use cases and add more features. In the short term, reusing it for our own needs in the various places where we have such things is likely to make local performance worse. Each local implementation is really tailored for the case at hand.

In my opinion, libraries such as Caffeine offer other options such as Least Frequently Used cache, which is interesting if unlike Spring Framework, you can afford a mandatory dependency.
Thanks!

jhoeller:
As part of #24197, we now got a public version of `ConcurrentLruCache`, so this issue is indeed now resolved for inclusion in 5.3. In the meantime, I feel confident enough to make that class public since it got quite a bit of feedback, and it's a clean solution to the problem in `NamedParameterJdbcTemplate`. That said, this is still mostly for internal use in the framework, as all the utilities in `org.springframework.util`, so use it at your own risk :-)

jhoeller:
Freshly closing this ticket with a different resolution.

PR #24197, opened 2019-12-12
Concurrent access to ParsedSql cache in NamedParameterJdbcTemplate

benelog:
`NamedParameterJdbcTemplate.getParsedSql(Sting)` can be a bottleneck under high load as it holds a global lock by `this.parsedSqlCache`. 

```java
		synchronized (this.parsedSqlCache) {
			ParsedSql parsedSql = this.parsedSqlCache.get(sql);
			if (parsedSql == null) {
				parsedSql = NamedParameterUtils.parseSqlStatement(sql);
				this.parsedSqlCache.put(sql, parsedSql);
			}
			return parsedSql;
		}
```

This PR tried to improve the performance by the combination of a ConcurrentHashMap and a LinkedHashMap.

I referenced https://github.com/spring-projects/spring-framework/commit/06c6cbb6b92 and 
https://github.com/spring-projects/spring-framework/issues/7831#issuecomment-453315164

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterJdbcTemplate.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterJdbcTemplate.java
@@ -21,6 +21,7 @@ import java.sql.SQLException;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 
 import javax.sql.DataSource;
@@ -77,13 +78,22 @@ public class NamedParameterJdbcTemplate implements NamedParameterJdbcOperations 
 
 	private volatile int cacheLimit = DEFAULT_CACHE_LIMIT;
 
-	/** Cache of original SQL String to ParsedSql representation. */
+	/** Fast access cache for ParsedSqls, returning already cached instances without a global lock */
+	private final Map<String, ParsedSql> parsedSqlAccessCache = new ConcurrentHashMap<>(DEFAULT_CACHE_LIMIT);
+
+	/** Cache of original SQL String to ParsedSql representation, synchronized for creation */
 	@SuppressWarnings("serial")
-	private final Map<String, ParsedSql> parsedSqlCache =
+	private final Map<String, ParsedSql> parsedSqlCreationCache =
 			new LinkedHashMap<String, ParsedSql>(DEFAULT_CACHE_LIMIT, 0.75f, true) {
 				@Override
 				protected boolean removeEldestEntry(Map.Entry<String, ParsedSql> eldest) {
-					return size() > getCacheLimit();
+					if (size() > getCacheLimit()) {
+						parsedSqlAccessCache.remove(eldest.getKey());
+						return true;
+					}
+					else {
+						return false;
+					}
 				}
 			};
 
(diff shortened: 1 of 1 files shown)


rengy-github:
Is it necessary to cache twice
why need to use synchronized when using concurrenthashmap
Can just modify synchronized

			ParsedSql parsedSql = this.parsedSqlCache.get(sql);
			if (parsedSql == null) {
                                synchronized (this.parsedSqlCache) {
				    parsedSql = NamedParameterUtils.parseSqlStatement(sql);
				    this.parsedSqlCache.put(sql, parsedSql);
                               }
			}
			return parsedSql;

benelog:
@rengy-github 
`LinkedHashMap.get()` is not thread-safe when `LinkedHashMap.put()` is called from other threads, even if `LinkedHashMap.put()` is synchronized.

rengy-github:
@benelog 
get doesn't need synchronization
ParsedSql parsedSql = this.parsedSqlCache.get(sql);
if (parsedSql == null) {
	synchronized (this.parsedSqlCache) {
		parsedSql = this.parsedSqlCache.get(sql);
		if(parsedSql==null) {
			parsedSql = NamedParameterUtils.parseSqlStatement(sql);
			this.parsedSqlCache.put(sql, parsedSql);
		}
	}
}

benelog:
@rengy-github 
Reading the following issues will help you understand my PR.

https://github.com/spring-projects/spring-framework/issues/7831

rengy-github:
I  think don't use sync or just put
because There is no problem that SQL is parsed twice occasionally

rengy-github:
My puzzle is whether there is   problem with the other thread get when one thread puts

Could you send me a link about this, thank you

benelog:
@rengy-github 
[The Javadoc of LinkedHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html) will be helpful for you.

> Note that this implementation is not synchronized. If multiple threads access a linked hash map concurrently, and at least one of the threads modifies the map structurally, it must be synchronized externally. 
....
> A structural modification is any operation that adds or deletes one or more mappings or, in the case of access-ordered linked hash maps, affects iteration order. In insertion-ordered linked hash maps, merely changing the value associated with a key that is already contained in the map is not a structural modification. In access-ordered linked hash maps, merely querying the map with get is a structural modification. )

For the above reason, `Collections.synchronizedMap(Map<K,V> m)` returns a wrapper to call `Map.get(Object key)` from synchronized block.

https://github.com/AdoptOpenJDK/openjdk-jdk/blob/master/src/java.base/share/classes/java/util/Collections.java#L2634

```java
        public V get(Object key) {
            synchronized (mutex) {return m.get(key);}
        }
```

Calling `LinkedHashMap.get()` from outside the synchronized block can cause memory leaks, the problem actually happened in our company.
The following article explains the incident.

https://d2.naver.com/helloworld/1326256
(Unfortunately, it was written in Korean language)

The following artice is about similar issues in HashMap.

https://mailinator.blogspot.com/2009/06/beautiful-race-condition.html

rengy-github:
@benelog 
thank you very much，Now I understand why use a ConcurrentHashMap and a LinkedHashMap

jhoeller:
I've addressed this using a revised version our own `ConcurrentLruCache` for 5.3 RC1 now, extracted from `MimeTypeUtils` (where it got introduced in 5.2), also available for reuse in other classes.

andrei-ivanov:
now #22789 got fixed too 👍

PR #24205, opened 2019-12-13
Simplify AdvisedSupport.removeAdvisor()

stsypanov:
1) it's not necessary to call `IntroductionAdvisor.getInterfaces()` twice for each iteration
2) instead of calling `ArrayList.get()` and then `ArrayList.remove()` we can once call `ArrayList.remove()` to get and remove item by its index


sbrannen:

Thanks

PR #24245, opened 2019-12-21
Fix example in RSocket docs

fjacobs:


sbrannen:

Thanks

PR #24281, opened 2020-01-02
Honor generic type information in BeanUtils.copyProperties()

### LINKED ISSUE #24187: Honor generic type information when copying properties with BeanUtils
**Affects:** `spring-beans-5.1.2.RELEASE`

```java
class XA{}

class XB{}

class A{
    List<XA> a = new ArrayList<>();
    private String b;
    private String c;
    public List<XA> getA() {
        return a;
    }
    public void setA(List<XA> a) {
        this.a = a;
    }

    public String getB() {
        return b;
    }

    public void setB(String b) {
        this.b = b;
    }

    public String getC() {
        return c;
    }

    public void setC(String c) {
        this.c = c;
    }
}

class B {

    List<XB> a = new ArrayList<>();


    private String b;


    private String c;

    public List<XB> getA() {
        return a;
    }

    public void setA(List<XB> a) {
        this.a = a;
    }

    public String getB() {
        return b;
    }

    public void setB(String b) {
        this.b = b;
    }

    public String getC() {
        return c;
    }

    public void setC(String c) {
        this.c = c;
    }
}
```

```java
public class TestBeanUtils{
            A a = new A();
            a.getA().add(new XA());
            B b= new B();
            BeanUtils.copyProperties(a,b);
            List<XB> a1 = b.getA();
            for (XB xb : a1) {
                System.out.println("AAA");
            }
            List<XA> a2 = a.getA();
            for (XA xb : a2) {
                System.out.println("AAA");
            }
        }
}
```

run the main method , then there will throw a `java.lang.ClassCastException`;

kpkunal1406:
Fixes #24187 Honor generic type information when copying properties with BeanUtils

Handled standard supported Type Casting of properties. Any custom Object level property casting is restricted.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -19,7 +19,6 @@ package org.springframework.beans;
 import java.beans.PropertyDescriptor;
 import java.beans.PropertyEditor;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -45,7 +44,7 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.core.KotlinDetector;
 import org.springframework.core.MethodParameter;
-import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.core.ResolvableType;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
@@ -679,8 +678,7 @@ public abstract class BeanUtils {
 	}
 
 	/**
-	 * Copy the property values of the given source bean into the given target bean
-	 * and ignored if 
+	 * Copy the property values of the given source bean into the given target bean.
 	 * <p>Note: The source and target classes do not have to match or even be derived
 	 * from each other, as long as the properties match. Any bean properties that the
 	 * source bean exposes but the target bean does not will silently be ignored.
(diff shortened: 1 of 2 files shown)


sbrannen:
Thanks for the PR.

In #24187, my understanding is that we would want to **honor** generics when _matching_ properties to copy.

In other words, I don't think we actually want to **convert** properties, since that might lead to unexpected results -- for example, loss of information when converting from `Float` to `Integer` in the tests in this PR.

kpkunal1406:
@sbrannen ,

Thanks for your feedback.
If we do not convert properties then target property's datatype will remain as source (As per current implementation no need to change anything). for e.g. `List<Integer>` is copied into `List<String>`. Here generic datatype is List. But when we actually fetch it target property it will be `List<Integer>`.

And for conversion from `Float` to `Integer` in test case is as sample example. User must be knowing while copying properties from source to target (what he/she is doing). This can be taken care by the developer.

Without any conversion how can we achieve this?

sbrannen:
> or else you mean to say allow to copy only if both (source & target) property's datatypes are matching otherwise skip (no` java.lang.ClassCastException`)?

Yes, I was saying that I think we would only want to copy properties if they are compatible without conversion.

kpkunal1406:
@sbrannen ,

I have pushed new implementation to copy properties without conversion. And it copies only if source and target property's types are matching.

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
```diff
@@ -677,7 +679,8 @@ public static void copyProperties(Object source, Object target, String... ignore
 	}
 
 	/**
-	 * Copy the property values of the given source bean into the given target bean.
+	 * Copy the property values of the given source bean into the given target bean
+	 * and ignored if
```
sbrannen, line 683: It looks like this change was made in error.

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
```diff
@@ -710,17 +713,25 @@ private static void copyProperties(Object source, Object target, @Nullable Class
 			if (writeMethod != null && (ignoreList == null || !ignoreList.contains(targetPd.getName()))) {
 				PropertyDescriptor sourcePd = getPropertyDescriptor(source.getClass(), targetPd.getName());
 				if (sourcePd != null) {
+					Field sourcefield = ReflectionUtils.findField(source.getClass(), sourcePd.getName());
+					Field targetfield = ReflectionUtils.findField(target.getClass(), targetPd.getName());
```
sbrannen, line 717: There is no requirement that the components have fields with those names. Thus, this approach (i.e., working with the fields) can potentially lead to null pointer exceptions.

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
```diff
@@ -710,17 +713,25 @@ private static void copyProperties(Object source, Object target, @Nullable Class
...
+										
+					TypeDescriptor sourceTypeDescriptor = new TypeDescriptor(sourcefield);
+					TypeDescriptor targetTypeDescriptor = new TypeDescriptor(targetfield);
```
sbrannen, line 720: Working with `TypeDescriptor`s is unnecessary here since we can work directly with `ResolvableType`s.

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
```diff
@@ -710,17 +713,25 @@ private static void copyProperties(Object source, Object target, @Nullable Class
...
+							sourceTypeDescriptor.getResolvableType().equals(targetTypeDescriptor.getResolvableType())) {
 						try {
 							if (!Modifier.isPublic(readMethod.getDeclaringClass().getModifiers())) {
 								readMethod.setAccessible(true);
 							}
 							Object value = readMethod.invoke(source);
 							if (!Modifier.isPublic(writeMethod.getDeclaringClass().getModifiers())) {
 								writeMethod.setAccessible(true);
-							}
+							}
```
sbrannen, line 734: Please make sure to run `./gradlew check` before submitting PRs in order to avoid Checkstyle violations that break the build.

spring-beans/src/test/java/org/springframework/beans/BeanUtilsTests.java
```diff
@@ -330,6 +331,49 @@ void isNotSimpleProperty(Class<?> type) {
 	private void assertSignatureEquals(Method desiredMethod, String signature) {
 		assertThat(BeanUtils.resolveSignature(signature, MethodSignatureBean.class)).isEqualTo(desiredMethod);
 	}
+	
+	@Test
+	void testCopiedParametersType() {
```
sbrannen, line 336: This method does not actually test the new support for generic types.

Instead, it verifies that non-matching types are not copied, which was already the case.

sbrannen (commented):
Thanks for the modifying the original PR.

I have reviewed the changes and added comments.

Please note, however, that I do not expect you to implement the changes, since I have already done so here (https://github.com/sbrannen/spring-framework/commits/issues/gh-24187-bean-utils-copy-properties-with-generic-types).

It turned out easier for me that way. 

Once we branch for 5.3, I will merge my feature branch and keep your commit to give you credit for the work you've done to get the ball rolling with this feature enhancement.

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
kpkunal1406: Okay I see

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
kpkunal1406: yeah got that

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
kpkunal1406: yes this make sense

spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
kpkunal1406: Yes. I will make sure to run `./gradlew check`  before committing. Thanks for your guidance

spring-beans/src/test/java/org/springframework/beans/BeanUtilsTests.java
kpkunal1406: yes I considered only to not to copy non-matching types as mentioned in issue.

kpkunal1406:
@sbrannen 
I appreciate your feedback and suggestions. It is very helpful.

You have already done changes in https://github.com/sbrannen/spring-framework/commits/issues/gh-24187-bean-utils-copy-properties-with-generic-types, I am resolving review comments.
However I have pushed changes into this request also.

Thanks

sbrannen:
This has been merged into `master` in 89ee0b077f7cc16441e056200257a54374dc2545 and revised in e74f868a2397def6b3f25c96fe08fcec36c36e26.

PR #24292, opened 2020-01-04
Add ContentLength support for InputStreamResource created in ResourceHttpMessageConverter and ResourceDecoder

ofaizulin:
**Issue**
InputStreamResource created in ResourceHttpMessageConverter /ResourceDecoder doesn't properly support contentLength.
contentLength() method consumes input stream to calculate size instead of just providing predefined value.

**Motivation**
Streaming use case, where application acts as a proxy for large amount of data.

**Fix**
ResourceHttpMessageConverter should read HTTP Content-Length header.
ResourceDecoder creates ISR based on byte array, so size is always known.

**Caveats**
contentLength() for InputStreamResource created in ResourceHttpMessageConverter  can be -1 if Content-Length header is missing.


rstoyanchev:
Thanks for the pull request. I made a small adjustment to fall back on the base class in case the content-length is not present.

PR #24293, opened 2020-01-04
Improve ExposeInvocationInterceptor exception message

lixiaolong11000:
### add exposeInvocationInterceptor exception msg
- because ExposeInvocationInterceptor use ThreadLocal to expose MethodInvocation in current Thread, thus ExposeInvocationInterceptor.currentInvocation()  that invoking in other thread will throw Exception.
- some spring users face the problem cann't know the reason. for example:
_https://stackoverflow.com/questions/7147031/spring-aspect-fails-when-join-point-is-invoked-in-new-thread?r=SearchResults_

Thank you.


sbrannen:
This has been merged into `master` in 132805133d and refined in 141dae6437.

Thanks

PR #24297, opened 2020-01-06
Update Stream.map() instead of Stream.peek()

NESOY:
#### Motivation
When I read `Stream.peek()` in java 8 document, It was for debugging.
and `Stream.peek()` has side effect in Java 9.
so I updated it with `Stream.map`

#### Modification
Update `Stream.peek()` -> `Stream.map()`

This method exists mainly to support debugging, where you want to see the elements as they flow past a certain point in a pipeline:

- [Java 8 Document](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#peek-java.util.function.Consumer-)

Java 9 Issue
- [Java9 Stream Count - Peek Issue](https://docs.oracle.com/javase/9/docs/api/java/util/stream/Stream.html#count--)

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/adapter/WebHttpHandlerBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/server/adapter/WebHttpHandlerBuilder.java
@@ -240,10 +240,12 @@ public final class WebHttpHandlerBuilder {
 		}
 
 		List<WebFilter> filtersToUse = this.filters.stream()
-				.peek(filter -> {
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Thanks for the PR but the Javadoc you're referring to is for `Stream#count` which is not in use here. Using Java 11 I confirmed the code works as expected. 

Also keep in mind this code is in support of `ForwardedHeaderFilter` which is deprecated since 5.1, and will likely be removed in 5.3.

PR #24300, opened 2020-01-06
Polish ResolvableType equals implementation

chenqimiao:
Please See the following code snippet of method
`org.springframework.core.ResolvableType#forType(Type, TypeProvider, VariableResolver)`

```java
ResolvableType resultType = new ResolvableType(type, typeProvider, variableResolver);
ResolvableType cachedType = cache.get(resultType);
if (cachedType == null) {
	cachedType = new ResolvableType(type, typeProvider, variableResolver, resultType.hash);
	cache.put(cachedType, cachedType);
}
```

Currently, this variable of `cachedType` is always `null`, because the property `componentType` of `resultType` is always `null`, which causes the `equals()` method to always return `false` and `cache.get()` always returns `null`.

The purpose of PR is to fix this problem.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/ResolvableType.java
+++ b/spring-core/src/main/java/org/springframework/core/ResolvableType.java
@@ -909,7 +909,7 @@ public class ResolvableType implements Serializable {
 				!ObjectUtils.nullSafeEquals(this.variableResolver.getSource(), otherType.variableResolver.getSource()))) {
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


sbrannen (changes requested):
Please introduce a unit test that fails before the change and passes after the change.

Thanks!

chenqimiao:
@sbrannen I wrote  a unit test and found that I understood it wrong.
Sorry!

sbrannen:
> @sbrannen I wrote a unit test and found that I understood it wrong.
> Sorry!

No problem.

Thanks for the feedback.

PR #24303, opened 2020-01-07
Use SingletonSupplier to create singletons

### LINKED ISSUE #24286: Use SingletonSupplier to create singletons
[SingletonSupplier](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/function/SingletonSupplier.java) is introduced since 5.1, we could reuse this to create singleton such as [DefaultConversionService](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/convert/support/DefaultConversionService.java)

for example
```
	private static SingletonSupplier<DefaultConversionService> singletonSupplier = SingletonSupplier
			.of(DefaultConversionService::new);

	public static DefaultConversionService getSharedInstance() {
		return singletonSupplier.obtain();
	}
```

quaff:
Fix https://github.com/spring-projects/spring-framework/issues/24286

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/ReactiveAdapterRegistry.java
+++ b/spring-core/src/main/java/org/springframework/core/ReactiveAdapterRegistry.java
@@ -36,6 +36,7 @@ import rx.RxReactiveStreams;
 import org.springframework.lang.Nullable;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.ReflectionUtils;
+import org.springframework.util.function.SingletonSupplier;
 
 /**
  * A registry of adapters to adapt Reactive Streams {@link Publisher} to/from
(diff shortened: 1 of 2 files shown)


sbrannen:
@jhoeller, what do you think about using `SingletonSupplier` for use cases like these?

quaff:
@sbrannen , please feedback since you adopt `SingletonSupplier` in https://github.com/spring-projects/spring-framework/commit/48052881222bddae82ddbf0a8a242e12a79db490

jhoeller:
This is well spotted in general, however, I'm rather defensive about core facilities there. At such a low level, I generally prefer to keep the code as straightforward as possible... for memory effectiveness, for inlining in HotSpot, etc. We apply similar reasoning for Java 8 Stream usage in our code: avoiding it in core places but selectively embracing it for readability higher up the stack.

So while we don't have strict rules in place here, I'd nevertheless avoid `SingletonSupplier` usage in `spring-core` itself. It's primarily meant for higher-level integration code such as our scheduling and cache integration; using it if for `XStreamMarshaller` seems appropriate from that perspective.

PR #24311, opened 2020-01-07
Add conversion support for Pattern/Regex

valfirst:


sbrannen:
@valfirst, can you please provide some background on the rationale for your proposal?

Do you have a particular use case that requires this support?

valfirst:
@sbrannen thank you for quick response!
We use conversion service to perform type conversion: https://github.com/vividus-framework/vividus/blob/master/vividus/src/main/resources/org/vividus/spring-configuration.xml#L51. 
At some point we faced with necessity to convert `String` value from the property to `Optional<Pattern>` type. One of the proposed solutions was to re-use all property editors: https://github.com/vividus-framework/vividus/pull/185/commits/cbe84d8234b0d2055dcd0c9c2f2ae6f7791900dc. For me this solution looks awkward.
I haven't found existing converter for `Pattern` type in Spring, so I've implemented own one and add it to our conversion service. It works as expected. I think the converter looks generic enough and could be useful for other users, so I've decided to create this PR.

sbrannen:
Thanks for providing the feedback on your use case.

We'll take that into consideration.

valfirst:
@sbrannen could you please review the changes when you have a chance? thanks

sokomishalov:
I think it makes sense to provide the same converter from String to `kotlin.text.Regex` if kotlin is present in the classpath due to closing the similar PR in the boot repository. 
https://github.com/spring-projects/spring-boot/pull/23706

ikalinin1:
@sbrannen  could you please take a look at this PR

snicoll:
Thanks @valfirst, I polished the proposal by adding support for Kotlin's  `Regex` as well.

PR #24316, opened 2020-01-08
Fix race condition for InputStreamResource

quaff:
Use AtomicBoolean instead of primitive boolean


jhoeller:
Do you have a specific use case where `InputStreamResource` is used in such a concurrent scenario? Generally, `InputStreamResource` is not really designed for concurrent use... not least of it all since the provided `InputStream` can only be read once in any case. From that perspective, defensive checks for concurrent scenarios seem unnecessary or even misleading since they suggest concurrent usability to begin with.

PR #24332, opened 2020-01-11
Polish populateBean implementation of AbstractAutowireCapableBeanFactory

chenqimiao:
From the context point of view, this property of filteredPds  must be null at the place I modified


sbrannen:
Thanks for the PR.

> From the context point of view, this property of `filteredPds` must be null at the place I modified

It will be `null` the first time that code path executes, but on subsequent iterations of the for-loop it may no longer be `null`.

Thus, the construct in place here is lazy initialization of the array. That is by design, and the proposal in this PR breaks this.

I am therefore closing this PR as invalid.

chenqimiao:
@sbrannen Thank you for pointing out my mistake

PR #24333, opened 2020-01-11
Make resolveAutowiringValue method of AutowireUtils support abstract class parameter

chenqimiao:
If the argument `requiredType ` of `resolveAutowiringValue` is an abstract class  and meet some other conditions, I think the argument that method return can be an object of subclass to the argument `requiredType`.

So I took the liberty to add some implementations, hoping to meet expectations.


sbrannen:
@chenqimiao, what would be the use case for such behavior?

chenqimiao:
@sbrannen 
Sorry for my negligence.
I wrote an example module for this pr.
Please review if you have time.
The address of the demo is 	https://github.com/chenqimiao/spring-framework/blob/demo/spring-demo/src/main/java/com/github/chenqimiao/Main.java
Just start the main method.

chenqimiao:
In other words,I want the parameter `dependencyType` of 
method `org.springframework.beans.factory.config.ConfigurableListableBeanFactory#registerResolvableDependency` can support abstract classes, and it can be auto-injected via convention-based autowiring by type. @sbrannen

PR #24350, opened 2020-01-14
Fix variable to static in OrderSourceProviderTests

hyeonisism:
Fix to static final


hyeonisism:
Thank you for sharing your ideas. I learned a lot from this PR.

PR #24356, opened 2020-01-14
continue for loop if the bean definition cannot be enhanced

chenqimiao:
As the log says,
`Cannot enhance @Configuration bean definition since its singleton instance has been created too early. The typical cause is a non-static @Bean method with a BeanDefinitionRegistryPostProcessor`.
So, I think this `if` block should `continue` for loop.


snicoll:
Duplicate of #22992

PR #24371, opened 2020-01-15
Publish a build scan only if authenticated

wilkinsona:
Previously, anyone could publish a scan and publishing was performed if the `GRADLE_ENTERPRISE_URL` environment variable was set. ge.spring.io has now been locked down to prohibit anonymous build scan publishing. This commit aligns with this change by only attempting to publish a build scan when authenticated with the server. Authentication is achieved via an access token that is made available via an environment variable on CI and locally via a file in `~/.gradle`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,8 @@ buildScan {
 	obfuscation {
 		ipAddresses { addresses -> addresses.collect { address -> '0.0.0.0'} }
 	}
-	publishAlwaysIf(System.getenv('GRADLE_ENTERPRISE_URL') != null)
+	publishAlways()
+	publishIfAuthenticated()
 	server = 'https://ge.spring.io'
 }
 


wilkinsona:
Please hold off on merging this. `publishIfAuthenticated()` does not appear to work as intended.

wilkinsona:
This is now ready to merge. If you want local builds to publish scans, you can obtain an access key by running `./gradlew provisionGradleEnterpriseAccessKey`.

sbrannen:

Thanks

snicoll:
@wilkinsona should the command above be invoked for every project or running it once is enough?

wilkinsona:
Once is enough for us as every project points to ge.spring.io.

etiennestuder:
> So it looks like we just need to do that once locally for all Spring projects using that server.

Yes, I can confirm that.

PR #24377, opened 2020-01-16
Set MessageProducer's deliveryMode after create Producer

ivyxjc:
I find out that the producer does not have the same deliveryMode that the JmsTemplate has.


snicoll:
Thanks for the PR. The Javadoc of `setDeliveryMode` states that it is only used with QoS:

> Since a default value may be defined administratively, this is only used when "isExplicitQosEnabled" equals "true".

`createProducer` being used internally by `JmsTemplate`, that hopefully explains why it isn't set on the producer. If you want to use `JmsTemplate` and override the QoS attribute, you'll need to invoke `setExplicitQosEnabled(true)` as well.

PR #24378, opened 2020-01-16
Simplify hasNext method() in compositeIterator

hyeonisism:
It can simplify this way


sbrannen:
Thanks for the PR.

The proposal is certainly more succinct; however, we generally avoid the use of the `Stream` API in potential hot spots within the core and instead favor plain-old iterative loops.

PR #24379, opened 2020-01-16
Use dedicated catch block for AnnotationConfigurationException

hyeonisism:
It can be separate.


sbrannen:
Thanks for the PR.

You're correct: that could be pulled out into a separate `catch` block; however, I've taken a slightly different approach in daa30a9f0b6bcc33cbb452e7a1df27776acfa935 with the goal of consistent handling of such exceptions within that package.

PR #24381, opened 2020-01-16
Polish constants name in jta transaction manager factory bean

hyeonisism:
Modified because this did not match the uppercase rule of the class name.


rstoyanchev (commented):
Thanks for the PR but both WebLogic and WebSphere are single names, so no underscore in the middle.

hyeonisism:
@rstoyanchev thanks to the reply. but I want to know that the class should also be "WeblogicJtaTransactionManager".

rstoyanchev:
Sorry, but I don't understand. The class supports both WebLogic and WebSphere.

PR #24383, opened 2020-01-17
Use dedicated catch block for ScriptException

hyeonisism:
It can be that.


spring-jdbc/src/main/java/org/springframework/jdbc/datasource/init/DatabasePopulatorUtils.java
```diff
@@ -52,10 +52,10 @@ public static void execute(DatabasePopulator populator, DataSource dataSource) t
 				DataSourceUtils.releaseConnection(connection, dataSource);
 			}
 		}
+		catch (ScriptException ex){
```
sbrannen: Please note that the above is missing a space: `) {`, but I'll fix that after merging.

sbrannen:

Thanks

PR #24385, opened 2020-01-17
Use try-with-resource in ScriptUtils

hyeonisism:
That can use try-with-resource.


sbrannen:

Thanks

PR #24386, opened 2020-01-17
Use computeIfAbsent to simplify getSessionAttributesHandler()

hyeonisism:
It can simplify using computeIfAbesnt.


rstoyanchev:
I revised this further to remove the surrounding synchronized block since computeIfAbsent on ConcurrentHashMap is already synchronized internally.

PR #24388, opened 2020-01-17
Hoist concatenation of two constant Strings out of loops

stsypanov:
In cases when concatenation arguments are the same they produce equal String at each iteration. This is why we can hoist them to reduce output of String garbage.


sbrannen:

Thanks

PR #24393, opened 2020-01-17
Add close() method in FileCopyUtils to reduce duplication

hyeonisism:
That can be reduced by the close method.


sbrannen:

Thanks

PR #24397, opened 2020-01-19
Polish copyConfigurationFrom implementation of AdvisedSupport

chenqimiao:
Polish copyConfigurationFrom implementation of AdvisedSupport


chenqimiao:
The modified code is redundant with the existing method named `addAdvisors`. 
So, I think we can reuse this method.

PR #24398, opened 2020-01-20
Add more lambda expressions in data-access code examples

maxxedev:
add more java8 lambda expressions in documentation for data-access

### DIFF
--- a/src/docs/asciidoc/data-access.adoc
+++ b/src/docs/asciidoc/data-access.adoc
@@ -2715,7 +2715,7 @@ The following query looks for a `String`:
 ----
 	String lastName = this.jdbcTemplate.queryForObject(
 			"select last_name from t_actor where id = ?",
-			new Object[]{1212L}, String.class);
+			String.class, 1212L);
 ----
 [source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
 .Kotlin
@@ -2730,17 +2730,15 @@ The following query finds and populates a single domain object:
 [source,java,indent=0,subs="verbatim,quotes",role="primary"]
 .Java
 ----
+	RowMapper<Actor> actorRowMapper = (resultRest, rowNum) -> {
+		Actor actor = new Actor();
+		actor.setFirstName(resultRest.getString("first_name"));
+		actor.setLastName(resultRest.getString("last_name"));
+		return actor;
+	};
 	Actor actor = this.jdbcTemplate.queryForObject(
 			"select first_name, last_name from t_actor where id = ?",
-			new Object[]{1212L},
-			new RowMapper<Actor>() {
-				public Actor mapRow(ResultSet rs, int rowNum) throws SQLException {
-					Actor actor = new Actor();
-					actor.setFirstName(rs.getString("first_name"));
-					actor.setLastName(rs.getString("last_name"));
-					return actor;
-				}
-			});
+			actorRowMapper, 1212L);
 ----
 [source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
 .Kotlin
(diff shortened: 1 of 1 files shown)


src/docs/asciidoc/data-access.adoc
```diff
@@ -5567,16 +5558,14 @@ declared type `ITEM_TYPE`:
 		public TestItemStoredProcedure(DataSource dataSource) {
 			// ...
 			declareParameter(new SqlOutParameter("item", OracleTypes.STRUCT, "ITEM_TYPE",
-				new SqlReturnType() {
-					public Object getTypeValue(CallableStatement cs, int colIndx, int sqlType, String 	) throws SQLException {
```
maxxedev: missing param name in `String typeName`

src/docs/asciidoc/data-access.adoc
```diff
@@ -2730,17 +2730,15 @@ The following query finds and populates a single domain object:
 [source,java,indent=0,subs="verbatim,quotes",role="primary"]
 .Java
 ----
+	RowMapper<Actor> actorRowMapper = (resultRest, rowNum) -> {
```
sbrannen: `resultRest` --> `resultSet`

Please make that change consistently in this PR.

sbrannen: Please inline the lambda expression as was previously the case with the anonymous inner class, since that then aligns with the explanatory text in the following paragraph.

src/docs/asciidoc/data-access.adoc
```diff
@@ -2757,16 +2755,15 @@ The following query finds and populates a number of domain objects:
...
```
sbrannen: Please inline this lambda expression as well. (see previous comment)

src/docs/asciidoc/data-access.adoc
```diff
@@ -2777,26 +2774,25 @@ The following query finds and populates a number of domain objects:
...
 ----
 	public List<Actor> findAllActors() {
-		return this.jdbcTemplate.query( "select first_name, last_name from t_actor", new ActorMapper());
+		return this.jdbcTemplate.query( "select first_name, last_name from t_actor", getActorMapper());
 	}
 
-	private static final class ActorMapper implements RowMapper<Actor> {
-
-		public Actor mapRow(ResultSet rs, int rowNum) throws SQLException {
+	private RowMapper<Actor> getActorMapper() {
```
sbrannen: The `RowMapper` implementation is stateless. Therefore, please store a single reference to it in a `final` field and remove the `getActorMapper()` method.

src/docs/asciidoc/data-access.adoc
```diff
@@ -3562,15 +3558,12 @@ on Oracle but may not work on other platforms:
...
-				return ps;
-			}
-		},
-		keyHolder);
+	PreparedStatementCreator statementCreator = connection -> {
+		PreparedStatement ps = connection.prepareStatement(INSERT_SQL, new String[] { "id" });
+		ps.setString(1, name);
+		return ps;
+	};
+	jdbcTemplate.update(statementCreator, keyHolder);
```
sbrannen: Please inline the lambda expression.

sbrannen (changes requested):

src/docs/asciidoc/data-access.adoc
maxxedev: Ok, stored as instance field `actorRowMapper`. Let know static field `ACTOR_ROW_MAPPER` is preferred

maxxedev:
@sbrannen anything else?

src/docs/asciidoc/data-access.adoc
sbrannen: That's true, but this is in the documentation that is being replaced. Thus we can ignore that for now.

sbrannen:
> @sbrannen anything else?

Thanks for making the changes. I think that looks pretty good now, and I'll review it locally before merging.

sbrannen:
This has been merged into `master` in 9b827283f1502c7bf7d517597cad47ce26944625 and refined in 7dea2686b81509f7d42901da5f0d68ae2e69bee9.

Thanks

sbrannen:
As a side note, the last Kotlin example in [this section](https://github.com/spring-projects/spring-framework/blob/master/src/docs/asciidoc/data-access.adoc#querying-select) still uses a local class instead of a lambda expression.

I'll see if we can that addressed as well.

maxxedev:
thanks for the review!

the [newer example](https://github.com/spring-projects/spring-framework/commit/7dea2686b81509f7d42901da5f0d68ae2e69bee9#r37967323) may not compile though.

sbrannen:
> thanks for the review!

You're welcome.

> the [newer example](https://github.com/spring-projects/spring-framework/commit/7dea2686b81509f7d42901da5f0d68ae2e69bee9#r37967323) may not compile though.

Fixed in b52136dfa5fcda00a6a5073470c6751ce96e1953.

Thanks again for pointing that out! 👍

PR #24404, opened 2020-01-20
Simplify canConvertElements method() in ConversionUtils class

hyeonisism:
It can simplify this way.


rstoyanchev:
Thanks but we'll keep this one the way it is for the implementation comments.

Issue #22083, opened 2018-11-29
Add String based URI template variants to RequestEntity  [SPR-17551]


spring-issuemaster:
**[Ryon Day](https://jira.spring.io/secure/ViewProfile.jspa?name=ryonday)** opened **[SPR-17551](https://jira.spring.io/browse/SPR-17551?redirect=false)** and commented

The `RestTemplateBuilder` documentation for the `rootUri()` method states:

"Set a root URL that should be applied to each request that starts with '/'."

However, the following code does not result in `RestTemplate` expanding the URI:

```java
UriTemplate template = new UriTemplate("/foo?bar={bar}");

this.restTemplate = restTemplateBuilder
 .rootUri(URI.create("http://foo.bar"))
 .build();

URI uri = template.expand("baz");
RequestEntity<Void> req = RequestEntity.get(uri)
 .accept(MediaType.APPLICATION_JSON)
 .header(accountLinkHeader, accountLinksHeader)
 .build();
String response = restTemplate.exchange(req, String.class);
```

 

Moreover, the exception received is EXTREMELY ambiguous when this doesn't work. In fact, the Spring exception thrown occludes the initial Apache exception (which states there's no host).

---

**Affects:** 5.1.3

aberkecz:
Hi,

I would love to help out with this issue if it's up for grabs. I spent some time investigating it but got stuck a bit when trying to figure out a nice solution.

So far, I've created an example project to see if the issue is still valid and found out that it is. However, the Javadoc for `RestTemplateBuilder#rootUri` has been clarified to describe the above behaviour. 

 Now, I’m a bit confused because I’m not entirely sure what would be a nice way to do the expansion when a `RequestEntity` is used. (Do I extend the `RequestEntity` with new public methods expecting a String templateUri? But how do I expand it there when I only have the `rootUri` set in the `RestTemplate` instance?)

Would it be helpful if I tried and submitted a PR for this? If so, could you please help me with some guiding on how to get started?

Thanks!

rstoyanchev:
@aberkecz thanks for considering this and yes this is marked "ideal-for-contribution".

For the implementation, I can suggest something like this:
1. Add extra static RequestEntity methods (per HTTP method) that accept a String-based URI template.
1. Add methods in the builder (`RequestEntity.HeadersBuilder` probably) that accept URI variables as either vararg or a Map.
[...]

PR #24406, opened 2020-01-20
Support URI templates in RequestEntity

parviz-93:
Add String based URI template and vararg to builder #22083

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/RequestEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/RequestEntity.java
@@ -25,12 +25,18 @@ import java.util.Arrays;
 import java.util.function.Consumer;
 
 import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
 import org.springframework.util.MultiValueMap;
 import org.springframework.util.ObjectUtils;
+import org.springframework.web.util.DefaultUriBuilderFactory;
+import org.springframework.web.util.UriTemplateHandler;
 
 /**
  * Extension of {@link HttpEntity} that adds a {@linkplain HttpMethod method} and
- * {@linkplain URI uri}. Used in {@code RestTemplate} and {@code @Controller} methods.
+ * {@linkplain URI uri} or String based uri template with placeholder. Uri variables can be provided
+ * through the builder {@link HeadersBuilder#uriVariables(Object...)}.
+ *
+ * Used in {@code RestTemplate} and {@code @Controller} methods.
  *
  * <p>In {@code RestTemplate}, this class is used as parameter in
  * {@link org.springframework.web.client.RestTemplate#exchange(RequestEntity, Class) exchange()}:
@@ -54,6 +60,19 @@ import org.springframework.util.ObjectUtils;
  * RequestEntity&lt;MyRequest&gt; request = RequestEntity.post(uri).accept(MediaType.APPLICATION_JSON).body(body);
  * </pre>
  *
+ * <p> if you would like to provide a string base URI template with variable, consider using
+ * {@link RequestEntity#method(HttpMethod, String)}}, {@link RequestEntity#get(String)},
+ * {@link RequestEntity#post(String)}, {@link RequestEntity#put(String)}, {@link RequestEntity#delete(String)},
+ * {@link RequestEntity#patch(String)}, {@link RequestEntity#options(String)}
+ *
+ * <pre class="code>
+ *    RequestEntity  request = RequestEntity
+ *    		.post(&quot;https://example.com/{foo}&quot;)
+ *    	    .uriVariables(&quot;bar&quot;)
+ *          .accept(MediaType.APPLICATION_JSON)
+ *          .build();
+ * </pre>
+ *
  * <p>Can also be used in Spring MVC, as a parameter in a @Controller method:
  * <pre class="code">
  * &#64;RequestMapping("/handle")
@@ -66,6 +85,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Arjen Poutsma
  * @author Sebastien Deleuze
+ * @author Parviz Rozikov
  * @since 4.1
  * @param <T> the body type
  * @see #getMethod()
@@ -73,10 +93,19 @@ import org.springframework.util.ObjectUtils;
  */
 public class RequestEntity<T> extends HttpEntity<T> {
 
+	private final static UriTemplateHandler DEFAULT_URI_BUILDER_FACTORY = new DefaultUriBuilderFactory();
+
 	@Nullable
 	private final HttpMethod method;
 
+	@Nullable
 	private final URI url;
+
+	@Nullable
+	private String uri;
+
+	@Nullable
+	private Object[] uriVariables;
 
 	@Nullable
 	private final Type type;
@@ -146,12 +175,30 @@ public class RequestEntity<T> extends HttpEntity<T> {
 	 * @since 4.3
 	 */
 	public RequestEntity(@Nullable T body, @Nullable MultiValueMap<String, String> headers,
-			@Nullable HttpMethod method, URI url, @Nullable Type type) {
+						 @Nullable HttpMethod method, URI url, @Nullable Type type) {
 
 		super(body, headers);
 		this.method = method;
 		this.url = url;
 		this.type = type;
+	}
+
+	/**
+	 * Private Constructor with method, URL, UriTemplate and varargs urivariables but without body nor headers.
+	 * @param method the method
+	 * @param url the URL
+	 * @param uri the UriTemplate
+	 * @param uriVariables the uriVariables
+	 */
+	private RequestEntity(MultiValueMap<String, String> headers, HttpMethod method, @Nullable URI url,
+						  @Nullable  String uri, @Nullable Object... uriVariables) {
+		super(null, headers);
+		Assert.isTrue(uri == null || url == null, "Either url or url must be not null");
+		this.method = method;
+		this.url = url;
+		this.type = null;
+		this.uri = uri;
+		this.uriVariables = uriVariables;
 	}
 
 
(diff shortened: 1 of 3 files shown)


spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -169,9 +195,27 @@ public HttpMethod getMethod() {
...
 	 */
 	public URI getUrl() {
-		return this.url;
+		if (uriTemplate == null) {
+			return this.url;
+		}
+
+		final DefaultUriBuilderFactory defaultUriBuilderFactory = new DefaultUriBuilderFactory();
+		defaultUriBuilderFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.URI_COMPONENT);
+		return defaultUriBuilderFactory.expand(uriTemplate.toString(), uriVariables);
```
rstoyanchev: This factory be a default instance, not customized, and it should be a static variable so it is not re-created for every request.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -138,22 +148,38 @@ public RequestEntity(@Nullable T body, @Nullable MultiValueMap<String, String> h
...
 	 * Constructor with method, URL, headers, body and type.
-	 * @param body the body
+	 * @param body    the body
 	 * @param headers the headers
-	 * @param method the method
-	 * @param url the URL
-	 * @param type the type used for generic type resolution
+	 * @param method  the method
+	 * @param url     the URL
+	 * @param type    the type used for generic type resolution
```
rstoyanchev: This alignment change is inconsistent with pretty much all of the Spring Framework codebase. Please correct it everywhere.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -138,22 +148,38 @@ public RequestEntity(@Nullable T body, @Nullable MultiValueMap<String, String> h
...
+	 * @param uriVariables the uriVariables
+	 */
+	private RequestEntity(MultiValueMap<String, String> headers,HttpMethod method, URI url, UriTemplate uriTemplate, Object... uriVariables) {
+		super(null, headers);
+		this.method = method;
+		this.url = url;
+		this.type = null;
+		this.uriTemplate = uriTemplate;
+		this.uriVariables = uriVariables;
+	}
```
rstoyanchev: Here you'll find our [Source Code Style](https://github.com/spring-projects/spring-framework/wiki/Code-Style) guidelines. Specifically the section on line wrapping (aim for 90, no more than 120, but only 80 for Javadoc).

rstoyanchev (changes requested):
Overall on the right path.

In addition to line specific comments, this is lacking the change where the `RestTemplate` uses the new `getUrl(UriTemplateHandler)` method instead of `getUrl()` which would add the benefits of a pre-configured base URL among others. That was the following step from the instructions:

> 6. Make use of this new method in RestTemplate.

Keep in mind also this cannot be merged at the moment until we switch master to 5.3 development.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
parviz-93: done

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -250,6 +305,15 @@ public static BodyBuilder method(HttpMethod method, URI url) {
...
 
+	/**
+	 * Create an HTTP GET builder with the given string base uri template.
+	 * @param uriTemplate the uri template
+	 * @return the created builder
+	 */
+	public static HeadersBuilder<?> get(UriTemplate uriTemplate) {
+		return method(HttpMethod.GET, uriTemplate);
+	}
+
```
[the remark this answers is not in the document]
parviz-93: done

spring-web/src/main/java/org/springframework/http/RequestEntity.java
parviz-93: done

spring-web/src/main/java/org/springframework/http/RequestEntity.java
parviz-93: done

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -241,6 +285,17 @@ public static BodyBuilder method(HttpMethod method, URI url) {
...
 	}
 
+	/**
+	 * Create a builder with the given method and given string base uri template.
+	 * @param method the HTTP method (GET, POST, etc)
+	 * @param uri the uri
+	 * @return the created builder
+	 */
+	public static BodyBuilder method(HttpMethod method, String uri) {
+		return new DefaultBodyBuilder(method, new UriTemplate(uri));
```
rstoyanchev: But why go to `UriTemplate` here, only to go back to `String` in the end in `getUrl()`? There is no benefit nor reason to create a `UriTemplate` at any point in this class.

rstoyanchev (changes requested):

spring-web/src/main/java/org/springframework/http/RequestEntity.java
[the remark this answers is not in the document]
rstoyanchev: Not completely done. See comment further below.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
parviz-93: I was dumb 😔. I will fix it

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -146,14 +157,31 @@ public RequestEntity(@Nullable T body, @Nullable MultiValueMap<String, String> h
...
+	 */
+	private RequestEntity(MultiValueMap<String, String> headers, HttpMethod method, URI url,
+						  String uri, Object... uriVariables) {
+		super(null, headers);
+		this.method = method;
+		this.url = url;
+		this.type = null;
+		this.uri = uri;
+		this.uriVariables = uriVariables;
+	}
```
rstoyanchev: The `url`, `uri`, and `uriVariables` arguments should be `@Nullable` and `org.springframework.util.Assert` should be used to check that either `url` or `uri` is provided. Otherwise the `getUrl` methods below would fail.

spring-web/src/test/java/org/springframework/http/RequestEntityTests.java
```diff
@@ -79,6 +81,16 @@ public void uriVariablesExpansion() throws URISyntaxException {
...
 
+	@Test
+	public void uriExpansion() throws URISyntaxException{
+		String template = "/foo?bar={bar}";
+		String bar = "foo";
+		final RequestEntity<Void> entity = RequestEntity.get(template).uriVariables(bar).build();
+
+		assertThat(entity.getUrl()).isEqualTo(new URI("/foo?bar=foo"));
+	}
+
```
rstoyanchev: This is lacking a test to demonstrate expanding with a `UriTemplateHandler`, e.g. one that has a baseUrl. That was actually the original motivation for the change.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -241,6 +284,17 @@ public static BodyBuilder method(HttpMethod method, URI url) {
...
 
+	/**
+	 * Create a builder with the given method and given string base uri template.
+	 * @param method the HTTP method (GET, POST, etc)
+	 * @param uri the uri
+	 * @return the created builder
+	 */
+	public static BodyBuilder method(HttpMethod method, String uri) {
+		return new DefaultBodyBuilder(method, uri);
+	}
```
rstoyanchev: All methods that take a `String` uri need to mention in the Javadoc that this is a URI template that may contain URI variable placeholders and mention that URI variables can be provided through the builder either as a vararg or as a map (probably best to `@link` to the corresponding builder methods). 

It be useful also to link to `getUrl()` which would expand through `DefaultUriBuilderFactory` and encode as explained in `DefaultUriBuilderFactory#setEncodingMode`, unless the `UriTemplateHandler` is provided.

To avoid repeating this in many places, you could document at the class level that there are `URI` and `String` uri based factory methods, and explain how the `String` uri variants work. Then refer to class-level Javadoc from each factory method.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -169,9 +197,24 @@ public HttpMethod getMethod() {
...
+
+	/**
+	 * Return the URL of the request.
+	 * @return the URL as a {@code URI}
+	 */
+	public URI getUrl(UriTemplateHandler uriTemplateHandler) {
+		if (uri == null) {
+			return this.url;
+		}
+		return uriTemplateHandler.expand(uri, uriVariables);
```
rstoyanchev: All new public methods should have `@since 5.3` tags in the Javadoc.

rstoyanchev (commented):
Making good progress. I've added more comments.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
parviz-93: Sorry for pause. I had a rest.
I added javadocs, but maybe also added grammatical errors, because I'm not a native English speaker

rstoyanchev: Okay thanks.

rstoyanchev:
Thanks for all the updates! This is now in good shape for inclusion. When master is switched to 5.3 development I'll process this right away.

parviz-93:
@rstoyanchev thank you for review. if there are issues which your hands do not reach, I am always ready to help

rstoyanchev:
Thanks! Keep an eye on the "ideal-for-contribution" label.

jhoeller:
@rstoyanchev I'm afraid we have do revisit this one since it turned out to create a package cycle between `org.springframework.http` (where `RequestEntity` lives) and `org.springframework.web` (where `UriTemplateHandler` lives). We'll have to put this `getUrl` overload into some utility method or introduce a dedicated `RequestEntity` subclass in the `org.springframework.web` package which may then integrate with specific facilities such as `UriTemplateHandler` there.

rstoyanchev:
This should be taken care of now.

PR #24409, opened 2020-01-21
Improve deleteRecursively() method in FileSystemUtil

hyeonisism:
Performance is too flawed in Java8. So I changed it.


spring-core/src/main/java/org/springframework/util/FileSystemUtils.java
```diff
@@ -77,7 +77,7 @@ public static boolean deleteRecursively(@Nullable Path root) throws IOException
 		if (root == null) {
 			return false;
 		}
-		if (!Files.exists(root)) {
+		if (!root.toFile().exists()){
```
marschall, line 80: That breaks if the path is not on the default file system.

PR #24429, opened 2020-01-26
odata uses MERGE method for updating Entity

met-pub:
If we dont add MERGE HttpMethod here, then [preflight request with asking MERGE method will be rejected](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/cors/DefaultCorsProcessor.java#L117).

https://www.odata.org/documentation/odata-version-2-0/operations/
2. Operations
The OData service interface has a fixed number of operations that have uniform meaning across all the resources it can act on. These operations are retrieve, create, update and delete and they map to the GET, POST, PUT/MERGE and DELETE HTTP methods.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpMethod.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpMethod.java
@@ -32,7 +32,7 @@ import org.springframework.lang.Nullable;
  */
 public enum HttpMethod {
 
-	GET, HEAD, POST, PUT, PATCH, DELETE, OPTIONS, TRACE;
+	GET, HEAD, POST, PUT, PATCH, DELETE, OPTIONS, TRACE, MERGE;
 
 
 	private static final Map<String, HttpMethod> mappings = new HashMap<>(16);


pivotal-issuemaster:
@bianbian-org Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24429)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24429) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@bianbian-org Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24429)!

rstoyanchev (changes requested):
To add `MERGE` to the `HttpMethod` enum would imply support for it, but we don't actually support it nor any other custom HTTP methods, which I believe `MERGE` to b e. It's not clear whether that change alone would be sufficient either but regardless I don't think adding it to the enum is the way to go.

I see that `DefaultCorsProcessor` calls `HttpRequest` which has a `getMethodValue()` method and likewise `HttpHeaders` could be used to get the header value without turning it into `HttpMethod`. Perhaps you can start by trying to change `DefaultCorsProcessor` to use that method instead and see if that gets you far enough.

bianbian-org:
OK. Finally I changed odata client to send PUT method.
[seems PUT/MERGE apply equally](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-odata/59d5abd3-7b12-490a-a0e2-9d9324b91893)

PR #24436, opened 2020-01-27
Support message conversion with kotlinx.serialization

### LINKED ISSUE #21188: Add kotlinx.serialization JSON support to Spring MVC
**[Alex Rader](https://jira.spring.io/secure/ViewProfile.jspa?name=rrader)** opened **[SPR-16647](https://jira.spring.io/browse/SPR-16647?redirect=false)** and commented

Please add support for Kotlin serialization library

https://github.com/kotlin/kotlinx.serialization

---

4 votes, 9 watchers

aahlenst:
This pull requests adds `HttpMessageConverter`s for kotlinx.serialization that support `application/json` and `application/cbor`. Tests are included but written in Kotlin because `@Serializable` works only for Kotlin data classes.

I have not added a converter for Protocol Buffers but might do so if this pull request gets accepted.

Closes gh-21188.


aahlenst:
@jhoeller Is there any chance that this might make it into 5.3 (happy to update/polish in that case) or would you prefer that I publish it as a separate artifact?

jhoeller:
@sdeleuze What's your take on this? If we move forward with it, 5.3 would be a fine timeframe from my perspective...

sdeleuze:
This would indeed be a very exciting feature, especially when you consider multiplatform scenarios like Spring Boot backend and KotlinJS frontend, but before giving an educated advice I need to talk with Kotlin team about the runtime API stability they plan to ensure for Kotlin 1.4, see https://github.com/Kotlin/kotlinx.serialization#current-project-status for more details. I will check with them and provide a feedback shortly.

vanniktech:
@aahlenst would you mind publishing this as a separate artifact in the meantime?

sdeleuze:
Ok so it seems we should be able to get stable support for the JSON part in Kotlin 1.4 timeframe. I think that's a good start since that enables multiplatform application with Spring backend, Kotlin/JS frontend and shared module.

@aahlenst Could you please update your PR to retain only the JSON support?

spring-web/src/main/java/org/springframework/http/converter/json/KotlinSerializationJsonHttpMessageConverter.java
```diff
@@ -0,0 +1,122 @@
...
+	 * Construct a new {@code KotlinSerializationJsonHttpMessageConverter} with default configuration.
+	 */
+	public KotlinSerializationJsonHttpMessageConverter() {
+		this(JsonConfiguration.getDefault(), EmptyModule.INSTANCE);
+	}
+
+	/**
+	 * Construct a new {@code KotlinSerializationJsonHttpMessageConverter} with custom configuration.
+	 */
+	public KotlinSerializationJsonHttpMessageConverter(JsonConfiguration jsonConfiguration, SerialModule serialModule) {
```
vanniktech, line 67: Can we also get a constructor that just takes the ready build `Json` class? That way we're much more flexible. I'd even go as far as removing this constructor in favor of just `Json`.

aahlenst: > That way we're much more flexible. I'd even go as far as removing this constructor in favor of just Json.

Could you please elaborate? Removing parameters/constructors is usually the opposite of "more flexible"?

aahlenst:
@sdeleuze Will do. Do you have additional information regarding CBOR? We're using both JSON/CBOR with Multiplatform, primarily for mobile.

sdeleuze:
CBOR won't be stable, it has been confirmed. But we will be happy to add it later. For the time being, I will make sure you can easily add it on your own.

spring-web/src/main/java/org/springframework/http/converter/json/KotlinSerializationJsonHttpMessageConverter.java
vanniktech, line 67:
```suggestion
	public KotlinSerializationJsonHttpMessageConverter(Json json) {
```

That way the consumer can construct the Json Instance in whichever way they want it to be. The other constructor would change like that:

```diff
public KotlinSerializationJsonHttpMessageConverter() {
-  this(JsonConfiguration.getDefault(), EmptyModule.INSTANCE);
+  this(new Json(JsonConfiguration.getDefault(), EmptyModule.INSTANCE)
}

Comment (vanniktech) [reply spring-web/src/main/java/org/springframework/http/converter/json/KotlinSerializationJsonHttpMessageConverter.java:67]: I replied but somehow it wasn't picked up correctly. It's here: https://github.com/spring-projects/spring-framework/pull/24436#discussion_r432472587

File: spring-web/src/main/java/org/springframework/http/converter/KotlinSerializationResolver.java [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+	 * @param type to find a serializer for.
+	 * @return resolved serializer for the given type.
+	 * @throws RuntimeException if no serializer supporting the given type can be found.
+	 */
+	public synchronized KSerializer<Object> resolve(Class<?> type) {
+		KSerializer<Object> cachedSerializer = RESOLVER_CACHE.get(type);
+		if (cachedSerializer != null) {
+			return cachedSerializer;
+		}
+		KSerializer<Object> resolvedSerializer = JvmResolvingKt.serializerByTypeToken(type);
```
qwwdfsad, line 50: Hi.
`serializerByTypeToken` is not yet stable, but we are working towards its stabilization.

It would be nice if you could report as much problems in `serializerByTypeToken` as possible, testing it on a various edge cases. E.g. we implemented it to be used along with `typeOf` or `typeTokenOf`: https://github.com/Kotlin/kotlinx.serialization/blob/master/runtime/jvmTest/src/kotlinx/serialization/features/ResolvingTest.kt#L102
You here are using it in a slightly different manner, invoking `.getClass` directly, so it would be nice if you could test it on a non-trivial types: wildcards, Kotlin nullable types, generic types (`List<String>`, nested maps, parameterized POJOs etc.). I am pretty sure some of these pattern are not going to work as is :)

vanniktech: I quickly tested this today and it looked like `List<T>` & `Map<T>` aren't supported.

aahlenst: Generics generally do not work at the moment. There are (and have been) tests in place that confirm this.

aahlenst:
@sdeleuze Ready for review. I rebased everything and dropped the CBOR support. Adding CBOR again is painless, there are no dependencies on other classes.

spring-web/src/main/java/org/springframework/http/converter/json/KotlinSerializationJsonHttpMessageConverter.java
aahlenst: Done.

aahlenst:
@sdeleuze Updated to Kotlin 1.4 and kotlinx.serialization 1.0.0-RC. Reading and writing of generics is now included, too.

vanniktech (approved):
Very nice 👍

sdeleuze:
Merged and polished via cd6085a31009e830b63461b71ae69e39afeac27c, thanks for this useful contribution @aahlenst.

sdeleuze:
I plan to add kotlinx.support to WebFlux (and its Coroutines support) in SpringFramework 5.3 RC2, see https://github.com/spring-projects/spring-framework/issues/25771 related issue.

vanniktech:
@aahlenst how/when will this make it's way into spring boot?

aahlenst:
@vanniktech I'm the wrong person to ask because I'm not involved with any Spring project or Pivotal. Usually, there's a Boot release shortly after the Framework release. So I assume that will have it.

snicoll:
@vanniktech the milestone  was removed by accident. This was added in Spring Framework `5.3.0-RC1` that got released last week. Each Spring Boot release [has a list of versions in the reference guide](https://docs.spring.io/spring-boot/docs/2.4.0-M3/reference/html/appendix-dependency-versions.html#dependency-versions). Spring Boot `2.4.0-M3` has been released with this work.

PR #24476, opened 2020-02-04
Produce less String-related garbage when generating bean names

stsypanov:


sbrannen:

Thanks

PR #24481, opened 2020-02-05
Upgrade to Spring Asciidoctor Extensions 0.4.0.RELEASE

wilkinsona:
0.4.0 provides built-in support for remembering a user's selections using local storage. This replaces the custom switch language JavaScript.

The selection is stored using a key derived from the options that were available. Concretely, when the options are Java or Kotlin, the local storage key is java-kotlin. Similarly, if the choices were Java, Kotlin, and XML, the key would be java-kotlin-xml.

Given local storage's domain and protocol scoping, the nature of the key that's used for storage will allow a user's selections to be applied across all documentation hosted on https://docs.spring.io that offer the same options.


sbrannen:

Thanks

PR #24484, opened 2020-02-06
Improve assertion message in PersistenceExceptionTranslationInterceptor

michael-simons:
It's a bit odd asserting the bean factory with != null and the given message.


sbrannen:
`Assert.state` throws an `IllegalStateException` which is appropriate here.

`Assert.notNull` throws an `IllegalArgumentException` which is inappropriate here, since the thing _missing_ is not provided as an argument to the method in question.

Spring is rather consistent in that regard, when multiple properties could have been set and Spring later determines that the current _state_ is illegal.

As for the exception message, "No XXX set" is also pretty standard across core Spring for missing properties; however, your proposal would provide more detailed information to the user and would align with the message in `setBeanFactory`.

So if you'd like to revert back to `Assert.state`, we'll consider the PR.

Cheers

michael-simons:
Thanks for letting me know, I understand and the argument makes perfect sense. As I applied the suggested change, you see that I still like the more detailed exception message better.

sbrannen:

Thanks

Issue #23905, opened 2019-11-01
Annotation retrieval performance degradation after 5.2


quaff:
I have a web application based on struts2 and spring, after upgrading to spring 5.2.0, I found performance degraded, I confirm it caused by [AutowireCapableBeanFactory.createBean()](https://github.com/apache/struts/blob/600203bbea77ba3c58f42a95b2b9970e2d928939/core/src/main/java/com/opensymphony/xwork2/spring/SpringObjectFactory.java#L191) .

I write an very simple test project [AutowirePerformance.zip](https://github.com/spring-projects/spring-framework/files/3796470/AutowirePerformance.zip), 1000000 iterations takes about 2100ms with 5.1.9 but 2800ms with 5.2.0, it means throughput decreased 25%.

https://github.com/spring-projects/spring-framework/issues/23792 doesn't improve much, @e-hubert would you run this test project and feedback?

quaff:
5.1.9:
<img width="1282" alt="5 1 9" src="https://user-images.githubusercontent.com/143040/67999210-46862080-fc96-11e9-861e-99fd0899f5b2.png">

5.2.0:
<img width="1295" alt="5 2 0" src="https://user-images.githubusercontent.com/143040/67999211-46862080-fc96-11e9-80c3-983f0d408f96.png">

e-hubert:
@quaff I run your test program locally on my box and can confirm your results. In addition to your tests with 5.1.9 and 5.2.0 I can add that in my tests 5.1.11 is still fast, while already the first milestone M1 shows a degredation. Right from the first glance on the profiler output this time I could not immediately spot an exact cause though.

quaff:
@e-hubert I have the same result, it degrade since 5.2.0.M1 .

quaff:
In a typical struts2 application, `Action` and `Result` are created by spring per request, I tested one scenario in my real application, this degradation could decrease 30% throughput.

quaff:
`ctx.getBean("prototypeBean")` also hit performance degradation, throughput decreased about 10% when upgrade from 5.1.x to 5.2.x, this is my simple test
```
public class GetBeanTest {
	public static void main(String[] args) throws Exception {
		AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext(MyConfiguration.class);
		// warming
		for (int i = 0; i < 1000; i++)
			ctx.getBean("action");
		int iterations = 1000000;
		StopWatch sw = new StopWatch();
[...]

PR #24485, opened 2020-02-06
Prevent unnecessary refresh for InjectionMetadata.EMPTY

yokotaso:
InjectionMetadata.EMPTY won't cached well in AutowiredAnnotationBeanPostProcessor#injectionMetadataCache

org.springframework.beans.factory.annotation.InjectionMetadata#needsRefresh
with InjectionMetadata.EMPTY always returns true,

And therefore AutowiredAnnotationBeanPostProcessor#find always calls
AutowiredAnnotationBeanPostProcessor#buildAutowiringMetadata

As a result of this, Performance degration of constructing component 
happen from 5.1.

Ref: https://github.com/spring-projects/spring-framework/issues/23905

This is a part of performance degration, And this pull request won't improve performance degration 
completely. performance degration remains a little.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
@@ -153,6 +153,9 @@ public class InjectionMetadata {
 	 * @return {@code true} indicating a refresh, {@code false} otherwise
 	 */
 	public static boolean needsRefresh(@Nullable InjectionMetadata metadata, Class<?> clazz) {
+		if (metadata == EMPTY) {
+			return false;
+		}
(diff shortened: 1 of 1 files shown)


jhoeller:
Good catch! I've implemented this slightly differently, introducing a `needsRefresh` method on `InjectionMetadata` itself (which is overridden to always return `false`  in the `EMPTY` instance). Thanks for the pull request, in any case!

yokotaso:
@jhoeller 
Thank you! code might be broken?
https://github.com/spring-projects/spring-framework/commit/669a689a50a1afd8f3cd97639753933822c8c79e#diff-59a9bf0a4cc5269a25aa9aba094bd2e7R2

jhoeller:
Indeed, fixing that comment typo as we speak :-) (and refining some javadoc a bit)

PR #24490, opened 2020-02-07
Extract a method named getRawClassForType to reuse some code

chenqimiao:
Extract a method named getRawClassForType to reuse some code and add a test case for the method.


chenqimiao:
Duplicate of #24525

PR #24494, opened 2020-02-09
Use equals method() instead of ==

hyeonisism:
Equals are better than comparisons using "=="
Also, Boolean returns each new object.


rstoyanchev:
I've changed this to `Boolean.TRUE.equals(..)` consistent with the rest of the framework, and shorter. `Boolean.TRUE` returns a constant.

PR #24496, opened 2020-02-10
ExecutorConfigurationSupport to allow millisecond precision for await termination period

ttddyy:
I noticed `ExecutorConfigurationSupport`(_`ThreadPoolTask[Executor|Scheduler]`_) only takes seconds for await termination.

I'm writing a graceful shutdown logic for k8s environment and awaiting by second is a bit large granularity to control the shutdown/await. Graceful shutdown get triggered by liveness probe and the frequency for liveness probe is not so long. 

This PR changes the minimum unit to milliseconds and adds a method to take `Duration` to specify the await termination.

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/concurrent/ExecutorConfigurationSupport.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/concurrent/ExecutorConfigurationSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.scheduling.concurrent;
 
+import java.time.Duration;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionHandler;
@@ -59,7 +60,7 @@ public abstract class ExecutorConfigurationSupport extends CustomizableThreadFac
 
 	private boolean waitForTasksToCompleteOnShutdown = false;
 
-	private int awaitTerminationSeconds = 0;
+	private Duration awaitTerminationDuration = Duration.ZERO;
 
 	@Nullable
 	private String beanName;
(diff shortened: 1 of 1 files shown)


spring-context/src/main/java/org/springframework/scheduling/concurrent/ExecutorConfigurationSupport.java
```diff
@@ -145,7 +146,16 @@ public void setWaitForTasksToCompleteOnShutdown(boolean waitForJobsToCompleteOnS
...
+		this.awaitTerminationDuration = Duration.ofSeconds(awaitTerminationSeconds);
+	}
+
+	/**
+	 *
+	 * @param awaitTerminationDuration time to await to shutdown internal executor
+	 * @since 5.2.4
+	 */
+	public void setAwaitTerminationDuration(Duration awaitTerminationDuration) {
+		this.awaitTerminationDuration = awaitTerminationDuration;
```
rstoyanchev: I think should be called just `setAwaitTermination`.

rstoyanchev: I think should be called just `setAwaitTermination`.

spring-context/src/main/java/org/springframework/scheduling/concurrent/ExecutorConfigurationSupport.java
```diff
@@ -145,7 +146,16 @@ public void setWaitForTasksToCompleteOnShutdown(boolean waitForJobsToCompleteOnS
 	 * @see java.util.concurrent.ExecutorService#awaitTermination
 	 */
 	public void setAwaitTerminationSeconds(int awaitTerminationSeconds) {
-		this.awaitTerminationSeconds = awaitTerminationSeconds;
+		this.awaitTerminationDuration = Duration.ofSeconds(awaitTerminationSeconds);
+	}
+
+	/**
+	 *
```
rstoyanchev: Add Javadoc (won't pass checkstyle).

rstoyanchev (changes requested):

rstoyanchev:
I've turned that into `awaitTerminationMillis` which is usable for XML config, consistent with the rest of the framework, itself a shortcut for a Duration.

ttddyy:
cool, thanks!!

PR #24499, opened 2020-02-10
AbstractJackson2HttpMessageConverter handles JsonMappingException

Romster:
JsonMappingException is treated as HttpMessageConversionException to keep the same behavior for new exceptions introduced in jackson-databind 2.10
Fixes issue 24455


jhoeller:
I've handled this a bit differently, specifically treating `MismatchedInputException` before we go into any other `JsonMappingException`, also applying it to our reactive Jackson encoder/decoder. Thanks for the PR, in any case!

PR #24524, opened 2020-02-14
MapSqlParameterSource.getValue Handling of null values

rengy-github:
Does the reading of multithreaded HashMap need to be synchronized? This problem has been bothering me. I don't think it's necessary

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterJdbcTemplate.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterJdbcTemplate.java
@@ -429,17 +429,9 @@ public class NamedParameterJdbcTemplate implements NamedParameterJdbcOperations 
 		if (getCacheLimit() <= 0) {
 			return NamedParameterUtils.parseSqlStatement(sql);
 		}
-		ParsedSql parsedSql = this.parsedSqlCache.get(sql);
-		if (parsedSql == null) {
-			synchronized (this.parsedSqlCache) {
-				parsedSql = this.parsedSqlCache.get(sql);
-				if(parsedSql==null) {
-					parsedSql = NamedParameterUtils.parseSqlStatement(sql);
-					this.parsedSqlCache.put(sql, parsedSql);
-				}
-			}
+		synchronized (this.parsedSqlCache) {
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@rengy-github Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24524)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24524) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@rengy-github Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24524)!

rengy-github:
Using synchronized only with put improves performance

spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterJdbcTemplate.java
```diff
@@ -429,9 +429,17 @@ protected ParsedSql getParsedSql(String sql) {
 		if (getCacheLimit() <= 0) {
 			return NamedParameterUtils.parseSqlStatement(sql);
 		}
-		synchronized (this.parsedSqlCache) {
-			return this.parsedSqlCache.computeIfAbsent(sql, NamedParameterUtils::parseSqlStatement);
+		ParsedSql parsedSql = this.parsedSqlCache.get(sql);
+		if (parsedSql == null) {
+			synchronized (this.parsedSqlCache) {
+				parsedSql = this.parsedSqlCache.get(sql);
```
hyeonisism: I think you can change the code more clearly with the computeIfAbsent() method.

I'd recommend you refer to [this](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-).

rengy-github:
I know it's not thread safe to use synchronization only when putting，But occasionally SQL is parsed twice without any problem

rengy-github:
i think parse a sql multiple times   better than blocking   other thread while get sql

rengy-github:
My puzzle is whether there is problem with the other thread get when one thread puts

Could send me a link about this, thank you

snicoll:
There has been a number of changes that leave the current proposal of returning null if the key is not set, which is a breaking change.

PR #24525, opened 2020-02-14
Optimize getDependencyType implementation of DependencyDescriptor by using ResolvableType

chenqimiao:
This pr optimizes `getDependencyType` implementation of  `DependencyDescriptor`  by using `ResolvableType` instead of native type operations.
Please help review the code by `Files changed`. Looking forward to your reply.


chenqimiao:
I think using `ResolvableType` to achieve the acquisition of nested raw class to optimize existing duplicate and long code.
Can someone help confirm the review?

snicoll:
Good catch @chenqimiao

PR #24529, opened 2020-02-15
Fix typo and polish if block

chenqimiao:
This pr is to polishing some `if` blocks in `ResolvableType` and fixing some typos in `ResolvableTypeTests`.
Looking forward to your reply.


rstoyanchev:
I think the code should remain as is returning at that point only if the condition evaluates to true, but otherwise continuing and eventually exiting through the false return at the end.

chenqimiao:
@rstoyanchev I get it, thank you for your feedback.

sbrannen:
@chenqimiao, I incorporated your typo fixes in 7778508e69f5bb09e58779fd4cdb347bdebe13bb.

PR #24530, opened 2020-02-15
Partial dependency injection into constructor

gzoom13:
With this commit it is possible to partially autowire constructors.
It is useful, for example, for lookup methods, creating prototype beans.
Sometimes prototype beans depend on singletons, and before to create
such prototypes it was required to write custom factories. This change
allows to use lookup methods.


goatfryed:
Any news on this topic?

gzoom13:
Hi @goatfryed. Since no one was reviewing this, I ended up copy-pasting original bean factory and modifying it in separate library: https://github.com/gzoom13/partial-autowire-bean-factory/. I have not published it, but if it will be useful for you, I could do it. It also requires some decomposition. What Spring version do you have?

snicoll:
Sorry for the late review. I can see you've moved this somewhere else and I don't think we want to support the additional complexity so I am going to close this now.

PR #24531, opened 2020-02-16
Correct formatting of MessageSource example in documentation

hrybs:
The formatting of MessageSource example is not correct. To have as an output the sentence "Ebagum lad, the 'userDao' argument is required, I say, required." we should add two pairs of single quotes.


sbrannen:
Good catch.

Thanks

PR #24535, opened 2020-02-17
PartsHttpMessageWriter introduced

kptfh:
Currently there is no option to pass `Flux<Part>` as body via `WebClient`. This pr extends `WebClient` functionality and allows accept  `Flux<Part>` as body.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/multipart/MultipartHttpMessageWriter.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/multipart/MultipartHttpMessageWriter.java
@@ -330,7 +330,7 @@ public class MultipartHttpMessageWriter extends LoggingCodecSupport
 	}
 
 
-	private Mono<DataBuffer> generateBoundaryLine(byte[] boundary, DataBufferFactory bufferFactory) {
+	static Mono<DataBuffer> generateBoundaryLine(byte[] boundary, DataBufferFactory bufferFactory) {
 		return Mono.fromCallable(() -> {
 			DataBuffer buffer = bufferFactory.allocateBuffer(boundary.length + 4);
 			buffer.write((byte)'-');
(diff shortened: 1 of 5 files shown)


rstoyanchev (changes requested):
@kptfh thanks for the pull request. I think a `Flux<Part>` and `Mono<MultiValueMap<String, ?>>` multipart writers are almost identical, one encoding a Flux sequence of parts, the other a sequence of parts extracted from a Map. So the code should be almost identical. So I would expect this to be solved through a base class from which both would extend. What do you think?

kptfh:
Hi. Do you mean to include this code into `MultipartHttpMessageWriter`?  
Initially I started with this approach but code get over-complicated so I decided to introduce small atomically tested change that meets single responsibility criteria.
IMHO I think that `MultipartHttpMessageWriter` should be split as it responsible for Form data encoding and multipart data and has a lot of ifs.

rstoyanchev:
No this can't be done in the current `MultipartHttpMessageWriter` because it's parameterized by `MultiValueMap<String, ?>`. 

However, and I haven't tried it, I think a base class can be extracted out from the current `MultiparthttpMesageWriter` with most of the logic for writing parts. So this new `PartHttpMessageWriter` (or perhaps `StreamingMultipartHttpMessageWriter`) could then extend from the same base class and both should then be pretty minimal, with most of the logic in the base class.

rstoyanchev:
> IMHO I think that MultipartHttpMessageWriter should be split as it responsible for Form data encoding and multipart data and has a lot of ifs.

`MultipartHttpMessageWriter` merely delegates to `FormHttpMessageWriter` so it acts as a composite, that's about as simple as it gets. There is a reason for that, because in `write` we can tell if the data is multipart or not even if the `MediaType` is not present. We can't do the same via `canWrite`.

PR #24541, opened 2020-02-17
Fixing logic error.

hebo6:
My application server provides HTTP(http://172.18.0.120) service on port 80, which runs behind the reverse proxy server providing HTTPS(https://domain.com) service on port 443. Then the websocket handshake to https://domain.com/websocket failed because the method org. Springframework. Web. CORS. Corsutils. Iscorsrequest() returns true(originUrl.getScheme() returns "https", request.getScheme() returns "http") when using HTTPS reverse proxy server to proxy HTTP service.


quaff:
There is no logic error here, you should consider [ForwardedHeaderFilter](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/filter/ForwardedHeaderFilter.java) .

PR #24551, opened 2020-02-19
Fix javadoc in DependencyDescriptor

chenqimiao:
The java doc description of `(never {@code null})` does not match the annotation of `@Nullable`.
Suggest to fix this description.


sbrannen:
Good catch.

Thanks

PR #24555, opened 2020-02-19
Simplify code in spring-test by using Collections.addAll

hyeonisism:
This can be changed like this.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
```diff
@@ -291,9 +292,7 @@ private void cookies(MockHttpServletRequest request) {
 
 		Cookie[] parentCookies = request.getCookies();
 		if (parentCookies != null) {
-			for (Cookie cookie : parentCookies) {
-				cookies.add(cookie);
-			}
+			cookies.addAll(Arrays.asList(parentCookies));
```
sbrannen: Please avoid the creation of the extra `ArrayList` as follows.

```suggestion
			Collections.addAll(cookies, parentCookies);
```

sbrannen (changes requested):

hyeonisism:
@sbrannen I fixed it. Thanks

sbrannen:

Thanks

Issue #24291, opened 2020-01-03
Inherited @Transactional methods should be able to use class-level TransactionManager qualifier from concrete class


elab:
Setup: In a multi-DB environment, there is an abstract Service/DAO class with `@Transactional` methods common to all entities from all DBs. Then there are multiple concrete Service/DAO beans, one per DB, each with specific transactionManager in the class-level `@Transactional` annotation.

Problem: When calling a base-class method via some DB-specific bean, a wrong transactionManager is used (always the `@Primary` one). The worst is that no exceptions are thrown, no error messages printed in log. The entities are simply silently e.g. not saved/updated in DB. Or, I afraid, they could even be saved to wrong DB.

The problem has already been described at least in #14011, #14295, #17080. Related issue is also #11839. But all the issues have been closed (as outdated) without a solution.

There were also some workarounds: https://stackoverflow.com/questions/51087660/dynamic-selection-of-transactionmanager-spring-boot, https://www.tirasa.net/en/blog/dynamic-springs-at-transactional, also #14295. But they all are not working (anymore).

How can it be solved? What do you, guys, do in such situation? Am I missing something obvious?
Thank you for your thoughts.

elab:
I couldn't find a reliable way to work around (e.g. defining an own TransactionInterceptor was practically impossible for me due to many unclear interdependencies in Spring code; would be glad to see a working example though). Therefore I don't see any other way as to make changes to the Spring codebase, to the place where transaction attributes are computed.

It's all about the class `AbstractFallbackTransactionAttributeSource`, method `computeTransactionAttribute`. There are two implementation of them:
- [in spring-framework/spring-tx](https://github.com/spring-projects/spring-framework/blob/5.1.x/spring-tx/src/main/java/org/springframework/transaction/interceptor/AbstractFallbackTransactionAttributeSource.java#L143-L187) (@jhoeller), important pieces of javadoc: [on class](https://github.com/elab/spring-framework/blob/5.1.x/spring-tx/src/main/java/org/springframework/transaction/interceptor/AbstractFallbackTransactionAttributeSource.java#L32-L52)
[...]

Issue #23473, opened 2019-08-16
Allow global default for @Transactional rollback behavior on checked exceptions


ilyastam:
In Kotlin all exceptions are effectively unchecked. Therefore, transactions will not be rolled back for functions like this:

```
@Transactional
fun example(): Unit {
  runQuery()
  throw MyCheckedException()
  runQuery()
}
```

to achieve a correct behavior, the above code needs to be refactored as follows:

```
@Transactional(rollbackFor = [Exception::class])
fun example(): Unit {
  runQuery()
  throw MyCheckedException()
  runQuery()
}
```

this isn't very intuitive and can lead to unexpected results. Furthermore, even if a developer is aware of this, he/she should not forget to specify `rollbackFor` for every `@Transactional` annotation.

**It should be possible to configure application-wide defaults for `rollbackFor` and `noRollbackFor`**

the-fine:
@ilyastam as stated here: https://docs.spring.io/spring/docs/5.2.0.RC1/spring-framework-reference/data-access.html#transaction-declarative-rolling-back
Transactions are rolled back by unchecked exceptions by default.
So it should work out of the box with Kotlin because exceptions are unchecked https://kotlinlang.org/docs/reference/exceptions.html
Do you have a sample projects that reproduces this issue?

kilink:
Kotlin doesn't distinguish between RuntimeException (unchecked) and Exception (checked), meaning Kotlin code can freely throw either type. Spring transactional support does distinguish between the two by not rolling back any checked exception by default (an exception subclassing Exception).

One can choose to only use RuntimeException derived exceptions in Kotlin to avoid that issue, but it can also arise when Kotlin code calls into a Java method that throws a checked Exception. Essentially it's a "foot gun", one mistake can lead to the unexpected behavior of a transaction not being rolled back.

raderio:
Why not to rollback by default by `Exception` and not by `RuntimeException` without any configs?
This will not break the Java code.

elab:
How about specifying rollback behavior in a `@Transactional` annotation at the class level?

snicoll:
@elab that would mark all public methods in the class to be transactional.

Fryie:
It would be really beneficial to have this as a configurable option (maybe even set by default when choosing Kotlin on `start.spring.io`). Since Kotlin doesn't force you to handle checked exceptions, it's easy to accidentally end up with a corrupted state because the transaction is not rolled back.

kilink:
[...]

PR #24571, opened 2020-02-22
Issue #24291 - Merge transactional attributes

elab:
Changes to spring-tx as discribed in #24291

### DIFF
--- a/spring-tx/src/main/java/org/springframework/transaction/annotation/AnnotationTransactionAttributeSource.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/annotation/AnnotationTransactionAttributeSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


elab:
@odrotbohm Since transaction attributes are merged in this new implementation, it could be used in spring-data as is (i.e. without [reverting](https://github.com/spring-projects/spring-data-commons/blob/2.2.x/src/main/java/org/springframework/data/repository/core/support/TransactionalRepositoryProxyPostProcessor.java#L102-L110) the [order of lookups](https://github.com/spring-projects/spring-data-commons/blob/2.2.x/src/main/java/org/springframework/data/repository/core/support/TransactionalRepositoryProxyPostProcessor.java#L377-L451)), what do you think?

elab:
Looking though [Reference Documentation (Data Access part)](https://docs.spring.io/spring/docs/current/spring-framework-reference/data-access.html) 5.2.3 I couldn't find anything that had to be changed.

jhoeller:
It looks like we'll address both #24291 and #23473 through a new `@TransactionConfig` annotation, analogous to our existing `@CacheConfig`, or similar global settings which are not trying to rely on class-level `@Transactional` declarations (which brings in the undesirable implicit transactional-for-all-public-methods semantic whereas we only try to have certain settings shared for methods which are explicitly marked as transactional). From that perspective, we are not going to use a merging approach here.

Issue #23915, opened 2019-11-03
Turn off useSuffixPatternMatching by default


rupert-madden-abbott:
When useSuffixPatternMatching is set to true, then 

> a method mapped to "/users" also matches to "/users.*"

The default value is true. Please can this be changed to false.

**Defaults shouldn't be surprising**
It is not obvious to developers that this will occur. They intend to setup a static mapping of "/users" and find that this mapping works. They have to explicit visit "/users.blah" to discover that they have unintentionally mapped more than they intended.

There are lots of examples of this causing surprise with developers being confused by the cause and purpose of this function:

* https://stackoverflow.com/questions/9688065/spring-mvc-application-filtering-html-in-url-is-this-a-security-issue
* https://stackoverflow.com/questions/30610607/how-to-change-spring-request-mapping-to-disallow-url-pattern-with-suffix
* https://stackoverflow.com/questions/30307678/why-does-requestmapping-spring-annotation-in-controller-capture-more-that-i-wan
* https://stackoverflow.com/questions/22845672/requestmapping-in-spring-with-weird-patterns

The behaviour is [clearly documented](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-requestmapping-suffix-pattern-match) but that isn't good enough to stop the surprise because it isn't clear from the application code itself. All a developer sees is that "/users" is mapped and there is nothing explicit to make them realise something else is going on.

**Defaults should encourage best practice**
As documented, this default is enabling a practice that used to be widespread but has fallen out of favour due to the problems it causes. Developers should be using the Accept header and not file extensions for content negotiation.

It would be better to allow developers to opt-in to this feature, rather than having the best practice be an opt-in.

**Defaults should be secure**
`antMatchers` still exists in Spring Security. Whilst everybody should be using `mvcMatchers`, it's still really easy to open up a security hole by having this switched on.

It is also much more risky than the optional trailing slash feature which developers are far more likely to expect and anticipate since such a feature is common to many web frameworks. Developers are very unlikely to add a file extension to their URL if they haven't mapped one and thus can still easily miss this vulnerability if they have failed to use `mvcMatchers`

rstoyanchev:
These are all valid arguments from our own reference docs, and in a greenfield situation like WebFlux, those options either don't even exist. In Spring MVC however, the surprise will also apply to all those who already rely on the current behavior. That is the main sticking point.

rstoyanchev:
**Team decision:** 

[...]

Issue #24179, opened 2019-12-10
Deprecate use of path matching and content negotiation via path extension


rstoyanchev:
The motivation for this has been described in the ["Suffix Match"](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-requestmapping-suffix-pattern-match) section of the reference docs for some time. This issue is about formally deprecating options for the use of path extensions for path matching and content negotiation.

When you have to consider all of the below together -- path patterns and URI variables, URL encoding (and double encoding tricks), path parameters ";"  and their treatment (or mistreatment) by frameworks and servers, relative paths `"../"`, the need for alignment between security and web framework path matching -- it is not surprising that incidental complexity makes it difficult to reason about potential attacks such as described in ["Suffix Match and RFD"](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-requestmapping-rfd) section of the reference docs. This is what makes path matching and content negotiation via path extensions an anti-pattern.

The preferred options should be to use the `Accept` header for content negotiation and `@RequestMapping(produces="...")` for path matching. These are and have been the defaults in Spring Boot auto-configuration. If a replacement for content negotiation via path extensions is needed, then using a query parameter should be preferred.

In 5.3, suffix pattern matching will be turned off by default, see https://github.com/spring-projects/spring-framework/issues/23915#issuecomment-563987147 for details.

kev22257:
Currently we are setting PathMatchConfigurer.setUseSuffixPatternMatch to false.  How do we advance beyond Spring 5.2.3 without a deprecated warning and keep the functionality the same?  Do we have to wait for Spring 5.3 when this will presumably be the default?

kev22257:
> Currently we are setting PathMatchConfigurer.setUseSuffixPatternMatch to false. How do we advance beyond Spring 5.2.3 without a deprecated warning and keep the functionality the same? Do we have to wait for Spring 5.3 when this will presumably be the default?
> 
> If that is true would it be possible to bridge the gap with a new method like disableUseSuffixPatternMatch() that explicitly sets the value to false without causing the warning considering GA for Spring 5.3 I think I saw was October 2020?

Never mind, I found the conversation you had in #24642.

PR #24576, opened 2020-02-23
Turn off useSuffixPatternMatching by default

mikesmithson:
Changes for issue #23915.


rstoyanchev (changes requested):
>  is this issue something a first-time contributor can submit a PR for? I saw that it was self-assigned back in December and wasn't sure if it was being actively worked on. I have a PR that I can submit if this issue is still available for outside contribution.

The necessary changes are quite small . The tricky part is reviewing the Javadoc in all the relevant classes and making updates, especially with regards to defaults. If you see the classes touched by commits associated with #24179, that will give you an idea. 

The reference documentation will also need to be checked and updated as needed.

I see that `RequestMappingInfo` has a similar property that needs changing too.

There might be more but this should give enough for another draft.

mikesmithson:
> > is this issue something a first-time contributor can submit a PR for? I saw that it was self-assigned back in December and wasn't sure if it was being actively worked on. I have a PR that I can submit if this issue is still available for outside contribution.
> 
> The necessary changes are quite small . The tricky part is reviewing the Javadoc in all the relevant classes and making updates, especially with regards to defaults. If you see the classes touched by commits associated with #24179, that will give you an idea.
> 
> The reference documentation will also need to be checked and updated as needed.
> 
> I see that `RequestMappingInfo` has a similar property that needs changing too.
> 
> There might be more but this should give enough for another draft.

@rstoyanchev  - OK, am I at least on the right track or do I need to regroup and start from the beginning?

rstoyanchev:
Yes that flag does need to change. My intention was to list what else would be needed.

mikesmithson:
> Yes that flag does need to change. My intention was to list what else would be needed.

@rstoyanchev  - OK, I pushed another change. Crossing my fingers...

mikesmithson:
is there any interest in the PR still?

kuberr:
> @rstoyanchev - Is there any interest in this PR still?

Yes.
We are facing a deprecation warning which I'm not sure if it will be solved by this PR. We currently  use this:

`contentNegotiationConfigurer.favorPathExtension(false);`

Bu it gives a deprecation warning. I was hoping this PR would allow us to remove this line of code. 

However, when I look at the actual changes I see that the `useSuffixMatch` variable is still set to true (alongside the `useRegisteredSuffixMatch` one).

rstoyanchev:
> Is there any interest in this PR still?

@mikesmithson yes but master is not tracking 5.3.x yet. The PR is incomplete as it stands. See comment by @kuberr above and also the Javadoc and documentation will need a more thorough review and updates.

I'm sorry that I can't give more detailed guidance and there is a reason this is not marked as ideal for contribution. It is tricky with lots of small changes that require a lot of attention. You can see how much work it took to apply deprecations in #24179 and you can also use that for clues if you'd like to take this further.

mikesmithson:
@kuberr @rstoyanchev  - I would gladly like to take this issue to the finish line if I can get incremental feedback along the way as I am mostly flying blind on a fair amount of this PR. I would probably have to make several small commits along the way and then make adjustments if necessary based on your feedback. Is that a reasonable proposal to you?

rstoyanchev:
@mikesmithson by all means, any further improvements would be welcome.

rstoyanchev:
@mikesmithson, I've resolved this with 147b8fb75550ebe25ebdae6a42af74c61b24439f. As I mentioned this really wasn't ideal for contribution. Thanks for the pull request in any case!

PR #24585, opened 2020-02-24
Improve documentation for @Autowired/@Value in @Configuration classes

chenqimiao:
Original：
`Otherwise, @Autowired and @Value do not work on the configuration class itself, since it is being created as a bean instance too early.`

I think this statement is not accurate enough. `@Autowired` actually works, if the priority of the definition of `BeanPostProcessor ` through `@Bean` lower than `AutowiredAnnotationBeanPostProcessor`.
So, I think we can adjust this description.


sbrannen:

Thanks

PR #24587, opened 2020-02-25
Polish some String equals()

hyeonisism:
This change can prevent NPE 
and will be consistent overall code using the String equals method.


quaff:
`method.getName()` always return non-null.

sbrannen:
Thanks for the PR.

Please note, however, that the `Method` and method name will never be `null` in this context. Thus, we will leave the code as it currently is.

PR #24588, opened 2020-02-25
Polish code by using matches() method in HttpMethod

hyeonisism:
I think it's safer to compare with using the matches() method than to compare with using the literal. So I changed it like this.


rstoyanchev:
Thanks for the PR but none of the changed places are exposed to `NullPointerException` and the current checks are fine, even a little shorter, so we'll keep them as they are.

PR #24598, opened 2020-02-26
Test HttpMethod class

hyeonisism:
I wrote the test code knowing that there is no test code in this class.


rstoyanchev:
I'm sorry but I don't see the point in this. Verifying the value of an enum has nothing to do with the actual enum, it's how all enums work. The resolve method is a map lookup, and matches is a trivial delegate to resolve. Testing all this with parameterized tests is quite unnecessary too. I'd say the whole test class is not necessary because the code paths for these trivial methods are exercised all over.

PR #24599, opened 2020-02-26
Add support for explicit generic type in PayloadApplicationEvent

chenqimiao:
This PR is to make event mechanisms better support generics.

When I publish an event which contains generics via `org.springframework.context.support.AbstractApplicationContext#publishEvent(java.lang.Object, org.springframework.core.ResolvableType)` , I find that corresponding listener does not get event message.  Please see the unit test cases in my PR. `The unit test method of named publishEventWithGeneric fail before this PR submission`.
So, I think we can add a payloadType field to PayloadApplicationEvent to fix this.
In addition, this fix code also can cache the type information of the payload to improve performance.
 
Thank you for reading and look forward to your reply.


spring-context/src/main/java/org/springframework/context/PayloadApplicationEvent.java
```diff
@@ -34,22 +36,38 @@
...
 		this.payload = payload;
+		this.payloadType = payloadType;
 	}
 
 
 	@Override
 	public ResolvableType getResolvableType() {
-		return ResolvableType.forClassWithGenerics(getClass(), ResolvableType.forInstance(getPayload()));
+		if(this.payloadType == null){
+			this.payloadType = ResolvableType.forInstance(getPayload());
```
jhoeller, line 68: Isn't this supposed to build the full nested `PayloadApplicationEvent` type still:
`this.payload = ResolvableType.forClassWithGenerics(getClass(), ResolvableType.forInstance(getPayload()));`

spring-context/src/main/java/org/springframework/context/PayloadApplicationEvent.java
```diff
@@ -34,22 +36,38 @@
...
 		Assert.notNull(payload, "Payload must not be null");
 		this.payload = payload;
+		this.payloadType = payloadType;
 	}
 
 
 	@Override
 	public ResolvableType getResolvableType() {
-		return ResolvableType.forClassWithGenerics(getClass(), ResolvableType.forInstance(getPayload()));
+		if(this.payloadType == null){
```
jhoeller, line 67: Our code style requires `if (this.payloadType == null) {` here.

chenqimiao: @jhoeller I have polished up this style.

snicoll:
@chenqimiao thank you!

chenqimiao:
@snicoll My pleasure

chenqimiao:
org.springframework.context.event.ApplicationListenerMethodAdapter#onApplicationEvent
-->
org.springframework.context.event.ApplicationListenerMethodAdapter#processEvent
-->
org.springframework.context.event.ApplicationListenerMethodAdapter#resolveArguments
-->
org.springframework.context.event.ApplicationListenerMethodAdapter#getResolvableType
-->
ResolvableType eventType = payloadEvent.getResolvableType();  // Generics may be lost, see PR test for specific cases .

PR #24603, opened 2020-02-26
Fix typo in URI Encoding section of reference docs

floftar:
EncodingMode enum has constant URI_COMPONENT,
but documentation uses URI_COMPONENTS.


fbergr:
Relevant source files:
- https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/util/DefaultUriBuilderFactory.java
- https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java

PR #24617, opened 2020-02-29
Avoid unnecessary sorting overhead

dreis2211:
Hi,

this PR avoids some unnecessary sorting overhead (e.g. if the collection is too small) for methods that are repeatably called and where collection sizes of <= 1 are fairly common (e.g. for the `ProducesRequestCondition`).

Cheers,
Christoph


rstoyanchev:
Thanks for the changes. It's a shame this check isn't in `Collections#sort` or `List#sort`.

dreis2211:
Indeed

PR #24618, opened 2020-02-29
Add missing field to MethodParameter copy constructor

dreis2211:
Hi,

I just noticed that the `MethodParameter` copy constructor misses to copy its `nestedMethodParameter` field.

Cheers,
Christoph

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -185,6 +185,7 @@ public class MethodParameter {
 		this.parameterAnnotations = original.parameterAnnotations;
 		this.parameterNameDiscoverer = original.parameterNameDiscoverer;
 		this.parameterName = original.parameterName;
+		this.nestedMethodParameter = original.nestedMethodParameter;
 	}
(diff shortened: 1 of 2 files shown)


jhoeller:
Well spotted... However, arguably we should not copy such a nested `MethodParameter` reference but rather either also clone it or simply let it be recreated on demand. Since this is effectively just a cached field for repeated nested generics access against the same parameter, we could simply leave it as-is and recreate the nested parameter on demand whenever the next `nested()` call comes in.

dreis2211:
I don't mind if you close this one, if you feel this is not wanted. In my opinion, it should at least be documented then, because one expects from a copy constructor to actually copy things.

dreis2211:
I'm cleaning my open PRs. Any wishes from you @jhoeller ? Closing or not?

snicoll:
Thanks for the PR, Christoph, and sorry for the delay. Juergen and I discussed this one and we're indeed going to close it. Given that the `nestedMethodParameter` is recreated on demand, this shouldn't have an impact and the cloning would have some unnecessary cost. We believe the status quo is the best we can do here.

PR #24619, opened 2020-03-01
Spring Tools 4 is now!

quxinyong:
The website [Spring Tool Suite](https://spring.io/tools/sts)  is not found.


snicoll:
Thank you for the PR but we already have one for that, see #24589.

PR #24624, opened 2020-03-02
Use native trim() of String instead of iterating through String in trimWhitespace of StringUtils

chenqimiao:
I think we can use the native method trim() of String instead of iterating through String in trimWhitespace of StringUtils.
This is simpler and secondarily more efficient.


snicoll:
Thanks for the PR but this was already done as part of #27769. Not that the replacement is `strip()`, not `trim()`.

PR #24631, opened 2020-03-02
Fix formatting in AOP examples in documentation

hrybs:
There is some problems with formatting. As a result we see incorrect Java code snippets. We can try to escape some of the symbols. i.e. using '\\' character. But I don't see any possibility to escape unconstrained quotes. i.e. '**'. So the best way is removing redundant 'quotes' subs.  
**The current documentation is incorrect because of that and contains misleading code fragments.**


rstoyanchev:
Indeed applying text formatting via [quotes substitution](https://asciidoctor.org/docs/user-manual/#quotes) is more likely to cause side effects than to be intentional.

PR #24632, opened 2020-03-03
Remove UTF-8 charset parameter from Content-Type in SseEmitter

sada-sigsci:
SseEmitter is adding header "Content-type:text/event-stream;charset=UTF-8".
Golang revproxy is expecting header "Content-Type: text/event-stream" for
immediately flushing the event stream.

HTML standard says [ https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model ]
ignore the MIME type param and the data is always expected in utf-8.

https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L374

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java
@@ -41,9 +41,6 @@ public class SseEmitter extends ResponseBodyEmitter {
 
 	static final MediaType TEXT_PLAIN = new MediaType("text", "plain", StandardCharsets.UTF_8);
 
-	static final MediaType TEXT_EVENTSTREAM = new MediaType("text", "event-stream", StandardCharsets.UTF_8);
-
-
 	/**
 	 * Create a new SseEmitter instance.
 	 */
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@sada-sigsci Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24632)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24632) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sada-sigsci Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24632)!

rstoyanchev:
The origin of this seems to be #18978 which was added as a workaround for https://github.com/spring-projects/spring-boot/issues/6230.

wilkinsona:
Thanks, @rstoyanchev. My memory of the Boot issues is hazy at best but I think the problem was reported when we were enforcing the encoding on both requests and responses. That was split out so that it was only done for requests by default. If that's correct, you'd have to explicitly set `spring.http.encoding.force-response=true` to encounter the problem. As it was Jetty-specific, there may also be a bit more to it. I think we'd have to try and recreate the original problem to learn more.

rstoyanchev:
Okay, I think we need to process this change to start. If that still causes the original side effect on Jetty with `spring.http.encoding.force-response=true`, we could call `response.setCharacterEncoding(null)` in the handling for SSE effectively resetting whatever forced charset might be.

sada-sigsci:
Thank you so much.

PR #24633, opened 2020-03-03
Add a unit test for getMostSpecificMethod of ClassUtils

chenqimiao:
The method `getMostSpecificMethod` of `ClassUtils` does not seem to be covered by unit test.
Consider adding a unit test method for this method.
Test this method from multiple latitudes of inheritance, method modifiers, implementation, etc.
Looking forward to your reply.


snicoll:
Thanks again @chenqimiao

Issue #24634, opened 2020-03-03
Modernize DataAccessException translation API


ttddyy:
I am implementing transaction retry mechanism to our applications and using `DataAccessException` mapping facility to translate some db failures to transient exception.
Since `DataAccessException` mapping is one of the early days feature of spring-framework, I felt some of the APIs are bit outdated.

Here is list of things I encountered:

### `SQLErrorCodeSQLExceptionTranslator` requires subclassing to perform custom mapping logic.

To add custom mappings to existing logic, `SQLErrorCodeSQLExceptionTranslator#customTranslate` need to be overridden by subclass.

I made a delegation subclass, then injected custom translators.

```java
public class CustomDelegatingSQLErrorCodeSQLExceptionTranslator extends SQLErrorCodeSQLExceptionTranslator {

	@Nullable
	private SQLExceptionTranslator delegate;

	// constructors

	@Override
	protected DataAccessException customTranslate(String task, String sql, SQLException sqlEx) {
		if (this.delegate == null) {
			return null;
		}
		return this.delegate.translate(task, sql, sqlEx);
	}
}
```

I think composition style API is nicer to have than sub-classing.

### Does not have a way to append/update *some* of the error codes on `SQLErrorCodes` (not just entire override)

Error code provided by `org/springframework/jdbc/support/sql-error-codes.xml`(`SQL_ERROR_CODE_DEFAULT_PATH`) can be overridden by `sql-error-codes.xml`(`SQL_ERROR_CODE_OVERRIDE_PATH`).
However, the override is done by bean(DB type). So, it cannot partially update error codes on a specific bean. For example, adding some error codes to transient exceptions to PostgreSQL.

To workaround repeating entire DB error code mapping, I created a `SQLErrorCodesUpdater` which uses the nature that the `SQLErrorCodes` is a mutable java bean.

```java
@AllArgsConstructor
@Getter
public class CustomSQLErrorCodes {

	/**
	 * name has to match with db vendor name from
	 * "DataSourceMetaData#getDatabaseProductName" which also matches to the bean id in
	 * "org/springframework/jdbc/support/sql-error-codes.xml"
	 */
	private String databaseName;

	/**
	 * Which exception to map
	 */
	private DataAccessExceptionType dataAccessExceptionType;

	/**
	 * Whether replace or add to the default sql error code
	 */
	private boolean add;

	/**
	 * SQL error code to add or replace
	 */
	private String[] codes;

}
```

```java
[...]

PR #24644, opened 2020-03-05
Sort detected PersistenceExceptionTranslator beans

ttddyy:
Apply sorting on detected `PersistenceExceptionTranslators`.

Relates to #24634

### DIFF
--- a/spring-tx/src/main/java/org/springframework/dao/support/PersistenceExceptionTranslationInterceptor.java
+++ b/spring-tx/src/main/java/org/springframework/dao/support/PersistenceExceptionTranslationInterceptor.java
@@ -16,6 +16,9 @@
 
 package org.springframework.dao.support;
 
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
 import java.util.Map;
 
 import org.aopalliance.intercept.MethodInterceptor;
@@ -27,6 +30,8 @@ import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.ListableBeanFactory;
+import org.springframework.beans.factory.support.DefaultListableBeanFactory;
+import org.springframework.core.OrderComparator;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ReflectionUtils;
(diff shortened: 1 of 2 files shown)


ttddyy:
The default `PersistenceExceptionTranslator`s, such as `AbstractEntityManagerFactoryBean`, are not ordered, so they always come last.
For my usecase, I need my custom one to be first, so it is fine. But, if custom translators need to come after defaults, then, default ones need to have some ordering numbers assigned.

spring-tx/src/main/java/org/springframework/dao/support/PersistenceExceptionTranslationInterceptor.java
```diff
@@ -167,8 +172,20 @@ protected PersistenceExceptionTranslator detectPersistenceExceptionTranslators(L
...
 		Map<String, PersistenceExceptionTranslator> pets = BeanFactoryUtils.beansOfTypeIncludingAncestors(
 				beanFactory, PersistenceExceptionTranslator.class, false, false);
+
+		Comparator<Object> comparatorToUse = null;
+		if (beanFactory instanceof DefaultListableBeanFactory) {
+			comparatorToUse = ((DefaultListableBeanFactory) beanFactory).getDependencyComparator();
+		}
+		if (comparatorToUse == null) {
+			comparatorToUse = OrderComparator.INSTANCE;
+		}
```
NathanQingyangXu: I think the above code can be more compact and elegant as
```
Comparator<Object> comparatorToUse = OrderComparator.INSTANCE;
if (beanFactory instanceof DefaultListableBeanFactory) {
	comparatorToUse = ((DefaultListableBeanFactory) beanFactory).getDependencyComparator();
}
```
or simply
```
Comparator<Object> comparatorToUse;
if (beanFactory instanceof DefaultListableBeanFactory) {
	comparatorToUse = ((DefaultListableBeanFactory) beanFactory).getDependencyComparator();
}
else  {
	comparatorToUse = OrderComparator.INSTANCE;
}
```

ttddyy: `DefaultListableBeanFactory#getDependencyComparator()` is  `@Nullable`, so it might return `null`.

jhoeller:
As indicated on #25559, I've resolved this in a different way, calling `getBeanProvider` with its `orderedStream` method (which takes the factory-level dependency comparator into account even for `@Order` definitions on `@Bean` methods). In order to preserve non-eager init semantics, we're specifically a new `getBeanProvider` variant with an explicit `allowEagerInit` flag there now.

PR #24648, opened 2020-03-05
Fix meta-annotations: @Timed and @Repeat

Ammar0:
The javadoc of the annotations `@Timed` & `@Repeat` says: "This annotation may be used as a meta-annotation to create custom composed annotations."

The `@Inherited` annotation is missing in the definitions of both annotation. We should add the `@Inherited` annotation to make them meta-annotations or update the javadoc.

// Steps to reproduce the problem:

```
@Timed(millis = 5000L)
@Retention(RUNTIME)
@Target(TYPE)
@interface MetaAnnotation {}

@MetaAnnotation
class TargetClass {
}

Timed timed = TargetClass.class.getAnnotation(Timed.class);
// timed is null
```

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/annotation/Repeat.java
+++ b/spring-test/src/main/java/org/springframework/test/annotation/Repeat.java
@@ -21,6 +21,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.lang.annotation.Inherited;
 
 /**
  * Test annotation for use with JUnit 4 to indicate that a test method should be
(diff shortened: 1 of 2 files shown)


sbrannen:
The `@Inherited` annotation has no effect on whether an annotation can be used as a meta-annotation. Thus, the Javadoc is correct.

The fact that `@Timed` and `@Repeat` are both meta-annotated with `@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})` _allows_ them to be declared on other annotations (i.e., "used as a meta-annotation").

`Class#getAnnotation(...)` does not support lookups of meta-annotations.

If you want meta-annotation support, you would need to use one of Spring's annotation search algorithms that supports it. For example, Spring itself uses `org.springframework.test.annotation.TestAnnotationUtils` to lookup `@Timed` and `@Repeat` via `org.springframework.core.annotation.AnnotatedElementUtils.findMergedAnnotation(AnnotatedElement, Class<Timed>)`.

I am therefore closing this PR as invalid.

PR #24651, opened 2020-03-06
Introduce getContentAsString and getContentAsByteArray to Resource

derrick-pericipio:
Projects that use java heavily rely on a clunky, verbose, and error prone method to access Resources for tests. 

Adding the ability to request a file's contents as a string for testing REST or WEBMVC allows for cleaner test classes and more readable code.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
+++ b/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
@@ -323,7 +323,7 @@ public abstract class AbstractFileResolvingResource extends AbstractResource {
 	}
 
 	@Override
-	public String getContentAsString(Charset encoding) throws IOException {
+	public String getContentAsString(Charset charset) throws IOException {
 
 		if( !exists() ) {
 			throw new FileNotFoundException(getDescription() + " cannot be found.");
@@ -331,7 +331,7 @@ public abstract class AbstractFileResolvingResource extends AbstractResource {
 		if ( !isReadable() ) {
 			throw new IOException(getDescription() + " cannot be opened for reading.");
 		}
-		return new String(Files.readAllBytes(Paths.get(getFile().getAbsolutePath())), encoding);
+		return new String(Files.readAllBytes(Paths.get(getFile().getAbsolutePath())), charset);
 
 	}
 
--- a/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
+++ b/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
@@ -25,7 +25,10 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.nio.channels.FileChannel;
 import java.nio.channels.ReadableByteChannel;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
+import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 
 import org.springframework.util.ResourceUtils;
@@ -298,6 +301,39 @@ public abstract class AbstractFileResolvingResource extends AbstractResource {
 		con.setRequestMethod("HEAD");
 	}
 
+	/**
+	 * This implementation returns the contents of a file as a string using the
+	 * system default Charset. Provided the resource exists and the context has
+	 * access to it, the contents will be returned as a single string with line
+	 * feed characters retained.
+	 * @return the contents of the requested file as a {@code String}.
+	 * @throws FileNotFoundException in the event the file path is invalid.
+	 * @throws IOException if the file can not be read or cannot be serialzied.
+	 */
+	@Override
+	public String getContentAsString() throws IOException {
+
+		if( !exists() ) {
+			throw new FileNotFoundException(getDescription() + " cannot be found.");
+		}
+		if ( !isReadable() ) {
+			throw new IOException(getDescription() + " cannot be opened for reading.");
+		}
+		return new String(Files.readAllBytes(Paths.get(getFile().getAbsolutePath())), Charset.defaultCharset());
+	}
+
+	@Override
+	public String getContentAsString(Charset encoding) throws IOException {
+
+		if( !exists() ) {
+			throw new FileNotFoundException(getDescription() + " cannot be found.");
+		}
+		if ( !isReadable() ) {
+			throw new IOException(getDescription() + " cannot be opened for reading.");
+		}
+		return new String(Files.readAllBytes(Paths.get(getFile().getAbsolutePath())), encoding);
+
+	}
 
 	/**
 	 * Inner delegate class, avoiding a hard JBoss VFS API dependency at runtime.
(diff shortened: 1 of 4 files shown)


sbrannen:
Thanks for the PR.

I'm not yet sure if we'll introduce functionality like this, but if we do, we'll need to take the following into account.

1. Adding a `compile` dependency on `commons-io` is not an option. Instead, we'd likely want to use our own `FileCopyUtils` (potentially introducing new methods if necessary to allow for configurable encoding).
2. The code must adhere to the project's [code style](https://github.com/spring-projects/spring-framework/wiki/Code-Style). Executing `./gradlew check` would be a good place to start to see Javadoc violations.
3. The name `getString()` is potentially misleading. A better name would be something like `getContentAsString()`.
4. Why would we want to limit this to `ClassPathResource`? Why not introduce it in `Resource`?

mp911de:
That's a neat optimization that sounds useful on the `Resource` level.

derrick-pericipio:
All great feedback:

- Not sure why my local build didnt flag the code style violation.. my bad. 
- I wondered what level to introduce this at so I started higher rather than lower. 
- I'll look at `FileCopyUtils`, can you elaborate on the desire to avoid commons?  It's already on the class path at the parent project level. Ultimately it just makes the implementation cleaner, it's not necessary for functionality. 

- Thoughts on throwing both `FileNotFoundException` and `IOException`?  I couldn't think of a more appropriate set when I put this together.

sbrannen:
As @jhoeller mentioned elsewhere, have you considered wrapping your `ClassPathResource` in an `EncodedResource` and then passing `encodedResource.getReader()` to `org.springframework.util.FileCopyUtils.copyToString(Reader)`?

derrick-pericipio:
@sbrannen 

My ultimate goal was a developer quality of life improvement to bring this on par with the Kotlin experience.  I will work on implementing this without Commons

derrick-pericipio:
I have pushed the implementation of the new function `getContentAsString()` down to the Resource interface and provided it with a default behavior of `toString()` to avoid having to override in both the Web and MVC modules.  

The Concrete implementation was provided at the `AbstractFileResolvingResource` level as any local or jar file should be able to be serialized. It no longer uses Apache Commons and uses the same base raw Java implementation as the Kotlin equivalent.  This does rely on Java version being 7 or higher which I did not consider an issue as the minimum for the project is 8.

Ultimately I expect this to be utilized during Contract Driven and Test Driven Development processes to avoid having to use Raw String variables and to pull developers away from using ResourceUtils directly with either Commons or the overtly verbose `Files.readBytes(File.path, Charset))` and having to handle the exceptions on their own (or forgetting) 

This is a largely QOL enhancement to the Java developer experience as Kotlin provides a convenience function known as `getResource(path).readText()` 

An alternative reasoning behind this enhancement is to close the need for access to ResourceUtils outside of the spring framework and allow for moving that Class to package private in a future change.

spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
```diff
@@ -298,6 +301,39 @@ protected void customizeConnection(HttpURLConnection con) throws IOException {
...
+			throw new FileNotFoundException(getDescription() + " cannot be found.");
+		}
+		if ( !isReadable() ) {
+			throw new IOException(getDescription() + " cannot be opened for reading.");
+		}
+		return new String(Files.readAllBytes(Paths.get(getFile().getAbsolutePath())), Charset.defaultCharset());
+	}
+
+	@Override
+	public String getContentAsString(Charset encoding) throws IOException {
```
NathanQingyangXu: I would suggest renaming 'encoding' to 'charset' to avoid confusion. Technically speaking, charset has different meaning than 'encoding'. I double checked naming convention in JDK and found either 'charset' or 'cs' is used. For string type parameter, 'charsetName' is used.

derrick-pericipio: I've changed the param name to `charset`

derrick-pericipio: Let me know if there's anything else that can be improved.

NathanQingyangXu: no further concern from my side

derrick-pericipio:
Please let me know if there is any other change desired for this enhancement. 

If not, I look forward to potentially seeing this contribution join the Spring Ecosystem.

rstoyanchev:
Couldn't this be implemented for `FileSystemResource` as well?

If the primary use case for this is in tests, I'm not sure I like seeing the methods directly on `Resource`. Another option could be `ResourceUtils` but even then the implementation looks so straight forward, we could just make it a test utility somewhere. 

You mentioned for "testing REST or WEBMVC". Is there a specific place you've noticed where we could add this so resources are logged?

derrick-anderson:
Picking this back up on my new account.  @rstoyanchev.

When i mentioned the use case it was in reference to situations when it made sense to serve static content from a jar file.  Often i've seen it for sample json payloads for dummy rest endpoints (when users could consume a stub jar etc) 

I can implement it for the `FileSystemResource` if that makes sense.  I pushed it down to Resource at @sbrannen 's reccomendation.  

Where do we think this should belong?

spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
```diff
@@ -298,6 +301,39 @@ protected void customizeConnection(HttpURLConnection con) throws IOException {
...
+	@Override
+	public String getContentAsString() throws IOException {
+
+		if( !exists() ) {
+			throw new FileNotFoundException(getDescription() + " cannot be found.");
+		}
+		if ( !isReadable() ) {
+			throw new IOException(getDescription() + " cannot be opened for reading.");
+		}
+		return new String(Files.readAllBytes(Paths.get(getFile().getAbsolutePath())), Charset.defaultCharset());
```
poutsma, line 322: Relying on the default charset is a recipe for disaster, as it can vary between environments and make code fail in collocation that runs fine in another.

Because we cannot determine the resource's charset with a 100% certainty, `Resource::getContentAsString()` should be removed, or it should be defined to use UTF-8 as default. (see `Files::readString`)

derrick-anderson: Thanks Poutsma! 

As this function is the core of this feature proposal I'll update the request to use utf-8 by default instead of throwing out the whole PR lol.

derrick-anderson:
Thank you for the update!  I'll review the updates this weekend.
Derrick Anderson (He/Him)
Practice Lead | Managed Services
–––––––––––––––––––

[image: HorizontalLockupBlack1666365038.png]
<https://www.kinandcarta.com/en-us/>


We exist to build a world that works better for everyone.
kinandcarta.com <https://www.kinandcarta.com/en-us/>

[image: linkedin40px1666368359.png]
<https://www.linkedin.com/company/kinandcarta> [image:
instagram40px1666368101.png] <https://www.instagram.com/kinandcarta> [image:
twitter35px1666365760.png] <https://twitter.com/kinandcarta> [image:
facebook40px1666369526.png] <https://www.facebook.com/kinandcarta>

Consulting Magazine's 2022 | Best Large Firm to Work For
<https://www.kinandcarta.com/en-us/insights/2022/09/2022-consulting-magazine-best-firm-to-work-for/>
Thread | Our Magazine <https://www.kinandcarta.com/en/thread/>

PR #24654, opened 2020-03-07
Fix common HashMap initial capacity issue in HttpMethod

NathanQingyangXu:
It is a common error to specify HashMap's initial capacity as expected size. It doesn't take `load factor` into consideration, which is never 1.0.
Fixed the issue in HttpMethod via the util method from guava library.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpMethod.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpMethod.java
@@ -16,9 +16,8 @@
 
 package org.springframework.http;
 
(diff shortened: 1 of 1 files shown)


chenqimiao:
@NathanQingyangXu The final mapping.size is 6, there seems to be no problem setting the initialization to 16.

NathanQingyangXu:
> @NathanQingyangXu The final mapping.size is 8, there seems to be no problem setting the initialization to 16.

Yeah, saw that. thanks for your feedback. I closed this PR.

PR #24666, opened 2020-03-10
Cache the default logger for recovery in FailFastProblemReporter

chenqimiao:
I feel that the default logger can be cached here.
If we do this, we can avoid duplicate logger creation, when calling setLogger with a null parameter.


snicoll:
Thanks for the PR but that complexifies the code quite a bit and an additional field for no concrete benefit.

PR #24668, opened 2020-03-10
use StringBuilder instead of StringBuffer in ReflectUtils#getClass

GungnirLaevatain:
It could use StringBuilder instead of StringBuffer in ```ReflectUtils#getClass```


sbrannen:

PR #24669, opened 2020-03-10
Optimize global advisor registration in ProxyFactoryBean

chenqimiao:
In the source code, `Map<Object, String> names = new HashMap<>(beans.size());`, `beans.size()` actually equal `0`. This problem will cause the hashMap to resize multiple times in subsequent `put` operations. This operation of resize is expensive when the expected size is large. So, I think we can set a suitable `initialCapacity` value to `names`  of  `HashMap` type, doing this prevents `resize`.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/ProxyFactoryBean.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/ProxyFactoryBean.java
@@ -48,6 +48,7 @@ import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
+import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


spring-aop/src/main/java/org/springframework/aop/framework/ProxyFactoryBean.java
```diff
@@ -522,8 +522,13 @@ private void addGlobalAdvisor(ListableBeanFactory beanFactory, String prefix) {
 				BeanFactoryUtils.beanNamesForTypeIncludingAncestors(beanFactory, Advisor.class);
 		String[] globalInterceptorNames =
 				BeanFactoryUtils.beanNamesForTypeIncludingAncestors(beanFactory, Interceptor.class);
-		List<Object> beans = new ArrayList<>(globalAdvisorNames.length + globalInterceptorNames.length);
-		Map<Object, String> names = new HashMap<>(beans.size());
```
chenqimiao: Beans.size() is equal to 0, because there is no element in beans at this time, even if the initial capacity is set for it.

jhoeller:
On review, I've applied a more general revision: avoiding unnecessary trace logging in ProxyFactoryBean, allowing us to skip that beans map completely and therefore never having to compute its capacity to begin with. Thanks for pointing this out, and thanks for the pull request in any case!

PR #24677, opened 2020-03-11
Minor refactoring

YooInKeun:
Minor refactoring to have uniformity of code and remove unnecessary import


sbrannen:
Thanks for the PR.

Although static imports for methods are typically not allowed in production code,
they are permissible in test classes and are primarily a matter of taste.

In light of that, I am declining the proposed change.

YooInKeun:
Thank you for your kine review👍

PR #24679, opened 2020-03-11
Extract traceEnabled for reuse in RuleBasedTransactionAttribute

chenqimiao:
Fix logger.isTraceEnabled check before the last logger.trace.
And extract an traceEnabled variable for reuse.

### DIFF
--- a/spring-tx/src/main/java/org/springframework/transaction/interceptor/RuleBasedTransactionAttribute.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/interceptor/RuleBasedTransactionAttribute.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


spring-tx/src/main/java/org/springframework/transaction/interceptor/RuleBasedTransactionAttribute.java
```diff
@@ -147,13 +148,15 @@ public boolean rollbackOn(Throwable ex) {
...
-		if (logger.isTraceEnabled()) {
+		if (traceEnabled) {
 			logger.trace("Winning rollback rule is: " + winner);
 		}
 
 		// User superclass behavior (rollback on unchecked) if no rule matches.
 		if (winner == null) {
-			logger.trace("No relevant rollback rule found: applying default rules");
+			if(traceEnabled) {
+				logger.trace("No relevant rollback rule found: applying default rules");
```
NathanQingyangXu: The above statement won't incur unnecessary perf overhead like in other cases (string concatenation has to be done beforehand) and IMO the logging level checking is not that necessary.

chenqimiao: Log level check whether it is necessary depends on your logging system.
I think Spring needs this check for compatibility.

chenqimiao: > The above statement won't incur unnecessary perf overhead like in other cases (string concatenation has to be done beforehand) and IMO the logging level checking is not that necessary.

Log level check whether it is necessary depends on your logging system.
I think Spring needs this check for compatibility.

sbrannen:
Thanks for the PR.

For these particular use cases, we don't think there is much benefit in applying the proposed changes. If trace logging becomes a performance issue, we might revisit whether we need to trace logging at all.

chenqimiao:
@sbrannen I get it, thank you for your pointing.

Issue #22060, opened 2018-11-21
Efficient type matching support for large number of beans [SPR-17528]


spring-projects-issues:
**[Rahul Shinde](https://jira.spring.io/secure/ViewProfile.jspa?name=rsjira22)** opened **[SPR-17528](https://jira.spring.io/browse/SPR-17528?redirect=false)** and commented

During profiling the application, it was found that around 19% of time was spent in the various BeanFactory methods.
See JFR snapshot:

![sbf_jfr_before.png](https://jira.spring.io/secure/attachment/26106/sbf_jfr_before.png)

Our application has over 10K beans and around 3K beans loaded at startup (~10% of them during and remaining immediately after ApplicationContext refresh).
The current implementation of _DefaultListableBeanFactory::doGetBeanNamesForType_ invokes _AbstractBeanFactory::predictBeanType_ twice for every type lookup (one via _isFactoryBean_ and the other during _isTypeMatch_)
As the _doGetBeanNamesForType_ implementation iterates over all the available bean definitions, in our case, the number of calls to predictBeanType quickly escalate to 10K * 3K * 2 = 60M.

The existing cache inside of _DefaultListableBeanFactory_ does help for subsequent lookups though.
I initially tried to minimize the number of calls to predictBeanType by adding a caching layer however the code in doGetBeanNamesForType seemed a bit hard to understand given the various corner cases it deals with.

![sbf_code.png](https://jira.spring.io/secure/attachment/26108/sbf_code.png) (_107.01 kB_)

It gave 10 seconds improvements however the lookup in the cache started showing up as the new hotspot.

Since the iteration over all the 10K bean definitions for every type lookup seemed to be the root cause, after several attempts, this is what I came up with (See PR).
The implementation goes over all the bean definitions and caches the type to name mapping once. It grabs the super types as well to avoid supplying a partial list of bean names in some cases.

Here is the JFR Hot Methods after applying the change:

![sbf_jfr_after.png](https://jira.spring.io/secure/attachment/26107/sbf_jfr_after.png) (_89.60 kB_)

This has helped speed up our application by around 25 seconds with no other change.
So far, it has been working fine for us on all our automated tests.
I would like to get your opinion on the change and if you think this would cause any long term breakages as Spring evolves.

Also it would be great if the core implementation itself can support more efficient lookups for types with sizeable beans in the registry.

 

---

**Affects:** 5.1 GA

**Attachments:**
[...]

PR #24681, opened 2020-03-11
Improve performance of AbstractAutowireCapableBeanFactory#predictBeanType

dreis2211:
Hi,

I am currently trying to improve startup time of a medium sized Spring-Boot app that currently takes around 20 seconds to start. While profiling I noticed that the top hotspot is `AbstractAutowireCapableBeanFactory#predictBeanType` as shown in the screenshot below:

image

Additionally you can see`Iterator.hasNext()/next()` which are also in the top of the list. I found out that a reasonable time is spent iterating over `BeanPostProcessors` (23 in my case) for every bean, although most of the post-processors are not a `SmartInstantiationAwareBeanPostProcessor` at all (only 3 in my case), which are the only post-processors of interest for predicting bean types. Even worse, almost all standard smart post-processors are anyway returning null.

The idea of this PR is to hold an additional list of smart post-processors that can be used to minimize the iteration overhead in cases where only `SmartInstantiationAwareBeanPostProcessor` instances are of interest. Applying this patch shows the following results in the profiling and saves around 1-2 seconds for me:

image

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/autoproxy/target/AbstractBeanFactoryBasedTargetSourceCreator.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/autoproxy/target/AbstractBeanFactoryBasedTargetSourceCreator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


dreis2211:
If #24756 is merged, this PR needs to adapt as well.

dreis2211:
Anyone willing to review?

jhoeller:
I suppose you are trying to optimize singleton creation performance there, that is, many separate beans of distinct types? I was wondering before whether to cache post-processor applicability to individual beans but that would only improve repeated creation attempts for non-singleton beans...

dreis2211:
It's indeed many separate `@Service`s or `@Component`s we're talking about, each of them autowiring a couple of other beans (let's say 5 on average - but that is only based on a gut-feeling)

dreis2211:
Any chance this might be included in the next release? (5.3.0 or even better: 5.2.7)

dreis2211:
I think that could also help on https://github.com/spring-projects/spring-framework/issues/22060

jhoeller:
I've implemented a different approach: an internal cache for pre-filtered post-processors, used within the `AbstractBeanFactory` hierarchy with no impact on outside packages. This is as fine-grained as possible now, differentiating not only `SmartInstantiationAware` but regular `InstantiationAware`, `DestructionAware` and `MergedBeanDefinitionPostProcessor` as well, also replacing the existing `hasInstantiationAware` and `hasDestructionAware` fields with the new single volatile post-processor cache... which is getting reset in `addBeanPostProcessor`.

To be committed ASAP, I'll simply repurpose this PR for it (I hope you're ok with that approach).

dreis2211:
Obviously, I appreciate a contribution as much as anyone - but I care more about the problem being fixed, so feel free to go on. I'll take a look once you committed it, because I can't really tell from your description if that actually solves the problem.

PR #24682, opened 2020-03-12
Improve javadoc in RollbackRuleAttribute regarding nested classes

chenqimiao:
In `public RollbackRuleAttribute(Class<?> clazz) `, only store the class name of the specified class, and in `private int getDepth(Class<?> exceptionClass, int depth) `, matching logic is by
`exceptionClass.getName().contains(this.exceptionName)`.
So, the enclosed class of the specified parameter of `public RollbackRuleAttribute(Class<?> clazz)` also can be match.
Suggest Improve javadoc in corresponding method.

### DIFF
--- a/spring-tx/src/main/java/org/springframework/transaction/interceptor/RollbackRuleAttribute.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/interceptor/RollbackRuleAttribute.java
@@ -54,7 +54,7 @@ public class RollbackRuleAttribute implements Serializable{
 	/**
 	 * Create a new instance of the {@code RollbackRuleAttribute} class.
 	 * <p>This is the preferred way to construct a rollback rule that matches
-	 * the supplied {@link Exception} class (and subclasses).
+	 * the supplied {@link Exception} class (, subclasses and enclosed class).
 	 * @param clazz throwable class; must be {@link Throwable} or a subclass
 	 * of {@code Throwable}
 	 * @throws IllegalArgumentException if the supplied {@code clazz} is
--- a/spring-tx/src/main/java/org/springframework/transaction/interceptor/RollbackRuleAttribute.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/interceptor/RollbackRuleAttribute.java
@@ -54,7 +54,7 @@ public class RollbackRuleAttribute implements Serializable{
 	/**
 	 * Create a new instance of the {@code RollbackRuleAttribute} class.
 	 * <p>This is the preferred way to construct a rollback rule that matches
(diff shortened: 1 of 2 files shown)


sbrannen (changes requested):
Please introduce a test which demonstrates that an enclosing class would result in a positive match.

chenqimiao:
@sbrannen Thank you for your review, I have submitted a unit test to demonstrate that an enclosing class would result in a positive match.

spring-tx/src/main/java/org/springframework/transaction/interceptor/RollbackRuleAttribute.java
```diff
@@ -54,7 +54,7 @@
 	/**
 	 * Create a new instance of the {@code RollbackRuleAttribute} class.
 	 * <p>This is the preferred way to construct a rollback rule that matches
-	 * the supplied {@link Exception} class (and subclasses).
+	 * the supplied {@link Exception} class (, subclasses and enclosed class).
```
sbrannen:
```suggestion
	 * the supplied {@link Exception} class, its subclasses, and its nested classes.
```

sbrannen (changes requested):

spring-tx/src/test/java/org/springframework/transaction/interceptor/RollbackRuleTests.java
```diff
@@ -88,4 +88,19 @@ public void ctorArgExceptionStringNameVersionWithNull() {
...
 	}
 
+	@Test
+	public void foundEnclosedExceptionWithEnclosingException() {
+		RollbackRuleAttribute rr = new RollbackRuleAttribute(EnclosingException.class);
+		assertThat(rr.getDepth(new EnclosingException.EnclosedException())).isEqualTo(0);
+	}
+
+	@SuppressWarnings("serial")
+	static class EnclosingException extends RuntimeException{
```
sbrannen:
```suggestion
	static class EnclosingException extends RuntimeException {
```

spring-tx/src/test/java/org/springframework/transaction/interceptor/RollbackRuleTests.java
```diff
@@ -88,4 +88,19 @@ public void ctorArgExceptionStringNameVersionWithNull() {
...
+
+		@SuppressWarnings("serial")
+		static class EnclosedException extends RuntimeException{
```
sbrannen:
```suggestion
		static class EnclosedException extends RuntimeException {
```

sbrannen (changes requested):

sbrannen:

Thanks

PR #24684, opened 2020-03-12
Remove unnecessary setter in DefaultServerRequest

YooInKeun:
I propose to remove `public void setStatus(int sc, String sm)` setter method in `class CheckNotModifiedResponse`.

Because setting status message code is not being used in `public void setStatus(int sc, String sm)`.
ex) `this.statusMsg = sm;`

And I know `public void setStatus(int sc, String sm)` in HttpServletResponse is deprecated.

There is an explanation about this in https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/http/HttpServletResponse.html#setStatus(int,%20java.lang.String)

`As of version 2.1, due to ambiguous meaning of the message parameter. To set a status code use setStatus(int), to send an error with a description use sendError(int, String). Sets the status code and message for this response.`

Thank you for reading my PR!

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/function/DefaultServerRequest.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/function/DefaultServerRequest.java
@@ -486,11 +486,6 @@ class DefaultServerRequest implements ServerRequest {
 		}
 
 		@Override
-		public void setStatus(int sc, String sm) {
-			this.status = sc;
-		}
-
-		@Override
 		public int getStatus() {
 			return this.status;
 		}


rstoyanchev:
Uh, it's a compile error.

YooInKeun:
Thank you! I missed that part.
Then, how about changing `public void setStatus(int sc, String sm)` like below?
I propose to classify it as unsupported method.

```
// Unsupported

...
 
@Override
public void setStatus(int sc, String sm) {
	throw new UnsupportedOperationException();
}
```

rstoyanchev:
This is a private implementation of `HttpServletResponse` that is used internally for `checkNotModified` checks and will never be exposed to the outside. Thanks for the proposal but it has no practical benefit.

YooInKeun:
Thank you for your kind review👍

PR #24686, opened 2020-03-13
Fix exception message in getMergedBeanDefinition

chenqimiao:
From the context, it feels like it should be `ConfigurableBeanFactory`, not `AbstractBeanFactory`.


spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractBeanFactory.java
```diff
@@ -1343,7 +1343,7 @@ protected RootBeanDefinition getMergedBeanDefinition(
 							else {
 								throw new NoSuchBeanDefinitionException(parentBeanName,
 										"Parent name '" + parentBeanName + "' is equal to bean name '" + beanName +
-										"': cannot be resolved without an AbstractBeanFactory parent");
+										"': cannot be resolved without an ConfigurableBeanFactory parent");
```
sbrannen:
```suggestion
										"': cannot be resolved without a ConfigurableBeanFactory parent");
```

sbrannen (changes requested):

sbrannen:

Thanks

PR #24688, opened 2020-03-13
Improve Javadoc in AutoProxyRegistrar

chenqimiao:
`EnableAspectJAutoProxy` doesn't have `mode` attribute, a bit mismatch with the logic of `registerBeanDefinitions` method in `AutoProxyRegistrar`. Suggest adjusting `EnableAspectJAutoProxy` to `EnableCaching `.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/AutoProxyRegistrar.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/AutoProxyRegistrar.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


spring-context/src/main/java/org/springframework/context/annotation/AutoProxyRegistrar.java
```diff
@@ -33,7 +33,7 @@
  *
  * @author Chris Beams
  * @since 3.1
- * @see EnableAspectJAutoProxy
+ * @see org.springframework.cache.annotation.EnableCaching
```
sbrannen:
```suggestion
 * @see org.springframework.cache.annotation.EnableCaching
 * @see org.springframework.transaction.annotation.EnableTransactionManagement
```

sbrannen (changes requested):

sbrannen:

Thanks

PR #24691, opened 2020-03-13
Do not mention credentials as enabled by default in CorsRegistry

mdindoffer:
Credentials are not enabled by default contrary to the javadocs. 
The credentials support must be enabled via #allowCredentials.


sbrannen:
Good catch!

Thanks

PR #24696, opened 2020-03-14
Improve Javadoc in InstantiationAwareBeanPostProcessor

chenqimiao:
`Post-process the given property values before the factory applies them to the given bean`
If I understand correctly, `AbstractAutowireCapableBeanFactory#applyPropertyValues` represents `the factory applies them to the given bean`.
So, suggest adding the method link in Javadoc to make the documentation clearer.


snicoll:
Thanks for the PR and sorry it took so long to process it. You are right but `AbstractAutowireCapableBeanFactory` is an implementation detail and I am not keen to add a reference to it in the interface contract.

PR #24698, opened 2020-03-14
Use Collections.emptyList() instead of new ArrayList()

chenqimiao:
It can reduce object creation.


PascalSchumacher:
Before this change the method always returned a mutable list. Now it will return an immutable list if the list is empty.

No idea how relevant it is in this case, but in general mixed mutability of return types should be avoided. For example `ErrorProne` would generate a warning for this code (see: https://errorprone.info/bugpattern/MixedMutabilityReturnType).

chenqimiao:
> Before this change the method always returned a mutable list. Now it will return an immutable list if the list is empty.
> 
> No idea how relevant it is in this case, but in general mixed mutability of return types should be avoided. For example `ErrorProne` would generate a warning for this code (see: https://errorprone.info/bugpattern/MixedMutabilityReturnType).
I get it, thank you for your pointing, I will close this PR.

PR #24704, opened 2020-03-16
Avoid unnecessary sorting in AbstractHandlerMethodMapping

chenqimiao:
This PR can reduce useless sorting in AbstractHandlerMethodMapping.


sbrannen:
Good catch!

Thanks

PR #24708, opened 2020-03-17
fix NPE in SynchronossPartGenerator#hookOnNext

GungnirLaevatain:
``listener`` is Nullable, so we should check it before using it


sbrannen:
Please note that the first line in that method is:

```java
Assert.state(this.parser != null && this.listener != null, "Not initialized yet");
```

@GungnirLaevatain, do you have a concrete use case where you experienced a `NullPointerException` in this context?

PR #24712, opened 2020-03-17
Non-public Kotlin beans can't be instantiated

tomvandenberge:
When a bean definition from an XML file is instantiated, and the bean class is a Kotlin class that is not public (e.g. protected), or a non-public constructor is used, this fails with an `IllegalAccessException: class kotlin.reflect.jvm.internal.calls.CallerImpl$Constructor cannot access a member of class MyKotlinClass with modifiers "protected"`. 

This works fine for Java classes, so I believe that this should also work for Kotlin classes.

I managed to locate and fix the cause of this problem. It is in `BeanUtils.instantiateClass(Constructor<T> ctor, Object... args)`. 

Here, the provided constructor object is made accessible before it is being used to instantiate the bean. 

Then, if the class that declares the constructor is a Kotlin class (and the `kotlin-reflect` jar is on the classpath; another prerequisite for this problem), the Kotlin `KFunction` object representing the Java constructor is obtained, and used to instantiate the bean.

Unfortunately, when this Kotlin constructor is used, the accessibility flag that was set on the original Java constructor object is not set. This is because the original Java constructor object is not the same object as the Java constructor object that is buried deep down in the Kotlin constructor object, what you might expect.

My fix sets the accessibility flag on the Kotlin constructor object itself, in a similar way as it is done for Java constructors.

I've added a test to `BeanUtilsTests` to ensure that a private Java class with a private constructor can be instantiated without problems. A few tests are added to `KotlinBeanUtilsTests` to ensure that this fix allows private and protected classes and constructors to be used.

I'm looking forward to your feedback. 
Regards,
Tom

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -38,6 +38,7 @@ import kotlin.jvm.JvmClassMappingKt;
 import kotlin.reflect.KFunction;
 import kotlin.reflect.KParameter;
 import kotlin.reflect.full.KClasses;
+import kotlin.reflect.jvm.KCallablesJvm;
 import kotlin.reflect.jvm.ReflectJvmMapping;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


sdeleuze:
@jhoeller I will take care of that one if that's fine with you.

jhoeller:
@sdeleuze Sure, I'm just keen on getting this into 5.2.5 :-)

sdeleuze:
@tomvandenberge Merged, thanks for your contribution.

PR #24717, opened 2020-03-18
Delete empty .gitignore file

YooInKeun:
I propose to delete empty .gitignore file.


sbrannen:
Please note that the `.gitignore` file is there to ensure that the `src/test/java` folder is present as required by certain build arrangements and IDEs.

In any case, I will merge this PR with the hope that it is no longer necessary.

PR #24721, opened 2020-03-18
Simplify Conventions.attributeNameToPropertyName()

YooInKeun:
I propose refactored code to simplify code and improve accuracy of Conventions.attributeNameToPropertyName() method.


rstoyanchev:
Thanks for the PR but the proposed solution is fewer in lines but is less optimal in terms of Object allocation and performance. As `Conventions` can be used in hot paths such as serving requests, so we'll keep the current solution.

YooInKeun:
Thank you for your kind review👍

PR #24728, opened 2020-03-18
Support for X-Forwarded-Prefix in UrlComponentsBuilder

### LINKED ISSUE #24723: Support for X-Forwarded-Prefix in Spring HATEOAS
Spring HATEOAS doesn't currently support the `X-Forwarded-Prefix` header.

> NOTE: Examples are illustrative

We are using a reverse proxy / API gateway to serve a HATEOAS webflux project. Our API gateway accepts requests at:

`https://api.example.com/context/resource`

Our backend service address for this is:

`https://backend.example.com/resource`

It returns the following HATEOAS links

```
{
  "_links": {
    "self": {
      "href": "https://backend.example.com/resource"
    }
  }
}
```

By using the following headers:

```
X-Forwarded-Host: api.example.com
X-Forwarded-Prefix: /context
```

We would expect the result to be

```
{
  "_links": {
    "self": {
      "href": "https://api.example.com/context/resource"
    }
  }
}
```

Instead we are seeing that the `X-Forwarded-Host` header is being applied correctly, but the `X-Forwarded-Prefix` header is not, e.g.:

```
{
  "_links": {
    "self": {
      "href": "https://api.example.com/resource"
    }
  }
}
```

The `WebFluxLinkBuilder` used to create HATEOAS links leverages the `UrlComponentsBuilder` from spring-web, which is [missing the handling logic for the `X-Forwarded-Prefix` header](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java#L739). The [ForwardedHeaderTransformer (also in spring-web) does handle this header](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/server/adapter/ForwardedHeaderTransformer.java#L60).

I'm proposing having this added to the `UrlComponentsBuilder`.

stuartwakefield:
Adds support in UriComponentsBuilder to handle X-Forwarded-Prefix header used by reverse proxies to specify the segment of the URL path that has been stripped by that reverse proxy. It is especially useful in conjunction with the X-Forwarded-Host header for building location-dependent content such as HATEOAS links, where the generated links should direct the client to further resources via the proxy rather than the target server.

The ForwardedHeaderTransformer does not need to handle the X-Forwarded-Prefix header as a special case as it is now handled by the UrlComponentsBuilder.

Closes #24723

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/adapter/ForwardedHeaderTransformer.java
+++ b/spring-web/src/main/java/org/springframework/web/server/adapter/ForwardedHeaderTransformer.java
@@ -94,11 +94,6 @@ public class ForwardedHeaderTransformer implements Function<ServerHttpRequest, S
 			if (!this.removeOnly) {
 				URI uri = UriComponentsBuilder.fromHttpRequest(request).build(true).toUri();
 				builder.uri(uri);
-				String prefix = getForwardedPrefix(request);
-				if (prefix != null) {
-					builder.path(prefix + uri.getRawPath());
-					builder.contextPath(prefix);
-				}
 			}
 			removeForwardedHeaders(builder);
 			request = builder.build();
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -778,7 +778,7 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 
 				String prefixHeader = headers.getFirst("X-Forwarded-Prefix");
 				if (StringUtils.hasText(prefixHeader)) {
-					PathComponent pathComponent = pathBuilder.build();
+					PathComponent pathComponent = this.pathBuilder.build();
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@stuartwakefield Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24728)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24728) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@stuartwakefield Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24728)!

rstoyanchev:
Closing this for now to keep the discussion in one place under #24723. Generally, you don't need to submit both an issue and a PR. We keep only one open.

PR #24731, opened 2020-03-19
Remove unnecessary condition in LocalVariableTableParameterNameDiscoverer

YooInKeun:
I propose to remove unnecessary condition in `LocalVariableTableParameterNameDiscoverer`.


sbrannen:
Thanks for the PR.

Although `org.springframework.asm.ClassReader.readStream(InputStream, boolean)` would eventually thrown an `IOException` for a `null` `InputStream`, we shouldn't rely on that.

In addition, aborting immediately when the class file is known not to exist is an optimization of the code in question.

PR #24740, opened 2020-03-20
Improve Javadoc in AnnotatedElementUtils

chenqimiao:
In fact, an empty set will be returned if none found in `getMetaAnnotationTypes` method.
In addition, `org.springframework.core.annotation.AnnotatedElementUtilsTests#getMetaAnnotationTypesOnNonAnnotatedClass`  method can also verify this。


sbrannen:

Thanks

PR #24744, opened 2020-03-20
Remove unnecessary null check in ConvertingEncoderDecoderSupport

chenqimiao:
`instanceof` check contains `null check`, I think we can remove leading `null` check.


rstoyanchev:
This has been merged.

PR #24756, opened 2020-03-21
Add BeanPostProcessors in bulk

stsypanov:
In one of my applications there's about 50 post processors added in loop to `CopyOnWriteArrayList`. I think this can be improved by using bulk method where lock/unlock happens only once for all PP and less garbage is produced inside of `CopyOnWriteArrayList`.


jhoeller:
I'll cover this as part of a larger revision which also covers #24681. Note that I opted for keeping the public surface limited as-is, only exposing `addBeanPostProcessors` on `AbstractBeanFactory` and checking for it in `PostProcessorRegistrationDelegate` (which we do there for other purposes already).

PR #24758, opened 2020-03-22
Fix empty domain handling in ResponseCookie

izeye:
This PR fixes empty domain handling in `ResponseCookie`.

This PR also updates its test to use its intended inputs.


rstoyanchev:
Thanks for the catch!

PR #24773, opened 2020-03-25
Support method injection when @Autowired(require=false) is used on mixed nullability args

chenqimiao:
When `@Autowired(require=false)` on mixed nullability args method, I think IOC container shoud call this method to inject partially non-null parameters, but in fact the current code logic will ignore this method injection, for example `:`

```java
	static class MixedNullableInjectionBean{
		public NonNullBean nonNullBean;
		public NullableBean nullableBean;

		@Autowired(required = false)
		public void nullabilityInjection(@Nullable NullableBean nullableBean, NonNullBean nonNullBean){
			if(nullableBean != null){
				this.nullableBean = nullableBean;
			}
			this.nonNullBean = nonNullBean;
		}
	}
```

```java
	@Test
	public void testMethodInjectionWithMultiMixedNullableArgs(){
		bf.registerBeanDefinition("nonNullBean", new RootBeanDefinition(
				NonNullBean.class));
		bf.registerBeanDefinition("mixedNullableInjectionBean", new RootBeanDefinition(MixedNullableInjectionBean.class));
		MixedNullableInjectionBean mixedNullableInjectionBean = bf.getBean(MixedNullableInjectionBean.class);
		assertThat(mixedNullableInjectionBean.nonNullBean).isNotNull();
		assertThat(mixedNullableInjectionBean.nullableBean).isNull();
	}
```
I liberated the `@Autowired(require=false)` annotation on the `nullabilityInjection` method of `MixedNullableInjectionBean`, this method has two parameters, `nullableBean` and `nonNullBean`, as the name of the parameter means, I only register `nonNullBean` to `DefaultListableBeanFactory`, and then register `MixedNullableInjectionBean` .
I think the IOC container should invoke `nullabilityInjection` method to inject `nonNullBean`,
but in fact the current code logic will ignore this method injection, I have expressed the nullability of the parameter `nullableBean`, so, I think this parameter is null should not affect the overall method injection. I slightly adjusted the logic of method injection in this pr, what do you think? Thank you for your reading and looking forward to your reply.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


snicoll:
Good stuff, thanks again @chenqimiao 

This was closed with the wrong issue reference: https://github.com/spring-projects/spring-framework/commit/d0fc6dd06d69d9095d909a02c395f456139f792d

PR #24774, opened 2020-03-25
Make resourceAnnotationTypes field final in CommonAnnotationBPP

chenqimiao:
The `resourceAnnotationTypes` field reference will not be reassigned. So suggest adding `final` keyword to this field.


sbrannen:

Thanks

PR #24782, opened 2020-03-26
Using UriComponentsBuilder.cloneBuilder does not copy uriVariables

### LINKED ISSUE #24772: Using UriComponentsBuilder.cloneBuilder does not copy uriVariables
According to the docs, when you do `UriComponentsBuilder.cloneBuilder` (`v5.2.5.RELEASE`) it should create a deep clone: 
https://github.com/spring-projects/spring-framework/blob/c08e31b7d613bf91cbe2beac2dad66714403faee/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java#L146-L162

However, if your builder has `pathSegment`s the `uriVariables` will not be preserved. 
Imagine the following scenario - you have a `base` builder which you are going to receive and you need to append path(s) to it. Here is an oversimplified example:

```java
public UriComponentsBuilder getStorageBuilder(int userId, String storageName) {
    HashMap<String, Object> uriVariables = new HashMap<>();
    uriVariables.put("userId", userId);
    uriVariables.put("storageId", storageName);

    UriComponentsBuilder baseBuilder = ServletUriComponentsBuilder.fromCurrentRequest();
    baseBuilder.replacePath("/storages");
    baseBuilder.pathSegment("{userId}");
    baseBuilder.pathSegment("{storageName}");
    baseBuilder.uriVariables(uriVariables);

    return baseBuilder;
}
....

// Receive a configured `builder` (something else calls `getStorageBuilder` before this method and it's passed as an argument here)
public UriComponentsBuilder getStorageDirectoryListing(UriComponentsBuilder baseBuilder, String path) {
    // http://localhost/storages/%7BuserId%7D/%7BstorageName%7D/some/path
    return baseBuilder.cloneBuilder().path(path).toUriString();
}
```

In this case the `cloned` builder will not have the `uriVariables` copied. Furthermore, there is no `getter` method for the `uriVariables` so you can't manually copy them. A quick'n'dirty workaround to this is would be:

```java
String fileUrl = UriComponentsBuilder.fromUriString(baseBuilder.toUriString())
                                     .path("/some/path").build().toUriString();
```

Unfortunately this removes the ability to change anything in the `pathSegments`.
Is there anything I'm missing here for `uriVariables` to be excluded when cloning or is this simply missing at the moment?

steve-todorov:
I took the liberty of adding a dedicated test which checks if all properties are properly copied when `UriComponentsBuilder.cloneBuilder()` is called. I intentionally added a separate test case since I wasn't sure if `testClone` isn't covering a use case I haven't though of.

Closes gh-24772

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -155,6 +155,7 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 		this.host = other.host;
 		this.port = other.port;
 		this.pathBuilder = other.pathBuilder.cloneBuilder();
+		this.uriVariables.putAll(other.uriVariables);
 		this.queryParams.putAll(other.queryParams);
 		this.fragment = other.fragment;
 		this.encodeTemplate = other.encodeTemplate;
--- a/spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
+++ b/spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
@@ -856,6 +856,37 @@ public class UriComponentsBuilderTests {
 		assertThat(result2.getFragment()).isEqualTo("f2");
 	}
 
+	@Test // gh-24772
+	public void testDeepClone() {
+		HashMap<String, Object> vars = new HashMap<>();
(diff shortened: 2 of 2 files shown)


rstoyanchev:
This is now merged, thanks!

steve-todorov:
Awesome, many thanks! :)

PR #24785, opened 2020-03-26
Simplify conversion of ByteArrayOutputStream to String

stsypanov:
Currently we often convert BAOS to String as
```java
String contents = new String(baos.toByteArray(), StandardCharsets.ISO_8859_1);
```
which can be made more simple and effective as
```java
String contents = baos.toString(StandardCharsets.ISO_8859_1.name());
```
`BAOS.toByteArray()` creates a copy of underlying array which is later decoded inside of `String` constructor, while `BAOS.toString()` doesn't create intermediate copy making convertion more memory-effective.

For JDK 8 I've got following results:
```
Benchmark                                                        (length)  Mode     Score     Error   Units
ByteArrayOutputStreamBenchmark.newString                               10  avgt    97.420 ±   7.340   ns/op
ByteArrayOutputStreamBenchmark.newString                              100  avgt   480.689 ±  20.615   ns/op
ByteArrayOutputStreamBenchmark.newString                             1000  avgt  4717.681 ± 175.550   ns/op

ByteArrayOutputStreamBenchmark.toString                                10  avgt    85.578 ±   5.233   ns/op
ByteArrayOutputStreamBenchmark.toString                               100  avgt   533.850 ±  23.902   ns/op
ByteArrayOutputStreamBenchmark.toString                              1000  avgt  4703.280 ± 113.713   ns/op

ByteArrayOutputStreamBenchmark.newString:·gc.alloc.rate.norm           10  avgt   134.400 ±  15.271    B/op
ByteArrayOutputStreamBenchmark.newString:·gc.alloc.rate.norm          100  avgt   722.401 ±  10.494    B/op
ByteArrayOutputStreamBenchmark.newString:·gc.alloc.rate.norm         1000  avgt  6380.016 ±  46.441    B/op

ByteArrayOutputStreamBenchmark.toString:·gc.alloc.rate.norm            10  avgt   109.600 ±  14.586    B/op
ByteArrayOutputStreamBenchmark.toString:·gc.alloc.rate.norm           100  avgt   664.002 ±  16.986    B/op
ByteArrayOutputStreamBenchmark.toString:·gc.alloc.rate.norm          1000  avgt  5433.618 ±  39.002    B/op

PR #24786, opened 2020-03-26
Simplify conversion of DataBuffer to String

stsypanov:
This PR is similar to https://github.com/spring-projects/spring-framework/pull/24785

Also there's one more place about which I'm not sure:
This code
```java
return DataBufferUtils.join(flux)
		.map(buffer -> {
			byte[] result = new byte[buffer.readableByteCount()];
			buffer.read(result);
			DataBufferUtils.release(buffer);
			return DigestUtils.md5DigestAsHex(result);
		});
```
could be rewritten as 
```java
return DataBufferUtils.join(flux)
		.map(buffer -> {
			try (InputStream inputStream = buffer.asInputStream(true)) {
				return DigestUtils.md5DigestAsHex(inputStream);
			} catch (Exception e) {
				throw new RuntimeException(e);
			}
		});
```
But I'm not sure whether it is worth the effort.


rstoyanchev:
Thanks for the pull request. I did some further refactoring along the same lines.

PR #24789, opened 2020-03-26
Remove obsolete comment in BeanDefinitionParserDelegate

chenqimiao:


spring-beans/src/main/java/org/springframework/beans/factory/xml/BeanDefinitionParserDelegate.java
```diff
@@ -90,7 +90,7 @@
 
 	/**
 	 * Value of a T/F attribute that represents true.
-	 * Anything else represents false. Case seNsItive.
+	 * Anything else represents false. Case sensitive.
```
sbrannen: I don't think this was a typo. Rather, I think it was intentional, to point out that case mattered.

However, in light of how the constant is actually used within the framework, this comment is no longer valid. Specifically, although the constant value itself is lowercase, it is sometimes used for case insensitive equality checks and other times for case sensitive equality checks.

I therefore propose to remove it entirely.

```suggestion
	 * Anything else represents false.
```

sbrannen (changes requested):

sbrannen (approved):

sbrannen:
This has been merged in a modified form into `master`.

See https://github.com/spring-projects/spring-framework/pull/24789#discussion_r398630817 for details.

Thanks

PR #24795, opened 2020-03-27
Remove the wrong @Nullable annotation in DynamicPropertiesContextCustomizer

chenqimiao:
The `buildDynamicPropertiesMap ` method may return an empty unmodifiableMap, but not null.


chenqimiao:
This PR is already included in #24796.

PR #24796, opened 2020-03-27
Remove incorrect @Nullable in DynamicPropertiesContextCustomizer

chenqimiao:
Hello~
This PR contains the following points：
* Remove incorrect `@Nullable` annotation in buildDynamicPropertiesMap, this method may return empty map, but not null.
* If the `methods ` parameter of Constructor of `DynamicPropertiesContextCustomizer` is null, it will produce NPE, and if the parameter is a empty set, it doesn't seem to make much sense, considering that this class is visible by the package, so I suggest adding a empty check in this Constructor.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizer.java
+++ b/spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizer.java
@@ -26,7 +26,6 @@ import java.util.function.Supplier;
 
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.core.env.MutablePropertySources;
-import org.springframework.lang.Nullable;
 import org.springframework.test.context.ContextCustomizer;
 import org.springframework.test.context.DynamicPropertyRegistry;
 import org.springframework.test.context.DynamicPropertySource;
--- a/spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizer.java
+++ b/spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizer.java
@@ -73,7 +73,6 @@ class DynamicPropertiesContextCustomizer implements ContextCustomizer {
 		sources.addFirst(new DynamicValuesPropertySource(PROPERTY_SOURCE_NAME, buildDynamicPropertiesMap()));
 	}
 
-	@Nullable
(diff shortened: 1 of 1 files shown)


spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizer.java
```diff
@@ -52,6 +51,7 @@
 
 
 	DynamicPropertiesContextCustomizer(Set<Method> methods) {
+		Assert.notEmpty(methods, "'methods' must not be null or empty");
```
sbrannen: Since this is a package private class and we know that this constructor is only ever invoked by `org.springframework.test.context.support.DynamicPropertiesContextCustomizerFactory.createContextCustomizer(Class<?>, List<ContextConfigurationAttributes>)`, this assertion does not add much value.

Furthermore, `DynamicPropertiesContextCustomizerFactory` will never invoke this constructor with a null or empty set of methods. If something were to invoke the constructor with `null`, that would be in violation of the package-level `@NonNullApi` declaration, and supplying an empty set would result in the registration of a `DynamicValuesPropertySource` that is effectively a no-op.

In light of that, please revert these changes in this PR.

spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizer.java
```diff
@@ -73,7 +73,6 @@ public void customizeContext(ConfigurableApplicationContext context,
 		sources.addFirst(new DynamicValuesPropertySource(PROPERTY_SOURCE_NAME, buildDynamicPropertiesMap()));
 	}
 
-	@Nullable
```
sbrannen: Good catch!

sbrannen (changes requested):
Please see my comments and refocus this PR solely on the incorrect `@Nullable` declaration.

Thanks

spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizer.java
chenqimiao: I get it, and thank you for your pointing, I have already reverted these changes.

chenqimiao:
@sbrannen I have submitted the update, thank you for your pointing.

sbrannen:

Thanks

PR #24799, opened 2020-03-27
Remove unnecessary mem alloc in CollectingReaderEventListener

chenqimiao:
The parameter of `collection.toArray()` method only need to express the type convert to , no need to specify the length of the parameter array.


sbrannen:

Thanks

PR #24800, opened 2020-03-27
Set initialValue of NamedThreadLocal in XmlBeanDefinitionReader

chenqimiao:
Implement `initialValue` method of `NamedThreadLocal` to lazy init `set` in `XmlBeanDefinitionReader`.
This will look more concise.
In addition, can you consider this PR #24705 ? Because with this PR code, the writing will be simpler.


sbrannen:

Thanks

PR #24801, opened 2020-03-27
Use autoboxing instead of explicit wrapping

chenqimiao:
This PR is considering using auto Box/Unbox instead of manual, because constructors like `new Integer(1)` have been `deprecated` since JDK 9.

There are a lot of changes in this PR.

I already ran `./gradlew check` and got a successful result, so I hope this will reduce your workload.

### DIFF
--- a/spring-context/src/test/java/org/springframework/aop/framework/ProxyFactoryBeanTests.java
+++ b/spring-context/src/test/java/org/springframework/aop/framework/ProxyFactoryBeanTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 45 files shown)


spring-context/src/test/java/org/springframework/aop/framework/ProxyFactoryBeanTests.java
```diff
@@ -10,7 +10,7 @@
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
+ * See the License for the specific languaspring-context/src/test/java/org/springframework/context/annotation/ReflectionUtilsIntegrationTests.javage governing permissions and
```
sbrannen, line 13: I'm not quite sure how you managed to modify that, but I'll fix it when merging. 😉

chenqimiao: Sorry, it should be my hand shaking.😿

sbrannen:

Thanks

PR #24803, opened 2020-03-28
Optimize hard code in PropertiesBeanDefinitionReader

chenqimiao:
This PR removes some hard codes. 
Please see my PR changes, although `CONSTRUCTOR_ARG_PREFIX` is final, but it is still possible to update iteratively, although it is unlikely, but it is still recommended to modify the hard code.


snicoll:
Thanks for the PR and sorry it took so long to process it. I don't mind too much about the number there and the class has been deprecated in the meantime so we'll keep things as they are.

PR #24864, opened 2020-04-04
Use single String instead of String[] for varargs argument

chenqimiao:


sbrannen:

Thanks

Issue #24615, opened 2020-02-28
Add Encoder and Decoder for ByteBuf


linux-china:
Now it's not possible to output ByteBuf to browser with data from RSocket, gRPC or other Netty based RPC frameworks with Spring WebFlux, and you should convert ByteBuf to ByteBuffer.  Any consideration to support ByteBuf directly.   Maybe performance enhancement without convert, and API friendly. 

```
   @GetMapping(value = "/bytebuf", produces = {MediaType.APPLICATION_JSON_VALUE})
    public Mono<ByteBuf> byteBuf() {
          //data from RSocket payload
    }
```

s-m-sufyian:
I am willing to work on this issue

poutsma:
Not sure if a codecs is the proper place for this, as I am not sure what they would do in scenario's where `ByteBuf` is not available (i.e. WebFlux on Servlet). Also, there are workarounds:

For extracting a `ByteBuf`, you can easily cast to a NettyDataBuffer, and call [`getNativeBuffer()`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/io/buffer/NettyDataBuffer.html#getNativeBuffer--). To create a `DataBuffer` from a `ByteBuf`, use [wrap](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/io/buffer/NettyDataBufferFactory.html#wrap-io.netty.buffer.ByteBuf-).

rstoyanchev:
Typically in a Spring application you with `WebClient` or `RSocketRequester` what you'll have is `DataBuffer` instead but if what you have is `ByteBuf` then we should be able to adapt. I can imagine `ByteBufEncoder` that simply with `NettyDataBufferFactory`, or otherwise copy to `ByteBuffer` and wrap, and releases the `ByteBuf`. A matching `Decoder` would make sense too and would be a simple pass-through.

Scheduling tentatively for 5.3 M1 but with a PR, we could probably include it sooner.

encircled:
Looks like s-m-sufyian is not working on it so I took it :)

rstoyanchev:
Superseded by #24866.

PR #24866, opened 2020-04-05
Add Encoder and Decoder for Netty ByteBuf

encircled:
Add support for encoding and decoding of `ByteBuf`, see #24615.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import org.springframework.core.ResolvableType;
+import org.springframework.core.io.buffer.DataBuffer;
+import org.springframework.core.io.buffer.DataBufferUtils;
+import org.springframework.core.io.buffer.NettyDataBuffer;
+import org.springframework.lang.Nullable;
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
+
+import java.util.Map;
+
+/**
+ * Decoder for {@link ByteBuf ByteBufs}.
+ *
+ * @author Vladislav Kisel
+ * @since 5.3
+ */
+public class ByteBufDecoder extends AbstractDataBufferDecoder<ByteBuf> {
+
+	public ByteBufDecoder() {
+		super(MimeTypeUtils.ALL);
+	}
+
+
+	@Override
+	public boolean canDecode(ResolvableType elementType, @Nullable MimeType mimeType) {
+		return (ByteBuf.class.isAssignableFrom(elementType.toClass()) &&
+				super.canDecode(elementType, mimeType));
+	}
+
+	@Override
+	public ByteBuf decode(DataBuffer dataBuffer, ResolvableType elementType,
+			@Nullable MimeType mimeType, @Nullable Map<String, Object> hints) {
+
+		// Copies the dataBuffer if needed only
+		ByteBuf byteBuf;
+		if (dataBuffer instanceof NettyDataBuffer) {
+			byteBuf = ((NettyDataBuffer) dataBuffer).getNativeBuffer();
+		} else {
+			byteBuf = Unpooled.wrappedBuffer(dataBuffer.asByteBuffer());
+			DataBufferUtils.release(dataBuffer);
+		}
+
+		if (logger.isDebugEnabled()) {
+			logger.debug(Hints.getLogPrefix(hints) + "Read " + byteBuf.readableBytes() + " bytes");
+		}
+		return byteBuf;
+	}
+
+}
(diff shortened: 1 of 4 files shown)


encircled:
Encoder&Decoder itself is ready I suppose. 
Please advise where do I find the tests for webflux server response encoders

spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+
+	@Override
+	public ByteBuf decode(DataBuffer dataBuffer, ResolvableType elementType,
+			@Nullable MimeType mimeType, @Nullable Map<String, Object> hints) {
+
+		int byteCount = dataBuffer.readableByteCount();
+		ByteBuf copy = Unpooled.buffer(byteCount);
+		copy.writeBytes(dataBuffer.asByteBuffer());
+		copy.readerIndex(0);
+		DataBufferUtils.release(dataBuffer);
```
rstoyanchev, line 58: If running on Netty, the copy here is unnecessarily inefficient. If `DataBuffer` is already an instance of `NettyDataBuffer` then we should simply return `((NettyDataBuffer) dataBuffer).getNativeBuffer()`.

spring-core/src/main/java/org/springframework/core/codec/ByteBufEncoder.java
```diff
@@ -0,0 +1,71 @@
...
+
+		return Flux.from(inputStream).map(byteBuffer ->
+				encodeValue(byteBuffer, bufferFactory, elementType, mimeType, hints));
+	}
+
+	@Override
+	public DataBuffer encodeValue(ByteBuf byteBuffer, DataBufferFactory bufferFactory,
+			ResolvableType valueType, @Nullable MimeType mimeType, @Nullable Map<String, Object> hints) {
+
+		DataBuffer dataBuffer = bufferFactory.wrap(byteBuffer.nioBuffer());
```
rstoyanchev, line 63: If running on Netty, it would be better to simply wrap the given `ByteBuf`, e.g. `((NettyDataBufferFactory) bufferFactory.wrap(byteBuffer)`.

spring-core/src/test/java/org/springframework/core/codec/ByteBufEncoderTests.java
```diff
@@ -0,0 +1,72 @@
...
+
+		// SPR-15464
+		assertThat(this.encoder.canEncode(ResolvableType.NONE, null)).isFalse();
+	}
+
+	@Override
+	@Test
+	public void encode() {
+		Flux<ByteBuf> input = Flux.just(this.fooBytes, this.barBytes)
+				.map(Unpooled::copiedBuffer);
```
rstoyanchev, line 62: We don't automatically wrap. This is perfectly okay as a single line. See https://github.com/spring-projects/spring-framework/wiki/Code-Style#line-wrapping.

spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
```
rstoyanchev, line 2: Please correct this in all files (current year is 2020).

rstoyanchev (changes requested):

spring-core/src/test/java/org/springframework/core/codec/ByteBufEncoderTests.java
encircled: Ok. Does it apply for sequential invocations as well? Like multiple `.map` etc

spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+	public boolean canDecode(ResolvableType elementType, @Nullable MimeType mimeType) {
+		return (ByteBuf.class.isAssignableFrom(elementType.toClass()) &&
+				super.canDecode(elementType, mimeType));
+	}
+
+	@Override
+	public ByteBuf decode(DataBuffer dataBuffer, ResolvableType elementType,
+			@Nullable MimeType mimeType, @Nullable Map<String, Object> hints) {
+
+		int byteCount = dataBuffer.readableByteCount();
```
rstoyanchev, line 54: This variable is no longer needed here.

spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+		// Copies the dataBuffer if needed only
+		ByteBuf copy = NettyDataBufferFactory.toByteBuf(dataBuffer);
+
+		DataBufferUtils.release(dataBuffer);
```
rstoyanchev, line 58: This should not be released. We are no longer making a copy, so releasing it is trouble.

spring-core/src/main/java/org/springframework/core/codec/ByteBufEncoder.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.core.io.buffer.NettyDataBufferFactory;
+import org.springframework.lang.Nullable;
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
+import reactor.core.publisher.Flux;
+
+import java.util.Map;
+
+/**
+ * Encoder for {@link ByteBuf ByteBuffers}.
```
rstoyanchev, line 33: Copy and paste error.

rstoyanchev (changes requested):
What's missing still is the conditional registration of this encoder and decoder in the default codecs of `CodecConfigurer` when Netty `ByteBuf` is present on the classpath.

spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
encircled: Good point, thanks

spring-core/src/main/java/org/springframework/core/codec/ByteBufDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.core.io.buffer.NettyDataBufferFactory;
+import org.springframework.lang.Nullable;
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
+
+import java.nio.ByteBuffer;
+import java.util.Map;
+
+/**
+ * Decoder for {@link ByteBuffer ByteBuffers}.
```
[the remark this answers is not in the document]
encircled: renamed

Issue #24498, opened 2020-02-10
Support Optional in Jackson 2 message converters


daniel-shuy:
**Affects:** 5.2.3.RELEASE

---

Given an endpoint that returns a polymorphic JSON type, eg.

```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME)
@JsonSubTypes({Impl1.class, Impl2.class})
public interface Parent {
}

public class Impl1 implements Parent {
}

public class Impl2 implements Parent {
}

@GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
public Parent get() {
    return new Impl1();
}
```

Jackson will serialize it with the property `@type`, eg.
```json
{
    "@type": "Impl1"
}
```

However, when wrapping it with an `Optional`/[Page](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Page.html), the JSON polymorphism no longer works. Eg.
```java
@GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
public Optional<Parent> getOptional() {
    return Optional.of(new Impl1());
}
```
returns:
```json
{}
```

According to FasterXML/jackson-databind#1594 this is due to type erasure, however the JSON polymorphism still works when wrapping with built-in `Collection`s (eg. `List`/`Set`/`Map`), eg.
```java
@GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
public Collection<Parent> getCollection() {
    return Collections.singleton(new Impl1());
}
```

So it seems like Spring has special handling for built-in `Collection`s. Is it possible to apply them to `Optional`/[Page](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Page.html) as well?

Currently the workaround is to create custom `JsonSerializer`s for `Optional`/[Page](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Page.html), eg.
```java
@JsonComponent
public class OptionalSerializer extends JsonSerializer<Optional<?>> {
    @Override
    public void serialize(Optional<?> value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
        if (value.isPresent()) {
            gen.writeObject(value.get());
        }
    }
}
```

encircled:
Hi.

I can confirm this issue. It works for collections thanks to Jackson support.

My PR fixes "Optional" only, for the Page it should be placed in a different repo I believe, spring-data-rest maybe?

[...]

PR #24867, opened 2020-04-06
Fix json type info when using Optional

encircled:
Preserve json type info during serialization when value is wrapped with java.util.Optional #24498


rstoyanchev:
Thanks for the PR but this is a `GenericHttpMessageConverter` and [typically invoked](https://github.com/spring-projects/spring-framework/blob/8265db0ae1d83a1e204e602fa96e542c2677977f/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/AbstractMessageConverterMethodProcessor.java#L276-L277) through the 4-arg `write` method that also takes a `Type`. So I don't think this is a good place to solve it and we'd have to put similar logic in all converters. 

I think it is better to address this more generally for all supported return types so I'll close this.

Issue #24873, opened 2020-04-07
Avoid using Regex for matching URL paths when possible as doing so causes more of the request time to be spent in regex then doing the API call itself.


LukeButters:
**Affects:** I don't know of a version this does not affect.

Hi, profiling reveals that a bunch of time is spent in in this stack trace:
```
    at java.base@11.0.3/java.util.regex.Matcher.<init>(Matcher.java:248)
    at java.base@11.0.3/java.util.regex.Pattern.matcher(Pattern.java:1133)
    at org.springframework.util.AntPathMatcher$AntPathStringMatcher.matchStrings(AntPathMatcher.java:686)
    at org.springframework.util.AntPathMatcher.matchStrings(AntPathMatcher.java:421)
    at org.springframework.util.AntPathMatcher.doMatch(AntPathMatcher.java:218)
    at org.springframework.util.AntPathMatcher.match(AntPathMatcher.java:177)
    at org.springframework.web.servlet.mvc.condition.PatternsRequestCondition.getMatchingPattern(PatternsRequestCondition.java:258)
    at org.springframework.web.servlet.mvc.condition.PatternsRequestCondition.getMatchingPatterns(PatternsRequestCondition.java:223)
    at org.springframework.web.servlet.mvc.condition.PatternsRequestCondition.getMatchingCondition(PatternsRequestCondition.java:205)
    at org.springframework.web.servlet.mvc.method.RequestMappingInfo.getMatchingCondition(RequestMappingInfo.java:229)
    at org.springframework.web.servlet.mvc.method.RequestMappingInfoHandlerMapping.getMatchingMapping(RequestMappingInfoHandlerMapping.java:93)
    at org.springframework.web.servlet.mvc.method.RequestMappingInfoHandlerMapping.getMatchingMapping(RequestMappingInfoHandlerMapping.java:57)
    at org.springframework.web.servlet.handler.AbstractHandlerMethodMapping.addMatchingMappings(AbstractHandlerMethodMapping.java:428)
    at org.springframework.web.servlet.handler.AbstractHandlerMethodMapping.lookupHandlerMethod(AbstractHandlerMethodMapping.java:394)
    at org.springframework.web.servlet.handler.AbstractHandlerMethodMapping.getHandlerInternal(AbstractHandlerMethodMapping.java:368)
    at org.springframework.web.servlet.handler.AbstractHandlerMethodMapping.getHandlerInternal(AbstractHandlerMethodMapping.java:65)
[...]

Issue #24830, opened 2020-03-31
Revisit "fast enough" performance tests


sbrannen:
## Overview

The Spring Framework test suite contains numerous tests that are assigned to the PERFORMANCE CI build, and many of these tests have names along the lines of "ensure that XYZ is fast enough".

When these tests were originally written, they were feasible when executed on an average developer's workstation; however, over time these tests have proven to be _flaky_ when executed on the CI server with various load levels for the various CI agents.

We are opening this issue in order to reassess whether such "performance" tests make sense in the current Spring Framework build.

## Candidates

- [ ] `org.springframework.aop.framework.IntroductionBenchmarkTests`

## Deliverables

- [x] Introduce a JMH benchmark infrastructure in the build
- [ ] Search for test methods and test classes that are annotated with `@EnabledForTestGroups(PERFORMANCE)` (or `@EnabledForTestGroups(TestGroup.PERFORMANCE)`) or that contain "FastEnough" in their titles and determine if the tests should be deleted or modified to make them more robust.
- [ ] Delete or modify each such test method or test class.

bclozel:
Thanks for creating this issue. We’ll be moving to concourse CI for this milestone and we won’t be having a « performance » build in that new pipeline.

We could however consider transforming some performances tests into proper JMH benchmarks in a dedicated module.

rstoyanchev:
+1 for jmh benchmakrs. Wouldn't it make sense to have these in the respective modules? Looking at [this plugin](https://github.com/melix/jmh-gradle-plugin#configuration) it could be `src/jmh/java` under spring-context or spring-web, etc.

bclozel:
I've introduced the basic JMH infrastructure in 67547e6 as well as a [new Wiki page on Micro-Benchmarks](https://github.com/spring-projects/spring-framework/wiki/Micro-Benchmarks) (feel free to edit and chime in!).

As an example, I've used this new infrastructure in #24769.

sbrannen:
I've added a **Candidates** section to this issue's description as a way to track classes we should be looking at. The list is currently a _work in progress_.

PR #24887, opened 2020-04-09
Avoid using regex matching for static patterns

encircled:
This fixes the issue #24873

As of now, `AntPathStringMatcher` uses regex matching even when:
- the pattern is a constant. In such cases, it should be enough to use strings comparison
- the pattern is `.*`. In such cases it is always a positive match

Applying such optimizations increased `PatternsRequestCondition` evaluation (on my machine) from ~440 ops/ms to ~700 ops/ms

`AntPathStringMatcher` is used not only for URL matching, so it would be nice to double-check carefully

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java
+++ b/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java
@@ -68,6 +68,7 @@ import org.springframework.lang.Nullable;
  * @author Arjen Poutsma
  * @author Rossen Stoyanchev
  * @author Sam Brannen
+ * @author Vladislav Kisel
  * @since 16.07.2003
  */
 public class AntPathMatcher implements PathMatcher {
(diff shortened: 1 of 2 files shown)


bclozel:
I've run these changes and variants as JMH benchmarks and I'm not seeing significant improvements (a bit better or worse than the current implementation depending on the seed data, but always within the error margin).

I guess the JVM is already optimizing those bits for us; the performance test strategy used in this PR is flawed. I'd suggest to look at JMH for a proper benchmark harness. I'm closing this PR for now. Thanks!

encircled:
I don’t think JVM is able to do such optimization. Can you provide your test?

encircled:
Alright, the same test using JMH:
```
    @Benchmark
    public void test() {
        matcher.match("/something/things/foos/{foo}/bar/{id}", "/something/things/foos/test/bar/qwerty");
    }
```

Result is `767.614 ns/op` vs `344.477 ns/op`. Is not it significant @bclozel ? ;) I know this is just one particular type of URL pattern tho.

bclozel:
Alright this is a very targeted improvement but an improvement still.
It's a small change but in a core part of our infrastructure that tends to lead to unintended behavior change quite easily. So I'll schedule that for our next 5.3 because it's a bit late to introduce that in a bugfix release.

Thanks!

encircled:
Great, thanks!

Shall I keep the new perf test and add `@EnabledForTestGroups(PERFORMANCE)`?

bclozel:
You can remove the performance test, as we're about to remove them anyway in #24830. I'll ping  you if I need some input for contributing JMH tests but I think we should be good.

encircled:
I've fixed one more issue, should be final now

PR #24897, opened 2020-04-11
Fixed typos

brianhartung:
Fixed minor typos in the documentation.
Note: if it really was intended to set country to age (?) then my mistake.


brianhartung:
The more I look at this, the more I think it was an intentional choice (though that wasn't clear) so I'll close this PR and submit a new one.

PR #24899, opened 2020-04-13
Closes #24894, don't close input stream after json parsing

### LINKED ISSUE #24894: Support for annotating multiple arguments with @RequestBody
I am using `spring-webmvc` 5.2.3.

I want to use Spring's Default MethodArgumentResolver to resolve the json payload into multiple beans. But I am not able to do that right now because `AbstractMessageConverterMethodArgumentResolver.EmptyBodyCheckingHttpInputMessage(httpRequest)`is passed the HttpRequest object and is used directly. While attempting to resolve the second argument the HttpRequest's inputStream is closed, so I get an exception saying that the input stream is closed. Ergo I am able to annotate only one argument with `@RequestBody`.

```java
    @PostMapping("/endpoint")
    public void  method(@Valid @RequestBody CustomBean bean1, 
                                @Valid @RequestBody CustomBean2 bean2) 
    {
        // using both beans.
    }
```

Why not use a `ContentCachingRequestWrapper` instead of the actual HttpRequest for argument resolution ?

juzerali:


juzerali:
Closing as the feature request was closed.

PR #24905, opened 2020-04-14
Remove unnecessary sorting in PostProcessorRegistrationDelegate

chenqimiao:
I think the sorting is unnecessary, because `internalPostProcessors ` has been sorted in every previous stage of traversing `ppNames`.
Secondly, this sorting will break some rules, For example, it will cause nonOrderedPostProcessors to sort by `@Order`. But in JavaDoc of BeanPostProcessor ,this is not allowed.

The Ordering section of the class-level Javadoc for BeanPostProcessor states the following.
> > `BeanPostProcessor` beans that are autodetected in an `ApplicationContext` will be ordered according to `PriorityOrdered` and `Ordered` semantics. In contrast, `BeanPostProcessor` beans that are registered programmatically with a `BeanFactory` will be applied in the order of registration; any ordering semantics expressed through implementing the `PriorityOrdered` or `Ordered` interface will be ignored for programmatically registered post-processors. Furthermore, the `@Order` annotation is not taken into account for `BeanPostProcessor` beans.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/PostProcessorRegistrationDelegate.java
+++ b/spring-context/src/main/java/org/springframework/context/support/PostProcessorRegistrationDelegate.java
@@ -45,6 +45,7 @@ import org.springframework.lang.Nullable;
  *
  * @author Juergen Hoeller
  * @author Sam Brannen
+ * @author Qimiao Chen
  * @since 4.0
  */
 final class PostProcessorRegistrationDelegate {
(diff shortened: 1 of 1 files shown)


snicoll:
@chenqimiao thanks for the PR but we'd like to keep this as is. Please also review [this note](https://github.com/spring-projects/spring-framework/blob/2c0a9e0e30d5b18734077abf37e34d4eb1e49d4d/spring-context/src/main/java/org/springframework/context/support/PostProcessorRegistrationDelegate.java#L68-L74) on the class you've modified.

PR #24908, opened 2020-04-15
Improve Javadoc

chenqimiao:
The `getBeanNamesOfType` method does not seem to exist, this looks like it should refer to the method called `getBeanNamesForType`.


snicoll:
@chenqimiao good catch, thank you!

Issue #18551, opened 2016-02-24
spring-test HtmlUnit does not recognize files for multipart uploading [SPR-13979]


spring-issuemaster:
**[Stephen Liang](https://jira.spring.io/secure/ViewProfile.jspa?name=stephenliang)** opened **[SPR-13979](https://jira.spring.io/browse/SPR-13979?redirect=false)** and commented

#### Overview

The params method in `HtmlUnitRequestBuilder` does not check if a parameter contains a File payload, which results in a File input box to get its value (the path to the file) and submit it as a key/value rather than the payload being uploaded as multipart form data.

#### For Reference

https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java#L361

`webRequest.getRequestParameters()` is a `KeyDataPair` which contains a field called `fileObject_`. This field should be recognized and used in place of the `value_` field which is the name of the field, rather than the content of the file (when the parameter is has a file).

#### To Reproduce

1. Create a simple file upload form, e.g.:

```html
<html><head></head><body>
<form enctype='multipart/form-data' action='"/upload"' method='POST'>
<input type='file' name='image' />
<input type='submit' id='clickMe'>
</form>
</body>
</html>
```

1. Create a html unit test that sends the file path to the "file" input type
2. Click on the submit button

**Expected Result**: The file contents are read and sent as form data
**Actual Result**: The file's name is sent as a key/value pair

---

**Affects:** 4.2 GA

0 votes, 5 watchers

djkirby:
Any update on this? Would be nice to have this functionality 🤞

sbrannen:
I've added the `ideal-for-contribution` label so that someone from the community can pick this one up.

govi20:
Would like to take it up.

sbrannen:
@govi20, great!

Feel free to submit a PR.

If for some reason you cannot get to it, please post back here so that we know you're not actively working on it.

govi20:
@sbrannen Apologies I missed this message. I will look into this over the weekend. One newbie question: is there a setting file that could be used to mark IntelliJ source and test root directories or it needs to be marked manually?

kobaeugenea:
Hi. Created PR for this issue https://github.com/spring-projects/spring-framework/pull/24926

sbrannen:
Superseded by #24926

PR #24926, opened 2020-04-18
Add file parameters as request parts in HtmlUnitRequestBuilder

kobaeugenea:
https://github.com/spring-projects/spring-framework/issues/18551
Solved this issue. If param is an instance of KeyDataPair so add it as part, not as parameter.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
@@ -16,11 +16,14 @@
 
 package org.springframework.test.web.servlet.htmlunit;
 
+import java.io.File;
+import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.nio.charset.Charset;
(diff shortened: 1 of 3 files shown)


sbrannen:
Thanks for the PR!

We'll review it with the intent to include it in 5.3.

In the interim, @rstoyanchev, in https://github.com/spring-projects/spring-framework/issues/18551#issuecomment-453436578 you mentioned that the implementation would need to switch from a `MockHttpServletRequest` to a `MockMultipartHttpServletRequest`; whereas, this PR works directly with the existing `Part` support in the `MockHttpServletRequest`. Do you still feel that the HtmlUnit support would need the `MultipartHttpServletRequest` API?

rstoyanchev:
Creating `MockMultipartHttpServletRequest` would be closer to simulating an application with a `MultipartPartResolver` configured. However, for Servlet 3 multipart requests that's not strictly required, and for example `MultipartResolutionDelegate` will wrap the request with `StandardMultipartHttpServletRequest` if necessary.

dadmin-admin:
So can we leave this way of implementation or do I need to re-implement it using MockMultipartHttpServletRequest?

kobaeugenea:
So can we leave this way of implementation or should I re-implement to use MockMultipartHttpServletRequest?

PR #24936, opened 2020-04-20
Simplify if-else constructions in spring-web

mikaelparsekyan:
Removed redundant else-s in `ReadOnlyHttpHeaders.java` and in `ResponseEntity.java`


sbrannen:
Thanks for the PR.

For these particular scenarios, the inclusion of the `else` block is technically unnecessary; however, the original author introduced them for a reason, typically based on readability.

Since we do not have a project policy for "unnecessary else blocks", we tend to leave them in place.

PR #24962, opened 2020-04-23
Make imported config always override the one found by package scan

### LINKED ISSUE #24643: Name of scanned @Configuration class affects @Import to work or fail
Our projects rely heavily on bean overriding (yes, we know, disabled now by default for good reasons, but still useful at times and a different story). 

We now traced down a phenomenon that the name of a `@Configuration` class found by the `@ComponentScan` seem to affect, whether overriding with `@Import` (or sub-classing) works properly. This is though the `ConfigurationClassParser` already does a lot of handling on this case, e. g. in `processConfigurationClass`.

Please find a test case in https://github.com/abenneke/sandbox/tree/master/spring-import-order:
It contains four `@Configuration` classes: Two for a "foo" bean (`FooConfigurationA+B`) and two for a "bar" bean (`BarConfigurationA+B`). However, the `@Import` order is different between both configurations: `FooConfigurationB` is importing `FooConfigurationA` while `BarConfigurationA` is importing `BarConfigurationB`.

In the test we now see that overriding works as expected for "foo", but fails for "bar"!?

Just to make sure we did not miss anything, we added the very same test using standard Spring methods and again using Spring Boot - but it does not make any difference here.

encircled:
Fixes #24643

**Current state**

Whether imported user config class will override a scanned one depends on its names.
Unit test which is reproducing it: 
`ImportTests#importedConfigOverridesScannedOne`

During nested configs processing (`ConfigurationClassParser#processMemberClasses`), all nested configs are being overridden by its scanned version loosing `importedBy` information. This is not intended behavior according to the documentation of import feature. But this accidentally made it work, because otherwise the config would have been registered twice: first by package scan, then by `ConfigurationClassBeanDefinitionReader#loadBeanDefinitionsForConfigurationClass`. 

Unit test which is reproducing it: `ComponentScanAnnotationIntegrationTests#viaContextRegistration_WithComposedAnnotation`

**Fixed version**

Change in `ConfigurationClassParser` is needed to make imported configs always override the scanned ones, preserving `importedBy` info. Due to preserving `importedBy` I had to change `ConfigurationClassBeanDefinitionReader` as well, which registers imported configs, but it did not expect that it could have been already registered (during package scanning). I used `AnnotationBeanNameGenerator` for naming to make it 100% backward compatible by preserving configs bean names.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassBeanDefinitionReader.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassBeanDefinitionReader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -68,6 +68,7 @@ import org.springframework.util.StringUtils;
  * @author Juergen Hoeller
  * @author Phillip Webb
  * @author Sam Brannen
+ * @author Vladislav Kisel
  * @since 3.0
  * @see ConfigurationClassParser
  */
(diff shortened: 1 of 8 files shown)


encircled:
@jhoeller just a kind reminder. Lets review&merge or close it. This bug should be addressed somehow thought.

jhoeller:
I have a different implementation strategy in mind as part of a slightly wider revision for 6.2 M1, so I'll close the PR at this point, keeping #24643 open instead. Thanks for the PR in any case! I'm going to make sure to cover the test cases sketched in the PR.

PR #24967, opened 2020-04-24
Add KClass variant to Coroutines web API 

imanushin:
Extend API - add ability to work with Kotlin Coroutines and generic deserialization.

### DIFF
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/ClientResponseExtensions.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/ClientResponseExtensions.kt
@@ -24,6 +24,7 @@ import org.springframework.core.ParameterizedTypeReference
 import org.springframework.http.ResponseEntity
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
+import kotlin.reflect.KClass
 
 /**
  * Extension for [ClientResponse.bodyToMono] providing a `bodyToMono<Foo>()` variant
@@ -57,6 +58,16 @@ inline fun <reified T : Any> ClientResponse.bodyToFlow(): Flow<T> =
 		bodyToFlux<T>().asFlow()
 
 /**
+ * Generic Coroutines [kotlinx.coroutines.flow.Flow] based variant of [ClientResponse.bodyToFlux].
+ * Please consider inline variant `bodyToFlow<Foo>` if possible
+ *
+ * @author Igor Manushin
+ * @since 5.2
+ */
+fun <T : Any> ClientResponse.bodyToFlow(clazz: KClass<T>): Flow<T> =
+		bodyToFlux(clazz.java).asFlow()
+
+/**
  * Extension for [ClientResponse.toEntity] providing a `toEntity<Foo>()` variant
  * leveraging Kotlin reified type parameters. This extension is not subject to type
  * erasure and retains actual generic type arguments.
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@imanushin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24967)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24967) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@imanushin Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=24967)!

imanushin:
@sdeleuze, could you please take a look?

imanushin:
@sdeleuze, could you please take a look?

sdeleuze:
@imanushin Thanks for this PR, could you please give me more detail about your use case? In particular, is it a matter of coding style or are you blocked by the reified type parameter variant? And why does it take a `Class` and not a `KClass` parameter ?

imanushin:
@sdeleuze,

> are you blocked by the reified type parameter variant?

Yes. I have generic service, which has ```KClass``` on input. It wraps common input logic, when you receive hook-like requests from the different services. In this case I need generic-like deserialization. As workaround, I wrote code ```bodyToMono(clazz).awaitSingle()```.

I think, there are people who have the same task. So in additional I extended other methods: for nullable return and for ```Flow```.

> And why does it take a Class and not a KClass parameter ?

Just because ```Class``` is standard way in Java to pass the instance type. However I can replace it to ```KClass``` (moreover, ```KClass``` I have in my code on input). What do you think, will ```KClass``` be better?

sdeleuze:
Ok that was I expected. We usually don't provide `Class` or `KClass` variant for reified type parameter extensions because the `Class` based Java variant is enough, but here indeed I can understand how it is annoying to be forced to call the Reactive API then the Reactive to Coroutines conversion API.

I think using `KClass` here is more idiomatic, so yes please modify this PR to use it instead of `Class`. Also for the sake of consistency, could you please add similar extensions to `ClientResponseExtensions` ?

I am also curious if these additional extensions change IntelliJ IDEA autocomplete experience or not, could you please test that with and without the additional extensions? I would like to make sure we do not impact too much 99% of the use cases with this 1% use cases improvement.

imanushin:
> I think using KClass here is more idiomatic, so yes please modify this PR to use it instead of Class.

Yes, of course.

> Also for the sake of consistency, could you please add similar extensions to ClientResponseExtensions ?

Yes, will do.

> I am also curious if these additional extensions change IntelliJ IDEA autocomplete experience or not, could you please test that with and without the additional extensions? I would like to make sure we do not impact too much 99% of the use cases with this 1% use cases improvement.

Yes... Probably, I didn't catch this fully. Am I right, that you'd like to test, that ```inline``` methods will be at the IntelliSence (and at the first places, if possible)? If yes, I'll attach the screenshots

sdeleuze:
I will check myself  the developer experience (last point), please ping me when you have updated the PR with the 2 other points.

imanushin:
@sdeleuze , done. I also extended API, which returns Kotlin classes.

imanushin:
@sdeleuze , could you please take look on PR?

imanushin:
@jhoeller, could you please take a look on PR?

imanushin:
@sdeleuze, @sdeleuze, all changes were applied.

May we merge the improvement?

imanushin:
@sdeleuze, @sdeleuze, could you please take a look on the PR?

sdeleuze:
Merged thanks for your patience.

Issue #24844, opened 2020-04-02
Add support for relative ordering


snicoll:
Our current support for `@Order` is using an absolute number with a default value of "lowest precedence" (i.e. `Integer.MAX_VALUE`). We also have `@PriorityOrdered` to override such ordering in certain scenarios.

One problem with the current default value is that it is not natural for customizers where you'd want the "default" ones to be at a given order with the ability of custom instance to run before, or after. Such default must provide a value as there is no way to run "after" with the current default value. A natural value for that is zero with negative values being processed before and positive values being processed after.

A better way to express ordering is to express it relative to another concept. You may need to run/customize/execute after or before another "instance".  In Spring Boot, we've essentially implemented that with `@AutoConfigureBefore`, `@AutoConfigureAfter` to order auto-configurations. 

This issue is to discuss if there is an appetite to provide such constructs in the core framework, with an SPI that would let us order things (potentially being honored on `ObjectProvider#orderedStream()`.

andersonkyle:
This would be particularly nice for servlet filter registration.

mjustin:
As a user, I can confirm this would be useful.  The current approach I have to follow is open the bean, see if there's an order defined, and then order relative to that (hoping that it's not at the max/min in a way that would prevent going before/after it).  When there isn't an order defined, I need to pull open `@Order.value()` to see what the default is (`LOWEST_PRECEDENCE`) and make sure I order relative to that.

Then I write something like the following, where the reasoning is in the comments, rather than being declarative in the code itself:

```java
@Order(Ordered.LOWEST_PRECEDENCE - 1) // Higher precedence than SomeOtherThing
```

rstoyanchev:
**Team Decision:** this is worth doing to provide sufficient control over ordering. There are some design decisions to be thought through.

sbrannen:
This is also related to:

- #34670

YongGoose:
Hello @sbrannen @snicoll 

I've been following this issue with interest and have put some thought into a potential design that I'd like to share. 
I'm also very keen to contribute an implementation for this; I'm confident I have the skills and motivation to see it through 🚀

My proposal is to introduce new annotations `@OrderBefore` and `@OrderAfter`.  (The annotation names may be changed later if more suitable names are found. Therefore, I would appreciate it if you could focus on their roles)

### Quick Example
```java
// 1. The component that runs first
@Component
public class DatabaseInitializer {  }

// 3. The component that runs last
@Component
[...]

PR #24977, opened 2020-04-26
Aggregate TestExecutionListener orders

andreisilviudragnea:
Aggregated `TestExecutionListener` orders into `SpringTestExecutionListenerOrder`, in order to help third-party integrators when choosing the order for a new listener.

I have done two similar PRs in [spring-boot](https://github.com/spring-projects/spring-boot/pull/21133) and [spring-security](https://github.com/spring-projects/spring-security/pull/8436) projects.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/SpringTestExecutionListenerOrder.java
+++ b/spring-test/src/main/java/org/springframework/test/context/SpringTestExecutionListenerOrder.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


snicoll:
See also #24844

cunla (approved):

andreisilviudragnea:
I have thought more about the changes in this PR. The introduced enum would couple all the listeners together and would hinder a possible future separation of the functionality of `spring-test` module in more modules, if this is possible now.

sbrannen:
Thanks for the PR.

I certainly understand the desire for third-party integrators to be able to _look up_ the order for a given `TestExecutionListener`, but I don't think we should provide such a mechanism via a centralized enum. In addition, there is no requirement that a given `TestExecutionListener` provide an explicit, static order. Listeners do not have to provide any order value at all, and they may choose to somehow _calculate_ the order programmatically instead of returning a static value.

Excerpt from https://github.com/spring-projects/spring-boot/pull/21133#issuecomment-619571205:

> Maybe I overthought this and only extracting each order to a constant on each listener would have been enough.

I also considered that at one point to make the order value statically available via the class reference, but for the above reasons I did not opt to implement that.

An enum or local static constants would only partially help to achieve the ultimate goal of this PR. Specifically, the goal is for `TestExecutionListener` implemetors to be able to have their listeners ordered before or after other listeners (some from core Spring Framework, Spring Boot, and Spring Security, and some from other third-party frameworks). Thus, the only robust way to address this need is via some form of relative ordering as proposed by @snicoll in #24844.

Thus, after thoughtful consideration, and given that @wilkinsona closed the related PR for Spring Boot (https://github.com/spring-projects/spring-boot/pull/21133#issuecomment-760861928), I am closing this PR.

PR #24988, opened 2020-04-28
Polishing

ijliym:


sbrannen:
Thanks for the PR.

The code in question was intentionally structured like that for the sake of readability.

Thus, we will retain that structure, and I am closing this PR.

Issue #23096, opened 2019-06-07
WebFlux Tomcat, Intermittent failure to create response content 


danielra:
**Affects:** \5.1.7.RELEASE

---

I am running a spring-boot application which is using WebFlux with Tomcat as the underlying http server. In my real application with fairly high request rate, I noticed in metrics what appeared to be a slow leak in tomcat connections. I have since added a WebFilter to implement a response timeout, and this confirmed that a low volume of responses were never being completed. With the WebFilter in place, the observed connection leak is fixed - but I would like to resolve the response timeouts.

I have created a relatively simple isolated demo project which reproduces the issue here: https://github.com/danielra/webflux_response_timeout_repro

The project can be built via `./gradlew clean build` and then run via `java -jar ./build/libs/demo-0.0.1-SNAPSHOT.jar`. To reproduce the issue with the demo project I have then been using `wrk` ( https://github.com/wg/wrk ) to throw load against the running application.

For example:
```
$ ./wrk -c 20 -t 16 -d 300 http://localhost:8080/get --latency
Running 5m test @ http://localhost:8080/get
  16 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.04ms    5.38ms 335.15ms   98.55%
    Req/Sec     1.62k   280.92     4.75k    79.07%
  Latency Distribution
     50%  508.00us
     75%  762.00us
     90%    2.00ms
     99%    7.48ms
  7719452 requests in 5.00m, 28.98GB read
  Socket errors: connect 0, read 7, write 0, timeout 1
Requests/sec:  25723.26
Transfer/sec:     98.89MB
```
Note that for this 5 minutes of load, 1 timeout was observed. The corresponding application console output was as follows:
```
$ java -jar ./build/libs/demo-0.0.1-SNAPSHOT.jar

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::        (v2.1.5.RELEASE)

[...]

PR #24989, opened 2020-04-28
Fix race condition in WriteResultPublisher between subscription and completion in publishComplete

danielra:
The bug I was experiencing was that sometimes with a WebFlux on Tomcat setup, responses were not being emitted (despite the Publisher being returned from my Controller methods completing successfully). Similar to https://github.com/spring-projects/spring-framework/issues/23096


rstoyanchev:
Thanks for spotting this!

Issue #24964, opened 2020-04-23
Include response body in RestTemplate exception when there is no suitable HttpMessageConverter


devinabyss:
When RestTemplate receive Response from target server, but there's no body readerble converter, throws RestClientException with only situation describing message. 

In this situation, even though communication with the server is normally completed, the content of the received text is lost only because the object conversion of the specified type is impossible.

The results seem unfriendly to developers.

It is extremely natural that Exception is thrown because object conversion is impossible, but preserving the response value so that the developer can confirm that the object was not converted to the intended object because the value was returned in the response is helpful in the development or production stage. I think it need. (Especially production)

sually, if the values of the object conversion failure case are unexpectedly returned in the body, the status code may not 2xx, so an HttpServerErrorResponse or HttpClientErrorException will be thrown, but there is no 100% guarantee that the other server will follow the general status code rules. It may or may not be intentional.

Also, may have misplaced the conversion specification simply by the carelessness of the developer writing the communication code.

Anyway, the important thing is that it's a good idea to provide more effective information for trace exception by providing the full value of the response body returned to the developer.

I searched similar issue on github and stackoverflow, but could not find.

So I modified codes of `HttpMessageConverterExtractor.class` near line 120 as below, and I'd like to hear your opinion.

I would appreciate it if you would like to keep the body values or if there is a better way.
thanks.

## Current

```java
		...
		catch (IOException | HttpMessageNotReadableException ex) {
			throw new RestClientException("Error while extracting response for type [" +
					this.responseType + "] and content type [" + contentType + "]", ex);
		}

		throw new RestClientException("Could not extract response: no suitable HttpMessageConverter found " +
				"for response type [" + this.responseType + "] and content type [" + contentType + "]");
	}
	...
```

## Suggest
```java
		...
		} 
		catch (IOException | HttpMessageNotReadableException ex) {
			throw new RestClientException("Error while extracting response for type [" +
					this.responseType + "] and content type [" + contentType + "]", ex);
		}

		String extractErrorMessage = "Could not extract response: no suitable HttpMessageConverter found " +
				"for response type [" + this.responseType + "] and content type [" + contentType + "]";

		try (InputStream is = responseWrapper.getBody()) {
			byte[] bytes = StreamUtils.copyToByteArray(is);
			throw new RestClientResponseException(extractErrorMessage, response.getStatusCode().value(),
					response.getStatusText(), response.getHeaders(), bytes, StandardCharsets.UTF_8);
[...]

PR #24994, opened 2020-04-29
Preserve Http Response Info When MessageConverter Not Found Case

devinabyss:
And Introduce Detail Purpose When Response Could Not Extract

https://github.com/spring-projects/spring-framework/issues/24964

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/HttpMessageConverterExtractor.java
+++ b/spring-web/src/main/java/org/springframework/web/client/HttpMessageConverterExtractor.java
@@ -17,7 +17,9 @@
 package org.springframework.web.client;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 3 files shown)


rstoyanchev (changes requested):
@devinabyss I'm not sure we need to expand with 3 new exceptions. Just one `RestClientResponseException` to replace the "no suitable HttpMessageConverter" case.

devinabyss:
> @devinabyss I'm not sure we need to expand with 3 new exceptions. Just one `RestClientResponseException` to replace the "no suitable HttpMessageConverter" case.

sorry about no more discuss about that on issue thread.

`RestClientResponseException.FailedToReadResponseBody` arises when converter body extracting failed. you first mentioned case on issue thread.  

After modifying the `RestClientResponseException.MessageConverterNotFound`, I thought that preserving the rest of the information (headers, specific status like 200, 201, 204..) is intentionally right, although in the sense of preserving the information of response, of course the body is impossible.

Can I just keep RestClientResponseException.MessageConverterNotFound?

rstoyanchev:
> Can I just keep RestClientResponseException.MessageConverterNotFound?

I don't see the value. It's not something that can be handled and it's in the root cause.

rstoyanchev:
I've resolved this by adding an `UnknownContentTypeException` sub-class. Thanks for the PR in any case.

devinabyss:
thanks.
sorry for my reaction was late.

PR #25002, opened 2020-05-01
Supplement Test Cases In XmlBeanFactoryTests : Resolve Circle Reference When Use Aop

lixiaolong11000:
Many people implementate `BeanPostProcessor` to realize Aop, but it has a problem with cicle reference. Implementation `SmartInstantiationAwareBeanPostProcessor.getEarlyBeanReference` is a good practice, for example `AbstractAutoProxyCreator`.
I think the case could make learners understand better.


lixiaolong11000:
hello,please take a look.

snicoll:
Thanks for the PR but cases like this are better served using `ObjectProvider` or `@Lazy` on the target component to indicate the lifecycle explicitly.

PR #25016, opened 2020-05-05
Support for pattern based origin cors configuration

### LINKED ISSUE #24763: CorsConfiguration based on pattern matching
**Affects:** Spring Web 5.2.4.RELEASE

---
Initial issue opened in Spring Cloud Gateway: https://github.com/spring-cloud/spring-cloud-gateway/issues/1607

Application APIs can be called by multiple origins. For applications that represent some kind of platform within a company it often makes sense to accept all request from orings coming from company domain, basically accept all request with origin matching a pattern **\*.company.com**. Currently default spring `CorsConfiguration ` supports only list of origins, I think this might not be flexible enough, as it will require configuration change every time new application within a company wants to call that REST api (especially in case of an API Gateway).

**Proposal**: add support for CORS configuration based on a pattern matching.
One option would'be, I guess, to extend existing `CorsConfiguration`  class with additional fields `allowedOriginPatterns` and modify `checkOrigin` to use this property.
Or have separate `CorsConfiguration` implementation that is using pattern matching.

While current solution is extendible and this approach can be implemented within the application itself, I think this should be a part of framework because it's a widely used functionality and is supported by other frameworks that are used for API implementation.

For example vert.x CorsHandler suipports that: https://github.com/vert-x3/vertx-web/blob/5cd7ecaa6bead1b246b5327537ee9b82c22187bc/vertx-web/src/main/java/io/vertx/ext/web/handler/CorsHandler.java#L42
Also, [Kong](https://docs.konghq.com/hub/kong-inc/cors/) supports pattern based origins as well.

korektur:
Fix https://github.com/spring-projects/spring-framework/issues/24763

I have modified `CorsConfiguration` to support pattern based origins.

Obviously that can be done in a different ways, like using somehing else instead of standard java `Pattern` or creating a new delegate`PatternCorsConfiguration` instead of modifying existing one, so please let me know if the approach I've chosen is not the most suitable.


korektur:
@sdeleuze @rstoyanchev can you review please?

rstoyanchev:
@korektur, on a first pass this looks okay to me and I can take it from here. 

One thing I'd like to refine is the pattern syntax to be just wildcards, as in your original https://github.com/spring-projects/spring-framework/issues/24763#issue-585831901 `"*.company.com"` rather than a full regex expression `".*\\.company\\.com"`. Internally it can still be converted to a regex so this is more about keeping the syntax simpler and more readable. Does that work for your cases?

Note also that typically it is better to rebase your PR rather than doing merges. I have corrected that locally, so please don't make any further changes.

korektur:
@rstoyanchev yes, that makes sense.

michael-o:
Can someone for the stupid explain why this violates CORS specs with "*" and `allowCredentials: true`? From an external point of view it does not matter how you match internally, this should just work like with previous fixed strings `allowedOrigins`.

FTR: https://stackoverflow.com/a/19744754/696632

PR #25021, opened 2020-05-07
Fix @Transactional syntax error in reference manual

quaff:


sbrannen:
Good catch!

Thanks

PR #25024, opened 2020-05-07
Drop unnecessary intermediate StringBuilder.toString() call in ServerSentEventHttpMessageReader

stsypanov:
Apart from simplification this transformation is likely to bring faster code which consumes less memory.


jhoeller:
@rstoyanchev This looks entirely valid, worth applying to 5.2.7 and also a 5.1.16 backport candidate from my perspective. Maybe an opportunity to revisit related code for efficient `StringBuilder` usage?

jhoeller:
I've taken this a bit further by enforcing `StringBuilder` exposure throughout the encode/decode methods in both `ServerSentEventHttpMessageReader` and `ServerSentEventHttpMessageWriter`, in particular also for the event content, not just for the comment part. Thanks for raising this!

PR #25027, opened 2020-05-07
Skip null values for primitive target properties in BeanUtils.copyProperties

importsource:
when we copy some property ,for example : 

```java
private Integer status;//source 

private int status;//target
```

if the source property is null, then throw new FatalBeanException(
									"Could not copy property '" + targetPd.getName() + "' from source to target", ex);

to fix it , maybe need to skip the null value

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -727,6 +727,7 @@ public abstract class BeanUtils {
 									readMethod.setAccessible(true);
 								}
 								Object value = readMethod.invoke(source);
+								if(value==null)continue;
 								if (!Modifier.isPublic(writeMethod.getDeclaringClass().getModifiers())) {
 									writeMethod.setAccessible(true);
 								}


importsource:
in my case, when i want to transfer one entity to thrift entity,then throw this exception

sbrannen:
Actually, a `null` value is permissible if the target property is not a primitive. So, I think it would be inappropriate to simply skip all `null` values. Furthermore, doing so would be a breaking change.

However, for the use case you supplied, the source is an `Integer` (which may be `null`); whereas, the target is an `int` which can never be `null`. Thus, for such use cases, I suppose we have two options:

1. Leave it as-is, throwing a `FatalBeanException` to indicate the failure.
2. Skip `null` values only for primitive target types, potentially logging a `DEBUG` statement.

importsource:
> Actually, a `null` value is permissible if the target property is not a primitive. So, I think it would be inappropriate to simply skip all `null` values. Furthermore, doing so would be a breaking change.
> 
> However, for the use case you supplied, the source is an `Integer` (which may be `null`); whereas, the target is an `int` which can never be `null`. Thus, for such use cases, I suppose we have two options:
> 
> 1. Leave it as-is, throwing a `FatalBeanException` to indicate the failure.
> 2. Skip `null` values only for primitive target types, potentially logging a `DEBUG` statement.
> 
> @jhoeller, what are your thoughts on the matter?



> Actually, a `null` value is permissible if the target property is not a primitive. So, I think it would be inappropriate to simply skip all `null` values. Furthermore, doing so would be a breaking change.
> 
> However, for the use case you supplied, the source is an `Integer` (which may be `null`); whereas, the target is an `int` which can never be `null`. Thus, for such use cases, I suppose we have two options:
> 
> 1. Leave it as-is, throwing a `FatalBeanException` to indicate the failure.
> 2. Skip `null` values only for primitive target types, potentially logging a `DEBUG` statement.
> 
> @jhoeller, what are your thoughts on the matter?

For my case, now I have to copy the BeanUtils file and then override in my local project. So ,I  suppose to skip the `null` value for the primitive target types.

PR #25038, opened 2020-05-09
Use Map::computeIfAbsent to simplify scope implementations

quaff:
It's a tiny improvement.


sbrannen:

Thanks

PR #25049, opened 2020-05-11
Replace HttpClient.websocket() deprecation

izeye:
This PR replaces deprecated `HttpClient.websocket()` with recommended one.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java
@@ -17,21 +17,23 @@
 package org.springframework.web.reactive.socket.client;
 
 import java.net.URI;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Thanks for the change. This is a very recent change in Reactor Netty, so I'll schedule it for 5.3. 

If you want to take the change further, we'll want to do something similar as [on the server side](https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/socket/server/upgrade/ReactorNettyRequestUpgradeStrategy.java) where we added a `Supplier<WebsocketServerSpec.Builder>` and deprecated individual properties in favor of using that.

izeye:
@rstoyanchev Thanks for the feedback!

Actually, I'm not familiar with the code, but I just gave it a try simply by mirroring the code you suggested. Let me know if anything is missing or wrong.

rstoyanchev:
I ended up applying some of the changes as part of the Reactor 2020.0.0 upgrade since some methods in Reactor Netty were removed. I applied your commit on top, still bringing in the deprecations.

PR #25054, opened 2020-05-12
Use computeIfAbsent in ResourceBundleMessageSource

chenqimiao:
This PR is to optimize code by using `computeIfAbsent` instead of `putIfAbsent`, in addition, the content of some Javadoc has been adjusted.

PR #25072, opened 2020-05-14
Use computeIfAbsent in ResourceBundleMessageSource

chenqimiao:
This PR is to optimize code by using computeIfAbsent instead of putIfAbsent, similar to #25054


sbrannen:

Thanks

PR #25080, opened 2020-05-14
Avoid unnecessary boxing/unboxing of primitives

stsypanov:
Wrapping primitives is pointless for the cases primitive can be used (e.g. in concatenation of arithmetics).

### DIFF
--- a/spring-core/src/main/java/org/springframework/cglib/proxy/Enhancer.java
+++ b/spring-core/src/main/java/org/springframework/cglib/proxy/Enhancer.java
@@ -1254,7 +1254,7 @@ public class Enhancer extends AbstractClassGenerator {
 			}
 
 			public int getOriginalModifiers(MethodInfo method) {
-				return (Integer) originalModifiers.get(method);
+				return ((Integer) originalModifiers.get(method)).intValue();
 			}
 
 			public int getIndex(MethodInfo method) {
--- a/spring-core/src/main/java/org/springframework/cglib/proxy/Enhancer.java
+++ b/spring-core/src/main/java/org/springframework/cglib/proxy/Enhancer.java
@@ -1254,11 +1254,11 @@ public class Enhancer extends AbstractClassGenerator {
 			}
 
 			public int getOriginalModifiers(MethodInfo method) {
-				return ((Integer) originalModifiers.get(method)).intValue();
+				return (Integer) originalModifiers.get(method);
 			}
 
 			public int getIndex(MethodInfo method) {
-				return ((Integer) indexes.get(method)).intValue();
+				return (Integer) indexes.get(method);
 			}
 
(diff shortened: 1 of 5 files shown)


spring-core/src/main/java/org/springframework/cglib/proxy/Enhancer.java
```diff
@@ -1254,11 +1254,11 @@ public ClassLoader getClassLoader() {
 			}
 
 			public int getOriginalModifiers(MethodInfo method) {
-				return ((Integer) originalModifiers.get(method)).intValue();
+				return (Integer) originalModifiers.get(method);
```
sbrannen, line 1257: Please refrain from modifying classes under `org.springframework.asm`, `org.springframework.cglib`, and `org.springframework.objenesis`. Those include repackaged forks of the third-party libraries ASM, CGLIB, and Objenesis. Any refactoring to those classes should take place upstream in the originating repository. The Spring Framework will then pick up the changes when syncing with official updates of the forked third-party libraries.

```suggestion
				return ((Integer) originalModifiers.get(method)).intValue();
```

spring-core/src/main/java/org/springframework/cglib/proxy/Enhancer.java
```diff
@@ -1254,11 +1254,11 @@ public ClassLoader getClassLoader() {
...
 			}
 
 			public int getIndex(MethodInfo method) {
-				return ((Integer) indexes.get(method)).intValue();
+				return (Integer) indexes.get(method);
```
sbrannen, line 1261: ditto

```suggestion
				return ((Integer) indexes.get(method)).intValue();
```

spring-web/src/main/java/org/springframework/http/converter/ResourceRegionHttpMessageConverter.java
```diff
@@ -201,10 +201,10 @@ private void writeResourceRegionCollection(Collection<ResourceRegion> resourceRe
 				print(out, "--" + boundaryString);
 				println(out);
 				if (contentType != null) {
-					print(out, "Content-Type: " + contentType.toString());
+					print(out, "Content-Type: " + contentType);
 					println(out);
 				}
-				Long resourceLength = region.getResource().contentLength();
+				long resourceLength = region.getResource().contentLength();
```
sbrannen, line 207: Good catch.

That one was obviously just an accident.

sbrannen (changes requested):

sbrannen:

Thanks

spring-web/src/main/java/org/springframework/http/converter/ResourceRegionHttpMessageConverter.java
stsypanov: No more commits into CGLIB )

PR #25097, opened 2020-05-18
BeanDefinitionDsl composition

krzykrucz:
Added new `beans` method to BeanDefinitionDsl, to allow nesting beans definitions. Currently it's not possible.

PR #25098, opened 2020-05-18
BeanDefinitionDsl: get environment property in a nice way

krzykrucz:
With this PR I'd like to open a discussion about a way to easily access Environment properties in `BeanDefinitionDsl`.


krzykrucz:
I am not sure as to the name of the method (`props`), because this method can be used not only to fetch single properties, but also whole configuration trees (as an alternative to a `@ConfigurationProperties` mechanism)

krzykrucz:
@sdeleuze Could you have a look, please?

krzykrucz:
@jhoeller @sdeleuze Would it be possible to speed up a bit this PR and https://github.com/spring-projects/spring-framework/pull/25097 ? 🙂

sdeleuze:
Hey, sorry for the delay. You can already access pretty easily to the environment via `env["foo"]`, `env.getProperty<Foo>("foo")` and `env.getRequiredProperty<Foo>("foo")` so not need for additional capabilities IMO.

PR #25126, opened 2020-05-26
Polish StringHttpMessageConverter.getContentTypeCharset()

quaff:
Bug: Possible null pointer dereference in org.springframework.http.converter.StringHttpMessageConverter.getContentTypeCharset(MediaType) due to return value of called method 
The return value from a method is dereferenced without a null check, and the return value of that method is one that should generally be checked for null.  This may lead to a NullPointerException when the code is executed.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/StringHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/StringHttpMessageConverter.java
@@ -143,7 +143,8 @@ public class StringHttpMessageConverter extends AbstractHttpMessageConverter<Str
 			Charset charset = contentType.getCharset();
 			if (charset != null) {
 				return charset;
-			} else if (contentType.isCompatibleWith(MediaType.APPLICATION_JSON)) {
+			}
+			else if (contentType.isCompatibleWith(MediaType.APPLICATION_JSON)) {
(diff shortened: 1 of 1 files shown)


spring-web/src/main/java/org/springframework/http/converter/StringHttpMessageConverter.java
```diff
@@ -139,18 +139,18 @@ protected void writeInternal(String str, HttpOutputMessage outputMessage) throws
...
-		}
-		else {
-			Charset charset = getDefaultCharset();
-			Assert.state(charset != null, "No default charset");
-			return charset;
+		if (contentType != null) {
+			Charset charset = contentType.getCharset();
+			if (charset != null) {
+				return charset;
+			} else if (contentType.isCompatibleWith(MediaType.APPLICATION_JSON)) {
```
sbrannen, line 146: Please adhere to project coding standards.

```suggestion
			}
			else if (contentType.isCompatibleWith(MediaType.APPLICATION_JSON)) {
```

sbrannen (changes requested):

sbrannen:
The non-null value returned by `contentType.getCharset()` cannot change between the `if` condition check and the `return` statement. So this is not a bug; however, invoking `contentType.getCharset()` only once is certainly an improvement.

sbrannen (approved):

sbrannen:

Thanks

Issue #25109, opened 2020-05-20
Custom HTTP Method on MockServerHttpRequest


GuyLewin:
In some cases we'd like to use `MockServerHttpRequest` with methods not in Spring's `HttpMethod` enum (for testing corner cases for example).
My scenario was wanting to use `MockServerHttpRequest` with the CONNECT method, which isn't in HttpMethod.
When `ServerHttpRequest` receives a request from Reactor Netty it can accept CONNECT as a string method even though it's not in the enum.
I believe we should also add a string-based method initialization to `MockServerHttpRequest`.

This was originally described in #24820 but I was asked to create a separate issue for this.
@rstoyanchev FYI

sbrannen:
As pointed out by @eleftherias, the change associated with this issue is actually a breaking change for anyone that previously supplied `null` for the `HttpMethod`.

For example, the following used to be supported but now throws an `IllegalArgumentException`.

```java
 MockServerHttpRequest.method(null, "/").build();
```

In light of that, perhaps we should reopen this issue to again support `null` for the `HttpMethod`.

@rstoyanchev Thoughts?

----

See also: https://github.com/spring-projects/spring-security/issues/8592

rstoyanchev:
My only concern is that Spring Security had no choice but to pass `null` since it was the only way to simulate a custom method. Aside from that there is no reason to pass `null`. My thought is to leave it the way it is but have a more helpful error message along the lines of "if you are passing `null` to simulate a custom method, please use ... instead". After all it's a test that's failing and it's easy to correct.

sbrannen:
> My thought is to leave it the way it is but have a more helpful error message

Fair enough. Let's just hope there weren't many projects depending on the previous unintentional _support_ for `null`.

midumitrescu:
Hello everybody, 

I played during the weekend with the idea of changing the HttpMethod to an interface and move the HttpMethod enum to HttpMethods. 

This introduces quite a few of necessary actions during an update to the newerst Spring version. 

You can check a comparison done under: https://github.com/midumitrescu/httpmethod-refactorings 

However, HTTP/1.1 and HTTP/2 both mention CONNECT as valid HTTP Method. 

Maybe one could consider simply adding the value to the enum? :)

midumitrescu:
Now, about changing just *MockServerHttpRequest*.

I haven't considered changing MockServerHttpRequest to allow for 
```
[...]

PR #25148, opened 2020-05-28
Refactoring for MockServerHttpRequest

midumitrescu:
Created a small refactoring for gh-25109. 

Changed: 

1. Field duplication
2. Set DefaultBodyBuilder as protected, to allow children from other packages to call constructor (it seemed very constrictive to allow the call of the contructor only from the existing package)
3. Moved the 
```
Assert.notNull(method, "HttpMethod is required. If testing a custom HTTP method, " +
					"please use the variant that accepts a String based HTTP method.");
```  
to DefaultBodyContructor to also catch the contructor call 
```
public static BodyBuilder method(HttpMethod method, URI url) {
		return new DefaultBodyBuilder(method, url);
	}
```
to leave the main constructor easier to understand and failfast when constructing he builder.
Also, messages when calling the 
``` 
MockServerHttpRequest.method(null, *any URI*)
```
would have failed with another kind of assertion message than 
```
MockServerHttpRequest.method(null, urlTemplate, param1, param2)
```

3. Extended the check for the string value of HttpMethod to exclude the set of HttpMethod to only whitespaces, as this, imho, makes no sense

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
+++ b/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
@@ -452,7 +452,8 @@ public final class MockServerHttpRequest extends AbstractServerHttpRequest {
 
 		protected DefaultBodyBuilder(String methodValue, URI url) {
 			Assert.isTrue(StringUtils.hasLength(methodValue) &&
-					StringUtils.hasLength(methodValue.trim()), "HTTP methodValue must not be empty");
+					StringUtils.hasLength(methodValue.trim()), "HttpMethod is required. " +
+					"Please initialize it to non empty value");
 			this.methodValue = methodValue.trim();
 			this.url = url;
 		}
--- a/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
+++ b/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
@@ -44,6 +44,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MimeType;
 import org.springframework.util.MultiValueMap;
+import org.springframework.util.StringUtils;
 import org.springframework.web.util.UriComponentsBuilder;
 
 /**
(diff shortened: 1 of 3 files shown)


midumitrescu:
Hey Rosen (@rstoyanchev ),

would you please mind taking a short look if this makes any sense to you?

spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
```diff
@@ -74,16 +74,13 @@
...
-	private MockServerHttpRequest(@Nullable HttpMethod httpMethod, @Nullable String customHttpMethod,
+	private MockServerHttpRequest(String httpMethodValue,
 			URI uri, @Nullable String contextPath, HttpHeaders headers, MultiValueMap<String, HttpCookie> cookies,
 			@Nullable InetSocketAddress remoteAddress, @Nullable InetSocketAddress localAddress,
 			@Nullable SslInfo sslInfo, Publisher<? extends DataBuffer> body) {
 
 		super(uri, contextPath, headers);
-		Assert.isTrue(httpMethod != null || customHttpMethod != null, "HTTP method must not be null");
-		this.httpMethod = httpMethod;
-		this.customHttpMethod = customHttpMethod;
```
rstoyanchev, line 86: We still can assert the httpMethodValue is not null.

rstoyanchev (commented):
Seems like a reasonable simplification and the `resolve` is a map lookup so it shouldn't be a concern.

spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
midumitrescu: I thought that all the checks are done in the Builder and the constructor is private. 

All classes extending the constructor must call super(), thus made sure there is no path able to initialize a null HttpMethod. 

But if you want I can surely add it

midumitrescu:
Quick question, sorry for asking this. 

There is another *MockServerHttpRequest* in
```
org.springframework.web.testfixture.http.server.reactive.MockServerHttpRequest
```

Is that auto generated or should it also receive this functionality ?

sbrannen:
> There is another _MockServerHttpRequest_ in

You will need to manually copy the changes to that class to keep them in sync,

midumitrescu:
> > There is another _MockServerHttpRequest_ in
> 
> You will need to manually copy the changes to that class to keep them in sync.

Done

spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
midumitrescu: @rstoyanchev I added some easy tests, testing all possible build paths. 

Should I anyways add the assertion here?

PR #25149, opened 2020-05-28
ExchangeStrategies custom codec's reader & writer are not registered

### LINKED ISSUE #25146: ExchangeStrategies custom codec's reader and writer are not registered
**Affects:** 5.2.6.RELEASE

---

I added my custom encoder and decoder like following.

```java
ExchangeStrategies
    .empty()
    .codecs(configurer -> {
        configurer.defaultCodecs().maxInMemorySize(properties.getMaxInMemorySize());
        configurer.customCodecs().registerWithDefaultConfig(encoder);
        configurer.customCodecs().registerWithDefaultConfig(decoder);
    })
    .build();
```

But there is no `messageWriters` when writing body with `BodyInserts.fromValue()`. 
It seems that encoder and decoder that I added above are not registered at `ExchangeStrategies`.

I think following code block has bug. Why put all readers and writers from `this` to `other`? It seems that `this` and `other` are swapped.

https://github.com/spring-projects/spring-framework/blob/90ccabd60bfe24249b3c4cbe43a25ffd0efa6eba/spring-web/src/main/java/org/springframework/http/codec/support/BaseCodecConfigurer.java#L144-L149

If it is a bug, I would like to create a PR. :)

dlsrb6342:
Resolves #25146

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/support/BaseCodecConfigurer.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/support/BaseCodecConfigurer.java
@@ -142,10 +142,10 @@ abstract class BaseCodecConfigurer implements CodecConfigurer {
 		 * @since 5.1.12
 		 */
 		DefaultCustomCodecs(DefaultCustomCodecs other) {
-			other.typedReaders.putAll(this.typedReaders);
-			other.typedWriters.putAll(this.typedWriters);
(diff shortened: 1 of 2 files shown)


sbrannen:
This has been merged into `5.2.x` and `master` and backported to `5.1.x`.

Thanks

Issue #25140, opened 2020-05-27
MultiValueMapAdapter.getFirst fails with IndexOutOfBoundsException in case of empty List


vlavasa:
If You create `MultiValueMap` via `MultiValueMapAdapter`, and value list is empty, call of method `getFirst` fail on `IndexOutOfBounds` exception:

```java
private static class MultiValueMapAdapter<K, V> implements MultiValueMap<K, V>, Serializable {

		private final Map<K, List<V>> map;

		public MultiValueMapAdapter(Map<K, List<V>> map) {
			Assert.notNull(map, "'map' must not be null");
			this.map = map;
		}

		@Override
		@Nullable
		public V getFirst(K key) {
			List<V> values = this.map.get(key);
			return (values != null ? values.get(0) : null);
		}
```

Good solution is:

```java
return (values != null && !values.isEmpty() ? values.get(0) : null);
```

jhoeller:
The common scenario is that such a list either contains one or more values or isn't set at all (in case of no value associated with a key). That said, point taken, we should align this with the existing `LinkedMultiValueMap` implementation where such a defensive emptiness check is present already.

vlavasa:
> 
> 
> The common scenario is that such a list either contains one or more values or isn't set at all (in case of no value associated with a key). That said, point taken, we should align this with the existing `LinkedMultiValueMap` implementation where such a defensive emptiness check is present already.

OK, but the adapter is used in method:
org.springframework.util.CollectionUtils#toMultiValueMap

and there is accepted any Map<K, List<V>> map

jhoeller:
Good point, it can be constructed with any target map content, including empty list values there. Also, an `addAll` call with an empty list argument will also lead to such empty list state, just like in `LinkedMultiValueMap`. I was just pointing out why it is uncommon to encounter empty lists there, otherwise it would be surprising for this bug to remain unnoticed so many years.

Thanks for raising this, in any case! I've marked it for a fix in the upcoming 5.2.7 and 5.1.16 releases.

midumitrescu:
@jhoeller could I give it a go?

my fix would be to have 

```
@Override
		@Nullable
		public V getFirst(K key) {
			List<V> values = this.map.getOrDefault(key, Collections.singletonList(null));
			return values.get(0);
		}
```

Does this make sense or should I better do the *?* contruct. 

I have also prepared unit tests.

jhoeller:
[...]

PR #25155, opened 2020-05-28
Resolves gh-25140

midumitrescu:
Simple fix for GH-25140

Just to be sure, the call 

```
multiValueMap.addAll("some-key", null);
```

will fail with NullPointer. 

```
multiValueMap.addAll("some-other-key", Collections.emptyList());
```

produces

```
multiValueMap.getFirst("some-other-key") == null
```


midumitrescu:
already resolved by @jhoeller with another commit

PR #25159, opened 2020-05-28
Tests for gh-25140

midumitrescu:
Created unit tests for 

```
CollectionUtils.toMultiValueMap
```

as suggested by @jhoeller in gh-25140 

I just followed the instructions given there although I could imagine a better solution which is a lot more DRY. 

I will create a second PR for comparison.

I would have 2 small questions: 

1. The code actually being tested lies in

```
CollectionUtils
```

but the test I have named **CollectionUtilsMultiValueMapAdapterTests**. 

Is this OK with you or are you following a strong name convention, thus I should move the tests to *CollectionUtilsTest* ? 

2. Is the gh issue number ok or should I create a separate one?


midumitrescu:
in favor of gh-25160

midumitrescu:
in favor of gh-25160

PR #25160, opened 2020-05-28
Introduce tests for MultiValueMap

midumitrescu:
imho improved tests for gh-25140, as solicited by @jhoeller 

As promised in gh-25159, I am submitting a second suggestion for Testing the exiting code of 

```
CollectionUtils.toMultiValueMap
```

The suggestion was to create for this MultiValueMap same tests as for 

```
LinkedMultiValueMap
```

present in
```
LinkedMultiValueMapTests
```

This, however, produced a very high code duplication, which I did not find very elegant. 

I think the better way is to use Jupiter properly, to instantiate various types of MultiValueMap and test the contract for each of them. 

Like this test code is not duplicated and the contracts are consistent with each other. 

This is just a suggestion. 

I am open for suggestions  in case this does not fit your expectations.

Kind regards

### DIFF
--- a/spring-core/src/test/java/org/springframework/util/MultiValueMapRelatedTests.java
+++ b/spring-core/src/test/java/org/springframework/util/MultiValueMapRelatedTests.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.util;
+
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Mihai Dumitrescu
+ * @author Arjen Poutsma
+ * @author Juergen Hoeller
+ */
+class MultiValueMapRelatedTests {
+
+	@ParameterizedTest
+	@MethodSource("objectsUnderTest")
+	void add(MultiValueMap<String, String> objectUnderTest) {
+		int startingSize = objectUnderTest.size();
+		objectUnderTest.add("key", "value1");
+		objectUnderTest.add("key", "value2");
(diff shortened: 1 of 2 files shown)


spring-core/src/test/java/org/springframework/util/MultiValueMapRelatedTests.java
```diff
@@ -0,0 +1,214 @@
...
+				() -> Assertions.assertThrows(UnsupportedOperationException.class,
+						() -> asUnmodifiableMultiValueMap.addIfAbsent("key", "value")),
+				() -> Assertions.assertThrows(UnsupportedOperationException.class,
+						() -> asUnmodifiableMultiValueMap.addAll("key", exampleListOfValues())),
+				() -> Assertions.assertThrows(UnsupportedOperationException.class,
+						() -> asUnmodifiableMultiValueMap.addAll(exampleMultiValueMap())),
+				() -> Assertions.assertThrows(UnsupportedOperationException.class,
+						() -> asUnmodifiableMultiValueMap.set("key", "value")),
+				() -> Assertions.assertThrows(UnsupportedOperationException.class,
+						() -> asUnmodifiableMultiValueMap.setAll(exampleHashMap())),
```
sbrannen, line 172: We tend to use `Collections.singletonMap` for such use cases.

```suggestion
						() -> asUnmodifiableMultiValueMap.setAll(Collections.singletonMap("key2", "key2.value1")),
```

spring-core/src/test/java/org/springframework/util/MultiValueMapRelatedTests.java
```diff
@@ -0,0 +1,214 @@
...
+		assertThat(objectUnderTest.getFirst("key")).isNull();
+	}
+
+	@ParameterizedTest
+	@MethodSource("objectsUnderTest")
+	void getFirst(MultiValueMap<String, String> objectUnderTest) {
+		List<String> values = new ArrayList<>(2);
+		values.add("value1");
+		values.add("value2");
+		objectUnderTest.put("key", values);
```
sbrannen, line 103: We tend to use `Arrays.asList` for such use cases.
```suggestion
		objectUnderTest.put("key", Arrays.asList("value1", "value2"));
```

spring-core/src/test/java/org/springframework/util/MultiValueMapRelatedTests.java
```diff
@@ -0,0 +1,214 @@
...
+		o2.put("key1", Collections.singletonList("value1"));
+		assertThat(o2).isEqualTo(objectUnderTest);
+		assertThat(objectUnderTest).isEqualTo(o2);
+	}
+
+	@ParameterizedTest
+	@MethodSource("objectsUnderTest")
+	void canNotChangeAnUnmodifiableMultiValueMap(MultiValueMap<String, String> objectUnderTest) {
+		MultiValueMap<String, String> asUnmodifiableMultiValueMap = CollectionUtils.unmodifiableMultiValueMap(objectUnderTest);
+		Assertions.assertAll(
```
sbrannen, line 160: It is not permissible to use JUnit Jupiter's `Assertions`. Doing so will actually fail the build.

Thus, please switch to AssertJ assertions and be sure to execute `./gradlew check` to verify that you don't have any Checkstyle violations in this regard.

Note: if you want something similar to `assertAll`, you may wish to use `SoftAssertions` from AssertJ.

sbrannen (changes requested):
I think the use of parameterized tests is a good idea.

I've requested a few minor changes and will review again once those have been made.

sbrannen (approved):

midumitrescu:
@sbrannen i am on it :)

I need a couple of hours because of personal stuff and I will finish the PR afterwards.

midumitrescu:
@sbrannen done

snicoll:
Thanks @midumitrescu, I've enabled the tests that were disabled as the fix has been added in the meantime.

PR #25194, opened 2020-06-05
CaffeineCache with LoadingCache

### LINKED ISSUE #25173: CaffeineCache does not support LoadingCache consistently
Hi all
CaffeineCache implementation override method `public ValueWrapper get(Object key)` and there is a specific check when a LoadingCache is provided.
In this case, the CacheLoader of the LoadingCache is used to retrieve the data (if not already present int the cache).

I don't see the same behavior for the method `public <T> T get(Object key, @Nullable Class<T> type)`.
In this case, the CacheLoader of the LoadingCache is never used. Using only this method... the cache will be always empty.

Is there any reason for this different behavior?

Thanks
Stefano

stefano-salmaso:
When CaffeineCache is created with a LoadingCache, the LoadingCache is
used with the method `public ValueWrapper get(Object key)` but not
with method `public <T> T get(Object key, @Nullable Class<T> type)`.
Using LoadingCache is incompatible with @Cacheable pattern but
consistently call the LoadingCache in both methods is better than now.
Removing the usage is a backward-incompatible change.

Closes gh-25173

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCache.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCache.java
@@ -29,6 +29,10 @@ import org.springframework.util.Assert;
  * Spring {@link org.springframework.cache.Cache} adapter implementation
  * on top of a Caffeine {@link com.github.benmanes.caffeine.cache.Cache} instance.
  *
+ * <p>IMPORTANT!
+ * <p>You can pass also a {@link com.github.benmanes.caffeine.cache.LoadingCache} to CaffeineCache, but
+ * using LoadingCache is incompatible with the @Cacheable pattern!
+ *
  * <p>Requires Caffeine 2.1 or higher.
  *
  * @author Ben Manes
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@stefano-salmaso Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25194)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25194) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@stefano-salmaso Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25194)!

stefano-salmaso:
@snicoll ... any chance to take this PR into consideration? 
(also close it is a possibility.. 😀)

snicoll:
It very much was @stefano-salmaso but I've been busy on other things lately. I apologize for the delay and I've flagged the original issue for the `5.3` bucket. I am not sure if we're going to fix it the way you've suggested so I'll keep both the issue and the PR open for now.

snicoll:
Thanks for the PR but I've decided to use a different approach, see https://github.com/spring-projects/spring-framework/commit/f3cedf268b3ef0c3ef14766db4e02d17346f8b24 for more details.

Issue #25182, opened 2020-06-03
Different behaviour with controller query parameters mapping in object in Kotlin and Java


Ivanbmstu:
Spring boot 2.3.0, java 11.

Hello everyone, i trying to map query params into object with kotlin data class.
When i query my controller with params /?values=1,2,3, then list parameter maps into one string instead of list of strings.
After that i decided to reproduce this error with java service, on pojo without constructor values maps correctly, on pojo with constructor values maps into single string.

Expected behaviour: query parameter maps in both cases into list divided by coma separator

Is it correct behaviour or bug? I dont think it is spring boot issue, so i created it in the spring-framework repository

Sorry for my english.

Repo to reproduce: https://github.com/Ivanbmstu/spring_query_mapping

Run build in both modules (Kotlin and Java)

encircled:
Hi,

I can confirm there is a difference in the params resolving between constructor and setter injections. Will prepare a PR soon.

sbrannen:
Superseded by #25200

PR #25200, opened 2020-06-06
Fix web parameters resolution when injected via constructor

encircled:
This fixes the problem described in #25182 

Comma-separated multi value parameters are wrongly injected if constructor injection is used. Only spring-web is affected, webflux works correctly.

For example `...?listOfStrings=1,2` becomes `List<String> listOfStrings = ["1,2"]`, instead of expected `List<String> listOfStrings = ["1", "2"]`. The reason is that `WebRequest#getParameterValues` exposes single item params as an array which breaks conversion service later on.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/annotation/ModelAttributeMethodProcessor.java
+++ b/spring-web/src/main/java/org/springframework/web/method/annotation/ModelAttributeMethodProcessor.java
@@ -17,6 +17,7 @@
 package org.springframework.web.method.annotation;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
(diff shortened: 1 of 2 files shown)


sbrannen:
Thanks for the reminder.

Since this appears to be a bug, we'll see about getting it merged into `main` for inclusion in 5.3.7 and potentially for a backport to 5.2.x.

encircled:
Alright, I will rebase it on the latest master then

sbrannen:
> Alright, I will rebase it on the latest master then

Thanks for rebasing it on `main`. 😉

sbrannen:
This has been merged into `main` in 9ddab9e69b5a1e1e1c87c8542eaa6aac060cc8e9 and revised in 355d394d7fff72844445839a7d8d2b359aec0ced.

Thanks for providing the fix!

PR #25261, opened 2020-06-17
Remove redundant assignment of default values to volatile fields

stsypanov:
Assignment of default values to fields in general case is redundant, because the value will be there according to JLS. However in case of volatile field there is also performance issue, consider benchmark
```java
@State(Scope.Thread)
@BenchmarkMode(value = Mode.AverageTime)
@Fork(jvmArgsAppend = {"-Xms4g", "-Xmx4g"})
public class VolatileFieldBenchmark  {
  @Benchmark
  public Object explicitInit() {
    return new Class1();
  }
  @Benchmark
  public Object noInit() {
    return new Class2();
  }
  private static class Class1 {
    private volatile boolean field = false;
    public boolean isField() {
      return field;
    }
    public void setField(boolean field) {
      this.field = field;
    }
  }
  private static class Class2 {
    private volatile boolean field;
    public boolean isField() {
      return field;
    }
    public void setField(boolean field) {
      this.field = field;
    }
  }
}
```
The benchmark demonstrates that instantiation of a class without default value assignment takes less time:
```
Benchmark                            Mode  Cnt   Score   Error  Units
VolatileFieldBenchmark.explicitInit  avgt   40  11.087 ± 0.140  ns/op
VolatileFieldBenchmark.noInit        avgt   40   3.367 ± 0.131  ns/op
```

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/MutablePropertyValues.java
+++ b/spring-beans/src/main/java/org/springframework/beans/MutablePropertyValues.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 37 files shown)


jhoeller:
Well spotted, a good general rule to follow! Thanks for raising this.

PR #25298, opened 2020-06-22
DefaultSubscriptionRegistry: Reduced thread contention

### LINKED ISSUE #24395: Thread BLOCKED issues of DestinationCache in DefaultSubscriptionRegistry
We are using the spring-messaging to implement a STOMP server (using SimpleBrokerMessageHandler).
The Client will subscribe on 5 channel and everything is ok when there are only a few users.
However, when the online user is above ~ 700, the websocket channel is "out of response".

After analysis, I found many other thread has been "BLOCKED" by DestinationCache, as follows:

```bash
"pk-ws-worker-100-thread-78" #560 prio=5 os_prio=0 tid=0x00007ff19c182000 nid=0x1252 waiting for monitor entry [0x00007ff0a8d9f000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.springframework.messaging.simp.broker.DefaultSubscriptionRegistry$DestinationCache.getSubscriptions(DefaultSubscriptionRegistry.java:269)
        - waiting to lock <0x00000004c007ec20> (a org.springframework.messaging.simp.broker.DefaultSubscriptionRegistry$DestinationCache$1)
```

And part of the code are as follows:

```java
public LinkedMultiValueMap<String, String> getSubscriptions(String destination, Message<?> message) {
  LinkedMultiValueMap<String, String> result = this.accessCache.get(destination);
  if (result == null) {
    synchronized (this.updateCache) {
      result = new LinkedMultiValueMap<>();
      for (SessionSubscriptionInfo info : subscriptionRegistry.getAllSubscriptions()) {
        for (String destinationPattern : info.getDestinations()) {
          if (getPathMatcher().match(destinationPattern, destination)) {
            for (Subscription sub : info.getSubscriptions(destinationPattern)) {
              result.add(info.sessionId, sub.getId());
            }
          }
        }
      }
      if (!result.isEmpty()) {
        this.updateCache.put(destination, result.deepCopy());
        this.accessCache.put(destination, result);
      }
    }
  }
  return result;
}
```
As you can see, the code inside synchronized will traverse all subscription, which will cost too much time and block other Thread.

Also, the accessCache / updateCache is not works if the client has not success make the subscription, which will make the situation worse.

We try to increase the cache limit and it does't work for our case.

To solve the problem, we remove the DestinationCache and
reimplement an Map -> <sessionId -> subsId> inside SessionSubscriptionRegistry.
(in our own codebase of course)

After theses change, the server can handle more than 5K online users with no problem.

Meanwhile, I noticed that DefaultSubscriptionRegistry and DestinationCache has been there for many years.

So, I just wonder is it ok to make a pr?
Or the existing DestinationCache is good for some other reason?

trim09:
* DestinationCache is now synchronized on multiple 'destination' locks
  (previously a single shared lock)
* DestinationCache keeps destinations without any subscriptions
 (previously such destinations were recomputed over and over)
* SessionSubscriptionRegistry is now a
  'sessionId -> subscriptionId -> (destination,selector)' map
  for faster lookups
  (previously 'sessionId ->  destination -> set of (subscriptionId,selector)')

closes gh-24395

### DIFF
--- a/spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
+++ b/spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.messaging.simp.broker;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.Blackhole;
+
+import org.springframework.messaging.Message;
+import org.springframework.messaging.simp.SimpMessageHeaderAccessor;
+import org.springframework.messaging.simp.SimpMessageType;
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.util.MultiValueMap;
+
+@BenchmarkMode(Mode.Throughput)
+public class DefaultSubscriptionRegistryBenchmark {
+
+	@State(Scope.Benchmark)
+	public static class ServerState {
+		@Param("1000")
+		public int sessions;
+
+		@Param("10")
+		public int destinations;
+
+		@Param({"0", "1024"})
+		int cacheSizeLimit;
+
+		@Param({"none", "patternSubscriptions", "selectorHeaders"})
+		String specialization;
+
+		public DefaultSubscriptionRegistry registry;
+
+		public String[] destinationIds;
+
+		public String[] sessionIds;
+
+		public AtomicInteger uniqueIdGenerator;
+
+		public Message<?> findMessage;
+
+		@Setup(Level.Trial)
+		public void doSetup() {
+			this.findMessage = MessageBuilder.createMessage("",  SimpMessageHeaderAccessor.create().getMessageHeaders());
+			this.uniqueIdGenerator = new AtomicInteger();
+
+			this.registry = new DefaultSubscriptionRegistry();
+			this.registry.setCacheLimit(this.cacheSizeLimit);
+			this.registry.setSelectorHeaderName("selectorHeaders".equals(this.specialization) ? "someSelector" : null);
+
+			this.destinationIds = IntStream.range(0, this.destinations)
+					.mapToObj(i -> "/some/destination/" + i)
+					.toArray(String[]::new);
+
+			this.sessionIds = IntStream.range(0, this.sessions)
+					.mapToObj(i -> "sessionId_" + i)
+					.toArray(String[]::new);
+
+			for (String sessionId : this.sessionIds) {
+				for (String destinationId : this.destinationIds) {
+					registerSubscriptions(sessionId, destinationId);
+				}
+			}
+		}
+
+		public void registerSubscriptions(String sessionId, String destination) {
+			if ("patternSubscriptions".equals(this.specialization)) {
+				destination = "/**/" + destination;
+			}
+			String subscriptionId = "subscription_" + this.uniqueIdGenerator.incrementAndGet();
+			this.registry.registerSubscription(subscribeMessage(sessionId, subscriptionId, destination));
+		}
+	}
+
+	@State(Scope.Thread)
+	public static class Requests {
+		@Param({"none", "sameDestination", "sameSession"})
+		String contention;
+
+		public String session;
+
+		public Message<?> subscribe;
+
+		public String findDestination;
+
+		public Message<?> unsubscribe;
+
+		@Setup(Level.Trial)
+		public void doSetup(ServerState serverState) {
+			int uniqueNumber = serverState.uniqueIdGenerator.incrementAndGet();
+
+			if ("sameDestination".equals(this.contention)) {
+				this.findDestination = serverState.destinationIds[0];
+			}
+			else {
+				this.findDestination = serverState.destinationIds[uniqueNumber % serverState.destinationIds.length];
+			}
+
+			if ("sameSession".equals(this.contention)) {
+				this.session = serverState.sessionIds[0];
+			}
+			else {
+				this.session = serverState.sessionIds[uniqueNumber % serverState.sessionIds.length];
+			}
+
+			String subscription = String.valueOf(uniqueNumber);
+			String subscribeDestination = "patternSubscriptions".equals(serverState.specialization) ?
+					"/**/" + this.findDestination : this.findDestination;
+			this.subscribe = subscribeMessage(this.session, subscription, subscribeDestination);
+
+			this.unsubscribe = unsubscribeMessage(this.session, subscription);
+		}
+	}
+
+	@State(Scope.Thread)
+	public static class FindRequest {
+		@Param({"none", "noSubscribers", "sameDestination"})
+		String contention;
+
+		public String destination;
+
+		@Setup(Level.Trial)
+		public void doSetup(ServerState serverState) {
+			switch (this.contention) {
+				case "noSubscribers":
+					this.destination = "someDestination_withNoSubscribers_" + serverState.uniqueIdGenerator.incrementAndGet();
+					break;
+				case "sameDestination":
+					this.destination = serverState.destinationIds[0];
+					break;
+				case "none":
+					int uniqueNumber = serverState.uniqueIdGenerator.getAndIncrement();
+					this.destination = serverState.destinationIds[uniqueNumber % serverState.destinationIds.length];
+					break;
+				default:
+					throw new IllegalStateException();
+			}
+		}
+	}
+
+	@Benchmark
+	public void registerUnregister(ServerState serverState, Requests request, Blackhole blackhole) {
+		serverState.registry.registerSubscription(request.subscribe);
+		blackhole.consume(serverState.registry.findSubscriptionsInternal(request.findDestination, serverState.findMessage));
+		serverState.registry.unregisterSubscription(request.unsubscribe);
+		blackhole.consume(serverState.registry.findSubscriptionsInternal(request.findDestination, serverState.findMessage));
+	}
+
+	@Benchmark
+	public MultiValueMap<String, String> find(ServerState serverState, FindRequest request) {
+		return serverState.registry.findSubscriptionsInternal(request.destination, serverState.findMessage);
+	}
+
+	public static Message<?> subscribeMessage(String sessionId, String subscriptionId, String dest) {
+		SimpMessageHeaderAccessor accessor = SimpMessageHeaderAccessor.create(SimpMessageType.SUBSCRIBE);
+		accessor.setSessionId(sessionId);
+		accessor.setSubscriptionId(subscriptionId);
+		accessor.setDestination(dest);
+		accessor.setNativeHeader("someSelector", "true");
+		return MessageBuilder.createMessage("", accessor.getMessageHeaders());
+	}
+
+	public static Message<?> unsubscribeMessage(String sessionId, String subscriptionId) {
+		SimpMessageHeaderAccessor accessor = SimpMessageHeaderAccessor.create(SimpMessageType.UNSUBSCRIBE);
+		accessor.setSessionId(sessionId);
+		accessor.setSubscriptionId(subscriptionId);
+		return MessageBuilder.createMessage("", accessor.getMessageHeaders());
+	}
+}
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
@@ -16,15 +16,17 @@
 
 package org.springframework.messaging.simp.broker;
 
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
+import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BiConsumer;
 
 import org.springframework.expression.EvaluationContext;
 import org.springframework.expression.Expression;
@@ -34,6 +36,7 @@ import org.springframework.expression.TypedValue;
 import org.springframework.expression.spel.SpelEvaluationException;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.SimpleEvaluationContext;
+import org.springframework.lang.NonNull;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
@@ -72,7 +75,7 @@ public class DefaultSubscriptionRegistry extends AbstractSubscriptionRegistry {
 
 	private PathMatcher pathMatcher = new AntPathMatcher();
 
-	private volatile int cacheLimit = DEFAULT_CACHE_LIMIT;
+	private int cacheLimit = DEFAULT_CACHE_LIMIT;
 
 	@Nullable
 	private String selectorHeaderName = "selector";
@@ -106,6 +109,7 @@ public class DefaultSubscriptionRegistry extends AbstractSubscriptionRegistry {
 	 */
 	public void setCacheLimit(int cacheLimit) {
 		this.cacheLimit = cacheLimit;
+		this.destinationCache.ensureCacheLimit();
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
```diff
@@ -0,0 +1,206 @@
...
+import org.springframework.messaging.simp.SimpMessageHeaderAccessor;
+import org.springframework.messaging.simp.SimpMessageType;
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.util.MultiValueMap;
+
+@BenchmarkMode(Mode.Throughput)
+@Fork(value = 1, warmups = 1)
+@Threads(4)
+@Measurement(iterations = 3, time = 2)
+@Warmup(time = 1, iterations = 3)
```
bclozel, line 46: Could you remove other annotations (besides `@BenchmarkMode`) from the class? We're trying to keep those options at the CLI level only.

spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
```diff
@@ -0,0 +1,206 @@
...
+
+		@Setup(Level.Trial)
+		public void doSetup(ServerState serverState) {
+			this.destination = serverState.destinations[this.counter.getAndIncrement() % serverState.destinations.length];
+			this.message = null;
+		}
+	}
+
+	@Benchmark
+	public void registerSubscription(ServerState serverState, Requests request) {
```
bclozel, line 169: Benchmark methods should either return a value or use `Blackhole` to consume values produced by the benchmark method. Without that, we're at risk that the JVM performs dead code elimination. See https://github.com/spring-projects/spring-framework/wiki/Micro-Benchmarks for more information.

The register/unregister subscriptions return `void`, but using `Blackhole` on `findSubscriptionsInternal` should do the trick.

spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
```diff
@@ -0,0 +1,206 @@
...
+		serverState.registry.registerSubscription(request.subscribe);
+		serverState.registry.findSubscriptionsInternal(request.findDestination, null);
+		serverState.registry.unregisterSubscription(request.unsubscribe);
+	}
+
+	@Benchmark
+	public void unregisterSession(ServerState serverState, UnregisterAllSubscriptionRequest request) {
```
bclozel, line 176: without a way to consume the result of the operation (maybe consuming the list of subscriptions after unregistering?), maybe removing the benchmark is the best course of action...

spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
```diff
@@ -208,304 +216,245 @@ public void unregisterAllSubscriptions(String sessionId) {
...
-					if (logger.isDebugEnabled()) {
-						logger.debug("Failed to evaluate selector: " + ex.getMessage());
-					}
-				}
-				catch (Throwable ex) {
-					logger.debug("Failed to evaluate selector", ex);
-				}
+			SessionSubscriptionInfo subscriptions = this.subscriptionRegistry.getSubscriptions(sessionId);
+			if (subscriptions != null) {
+				subIds.stream()
```
bclozel, line 221: We've found in the past (especially on JDK8) that using java.util.Steam on hot code paths hinders performance. Could you change this block accordingly?

bclozel (changes requested):
Thanks for this PR!
I really like the fact that you added benchmarks to discuss the performance improvements, thank you!

rstoyanchev (commented):
I've had a detailed first look and scheduled it for 5.3 M2. The only comment I have at this time is to skip the use of `Stream` to reduce Object allocation along with the comments about the benchmark from @bclozel.

Let me know if you plan to update those, or if not we can also take it from here.

Thanks for all this!

spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
```diff
@@ -0,0 +1,206 @@
...
+
+		public String[] sessionIds;
+
+		@Setup(Level.Trial)
+		public void doSetup() {
+			this.registry = new DefaultSubscriptionRegistry();
+
+			this.destinations = IntStream.range(0, this.numberOfDestinations)
+					.mapToObj(i -> "/some/destination/" + i)
+					.toArray(String[]::new);
```
rstoyanchev, line 72: There probably should be some pattern destinations.

7erg:
Hey! It's very good implementation (especially its performance), but I have run into a deadlock:)
```
private final class DestinationCache {
...
        private final Queue<String> cacheEvictionPolicy = new LinkedList<>();

        public LinkedMultiValueMap<String, String> getSubscriptions(String destination) {
            LinkedMultiValueMap<String, String> subscriptions = this.destinationCache.get(destination);
            if (subscriptions == null) {
                subscriptions = this.destinationCache.computeIfAbsent(destination, dest -> { // catalina-exec-1 locked 0x0000000532af8070
                    LinkedMultiValueMap<String, String> sessionSubscriptions = calculateSubscriptions(destination);
                    synchronized (this.cacheEvictionPolicy) { //catalina-exec-1 waiting for lock 0x0000000506bb19b0
                        this.cacheEvictionPolicy.add(destination);
                    }
                    return sessionSubscriptions;
                });
                ensureCacheLimit();
            }
            return subscriptions;
        }

        private void ensureCacheLimit() {
            synchronized (this.cacheEvictionPolicy) { //catalina-exec-2 locked 0x0000000506bb19b0
                int limit = cacheLimit;
                while (this.cacheEvictionPolicy.size() > limit) {
                    this.destinationCache.remove(this.cacheEvictionPolicy.poll()); // catalina-exec-2 waiting to lock 0x0000000532af8070 in destinationCache.remove(...)
                }
            }
        }
...
}
```

I think it could be fixed removing first `synchronized (this.cacheEvictionPolicy)` and using `private final Queue<String> cacheEvictionPolicy = new ConcurrentLinkedQueue<>();`

spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
trim09: 👍 I will remove them

spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
trim09: thanks for this finding, I will add the blackhole

spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
trim09: I have never spot any observable performance degradation on JDK8 related to streams, but I will follow your advice and remove all streams for now.
I almost forgot that someone on the world is still riding that dead horse 😄

trim09:
> Hey! It's very good implementation (especially its performance), but I have run into a deadlock:)
ah, my apologize. I've found the same this morning.
ConcurrentLinkedQueue is a great lock-free implementation, but have a look on it's size() method implementation. It traverse whole linked list to count the size :-/
LinkedBlockingQueue has two locks - one for add operation and one for take and fast size() operation. I think this can fix it.

spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
trim09: I will add a benchmark for pattern destinations. Actually, there two places where I can put a pattern.
1) new subscription can be a pattern destination
2) other already registered subscriptions can be a pattern
I will add a flag at both places.

spring-messaging/src/jmh/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistryBenchmark.java
trim09: Your are right, the JVM eliminated that code and when I add a findSubscriptionsInternal() the "performance" dropped. Not it looks like it is measuring something meaningful. Good finding, thank you.

trim09:
Thank you all for the review. I have fixed all findings mentioned above. Please have a look.

spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
```diff
@@ -208,304 +214,250 @@ public void unregisterAllSubscriptions(String sessionId) {
...
-								for (Subscription sub : info.getSubscriptions(destinationPattern)) {
-									result.add(info.sessionId, sub.getId());
-								}
-							}
-						}
-					}
-					if (!result.isEmpty()) {
-						this.updateCache.put(destination, result.deepCopy());
-						this.accessCache.put(destination, result);
+		private final Queue<String> cacheEvictionPolicy = new LinkedBlockingQueue<>();
```
trim09, line 259: I did some perf testing of "LinkedBlockingQueue vs ConcurrentLinkedQueue" and I have no clear winner.
I think ConcurrentLinkedQueue.size() would cause lots of CPU cache miss while traversing through the whole linked list. That is something I don't know hot to test on local machine without some bigger test preparation. That is why I left LinkedBlockingQueue there, but I will change it to ConcurrentLinkedQueue if you like it more.

trim09: finally, I used ConcurrentLinkedQueue and AtomicInteger to maintain the size.

bclozel (approved):
I think there's only one remaining task being discussed with Rossen: pattern destinations in the benchmark.

Other than that, it looks good! Thanks very much!

trim09:
I dedicated my Laptop i5-8250U @ 1.6Ghz, ram @ 2.4Ghz to perf tests for two days .Because it took ~6hours to run one set of JMH tests, I would like to share them with you. I ran the following:

1. results-newImpl-4-threads.txt - proposed implementation with @Threads(4) 
2. results-newImpl-single-thread.txt - proposed implementation
3. results-oldfixed-single-thread.tx - old implementation with a cacheLimit(0) fix described below
4. results-oldFixed-4-threads.txt] - old implementation with @Threads(4) and the cacheLimit(0) fix

I had to fix setCacheLimit(0) on the old implementation by swapping these two lines:
`
this.updateCache.put(destination); // check cache limit and remove eldest if over size limit
this.accessCache.put(destination); // add item int the cache
`
They were in a wrong order and could not keep the cache empty which I need for the perf test.

image

[results-oldFixed-4-threads.txt](https://github.com/spring-projects/spring-framework/files/4905344/results-oldFixed-4-threads.txt)
[results-oldfixed-single-thread.txt](https://github.com/spring-projects/spring-framework/files/4905346/results-oldfixed-single-thread.txt)


[results-newImpl-single-thread.txt](https://github.com/spring-projects/spring-framework/files/4905343/results-newImpl-single-thread.txt)
[results-newImpl-4-threads.txt](https://github.com/spring-projects/spring-framework/files/4905342/results-newImpl-4-threads.txt)

spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
```diff
@@ -142,14 +146,17 @@ public String getSelectorHeaderName() {
...
 		Expression expression = getSelectorExpression(message.getHeaders());
-		this.subscriptionRegistry.addSubscription(sessionId, subsId, destination, expression);
-		this.destinationCache.updateAfterNewSubscription(destination, sessionId, subsId);
+		boolean isAntPattern = this.pathMatcher.isPattern(destination);
+		Subscription subscription = new Subscription(subscriptionId, expression, destination, isAntPattern);
+
+		Subscription previousValue = this.subscriptionRegistry.addSubscription(sessionId, subscriptionId, subscription);
+		if (previousValue == null) {
+			this.destinationCache.updateAfterNewSubscription(destination, isAntPattern, sessionId, subscriptionId);
+		}
```
rstoyanchev, line 159: Subscription id's are supposed to be unique within a session/connection. That means previousValue should always be null. Do you know of a case where it wouldn't be?

rstoyanchev:
One thing that would still be useful for the benchmark is a setup category that has both finding and registering/unrestering. Within that category, patterns vs no patterns, where patterns should probably exceed the cache size, since they allow variation in destinations and a greater number of destinations, e.g. with path variables `"/some/destination/{id}"`.

trim09:
> ....where patterns should probably exceed the cache size, since they allow variation in destinations and a greater number of destinations, e.g. with path variables `"/some/destination/{id}"`.
@rstoyanchev 
I think I see your point, but I am not sure if I fully understand. I think there are no "pattern destinations". There are only "pattern subscriptions". Patterns can be passed in only on a subscription registration (the client registers on e.g. _"/some/destination/*"_ via _addSubscriptionInternal()_).  There is no possibility to use patterns from the server side. The server cannot send to  _"/some/destination/*"_, he needs to pick and choose one of static destinations e.g.  _"/some/destination/id12345"_.

Any (static/pattern) subscription registration updates only destinations that are currently cached. In other words subscription registration does not add any new cache entries and could not trigger a cache overflow. 

The cache overflow can happen then server sends (_findSubscriptionsInternal()_ method) to more destinations than is the cache size limit. This case is already covered by a benchmark, please see a _cacheSizeLimt=0_ column in a table above. 

I am happy to implement more benchmarks. Could you please describe the benchmark setup you proposed in more details so I can get on the same page?

rstoyanchev:
Sorry, I meant to say where destinations (not patterns) exceed the cache size. I was merely pointing out the current setups demonstrate well individual aspects but it would be useful to have a combined setup (both find and subscribe/unsubscribe) as it happens at runtime. My second thought was that when patterns are used, it's easy to imagine the number of destinations exceeding the cache size and that would also be a good match for such a combined setup. Does that make sense?

By the way I'm already reviewing and polishing locally so please don't submit any further changes. I was just thinking out loud but it can be done separately later.

spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
trim09: I had "better to be safe than sorry" on my mind. Actually, you are right and it complicated the code. I removed it. Thanks for this finding.

trim09:
@rstoyanchev 
Ok, thanks a lot. I will leave it as is for you without further changes to save you from ugly merges.
I left removal of a _this.subscriptionRegistry.addSubscription_  return value in my local repo, please, do It as a part of your polishing.

I am sorry for closing and reopening this issue. I unintentionally deleted (and then resurrected) a remote branch in my github repo and it most likely closed this PR. Please force push your changes if needed.

>combined setup (both find and subscribe/unsubscribe) as it happens at runtime

I really like your idea of a more realistic performance test.  We can do a setup where e.g.:
* server is publishing messages on 10000 different destinations 
* cache size is 1024
* 1000? client sessions
* each session has x% of static destination subscriptions and 100-x% of pattern subscriptions
Honestly, I am not sure if I am able to setup such a test, because I think usages of websocket can differ so much, that I cannot think of all in a test.  

E.g. our case on a single server in production is:
* 2000 sessions
* we use static subscriptions only (no patterns)
* ~300 destinations in total. Some has no subscribers. Some (4 destinations) are subscribed by all sessions.
* in average ~7 subscriptions per session. 4 common and then e.g. 3 other subscriptions.
* 4000 delivered msg/s

I can imagine that someone is not using static destinations, but is using patterns only, or is registered to much more destinations.
But you are right, we can at least try to deliver some kind setup and do a benchmark. Maybe I will get back to it once you are done with polishing/merging. 

>destinations (not patterns) exceed the cache size.

I think, this kind of benchmark is already there in a form of cache size = 0. In such case any _find_ operation does not find an cached entry and perform recalculation, store and evict.

rstoyanchev:
@trim09 your changes are now in master. 

I did a little polishing, completely  functionally neutral. That aside I did a review with @jhoeller and spotted one issue. Since the cacheSize is updated before the queue, it can lead to `cacheSize` > `queue.size()` if competing threads have updated one but not yet the other. Then a thread checking `cacheSize` can `poll` more than is available in the queue getting back `null` and leading to NPE from `ConcurrentHashMap`. Switching the order ensures the queue always has more elements than the size currently indicates but eventually they'll get consistent as threads catch up. So know we also replace `poll` with `remove` since the queue should return an item, or otherwise it is a bug.

There is one more point that I wanted to raise before making any changes. In the previous implementations, removal actually iterated over the cache and looked for a match on the `subscriptionId` to find matching destinations rather than using pattern matching. Is there any reason we can't do that still for subscriptions with a pattern?

> I am sorry for closing and reopening this issue.

No worries.

For the more realistic test, yes we can work on that next and it can have a couple of sub-categories with static destinations like your scenario and also with patterns. Here is [one example](https://github.com/rstoyanchev/spring-websocket-portfolio/blob/774e21dd6a6efae7b81645f20c7ced9123c47816/src/main/webapp/static/js/services.js#L75) with stock quotes by ticker which would then [broadcasts](https://github.com/rstoyanchev/spring-websocket-portfolio/blob/774e21dd6a6efae7b81645f20c7ced9123c47816/src/main/java/org/springframework/samples/portfolio/service/QuoteService.java#L66).

trim09:
> In the previous implementations, removal iterated over the cache looking for a match by `subscriptionId` as a way of finding matching destinations rather than pattern matching destinations. Is there any reason we can't do the same still for subscriptions with a pattern, i.e. [this part](https://github.com/spring-projects/spring-framework/blob/b451379c2f6da834a24d230bc9f1355660da3e7c/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java#L344-L349)?

No, there was no reason. The older way looks better and is maybe even faster. Please, feel free to change [it](https://github.com/spring-projects/spring-framework/blob/b451379c2f6da834a24d230bc9f1355660da3e7c/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java#L344-L349) to  something like this:
```
String subscriptionId = subscription.getId();
this.destinationCache.forEach((destination, sessionIdToSubscriptionIds) -> {
	List<String> subscriptionIds = sessionIdToSubscriptionIds.get(sessionId);
	if (subscriptionIds != null && subscriptionIds.contains(subscriptionId)) {
		removeInternal(destination, sessionId, subscriptionId);
	}
});
```

> `cacheSize` is updated before the queue

Great catch! I agree, it was in a wrong order. 👍  

> use `remove` instead of `poll` since the absence of an item would be a bug.

👍 

> I did a little polishing

It looks great! Thanks.

> your changes are now in master. Thanks again for the very detailed contribution!

Cool! I really enjoyed my first contribution here and I am considering a next performance improvement.

trim09:
I closed this review as it has been already merged in master.

rstoyanchev:
I've updated the removal logic so we should be all good here.

> I really enjoyed my first contribution here and I am considering a next performance improvement.

Thanks again, much appreciated!

PR #25308, opened 2020-06-24
simplify the code, replace if with switch

ghost:
simplify the code, replace if with switch


sbrannen:

Thanks

PR #25311, opened 2020-06-24
Use @SafeVarargs in Jackson builder and factory

### LINKED ISSUE #25293: Use @SafeVarargs in Jackson2ObjectMapperBuilder#modulesToInstall
**Affects:** 5.2.7

---

When using `Jackson2ObjectMapperBuilder#modulesToInstall(Class<? extends Module>...)` you get the following warning

    Type safety: A generic array of Class<? extends Module> is created for a varargs parameter

This can for example be seen in `Jackson2ObjectMapperBuilderTests#customizeWellKnownModulesWithModuleClass()` which needs an additional `@SuppressWarnings("unchecked")`

This could easily be solved by annotating `#modulesToInstall` with `@SafeVarargs` instead of `@SuppressWarnings("unchecked")`. The downside of this is that `#modulesToInstall` would have to be declared `final`. If this is acceptable I can submit a PR.

This also applied to `Jackson2ObjectMapperFactoryBean#setModulesToInstall(Class<? extends Module>...)`.

marschall:
Using @SafeVarargs in Jackson mapper builder and factory bean classes
allows the varargs methods to be used without a compiler warning. The
implementations of these methods do not perform unsafe operations on
their varargs parameter. It is therefore safe to add this annotation.

The following two methods are changed:

- add @SafeVarargs to Jackson2ObjectMapperBuilder#modulesToInstall
  and make it final
- add @SafeVarargs to
  Jackson2ObjectMapperFactoryBean#setModulesToInstall and make it final

This is a backwards incompatible change as these methods now have to be
declared final. Existing subclasses that override one of these methods
will break.

Closes gh-25293

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java
@@ -595,8 +595,8 @@ public class Jackson2ObjectMapperBuilder {
 	 * @see #modulesToInstall(Module...)
 	 * @see com.fasterxml.jackson.databind.Module
 	 */
-	@SuppressWarnings("unchecked")
-	public Jackson2ObjectMapperBuilder modulesToInstall(Class<? extends Module>... modules) {
+	@SafeVarargs
+	public final Jackson2ObjectMapperBuilder modulesToInstall(Class<? extends Module>... modules) {
 		this.moduleClasses = modules;
 		this.findWellKnownModules = true;
 		return this;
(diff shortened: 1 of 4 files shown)


sbrannen:
When running the build with JDK 9+, it fails with the following:

```
> Task :spring-web:compileJava FAILED
/Users/XXX/spring-framework/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java:601: warning: [varargs] Varargs method could cause heap pollution from non-reifiable varargs parameter modules
                this.moduleClasses = modules;
```

But I'll sort it out. 😉

sbrannen:

Thanks

marschall:
Thank you!

PR #25314, opened 2020-06-24
Wording changes

Buzzardo:
Replace potentially insensitive language with more neutral language.

Note that there are lots of code items (method names, for example) that
also need to change.


sbrannen:
Thanks for the PR!

We'll get this merged in for 5.2.x and consider backporting.

Do you have plans to make a pass through the Javadoc as well (perhaps in a follow up PR)?

Buzzardo:
I forgot about the Javadoc. Thanks for the reminder, Sam. I'll add it to my todo list.

jhoeller:
@sbrannen @Buzzardo While finding a neutral tone for such statements in the documentation is certainly a worthwhile goal, we need to pay attention that we're not making the javadocs less clear. What I mean here is in particular the term "execute" (as commonly handled in this PR); that term is very common in many core Java APIs. Our own use of the term is almost always connected to its use in core Java, so there isn't much we can do about it in our own code-level language either. All in all, arguably we should preserve the use of "execute" where it directly refers to API artifacts (which is in particular the case in javadocs), keeping the language in the javadoc in line with the language of the API itself.

PR #25319, opened 2020-06-26
Replace anonymous inner classes with lambda expressions

ghost:
In SimpleBufferingAsyncClientHttpRequest And SimpleStreamingAsyncClientHttpRequest class，Anonymous type can be replaced with lambda


sbrannen:
Although we typically don't seek to enhance deprecated classes, I've gone ahead and merged this PR into `master`.

Please note as well that your changes broke the build. So be sure to always run `./gradlew check` before submitting a PR.

Thanks

PR #25321, opened 2020-06-26
Improve Random seed in SocketUtils

jryan128:
I had an issue where `System.currentTimeMillis()` was a bad seed when a few tests were running in parallel. I want to be able to override the random to use `nanoTime()`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/SocketUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/SocketUtils.java
@@ -54,7 +54,7 @@ public class SocketUtils {
 	public static final int PORT_RANGE_MAX = 65535;
 
 
-	private static final Random random = new Random(System.currentTimeMillis());
+	private static Random random = new Random(System.currentTimeMillis());
 
 
 	/**
@@ -76,6 +76,13 @@ public class SocketUtils {
 	public SocketUtils() {
 	}
 
+	/**
+	 * Set custom random source.
+	 * Default uses {@code System.currentTimeMillis()} instead of {@code Random}'s default.
+	 */
+	public static void setRandom(Random random) {
+		SocketUtils.random = random;
+	}
 
 	/**
 	 * Find an available TCP port randomly selected from the range


pivotal-issuemaster:
@jryan128 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25321)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25321) to manually synchronize the status of this Pull Request.

spring-core/src/main/java/org/springframework/util/SocketUtils.java
```diff
@@ -76,6 +76,13 @@
 	public SocketUtils() {
 	}
 
+	/**
+	 * Set custom random source.
+	 * Default uses {@code System.currentTimeMillis()} instead of {@code Random}'s default.
+	 */
+	public static void setRandom(Random random) {
+		SocketUtils.random = random;
```
NielsUll, line 84: Add null validation:
		`Assert.notNull(random, "'random' must not be null")`;

NielsUll (commented):
Needs unit tests

NielsUll:
While I understand your problem, having mutable static state which affect behaviour of static methods is very dangerous from an architectural point of view. 

Any code could call `SocketUtils.setRandom(new FixedRandom(0.5))` and it would break the rest of the application in hard-to-find ways.

sbrannen:
Instead of making the `Random` instance configurable, we are rather considering instantiating it as follows.

```java
private static final Random random = new Random(System.nanoTime());
```

Would that suit your needs?

jryan128:
Yes

sbrannen:
Thanks for the PR.

As mentioned above, I implemented it slightly differently in 35582ded0e6619fa0d86fe1eb0182274456767b7 for `5.2.8` and `5.3 M2`.

jryan128:
Thank you so much!!

PR #25342, opened 2020-06-30
fixed #25263 and #25350

### LINKED ISSUE #25263: Overloaded @Bean method with name mismatch causes bean to be created twice (in case of ASM processing)
Reproduced in Spring 5.2.7.RELEASE (Spring Boot 2.3.1.RELEASE):

```java
@Configuration
public class SomeConfig {
    @Bean(name = "other")
    SomeOtherBean foo() {
        System.out.println("constructing SomeOtherBean");
        return new SomeOtherBean();
    }

    @Bean(name = "foo")
    SomeBean foo(@Qualifier("other") SomeOtherBean other) {
        System.out.println("constructing SomeBean");
        return new SomeBean(other);
    }
}
```

With the above configuration class, `SomeOtherBean` is constructed twice, and `SomeBean` is never constructed. Why? Because the two methods have the same name. If you rename the second method to `bar`, everything is constructed once, as expected. 
Admittedly, using the same method name is not a good idea. I just happened to have done it by accident.

lifejwang11:
I tried to fiexed [#25263](https://github.com/spring-projects/spring-framework/issues/25263),The test case is in **ImportedConfigurationClassEnhancementTests#checkOverloadedBean**,The test passed,It's in  the Spring-Context module,You can verify it,thanks.


rstoyanchev:
The referenced issues have been fixed.

PR #25349, opened 2020-07-01
Avoid resizing of fixed-size HashMap

quaff:
loadFactor should be 1.0 instead of default 0.75

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java
@@ -325,7 +325,7 @@ class CglibAopProxy implements AopProxy, Serializable {
 		if (isStatic && isFrozen) {
 			Method[] methods = rootClass.getMethods();
 			Callback[] fixedCallbacks = new Callback[methods.length];
-			this.fixedInterceptorMap = new HashMap<>(methods.length);
+			this.fixedInterceptorMap = new HashMap<>(methods.length, 1);
 
 			// TODO: small memory optimization here (can skip creation for methods with no advice)
 			for (int x = 0; x < methods.length; x++) {
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -792,7 +792,7 @@ public abstract class BeanUtils {
 			}
 
 			List<KParameter> parameters = kotlinConstructor.getParameters();
-			Map<KParameter, Object> argParameters = new HashMap<>(parameters.size());
+			Map<KParameter, Object> argParameters = new HashMap<>(parameters.size(), 1);
 			Assert.isTrue(args.length <= parameters.size(),
 					"Number of provided arguments should be less of equals than number of constructor parameters");
 			for (int i = 0 ; i < args.length ; i++) {
--- a/spring-beans/src/main/java/org/springframework/beans/factory/config/MapFactoryBean.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/config/MapFactoryBean.java
@@ -85,7 +85,7 @@ public class MapFactoryBean extends AbstractFactoryBean<Map<Object, Object>> {
 			result = BeanUtils.instantiateClass(this.targetMapClass);
 		}
 		else {
-			result = new LinkedHashMap<>(this.sourceMap.size());
+			result = new LinkedHashMap<>(this.sourceMap.size(), 1);
 		}
 		Class<?> keyType = null;
 		Class<?> valueType = null;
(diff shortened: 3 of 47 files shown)


jhoeller:
This is rather subtle, unfortunately. It would be ideal to specify the number of expected entries upfront, but given that there is no such API on `HashMap`, specifying the expected size as initial capacity is an acceptable trade-off for our scenarios.

We generally avoid using custom load factors since that factor affects not just resize steps but also lookup performance. The default load factor of 0.75 is also used in the `HashMap` copy constructor, and an effective load factor of 0.5 is applied by the new `Map.of` method (through its "expand factor" of 2 for the input length), supposedly for the same reason: a better trade-off in terms of lookup performance through generously distributed hash buckets with less chance for collisions.

Since a `HashMap` does not use the given capacity value as-is but rather upgrades it to the next power of two anyway, actual resize steps do not happen as often as expected, so in ~half the cases a load factor of 1 won't make a difference, and even in the other half it'll only trigger a single resize step. Ironically, whatever load factor it choses to use itself, once such a `HashMap` is copied through the copy constructor, it'll revert to a default load factor of 0.75 again.

We'd have to look at the typical distribution in such narrow-sized maps with load factor 1 in our scenarios: If the elements expose a well-formed hashcode and collisions are typically none and there are no lookup operations for non-existing elements and no additional elements are added, I suppose we could consider a load factor of 1. That said, there are rather strong recommendations against such a load factor in general... and a potential resize step isn't so bad.

quaff:
Those maps will always resize once before this commit, what about `new HashMap<>((int) (size / 0.75))`?

jhoeller:
Well, the maps resize ~half of the time on average since the upgrade to the next power of two covers some ground as well (e.g size 9 -> capacity 16 is sufficient but size 13 -> capacity 16 isn't in case of load factor .75). In any case, it's not ideal as-is, I agree.

Calculating against the load factor in every such place would help indeed but look somewhat convoluted. Maybe we'd have to put it into a common utility method, or even into convenient map factory methods on our `CollectionFactory` class. It's a shame that `HashMap` does not provide a corresponding constructor itself :-(

I'm not convinced that this makes enough difference in practice to be worth the extra "noise" at code level but we can consider doing a consistent revision for 5.3 here.

diguage:
Extract the creation of `HashMap` into a Builder may be a better solution. It can change the factor at the Builder for the global.

wind57:
@jhoeller here is my interaction with Stuart Marks on this subject :  https://stackoverflow.com/questions/52692803/hashmap-rehash-resize-capacity

I will simplify this in a few sentences if you do not have the time to read it entirely. You _can_ say to a helper method : "please give me a REAL capacity under which no rehash operations will occur". At the moment, this can be done via (for example):

```
	/**
	 * Returns a capacity (maximum number of entries) that this Map can hold before the inner table is resized.
	 *
	 * @param desiredSize how many entries will this map contain
	 * @param cls the class of the Map
	 * @return a capacity under which no rehash operations will occur for the given Map
	 */
	public static int capacity(int desiredSize, Class<? extends Map<?, ?>> cls) {
		// this is empty on purpose for the point of discussion
	}
```

and you could say : `HashMap<String, String> map = new HashMap(capacity(9))`. But this has subtle problems.

 - we have to treat `ConcurrentHashMap` differently, because it does the correct thing (with a small gotcha: https://bugs.openjdk.java.net/browse/JDK-8202422).

- the callers of this should be confused (I would). passing a `TreeMap` or `EnumMap` here, makes no sense as they do not 
have constructors that take a size in. But that is most probably nitpicking on my side.

- `HashMap/LinkedHashMap` make perfect sense and are good candidates.

- `IdentityHashMap` and `Hashtable` have weird inner defaults as far as I remember, that must be handled in the method body as separate scenarios, most probably.

- this entire method is going to be a false promise anyway (https://bugs.openjdk.java.net/browse/JDK-8211831 + the very first link I gave). Specifically: 

> ... "no rehash operations will ever occur".... Strictly speaking the above statement is now incorrect, so it should probably be adjusted to say something like "rehash operations will generally not occur.

While it can surely be done, this is not obvious, at least IMHO.

jhoeller:
I've introduced `CollectionUtils.newHashMap` and `newLinkedHashMap` methods that take an `expectedSize` argument and repurposed the existing `int` constructors of our own `LinkedCaseInsensitiveMap` and `LinkedMultiValueMap` to work in `expectedSize` style as well (since for those two we never exposed load factor arguments to begin with, it seems sensible to simply extend the semantics of those existing capacity constructors to the expected number of elements). To be committed soon.

wind57:
why not say in the documentation exactly like `java.util.HashMap` was supposed to, and will after the defect will be closed? :

> `expectedSize` the expected number of elements under which no rehash operations will generally occur

otherwise what is the point if that documentation? providing the size in case of a `HashXXX` structure is _all_ about rehashes, to me. Am I missing something?

Also what is wrong with a diamond operator in `return new HashMap<>(....)`  in `CollectionUtils`?

jhoeller:
Good catch about the diamond operator, that was an accident that sneaked into the commit. I'll fix this right away.

As for the documentation, we can certainly hint at no resize/rehash operations expected there. Generally speaking, I find referring to the expected size (as per the `Map.size()` method) and the expected number of elements pretty intuitive, matching a simplified user way of thinking about it, as opposed to referring to capacity, load factor or the occurrence of rehashes.

wind57:
I don't fully agree on the wording that is done, but is _far_ better now. Thank you and I will stop commenting here.

PR #25358, opened 2020-07-06
Support for customization of 404 response when RouterFunctionWebHandler finds no routes

ivd-git:
**Context**
Next to the default WebServer a service starts another server to handle admin requests on a different port. The admin webserver is connected to the `HttpHandler` created by `RouterFunctions.toHttpHandler`.

When the admin server processes a request for a path not mapped to any `RouterFunction` then the admin server responds with a 404 NotFound status code and an empty response body.

Currently it is not possible to have a different response body. 

Changing the `RouterFunctionWebHandler` to return a `Mono.error(new ResponseStatusException(HttpStatus.NOT_FOUND))` instead of `ServerResponse.notFound().build()`  allows clients to implement an exception handler that customizes the response body.

Default behavior does not change because `WebFluxResponseStatusExceptionHandler` is the default handler.


poutsma:
Merged in 2f557d9. 

Thank you for submitting a PR, and we are sorry it took this long to resolve.

PR #25367, opened 2020-07-07
Allow auto grow for entries without default constructor in SpEL

MartinKnopf:
Hi,

I wanted to add a `BigDecimal` to a list using SpEL. Auto grow only works for element types with a default constructor and since `BigDecimal` does not have a default constructor auto grow does not work with my list.

Here is an example:

```java
public class SpelTest {

    public List<BigDecimal> values;
    
    SpelExpressionParser parser;

    @Before
    public void setup() {
        values = new ArrayList<>();
        parser = new SpelExpressionParser(new SpelParserConfiguration(true, true));
    }

    @Test
    public void shouldChangeValue() {
        values.add(BigDecimal.ONE);

        parser.parseExpression("values[0]").setValue(this, "123.4");

        assertThat(values.get(0)).isEqualTo(BigDecimal.valueOf(123.4)); // passes
    }

    @Test
    public void shouldAddValue() {
        parser.parseExpression("values[0]").setValue(this, "123.4");

        assertThat(values.get(0)).isEqualTo(BigDecimal.valueOf(123.4)); // fails
    }
}
```
Changing the first entry passes but adding an entry fails with

```
Caused by: java.lang.NoSuchMethodException: java.math.BigDecimal.<init>()
    at java.base/java.lang.Class.getConstructor0(Class.java:3349)
    at java.base/java.lang.Class.getDeclaredConstructor(Class.java:2553)
    at org.springframework.util.ReflectionUtils.accessibleConstructor(ReflectionUtils.java:185)
    at org.springframework.expression.spel.ast.Indexer$CollectionIndexingValueRef.growCollectionIfNecessary(Indexer.java:715)
    ... 55 more
```

My solution inserts `null` when no default constructor can be found for the element type. Not sure if this is a good solution but at least it does not seem to break any tests. Please be nice :-)

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/ast/Indexer.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/ast/Indexer.java
@@ -728,7 +728,8 @@ public class Indexer extends SpelNodeImpl {
 		Constructor<?> getConstructor(Class<?> type) {
 			try {
 				return ReflectionUtils.accessibleConstructor(type);
-			} catch (Throwable ex) {
+			}
+			catch (Throwable ex) {
 				return null;
 			}
 		}
--- a/spring-expression/src/main/java/org/springframework/expression/spel/ast/Indexer.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/ast/Indexer.java
@@ -712,16 +712,24 @@ public class Indexer extends SpelNodeImpl {
 				}
 				TypeDescriptor elementType = this.collectionEntryDescriptor.getElementTypeDescriptor();
 				try {
-					Constructor<?> ctor = ReflectionUtils.accessibleConstructor(elementType.getType());
+					Constructor<?> ctor = getConstructor(elementType.getType());
 					int newElements = this.index - this.collection.size();
 					while (newElements >= 0) {
-						this.collection.add(ctor.newInstance());
+						this.collection.add(ctor == null ? null : ctor.newInstance());
 						newElements--;
 					}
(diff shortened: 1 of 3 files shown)


sbrannen:
@aclement, would you mind taking a quick look at this to assess the implementation within the context of SpEL?

aclement:
I think I'm ok with that change. Might be worth a line in the section in the docs around `Parser configuration` to mention you might get nulls in the array if it is expanded and there is no default ctor for that element type. As I recall SpEL compilation doesn't currently support collection growing so there is no impact on that mode with changes in this area.

sbrannen:
> I think I'm ok with that change. 

Thanks for the quick review, @aclement. Much appreciated!

> Might be worth a line in the section in the docs around `Parser configuration` to mention you might get nulls in the array if it is expanded and there is no default ctor for that element type. As I recall SpEL compilation doesn't currently support collection growing so there is no impact on that mode with changes in this area.

@Horsed, would you mind adding such documentation to the reference manual (`core-expressions.adoc`) and including that in this PR?

Horsed:
Hey,

I added a note to the reference manual plus one more unit test.

Have a good day!

sbrannen:
> I added a note to the reference manual plus one more unit test.

Thanks

> Have a good day!

You, too!

sbrannen:
This has been merged into `master` in 35c0ae7b0c304599b26769378873856e398d0d38 and slightly revised in b0570cd3a65d73ccf29558fa994975260a8651d3.

Thanks

PR #25368, opened 2020-07-08
Use correct link to MaxInMemorySize

izeye:
This PR fixes a broken link to the "MaxInMemorySize" section.

### DIFF
--- a/src/docs/asciidoc/web/webflux.adoc
+++ b/src/docs/asciidoc/web/webflux.adoc
@@ -835,7 +835,7 @@ exposes a `maxInMemorySize` property and if so the Javadoc will have details abo
 values. On the server side, `ServerCodecConfigurer` provides a single place from where to
 set all codecs, see <<webflux-config-message-codecs>>. On the client side, the limit for
 all codecs can be changed in
-<<web-reactive.adoc#webflux-client-builder-maxinmemorysize, WebClient.Builder>>.
+<<webflux-webclient.adoc#webflux-client-builder-maxinmemorysize, WebClient.Builder>>.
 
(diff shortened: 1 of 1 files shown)


sbrannen:
The currently published [HTML version of the reference manual](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html#webflux-codecs-limits) does not exhibit a broken link for this.

So, out of curiosity, where are you seeing the broken link? Just in the files on GitHub like [here](https://github.com/spring-projects/spring-framework/blob/5.2.x/src/docs/asciidoc/web/webflux.adoc#limits)?

izeye:
@sbrannen Sorry, I didn't check the reference docs. I just saw it on the source file. Feel free to close this if it's working intentionally somehow.

sbrannen:
> Sorry, I didn't check the reference docs. 

No problem.

> I just saw it on the source file.

I was just curious if the link was broken in some other published format.

> Feel free to close this if it's working intentionally somehow.

I think it's actually working by accident currently.

In fact, it only works because the document you edited is _included_ in `web-reactive.adoc`, which makes the link effectively local. But... it's definitely not "correct" as it is, since a refactoring of the documents could actually cause the link to be broken in the generated documentation. There are likely several other such "accidental" local links.

sbrannen:

Thanks

sbrannen:
Open mouth.... insert foot!

It actually only works the way it was before (by cross referencing based on the including document). Thus, I reverted the change in c972d861edee89c0ca97ddfe455d803f3d046ab9.

That's one of the nuances of Asciidoctor, but... it is what it is.

PR #25381, opened 2020-07-12
Use Math.min() in ExponentialBackOff

ghost:
Reports cases where the minimum or the maximum of two numbers can be calculated by means of Math.max() or Math.min() instead of doing it manually.

### DIFF
--- a/spring-core/src/main/java/org/springframework/asm/ByteVector.java
+++ b/spring-core/src/main/java/org/springframework/asm/ByteVector.java
@@ -354,7 +354,7 @@ public class ByteVector {
   private void enlarge(final int size) {
     int doubleCapacity = 2 * data.length;
     int minimalCapacity = length + size;
-    byte[] newData = new byte[doubleCapacity > minimalCapacity ? doubleCapacity : minimalCapacity];
+    byte[] newData = new byte[Math.max(doubleCapacity, minimalCapacity)];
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/asm/ByteVector.java
```diff
@@ -354,7 +354,7 @@ public ByteVector putByteArray(
   private void enlarge(final int size) {
     int doubleCapacity = 2 * data.length;
     int minimalCapacity = length + size;
-    byte[] newData = new byte[doubleCapacity > minimalCapacity ? doubleCapacity : minimalCapacity];
+    byte[] newData = new byte[Math.max(doubleCapacity, minimalCapacity)];
```
sbrannen, line 357: Please refrain from modifying classes under `org.springframework.asm`, `org.springframework.cglib`, and `org.springframework.objenesis`. Those include repackaged forks of the third-party libraries ASM, CGLIB, and Objenesis. Any refactoring to those classes should take place upstream in the originating repository. The Spring Framework will then pick up the changes when syncing with official updates of the forked third-party libraries.

----

Note, however, that there is no need to update this PR. I'll simply revert that before merging.

sbrannen:
This has been revised and merged into `master` in 490bdd11a562b665f1b8695d4b8f61f6bb7949a1.

Thanks

PR #25415, opened 2020-07-19
Simplified return statement

ghost:
return value if statements which can be simplified to a single assignment, return or assert statement.


poutsma:
Merged. Thanks for the PR!

PR #25429, opened 2020-07-20
Fix infinite loop bug in AnnotationScanner

yilianhuaixiao:
The original code
`isFiltered(source, context, classFilter)` when true,
 would cause  dead circle.
Maybe new code is right.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationsScanner.java
@@ -146,14 +146,14 @@ abstract class AnnotationsScanner {
 			Class<?> root = source;
 			while (source != null && source != Object.class && remaining > 0 &&
 					!hasPlainJavaAnnotationsOnly(source)) {
+				R result = processor.doWithAggregate(context, aggregateIndex);
+				if (result != null) {
+					return result;
+				}
 				if (isFiltered(source, context, classFilter)) {
 					source = source.getSuperclass();
 					aggregateIndex++;
 					continue;
-				}
-				R result = processor.doWithAggregate(context, aggregateIndex);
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@yilianhuaixiao Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25429)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25429) to manually synchronize the status of this Pull Request.

sbrannen:
Have you encountered an exception or some form of infinite loop in your code regarding your proposed change?

If so, please elaborate.

In any case, please add a test to this PR that fails before the proposed change and passes after the proposed change.

Thanks

yilianhuaixiao:
I add the testcase.
Before the bugfix, the testcase would cause infinite loop. After, the testcase could be passed.

yilianhuaixiao:
@sbrannen Thanks for your advice

sbrannen:
> I add the testcase.
> Before the bugfix, the testcase would cause infinite loop. After, the testcase could be passed.

Thanks!

We'll look into it.

sbrannen:
Tentatively slated for 5.3 M2 for investigation.

sbrannen:
This has been merged into `master` in 650cbeee1433aa38619691946c316d9d03a6512d and revised in 5442d8779aa6ae3d5495255853ef7be8a14ca6da.

Thanks!

----

@yilianhuaixiao, out of curiosity, since this code path is not executed in production code, how did you notice this bug?

sbrannen:
See also #25477

yilianhuaixiao:
@sbrannen  I am learning the spring-core  source codes rencently.

PR #25446, opened 2020-07-21
Use instanceof instead of Class#isInstance where feasible

XenoAmess:
as title


sbrannen:
This has been merged into `master` in c547809e89bb7da739519203551832f8e7ce35fc and revised in cdc234d7db97189e7255323921bdf84897795f17.

Thanks

PR #25449, opened 2020-07-21
Simplify if-statements with instanceof checks

XenoAmess:
as title

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/web/AbstractGenericWebContextLoader.java
+++ b/spring-test/src/main/java/org/springframework/test/context/web/AbstractGenericWebContextLoader.java
@@ -181,7 +181,7 @@ public abstract class AbstractGenericWebContextLoader extends AbstractContextLoa
 
 		// If the WebApplicationContext has no parent or the parent is not a WebApplicationContext,
 		// set the current context as the root WebApplicationContext:
-		if (parent == null || (!(parent instanceof WebApplicationContext))) {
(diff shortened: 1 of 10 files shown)


spring-test/src/main/java/org/springframework/test/context/web/AbstractGenericWebContextLoader.java
```diff
@@ -181,7 +181,7 @@ protected void configureWebResources(GenericWebApplicationContext context,
 
 		// If the WebApplicationContext has no parent or the parent is not a WebApplicationContext,
 		// set the current context as the root WebApplicationContext:
-		if (parent == null || (!(parent instanceof WebApplicationContext))) {
+		if ((!(parent instanceof WebApplicationContext))) {
```
sbrannen, line 184:
```suggestion
		if (!(parent instanceof WebApplicationContext)) {
```

sbrannen (changes requested):
Thanks for the PR.

I've requested a simple change.

spring-test/src/main/java/org/springframework/test/context/web/AbstractGenericWebContextLoader.java
XenoAmess: @sbrannen done. and commits squashed.

XenoAmess:
feedback provided above.

sbrannen (approved):

sbrannen:

Thanks

PR #25450, opened 2020-07-21
Use StringBuilder instead of StringBuffer in tests

XenoAmess:


spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
```diff
@@ -260,7 +260,7 @@ private static Class getClass(String className, ClassLoader loader, String[] pac
 		while ((index = className.indexOf("[]", index) + 1) > 0) {
 			dimensions++;
 		}
-		StringBuffer brackets = new StringBuffer(className.length() - dimensions);
+		StringBuilder brackets = new StringBuilder(className.length() - dimensions);
```
sbrannen, line 263: In light of the following policy, please revert this change.

----

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/cglib/core/ReflectUtils.java
XenoAmess: @sbrannen done.

sbrannen (approved):

sbrannen:

Thanks

PR #25452, opened 2020-07-21
Use `private static` instead of `static private` for method declarations

XenoAmess:
as title

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -910,7 +910,7 @@ public class MethodParameter {
 		 * Return the generic return type of the method, with support of suspending
 		 * functions via Kotlin reflection.
 		 */
-		static private Type getGenericReturnType(Method method) {
+		private static Type getGenericReturnType(Method method) {
 			try {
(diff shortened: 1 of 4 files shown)


sbrannen:

Thanks

izeye:
The title seems to be the opposite to what has been done.

sbrannen:
> The title seems to be the opposite to what has been done.

Indeed. That was my mistake when I reworded this PR's title incorrectly. I'll fix it now, but the associated commit message will unfortunately remain misleading.

XenoAmess:
> > The title seems to be the opposite to what has been done.
> 
> Indeed. That was my mistake when I reworded this PR's title incorrectly. I'll fix it now, but the associated commit message will unfortunately remain misleading.

@sbrannen @izeye 
Sorry for the mistake I made in commit message, and the delay of reply.

sbrannen:
> > > The title seems to be the opposite to what has been done.
> > 
> > 
> > Indeed. That was my mistake when I reworded this PR's title incorrectly. I'll fix it now, but the associated commit message will unfortunately remain misleading.
> 
> @sbrannen @izeye
> Sorry for the mistake I made in commit message, and the delay of reply.

No worries. It was **I** who made the mistake in the commit message anyway. 😉

PR #25455, opened 2020-07-21
use regionMatch to reduce subString creation in CodeFlow.isPrimitiveOrUnboxableSupportedNumber

XenoAmess:
as title.


sbrannen:
Thanks for the PR.

Upon further reflection, we have decided that we will keep that code block as-is -- even with the extra substring allocation -- in favor of readability and due to the fact that this is not a critical code path.

PR #25456, opened 2020-07-21
Use Integer.parseInt instead of Integer.valueOf for primitive int

XenoAmess:
as title.


sbrannen:

Thanks

PR #25459, opened 2020-07-22
Removed unused imports

MattMFX:
Removed Java.util.List and Java.util.stream.Collectors from TestCompiler.java


sbrannen:
Thanks for the PR, but I already addressed this in 2719dcb29b871ec24d6f10028dce76e440691d20.

PR #25477, opened 2020-07-25
Remove unused class filtering support in AnnotationScanner

sbrannen:
PR gh-25429 brought it to our attention that there was a bug in
AnnotationScanner when using a non-null class filter that filtered out
classes; however, it turns out that there is no production code that
utilizes the package-private class filtering support.

This commit therefore removes all class filtering support from
AnnotationScanner since that functionality was effectively unused.


sbrannen:
@philwebb and @jhoeller, I've requested your review of this PR.

I assume there was originally an intent to utilize the class filtering functionality in `AnnotationScanner`, but since this functionality is currently not used, I figured we should simply remove it altogether.

Thus, if you don't foresee any need to keep this functionality in place, I'll merge this PR for `5.2.x` and `master`.

PR #25478, opened 2020-07-27
Use Comparable instead of dedicated implementations

wind57:


pivotal-issuemaster:
@wind57 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25478)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25478) to manually synchronize the status of this Pull Request.

spring-core/src/main/java/org/springframework/util/comparator/BooleanComparator.java
```diff
@@ -63,8 +64,9 @@ public BooleanComparator(boolean trueLow) {
 
 
 	@Override
-	public int compare(Boolean v1, Boolean v2) {
-		return (v1 ^ v2) ? ((v1 ^ this.trueLow) ? 1 : -1) : 0;
```
wind57, line 67: some time ago I had an answer about this here : https://stackoverflow.com/questions/55598277/is-there-a-useful-difference-between-p-q-and-p-q-for-booleans/55601399#55601399 and it turns out that the performance of XOR vs plain `!=` is same. This also makes it far easier to read

spring-core/src/main/java/org/springframework/util/comparator/BooleanComparator.java
```diff
@@ -76,7 +78,7 @@ public boolean equals(@Nullable Object other) {
 
 	@Override
 	public int hashCode() {
-		return getClass().hashCode() * (this.trueLow ? -1 : 1);
```
wind57, line 79: since `java-8` there is a way to get the hashCode without calling into `getClass`

spring-core/src/test/java/org/springframework/util/comparator/BooleanComparatorTests.java
```diff
@@ -29,34 +29,35 @@
  * @author Keith Donald
  * @author Chris Beams
  * @author Phillip Webb
+ * @author Eugene Rabii
  */
 class BooleanComparatorTests {
```
wind57, line 35: tests have changed because the documentation of `Boolean::compare` does not guarantee `-1, 0, 1`; but it does less than 1, 0, greater than 1

wind57:
Will provide a broader PR, against the entire package.

snicoll:
@wind57 please do not do that. If you want to update your PR, you can push more changes to your `comparator-cleanup` branch and this will update this PR.

spring-core/src/main/java/org/springframework/util/comparator/Comparators.java
```diff
@@ -29,49 +29,44 @@
 
 	/**
 	 * Return a {@link Comparable} adapter.
```
wind57, line 31: I've used `Comparator::naturalOrder` and `Comparator::nullLast/nullFirst` instead of `NullSafeComparator`

wind57:
@snicoll thank you, I am still new, not sure about the process yet. re-opened.

snicoll:
@wind57 thanks very much for making your first contribution to the Spring Framework. I've polished your change and deprecated those two implementations as we can now replace them by the convenience method in the JDK.

PR #25483, opened 2020-07-28
AnnotationTypeMappings does not filter repeatable annotations

yilianhuaixiao:
since
`					if (!isMappable(source, metaAnnotation)) {
						continue;
					}`
is called before and 'source' ,'metaAnnotation' did not changed. 
The code should be deleted.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMappings.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMappings.java
@@ -91,6 +91,9 @@ final class AnnotationTypeMappings {
 					.findRepeatedAnnotations(metaAnnotation);
 			if (repeatedAnnotations != null) {
 				for (Annotation repeatedAnnotation : repeatedAnnotations) {
+					if (!isMappable(source, repeatedAnnotation)) {
+						continue;
+					}
 					addIfPossible(queue, source, repeatedAnnotation);
 				}
 			}
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMappings.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMappings.java
@@ -91,9 +91,6 @@ final class AnnotationTypeMappings {
 					.findRepeatedAnnotations(metaAnnotation);
 			if (repeatedAnnotations != null) {
 				for (Annotation repeatedAnnotation : repeatedAnnotations) {
-					if (!isMappable(source, metaAnnotation)) {
-						continue;
-					}
 					addIfPossible(queue, source, repeatedAnnotation);
 				}
 			}
--- a/spring-core/src/test/java/org/springframework/core/annotation/AnnotationTypeMappingsTests.java
+++ b/spring-core/src/test/java/org/springframework/core/annotation/AnnotationTypeMappingsTests.java
@@ -463,6 +463,14 @@ class AnnotationTypeMappingsTests {
 		assertThat(mapping.isEquivalentToDefaultValue(0, OutputStream.class, ReflectionUtils::invokeMethod)).isFalse();
 	}
 
+	@Test
+	void forAnnotationTypeWhenRepeatableMetaAnnotationIsFiltered() {
+		AnnotationTypeMappings mappings = AnnotationTypeMappings.forAnnotationType(WithRepeatedMetaAnnotations.class,
+			Repeating.class.getName()::equals);
+		assertThat(getAll(mappings)).flatExtracting(AnnotationTypeMapping::getAnnotationType)
+			.containsExactly(WithRepeatedMetaAnnotations.class);
+	}
+
 	private Method[] resolveMirrorSets(AnnotationTypeMapping mapping, Class<?> element,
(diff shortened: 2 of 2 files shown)


spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMappings.java
```diff
@@ -91,9 +91,6 @@ private void addMetaAnnotationsToQueue(Deque<AnnotationTypeMapping> queue, Annot
 					.findRepeatedAnnotations(metaAnnotation);
 			if (repeatedAnnotations != null) {
 				for (Annotation repeatedAnnotation : repeatedAnnotations) {
-					if (!isMappable(source, metaAnnotation)) {
```
sbrannen, line 94: I'm wondering if that was perhaps a copy-and-paste error and if it should rather be:

```java
if (!isMappable(source, repeatedAnnotation)) {
    continue;
}
```

@philwebb, what are your thoughts on the matter?

yilianhuaixiao:
@sbrannen  Yes, I think you are right

sbrannen:
@yilianhuaixiao with your latest commit that _fixes_ the issue, please introduce a test that fails before the fix and passes after the fix.

yilianhuaixiao:
@sbrannen  I added the testcase which can be passed after changing.

spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMappings.java
sbrannen: @philwebb, I'm assuming my above assumption is correct. (I know: a lot of assuming).

In light of that, we'll go forward with the change in this PR.

spring-core/src/test/java/org/springframework/core/annotation/AnnotationTypeMappingsTests.java
```diff
@@ -474,6 +474,16 @@ void isEquivalentToDefaultValueWhenNotMatchingReturnsFalse() {
 		return result;
 	}
 
+	@Test
+	void forAnnotationTypeWhenRepeatableMetaAnnotationFilterd() {
+		AnnotationTypeMappings mappings = AnnotationTypeMappings.forAnnotationType(WithRepeatedMetaAnnotations.class,
+		annotationType ->
+				ObjectUtils.nullSafeEquals(annotationType, Repeating.class.getName()));
```
sbrannen, line 481: You omitted the import for `ObjectUtils`, but don't worry about it. I'll add it before merging.

sbrannen: For future reference, I was able to simplify the annotation filter in that test method as follows.

```java
@Test
void forAnnotationTypeWhenRepeatableMetaAnnotationIsFiltered() {
	AnnotationTypeMappings mappings = AnnotationTypeMappings.forAnnotationType(WithRepeatedMetaAnnotations.class,
			Repeating.class.getName()::equals);
	assertThat(getAll(mappings)).flatExtracting(AnnotationTypeMapping::getAnnotationType)
			.containsExactly(WithRepeatedMetaAnnotations.class);
}
```

However, I won't be merging that change due to reasons I'll explain later.

sbrannen:
@yilianhuaixiao, thanks again for the PR and the test case!

It turns out that `AnnotationTypeMappings.forAnnotationType(Class<? extends Annotation>, AnnotationFilter)` is not invoked outside of the `AnnotationTypeMappings` class itself.

sbrannen:
@yilianhuaixiao, thanks again for the PR and the test case!

It turns out that `AnnotationTypeMappings.forAnnotationType(Class<? extends Annotation>, AnnotationFilter)` is not invoked outside of the `AnnotationTypeMappings` class itself.

It is only invoked from `AnnotationTypeMappings.forAnnotationType(Class<? extends Annotation>)`, with `AnnotationFilter.PLAIN` as the supplied annotation filter, which filters out annotations from the `java.lang` and `org.springframework.lang` packages.

Since none of the annotations from the `java.lang` and `org.springframework.lang` packages is _repeatable_, the filtering of repeatable annotations is unnecessary.

In light of that, I am changing this from a "bug" to a "task" issue in order to remove the method from the package-private API and to remove the filtering of repeatable annotations altogether.

sbrannen:
> It turns out that `AnnotationTypeMappings.forAnnotationType(Class<? extends Annotation>, AnnotationFilter)` is not invoked outside of the `AnnotationTypeMappings` class itself.

Actually, I made a mistake while analyzing this in my IDE. An `AnnotationFilter` can be supplied to `AnnotationTypeMappings` via the `from` factory methods in `TypeMappedAnnotations` via the `from` factory methods in `MergedAnnotations` which are part of the public API.

Thus, we will merge this PR as a bug fix.

spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMappings.java
philwebb: I think it was a typo. A little concerning that no test covered it.

sbrannen:
This has been merged into `5.2.x` and `master`.

Thanks

PR #25487, opened 2020-07-28
Updated as per standard design practice

abhishek-abhi:
As this class would be used in other Comparators, we should make this as singleton design pattern. I have just modified this class, so that it can be used in concurrent environments.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/comparator/ComparableComparator.java
+++ b/spring-core/src/main/java/org/springframework/util/comparator/ComparableComparator.java
@@ -39,7 +39,7 @@ public class ComparableComparator<T extends Comparable<T>> implements Comparator
 	
 	private ComparableComparator(){}
 	
-	public ComparableComparator getInstance(){
+	public static ComparableComparator getInstance(){
 		return INSTANCE;
 	}
 
--- a/spring-core/src/main/java/org/springframework/util/comparator/ComparableComparator.java
+++ b/spring-core/src/main/java/org/springframework/util/comparator/ComparableComparator.java
@@ -23,7 +23,7 @@ import java.util.Comparator;
  * Mainly for internal use in other Comparators, when supposed
  * to work on Comparables.
  *
- * @author Keith Donald
+ * @author Keith Donald , Abhishek Kumar Mandal
  * @since 1.2.2
  * @param <T> the type of comparable objects that may be compared by this comparator
  * @see Comparable
@@ -35,8 +35,13 @@ public class ComparableComparator<T extends Comparable<T>> implements Comparator
 	 * @see Comparators#comparable()
 	 */
 	@SuppressWarnings("rawtypes")
-	public static final ComparableComparator INSTANCE = new ComparableComparator();
-
+	private static final ComparableComparator INSTANCE = new ComparableComparator();
+	
+	private ComparableComparator(){}
+	
+	public ComparableComparator getInstance(){
+		return INSTANCE;
+	}
 
 	@Override
 	public int compare(T o1, T o2) {


spring-core/src/main/java/org/springframework/util/comparator/ComparableComparator.java
```diff
@@ -35,8 +35,13 @@
 	 * @see Comparators#comparable()
 	 */
 	@SuppressWarnings("rawtypes")
-	public static final ComparableComparator INSTANCE = new ComparableComparator();
-
+	private static final ComparableComparator INSTANCE = new ComparableComparator();
+	
+	private ComparableComparator(){}
+	
+	public ComparableComparator getInstance(){
```
wind57, line 42: have a look [https://github.com/spring-projects/spring-framework/pull/25478/files](here). This class should be deprecated and not longer be updated, at all.

abhishek-abhi: No, as this class is present in the repository, there is a valid reason to update this as per the latest design. If you feel , you can raise different PR if required.

spring-core/src/main/java/org/springframework/util/comparator/ComparableComparator.java
```diff
@@ -35,8 +35,13 @@
 	 * @see Comparators#comparable()
 	 */
 	@SuppressWarnings("rawtypes")
-	public static final ComparableComparator INSTANCE = new ComparableComparator();
-
+	private static final ComparableComparator INSTANCE = new ComparableComparator();
+	
+	private ComparableComparator(){}
+	
+	public static ComparableComparator getInstance(){
```
wind57, line 42: - can you exactly explain what _concurrent environments_ in your reasoning is supposed to mean? this class has no state, it is _already_ as safe as it can get. 

- adding a private constructor were a default was implied means you are breaking every caller that relied on such a constructor being present. You do understand that, right?

- what is the difference between having that `INSTANCE` being `public` and now in your changes accessed via a `getInstance`?  And of course the fact that you made it `private` breaks every caller that relied on it being `public`, without any benefit at all.

abhishek-abhi: This is the way how we define any utility kind of classes. Here all the methods are static. Default constructor and private constructor is completely different. private constructor means , if you really need to use this class you can only have to use by calling getInstance . well one more thing I want to mention that this class is having state, so the argument that you put is not valid.

wind57: > Default constructor and private constructor is completely different

every caller that could have done `new ComparableComparator...` until now, will break if these changes that you propose will be integrated; since you made the `ComparableComparator` constructor `private`. It does not matter how utility classes are defined or not in this context - your changes introduce a _breaking_ change.

> well one more thing I want to mention that this class implementation is not breaking anything, so the argument that you put is not valid

read above. Plus add the fact that `INSTANCE` _was_ `public` before your changes, so callers could have done `ComparableComparator.INSTANCE`; and now they will not. Think about this way - if these changes are integrated and some clients upgrade spring-core library, their code will not compile anymore.

> avoid extra object creation cost

_In general_ yes, but this is overall a myth _also_ that developers stop thinking about and take for granted. More details here for example : https://stackoverflow.com/questions/59718427/impact-of-java-streams-from-gc-perspective-or-handling-short-lived-objects-by-th/59745276#59745276

_In general_ yes; but in this case for such small and easy to create Objects this is a myth. 

Btw, I am not the maintainer or a committer in this library, I am just trying to understand your reasonings and so far I don't really see them.

abhishek-abhi: Thank you for your comments. I believe that I have already kept the required justification in my comments. In addition to that  this class is a helper class for other comparators and we can use the getInstance method for creation of object for this class, as there could be several other comparators that can use this class as an utility when they end up creating a brand new instance of this object with traditional default constructor. As we are not exposing this class for the user, we can definitely mark it as singleton for better design. Anyways let us stop the chain of comments and the actual maintainer or the person who is having the merge rights can decide on this ground with his own perspective with my proposed changes.

wind57: > As we are not exposing this class for the user

- this class _is_ `public`

- it is not hidden via the manifest file

- it is not hidden via the module system

- it _can't_ be hidden anymore since it was already exposed.

- if they _really_ wanted to hide it, it would have been `package-protected`, which it is not.

abhishek-abhi: If you see the comments kept by the creator of this class mentioned that this class is mainly for internal use. Thats why it is evident that no user would prefer to use this when there are other specific comparator present over there with justified reasons.

abhishek-abhi:
Hi, @jhoeller please review my PR, please give suggestions and merge this if its okay.

snicoll:
Thank you for the PR but there is nothing wrong with the current code. Changing the visibility of the default instance to `private ` is a breaking change. Please review [the comment I've added to your other PR](https://github.com/spring-projects/spring-framework/pull/25566#issuecomment-671021037).

PR #25489, opened 2020-07-29
specify null and zero length data

minuk8932:
String: length of zero and null value are not same.
but, this code return null when string has length of zero.

So, I think it need to be more specific.
if the parameter String(source) value is null -> then NullPointException is occured (automatically).
and if String length is not valid (that is length is not 1) -> then IllegalArgumentException ocurred by the code.

Please check and leave a comment. thank you.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/StringToCharacterConverter.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/StringToCharacterConverter.java
@@ -28,10 +28,7 @@ final class StringToCharacterConverter implements Converter<String, Character> {
 
 	@Override
 	public Character convert(String source) {
-		if (source.isEmpty()) {
-			return null;
-		}
-		if (source.length() > 1) {
+		if (source.length() != 1) {
 			throw new IllegalArgumentException(
 					"Can only convert a [String] with length of 1 to a [Character]; string value '" + source + "'  has length of " + source.length());
 		}


pivotal-issuemaster:
@minuk8932 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25489)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25489) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@minuk8932 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25489)!

spring-core/src/main/java/org/springframework/core/convert/support/StringToCharacterConverter.java
```diff
@@ -28,10 +28,7 @@
 
 	@Override
 	public Character convert(String source) {
-		if (source.isEmpty()) {
```
wind57, line 31: as far as I understand this was here because if there is an empty String, you can't represent an "empty char" (there is not such thing), thus a `null` probably makes the most sense here. I wonder if there is code that actually relies on this. 

what I mean is that someone might rely on such an un-documented specification like this:

```
StringToCharacterConverter converter = ....;
Character c = converter.convert("");
if(c == null){
     // assume "empty" and do something
}
```

the way you refactored means that every client is now broken, since you throw an `Exception`...

does this make sense? btw, I don't know what is the correct way, just adding my thoughts in here

minuk8932: Thank you for your review.

I think someone would perceive null as just null, not an empty string. (I think, it is right.)
And also, the code you're talking about is a code to handle null strings, not an empty string.
Therefore, I thought it was correct to explicitly specify whether it is an empty string or null, and then intend to cope with it later. (when `Exception` throwed)

On the other hand, someone can be as intended as you say. And I sympathize with ' every client is now broken, since you throw an Exception...'.

In conclusion, this is also just my opinion. Thank you for a nice opinion. :)

PR #25491, opened 2020-07-30
Add MediaType.APPLICATION_JSON_SEQ for application/json-seq

quaff:
[JSON Text Sequences](https://tools.ietf.org/html/rfc7464) should be added to MediaType like NDJSON, jackson will supports it later, please see https://github.com/FasterXML/jackson-core/issues/633


bclozel:
Closing this as FasterXML/jackson-core#633 is not supported and NDJSON seems to be enough for now.
Thanks!

PR #25506, opened 2020-08-02
Optimize DefaultLifecycleProcessor::startBeans

chenqimiao:
Use lambda for minor refactoring

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/DefaultLifecycleProcessor.java
+++ b/spring-context/src/main/java/org/springframework/context/support/DefaultLifecycleProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-context/src/main/java/org/springframework/context/support/DefaultLifecycleProcessor.java
+++ b/spring-context/src/main/java/org/springframework/context/support/DefaultLifecycleProcessor.java
@@ -143,11 +143,8 @@ public class DefaultLifecycleProcessor implements LifecycleProcessor, BeanFactor
 		lifecycleBeans.forEach((beanName, bean) -> {
 			if (!autoStartupOnly || (bean instanceof SmartLifecycle && ((SmartLifecycle) bean).isAutoStartup())) {
 				int phase = getPhase(bean);
-				LifecycleGroup group = phases.get(phase);
-				if (group == null) {
-					group = new LifecycleGroup(phase, this.timeoutPerShutdownPhase, lifecycleBeans, autoStartupOnly);
-					phases.put(phase, group);
-				}
+				LifecycleGroup group = phases.computeIfAbsent(phase,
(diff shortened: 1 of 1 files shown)


spring-context/src/main/java/org/springframework/context/support/DefaultLifecycleProcessor.java
```diff
@@ -143,11 +143,8 @@ private void startBeans(boolean autoStartupOnly) {
 		lifecycleBeans.forEach((beanName, bean) -> {
 			if (!autoStartupOnly || (bean instanceof SmartLifecycle && ((SmartLifecycle) bean).isAutoStartup())) {
 				int phase = getPhase(bean);
-				LifecycleGroup group = phases.get(phase);
-				if (group == null) {
-					group = new LifecycleGroup(phase, this.timeoutPerShutdownPhase, lifecycleBeans, autoStartupOnly);
-					phases.put(phase, group);
-				}
+				LifecycleGroup group = phases.computeIfAbsent(phase,
+						p -> new LifecycleGroup(phase, this.timeoutPerShutdownPhase, lifecycleBeans, autoStartupOnly));
```
wind57, line 147: documentation of `computeIfAbsent` says:

>  return the current (**existing or computed**) value...

so you could do:

```
phases.computeIfAbsent(
						phase,
						p -> new LifecycleGroup(phase, timeoutPerShutdownPhase, lifecycleBeans, autoStartupOnly)
				).add(beanName, bean);
```

But looking at the implementation, `phases` is really just a sorted by Keys, which mean it's really supposed to be a `TreeMap`? So that code could be written as :

```
	private void startBeans(boolean autoStartupOnly) {
		Map<String, Lifecycle> lifecycleBeans = getLifecycleBeans();
		Map<Integer, LifecycleGroup> phases = new TreeMap<>();

		lifecycleBeans.forEach((beanName, bean) -> {
			if (!autoStartupOnly || (bean instanceof SmartLifecycle && ((SmartLifecycle) bean).isAutoStartup())) {

				int phase = getPhase(bean);
				phases.computeIfAbsent(
						phase,
						p -> new LifecycleGroup(phase, timeoutPerShutdownPhase, lifecycleBeans, autoStartupOnly)
				).add(beanName, bean);
			}
		});

		phases.forEach((key, value) -> value.start());
	}
```

The use of `forEach` is safe since it is documented as :

>  Unless otherwise specified by the implementing class, actions are performed in **the order of entry set** iteration

and `TreeMap::entrySet` is documented as :

> The set's iterator returns the entries in ascending key order...

What do you think?

chenqimiao: I think what you said makes sense, but I added a judgment to reduce the automatic code generation of the iterator.
Thanks !

chenqimiao: I think what you said makes sense, but I added a judgment to reduce the execution of the iterator automatic code.
Thanks !

wind57: I'm sorry, but can you explain "but I added a judgment to reduce the execution of the iterator automatic code" - what this means? I don't get it. thank you

sbrannen:
This has been merged into `master` and slightly revised in 1f35cc5cae5027a4cc78a2e5e30ce89f8418b607.

Thanks

PR #25507, opened 2020-08-03
CloseStatus implements Serializable

candrews:
I wanted to use `CloseStatus` in another serializable class and noticed that it wasn't serializable.

There's no harm in making it serializable... so why not :)

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/CloseStatus.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/CloseStatus.java
@@ -16,6 +16,8 @@
 
 package org.springframework.web.socket;
 
+import java.io.Serializable;
+
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
@@ -30,7 +32,9 @@ import org.springframework.util.ObjectUtils;
  * @author Rossen Stoyanchev
  * @since 4.0
  */
-public final class CloseStatus {
+public final class CloseStatus implements Serializable {
+
+	private static final long serialVersionUID = 5199057709285570947L;
 
 	/**
 	 * "1000 indicates a normal closure, meaning that the purpose for which the connection


spring-websocket/src/main/java/org/springframework/web/socket/CloseStatus.java
```diff
@@ -30,7 +32,9 @@
  * @author Rossen Stoyanchev
  * @since 4.0
  */
-public final class CloseStatus {
+public final class CloseStatus implements Serializable {
+
+	private static final long serialVersionUID = 5199057709285570947L;
```
wind57, line 37: you don't have to provide one btw, though this does give you greater control. 

A few points here:

 - It's easy to add `implements Serializable` and move one, yes. The problem is that if you do that (without adding custom code for serialization) means that you have just "exposed" all it's internals and your future changes are almost impossible. It's like exposing an `API`. Look for example what `HashMap` does when it serializes and deserializes : it saves only "raw data", so that in future it could deserialize to any implementation it would feel like, without breaking expectations of clients.

- Then the is the problem that during deserialization no constructor is called, which might cause problems too. So for example invariants like :

```
Assert.isTrue((code >= 1000 && code < 5000), "Invalid status code");
```

from within this class constructor must be carried over to the deserialization code also. 

- You need (usually) to maintain good unit tests to ensure that this process of serialization/deserialization works across versions; this adds complexity to the code base.

I am sure that if you go to `Effective Java` you will find a lot more reasons (or many more details) for not using `Serializable`. 

The only good thing here is that this class is `final`, which relaxes many of the "bad things" that can happen during serialization. That said, I don't really know what a) Spring committers and maintainers think about this subject in general (and I am neither of these) b) why do you need to serialize something using java's default way?

I hope you do not mind me putting so many works it for such a small PR :)

rstoyanchev:
It is a simple holder of 2 fields and that isn't going to change.

PR #25508, opened 2020-08-03
Optimize `AggregatesSpliterator` to avoid repetitive work

yilianhuaixiao:
change 

```java
if (mergedAnnotation == null) {
    return tryAdvance(aggregate, action);
}
```

to

```java
if (mergedAnnotation == null) {
    return tryAdvance(aggregate, action, lowestDistance, annotationResult);
}
```

to avoid repetitive visiting to the former code.

Testcase passed.


sbrannen:
Hi @yilianhuaixiao,

Thanks for the PR.

Out of curiosity, what prompted you to make these changes?

Did you notice a performance issue via benchmarking?

snicoll:
Closing due to the lack of feedback.

PR #25528, opened 2020-08-04
PingPongWebSocketHandlerDecorator schedules pings and replies to pongs

candrews:
I've found it to be common for WebSockets clients to be required to issue pings and reply to pongs, or if not required to do so, want to do so in order to detect disconnects. In the interest of making this common chore easier for everyone to do, it would be nice for Spring to include `PingPongWebSocketHandlerDecorator` making it easy to implement that functionality.

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/handler/PingPongWebSocketHandlerDecorator.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/handler/PingPongWebSocketHandlerDecorator.java
@@ -1,0 +1,92 @@
+package org.springframework.web.socket.handler;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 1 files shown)


candrews:
@rstoyanchev can you please take a look at this PR?

rstoyanchev:
@candrews this is for a heartbeat mechanism? It can work fine for an application but I'm afraid it is a bit incomplete to promote at the framework level where it would likely need to be extended over time to address more needs. 

Heartbeats are usually supported at a higher level, e.g. STOMP or RSocket, where it is built into the protocol, it is bi-directional, and allows each side to know how frequently the other side wants to receive. This mutual coordination is important because each side then knows how frequently it needs to write, and when to decide the other side is no longer connected. 

In such implementations, heartbeats are avoided if the application is already sending and receiving messages at the desired frequency, thus avoiding additional traffic and the possibility for a pong to get jammed behind others. An implementation could also be improved to use a single task that wakes up as frequently as necessary (depending on the shortest heartbeat duration) to check on all sessions).

While the implementation can be improved, I think such a mechanism is better at a higher protocol level.

PR #25537, opened 2020-08-06
Fix comment syntax in xml examples

diguage:


snicoll (approved):

snicoll:
Good catch @diguage, thank you.

PR #25538, opened 2020-08-06
Use lowercase property names in SpEL examples

quaff:
According to java code conventions

### DIFF
--- a/src/docs/asciidoc/core/core-expressions.adoc
+++ b/src/docs/asciidoc/core/core-expressions.adoc
@@ -835,12 +835,12 @@ shows:
 	// Members List
 
 	// evaluates to "Nikola Tesla"
-	String name = parser.parseExpression("Members[0].Name").getValue(
+	String name = parser.parseExpression("members[0].name").getValue(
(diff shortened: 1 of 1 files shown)


sbrannen:
For the record, the proposed changes are to SpEL "property names", not variable names.

As stated in the same chapter:

> Case insensitivity is allowed for the first letter of property names.

Thus, `person.name`, `person.Name`, `Person.Name`, and `person.getName()` are all equivalent with regard to property access in SpEL.

It's ultimately a matter of taste. Some people prefer to follow Java conventions. Others prefer capitalized property names in expressions for improved readability.

In any case, I do think it would be beneficial to focus on Java coding conventions in the examples, with a `NOTE` pointing out more explicitly what I've stated above.

In light of that, I'll probably merge this PR as-is and add such a note.

sbrannen:
This has been merged into `master` in 9c11887fe8d119efb91403975c17a3bb231735b4 and revised in b01adadf6005867216858d7e2404d416f25faa63.

Thanks

PR #25540, opened 2020-08-06
Add necessary semicolons

diguage:

PR #25554, opened 2020-08-06
Use Groovy syntax highlighting in reference manual

diguage:
Change the language name of the source block for highlight.

This is an improved version of  #25540.


sbrannen:

Thanks

PR #25556, opened 2020-08-07
Let the DataSize.toString method returns the closest unit to represen…

ijliym:
…t the data size

So I know I configured the max upload file size is `4MB`, not `4194304B` when I print the log message.

### DIFF
--- a/spring-core/src/test/java/org/springframework/util/unit/DataSizeTests.java
+++ b/spring-core/src/test/java/org/springframework/util/unit/DataSizeTests.java
@@ -200,12 +200,32 @@ class DataSizeTests {
 
 	@Test
 	void toStringUsesBytes() {
-		assertThat(DataSize.ofKilobytes(1).toString()).isEqualTo("1024B");
+		assertThat(DataSize.ofBytes(1023).toString()).isEqualTo("1023B");
+	}
+
+	@Test
+	void toStringUsesKilobytes() {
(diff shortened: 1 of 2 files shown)


quaff:
You shouldn't override default implementation, you could add a new method.

snicoll (changes requested):
That's an interesting idea. Please update your PR to include tests that cover the various scenarios. We can figure out if we want move to a dedicated method or update the existing `toString` representation.

spring-core/src/test/java/org/springframework/util/unit/DataSizeTests.java
```diff
@@ -200,12 +200,32 @@ void isNegativeWithNegative() {
 
 	@Test
 	void toStringUsesBytes() {
-		assertThat(DataSize.ofKilobytes(1).toString()).isEqualTo("1024B");
+		assertThat(DataSize.ofBytes(1023).toString()).isEqualTo("1023B");
+	}
+
+	@Test
+	void toStringUsesKilobytes() {
+		assertThat(DataSize.ofKilobytes(1).toString()).isEqualTo("1KB");
```
snicoll, line 208: Rather than only testing the happy path, I'd like to see some tests where we can't fallback to the optimisation.

snicoll (changes requested):

snicoll:
There hasn't been much traction on this one and no feedback for over a year so I am going to close this one.

Issue #25323, opened 2020-06-27
Need to know transaction.isReadonly while getting connection - spring-jdbc


johnny2002:
Backgroup:  I have a large project with more than 500 nodes need to connection to one Database,  so connection count is too high. In order to reduce db connections, I want to reuse one connection for all readonly transactions. So, I need to get the "TransactionSynchronizationManager.isCurrentTransactionReadOnly()" in Datasource.getConnection() method. While, in spring-jdbc-5.2.2, DataSourceTransactionManager.doBegin method, the connection is gotten at line 263, the readonly flag is setted at line 298.

Suggestion: Set the readonly flag before getting connection. e.g.  in DataSourceTransactionManager.doBegin method

> protected void doBegin(Object transaction, TransactionDefinition definition) {
>     **TransactionSynchronizationManager.setCurrentTransactionReadOnly(definition.isReadOnly());**
>     DataSourceTransactionObject txObject = (DataSourceTransactionObject) transaction;
>     ...
>     **Connection newCon = obtainDataSource().getConnection();**
>     ...

Then I will override a Datasource, that, return one shared connection to all readonly transactions.

quaff:
Is it safe to reuse connection across threads?

johnny2002:
> Is it safe to reuse connection across threads?

for readonly operations, it is.

quaff:
> > Is it safe to reuse connection across threads?
> 
> for readonly operations, it is.

It depends on driver implementation.

johnny2002:
> > > Is it safe to reuse connection across threads?
> > 
> > 
> > for readonly operations, it is.
> 
> It depends on driver implementation.

At least, as I tested, oracle, mysql, pgsql are no problem.

johnny2002:
further more, no matter share connection is safe or not, let developers can get the readonly properties as earlier as possible is a good manner

johnny2002:
> > > Is it safe to reuse connection across threads?
> > 
> > 
> > for readonly operations, it is.
> 
> It depends on driver implementation.

So, what's the decision? will you adopt the change request?

eitan613:
Isn't the readonly flag set at line 277?

eitan613:
@johnny2002  I am trying to write a test method after adding your extra line of code.  But I can't seem to get doBegin to work properly.  It blows up no matter what I pass in as the transaction object. any ideas?

snicoll:
The PR has been declined, reopening to reconsider this.

[...]

PR #25558, opened 2020-08-07
fix for issue #25323 with test

eitan613:


pivotal-issuemaster:
@eitan613 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25558)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25558) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@eitan613 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25558)!

spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
```diff
@@ -49,6 +49,18 @@
...
+		URI uri = new URI("http://example.com/some/path?query=[fromto]");
+		URI finalUri = uri;
+		assertThatIllegalArgumentException().isThrownBy(() -> UriComponentsBuilder.fromUri(finalUri).build(true));
+		//ok, then encode it
+		uri = UriComponentsBuilder.fromUri(uri).build(false).encode().toUri();
+		//confirm encoded
+		URI checkUri = UriComponentsBuilder.fromUri(uri).build(true).toUri();
+		assertThat(uri.toString()).isEqualTo(checkUri.toString());
+	}
+
```
rstoyanchev, line 63: Looks like you accidentally included this, which I recognize as a scenario from an unrelated issue.

spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
```diff
@@ -440,7 +440,7 @@ public UriComponents build() {
 	/**
 	 * Build a {@code UriComponents} instance from the various components
 	 * contained in this builder.
-	 * @param encoded whether all the components set in this builder are
+	 * @param encoded asserts whether all the components set in this builder are already
```
rstoyanchev, line 443: Likewise here, this looks accidentally included here.

spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
eitan613: The purpose of this was to check that even though it threw an exception before it works now and is encoded as it should be. I can remove it though if it's unnecessary...

spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
eitan613: this was on purpose to prevent confusion by people reading the javadoc. Looking at issue #25323 and just different forums for questions answers there seems to be confusion about the purpose of the method. People think it's supposed to be encoding. This is to clarify it doesn't encode and the boolean passed in is to flag if it is already encoded

eitan613: Yes, I accidentally included two separate fixes in one PR. Will submit it in a new one.

eitan613:
will re-open as separate pull requests. sorry for the confusion

Issue #22845, opened 2019-04-27
CrudRepository add support for Optional


amalic:
Currently I use CrudRepository this way
```java
@Repository
public interface StuffRepository extends CrudRepository<Stuff, Long> { 
    public Stuff findStuffByOtherId(String otherId);
}
```
would be great if this were possible
```java
@Repository
public interface StuffRepository extends CrudRepository<Stuff, Long> { 
    public Optional<Stuff> findStuffByOtherId(String otherId);
}
```

This would enable more compact code like this
```java
@Autowired StuffRepository stuffRepository;

public Stuff getOrCreateStuff(String otherId) {
  return stuffRepository.findByOtherId(otherId)
      .orElse(stuffRepository.save(new Stuff(otherId)));
}
```

snicoll:
This is the issue tracker for spring framework and spring data requests are not handled here. As far as I am aware this is possible since the Kay release train.

PR #25572, opened 2020-08-10
Call StartupStep.end in finally block

mdeinum:
Prior to this commit it was possible that a StartupStep was
started but never ended. This was the case when an exception
occured during bean initializing. To always call the method
regardless of the outcome, the call to StartupStep.end has
been moved to a finally block.

When an exception occurs the StartupStep is also enriched with
the exception class and message for diagnostic purposes.

Related: spring-projects/spring-boot#22776


mdeinum:
@bclozel this fixes the missing end event detected in #22845.

mdeinum:
Maybe more events need to be in a `try/finally` approach due to possible errors being generated? I didn't investigate all other events (as this was the one that was missing).

bclozel:
Thanks @mdeinum , I totally missed that one; I'll take another look at other events.

PR #25584, opened 2020-08-14
FileUrlResource isWritable method returns true if URL protocol is not indicating a file

yilianhuaixiao:
if add testcase
`	@Test
	void test() throws IOException {
		Resource resource = new FileUrlResource(new URL("https://spring.io/"));
		assertThat(resource.isReadable()).isFalse();
	}
`
the former code cannot pass.
Then I restricted   FileUrlResource must be contructed with a URL about file.


jhoeller:
I've addressed this somewhat differently, exclusively relying on `getFile()` resolution (which could be overridden in subclasses) and not checking the URL protocol in the `isWritable()` implementation at all anymore.

Thanks for spotting this - and thanks for the pull request, in any case!

PR #25594, opened 2020-08-15
Optimize ObjectUtils.nullSafeHashCode() implementations

wind57:


wind57:
this is mainly a continuation of [this PR](https://github.com/spring-projects/spring-framework/pull/25591). 

 - All `nullSafeHashCode` are deprecated in favor of `Arrays::hashCode`. The only difference between existing methods and the ones from `java.util.Arrays` is `INITIAL_HASH`; which in case of hashCode makes no difference. As such, why keep the implementation at all? Deprecate and delegate.

 - The look-up is now done via a `Map`, instead of nested `if/else` statements, just like in the previous PR. All the same reasons apply (as to why that `innerWildcardCapture` exists). I have introduce one more type parameter, so that the previous PR can re-use the same method. One needs `Function<?, String>` and this PR needs `Function<?, Integer>`, so adding one more type parameter, will make both happy.

- I have also changed the tests; I find it weird to test hashCode relying on implementation details. hashCode/equals should only test the _contract_ of the methods.

snicoll:
Closing for the same reason exposed in https://github.com/spring-projects/spring-framework/pull/25591#issuecomment-1693326786. Thanks for the PR, in any case!

PR #25604, opened 2020-08-17
Prevent confusion about build(boolean) method in UriComponentsBuilder

eitan613:
To prevent confusion about the build(boolean) method


rstoyanchev:
@eitan613 I'm fine to make some Javadoc improvements but I'm not sure if the proposed change makes much of a difference. Either way the `encoded` flag states what it means. The fact that this will be validated/asserted doesn't aid in understanding.

I was thinking rather that the `UriComponentsBuilder.fromUri(URI)` method should state that you must then use `.build(true)` and reversely on `.build(true)` point out that if the builder was created from a `URI` the flag should be `true`. 

Or otherwise please clarify what confused you so we can find a way to improve.

rstoyanchev:
Okay I see. I think it already says that but we can certainly strengthen the language used. It does sound more like the confusion might be from looking at the method and assuming what it means.

eitan613:
Anything else I can/should do then?

PR #25605, opened 2020-08-17
Expose readOnly information on transaction

eitan613:


snicoll:
Thanks for the PR but that's not the right way of fixing the issue. I'll reopen the related issue and we'll give it another go.

PR #25621, opened 2020-08-21
SocketUtils ServerSocket、DatagramSocket use try-with-resource.

ChunMengLu:
SocketUtils ServerSocket、DatagramSocket use try-with-resource.


jhoeller:
On review, I prefer the explicitness of the `close()` calls in the current code. Since there is no actual resource operation to "try", I'd rather keep the immediate open-close steps explicit... instead of hiding the close step in the try-with-resources statement.

Issue #25623, opened 2020-08-21
PathMatchingResourcePatternResolver  gives me a FileSystemResource instead of ClassPathResource while I specify the prefix "classpath:" 


justmehyp:
Recently, I had some trouble loading classpath resources. During local development, I can obtain the resources under classpath
through classloader. And If this resource is a folder, I can get sub-files/ sub-directories through the `File#listfiles` method.
But this doesn't work when the program is packaged in a jar package, the file in a jar is not a file, and the folder is not a folder.
In this case, they are all `JarEntry`.

So I found out the PathMatchingResourcePatternResolver, which I hoped would solve my problem, but it couldn't.

The problem is described as follows:

First, we create a new PathMatchingResourcePatternResolver instance:
```
package com.xxx;

public class Main {
  public static void main(String[] args) {
    ResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
  }
}
```

I want to get the resources under classpath, such as all sub-resources under 'org/springframework/core/io/support/',
```
resolver.getResources("classpath:org/springframework/core/io/support/*")
```

I specified the prefix "classpath:" just as I wanted to get resources under classpath.
The statement returns an Array with component type `ClassPathResource`. That is exactly what I want. And the class name
`ClassPathResource` is just so NICE.

OK, We just tried to get the resources inside a jar.
Next, We are going to get something in expanded directory(file system). My program has a package called` com.xxx `. Next, let's try to obtain the resources under "com/xxx":
```
resolver.getResources("classpath:com/xxx");
```

This time, take a closer look. It returns an Array with concrete component type `FileSystemResource`. I am confused now.
I specified the prefix "classpath:", instead of "file:///", so why give me a FileSystemResource?

I am expecting the perfect `ClassPathResource` because I need its `getPath` method to do other things.

Finally, I am thinking about that this may be a defect in implementation, and should be resolve. What do you guys think?

Martin-Wegner:
Since 5.3.0 my package scanner is broken, which uses the ResourcePatternResolver to find classpath resources inside a JAR. I think this is related to this issue.

Martin-Wegner:
@rstoyanchev why do you closed this issue? The bug is still present in Spring 5.3.13 :(

rstoyanchev:
@Martin-Wegner it is closed as superseded by PR#25624. If both an issue and a PR are opened, we keep only one open. They are linked and there is no need to have both.

Martin-Wegner:
[...]

PR #25624, opened 2020-08-21
Consistently resolve `ClassPathResource` for `classpath:` prefix in `PathMatchingResourcePatternResolver`

justmehyp:
- addresses #25623

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java
+++ b/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java
@@ -41,7 +41,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.core.io.DefaultResourceLoader;
-import org.springframework.core.io.FileSystemResource;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.core.io.UrlResource;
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java
```diff
@@ -712,31 +711,58 @@ protected JarFile getJarFile(String jarFileUrl) throws IOException {
...
+		return filePath.substring(rootPath.length());
+	}
+
+	/**
+	 * Get the given file's absolute path. If the file is a directory then a slash is appended.
+	 * @param file the file to get absolute path
+	 * @return the absolute path
+	 */
+	private String getAbsolutePath(File file) {
+		String absolutePath = StringUtils.replace(file.getAbsolutePath(), File.separator, "/");
```
RemcoBuddelmeijer, line 759: Why do you want to replace the `File.separator`, which is system specific, with a `"/"`? Even though it _should_ be safe to use because it _should_ be translated by any libraries to a platform-specific file separator.

Nikunjksanghavi:
@justmehyp, @rstoyanchev, @NLRemco  When this issue is fixed, it's been opened for around 2+ years.

sbrannen:
`PathMatchingResourcePatternResolver` has undergone massive changes in Spring Framework 6.

@justmehyp, is this issue still relevant?

justmehyp:
Since I'm not working on it now,  I haven't tested the new version yet. 
It was a bug for me,  but as you can see,  this issue is been opened for around 2+ years, looks like other people don't think so.
I mean, Spring Framework has undergone significant changes, and if this problem still persists, you may need to rewrite the repair code based on the new version Spring.
Let me turn this issue off first.

Issue #25625, opened 2020-08-21
UriBuilder replaceQueryParam suggests a way to handle nulls.


kingbbode:
I expect replaceQueryParam to skip null values.

```
UriComponentsBuilder.newInstance()
                .queryParam("baz", "qux", 42)
                .replaceQueryParam("baz", (Object) null)
               .build()
```

expected query is null.

```
UriComponentsBuilder.newInstance()
                .queryParam("baz", "qux", 42)
                .replaceQueryParam("baz", null, 24)
               .build()
```

expected query is "baz=24"

Tell me what you think. Thanks.

rstoyanchev:
It is possible to have query parameters without a value, so `baz&baz=&baz=42` so we can't do that. You'll need to clean null values 
from the List or check the single value for null. At best we can clarify in the Javadoc that `null` is a valid value which results in a query parameter without a value.

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

rstoyanchev:
Note also that in 5.3 there is a new `queryParamIfPresent(Optional)` method.

PR #25626, opened 2020-08-21
fix from #25625.

kingbbode:
see #25625


rstoyanchev:
Closing this for now. See comments under #25625.

PR #25631, opened 2020-08-23
Simplify code in AbstractFallbackJCacheOperationSource

EruDev:
I found that `operation` has been judged not to be null before, so it can be returned directly


spring-context-support/src/main/java/org/springframework/cache/jcache/interceptor/AbstractFallbackJCacheOperationSource.java
```diff
@@ -96,9 +96,7 @@
 		if (specificMethod != method) {
 			// Fallback is to look at the original method.
 			operation = findCacheOperation(method, targetClass);
```
woniper, line 98: `return findCacheOperation(method, targetClass);` How about this code?

sbrannen:
Thanks for submitting the PR!

Although the proposal does simplify the code, it breaks from the style of the rest of the method. In such circumstances we prefer to retain consistent style rather than aim for 100% code simplification.

In light of that I am closing this PR.

PR #25635, opened 2020-08-24
Remove never use code in AbstractNestablePropertyAccessor

yilianhuaixiao:
'if (propValue == null)' ,' isAutoGrowNestedPaths()' always returns false.  So remove the never accessable code, direct throw exception.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
@@ -397,17 +397,9 @@ public abstract class AbstractNestablePropertyAccessor extends AbstractPropertyA
 		}
 
 		if (propValue == null) {
-			// null map value case
-			if (isAutoGrowNestedPaths()) {
-				int lastKeyIndex = tokens.canonicalName.lastIndexOf('[');
-				getterTokens.canonicalName = tokens.canonicalName.substring(0, lastKeyIndex);
-				propValue = setDefaultValue(getterTokens);
-			}
-			else {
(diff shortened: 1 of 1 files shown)


mdeinum (commented):
I would say that this is a false positive from your IDE. Although it appears that the value never changes it actually does. The `BeanWrapperFieldSetMapper` is such a usage in the Spring Batch Core project. However it is used in Spring itself as well as the `DataBinder` used a `BeanWrapperImpl` internally with the default value of `true` for the `autogrowNestedPaths` property. 

So the code is actually not dead code but still used.

yilianhuaixiao:
@mdeinum  I am not  sure about your words.   The 'propValue'  represents the direct souce of  the 'tokens' paremeter.
For example, if the 'tokens.canonicalName' is 'listOfMaps[0]['luckyNumber']'.  the  'propValue' represents  'listOfMaps[0]' 
object.

And in getPropertyValue method in AbstractNestablePropertyAccessor:
```
 if (isAutoGrowNestedPaths()) {
						value = setDefaultValue(new PropertyTokenHolder(tokens.actualName));
					}
```
in
```	private Object setDefaultValue(PropertyTokenHolder tokens) {
		PropertyValue pv = createDefaultPropertyValue(tokens);
		setPropertyValue(tokens, pv);
		Object defaultValue = getPropertyValue(tokens);
		Assert.state(defaultValue != null, "Default value must not be null");
		return defaultValue;
	}
```
Default value must not be null.

So once the 'propValue == null',
it means isAutoGrowNestedPaths   always  be false.

snicoll:
Yes I agree with @mdeinum here. Thanks for the PR, in any case.

PR #25647, opened 2020-08-26
Avoid use of String's for ContentDisposition type

kashike:
This adds constants for three common Content-Disposition types: `attachment`, `form-data`, and `inline`.

I have also updated the tests to use these new constants.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
+++ b/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
@@ -48,6 +48,18 @@ public final class ContentDisposition {
 	private static final String INVALID_HEADER_FIELD_PARAMETER_FORMAT =
 			"Invalid header field parameter format (as defined in RFC 5987)";
 
+	/**
+	 * The {@literal attachment} content-disposition type.
+	 */
+	private static final String ATTACHMENT = "attachment";
+	/**
+	 * The {@literal form-data} content-disposition type.
+	 */
+	private static final String FORM_DATA = "form-data";
+	/**
+	 * The {@literal inline} content-disposition type.
+	 */
+	private static final String INLINE = "inline";
 
 	@Nullable
 	private final String type;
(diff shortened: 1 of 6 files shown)


kashike:
Is there anything holding this pull request back? It is very simple and non-breaking.

rstoyanchev:
I think adding alternatives to the static `builder(String)` such as `formData()`, `inline()`, and `attachment()` would make the case of building more convenient than using constants and that be good enough since I'm guessing that building is the more common case.

kashike:
@rstoyanchev I have added the helper methods, but I have also kept the new constants as well, as they are still useful - for example, checking the type: `disposition.getType().equals(ContentDisposition.ATTACHMENT)`.

Let me know if this is okay with you now.

spring-web/src/main/java/org/springframework/http/ContentDisposition.java
```diff
@@ -264,6 +276,33 @@ public static Builder builder(String type) {
 		return new BuilderImpl(type);
 	}
 
+	/**
+	 * Return a builder for a {@code ContentDisposition} with
+	 * the {@link #ATTACHMENT attachment} type.
+	 * @return the builder
```
rstoyanchev, line 282: Please add `@since 5.3` tags to all new methods.

spring-web/src/main/java/org/springframework/http/ContentDisposition.java
```diff
@@ -48,6 +48,18 @@
...
+	 */
+	public static final String ATTACHMENT = "attachment";
+	/**
+	 * The {@literal form-data} content-disposition type.
+	 */
+	public static final String FORM_DATA = "form-data";
+	/**
+	 * The {@literal inline} content-disposition type.
+	 */
+	public static final String INLINE = "inline";
```
rstoyanchev, line 62: > I have also kept the new constants as well, as they are still useful - for example, checking the type: disposition.getType().equals(ContentDisposition.ATTACHMENT).

Why don't we then add `isInline`, `isAttachment`, and `isFormData` methods then?

rstoyanchev (changes requested):
Thanks I've schedule it for 5.3.

kashike:
@rstoyanchev How does that look now?

Issue #25557, opened 2020-08-07
Optimize NumberToNumberConverterFactory and StringToEnumConverterFactory


yilianhuaixiao:
When I look at `NumberToNumberConverterFactory`, I find no instance cache for `NumberToNumber`.

```java
public <T extends Number> Converter<Number, T> getConverter(Class<T> targetType) {
	return new NumberToNumber<>(targetType);
}
```

Therefore a new `NumberToNumber` object is instantiated for every conversion.

I added the following test case in `DefaultConversionServiceTests`, debugged it, and it really does create a new instance every time.

```java
	@Test
	void frequencyConvert() {
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
		assertThat(conversionService.convert("1.0", Float.class)).isEqualTo(Float.valueOf("1.0"));
	}
```

Should `NumberToNumberConverterFactor` and `StringToEnumConverterFactory` introduce caches to avoid frequently creating new objects?

kostya05983:
Hello @sbrannen, I think we can add some static caches, but I'm not sure. Maybe it must work in more clever way with another caches. I attached a pr with static caches.

bclozel:
Are those allocations an actual performance issue? @yilianhuaixiao do you have an actual application showing performance problems because of this, or is this issue a simple observation?

[...]

PR #25715, opened 2020-09-06
Fix Issue #25557

kostya05983:
Please see #25557

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
@@ -22,6 +22,9 @@ import org.springframework.core.convert.converter.Converter;
 import org.springframework.core.convert.converter.ConverterFactory;
 import org.springframework.util.NumberUtils;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
 /**
  * Converts from any JDK-standard Number implementation to any other JDK-standard Number implementation.
  *
@@ -41,10 +44,16 @@ import org.springframework.util.NumberUtils;
  * @see NumberUtils
  */
 final class NumberToNumberConverterFactory implements ConverterFactory<Number, Number>, ConditionalConverter {
+	private final static Map<Class<?>, NumberToNumber<?>> cache = new ConcurrentHashMap<>();
 
 	@Override
 	public <T extends Number> Converter<Number, T> getConverter(Class<T> targetType) {
-		return new NumberToNumber<>(targetType);
+		NumberToNumber converter = cache.get(targetType);
+		if (converter == null) {
+			converter = new NumberToNumber<>(targetType);
+			cache.put(targetType, converter);
+		}
+		return converter;
 	}
 
 	@Override
--- a/spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
@@ -16,14 +16,14 @@
 
 package org.springframework.core.convert.support;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
 import org.springframework.core.convert.TypeDescriptor;
 import org.springframework.core.convert.converter.ConditionalConverter;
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
```diff
@@ -40,11 +43,18 @@
  * @see java.math.BigDecimal
  * @see NumberUtils
  */
+@SuppressWarnings({"rawtypes", "unchecked"})
 final class NumberToNumberConverterFactory implements ConverterFactory<Number, Number>, ConditionalConverter {
+	private final static Map<Class<?>, NumberToNumber<?>> cache = new ConcurrentHashMap<>();
```
wind57, line 48: upper-case; java naming convention

spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
```diff
@@ -40,11 +43,18 @@
...
 
 	@Override
 	public <T extends Number> Converter<Number, T> getConverter(Class<T> targetType) {
-		return new NumberToNumber<>(targetType);
+		NumberToNumber converter = cache.get(targetType);
+		if (converter == null) {
```
wind57, line 53: `Map::computeIfAbsent` instead

spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
wind57: the thing is, when you add these `Class<?>` to a static cache - you make them GC roots, forever, in this case. Unless there is serious proof that this allocation is perf critical, this might harm more then fix.

kostya05983: Thanks, use Class<?> getName as a key.

spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
kostya05983: Now code is simpler! Thanks

spring-core/src/main/java/org/springframework/core/convert/support/NumberToNumberConverterFactory.java
wind57: that's... not what I meant. I mean the fact that you declare a `private static final Map... ` means that you declare an additional GC root. In general, you should not think about these details and just code the way it is easier to read.

Let's change the angle of this PR a bit, may be that will make more sense. _Why_ is there a need to fix _anything_ here? Because on each invocation there will be a new instance allocated, right? And let me ask you then - why is that bad? Or is it bad at all?

kostya05983: Yes, we have an instance allocated on each invocation but also we call ConversionUtils.getEnumType(targetType) for each invocation. When we use static cache we save small amount of time on this operations, but we need some warm up the server to work it the fastest way. Also, I want notice that ConversionUtils.getEnum() have a cycle inside. Also, we lost some memory 'cause of cache. I'l be glad if you say some words about how harmful GC root can be.

So, I want just to save some time inconversions by this pr.

wind57: - I was not saying that the addition of this Map is _harmful_; all I said is that I am not convinced this is something that needs to be "fixed". The allocation of such a small Object looks like a very tiny optimization (especially considering that the world is moving towards concurrent GCs, like `Shenandoah` or `ZGC` - where these allocations will not even matter).

- do you have any proof that you are saving some time, actually? and of course even if that is true, you are paying with memory for that. But now that you add a GC root - you are also _adding_ time for the scanning itself (though it will be very, very little; and for most GCs it will be concurrent).

 - overall, as I said initially, this has to be coded in manner that is easier to read and not think (much) about the details above. I thought that may be you have done some experiments and have some numbers, thus the initial questions.

sbrannen:
Closing in line with https://github.com/spring-projects/spring-framework/issues/25557#issuecomment-691503210.

PR #25724, opened 2020-09-07
Add annotation judgment in method BeanFactoryUtils#beanNamesForAnnotationIncludingAncestors

xiaobei-ihmhny:
When the local BeanFactory and the parent BeanFactory contain local bean of the same name and type, but only the parent BeanFactory contains specified annotations, should the parent Bean be used instead of the local Bean


pivotal-issuemaster:
@xiaobei-ihmhny Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25724)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25724) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@xiaobei-ihmhny Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25724)!

jhoeller:
That merge step has `getBeanNamesForAnnotation` results as input, so those beans are known to have the annotation already. The extra check added in this PR seems to prefer an annotation-carrying bean in a parent factory over a locally defined bean without the annotation which is at odds with our general `BeanFactory` lookup rules: A local bean of the same name always overrides a parent bean; a caller cannot retrieve the parent-defined bean from the local `BeanFactory` handle. So even if the bean in the parent factory had the annotaton, the local `getBean` call would always expose the local non-annotated bean instead.

From that perspective, the existing code seems to be working as designed, ignoring any parent match if a local bean overrides it (even if the local bean does not have the annotation and therefore not show up in the result). Thanks for raising it, in any case!

PR #25732, opened 2020-09-08
change to deque implementation

wind57:


wind57:
I do not know all the complete internals, but I found this while working with a different spring-project. Specifically, [here](https://github.com/spring-cloud/spring-cloud-kubernetes/blob/master/spring-cloud-kubernetes-config/src/main/java/org/springframework/cloud/kubernetes/config/SecretsPropertySourceLocator.java#L73). Accidentally, I looked up the implementation and I though this can be improved.

wind57:
I missed the fact that these are supposed to be unique, a Deque is no good.

PR #25742, opened 2020-09-08
Pass the exception to the HandlerExecutionChain

omradouane:
Hello,

If an exception is thrown, it should be passed to the `HandlerExecutionChain` which in turn passed it to any existing `HandlerInterceptor` that can used it for any business logic inside `afterCompletion()` method.

In fact, when an exception is thrown and handled, the `afterCompletion()` method of the interceptor (aka. `HandlerInterceptorAdapter`) has no clue on that exception since the `DispatcherServlet` pass `null` even if there is an exception.

A business case that can use this exception is statistics or logging (mail, jms, ...).

Thank you.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
@@ -1153,8 +1153,8 @@ public class DispatcherServlet extends FrameworkServlet {
 		}
 
 		if (mappedHandler != null) {
-			// Exception (if any) is already handled..
-			mappedHandler.triggerAfterCompletion(request, response, null);
+			// Exception (if any) is already handled, but passed in case of any business logic needs..
+			mappedHandler.triggerAfterCompletion(request, response, exception);
 		}
 	}
 


pivotal-issuemaster:
@omradouane Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25742)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25742) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@omradouane Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25742)!

omradouane:
Hello,

I found a workaround. The method `processHandlerException(HttpServletRequest request, HttpServletResponse response, @Nullable Object handler, Exception ex)` may save the exception as an attribute in the request.

So to get the exception, all you need to do is `request.getAttribute(DispatcherServlet.EXCEPTION_ATTRIBUTE);`

Thank you !

rstoyanchev:
@omradouane indeed, `HandlerInterceptor#afterCompletion` says it does not pass handled exceptions:
```java
	 * @param ex any exception thrown on handler execution, if any; this does not
	 * include exceptions that have been handled through an exception resolver
```

The option remains to check `DispatcherServlet.EXCEPTION_ATTRIBUTE`:
```
	 * Name of request attribute that exposes an Exception resolved with a
	 * {@link HandlerExceptionResolver} but where no view was rendered
	 * (e.g. setting the status code).
```

So this is by design and we can't change those contracts.

PR #25757, opened 2020-09-11
AbstractRequestLoggingFilter: split client info, conditional before/after

candrews:
* New methods to configure each part of the client info logging
* Session id is no longer logged by default when client info logging is enabled
  The session id is sensitive information (similar to user credentials)
* Methods to separately and conditional enable/disable before and after
  logging

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/filter/AbstractRequestLoggingFilter.java
+++ b/spring-web/src/main/java/org/springframework/web/filter/AbstractRequestLoggingFilter.java
@@ -95,6 +95,12 @@ public abstract class AbstractRequestLoggingFilter extends OncePerRequestFilter 
 
 	private boolean includeClientInfo = false;
 
+	private boolean includeClientInfoRemoteAddr = true;
+
+	private boolean includeClientInfoSessionId = false;
+
+	private boolean includeClientInfoUser = true;
+
 	private boolean includeHeaders = false;
 
 	private boolean includePayload = false;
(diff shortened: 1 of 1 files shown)


candrews:
@jhoeller can you (or anyone else applicable) please review and (hopefully) merge this PR?

Thank you!

jhoeller:
My impression is that there are too many boolean flags now, with the client info logging possibly being better off in an overridable template method? This would also allow for excluding the session id at that level.

rstoyanchev:
Yes too many flags indeed. Maybe we can add a protected method that extracts the desired request properties to map first:
```java
protected Map<String, Object> extractRequestProperties(HttpServletRequest request) {
	Map<String, Object> result = new LinkedHashMap<>();
	if (isIncludeClientInfo()) {
		result.put("client", request.getRemoteAddr());
		HttpSession session = request.getSession(false);
		if (session != null) {
			result.put("session", session.getId());
		}
		result.put("user", request.getRemoteUser());
	}
	if (isIncludeHeaders()) {
		HttpHeaders headers = new ServletServerHttpRequest(request).getHeaders();
		if (getHeaderPredicate() != null) {
			Enumeration<String> names = request.getHeaderNames();
			while (names.hasMoreElements()) {
				String header = names.nextElement();
				if (!getHeaderPredicate().test(header)) {
					headers.set(header, "masked");
				}
			}
		}
		result.put("headers", headers);
	}

	if (isIncludePayload()) {
		String payload = getMessagePayload(request);
		if (payload != null) {
			result.put("payload", payload);
		}
	}
	return result;
}
```

This then easy to format and a sub-class can override this method to further customize what goes in the Map.

PR #25777, opened 2020-09-15
Unify WebRequest parameter names

jaeyeonling:
In all other methods, `WebRequest` uses the name `request`.
Using `webRequest` only for `handleAsyncRequestTimeoutException()`.
Unify names for clean code.

Thanks!


rstoyanchev:
Thanks for the suggestion, although it is sensible on its own, arguably the code is also fine as it is. Local variable names do not need to be named the same way and many times there are good reasons to choose differently depending on the context such as how long the method is, how many other variables there are, how it fits in terms of line length, and so on.

PR #25780, opened 2020-09-16
Resolve Principal argument only when not annotated

### LINKED ISSUE #4151: AuthenticationPrincipalArgumentResolver not used when argument class implements Principal
### Summary

putting an @AuthenticationPrincipal annotation on a Controller parameter, where the parameter's type implements Principal causes the servlet to throw an IllegalArgumentException

### Actual Behavior

Basically, since ServletRequestMethodArgumentResolver supports Principal objects, it gets used, and it populates the parameter with an Authentication object, causeing a casting exception when the resolving AuthenticationObject is forced into my UserDetails object

### Expected Behavior

AuthenticationPrincipalArgumentResolver should be defaulted to resolve AuthenticationPrincipal arguments, which will do the correct resolution.

The workaround is not having my UserDetails object inherit Principal, or use the actual UserDetails class in the parameter, and then cast it to my subclass, but this is silly. It makes perfect sense that a UserDetails object is also a Princial, and it is in fact retrieved via Authentication _Principal_

#### Fix

The fix would be forcing ServletRequestMethodArgumentResolver to be lower priority in argument resolution than AuthenticationPrincipalArgumentResolver 

### Version

All

anthonyraymond:
`ServletRequestMethodArgumentResolver` happend early in the `ArgumentResolver` chain. One of his ability is to resolve `Principal`, that's great and it works well.

But when a parameter of type `Principal` is annotated we don't want to get the `Principal` from the `HttpServletRequest.getUserPrincipal()`.
This feature is even in conflict with the **spring-security** documentation and the [`@AuthenticationPrincipal`](https://docs.spring.io/spring-security/site/docs/4.2.15.RELEASE/apidocs/org/springframework/security/web/bind/annotation/AuthenticationPrincipal.html) annotation which is supposed to resolve the Principal from Authentication.getPrincipal().

Having the `ServletRequestMethodArgumentResolver` resolving annotated `Principal` makes the `@AuthenticationPrincipal` and `AuthenticationPrincipalArgumentResolver` useless and missleading.

Fix spring-projects/spring-security#4151

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolver.java
@@ -84,12 +84,13 @@ public class ServletRequestMethodArgumentResolver implements HandlerMethodArgume
 	@Override
 	public boolean supportsParameter(MethodParameter parameter) {
 		Class<?> paramType = parameter.getParameterType();
+		final Annotation[] parameterAnnotations = parameter.getParameterAnnotations();
 		return (WebRequest.class.isAssignableFrom(paramType) ||
 				ServletRequest.class.isAssignableFrom(paramType) ||
 				MultipartRequest.class.isAssignableFrom(paramType) ||
 				HttpSession.class.isAssignableFrom(paramType) ||
 				(pushBuilder != null && pushBuilder.isAssignableFrom(paramType)) ||
-				Principal.class.isAssignableFrom(paramType) ||
+				(Principal.class.isAssignableFrom(paramType) && parameterAnnotations.length == 0) ||
 				InputStream.class.isAssignableFrom(paramType) ||
 				Reader.class.isAssignableFrom(paramType) ||
 				HttpMethod.class == paramType ||
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolver.java
@@ -19,6 +19,7 @@ package org.springframework.web.servlet.mvc.method.annotation;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
+import java.lang.annotation.Annotation;
 import java.security.Principal;
 import java.time.ZoneId;
 import java.util.Locale;
--- a/spring-webmvc/src/test/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolverTests.java
+++ b/spring-webmvc/src/test/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolverTests.java
@@ -16,24 +16,6 @@
 
 package org.springframework.web.servlet.mvc.method.annotation;
 
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
-import org.springframework.core.MethodParameter;
-import org.springframework.http.HttpMethod;
-import org.springframework.web.context.request.ServletWebRequest;
-import org.springframework.web.context.request.WebRequest;
-import org.springframework.web.method.support.ModelAndViewContainer;
-import org.springframework.web.multipart.MultipartRequest;
-import org.springframework.web.servlet.DispatcherServlet;
-import org.springframework.web.servlet.i18n.FixedLocaleResolver;
-import org.springframework.web.testfixture.servlet.MockHttpServletRequest;
-import org.springframework.web.testfixture.servlet.MockHttpServletResponse;
-import org.springframework.web.testfixture.servlet.MockHttpSession;
-
-import javax.servlet.ServletRequest;
-import javax.servlet.http.HttpSession;
-import javax.servlet.http.PushBuilder;
 import java.io.InputStream;
 import java.io.Reader;
 import java.lang.annotation.ElementType;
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@anthonyraymond Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25780)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25780) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@anthonyraymond Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25780)!

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolver.java
```diff
@@ -84,12 +84,13 @@
...
 	public boolean supportsParameter(MethodParameter parameter) {
 		Class<?> paramType = parameter.getParameterType();
+		final Annotation[] parameterAnnotations = parameter.getParameterAnnotations();
 		return (WebRequest.class.isAssignableFrom(paramType) ||
 				ServletRequest.class.isAssignableFrom(paramType) ||
 				MultipartRequest.class.isAssignableFrom(paramType) ||
 				HttpSession.class.isAssignableFrom(paramType) ||
 				(pushBuilder != null && pushBuilder.isAssignableFrom(paramType)) ||
-				Principal.class.isAssignableFrom(paramType) ||
+				(Principal.class.isAssignableFrom(paramType) && parameterAnnotations.length == 0) ||
```
sbrannen, line 93: Merely checking for the absence of an annotation would not be sufficient, since the parameter could be annotated with annotations other than Spring Security's `@AuthenticationPrincipal`.

sbrannen: I think the Spring Security issue would need to addressed within Spring Security, potentially making use of `RequestMappingHandlerAdapter.setArgumentResolvers(List<HandlerMethodArgumentResolver>)` (as opposed to `setCustomArgumentResolvers(...)`).

@rstoyanchev and @rwinch, thoughts?

imjmakki (commented):
great job

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolver.java
rstoyanchev: I don't think setting the default resolvers is a good way to go.

We have some precedent with arguments of type `Map` where we do check that it isn't annotated in some way, see 5a3ff3521538725695da77d250c97fb49783b10f. We could so something similar here given that `Principal` is also a very high level interface with many possible uses and implementations. What I'm wondering though is why this problem surfaces only now, was there some change in Spring Security or did it never work?

rstoyanchev: Or does it depend on the user Object of the application? 

Under https://github.com/spring-projects/spring-security/issues/4151 it says one option is for UserDetails to not inherit Principal but from what I see it doesn't https://docs.spring.io/spring-security/site/docs/4.2.15.RELEASE/apidocs/org/springframework/security/core/userdetails/UserDetails.html.

sbrannen: > We have some precedent with arguments of type Map where we do check that it isn't annotated in some way, see 5a3ff35.

OK. I wasn't aware of that.

sbrannen: > Or does it depend on the user Object of the application?

Yes, it depends on the user-specific implementation of `UserDetails`.

> Under [spring-projects/spring-security#4151](https://github.com/spring-projects/spring-security/issues/4151) it says one option is for UserDetails to not inherit Principal but from what I see [UserDetails](https://docs.spring.io/spring-security/site/docs/4.2.15.RELEASE/apidocs/org/springframework/security/core/userdetails/UserDetails.html) doesn't .

Right. The `UserDetails` interface does not extend `Principal` by default, though a developer may of course choose to implement `Principal` if they want (or need) to.

I have implemented in `UserDetails` in several applications of the years, and I have never had it implement `Principal` as well. I also have never seen anyone do that. But... as I mentioned above, it's certainly an option if it makes sense in a given application.

anthonyraymond: Hi guys thanks for your investment in this issue.

to answer @rstoyanchev 
> What I'm wondering though is why this problem surfaces only now, was there some change in Spring Security or did it never work?

My guess is that it's never worked, people use the `@AuthenticationPrincipal` and they get what they asks for: a `Principal`. It's not the expected one, because it's not coming from the right place, but it's a principal and it's good enough for most devs.

But this behavior has always been a problem:
-  [2019 SO post](https://stackoverflow.com/questions/58817175/java-lang-illegalstateexceptioncurrent-user-principal-is-not-of-type-when-i-try): `HttpServletRequest.getUserPrincipal()` contains the `Authentication` (which extends Principal) so it returns it, and it does not extract the Principal from the Authentication. There are multiple solutions for this around SO and most of them say : *Use `org.keycloak.adapters.springsecurity.token.KeycloakAuthenticationToken` in place of `KeycloakPrincipal`* (which is an implementation of `Authentication`) and magically it starts to work, but not for the good reasons.
- [2015 SO post](https://stackoverflow.com/questions/28834154/why-does-authenticationprincipal-return-the-authentication-instead-of-the-princ): Because `HttpServletRequest.getUserPrincipal()` contains the `Authentication` (which extends Principal).

Basically what you can get with the Google search "**Current user principal is not of type**" is related to this issue.

Both of these issues come from the fact that the Principal resolved is from `HttpServletRequest.getUserPrincipal()` instead of `SecurityContext.getAuthentication.getPrincipal()`.

The nasty thing about that is that the `@AuthenticationPrincipal` has not only become not relevant but also missleading, because all class extending Principal can be injected into a `Principal` type (thanks to the spring-framework, the annotation takes no part in this), but as soon as you want to get your implementation of `Principal` you are doomed.

rstoyanchev:
Thanks for the extra detail. This is scheduled for 5.3 now.

> Why is the spring-framework trying to inject a Principal ?

It is one of many request properties that `ServletRequestMethodArgumentResolver` aims to expose.

anthonyraymond:
Thanks for the quick reply.

Since this will most likely get merged, i added a test to ensure it wont break in the future.

spring-webmvc/src/test/java/org/springframework/web/servlet/mvc/method/annotation/ServletRequestMethodArgumentResolverTests.java
```diff
@@ -44,6 +31,21 @@
 import org.springframework.web.testfixture.servlet.MockHttpServletResponse;
 import org.springframework.web.testfixture.servlet.MockHttpSession;
 
+import javax.servlet.ServletRequest;
```
sbrannen, line 34: Please adhere to the [project guidelines for imports](https://github.com/spring-projects/spring-framework/wiki/Code-Style#import-statements).

In addition, please execute `./gradlew check` before submitting a PR or pushing changes to a PR.

anthonyraymond: Done, sorry about that

imjmakki (commented):
nice

imjmakki (commented):
good

PR #25810, opened 2020-09-23
Optimize StringUtils trimLeading/Trailing methods

Shenker93:
This PR includes some StringUtils methods optimization. The problem is that trimLeading/TrailingCharacter and trimLeading/TrailingWhitespace methods create a StringBuilder object as a local var instead of using String.substring method. Such approach is leading to unnecessary object creation on every method invocation. Also it runs slower because of internal StringBuilder operations and conversions String->StringBuilder->String. According to simple JMH benchmark, substring is x2/x3 faster on avg scenarios. The situation with long strings that are not changed at all(if the method returns an equal string) is especially dramatic for StringBuilder (but substring will return its param instead of new a string creation)
Also found out that trimWhitespace(String str) method is already implemented in similar style.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -271,11 +271,11 @@ public abstract class StringUtils {
 			return str;
 		}
 
-		StringBuilder sb = new StringBuilder(str);
-		while (sb.length() > 0 && Character.isWhitespace(sb.charAt(0))) {
-			sb.deleteCharAt(0);
+		int beginIdx = 0;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@Shenker93 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25810)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25810) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Shenker93 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25810)!

PR #25813, opened 2020-09-24
Support custom TimeUnits in StopWatch

### LINKED ISSUE #25803: Support custom TimeUnits in StopWatch
I use `spring-core:5.2.9.RELEASE`.

When I invoke `StopWatch#prettyPrint`,

```text
StopWatch '': running time = 9624841500 ns
---------------------------------------------
ns         %     Task name
---------------------------------------------
1202450400  012%  createRandoms
8422391100  088%  write
```

I want `StopWatch` to provide `setTimeUnit(java.util.concurrent.TimeUnit)`, `getLastTime(TimeUnit)` and `getTotalTime(TimeUnit)`; the return type of these methods is `double`.

For example,

```java
StopWatch stopWatch = new StopWatch();
stopWatch.setTimeUnit(TimeUnit.MINUTES); // If not set, StopWatch is applied default TimeUnit.
stopWatch.start();

/* working... */

stopWatch.stop();
getTotalTime(TimeUnit.SECONDS); // 9.6248415
getTotalTime(); // 0.160414025 / Without param, it depends on StopWatch#timeUnit
System.out.println(stopWatch.prettyPrint());
```
is like ...

```text
StopWatch '': running time = 0.160414025 min
---------------------------------------------
min         %     Task name
---------------------------------------------
0.02004084   012%  createRandoms
0.140373185  088%  write
```

Thanks!

ArtemKudriaC:
Add `TimeUnit` support in `StopWatch` util class.

- Provide `StopWatch#getLastTime` and `StopWatch#getTotalTime`
- Modify the print format for tasks  / include default `TimeUnit` in nanoseconds

See #25803


spring-core/src/main/java/org/springframework/util/StopWatch.java
```diff
@@ -114,6 +117,12 @@ public void setKeepTaskList(boolean keepTaskList) {
 		this.keepTaskList = keepTaskList;
 	}
```
choi1135, line 118: good

spring-core/src/main/java/org/springframework/util/StopWatch.java
```diff
@@ -338,6 +376,34 @@ private static double nanosToSeconds(long duration) {
...
 
+	private static double convertTimeUnit(double amount, TimeUnit from, TimeUnit to) {
+		if (from.ordinal() < to.ordinal()) {
+			return amount / from.convert(1, to);
+		} else {
+			return amount * to.convert(1, from);
+		}
+	}
+
+	private static String getTimeUnitAbbreviation(TimeUnit timeUnit) {
```
yun-yun, line 387: Why use name() string rather than enum for lookups?
What's the good of it?
eg:
```
private static String getTimeUnitAbbreviation(TimeUnit timeUnit) {
      switch (timeUnit) {
          case NANOSECONDS:
              return "ns";
          case MICROSECONDS:
              return "μs";
          case MILLISECONDS:
              return "ms";
          case SECONDS:
              return "sec";
          case MINUTES:
              return "min";
          case HOURS:
              return "hrs";
          case DAYS:
              return "days";
          default:
              throw new IllegalArgumentException("No TimeUnit equivalent for " + timeUnit);
      }
  }
```

msangel (approved):

msangel:
Precisely what I was looking for. I don't know how many nanoseconds are in one second, so this will simplify my life.

spring-core/src/main/java/org/springframework/util/StopWatch.java
ak98neon: Updated

ventulus95:
The function itself is implemented, but it hasn't passed the CI yet.  So won't this PR be merged in the foreseeable future? It is very useful for debugging and I would like to use it in a version, so when will this PR be merged? @ak98neon

snicoll:
@ventulus95 it's in the general backlog so there's no commitment on our end. That said, I am not sure we should implement it this way. We could reuse `Duration` which sounds a more rich concept for what you're trying to do.

jhoeller:
I don't see the need for a new group of TimeUnit-based accessors and see them as rather confusing next to our existing `getLaskTaskTimeX` and `getTotalTimeX` accessors, in particular if the parameterless variant returns the time in a unit based on a `StopWatch` instance property.

That said, I do see the point with `prettyPrint`. The pre-5.2 representation with milliseconds was arguably more useful than the present implementation is. This can be addressed through an overloaded `prettyPrint(TimeUnit)` method which I'm inclined to introduce under #25803.

That aside, there is the question of `Duration`-returning time accessors... and of `ChronoUnit` for `prettyPrint` if `java.time` alignment is more appropriate than `java.util.concurrent` alignment. Either way, this is worth a broader revision for 6.1.

jhoeller:
See https://github.com/spring-projects/spring-framework/issues/25803#issuecomment-1668705268 for a summary of the 6.1 revision. This ended up preferring `TimeUnit` based accessors over `Duration` since the `double` precision is rather important here. Note that there is no instance-level `TimeUnit` setting, just a `TimeUnit` method parameter for specific accessors.

PR #25846, opened 2020-09-30
Drop explicit zeroing at instantiation of Atomic* objects

stsypanov:
Often fields of atomic types e.g. AtomicInteger are explicitly zeroed at initialization, which is redundant and slower than relying on default values:
```java
@State(Scope.Thread)
@BenchmarkMode(value = Mode.AverageTime)
public class AtomicBenchmark {
  @Benchmark
  public Object defaultValue() {
    return new AtomicInteger();
  }
  @Benchmark
  public Object explicitValue() {
    return new AtomicInteger(0);
  }
}
```
Semantically both new AtomicInteger() and new AtomicInteger(0) are the same, but explicitValue() is much slower:
```java
Benchmark                      Mode  Cnt   Score   Error  Units
AtomicBenchmark.defaultValue   avgt   30   4.778 ± 0.403  ns/op
AtomicBenchmark.explicitValue  avgt   30  11.846 ± 0.273  ns/op
```
This is related to https://github.com/spring-projects/spring-framework/pull/25261

### DIFF
--- a/integration-tests/src/test/java/org/springframework/scheduling/annotation/ScheduledAndTransactionalAnnotationIntegrationTests.java
+++ b/integration-tests/src/test/java/org/springframework/scheduling/annotation/ScheduledAndTransactionalAnnotationIntegrationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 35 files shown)


wind57:
very interesting. this seems to happen because there is an explicit `volatile` write, via : `AtomicInteger(0)`.

stsypanov:
@wind57 yes, I've mentioned a previous PR being about the same: https://github.com/spring-projects/spring-framework/pull/25261

Also see https://bugs.openjdk.java.net/browse/JDK-8145680

PR #25848, opened 2020-10-01
UserCredentialsConnectionFactoryAdapter createContext changes.

dotbg:
UserCredentialsConnectionFactoryAdapter holds the username/password and uses them for the createConnection mehod. 
This PR makes use of the same technic for  the `createContext` methods


pivotal-issuemaster:
@dotbg Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25848)!

snicoll:
@dotbg thanks for the PR but keeping the credentials in a thread local like this doesn't sound right to me. The JMS spec states that the context is created with the "default user identity" which should match the one that was used when the connection factory was created. 

As for the session mode, we're retaining compatibility with `1.1` so creating the context with no specified mode is what we should be doing.

It may be that you experienced an issue and that code fixed it for you. If that's the case, please create an issue with a small sample that we can run ourselves to reproduce the issue and we'll take it from there.

PR #25855, opened 2020-10-03
Update javax.mail reference to jakarta.mail

thanus:
When I wanted to use `spring-boot-starter-mail` I noticed that `jakarta.mail` should be used instead of `javax.mail`. This has now been changed in the documentation.

### DIFF
--- a/src/docs/asciidoc/integration.adoc
+++ b/src/docs/asciidoc/integration.adoc
@@ -4421,10 +4421,10 @@ This section describes how to send email with the Spring Framework.
 The following JAR needs to be on the classpath of your application in order to use
 the Spring Framework's email library:
 
-* The https://javaee.github.io/javamail/[JavaMail] library
+* The https://eclipse-ee4j.github.io/mail/[JavaMail] library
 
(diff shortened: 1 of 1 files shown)


src/docs/asciidoc/integration.adoc
```diff
@@ -4421,10 +4421,10 @@ This section describes how to send email with the Spring Framework.
 The following JAR needs to be on the classpath of your application in order to use
 the Spring Framework's email library:
 
-* The https://javaee.github.io/javamail/[JavaMail] library
+* The https://eclipse-ee4j.github.io/mail/[JavaMail] library
```
thanus, line 4424: It seems like Jakarta Mail is the new name for JavaMail. Should this also be changed? JavaMail is mentioned in several places.

jhoeller: We could mention both names. JavaMail is still the canonical name for the API in Java EE 7/8 and therefore our baseline. Jakarta EE 9 and Jakarta Mail are not common enough (yet) to switch the primary names that we use for those APIs.

PR #25877, opened 2020-10-07
Introduce SQL statement filter support in JDBC `ScriptUtils`

qxo:
so we can skip some sql statements by regex pattern,
such as: we do not want run delete/update/drop statements on production environment.
just call ResourceDatabasePopulator.ignoreDeleteUpdateDrop to lock sqlIgnorePattern=^\\s*(delete|drop|update).+

sorry for my poor english...and someone please help me fix  the english javadocs!

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/init/InitSqlFilter.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/init/InitSqlFilter.java
@@ -1,0 +1,41 @@
+/* Copyright 2002-2017 the original author or authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


quaff:
I think it's a specific use case, your application should handle it, not framework.

quaff:
Create a custom `Resource` implementation (e.g. `SqlSafeResource`) is better.

qxo:
> I think it's a specific use case, your application should handle it, not framework.

It'll be great if this feature native ready by the spring framework :)

SqlSafeResource is a better naming. 
And with the InitSqlFilter  for mini modify:)

snicoll:
I agree with @quaff, and the high-level method to ignore drops if the line starts with `drop` looks very specific to me. In general, skipping something by regexp is better done in a project where such assumption about the format can be done.

Issue #25896, opened 2020-10-12
readWithMessageConverters doesn't close FileInputStream


SenseiFisher:
I have an issue with readWithMessageConverters(HttpInputMessage inputMessage, ...)
in AbstractMessageConverterMethodArgumentResolver class.

It seems like this method initialize a new FileInputStream on calling inputMessage.getBody(),
then the FileInputStream is read by calling:
1. genericConverter.read(..., inputMessage);
OR
2. ((HttpMessageConverter<T>) converter).read(targetClass, inputMessage);

This FileInputStream is getting closed on calling genericConverter.read(..., inputMessage) (Jackson as a genericConverter).
But on calling ((HttpMessageConverter<T>) converter).read(targetClass, inputMessage);
with a converter of type StringHttpMessageConverter
the FileInputStream is not closed.

There is no decumentation on HttpMessageConverter that indicates the read method should close the FileInputStream.

So I guess the "readWithMessageConverters" method should ensure the FileInputStream is closed.

PR #25897, opened 2020-10-12
Close Body after reading

SenseiFisher:
readWithMessageConverters opens a FileInputStream but does not ensure it closes after reading from it.
Ensure the FileInputStream is closed.

Solves Issue:
https://github.com/spring-projects/spring-framework/issues/25896

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/AbstractMessageConverterMethodArgumentResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/AbstractMessageConverterMethodArgumentResolver.java
@@ -216,6 +216,11 @@ public abstract class AbstractMessageConverterMethodArgumentResolver implements 
 		catch (IOException ex) {
 			throw new HttpMessageNotReadableException("I/O error while reading input message", ex, inputMessage);
 		}
+		finally {
+			if (message.hasBody()) {
+				message.getBody().close();
+			}
+		}
 
 		if (body == NO_VALUE) {
 			if (httpMethod == null || !SUPPORTED_METHODS.contains(httpMethod) ||


pivotal-issuemaster:
@SenseiFisher Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25897)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25897) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@SenseiFisher Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=25897)!

akashchandwani (approved):

rstoyanchev:
I don't think it should be closed as it may not be the end of the response, for example in a multipart request but it could also be in another streaming scenario. See for example #14728 or the more recent regression #25989 caused by a similar fix.

Is there a specific issue that this presents?

SenseiFisher:
Hi,
I recived a "Too many opened files" (50,000)
a few times at the production environment.
After investigating it seems like the files belongs to tomcat.

I ran some tests at our development environment and saw the files only
closed at GC when using the StringHttpConverter (finalize).

This led me to belive it just a matter of luck, if GCs are not executed
frequent enough, every system will recive this exact same error.



בתאריך יום ב׳, 2 בנוב׳ 2020, 15:44, מאת Rossen Stoyanchev ‏<
notifications@github.com>:

jhoeller:
Since this is on the input side, we could potentially close it once the request body has been fully read. I'm less concerned there than on the output side but still wondering why it would be necessary.

Where does the `FileInputStream` come from, actually? Is it Tomcat opening such a stream?

SenseiFisher:
Well...
The InputSteam is created on the
EmptyBodyCheckingHttpInputMessage constructor, and the instance is stored
in "message".

After some processing, the function ends and return the processed body.
Making the "message" unreachable.

As a result once the function ends the InputStream cannot be closed
manually.


בתאריך יום ג׳, 1 בדצמ׳ 2020, 10:12, מאת Juergen Hoeller ‏<
notifications@github.com>:

rstoyanchev:
Scheduling for consideration in 6.0 M2 along with #27773.

> Where does the FileInputStream come from, actually? Is it Tomcat opening such a stream?

If if it is related to reading parts, as in #27773, then the the underlying input stream could be a temp file for the part.

rstoyanchev:
We're going to address this in a similar way but with a slightly different implementation, so I'm closing this as superseded by #27773.

PR #25899, opened 2020-10-12
Avoid multiple volatile reads/writes in a row where only one is enough

stsypanov:
- in `MethodInvocationProceedingJoinPoint` use the same approach as in `InjectionPoint.getAnnotations()` to do only one volatile read for all calls after field initialization instead of two.
- in `AutowiredAnnotationBeanPostProcessor` we are guareded by `synchronized`
- `AbstractCacheManager` newly created `LinkedHashSet` always has some extra capacity, so calculation of it's size +1 is redundant

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/aspectj/MethodInvocationProceedingJoinPoint.java
+++ b/spring-aop/src/main/java/org/springframework/aop/aspectj/MethodInvocationProceedingJoinPoint.java
@@ -221,10 +221,11 @@ public class MethodInvocationProceedingJoinPoint implements ProceedingJoinPoint,
 		@Override
 		@Nullable
 		public String[] getParameterNames() {
-			if (this.parameterNames == null) {
-				this.parameterNames = parameterNameDiscoverer.getParameterNames(getMethod());
+			String[] parameterNames = this.parameterNames;
+			if (parameterNames == null) {
+				this.parameterNames = parameterNames = parameterNameDiscoverer.getParameterNames(getMethod());
 			}
-			return this.parameterNames;
+			return parameterNames;
 		}
 
 		@Override
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java
@@ -644,21 +644,20 @@ public class AutowiredAnnotationBeanPostProcessor implements SmartInstantiationA
 				}
 				synchronized (this) {
 					if (!this.cached) {
+						Object cachedFieldValue = null;
 						if (value != null || this.required) {
-							this.cachedFieldValue = desc;
+							cachedFieldValue = desc;
 							registerDependentBeans(beanName, autowiredBeanNames);
 							if (autowiredBeanNames.size() == 1) {
 								String autowiredBeanName = autowiredBeanNames.iterator().next();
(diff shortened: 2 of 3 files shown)


spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java
```diff
@@ -644,21 +644,20 @@ protected void inject(Object bean, @Nullable String beanName, @Nullable Property
...
-									this.cachedFieldValue = new ShortcutDependencyDescriptor(
+									cachedFieldValue = new ShortcutDependencyDescriptor(
 											desc, autowiredBeanName, field.getType());
 								}
 							}
 						}
-						else {
-							this.cachedFieldValue = null;
-						}
+						this.cachedFieldValue = cachedFieldValue;
```
wind57, line 660: hmm, but `cachedFieldValue` _and_ `cached` is `volatile`? A volatile write (`this.cached = true`) ensures that all _subsequent_ reads that see this write, will also observer a fully and initialized `cachedFieldValue`... what I am missing?

spring-aop/src/main/java/org/springframework/aop/aspectj/MethodInvocationProceedingJoinPoint.java
```diff
@@ -221,10 +221,11 @@ public Method getMethod() {
 		@Override
 		@Nullable
 		public String[] getParameterNames() {
-			if (this.parameterNames == null) {
-				this.parameterNames = parameterNameDiscoverer.getParameterNames(getMethod());
+			String[] parameterNames = this.parameterNames;
```
wind57, line 224: I really like this. the performance impact is tiny: on `x86` this is "free", on `arm` this would translate to `lsync` (_lightweight_ sync); but it just sets a pattern - which I really like.

spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java
stsypanov: Nothing changes here as of seeing `cachedFieldValue`, because 1) we are inside of synchronized block and 2) `cachedFieldValue` is anyway assigned. Here I try to avoid the case when we assign `this.cachedFieldValue = desc;` (line 648 in original code) and then reassign it with `new ShortcutDependencyDescriptor()` (line 654 in original code) which is redundant. Instead we can write the values into an intermediate variable and then write it into volatile field only once

wind57: may be I was not clear enough, sorry. I perfectly understand what you are trying to do - yes, it also makes sense. No question about it. There is a minor problem with your reasoning (or wording) : indeed you are doing the _write_ in a `synchronized` block, but the _read_ happens outside:

```
if (this.cached) {
	value = resolvedCachedArgument(beanName, this.cachedFieldValue);
}
```

anyway, that is not the point I was trying to make. 

My question is rather different, why keep `cachedFieldValue` as `volatile`? It surely looks like `cached` being `volatile` would be enough.  Does that make sense now?

stsypanov: Yes,now I clearly understand your point. I'd keep `cachedFieldValue`volatile because it's not a bottleneck, and there's a risk of breaking the code.

Issue #22146, opened 2018-12-20
SpEL, InlineMap with negative values are not cached  [SPR-17614]


spring-projects-issues:
**[Davide Pallaoro](https://jira.spring.io/secure/ViewProfile.jspa?name=davide.pallaoro)** opened **[SPR-17614](https://jira.spring.io/browse/SPR-17614?redirect=false)** and commented

Hi there,

I’ve seen that InlineMap class have a nice mechanism that, cache the value of the map if it is constant.

But if the map contains a negative number this caching mechanism is not working.

 

Take for example the following test

```java
package test;

import org.springframework.expression.Expression;
import org.springframework.expression.ExpressionParser;
import org.springframework.expression.spel.standard.SpelExpressionParser;
import org.testng.Assert;
import org.testng.annotations.Test;

public class MapCacheTest {

    @Test
    public void testMapCached() {
        parseMap("{1 : 2, 3 : 4}");
    }

    @Test
    public void testMapNOTCached() {
        parseMap("{-1 : 2, 3 : 4}");
    }

    private void parseMap(String s) {
        ExpressionParser parser = new SpelExpressionParser();
        Expression expression = parser.parseExpression(s);
        Object value = expression.getValue();
        Assert.assertNotNull(value);
    }
}
```

 If you run the previous code, (using org.springframework.spring-expression:5.1.3.RELEASE) and set a breakpoint in

```java
org.springframework.expression.spel.ast.InlineMap#getValueInternal
```

which code is the following..
 

```java
@Override
public TypedValue getValueInternal(ExpressionState expressionState) throws EvaluationException {
     if (this.constant != null) {
          return this.constant;
     }
     else {
          Map<Object, Object> returnValue = new LinkedHashMap<>();
          int childcount = getChildCount();
          for (int c = 0; c < childcount; c++) {
```

You will notice that when executing **MapCacheTest.testMapCached()** this.constant is returned.

But when running **MapCacheTest.testMapNOTCached()** the code jump into the else statement and create a new map each time this method is called.

 

In a scenario with an high concurrency and with pretty big maps, this could decrease the performance of the app.

 

Let me know if everything is clear or you need some further details.

Many thanks and kind regards.

 

---

**Affects:** 5.1.3

**Issue Links:**
- #22137 SpEL, error parsing big InlineMap
- #22157 Revisit restrictions in SpEL position handling

masosky:
Hi there,

I would like to add that this is also occurring in InlineMap.java class.
[...]

PR #25921, opened 2020-10-16
Improve SpEL inline collection caching

SammyVimes:
This is a fix for https://github.com/spring-projects/spring-framework/issues/22146

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/ast/InlineList.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/ast/InlineList.java
@@ -27,6 +27,7 @@ import org.springframework.expression.TypedValue;
 import org.springframework.expression.spel.CodeFlow;
 import org.springframework.expression.spel.ExpressionState;
 import org.springframework.expression.spel.SpelNode;
+import org.springframework.expression.spel.support.StandardEvaluationContext;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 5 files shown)


sbrannen:
For future reference, there was no need to close #25047.

If you need to make changes to a PR, you can push additional commits to the PR branch and rebase on the latest `master` if necessary.

sbrannen:
@aclement, do you have time to review the proposal?

aclement:
It looks reasonable to me @sbrannen . The only concern I'd have is that we do allowable SpEL compilation of inline lists (not inline maps yet):

In InlineList:
```
	public boolean isCompilable() {
		return isConstant();
	}
```

So maybe a few tests verifying compiled expressions using inline lists are still OK with the constant computation extended to include OpMinus nodes. I *think* it'll be ok from browsing the code, but worth testing.

SammyVimes:
@aclement I added tests verifying compilation, looks like it's still OK

SammyVimes:
Hello! I wonder if we can proceed with the review and merge this? :)

sdeleuze:
Rebased and merged, thanks for your contribution and sorry for the delay.

sbrannen:
Indeed, thanks for the PR, @SammyVimes! 👍 

I unfortunately forgot about this PR and so only got around to reviewing it after it had been merged.

You can see the result of my _belated review_ in e7cf54d4e236ac1e5e5dd59899aae268bcace21d.

PR #25927, opened 2020-10-16
Improve ResourceArrayPropertyEditor time complexity

shevtsiv:
The previous implementation uses ArrayList for storing resolved
resources and ArrayList has O(n) time complexity for the contains
operation. By switching to the HashSet for storing resolved
resources we improve the time complexity of this operation to be O(1).

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/support/ResourceArrayPropertyEditor.java
+++ b/spring-core/src/main/java/org/springframework/core/io/support/ResourceArrayPropertyEditor.java
@@ -18,10 +18,10 @@ package org.springframework.core.io.support;
 
 import java.beans.PropertyEditorSupport;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


jijilki (approved):

rstoyanchev:
Doesn't this changes the order of resources and eliminate duplicates both of which could be significant? Are you seeing an actual performance hotspot with this in some scenario?

shevtsiv:
> Doesn't this changes the order of resources and eliminate duplicates both of which could be significant? Are you seeing an actual performance hotspot with this in some scenario?

I don't have any specific performance metrics, just found a spot in the scope of some bugfix, and decided to suggest an improvement. You are right, this may change the order of the resources, but the elimination should be done right as the initial code also checks for duplicates. If the ordering could not be fixed with something like TreeSet (and I suppose it could be not), then I think this PR can be closed, thanks!

rstoyanchev:
Thanks, I've updated it to use `LinkedHashSet`.

PR #25951, opened 2020-10-21
Support Optional for query parameters in UriBuilder and UriComponentsBuilder

robinroos:
Add a method optionalQueryParam(String, Optional) to the UriBuilder interface.

The implementation adds the query parameter contained by the optional to the URL.  If the optional is empty then no action is taken; specifically the query parameter name is not added to the URL.

This PR replaces #25925 and #25950.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriBuilder.java
@@ -19,6 +19,7 @@ package org.springframework.web.util;
 import java.net.URI;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Optional;
 
 import org.springframework.lang.Nullable;
 import org.springframework.util.MultiValueMap;
@@ -184,6 +185,15 @@ public interface UriBuilder {
 	 * @see #queryParam(String, Collection)
 	 */
 	UriBuilder queryParam(String name, Object... values);
+
+	/**
+	 * Append the given query parameter and value if and only if a value is present in the Optional.
+	 * No action will be taken, and the query parameter name will not be added, if the passed Optional is empty.
+	 * @param name the query parameter name
+	 * @param optionalValue an Optional, either empty or holding the query parameter value.
+	 * @return
+	 */
+	UriBuilder optionalQueryParam(String name, Optional<?> optionalValue);
 
 	/**
 	 * Variant of {@link #queryParam(String, Object...)} with a Collection.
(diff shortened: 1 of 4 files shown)


spring-web/src/main/java/org/springframework/web/util/UriBuilder.java
```diff
@@ -185,6 +186,15 @@
...
 	UriBuilder queryParam(String name, Object... values);
 
+	/**
+	 * Append the given query parameter and value if and only if a value is present in the Optional.
+	 * No action will be taken, and the query parameter name will not be added, if the passed Optional is empty.
+	 * @param name the query parameter name
+	 * @param optionalValue an Optional, either empty or holding the query parameter value.
+	 * @return
+	 */
+	UriBuilder optionalQueryParam(String name, Optional<?> optionalValue);
```
rstoyanchev, line 196: `queryParamIfPresent` aligns better with terminology from `Optional` and it sorts better alphabetically with the existing `queryParam` methods.

spring-web/src/main/java/org/springframework/web/util/UriBuilder.java
```diff
@@ -185,6 +186,15 @@
 	 */
 	UriBuilder queryParam(String name, Object... values);
 
+	/**
+	 * Append the given query parameter and value if and only if a value is present in the Optional.
+	 * No action will be taken, and the query parameter name will not be added, if the passed Optional is empty.
+	 * @param name the query parameter name
+	 * @param optionalValue an Optional, either empty or holding the query parameter value.
+	 * @return
+	 */
```
rstoyanchev, line 195: This can be described in terms of delegating to the existing `queryParam(String, Object...)` or `queryParam(String, Collection<?>)` methods if and only if the Optional has a value.

Note the case of the value being a Collection. That should be explicitly addressed in the Javadoc and the implementation should ensure it delegates to the correct method.

rstoyanchev (changes requested):

spring-web/src/main/java/org/springframework/web/util/UriBuilder.java
robinroos: Done.

spring-web/src/main/java/org/springframework/web/util/UriBuilder.java
robinroos: Done.

PR #26011, opened 2020-11-02
Optimize AbstractBeanFactory.afterPrototypeCreation()

hzmpay:
When curVal is of type Set, The instanceof judgment only needs to be performed once, And when the length of the Set is 1, Call method prototypesCurrentlyInCreation.remove() directly.


snicoll:
Thanks for the PR but the code is fine as it is and reads better than your proposal IMO.

PR #26012, opened 2020-11-02
Improve URI/query strings sanitization

stsypanov:
Use `StringBuilder.deleteCharAt(int)` and `StringBuilder.delete(int, int)` to handle sanitization more effectively

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -1023,15 +1023,21 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 			if (this.path.length() == 0) {
 				return null;
 			}
-			String path = this.path.toString();
-			while (true) {
-				int index = path.indexOf("//");
-				if (index == -1) {
-					break;
+			String sanitized = getSanitizedPath(this.path);
+			return new HierarchicalUriComponents.FullPathComponent(sanitized);
+		}
+
+		private static String getSanitizedPath(final StringBuilder path) {
+			int index = path.indexOf("//");
+			if (index >= 0) {
+				StringBuilder sanitized = new StringBuilder(path);
+				while (index != -1) {
+					sanitized.deleteCharAt(index);
+					index = sanitized.indexOf("//", index);
 				}
-				path = path.substring(0, index) + path.substring(index + 1);
+				return sanitized.toString();
 			}
-			return new HierarchicalUriComponents.FullPathComponent(path);
+			return path.toString();
 		}
 
 		public void removeTrailingSlash() {
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterUtils.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/namedparam/NamedParameterUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
```diff
@@ -1023,18 +1023,11 @@ public PathComponent build() {
...
 			}
-			String path = this.path.toString();
-			while (true) {
-				int index = path.indexOf("//");
-				if (index == -1) {
-					break;
-				}
-				path = path.substring(0, index) + path.substring(index + 1);
-			}
+			String path = UrlPathHelper.getSanitizedPath(this.path);
```
rstoyanchev, line 1026: I would prefer to avoid any dependency from `UriComponentsBuilder` on `UrlPathHelper`. The latter is more of a Servlet API specific component while the former is more general. Even if the re-use in this case doesn't touch on anything Servlet specific, it seems very minor and not worth the gain.

rstoyanchev (changes requested):

spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
jhoeller: Good catch, Rossen. I'd rather accept some minor duplication there as well.

stsypanov: @rstoyanchev should I move the method into a separate utility class or keep duplicated code fragment?

jhoeller: I'd rather keep the duplicated code fragment, unless there is an existing common place to add this to, @rstoyanchev?

BTW a minor thing: The `public void removeTrailingSlash() ` method was intentionally public. We tend to declare API methods as public if they're called from outside even for inner classes, along the lines of "what if this inner class was an externally declared class, would it have to expose this as public method then".

rstoyanchev: Yes, the minor duplication is the preferred outcome.

stsypanov: Done

rstoyanchev (commented):
Looks fine as far as I UriComponentsBuilder and UrlPathHelper are concerned. 

Very minor, I would say `getSanitizedPath` and `doubleToSingleSlash` can be combined into one method. Also the same identical method could be created in `UriComponentsBuilder`. When there is identical logic it's useful to make it look as identical as possible.

stsypanov:
@rstoyanchev done.

I've also measured performance impact with simple benchmark
```java
@Fork(jvmArgsAppend = {"-Xms2g", "-Xmx2g"})
public class DropDoubleSlashBenchmark {

  @Benchmark
  public String ineffective(Data data) {
    String path = data.path;
    while (true) {
      int index = path.indexOf("//");
      if (index == -1) {
        break;
      }
      path = path.substring(0, index) + path.substring(index + 1);
    }
    return path;
  }

  @Benchmark
  public String effective(Data data) {
    StringBuilder path = new StringBuilder(data.path);
    while (true) {
      int index = path.indexOf("//");
      if (index == -1) {
        break;
      }
      path.deleteCharAt(index);
    }
    return path.toString();
  }

  @State(Scope.Thread)
  public static class Data {
    private final String path = "/home/" + "/path";
  }
}
```
and got the following results on my machine
```
JDK 8

effective                          avgt    47.882 ±   1.801   ns/op
ineffective                        avgt    51.620 ±   0.907   ns/op
effective:·gc.alloc.rate.norm      avgt   136.000 ±   0.001    B/op
ineffective:·gc.alloc.rate.norm    avgt   224.000 ±   0.001    B/op

JDK 11

effective                          avgt    33.949 ±   0.188   ns/op
ineffective                        avgt    47.735 ±   0.519   ns/op
effective:·gc.alloc.rate.norm      avgt   104.000 ±   0.001    B/op
ineffective:·gc.alloc.rate.norm    avgt   152.000 ±   0.001    B/op
```
On longer strings I think we'll have even better improvement.

rstoyanchev (approved):
Thanks!

PR #26025, opened 2020-11-03
Add FullyQualifiedAnnotationBeanNameGenerator.INSTANCE

evpaassen:
Spring 5.2.3 (b4c91e7dac91c6176a5412ed930d2c048cc5c42f) introduces the `FullyQualifiedAnnotationBeanNameGenerator`, but does not override/hide the static `INSTANCE` field from its parent, which seems confusing to me. Someone might access `FullyQualifiedAnnotationBeanNameGenerator.INSTANCE` and get a `AnnotationBeanNameGenerator` instead. Just like I did. 😅 

This PR aims to solve that by adding a static `INSTANCE` field to `FullyQualifiedAnnotationBeanNameGenerator`, like its parent does too.


evpaassen:
Thanks for merging! :)

PR #26050, opened 2020-11-09
Fix wrong reference in UrlPathHelper.removeSemicolonContentInternal()

izeye:
This PR fixes a wrong reference in `UrlPathHelper.removeSemicolonContentInternal()`.

This PR also changes to short-circuit when `slashIndex` is -1.


jhoeller:
Good catch, that was an accidental leftover from #26012. Thanks for the timely PR!

PR #26058, opened 2020-11-10
Solve duplicate registration of beans of type MergedBeanDefinitionPostProcessor.

hzmpay:
Solve duplicate registration of beans of type MergedBeanDefinitionPostProcessor,although it is deleted and then added in the addBeanPostProcessor process, it is not necessary.


snicoll:
@hzmpay thanks for the PR but we can't accept such a code change in the core container that isn't exercised by additional tests. If you're experiencing a problem, please open a new issue with a small sample that we can use to reproduce the problem and we can take it from there.

PR #26070, opened 2020-11-11
Polish Javadoc for InjectionMetadata.forElements()

izeye:
This PR polishes `InjectionMetadata.forElements()` by removing an unnecessary branch and updating outdated Javadoc.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
@@ -141,8 +141,7 @@ public class InjectionMetadata {
 	 * Return an {@code InjectionMetadata} instance, possibly for empty elements.
 	 * @param elements the elements to inject (possibly empty)
 	 * @param clazz the target class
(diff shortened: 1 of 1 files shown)


jhoeller:
Good catch, the javadoc is definitely out of date now. However, with respect to the implementation, the use of the empty list constant is intentional since it reduces the memory footprint for the common empty case... the next best thing we can do after using the `InjectionMetadata.EMPTY` constant itself.

izeye:
@jhoeller Thanks for the feedback! I just thought that using the provided empty collection and using the empty list constant are the same in terms of memory footprint. Am I missing something?

jhoeller:
The provided collection is the fresh result from a new search, so a new instance every time. Since `InjectionMetadata` is being cached, we'd make each such newly built empty collection instance long-lived. Replacing it with an empty constant makes the cached empty `InjectionMetadata` instances all refer to a single long-lived collection constant. It'd be even better to replace `InjectionMetadata` itself with a constant but that's unfortunately not possible due to the distinct class for the refresh check.

izeye:
@jhoeller Thanks for the explanation! Sorry, I just saw the code snippet only and didn't know the caching part. I reverted it.

PR #26076, opened 2020-11-12
Make tests meaningful in DefaultListableBeanFactoryTests

limo520:


limo520:
such as assertThat(myself).isEqualTo(myself)

PR #26077, opened 2020-11-12
Avoid full singleton lock for getSingleton(beanName, singletonFactory)

hzmpay:
Quick check for existing instance without full singleton lock.


jhoeller:
That variant of `getSingleton` is not exposed in the registry/factory interfaces and is only intended to be used in a complex retrieval algorithm where a quick check should be performed through a plain `getSingleton(beanName)` call first anyway. Introducing another quick check here would effectively introduce unnecessary overhead in the `AbstractBeanFactory.doGetBean` algorithm, so I'd rather leave it as-is.

PR #26087, opened 2020-11-13
Add "application/java-archive" MediaType

kashike:
I have found myself using this type in a few projects, so I figured I would throw a PR up for it.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/MediaType.java
+++ b/spring-web/src/main/java/org/springframework/http/MediaType.java
@@ -387,6 +387,18 @@ public class MediaType extends MimeType implements Serializable {
 	 */
 	public static final String TEXT_XML_VALUE = "text/xml";
 
+	/**
+	 * Public constant media type for {@code application/java-archive}.
+	 * @since ?.?.?
+	 */
+	public static final MediaType APPLICATION_JAVA_ARCHIVE;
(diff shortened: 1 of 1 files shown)


spring-web/src/main/java/org/springframework/http/MediaType.java
```diff
@@ -387,6 +387,18 @@
 	 */
 	public static final String TEXT_XML_VALUE = "text/xml";
 
+	/**
+	 * Public constant media type for {@code application/java-archive}.
+	 * @since ?.?.?
```
kashike, line 392: Will change this once the version is known.

spring-web/src/main/java/org/springframework/http/MediaType.java
```diff
@@ -387,6 +387,18 @@
...
+	 */
+	public static final MediaType APPLICATION_JAVA_ARCHIVE;
+
+	/**
+	 * A String equivalent of {@link MediaType#APPLICATION_JAVA_ARCHIVE}.
+	 * @since ?.?.?
```
kashike, line 398: Will change this once the version is known.

snicoll:
Thanks for the PR but the `MediaType` class in the core framework is not meant to define all possible media types out there.

PR #26100, opened 2020-11-17
Avoid char array creation in AbstractAspectJAdvice.isVariableName()

izeye:
This PR changes to avoid `char` array creation in `AbstractAspectJAdvice.isVariableName()`.


snicoll:
Well spotted, thanks again @izeye.

PR #26101, opened 2020-11-17
Reduce overhead of char[] creation

mdeinum:
There are more locations which could benefit from not using a
toCharArray on a String, but rather use the charAt method from
the String itself. This to prevent an additional copy of the
char[] being created.


mdeinum:
Additional note, while inspecting the `AspectJAdviceParameterNameDiscoverer` I noticed that it might be beneficial to extract the `isVariableName` method from the `AbstractAspectJAdvice` to a helper/utility class and use that from both classes. They share the same logic to determine if something is a variable name.

PR #26107, opened 2020-11-17
Re-use the isVariableName method

mdeinum:
Prior to this change the checks for isVariableName were duplicated
in 2 different locations. The logic has been moved to a utility class
to allow for re-use in those places and so that it benefits from any
optimizations that are done.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
+++ b/spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
@@ -351,15 +351,7 @@ public abstract class AbstractAspectJAdvice implements Advice, AspectJPrecedence
 	}
 
 	private static boolean isVariableName(String name) {
-		if (!Character.isJavaIdentifierStart(name.charAt(0))) {
-			return false;
-		}
-		for (int i = 1; i < name.length(); i++) {
-			if (!Character.isJavaIdentifierPart(name.charAt(i))) {
-				return false;
-			}
-		}
-		return true;
+		return AspectJProxyUtils.isVariableName(name);
 	}
 
 
--- a/spring-aop/src/main/java/org/springframework/aop/aspectj/AspectJAdviceParameterNameDiscoverer.java
+++ b/spring-aop/src/main/java/org/springframework/aop/aspectj/AspectJAdviceParameterNameDiscoverer.java
@@ -470,22 +470,10 @@ public class AspectJAdviceParameterNameDiscoverer implements ParameterNameDiscov
 	 */
 	@Nullable
 	private String maybeExtractVariableName(@Nullable String candidateToken) {
-		if (!StringUtils.hasLength(candidateToken)) {
-			return null;
-		}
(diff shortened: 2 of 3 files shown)


spring-aop/src/main/java/org/springframework/aop/aspectj/VariableNameUtil.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Utility class containing the logic to check if a string could be a Java identifier.
+ * variable name.
+ *
+ * @author Marten Deinum
+ * @since 5.3.2
+ * @see Character#isJavaIdentifierStart(char)
+ * @see Character#isJavaIdentifierPart(char)
+ */
+public abstract class VariableNameUtil {
```
sbrannen, line 30:
```suggestion
abstract class VariableNameUtil {
```

There's no need to make this utility class `public` -- right?

spring-aop/src/main/java/org/springframework/aop/aspectj/VariableNameUtil.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.aop.aspectj;
+
+import org.springframework.util.StringUtils;
+
+/**
+ * Utility class containing the logic to check if a string could be a Java identifier.
+ * variable name.
```
sbrannen, line 23: Did you mean this?

```suggestion
 * Utility class containing the logic to check if a string could be a Java
 * variable name.
```

spring-aop/src/main/java/org/springframework/aop/aspectj/VariableNameUtil.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2020-2020 the original author or authors.
```
sbrannen, line 2:
```suggestion
 * Copyright 2002-2020 the original author or authors.
```

jhoeller:
Rather than introduce a new utility class, couldn't we add that `isVariableName` method to the existing `AspectJProxyUtils`? Since both of the callers live in the same package, we could even declare it as package-visible there.

sbrannen (changes requested):

sbrannen:
In light of @jhoeller's simultaneous review (https://github.com/spring-projects/spring-framework/pull/26107#issuecomment-728952313), you may ignore my review in favor of his suggestions.

spring-aop/src/main/java/org/springframework/aop/aspectj/VariableNameUtil.java
mdeinum: Not really no. I initially figured to make it package protected/default scoped. But as most utility classes are `public` I figured lets make it public.

mdeinum:
> Rather than introduce a new utility class, couldn't we add that `isVariableName` method to the existing `AspectJProxyUtils`? Since both of the callers live in the same package, we could even declare it as package-visible there.

Adding it in that place was my first thought, didn't consider making it package visible. Will move it there and force push.

PR #26114, opened 2020-11-18
Add ResponseSpec#toEntityFlux overload that accepts BodyExtractor

### LINKED ISSUE #26099: Add toEntityFlux with BodyExtractor to WebClient.ResponseSpec
**Affects:** Spring WebFlux 5.3.1

---

I started a conversation over on https://github.com/spring-projects/spring-framework/issues/26023#issuecomment-727119896 but it feels more appropriate to open a new issue.

If the plan is to remove the `WebClient#exchange` method entirely (as described [here](https://github.com/spring-projects/spring-framework/issues/25751)) how can we solve for those cases that aren't currently solved for using the alternative methods?

I have two scenarios in my case base which seem to require `WebClient#exchange`:

The first scenario is a series of endpoints which more or less act as a proxy to an upstream service.  We use `Flux<DataBuffer>` to stream the raw data to/from the client.  In this case `retrieve().entityToFlux(DataBuffer.class)` looks like it comes close to satisfying this use case, but I don't see a way with `ResponseSpec` to disable the default status code handling entirely, which is what I need.

The second scenario are some endpoints where we make an upstream service request and get back a large JSON response containing an array nested in a root object, e.g.:

```json
{
    "response": {
        "entities": [ ]
    }
}
```

We're currently using `exchange` with a custom `BodyExtractor<Flux<T>>` which can scan into the JSON response to the array and stream the individual elements which we return in a `ResponseEntity<Flux<T>>` so that we can also return the response headers and status code.  We then flow this `Flux<T>` back to the controller where it is mapped to transform each entity and stream it down to the client.  See #24951 for more details.

I've not fully evaluated where we're using `exchange` and whether or not they all fit into these two cases so I might be missing some scenarios.

HaloFour:
See #26099.

Adds an overload to the `ResponseSpec` interface to allow interpreting the response as a `ResponseEntity<Flux<T>>` using a `BodyExtractor<Flux<T>, ? super ClientHttpResponse>` to decode the body.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
@@ -43,11 +43,13 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.client.reactive.ClientHttpRequest;
+import org.springframework.http.client.reactive.ClientHttpResponse;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
+import org.springframework.web.reactive.function.BodyExtractor;
 import org.springframework.web.reactive.function.BodyInserter;
 import org.springframework.web.reactive.function.BodyInserters;
 import org.springframework.web.util.UriBuilder;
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@HaloFour Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26114)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26114) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@HaloFour Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26114)!

rstoyanchev:
This is merged now, thanks!

PR #26125, opened 2020-11-20
Use Collection View instead of a separate collection

bananayong:
- return a keySet view of map instead of cacheNames
- remove cacheNames
- simplify get-if-put -> computeIfAbsent
- remove synchronized block


jhoeller:
The main reason for the separate `cacheNames` collection is the ordering of cache names which can be important for startup-time cache registration at least, not so much for runtime-added caches. We generally recommend startup-time cache definitions, so it seems sensible to preserve the current arrangement.

If we were to give up on cache name ordering, your PR would make sense, of course. Thanks for your efforts, in any case!

bananayong:
Thank you for your kind explanation!
I missed javadoc of updateCacheNames method.

PR #26126, opened 2020-11-20
Delegate to AopUtils.invokeJoinpointUsingReflection from AbstractAspectJAdvice

polarbear567:
Code can be reused to make the code more concise.


sbrannen:
Thanks for submitting this PR to address that outstanding TODO.

After reviewing the code in question, we have decided to keep the code as-is, despite the code duplication.

The rationale is that the exception message in `AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs()` provides additional context via the pointcut expression, and we would lose that additional context if we simply delegate to `AopUtils.invokeJoinpointUsingReflection()`. We _could_ introduce an overloaded variant of `AopUtils.invokeJoinpointUsingReflection()` that accepts an additional argument to provide the additional context for the exception message, but we don't think that would be the best solution for this particular use case.

In light of that, we will simply remove the TODO.

PR #26133, opened 2020-11-23
new interface exposes inner response for testing

alexfeigin:
In my production setup, I use `ServerResponse.async`, and in tests, I need to access the wrapped 
 `CompletableFuture<ServerResponse>`. right now the only way to access is through reflection, and that depends on the field name remaining the same in the future.

This pull requests introduces interface `AsyncServerResponse` with access to inner `futureResponse`. and the class previously called AsyncServerResponse, is now `DefaultAsyncServerResponse`.

This is the same as with `EntityResponse`. that exposes the entity, of a `DefaultEntityResponse<T>`.

I realize that a user of this interface can act upon the inner future response, but to get at it, the user will have to manually cast to `AsyncServerResponse` and by that, take responsibility for their actions.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/function/AsyncServerResponse.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/function/AsyncServerResponse.java
@@ -18,8 +18,7 @@ package org.springframework.web.servlet.function;
 
 import java.io.IOException;
 import java.time.Duration;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
+import java.util.concurrent.*;
 import java.util.function.Function;
 
 import javax.servlet.ServletException;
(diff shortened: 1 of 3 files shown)


alexfeigin:
sorry for the formatting changes, out of habit.. if one of the maintainers can reformat back the the spring standard that would be helpful @poutsma

poutsma:
I am reluctant to expose the completable future, because it requires us to keep using that type forever, while we want to be able to use a asynchronous approach when necessary. In other words: the fact that the `AsyncServerResponse` uses a `CompletableFuture` is a implementation detail of that class.

As you've noted, the `EntityResponse` does expose its entity, which can also be a `CompletableFuture` to have a asynchronous body, i.e. a `EntityResponse<CompletableFuture<?>>`. So perhaps you can get access to the future in your tests this way?

alexfeigin:
My production code returns `ServerResponse` objects.. and the implementation for some returns immediate responses: some entity, some default responses with redirects, and some times the responses are async using the `ServerResponse.async`

There is no way to cast a ServerResponse created with `ServerResponse.async` to `EntityResponse`.. unless we will provide a spring web util method to make the translation, as of course, `AsyncServerResponse` does not implement `EntityResponse<CompletableFuture<ServerResponse>>`...

I understand your concerns about the current and future implementations of `AsyncServerResponse`. Maybe we should instead allow for a generic `AsyncServerResponse<T>` where a user that is aware of the implementation (i.e the user that used ServerResponse.async with a CompletableFuture) can get the Object `capture of ?` and cast it back to the CompletableFuture<ServerResponse>

alexfeigin:
I had a new idea, with your concerns in mind, please take a look at the change @poutsma.

Instead of making the interface `AsyncServerResponse` expose the inside implementation details, I have exposed a concrete "toSync" blocking `get` method. it will return a finished ServerResponse, this is good for tests, as they will likely need to block to check the result either way.

let me know what you think of this iteration.

poutsma:
I like the idea of the blocking `get` method, and will merge this PR for 5.3.2.

poutsma:
Manually merged, see https://github.com/spring-projects/spring-framework/commit/8f0ad73bfdb855467bf62f7c00e53868fe7b6e5a.

Thank you for providing a PR!

PR #26173, opened 2020-11-29
Add overloaded send methods of client StompSession

topikachu:
The methods allow a post processor to post-process or modify the created message when sending.


topikachu:
Hi, I add two overloaded send methods so I can calculate the Message Authentication Code (MAC) against the serialized payload.

rstoyanchev:
These are essentially post-conversion hooks. The same can be intercepted by wrapping the `MessageConverter`.

PR #26203, opened 2020-12-03
Fix comments to match behaviour in Tokenizer#lexNumericLiteral

limo520:
Real numbers must have leading digits. And SIGN should occur 0 or 1 time at EXPONENT_PART.


limo520:
Does [SpringExpressions.g](https://github.com/spring-projects/spring-framework/blob/master/spring-expression/src/main/resources/org/springframework/expression/spel/generated/SpringExpressions.g) be modified？I don't know how it works.

PR #26206, opened 2020-12-03
Allow passing Class<?> to resolveMethodByExceptionType in ExceptionHandlerMethodResolver

neiser:
This fits better to the return value of `java.lang.reflect.Method.getExceptionTypes` which only returns items of `Class<?>` and not of `Class<? extends Throwable>` (which I don't know why that's the case, but we should just cope with this)

A real world usage of `ExceptionHandlerMethodResolver#resolveMethodByExceptionType` is [here](https://github.com/qaware/openapi-generator-for-spring/blob/master/openapi-generator-for-spring-test/src/test/java/de/qaware/openapigeneratorforspring/test/app32/App32Configuration.java#L50).

This PR would remove this ugly uncheckedCast for users of `ExceptionHandlerMethodResolver`. Still the method `resolveMethodByExceptionType` throws `IllegalArgumentException` if anything else than Class<? extends Throwable> is passed to the method.

I've also added some test cases to show that it works as expected.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/annotation/ExceptionHandlerMethodResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/method/annotation/ExceptionHandlerMethodResolver.java
@@ -143,17 +143,27 @@ public class ExceptionHandlerMethodResolver {
 	/**
 	 * Find a {@link Method} to handle the given exception type. This can be
 	 * useful if an {@link Exception} instance is not available (e.g. for tools).
-	 * @param exceptionType the exception type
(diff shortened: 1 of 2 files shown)


neiser:
@sbrannen Is there any chance that this PR is merged? I'd rebase the PR to fix the merge conflicts in the meantime.

rstoyanchev:
@neiser we've discussed this among the team and decided that the current declaration provide good guidance and we do not want to relax the signature in this way. Furthermore this is an SPI that should be used infrequently, which makes it a good trade-off vs requiring a cast.

PR #26211, opened 2020-12-04
Correct satisfies statements in SpelParserTests.generalExpressions

limo520:


sbrannen:
Good catch!

Thanks

Issue #26174, opened 2020-11-29
Spring messaging support hints on client and server side for RSocket


rudy2steiner:
**Affects:** \<since 5.2>

---
To my knowledge,  for RSocket request, the hints on client and server side are both an empty map.  

 Even I add a metadata into header by RSocketRequester.RequestSpec#metadata, there is no way for me to use the metadata on server side.

To support hints on client and server side  will in favor of user to pass encode/decode context. 

And I'll pull a request if any one expect this change.For how to implement it, we could discuss on below. 

On RSocket client, in DefaultRequestSpec#encodeData method uses an empty hints to call encoder:
```
private <T> DataBuffer encodeData(T value, ResolvableType elementType, @Nullable Encoder<?> encoder) {
			if (encoder == null) {
				elementType = ResolvableType.forInstance(value);
				encoder = strategies.encoder(elementType, dataMimeType);
			}
			return ((Encoder<T>) encoder).encodeValue(
					value, bufferFactory(), elementType, dataMimeType, EMPTY_HINTS);
		}
```

On RSocket server side, in PayloadMethodArgumentResolver#decodeContent method uses an empty hints to call decoder:
```
private Mono<Object> decodeContent(MethodParameter parameter, Message<?> message,
			boolean isContentRequired, Flux<DataBuffer> content, MimeType mimeType) {

		ResolvableType targetType = ResolvableType.forMethodParameter(parameter);
		Class<?> resolvedType = targetType.resolve();
		ReactiveAdapter adapter = (resolvedType != null ? getAdapterRegistry().getAdapter(resolvedType) : null);
		ResolvableType elementType = (adapter != null ? targetType.getGeneric() : targetType);
		isContentRequired = isContentRequired || (adapter != null && !adapter.supportsEmpty());
		Consumer<Object> validator = getValidator(message, parameter);

		Map<String, Object> hints = Collections.emptyMap();

		for (Decoder<?> decoder : this.decoders) {
			if (decoder.canDecode(elementType, mimeType)) {
				if (adapter != null && adapter.isMultiValue()) {
					Flux<?> flux = content
							.map(buffer -> decoder.decode(buffer, elementType, mimeType, hints))
							.onErrorResume(ex -> Flux.error(handleReadError(parameter, message, ex)));
					if (isContentRequired) {
						flux = flux.switchIfEmpty(Flux.error(() -> handleMissingBody(parameter, message)));
					}
					if (validator != null) {
[...]

PR #26220, opened 2020-12-05
Support hints for codec on client and server side

rudy2steiner:
#26174

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/reactive/PayloadMethodArgumentResolver.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/reactive/PayloadMethodArgumentResolver.java
@@ -20,7 +20,6 @@ import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import java.util.function.Consumer;
 
 import org.apache.commons.logging.Log;
@@ -226,13 +225,11 @@ public class PayloadMethodArgumentResolver implements HandlerMethodArgumentResol
 		isContentRequired = isContentRequired || (adapter != null && !adapter.supportsEmpty());
 		Consumer<Object> validator = getValidator(message, parameter);
 
-		Map<String, Object> hints = Collections.emptyMap();
-
 		for (Decoder<?> decoder : this.decoders) {
 			if (decoder.canDecode(elementType, mimeType)) {
 				if (adapter != null && adapter.isMultiValue()) {
 					Flux<?> flux = content
-							.map(buffer -> decoder.decode(buffer, elementType, mimeType, hints))
+							.map(buffer -> decoder.decode(buffer, elementType, mimeType, message.getHeaders()))
 							.onErrorResume(ex -> Flux.error(handleReadError(parameter, message, ex)));
 					if (isContentRequired) {
 						flux = flux.switchIfEmpty(Flux.error(() -> handleMissingBody(parameter, message)));
@@ -245,7 +242,7 @@ public class PayloadMethodArgumentResolver implements HandlerMethodArgumentResol
 				else {
 					// Single-value (with or without reactive type wrapper)
 					Mono<?> mono = content.next()
-							.map(buffer -> decoder.decode(buffer, elementType, mimeType, hints))
+							.map(buffer -> decoder.decode(buffer, elementType, mimeType,  message.getHeaders()))
 							.onErrorResume(ex -> Mono.error(handleReadError(parameter, message, ex)));
 					if (isContentRequired) {
 						mono = mono.switchIfEmpty(Mono.error(() -> handleMissingBody(parameter, message)));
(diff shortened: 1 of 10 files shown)


pivotal-issuemaster:
@rudy2steiner Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26220)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26220) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@rudy2steiner Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26220)!

rudy2steiner:
@rstoyanchev  would you mind helping me review this PR!

rstoyanchev (changes requested):
@rudy2steiner thanks for those changes but I'm afraid it's not going to work out in its present form. What I had in mind was a lot less extensive, simply passing `MessageHeaders` in `PayloadMethodArgumentResolver` as a hint under a well-known key so that a custom `Decoder` can take into account message headers, including extracted metadata. 

I don't think `RSocketRequester` should be involved in any of these changes. Composite metadata already supports sending metadata of any mime type, and there is no reason for it to know what kind of metadata it is.

rudy2steiner:
> @rudy2steiner thanks for those changes but I'm afraid it's not going to work out in its present form. What I had in mind was a lot less extensive, simply passing `MessageHeaders` in `PayloadMethodArgumentResolver` as a hint under a well-known key so that a custom `Decoder` can take into account message headers, including extracted metadata.
> 
> I don't think `RSocketRequester` should be involved in any of these changes. Composite metadata already supports sending metadata of any mime type, and there is no reason for it to know what kind of metadata it is.

Yes, I see what you have in mind. Once we pass headers as hints to decoder, we could add any metadata with self-defined mime on client side, and the metadata will be available for server decoder. But how to keep encoder of client and decoder of server have same hints context is a problem.

rudy2steiner:
May we could introduce a new method on RequestSpec:
data with hints for local encoders
```
RetrieveSpec data(Object data, Map<String,Object> hints );
```

and hints for remote decoder could implement by add a metadata use MetadataSpec#metadata

In this way, only a few changes and providing the ability to attach hints for encoder and decoder

rstoyanchev:
> But how to keep encoder of client and decoder of server have same hints context is a problem.

I don't follow what you mean. Wouldn't the client just add some metadata which then the server can use for decoding?

rudy2steiner:
> I don't follow what you mean. Wouldn't the client just add some metadata which then the server can use for decoding?

I think we should provide the ability for add hints for client encoder as the same time.

rstoyanchev:
I think it would be useful if you could provide more details of what you are trying to achieve more concrete. That would help to think about possible solutions.

rudy2steiner:
> I think it would be useful if you could provide more details of what you are trying to achieve more concrete. That would help to think about possible solutions.

In my case, the hints will be used to affect behavior of encoder and decoder, but found that hints are always empty. I need to rewrite a lot of code to implement the requirement by myself, but I think Spring Messaging should built-in support user to add hints for encoder and decoder for per message , the hints are possibly different for per message.

We could discuss how to implement this feature if you agree with this would benefit for user.

rstoyanchev:
I'm afraid that doesn't tell me much more. I gather that you want to pass hints but what's not clear is why you want to pass hints into a client `Encoder` that would then be serialized and passed to the server side as well. By "provide more details" I meant provide a concrete scenario to illustrate the use case.

rudy2steiner:
I have messages from different sources and be wrapped into instance of same class, but I want to serialize it using different algorithm for message from different source .

If I can pass a hint that indicate where the message from,  then selecting corresponding algorithm to encode on client  and decode it on server.

So I need a way to pass a hint for local encoder and remote decoder. 

As i can seen, Hints as a metadata can be abstracted from headers and then passing it to decoder.
at the same time, I need pass the same hint to local encoder.

 Spring messaging has passing the hints into encoder and decoder, but it's alway empty and can't be customized by users.

rstoyanchev:
Sorry for the delay and thanks for the extra details. This is quite helpful although it's still a bit abstract. It would help to have a little more detail about the different encoding algorithms. Do they result in a different format on the wire for example, which could potentially be handled by using different data MIME types. Or perhaps something in the Object structure or type could be used to differentiate?

rudy2steiner:
Currently,  Data Mime is connection  not  message level which will double connections if using different data MIME types.
Object structure  isn't a general solution for this problem.

On production environment, we use the first solution as a temporary solution, but we should have more elegant solution.

* Supporting messaging level data MIME feature is a perfect solution in the future.
* Supporting hints on client and server also benefits for solving my problem.

Anyway, Supporting hints on client and server is necessary.

I'll try to simplify my changes.

rstoyanchev:
> Currently, Data Mime is connection not message level which will double connections if using different data MIME types.

Keep in mind there is this https://github.com/rsocket/rsocket/blob/master/Extensions/PerStreamDataMimeTypesDefinition.md which we need to add support for. Once that's implemented, doing the same via hints would amount to a workaround, so I think we should turn this into a ticket to support per-stream/request MIME types.

rudy2steiner:
Is there any body has work on it,  I'd like try if no body ?
So in this pr we will only support hints on server side?

rstoyanchev:
I'd rather focus on support for per-request data MIME types first and close this. I've created the issue #26379. You can give it a give it a try if you want. I don't have a good sense if it would work out well for a contribution.

rudy2steiner:
Ok, we could polish this feature  in the future.

PR #26240, opened 2020-12-08
remove not necessary judge and fields

xichaodong:
Because of the method of getReaderContext().error() will throw BeanDefinitionParsingException, and the following code will not be executed if an exception is thrown,So I think maybe can remove valid flag and judge.


quaff:
It depends on `ProblemReporter` implementation, user implementation may log error instead of throw exception.

cuspymd (commented):
According to the documents in the `ProblemReporter` interface below, `error()` method may not raise an exception depending on the implementation.
```
public interface ProblemReporter {
...
	/**
	 * Called when an error is encountered during the parsing process.
	 * <p>Implementations may choose to treat errors as fatal.
	 * @param problem the source of the error (never {@code null})
	 */
	void error(Problem problem);
...
}

```

snicoll:
Thanks everyone.

@xichaodong we can't merge  a PR in the core container without additional tests and the objections raised are legit.

Issue #26258, opened 2020-12-11
Fail to parse File URI with UriTemplateHandler


mvpcortes:
Hi!

I thinl I found a exception after update to spring-boot 2.4.0, 2.4.1.

A cannot parse a file uri using UriTemplateHandler. Before, in 2.3.x, I could do it.

A test example is:

```
 @Test
    public void verify() throws MalformedURLException {
        final UriTemplateHandler uriTemplateHandler;

        DefaultUriBuilderFactory uriFactory = new DefaultUriBuilderFactory();
        uriFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.URI_COMPONENT);  // for backwards compatibility..
        uriTemplateHandler = uriFactory;

        URL url = uriTemplateHandler.expand(file.toURI().toString(), Collections.emptyMap()).toURL();

        assertThat(url.toString()).isEqualTo(file.toURI().toString());
    }
```

file  is a File object with value (toString()): /tmp/junit5116856254058695014/arquivo.txt
Sorry If I made some mistake.

Thanks

encircled:
Hi,
caused by a new validation intended for ipv6, I've raised the PR fixing that

rstoyanchev:
Yes it looks like this dates back to a change for 5.3 under #25334 that adds a similar check to `fromUriString` as in `fromHttpUrl`. However according to RFC 8320, section 2.2, the semantics of URI authority is defined by the scheme. So while for "http/https" a host is expected, for other schemes it might not be.

PR #26272, opened 2020-12-15
Fix UriComponentsBuilder validation of file URI #26258

encircled:
Fixes gh #26258


rstoyanchev:
I have addressed this slightly differently by applying the check for HTTP URLs only and not making further assumptions about any other URIs. Thanks for the PR in any case.

PR #26291, opened 2020-12-17
Update DefaultListableBeanFactory's determineHighestPriorityCandidate

ghost:
Update method determineHighestPriorityCandidate of DefaultListableBeanFactory. Runs define the same priority, but highestPriority.


quaff:
I don't understand what problem you are trying to resolve.

york-lee:
> I don't understand what problem you are trying to resolve.

As I understand it, the purpose of this method is to get the highest priority bean, so we just need to ensure that the highest priority is not repeated. There is no need to restrict other low-priority beans.

quaff:
> > I don't understand what problem you are trying to resolve.
> 
> As I understand it, the purpose of this method is to get the highest priority bean, so we just need to ensure that the highest priority is not repeated. There is no need to restrict other low-priority beans. Or some other purpose That I don't understand？

Still not clear, are you fix a bug or improve code readability?

york-lee:
> > > I don't understand what problem you are trying to resolve.
> > 
> > 
> > As I understand it, the purpose of this method is to get the highest priority bean, so we just need to ensure that the highest priority is not repeated. There is no need to restrict other low-priority beans. Or some other purpose That I don't understand？
> 
> Still not clear, are you fix a bug or improve code readability?

I think it's a bug.

quaff:
> > > > I don't understand what problem you are trying to resolve.
> > > 
> > > 
> > > As I understand it, the purpose of this method is to get the highest priority bean, so we just need to ensure that the highest priority is not repeated. There is no need to restrict other low-priority beans. Or some other purpose That I don't understand？
> > 
> > 
> > Still not clear, are you fix a bug or improve code readability?
> 
> I think it's a bug.

Then you should include a test case in this PR.

york-lee:
> > > > > I don't understand what problem you are trying to resolve.
> > > > 
> > > > 
> > > > As I understand it, the purpose of this method is to get the highest priority bean, so we just need to ensure that the highest priority is not repeated. There is no need to restrict other low-priority beans. Or some other purpose That I don't understand？
> > > 
> > > 
> > > Still not clear, are you fix a bug or improve code readability?
> > 
> > 
> > I think it's a bug.
> 
> Then you should include a test case in this PR.

OK, I will complete it.

york-lee:
suspend

Issue #25650, opened 2020-08-26
Replace remaining usage of LinkedList with ArrayList/ArrayDeque


stsypanov:
As of 2020 there's no need to use `LinkedList`:
- empty `ArrayList` is more lightweight than `LinkedList`:
```java
@State(Scope.Thread)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Fork(jvmArgsAppend = {"-Xms2g", "-Xmx2g"})
public class ListBenchmark {
  @Benchmark
  public Object arrayList() {
    return new ArrayList<>();
  }

  @Benchmark
  public Object linkedList() {
    return new LinkedList<>();
  }
}
```
which on JDK 11 gives
```
ListBenchmark.arrayList                                    avgt   15     4.011 ±   0.120   ns/op
ListBenchmark.arrayList:·gc.alloc.rate.norm                avgt   15    24.002 ±   0.001    B/op

ListBenchmark.linkedList                                   avgt   15     4.243 ±   0.181   ns/op
ListBenchmark.linkedList:·gc.alloc.rate.norm               avgt   15    32.002 ±   0.001    B/op
```
if we put one element then `LinkedList` wins:
```
enchmark                                                  Mode  Cnt     Score     Error   Units
ListBenchmark.arrayList                                    avgt   15     9.188 ±   0.831   ns/op
ListBenchmark.arrayList:·gc.alloc.rate.norm                avgt   15    80.005 ±   0.001    B/op
ListBenchmark.linkedList                                   avgt   15     8.148 ±   0.285   ns/op
ListBenchmark.linkedList:·gc.alloc.rate.norm               avgt   15    56.004 ±   0.001    B/op
```
in other cases `LinkedList` looses:
```
2 elements
Benchmark                                                  Mode  Cnt     Score     Error   Units
ListBenchmark.arrayList                                    avgt   15    11.308 ±   0.290   ns/op
ListBenchmark.arrayList:·gc.alloc.rate.norm                avgt   15    80.006 ±   0.001    B/op
ListBenchmark.linkedList                                   avgt   15    11.898 ±   0.608   ns/op
ListBenchmark.linkedList:·gc.alloc.rate.norm               avgt   15    80.006 ±   0.001    B/op

3 elements
ListBenchmark.arrayList                                    avgt   15    14.763 ±   1.452   ns/op
[...]

Issue #25552, opened 2020-08-06
Optimize usage of "".equals()


sbrannen:
Investigate claims made in [Micro optimizations in Java. String.equals()](https://medium.com/javarevisited/micro-optimizations-in-java-string-equals-22be19fd8416) article.

sbrannen:
Related to #25553

PR #26316, opened 2020-12-23
Optimize memory allocations in StringUtils#cleanPath

knittl:
Investigating once again slow Spring Boot start times, I played around with Java Mission Control profiler and found out that `StringUtils#cleanPath` alone is responsible for 7% of all CPU time spent during setup of the application context. Looking at the implementation I was able to identify some potential hotspots (e.g. LinkedList). The application is currently using Spring Framework 5.1.10 and I was happy to find that some of these issues have already been fixed in more recent versions of Spring.

However there are still a few more opportunities to avoid memory allocations and copying of data:

* Set a sensible initial capacity for lists and string builders: optimize for the common cases. We will sometimes allocate a few bytes more than required, but at least we avoid re-allocating the backing array. For the StringBuilder we need to iterate the collection twice to calculate the capacity beforehand. But iterating is a lot cheaper compared to allocating memory two or more times.
* Re-use strings instead of concatenating their parts again: an earlier optimization introduced short-circuiting for a common case. We have computed the return value already before splitting it, so return it directly.
* Do not concatenate with the empty string: Future JDKs might alleviate this, but at least with Java 11, `return "" + somestring;` is slower than `return somestring;`. Obviously it is more expensive memory-wise due to the construction of the newly resulting string.

This PR relates to and builds on #24674, #25650, #25552, #25553, and others.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -695,7 +695,8 @@ public abstract class StringUtils {
 		}
 
 		String[] pathArray = delimitedListToStringArray(pathToUse, FOLDER_SEPARATOR);
-		Deque<String> pathElements = new ArrayDeque<>();
+		// we never require more elements than pathArray and in the common case the same number
+		Deque<String> pathElements = new ArrayDeque<>(pathArray.length);
 		int tops = 0;
 
 		for (int i = pathArray.length - 1; i >= 0; i--) {
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -733,7 +733,40 @@ public abstract class StringUtils {
 			pathElements.addFirst(CURRENT_PATH);
 		}
 
-		return prefix + collectionToDelimitedString(pathElements, FOLDER_SEPARATOR);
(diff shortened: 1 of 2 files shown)


vasilyl (approved):

knittl:
Is there still anything required from my part? Should I provide benchmarks to prove the optimization?

bclozel:
Closed with cc026fcb8ae5117

bclozel:
Thanks @knittl - this is now merged.
I ran a few benchmarks and it seems that this optimization helps both `StringUtils#delimitedListToStringArray` (+40% throughput) and `StringUtils#cleanPath` (+10% for the worst case).

knittl:
@bclozel thanks for merging, polishing, and benchmarks! I wonder how big the overhead of converting each collection item to a string when pre-computing the string builder size for non-string collections (in cc026fcb8ae51172f3d063e7ed07a32927e23d8e)? For string collections, it is obviously just a no-op, but for generic collections of a different type stringifying each of its elements twice might incur quite some memory pressure and runtime overhead (depending on the toString complexity)?

One of the reasons the method was duplicated was to have a specialization for string collections. One idea, I had was to pre-compute all string representations, copy them to a new collection, then use this collection to populate the string builder. With your JMH benchmarks (thanks!), copying the collection seems to incur quite some overhead itself.

PR #26339, opened 2021-01-03
Add null check for ExceptionHandlerMethodResolvers

refeccd:
Add null check to avoid subsequent locking and map operations


rstoyanchev:
Thanks for the pull request. I've handled this slightly differently by caching the no match.

Issue #26360, opened 2021-01-08
Add API to use ParameterizedTypeReference on RequestEntity Function on Get / Post function, like exchange.


Hubtwork:
**Affects:** 
<Spring Web 5.3.2.RELEASE>
<Spring Boot 2.2.4.RELEASE>
If this is not the right place for this issue, please let me know.

---

Hello

I think that **exchange, getForEntity, postForEntity function** which is in RestTemplate, Spring web Framework is designed to receive ResponseEntities from each http method request like below.
And there's no guide in docs ever, so users are using same purpose depending on their needs. 

```java
@Override
	public <T> ResponseEntity<T> exchange(String url, HttpMethod method,
			@Nullable HttpEntity<?> requestEntity, Class<T> responseType, Object... uriVariables)
			throws RestClientException {

		RequestCallback requestCallback = httpEntityCallback(requestEntity, responseType);
		ResponseExtractor<ResponseEntity<T>> responseExtractor = responseEntityExtractor(responseType);
		return nonNull(execute(url, method, requestCallback, responseExtractor, uriVariables));
	}
```

```java
@Override
	public <T> ResponseEntity<T> getForEntity(String url, Class<T> responseType, Object... uriVariables)
			throws RestClientException {

		RequestCallback requestCallback = acceptHeaderRequestCallback(responseType);
		ResponseExtractor<ResponseEntity<T>> responseExtractor = responseEntityExtractor(responseType);
		return nonNull(execute(url, HttpMethod.GET, requestCallback, responseExtractor, uriVariables));
	}
```

```java
@Override
	public <T> ResponseEntity<T> postForEntity(String url, @Nullable Object request,
			Class<T> responseType, Object... uriVariables) throws RestClientException {

		RequestCallback requestCallback = httpEntityCallback(request, responseType);
		ResponseExtractor<ResponseEntity<T>> responseExtractor = responseEntityExtractor(responseType);
		return nonNull(execute(url, HttpMethod.POST, requestCallback, responseExtractor, uriVariables));
	}
```

However, Spring Web supports **ParameterizedTypeReference** for mapping response to any Type like Collections in **exchange**. On the other hand, **getForEntity, postForEntity** are not supported. 

```java
@Override
	public <T> ResponseEntity<T> exchange(RequestEntity<?> entity, ParameterizedTypeReference<T> responseType)
			throws RestClientException {

		Type type = responseType.getType();
		RequestCallback requestCallback = httpEntityCallback(entity, type);
[...]

PR #26362, opened 2021-01-08
ADD getForEntity, postForEntity Overloading function using ParameterizedTypeReference

hubtwork:
GetForEntity, PostForEntity Overloading function using ParameterizedTypeReference

There's needs for usage-free to use ParameterizedTypeReference on any RequestEntity function.

ISSUE : https://github.com/spring-projects/spring-framework/issues/26360
MAKE : GetForEntity, PostForEntity Overloading function using ParameterizedTypeReference


rstoyanchev:
Thanks for the pull request but use of `ParameterizedTypeReference` is a less common case and it is a balance between having too many methods in `RestTemplate` vs convenience. If we overload these methods, it would be reasonable to expect others to be overloaded as well which makes 12 new methods.

For such cases please use the `exchange` methods.

PR #26393, opened 2021-01-16
Add support for java.util.Optional to GenericHttpMessageConverter #24498

### LINKED ISSUE #24498: Support Optional in Jackson 2 message converters

encircled:
Fixes #24498

Added unboxing of  `java.util.Optional`  to `AbstractHttpMessageConverter` and `AbstractGenericHttpMessageConverter`. If the object is not unboxed, the underlying message converter receives an instance of `java.util.Optional`, which was causing an issue described in the #24498 in case of Jackson converter. It will also allow to use `Optional<*>` in other message converters without explicitly registering a type converter.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/AbstractHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/AbstractHttpMessageConverter.java
@@ -23,9 +23,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 import org.apache.commons.logging.Log;
 
+import org.springframework.core.GenericTypeResolver;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpInputMessage;
 import org.springframework.http.HttpLogging;
@@ -45,6 +47,7 @@ import org.springframework.util.Assert;
  * @author Arjen Poutsma
  * @author Juergen Hoeller
  * @author Sebastien Deleuze
+ * @author Vladislav Kisel
  * @since 3.0
  * @param <T> the converted object type
  */
@@ -53,17 +56,23 @@ public abstract class AbstractHttpMessageConverter<T> implements HttpMessageConv
 	/** Logger available to subclasses. */
 	protected final Log logger = HttpLogging.forLogName(getClass());
 
+	/**
+	 * This converter type argument
+	 */
+	@Nullable
+	private Class<?> typeArgument;
+
 	private List<MediaType> supportedMediaTypes = Collections.emptyList();
 
 	@Nullable
 	private Charset defaultCharset;
-
 
 	/**
 	 * Construct an {@code AbstractHttpMessageConverter} with no supported media types.
 	 * @see #setSupportedMediaTypes
 	 */
 	protected AbstractHttpMessageConverter() {
+		resolveTypeArgument();
 	}
 
 	/**
@@ -72,6 +81,7 @@ public abstract class AbstractHttpMessageConverter<T> implements HttpMessageConv
 	 */
 	protected AbstractHttpMessageConverter(MediaType supportedMediaType) {
 		setSupportedMediaTypes(Collections.singletonList(supportedMediaType));
+		resolveTypeArgument();
 	}
 
 	/**
@@ -80,6 +90,7 @@ public abstract class AbstractHttpMessageConverter<T> implements HttpMessageConv
 	 */
 	protected AbstractHttpMessageConverter(MediaType... supportedMediaTypes) {
 		setSupportedMediaTypes(Arrays.asList(supportedMediaTypes));
+		resolveTypeArgument();
 	}
 
 	/**
@@ -92,6 +103,7 @@ public abstract class AbstractHttpMessageConverter<T> implements HttpMessageConv
 	protected AbstractHttpMessageConverter(Charset defaultCharset, MediaType... supportedMediaTypes) {
 		this.defaultCharset = defaultCharset;
 		setSupportedMediaTypes(Arrays.asList(supportedMediaTypes));
+		resolveTypeArgument();
 	}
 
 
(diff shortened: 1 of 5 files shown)


spring-web/src/main/java/org/springframework/http/converter/AbstractHttpMessageConverter.java
```diff
@@ -224,11 +227,23 @@ public HttpHeaders getHeaders() {
...
+			writeInternal(unboxed, outputMessage);
 			outputMessage.getBody().flush();
 		}
 	}
 
+	/**
+	 * Unbox the object value if it is stored in a container (like {@link java.util.Optional}.
+	 */
+	@SuppressWarnings("unchecked")
+	protected T unboxIfNeeded(T value) {
```
RemcoBuddelmeijer, line 239: Quick question about the unwrapping and returning type parameter: basically what ur doing here is convert a wrapped (Optional) value of type `T` to an unwrapped value and in the process of this changing the returned type. 
How are you going to be ensuring type safety? Especially for the abstract method `writeInternal` which may be implemented by anyone?

Let's say that you have type argument `Optional<String>` and the `unboxIfNeeded` method converts this type to `String`. This all while `writeInternal` and the rest of the code still expect `T` to be `Optional<String>`, this especially is an issue since you tell the compiler to expect type safety (by suppressing the warning). 
These seem like mismatching types which might, depending on the implementing code, throw a ClassCastException. This may fix the issue but you will now not know what the type of variable `t` really is.

_(I am fully aware that using Optional as a type argument is a bad practice, however I would value type safety over implementation of the user)_

encircled: Correct, in case when a converter has generic type `Optional<*>`, it would fail with the class cast exception. I will fix that

RemcoBuddelmeijer: I see that you are now "ensuring" type safety. However, there are still some issues which resolve around the second part of my comment.

In my comment I mentioned the following:
> This may fix the issue, but you will now not know what the type of variable t really is going to be. When you implement AbstractHttpMessageConverter by yourself, you are going to be having restrictions for a fix on Jackson's implementation. I would suggest not changing this class but look at the Jackson specific implementation (AbstractJackson2HttpMessageConverter) since you won't be needing the type safety here and it tackles this specific issue.

With this, I was referring to the strategy pattern used by these classes (`HttpMessageConverter`, `GenericHttpMessageConverter`, `AbstractHttpMessageConverter` and `AbstractGenericHttpMessageConverter`). These classes are to be implemented and allow the user to define, at run-time, the to be chosen strategy. 
Since this is the case, we do not want the user to be confused or to already have some sort of pre-processing on any implementation that has been made. Because this specific issue (the unboxing issue) lies within the `AbstractJackson2HttpMessageConverter` class and sub classes: why not fix it there as it will fit that specific use case? Now we are going to be introducing bugs and unexpected behavior for anyone implementing this base class in it's own converter, which is most certainly not what we would want.

encircled: As agreed here #24867, we want to avoid code duplication. You are still able to override the `unboxIfNeeded` in the impl in case you really don't want the `Optional` to be unwrapped

encircled:
@sbrannen just a kind reminder. Lets review&merge or close it if you think that such behaviour should not be part of the product

daniel-shuy:
Any update for this PR?

poutsma:
Thank for for submitting a PR, and our apologies for taking this long to resolve it.

We have decided to resolve the underlying issue differently than suggested in this PR,  in the Jackson converters rather than the generic converters. See https://github.com/spring-projects/spring-framework/commit/5502e6151967afaef7fa3d9d057f3e2d226a0c84 for more details.

sbrannen:
- Superseded by #24498

PR #26402, opened 2021-01-19
Make CommonsPool2TargetSource generic

quaff:


sbrannen:
Hi @quaff,

What is your use case for making `CommonsPool2TargetSource` parameterized?

quaff:
> Hi @quaff,
> 
> What is your use case for making `CommonsPool2TargetSource` parameterized?

It's worthy to make changes if full backward compatibility, for example, I can infer target type from java config
```java
@Bean
public CommonsPool2TargetSource<Foo> foo() {

}
```

sbrannen:
> for example, I can infer target type from java config

Why would you need to do that?

As stated in the Javadoc for `TargetSource`:

> Application developers don't usually need to work with `TargetSources` directly: this is an AOP framework interface.

In other words, the framework itself would not benefit from parameterizing `CommonsPool2TargetSource` -- `Object` is sufficient for the framework. So why are you working directly with a `TargetSource`?

PR #26436, opened 2021-01-24
Fix typo in URI Encoding section of reference docs

daisuzz:
> VALUES_ONLY: Does not encode the URI template and, instead, applies strict encoding to URI variables through UriUtils#encodeUriUriVariables prior to expanding them into the template.

I fixed `UriUtils#encodeUriUriVariables` into `UriUtils#encodeUriVariables` in  "[1.5.3. URI Encoding](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#web-uri-encoding)" section of reference docs.


sbrannen:
Good catch.

Thanks

PR #26442, opened 2021-01-25
Introduce 'idleReceivesPerTaskLimit' in DefaultMessageListenerContainer

### LINKED ISSUE #26195: Spring DMLC releasing worker tasks based upon a pure inactivity counter
What we need: Spring `DefaultMessageListenerContainer` releasing worker tasks based upon a pure inactivity counter. That is, the counter that counts how many times a "null" message has been received. For example, if the _receiveTimeout_ is set to 1 second and the (fictive, to be created) property _maxIdleReceivesPerTask_ is set to 60, the `DefaultMessageListenerContainer` would release the worker task after 60 times subsequently receiving a null value, and thus, be released 60 seconds after the last activity.

"But this already exists" : well, there is indeed _idleTaskExecutionLimit_ and the _maxMessagesPerTask_, but if my  understanding is correct, the _maxMessagesPerTask_ has the side effect that it also releases the task when *not* being idle. In those cases the task is immediately re-scheduled, but this is something we want to avoid. In our case we want to set _maxMessagesPerTask_ to "-1" but still let the worker task be released after a pre-configured amount of null messages have been received.

Is this change possible?

koen-serneels:
Closes GH-26195

### DIFF
--- a/spring-jms/src/main/java/org/springframework/jms/listener/DefaultMessageListenerContainer.java
+++ b/spring-jms/src/main/java/org/springframework/jms/listener/DefaultMessageListenerContainer.java
@@ -188,6 +188,8 @@ public class DefaultMessageListenerContainer extends AbstractPollingMessageListe
 
 	private int maxMessagesPerTask = Integer.MIN_VALUE;
 
+	private int idleReceivesPerTaskLimit = Integer.MIN_VALUE;
+
 	private int idleConsumerLimit = 1;
 
 	private int idleTaskExecutionLimit = 1;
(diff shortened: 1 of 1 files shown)


koen-serneels:
@jhoeller would it be possible to check if this change is ok/makes sense?

pivotal-issuemaster:
@koen-serneels Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26442)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26442) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@koen-serneels Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26442)!

koen-serneels:
@jhoeller any chance you have some time to check this? If sane/no other alternative we could already patch it ourselves, thanks.

jhoeller:
This is in master now in a revised fashion, ready to make it into the 5.3.5 release next Tuesday. Please give it a try with `5.3.5-SNAPSHOT´ (available from `https://repo.spring.io/snapshot/`) at your earliest convenience, double-checking that it works for you (in particular that you are seeing the desired effect between with and without a `idleReceivesPerTaskLimit` setting).

koen-serneels:
We'll try to check this asap. Thanks Jürgen.

PR #26446, opened 2021-01-26
Fix DelegatingWebMvcConfigurationTests.configurePathPatternParser()

izeye:
This PR fixes a variable reference in `DelegatingWebMvcConfigurationTests.configurePathPatternParser()` as it seems to be a wrong reference.


rstoyanchev:
Thanks @izeye!

PR #26450, opened 2021-01-27
Add Javadoc @since tag to TransactionSynchronizationUtils.unwrapResourceIfNecessary()

izeye:
It has been changed to be `public` in 4a7a2258f9ad8771e9a8225f5a18db156372162a, but its Javadoc `@since` tag seems to be missed. This PR simply adds it.


sbrannen:

Thanks

sbrannen:
FYI: this actually broke the build...

> [ant:checkstyle] [ERROR] /.../spring-framework/spring-tx/src/main/java/org/springframework/transaction/support/TransactionSynchronizationUtils.java:61: At-clauses have to appear in the order '[@param, @return, @throws, @since, @deprecated, @see]'. [AtclauseOrder]

I'll take the blame since I merged without verifying. Fixed in 5e561623944665f282f5549e3c21b5c1261d7359.

However, in the future, please make sure you execute `./gradlew check` before submitting a PR -- yes, even for _minor_ Javadoc changes. 😇

Issue #23460, opened 2019-08-14
ConcurrentModificationException in MockMvc.perform


mmyersDE:
Based on https://github.com/spring-projects/spring-security/issues/7224#issuecomment-520936156  by @rwinch:

> There appears to be an issue with MockMvc not allowing multiple Threads to operate on the headers at the same time. It happens in Spring Security because the `HeaderWriterFilter` will attempt to write the headers [when it completes](https://github.com/spring-projects/spring-security/blob/5.1.6.RELEASE/web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java#L77) and just before the [response is committed](https://github.com/spring-projects/spring-security/blob/5.1.6.RELEASE/web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java#L100).
>
> I [pushed a branch named nosecurity](https://github.com/rwinch/SPSS-7224/tree/nosecurity) that reproduces the issue without Spring Security being used to demonstrate the issue is in MockMvc rather than Spring Security.

rstoyanchev:
The stacktrace for context:
```
Caused by: java.util.ConcurrentModificationException
	at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:719)
	at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:742)
	at org.springframework.mock.web.HeaderValueHolder.getByName(HeaderValueHolder.java:98)
	at org.springframework.mock.web.MockHttpServletResponse.doAddHeaderValue(MockHttpServletResponse.java:617)
	at org.springframework.mock.web.MockHttpServletResponse.setHeaderValue(MockHttpServletResponse.java:579)
	at org.springframework.mock.web.MockHttpServletResponse.setHeader(MockHttpServletResponse.java:557)
	at test.Controller.lambda$nosecurity$0(Controller.java:23)
	at org.springframework.web.context.request.async.WebAsyncManager.lambda$startCallableProcessing$4(WebAsyncManager.java:327)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[...]

PR #26454, opened 2021-01-27
Synchronized access to MockHttpServletResponse headers

aplsup:
In order to support MVC async methods while testing with MockMvc, synchronizes any access to MockHttpServletResponse HTTP headers.
It avoids ConcurrentModificationException thrown when e.g. 2 different threads read
and write the header's map.

Link to a project reproducing the issue [here](https://github.com/aplsup/spring-mockmvc-aync-repro)
Related/similar to #23460

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletResponse.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletResponse.java
@@ -33,9 +33,12 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


rstoyanchev:
This is according to the Servlet spec 2.3.3.4:

> Other than the startAsync and complete methods, implementations of the request and response objects are not guaranteed to be thread safe. This means that they should either only be used within the scope of the request handling thread or the application must ensure that access to the request and response objects are thread safe.

So access to the response must be synchronized. The response itself is not.

#23460 should have been rejected as well. There was also an actual issue involving Spring Security and SSE and a similar exception but I can't find it at the moment. I believe it was transferred to be addressed within Spring Security.

PR #26458, opened 2021-01-27
Introduce SPI for automatic bean definition profile registration

raddatzk:
In our project we created a starter which contains most general stuff, like database, feign clients and so on. 
Now we want to configure this starter with different properties per stage. 
Currently we created a file "application-commons.yaml" in our starter and include this "profile" at `spring.profiles.active`, but this does not really scale well if we create more starters, or more applications. We would have to keep in mind to add all the profiles to all the services.

With this PR we are able to implement an `AdditionalProfileLoader`, which returns the name of the starters "profile", and then add the class to the `spring.factories`: `org.springframework.core.io.support.AdditionalProfileLoader=com.example.AdditionalCommonProfile`

This PR relates to https://github.com/spring-projects/spring-boot/pull/25033 which adds the ability to load additional files like `commons.yaml`, `commons-local.yaml` and so on.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/env/AbstractEnvironment.java
+++ b/spring-core/src/main/java/org/springframework/core/env/AbstractEnvironment.java
@@ -16,14 +16,8 @@
 
 package org.springframework.core.env;
 
-import java.security.AccessControlException;
-import java.util.*;
-import java.util.stream.Collectors;
-
-import jdk.internal.joptsimple.internal.Strings;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
 import org.springframework.core.SpringProperties;
 import org.springframework.core.convert.support.ConfigurableConversionService;
 import org.springframework.core.io.support.AdditionalProfileLoader;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@kevinraddatz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26458)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26458) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kevinraddatz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26458)!

philwebb:
I think this change is too invasive for Spring Framework which should not really be aware of Spring Boot concerns. We already have code in Spring Boot [to apply profiles to the Environment](https://github.com/spring-projects/spring-boot/blob/fd27c26e3b33b6f08dadfd524926e72796609937/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataEnvironment.java#L344-L349). I think it makes more sense for any additional profile logic to but put there.

PR #26480, opened 2021-01-31
URL and HTTP method in rest client exception message

jerzykrlk:
Dear team,

This is a followup of a discussion in a past PR: https://github.com/spring-projects/spring-framework/pull/1956

The rest client exception contains a preview of the body, which is great, and it helps a lot to quickly diagnose errors.

Now, the stacktrace looks like this:
```
Caused by: org.springframework.web.client.HttpServerErrorException$InternalServerError: 500 Server Error: [no body]
	at org.springframework.web.client.HttpServerErrorException.create(HttpServerErrorException.java:101)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:238)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:147)
```

I'd like to add the HTTP method and URL to the error, so that the stacktrace looks like this:
```
Caused by: org.springframework.web.client.HttpServerErrorException$InternalServerError: 500 Server Error after GET http://localhost:49709/status/server : [no body]
	at org.springframework.web.client.HttpServerErrorException.create(HttpServerErrorException.java:101)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:238)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:147)
```

Thank you!

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
@@ -20,11 +20,13 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.net.URI;
 import java.nio.CharBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
 import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.client.ClientHttpResponse;
@@ -109,20 +111,38 @@ public class DefaultResponseErrorHandler implements ResponseErrorHandler {
 	 * {@link HttpStatus} enum range.
 	 * </ul>
 	 * @throws UnknownHttpStatusCodeException in case of an unresolvable status code
-	 * @see #handleError(ClientHttpResponse, HttpStatus)
+	 * @see #handleError(URI, HttpMethod, ClientHttpResponse, HttpStatus)
 	 */
 	@Override
 	public void handleError(ClientHttpResponse response) throws IOException {
 		HttpStatus statusCode = HttpStatus.resolve(response.getRawStatusCode());
 		if (statusCode == null) {
 			byte[] body = getResponseBody(response);
-			String message = getErrorMessage(response.getRawStatusCode(),
-					response.getStatusText(), body, getCharset(response));
+			String message = getErrorMessage(response.getRawStatusCode(), response.getStatusText(), body, getCharset(response), null, null);
 			throw new UnknownHttpStatusCodeException(message,
 					response.getRawStatusCode(), response.getStatusText(),
 					response.getHeaders(), body, getCharset(response));
 		}
 		handleError(response, statusCode);
+	}
+
+	/**
+	 * Delegates to {@link #handleError(URI, HttpMethod, ClientHttpResponse, HttpStatus)} with the
+	 * response status code.
+	 * @throws UnknownHttpStatusCodeException in case of an unresolvable status code
+	 * @see #handleError(URI, HttpMethod, ClientHttpResponse, HttpStatus)
+	 */
+	@Override
+	public void handleError(URI url, HttpMethod method, ClientHttpResponse response) throws IOException {
+		HttpStatus statusCode = HttpStatus.resolve(response.getRawStatusCode());
+		if (statusCode == null) {
+			byte[] body = getResponseBody(response);
+			String message = getErrorMessage(response.getRawStatusCode(), response.getStatusText(), body, getCharset(response), url, method);
+			throw new UnknownHttpStatusCodeException(message,
+					response.getRawStatusCode(), response.getStatusText(),
+					response.getHeaders(), body, getCharset(response));
+		}
+		handleError(url, method, response, statusCode);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
```diff
@@ -132,9 +154,10 @@ public void handleError(ClientHttpResponse response) throws IOException {
 	 * </pre>
 	 */
 	private String getErrorMessage(
```
jerzykrlk, line 156: Ideally, I'd make this method protected - so that I can override it to extract some more specific information from responses that return HTML, XML or some other nonstandard data.

qw81641493 (approved):

spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
```diff
@@ -175,11 +207,29 @@ private String getErrorMessage(
 	 * @see HttpServerErrorException#create
 	 */
 	protected void handleError(ClientHttpResponse response, HttpStatus statusCode) throws IOException {
+		handleError(null, null, response, statusCode);
+	}
```
rstoyanchev, line 211: This is a protected method that is now bypassed. Something to consider for a minor/major release but not at this stage of the 5.3.x branch.

rstoyanchev (commented):
Given the change of protected methods I'd rather avoid this change at this stage of the 5.3.x branch. As far as I can see you can still make it happen in your own custom extension of DefaultResponseErrorHandler, right? It just wouldn't be out of the box.

spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
jerzykrlk: Thanks for looking into that!

Initially, I didn't think of squeezing it into 5.3. I'd be very happy to do so - but I think the previous changes broke something in spring-integration, so it might not be that easy.

I tried to not to change the behaviour of this method at all. I can also try to rewrite this to make it more 5.3-compatible. But If you think it makes more sense to wait - let's wait.

spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
jerzykrlk: Considering something beyond 5.3.x - what about this one? 

I'd love to have an extension point where I can further customize the response body preview - format it right, remove some tags, etc. 

A long time ago, here: https://github.com/spring-projects/spring-framework/pull/1956/files - I even thought about a separate class to extract the body. But a protected method would be just fine.

jerzykrlk:
> As far as I can see you can still make it happen in your own custom extension of DefaultResponseErrorHandler, right? It just wouldn't be out of the box.

I'm not sure if you're refering to making `getErrorMessage` protected, or the entire PR?

spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
```diff
@@ -109,20 +111,40 @@ protected boolean hasError(int unknownStatusCode) {
...
 			byte[] body = getResponseBody(response);
-			String message = getErrorMessage(response.getRawStatusCode(),
-					response.getStatusText(), body, getCharset(response));
+			String message = getErrorMessage(response.getRawStatusCode(), response.getStatusText(), body, getCharset(response), url, method);
 			throw new UnknownHttpStatusCodeException(message,
 					response.getRawStatusCode(), response.getStatusText(),
 					response.getHeaders(), body, getCharset(response));
 		}
-		handleError(response, statusCode);
+		handleError(url, method, response, statusCode);
```
rstoyanchev, line 147: This is the protected method that's now bypassed.

rstoyanchev:
>> As far as I can see you can still make it happen in your own custom extension of DefaultResponseErrorHandler, right? It just >> wouldn't be out of the box.

> I'm not sure if you're refering to making getErrorMessage protected, or the entire PR?

I'm talking about the call to the protected [handleError method](https://github.com/spring-projects/spring-framework/blob/53cafe728cf0fe1f333dc77e93f2674e54e597a2/spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java#L125) that is now bypassed.

spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
jerzykrlk: Thanks for the explanation!

Then it can definitely wait for a future Spring release. I'm not in a particular hurry - any custom extensions would well enough.

Should I close the PR, or just keep it here? 

Also - is it something that you'd see easy to port into WebClient? Would it make sense to make these two in sync?

rstoyanchev: I think we can close it for now. On the WebClient side it looks we do have the URL and HTTP method in `WebClientResponseException`.

jerzykrlk: Thanks for the feedback @rstoyanchev ! 

Is there a chance to squeeze this in a future Spring Framework release? Should I come back once it's announced?

rstoyanchev: Yes, of course. We can probably re-open this PR then.

PR #26487, opened 2021-02-01
Add application/vnd.google.protobuf to ProtobufCodecSupport

making:
While RSocket spec supports `application/vnd.google.protobuf` as a Well-known MIME Type, Protobuf Codec doesn't.
As a result, Spring RSocket app that has `ProtobufDecoder` configured can not receive a payload with `application/vnd.google.protobuf` data mime type.
https://github.com/rsocket/rsocket/blob/master/Extensions/WellKnownMimeTypes.md


bclozel:
@making `application/vnd.google.protobuf` sounds good, but is `"application/octet-stream"` really necessary? This one might be problematic.

making:
I think it's safe to keep it for the backwards compatibility.

sbrannen:
> @making `application/vnd.google.protobuf` sounds good, but is `"application/octet-stream"` really necessary? 

`"application/octet-stream"` was already there.

bclozel:
Ha, never mind I misread the diff. Thanks!

rstoyanchev:
I think this safe to do. It would expand the capability on the decoding side, but on the encoding side we still default to the first one on the list.

PR #26504, opened 2021-02-03
Fall back on awaitToBodylessEntity when awaitBody is used with Unit

gaerfield:
allow WebClient.ResponseSpec.awaitBody()
to be used in combination with Type "Unit"


sdeleuze:
Polished and merged, thanks.

PR #26556, opened 2021-02-16
Fix build output directory for ref docs in CONTRIBUTING.md

cprayer:
https://github.com/spring-projects/spring-framework/blob/fd49c8f598a9ce33f7ce14ace31c86856819776b/gradle/docs.gradle#L150-L168
(`fd49c8f` is latest commit of master branch)
As you can see, I think it should be `build/docs/ref-docs/html5` directory, not a `build/asciidoc/html5` directory.


sbrannen:
Good catch.

Thanks

PR #26569, opened 2021-02-18
DataClassRowMapper should not override Kotlin init properties

dennie170:
Since Kotlin data classes are already constructed with values we don't want to override all the properties later on again.
Properties modified in the init { } block in a class would be reverted back to the value of the column again.

An example of this would be setting the default scale of a BigDecimal inside an init block. The init would be called and set the scale after constructing. However, after constructing the BeanPropertyRowMapper would override the BigDecimal again so the scale would be 0 again.

This commit fixes this issue for Kotlin data classes.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/DataClassRowMapper.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/DataClassRowMapper.java
@@ -21,6 +21,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 
 import org.springframework.beans.BeanUtils;
+import org.springframework.beans.BeanWrapperImpl;
 import org.springframework.beans.TypeConverter;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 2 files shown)


dennie170:
Hi guys, this issue has not been commented on since February. 
Is this going to be picked up sometime?

dennie170:
Hi guys, I see nobody has replies to my issue since February. 
Are you gonna have a look at this? It is pretty "breaking" in Kotlin, and now demands some custom solutions.
Thanks!

jhoeller:
It turns out that Kotlin `var` declarations implicitly generate setter methods for the affected properties, which leads to our `DataClassRowMapper` populating that setter as well (since we're expecting either a constructor argument or a setter method, not both, we haven't been checking for this explicitly). Mixing and matching between constructor arguments and setter methods is allowed by design, so we can't cut off the latter completely (like the original PR does). Instead, I've revised the `DataClassRowMapper` algorithm to suppress setter method calls for all properties that have been populated through constructor arguments already. In addition to a Java-based test for this behavior, I've also added your Kotlin test case.

dennie170:
Thanks for figuring it out and the clear explanation! Glad to see it's fixed now.

PR #26575, opened 2021-02-20
Fix handling of "file:" paths to non-existent files on Windows

candrews:
On Windows, for `setAsText`, if the text argument is a `file:` URL for a path that does not exist, `Paths.get(text)` is called where text is a `file:` URL, which doesn't work - the result is an `InvalidPathException`. The Windows part is important - this issue only manifests when running on Windows.

To fix this issue, also check that the resource isn't a file before calling `Paths.get()`. That way, resources that are files skip to the other branch.


candrews:
I've added some tests to `PathEditorTests` to test for this issue. However, this issue only manifests when running on Windows (the newly added tests pass with and without this fix on other platforms).

PR #26581, opened 2021-02-21
fix initialization of spring-batch tables with datasource in auto-commit mode false

### LINKED ISSUE #27008: DataSourceInitializer does not work with spring.datasource.hikari.auto-commit=false
Hello, 

I would like to submit an issue. I am facing a problem with creation of spring-batch tables when you use a datasource in `auto-commit mode => false;`

My project is using spring-boot@2.2.7 and postgresql.

Here is my configuration :

```
spring:
  batch:
    initialize-schema: always

spring:
  datasource:
    type: com.zaxxer.hikari.HikariDataSource
    hikari:
      poolName: Hikari
      auto-commit: false
```	  

I have checked the `BatchDataSourceInitializer` class which is the subclass of `AbstractDataSourceInitializer`.
It appears that the code at line 65 `DatabasePopulatorUtils.execute(populator, this.dataSource);` is responsible for running the sql script of spring-batch tables, with spring-jdbc, but the transaction is never committed.
And consequently, my tables are never created in my database.

If I switch the configuration of my datasource, ie `auto-commit: true`, everything is working fine.

I was expecting my spring-batch tables be created (initialize-schema => always) even if my datasource is not in auto-commit mode.
It's not so easy and not very intuitive, at first sight, to discover why the tables are not created. We first think about a misconfiguration relative to spring-batch but nothing related to datasource configuration.

What do you think about that ?

Thanks for your help.

nguyensach:
When initialize spring-batch tables with postgresql datasource in auto-commit mode false,
the tables are not created in the database.

In ScriptUtils.executeSqlScript, after executing statement, there is not an explicit commit.
So that, when postgresql datasource is in auto-commit mode false, transaction is not committed

I added an explicit commit after executing statement when auto-commit mode is false

- Fixes #27008

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/init/ScriptUtils.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/init/ScriptUtils.java
@@ -599,6 +599,9 @@ public abstract class ScriptUtils {
 					stmtNumber++;
 					try {
 						stmt.execute(statement);
+						if(!connection.getAutoCommit()) {
+							connection.commit();
+						}
 						int rowsAffected = stmt.getUpdateCount();
 						if (logger.isDebugEnabled()) {
 							logger.debug(rowsAffected + " returned as update count for SQL: " + statement);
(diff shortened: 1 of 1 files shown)


sbrannen:
Thanks for the PR.

`ScriptUtils` is a set of low-level utilities for parsing and executing SQL scripts, with zero knowledge of transaction management. Transaction management should occur at a higher level in the application.

PR #26592, opened 2021-02-23
Fix MockHttpServletRequest reference in Javadoc

izeye:
This PR fixes `MockHttpServletRequest` reference in Javadoc.


sbrannen:

We had already noticed that copy-n-paste error but hadn't gotten around to fixing it yet.

So thanks for the nudge.

PR #26600, opened 2021-02-24
Correctly set auto-growing array's element

nullzl:
The implementation of `AbstractNestablePropertyAccessor.processKeyedProperty` results in a `java.lang.IllegalArgumentException: array element type mismatch` when the property expression has more than one property key and the last key causes the array to grow automatically. 

For example, given a property `int[][] multiArray = new int[2][2]` and property expression  `multiArray[1][3]`, the `processKeyedProperty` method creates a new array object and assigns it to `multiArray`; whereas, the new array object should be assigned to `multiArray[1]`.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
@@ -305,8 +305,10 @@ public abstract class AbstractNestablePropertyAccessor extends AbstractPropertyA
 					Class<?> componentType = propValue.getClass().getComponentType();
 					Object newArray = Array.newInstance(componentType, arrayIndex + 1);
 					System.arraycopy(propValue, 0, newArray, 0, length);
-					setPropertyValue(tokens.actualName, newArray);
-					propValue = getPropertyValue(tokens.actualName);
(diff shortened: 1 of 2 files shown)


sbrannen (changes requested):
Please provide a unit test that fails before the change and passes after the change -- for example, something analogous to `org.springframework.beans.BeanWrapperAutoGrowingTests.getPropertyValueAutoGrowMultiDimensionalArray()`.

Thanks

nullzl:
> Please provide a unit test that fails before the change and passes after the change -- for example, something analogous to `org.springframework.beans.BeanWrapperAutoGrowingTests.getPropertyValueAutoGrowMultiDimensionalArray()`.
> 
> Thanks

Thank you very much for your review. I have added  a unit test org.springframework.beans.BeanWrapperAutoGrowingTests.setPropertyValueAutoGrowMultiDimensionalArray().

sbrannen (approved):

sbrannen:
This has been merged into `master` in a33eac3ec05952424603d1fefc627b4a3ef36449, revised in 8baf404893037951ac29393a41d40af4fa11775b, and backported to `5.2.x`.

Thanks for catching _and_ fixing this bug!

PR #26612, opened 2021-02-26
Throw Exception when two primary beans are present in one factory

nullzl:
Prior to this commit, the implementation of `DefaultListableBeanFactory.determinePrimaryCandidate()` did not throws `NoUniqueBeanDefinitionException` in the flowing two situations:
* more than one primary bean in one ancestor factory.(test unit in `DefaultListableBeanFactoryTests.autowireBeanByTypeWithTwoPrimaryCandidatesInOneAncestor()`)
* more than one local primary bean which implements `FactoryBean`.(test unit in `DefaultListableBeanFactoryTests.autowireBeanByTypeWithTwoPrimaryFactoryBeans()`)

This commit fix these issues by prefer primary bean in the youngest factory, and if the youngest factory has more than one primary bean thorws `NoUniqueBeanDefinitionException`.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jhoeller:
We are likely to implement this is in a different fashion but the general point is valid: We ignore multiple matching primary beans in ancestors, just taking the first one, whereas we do insist on a unique primary bean to be available locally. We might simply add another `else` clause to the present checks when we got a primaryBeanName already, rejecting another matching primary bean when it is not a local override.

PR #26619, opened 2021-02-27
Local @CrossOrigin maxAge value should override global value

GungnirLaevatain:
The attribute of **CrossOrigin** that be applied to method will override that be applied to class, like **origins**, but **maxAge** is different，so I think we would modify it


sbrannen:
The Javadoc for `@CrossOrigin` states:

> The rules for combining global and local configuration are generally additive -- e.g. all global and all local origins. For those attributes where only a single value can be accepted such as `allowCredentials` and `maxAge`, the local overrides the global value.

In light of that, I have labeled this as a bug to be fixed in 5.3.5 and backported to 5.2.14.

Good catch!

sbrannen:
This has been merged into `master` in 90de1ab6d1cb85fe9d84a0985f9e1762494ede7e, revised in 01c2e12fef43f422b897bbf57aeac95a95c90ce8, and backported to `5.2.x`.

Thanks

----

For future reference, please make sure to run `./gradlew check` before submitting a PR in order to catch and correct any Checkstyle violations such as the unused imports in this PR.

Issue #26379, opened 2021-01-12
Support for RSocket per-stream data MIME type extension


rstoyanchev:
At present there can be only one data MIME for the every request/stream on the connection. This is negotiated in the SETUP frame. The [Stream Data MIME Types Metadata Extension](https://github.com/rsocket/rsocket/blob/master/Extensions/PerStreamDataMimeTypesDefinition.md) defines a way to declare the data MIME type for a specific request/stream. This includes a MIME type for the request input and an "acceptable" MIME type for the response.

rudy2steiner:
I'll work on it

spencergibb:
This would be useful for the RSocket Routing Broker

PR #26621, opened 2021-02-28
Message and stream level data  and accept  mime

rudy2steiner:
At present there can be only one data MIME for the every request/stream on the connection, 
Supporting content-type and accept header like http in this pr.
Implementation of  issue #26379

Class MimeTypeMetadataCodec will be delete once  rsocket-java v1.1.1 release


rudy2steiner:
@rstoyanchev PTAL

rstoyanchev (changes requested):
Some high level feedback. 

To start, the `MetadataMimeCodec` should be in RSocket-Java, not here. I've created https://github.com/rsocket/rsocket-java/issues/996 for that.

On the side of RSocketRequester, I can see potentially methods to set the mimeTypes or acceptMimeTypes for a request but no need more generally to introduce the concept of "headers".

Some classes, like `AbstractEncoderMethodReturnValueHandler.java`, are intended to be more generally used and this is why they are not in the `rsocket` package. Such classes cannot have RSocket dependencies.

rudy2steiner:
> Some classes, like `AbstractEncoderMethodReturnValueHandler.java`, are intended to be more generally used and this is why they are not in the `rsocket` package. Such classes cannot have RSocket dependencies.

I agree with MetadataMimeCodec should be in Rsocket-java, and do you have more specific feedbacks how to implement contentType and acceptMimes .

rudy2steiner:
@rstoyanchev thanks for your feedbacks, updated!

spring-messaging/src/main/java/org/springframework/messaging/rsocket/annotation/support/MessagingRSocket.java
```diff
@@ -203,16 +204,32 @@ private MessageHeaders createHeaders(
...
+	 **/
+	@SuppressWarnings({ "unchecked" })
+	public MimeType getMessageContentType(MessageHeaderAccessor headers){
+		Object messageContentType  = headers.getHeader(MessageHeaders.CONTENT_TYPE);
+		if (messageContentType instanceof List) {
+			List<Object> types = (List<Object>)messageContentType;
+			return types.size() == 0 ? this.dataMimeType : (MimeType) types.get(0);
+		}
+		return this.dataMimeType;
+	}
```
rstoyanchev, line 233: Likewise here, when accessed `CONTENT_TYPE` should be expected to be a single `MimeType`.

spring-messaging/src/main/java/org/springframework/messaging/rsocket/annotation/support/RSocketPayloadReturnValueHandler.java
```diff
@@ -50,22 +56,56 @@ public class RSocketPayloadReturnValueHandler extends AbstractEncoderMethodRetur
...
 
 	@Override
 	protected Mono<Void> handleEncodedContent(
 			Flux<DataBuffer> encodedContent, MethodParameter returnType, Message<?> message) {
 
 		AtomicReference<Flux<Payload>> responseRef = getResponseReference(message);
+		AtomicReference<MimeType> contentTypeAtomic = getResponseContentType(message);
 		Assert.notNull(responseRef, "Missing '" + RESPONSE_HEADER + "'");
-		responseRef.set(encodedContent.map(PayloadUtils::createPayload));
+		responseRef.set(encodedContent.switchOnFirst((signal, inner) -> {
```
rstoyanchev, line 78: We shouldn't automatically use `switchOnFirst`. If no content negotiation took place, and the application is using the connection-wide `dataMimeType`, there is no need to use this operator.

spring-messaging/src/main/java/org/springframework/messaging/rsocket/DefaultRSocketRequester.java
```diff
@@ -310,12 +349,27 @@ private <T> Flux<T> retrieveFlux(ResolvableType elementType) {
...
 					(T) decoder.decode(dataBuffer, elementType, dataMimeType, EMPTY_HINTS));
 		}
 
+		@SuppressWarnings("unchecked")
+		public <T> Flux<T> decodeFlux(Payload firstPayload, Flux<Payload> payloads,ResolvableType elementType){
+			MimeType contentMime =  messageContentType(firstPayload);
+			Decoder<?> decoder = strategies.decoder(elementType, contentMime);
+			return payloads.map(this::retainDataAndReleasePayload).map(dataBuffer ->
+					(T) decoder.decode(dataBuffer, elementType, contentMime, EMPTY_HINTS));
+		}
```
rstoyanchev, line 371: Likewise on the requester side, we should be aware if a request specific content type is being used, and if not then avoid using the `switchOnFirst` operator.

spring-messaging/src/main/java/org/springframework/messaging/handler/invocation/reactive/HandlerMethodReturnValueHandler.java
```diff
@@ -33,7 +33,10 @@ public interface HandlerMethodReturnValueHandler {
 
 	/** Header containing a DataBufferFactory for use in return value handling. */
 	String DATA_BUFFER_FACTORY_HEADER = "dataBufferFactory";
-
+	/**
+	 * Message header name of response content type.
+	 **/
+	String RESPONSE_CONTENT_TYPE = "responseContentType";
```
rstoyanchev, line 39: It's not clear what the purpose of this header is. It seems to be set as an empty AtomicReference from `MessagingRSocket` in the beginning, and then its value is set from `AbstractEncoderMethodReturnValueHandler`, but I don't see anything making use of it.

rstoyanchev (changes requested):
@rudy2steiner, I've reviewed the PR and added a few comments.

rstoyanchev:
Closing for now as the PR needs significant updates, but we have #26379 to keep track of the request.

PR #26634, opened 2021-03-03
Optimize initialization size of temporary ArrayLists in ReflectionUtils

chenqimiao:
I feel that there is no need to specify the initial size for this ArrayList, just use the default initial size is enough.


jhoeller:
Good catch, that number 32 seems a bit arbitrary (and as a power of two, rather suited for a Set or Map). However, since even `java.lang.Object` has >10 declared methods already, a higher default size does seem useful there in `ReflectionUtils`. I'm inclined to change it 20, expecting one always-triggered duplication of the default ArrayList capacity of 10.

chenqimiao:
Thanks for your reply, I think what you said makes sense, so I revised my PR
@jhoeller

jhoeller:
@chenqimiao It seems the PR contains 3 commits now, could you please squash them into a single commit?

chenqimiao:
> @chenqimiao It seems the PR contains 3 commits now, could you please squash them into a single commit?

Sure .

PR #26651, opened 2021-03-08
Update reference to deprecated CronSequenceGenerator

MichelTenVoorde:
Scheduled#cron() was pointing to CronSequenceGenerator for an explanation of the cron string format. This class however is deprecated, and has been replaced with CronExpression. This PR updates the JavaDoc accordingly.


snicoll:
@MichelTenVoorde good catch and thank for you making your first contribution to Spring Framework.

PR #26680, opened 2021-03-15
Add support for Oracle bind marker scheme using R2DBC

mp911de:
We now support Oracle's bind marker scheme that identifies dynamic parameters using names that are prefixed with a colon such as `:P0_myparam`.


sbrannen:

Thanks

PR #26684, opened 2021-03-16
Polish TestContextAnnotationsUtilsTests

Rebwon:
This PR improves code by removing unnecessary conditional statements.


sbrannen:
Thanks for the PR.

The use of if-blocks in `TestContextAnnotationsUtils` is intentional in terms of consistency between the various steps of the search algorithm. Thus, although your proposed changes are technically correct, I'll keep the code as it currently is.

I have therefore changed the focus of this PR to the polishing in `TestContextAnnotationsUtilsTests`.

Rebwon:
Thank you for your answer. I agree with your opinion.

PR #26700, opened 2021-03-18
Apply dynamic changes in ThreadPoolTaskExecutor before setting local value

filiphr:
If the ThreadPoolTaskExecutor is dynamically changed with an invalid value
the state of the ThreadPoolTaskExecutor does no longer correctly represent
the state of the underlying ThreadPoolExecutor


filiphr:
Hey team, I wanted to check if you have had the time to look into the PR. Is this something interesting for you? Is there something more that I should do about this?

jhoeller:
Thanks for raising attention here, this unfortunately remained unnoticed initially.
Looks good to from a quick glance, I'll roll it into 5.3.9.

PR #26706, opened 2021-03-19
Treat RowSet column labels as case-insensitive

navol:
CachedRowSetImpl doesn't allow for column label use (SPR-7506).

Looking up values by labels in RowSet is supported by ResultSetWrappingSqlRowSet, with fallback to wrapped ResultSet if nothing found. Unfortunately it works in case-sensitive manner (HashMap<String,Integer>).

Assume there is a table:
CREATE TABLE T(col int)

And code:
SqlRowSet srs = jdbcTemplate.queryForRowSet(query);
if ( srs.next() ) srs.getInt("col"); // do smth with value

Queries:
1. select col from T -- ok, column name here, and column name == column label
2. select 0 as col from T -- ok, column label, cache in ResultSetWrappingSqlRowSet handle this
3. select 0 as Col from T -- error, column name != column label, + cache miss 
3.1. column name != column label
3.2. code in ResultSetWrappingSqlRowSet can't find column by label, because "col" != "Col"
3.3. wrapped ResultSet (RowSet, CachedRowSetImpl) can't find column by name, because there is no column name

This commit change column-label treatment to case-insensitive.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/rowset/ResultSetWrappingSqlRowSet.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/rowset/ResultSetWrappingSqlRowSet.java
@@ -29,7 +29,7 @@ import java.util.Map;
 
 import org.springframework.jdbc.InvalidResultSetAccessException;
 import org.springframework.lang.Nullable;
-import org.springframework.util.CollectionUtils;
+import org.springframework.util.LinkedCaseInsensitiveMap;
 
 /**
  * The default implementation of Spring's {@link SqlRowSet} interface, wrapping a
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@navol Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26706)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26706) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@navol Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26706)!

snicoll:
@navol thank you for the PR and sorry it took so long to process it. 

The behavior of dealing with case insensitive values should be explicit. `JdbcTemplate` has a  `setResultsMapCaseInsensitive` to turn this feature on/off and looking at `SqlRowSet` I think we should keep things as they are.

Issue #26707, opened 2021-03-20
Resource.isFile() return true when the resource path actually not exists


drgnchan:
**Affects:** springframework-5.3.5; spring web 5.3.5; spring boot 2.4.4

**demo**:
```java
@SpringBootApplication
public class ResourcebugApplication {

    public static void main(String[] args) {
        ConfigurableApplicationContext context = SpringApplication.run(ResourcebugApplication.class, args);
        Resource notExistsResource = context.getResource("com/example/resourcebug/NotExists.java");
        System.out.println("notExistsResource is file: " + notExistsResource.isFile());
    }
}
```

**output**:
notExistsResource is file: true

**expected output**:
notExistsResource is file: false

drgnchan:
is this issue associated with #26702

PR #26709, opened 2021-03-21
check file with realPath is existing

drgnchan:
#26707 .Tomcat's `StandardContext.getRealPath(String path)` will return non null value when the file actually not exists.So add some code in `ServletContextResource` to check file is actually exists with jdk's `File` class.


cuspymd (commented):
There is already an `exists()` method, do you think it is correct to check the existence of the file in the `isFile()` method?

drgnchan:
> There is already an `exists()` method, do you think it is correct to check the existence of the file in the `isFile()` method?

maybe it is a directory.

bclozel:
Closing this PR in favor of #26707 - your contribution has been merged in b71e686cbd, but this outdated PR got wrongly updated with lots of duplicate commits (a rebase and a force-push without any merge would have made things better).

Sorry it took us so long to process this PR, this fix is going to be shipped with the next 5.3.x maintenance release.

PR #26725, opened 2021-03-25
Introduce `reverse` methods in AOP `MethodMatchers` and `ClassFilters`

Jinhui-Z:
Add reverse methed in MethodMatchers and ClassFilters.

Add support for MethodMatchers and ClassFilters to reverse the given MethodMatcher and ClassFilter.
It is convenient in some complex conditions,Or is there another way to do it?Thanks.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/support/ClassFilters.java
+++ b/spring-aop/src/main/java/org/springframework/aop/support/ClassFilters.java
@@ -85,6 +85,17 @@ public abstract class ClassFilters {
 		return new IntersectionClassFilter(classFilters);
 	}
 
+	/**
+	 * reverse the given ClassFilter match.
+	 * @param cf the ClassFilter
+	 * @return a distinct ClassFilter that not matches classes
+	 * of the given ClassFilter match
+	 */
+	public static ClassFilter reversion(ClassFilter cf) {
+		Assert.notNull(cf, "ClassFilter must not be null");
+		return new ReversionClassFilter(cf);
+	}
+
 
 	/**
 	 * ClassFilter implementation for a union of the given ClassFilters.
@@ -167,4 +178,39 @@ public abstract class ClassFilters {
 
 	}
 
+
+	/**
+	 * ClassFilter implementation for an reversion of the given ClassFilter.
+	 */
+	@SuppressWarnings("serial")
+	private static class ReversionClassFilter implements ClassFilter, Serializable {
+
+		private final ClassFilter filter;
+
+		ReversionClassFilter(ClassFilter filter) {
+			this.filter = filter;
+		}
+
+		@Override
+		public boolean matches(Class<?> clazz) {
+			return !filter.matches(clazz);
+		}
+
+		@Override
+		public boolean equals(@Nullable Object other) {
+			return (this == other || (other instanceof ReversionClassFilter &&
+					this.filter.equals(((ReversionClassFilter)other).filter)));
+		}
+
+		@Override
+		public int hashCode() {
+			return 37 * this.filter.hashCode();
+		}
+
+		@Override
+		public String toString() {
+			return getClass().getName() + ": " + this.filter;
+		}
+
+	}
 }
--- a/spring-aop/src/main/java/org/springframework/aop/support/ClassFilters.java
+++ b/spring-aop/src/main/java/org/springframework/aop/support/ClassFilters.java
@@ -213,4 +213,5 @@ public abstract class ClassFilters {
 		}
 
 	}
+	
 }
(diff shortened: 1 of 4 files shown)


cuspymd (commented):
Isn't it simpler to do "Logical NOT" on the results of the matches() of the original filter? Could you give me an example of how these classes can be useful?

Jinhui-Z:
@cuspymd  Thanks for the reply.
Some complex combinations are handy, for example in the SpringCloud, if I need to develop a spring-boot-starter project that collects logs for all requests by default and provides @logIgnore to ignore certain requests. For Feign, I could write the Advisor like this:

```
public class Example {

    /**
     * this FeignClient require logging
     */
    @FeignClient("svr-name")
    public interface Client extends Api {
    }


    /**
     * this FeignClient does not require logging
     */
    @LogIgnore
    @FeignClient("svr-name")
    public interface IgnoreClient extends Api {
    }


    /**
     * in other Jar
     */
    interface Api {
        /**
         * this method does not require logging
         */
        @LogIgnore
        @RequestMapping("/ignoreApi")
        void ignoreApi();

        /**
         * this method require logging
         */
        @RequestMapping("/api")
        void api();
    }


    /**
     * Annotated classes or methods do not need to be logged
     */
    @Target({ElementType.METHOD, ElementType.TYPE})
    @Inherited
    @Documented
    @interface LogIgnore {
    }


    /**
     * The Advisor for Feign
     */
    public class FeignAdvisor extends AbstractPointcutAdvisor {

        @Override
        public Pointcut getPointcut() {
            return new Pointcut() {
                @Override
                public ClassFilter getClassFilter() {
                    //This class mast contain @FeignClient
                    ClassFilter cf1 = new AnnotationClassFilter(FeignClient.class, true);
                    //This class cannot contain @LogIgnore
                    ClassFilter cf2 = ClassFilters.reversion(new AnnotationClassFilter(com.mifi.lc.vc.store.admin.basic.FeignAdvisor.LogIgnore.class, true));

                    //maybe more ClassFilter

                    return ClassFilters.intersection(cf1, cf2);
                }

                @Override
                public MethodMatcher getMethodMatcher() {
                    //This method mast contain @RequestMapping
                    MethodMatcher cf1 = new AnnotationMethodMatcher(RequestMapping.class, true);
                    //This method cannot contain @LogIgnore
                    MethodMatcher cf2 = MethodMatchers.reversion(new AnnotationMethodMatcher(com.mifi.lc.vc.store.admin.basic.FeignAdvisor.LogIgnore.class, true));

                    //maybe more MethodMatcher

                    return MethodMatchers.intersection(cf1, cf2);
                }
            };
        }

        @Override
        public Advice getAdvice() {
            //do something
            return (MethodInterceptor) (MethodInvocation invocation) -> invocation.proceed();
        }
    }
}
```

cuspymd:
> Some complex combinations are handy, for example in the SpringCloud, if I need to develop a spring-boot-starter project that collects logs for all requests by default and provides @logIgnore to ignore certain requests. For Feign, I could write the Advisor like this:
Thanks for the detailed explanation. It looks very useful.

Jinhui-Z:
@rstoyanchev        Could you please have a look into this?

Pulkit-Garg15:
Hi, I am a beginner with Open source contributions. This would be my first contribution, can anyone please guide me through it?

snicoll:
@Pulkit-Garg15 thanks for the offer but this is already a PR (see the "Files Changed" tab above) so this isn't open for contribution.

snicoll:
I think a `negate` on `ClassFilter` can be added as it is a functional interface. I am wondering if we should do the same for `MethodFilter` though, and how important the equals/hashCode bit is.

sbrannen:
> I think a `negate` on `ClassFilter` can be added as it is a functional interface.

I agree. It's analogous to `java.util.function.Predicate.negate()` in that sense.

Along the same lines, I could see it being useful to have something like the static `java.util.function.Predicate.not(Predicate<? super T>)` companion method.

> I am wondering if we should do the same for `MethodFilter` though,

I imagine that could be handy in `MethodMatcher`, too.

> and how important the equals/hashCode bit is.

The filters are used to differentiate pointcuts -- for example, in `org.springframework.aop.support.ComposablePointcut.equals(Object)`. So, unless I'm mistaken, proper `equals` and `hashCode` implementations are required.

snicoll:
Such a shame that we can't do the `negate` idea as a default method. We have several components in the framework that implements both interface so we're kinda stuck there :/

snicoll:
@Jinhui-Z thanks very much for making your first contribution to Spring Framework.

izeye:
This seems to have been shipped with 6.1.0-M5 but doesn't have a milestone. There's also a typo in "metheds" in the title.

snicoll:
Good catch @izeye. I've fixed the title and added the missing entry manually to the release notes.

Issue #26491, opened 2021-02-02
WebClients awaitBody throws Exception when expecting "Unit"


blatobi:
**Affected component**: spring-webflux
**Version**:  5.3.2

When using the Kotlin-Extensions for WebClient on Endpoints that do not return any data, `awaitBody<Unit>` is throwing `java.util.NoSuchElementException: No value received via onNext for awaitSingle`.

I was thinking I may accidentally misuse the API, but [spring-docs](https://docs.spring.io/spring-framework/docs/current/reference/html/web-reactive.html#webflux-client-body) document also this call:
```
val personDeferred: Deferred<Person> = ...

client.post()
        .uri("/persons/{id}", id)
        .contentType(MediaType.APPLICATION_JSON)
        .body<Person>(personDeferred)
        .retrieve()
        .awaitBody<Unit>()
```

**Reproduction**:
Given the following controller:
```
import org.springframework.http.HttpStatus
import org.springframework.web.bind.annotation.*

@RestController
class TestController {

    @PostMapping("/test")
    @ResponseStatus(HttpStatus.NO_CONTENT)
    fun postTest() {}

}
```
And this Unit-Test:
```
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
internal class TestControllerTest(
    @LocalServerPort private val port : Int
) {

    @Test
    fun postTest()  = runBlocking {
        WebClient.create("localhost:$port/test")
            .post()
            .retrieve()
            .awaitBody<Unit>()
    }

}
```
The test will fail with
```
java.util.NoSuchElementException: No value received via onNext for awaitSingle

	at |b|b|b(Coroutine boundary.|b(|b)
	at TestControllerTest$deleteTest$1.invokeSuspend(TestControllerTest.kt:52)
Caused by: java.util.NoSuchElementException: No value received via onNext for awaitSingle
	at kotlinx.coroutines.reactive.AwaitKt$awaitOne$$inlined$suspendCancellableCoroutine$lambda$1.onComplete(Await.kt:181)
	at reactor.core.publisher.StrictSubscriber.onComplete(StrictSubscriber.java:123)
```

blatobi:
Current workaround:
```
WebClient.create("localhost:$port/test")
            .post()
            .retrieve()
            .toBodilessEntity()
            .awaitSingleOrNull()
```

sdeleuze:
[...]

PR #26731, opened 2021-03-25
awaitBodyOrNull function to handle empty body

ValentinShakhov:
Created awaitBodyOrNull function to handle empty body

`suspend inline fun <reified T : Any> WebClient.ResponseSpec.awaitBody() : T `
does not handle nullable body

`suspend fun WebClient.ResponseSpec.awaitBodilessEntity()`
ignores body completely even when the content-length is above zero

### DIFF
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
@@ -144,6 +144,17 @@ suspend inline fun <reified T : Any> WebClient.ResponseSpec.awaitBody() : T =
 	}
 
 /**
+ * Coroutines variant of [WebClient.ResponseSpec.bodyToMono].
+ *
+ * @author Valentin Shakhov
+ */
+suspend inline fun <reified T : Any> WebClient.ResponseSpec.awaitBodyOrNull() : T? =
+	when (T::class) {
+		Unit::class -> awaitBodilessEntity().let { Unit as T }
+		else -> bodyToMono<T>().awaitSingleOrNull()
+	}
+
+/**
  * Coroutines variant of [WebClient.ResponseSpec.toBodilessEntity].
  */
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ValentinShakhov Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26731)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26731) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ValentinShakhov Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26731)!

rstoyanchev:
This looks like a duplicate of #26491, unless I'm mistaken @sdeleuze?

valentin-apc:
awaitBody<Unit>() is not something that I wanted to achieve here.
I am accessing an endpoint that might or might not return content, therefore I need to be able to handle both empty and non-empty body. And the return type expected is a specific nullable one that I can declare with a particular class (not Any).
My current workaround is like this:

    suspend fun get() : SomeDto? = webClient.get().retrieve().bodyToMono(SomeDto::class.java).awaitSingleOrNull()

Maybe I am missing or misusing something?

sdeleuze:
After a quick first look, it seems a valid proposal. The use case makes sense and we provide already such variant for `ClientResponse` but not `WebClient.ResponseSpec`.

valentin-apc:
> After a quick first look, it seems a valid proposal. The use case makes sense and we provide already such variant for `ClientResponse` but not for `WebClient.ResponseSpec`.

Could you please provide an example of doing that with a ClientResponse? I did try it as well, but it would just hang and wait for the response forever. It could be that I did it wrong, though. 
I would appreciate that.

sdeleuze:
For `ClientResponse`, we probably miss an `awaitExchangeOrNull()` variant to use it with `awaitBodyOrNull()`. Could you please submit a PR or create a related issue ?

PR #26735, opened 2021-03-26
avoid re-initializing the HandlerMethod

refeccd:


rstoyanchev:
This is invoked during initialization and should be of limited impact. In any case we prefer not to change APIs in a maintenance release in ways may cause failures in classes that use this method. Thanks for the pull request in any case.

PR #26755, opened 2021-04-02
Add `@since` tag to WebClient.ResponseSpec.awaitBodyOrNull()

izeye:
This PR adds `@since` tag to `WebClient.ResponseSpec.awaitBodyOrNull()` that has been introduced in https://github.com/spring-projects/spring-framework/pull/26731.


snicoll:
Excellent, thanks again Johnny!

PR #26758, opened 2021-04-04
Update FactoryBean generic description in ref docs

hrybs:
The **FactoryBean** is a generic type and as a result has different method returns types from ones described in documentation.


sbrannen:

Thanks

PR #26762, opened 2021-04-05
Added synchronization control on shared resources

gvma:
Problem:
The use of shared resources, like the system properties, may result in flaky test outcomes in the case of more than one execution thread or parallel execution.

Solution:
The synchronization control allows the test planner to organize executions, assuring that a test that writes to a shared resource will be executed isolated.

### DIFF
--- a/spring-context/src/test/java/org/springframework/context/annotation/configuration/ConfigurationClassWithPlaceholderConfigurerBeanTests.java
+++ b/spring-context/src/test/java/org/springframework/context/annotation/configuration/ConfigurationClassWithPlaceholderConfigurerBeanTests.java
@@ -18,6 +18,10 @@ package org.springframework.context.annotation.configuration;
 
 import org.junit.jupiter.api.Test;
 
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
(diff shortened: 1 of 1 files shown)


cuspymd:
How much time will it save when running the test in `CONCURRENT` mode?

gvma:
> How much time will it save when running the test in `CONCURRENT` mode?

According to the JUnit 5 User Guide in https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution, if not otherwise defined, the default dynamic configuration "computes the desired parallelism based on the number of available processors/cores". This way, we believe the execution time will be a fraction of the available cores and threads proportion. In order to avoid flaky test outcomes due to concurrent access to shared resources, we also added a failsafe annotation in the specific method declaration.

snicoll:
Thanks for the PR but there is no indication that this particular test is flaky. The argument you've made could be applied to many other tests and we prefer to act on it on a case-by-case basis when flakiness actually happens.

sbrannen:
Furthermore, we do not execute any of the tests in our test suite using JUnit Jupiter's parallel execution support.

If we execute tests in parallel, that is managed by the Gradle build, and consequently, the `@Execution` annotation from JUnit Jupiter has no effect on that.

PR #26763, opened 2021-04-05
Handle removal of query param when a single (null) value is sent to `replaceQueryParam`

dhirenships:
The earlier definition of `replaceQueryParam(String name, Object... values)` didn't  remove the query param if the passed value is null. So, `replaceQueryParam("param1" , null)` would still lead to presence of queryparam `param1` in the URI. 

So if value is null, one would need to write this piece of code:
`
if(value== null) {
replaceQueryParam("param1");
} else {
replaceQueryParam("param1", value);
}
`

Hence added a new definition of the function which takes in a single argument `Object value`.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -23,6 +23,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.List;
@@ -762,6 +763,15 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 		}
 		resetSchemeSpecificPart();
 		return this;
+	}
+
+	@Override
+	public UriComponentsBuilder replaceQueryParam(String name, @Nullable Object value) {
+		Collection<?> values = null;
+		if (!ObjectUtils.isEmpty(value)) {
+			values = Collections.singleton(value);
+		}
+		return replaceQueryParam(name, values);
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
```diff
@@ -764,6 +765,15 @@ public UriComponentsBuilder replaceQueryParam(String name, Object... values) {
 		return this;
 	}
 
+	@Override
+	public UriComponentsBuilder replaceQueryParam(String name, @Nullable Object value) {
+		Collection<?> values = null;
+		if (!ObjectUtils.isEmpty(value)) {
```
cuspymd, line 771: `value` is not an array type, so it seems strange to check with `ObjectUtils.isEmpty()`.

gdagitrep: `value` doesn't need to be an array. Please check the definition of `ObjectUtils.isEmpty()` here: https://github.com/gdagitrep/spring-framework/blob/621295dbd8e53dd66901348e6f23fbdfd78b8b22/spring-core/src/main/java/org/springframework/util/ObjectUtils.java#L134

cuspymd: Oh, I didn't realize if there were other functions with different parameters. As the number of overloaded functions increases, the convenience of using the function increases, but the readability of the code seems to decrease.

gdagitrep: Agreed

gdagitrep: @cuspymd Possible to approve if no concerns?

snicoll:
I wonder if we shouldn't update the check in the vararg case, rather than introducing a new method. WDYT @rstoyanchev ?

rstoyanchev:
Thank you for the suggestion. However, the proposed change would break existing applications which rely on the present behavior. Moreover the current behavior is by design where both the `queryParam` and `replaceQueryParam` methods and their overloaded variants, treat `null` as a query parameter without a value. This aligns with similar behavior in the Servlet API on the server side where a query parameter may be present with or without a value, or not present at all. So, yes you want to remove a parameter completely, you will need such conditional logic.

PR #26776, opened 2021-04-08
Fix javadoc link syntax

KENNYSOFT:
Fix the javadoc `@link` syntax should in curly braces.

image

image


snicoll:
@KENNYSOFT good catch and thank you for making your first contribution to Spring Framework.

PR #26780, opened 2021-04-09
Add handleErrorFrame callback for STOMP handler

ppshinebl:
The previous implementation invoke handleFrame no matter the frame is "MESSAGE" nor "ERROR".  This commit add a new callback "handleErrorFrame" to the handler to separate from the message frame handling.

Reason:
1. The error frame is created by server and not following bussiness message frame format.
2. Otherwise the handler can not tell if the frame is business message or it's an error message.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompFrameHandler.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompFrameHandler.java
@@ -38,11 +38,18 @@ public interface StompFrameHandler {
 	/**
 	 * Handle a STOMP frame with the payload converted to the target type returned
 	 * from {@link #getPayloadType(StompHeaders)}.
+	 * Method {@link #handleErrorFrame(StompHeaders, byte[])} is invoked if the frame
+	 * is an error frame.
 	 * @param headers the headers of the frame
 	 * @param payload the payload, or {@code null} if there was no payload
 	 */
 	void handleFrame(StompHeaders headers, @Nullable Object payload);
 
+	/**
+	 * Handle a STOMP error frame received from server.
+	 * @param headers the headers of the frame
+	 * @param payload the payload, or {@code null} if there was no payload
+	 */
 	void handleErrorFrame(StompHeaders headers, @Nullable byte[] payload);
 
 }
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompFrameHandler.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompFrameHandler.java
@@ -43,4 +43,6 @@ public interface StompFrameHandler {
 	 */
 	void handleFrame(StompHeaders headers, @Nullable Object payload);
 
+	void handleErrorFrame(StompHeaders headers, @Nullable byte[] payload);
+
 }
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/DefaultStompSession.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/DefaultStompSession.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@ppshinebl Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26780)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26780) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ppshinebl Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26780)!

spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompFrameHandler.java
```diff
@@ -43,4 +43,6 @@
 	 */
 	void handleFrame(StompHeaders headers, @Nullable Object payload);
 
+	void handleErrorFrame(StompHeaders headers, @Nullable byte[] payload);
```
cuspymd, line 46: documentation seems be needed.

cuspymd (commented):
It seems that the year needs to be updated in the license header of the modified files.
Please refere [License section in Code-Style](https://github.com/spring-projects/spring-framework/wiki/Code-Style#license).

rstoyanchev:
To handle an `ERROR` frame, use the global `StompSessionHandler` that is provided to the `connect` method. It is the only frame it should receive because [MESSAGE](https://stomp.github.io/stomp-specification-1.2.html#MESSAGE) frames have a `subscription` header and are passed into the `StompFrameHandler` provided to the `StompSession#subscribe`. Likewise the `StompFrameHandler` for a given subscription will never see an ERROR frame and it would be misleading to expose such a method in `StompFrameHandler`. 

If anything it would be an option to expose `handleErrorFrame` in `StompSessionHandler` but considering it already has `handleFrame` and `ERROR` is the only frame that can be handled globally, it seem unnecessary. It is also not backwards compatible.

Thanks for the suggestion in any case.

ppshinebl:
1. According to STOMP1.2 [spec](https://stomp.github.io/stomp-specification-1.2.html#SUBSCRIBE)

> If the server cannot successfully create the subscription, the server MUST send the client an ERROR frame and then close the connection.

The ERROR frame could be subscription specific.

2. As a mention in the merge request as following, headers are filtered and raw headers are not passed to the StompFrameHandler,  "command" type is also invisible, so it's necessary to create a new method to handle the ERROR frame.

> The error frame is created by server and not following bussiness message frame format.

> Otherwise`` the handler can not tell if the frame is business message or it's an error message.

royclarkson:
@ppshinebl did you mean to mention @rstoyanchev on your [previous comment](https://github.com/spring-projects/spring-framework/pull/26780#issuecomment-818820463)?

rstoyanchev:
> The ERROR frame could be subscription specific.

I think you mean it can be related or triggered by. The [ERROR frame](https://stomp.github.io/stomp-specification-1.2.html#ERROR) is not linked to a subscription in any formal way. It would have to have the subscription id, just like the `MESSAGE` frame does but it doesn't. Furthermore, the connection is closed immediately after. That means that only 1 `ERROR` frame can ever be received globally, i.e. not linked to a subscription.

I don't think it makes sense to add a dedicated `handleErrorFrame` method. Use the `handleFrame` on the `StompSessionHandler` and that will be the only frame it can receive. There is no ambiguity there.

ppshinebl:
> I don't think it makes sense to add a dedicated `handleErrorFrame` method. Use the `handleFrame` on the `StompSessionHandler` and that will be the only frame it can receive. There is no ambiguity there.

The STOMP specification defines an ERROR frame, but in the handler  just mix it with normal message frame handling, that's misleading and very amibigous.
Also when handling both message frame and error frame, how can tell if it is normal message or an error? If the message context type is json, and the error frame format is defined by broker, how can you deserialize both format type with limitted header? 
For current implementation, just trying to format the error frame like message frame using the same and throwing exceptions, I think that's just a work around solution.

rstoyanchev:
No there is no mixing. The global StompSessionHandler does not get any frames.

ppshinebl:
@rstoyanchev 
Any comment?

PR #26783, opened 2021-04-09
Update ref docs regarding RequiredAnnotationBeanPostProcessor registration

hrybs:
No *RequiredAnnotationBeanPostProcessor* is automatically registered.
According to source code of *AnnotationConfigBeanDefinitionParser* class we see that there is a using of *AnnotationConfigUtils* class. Investigating the code of letter one we see that there is registration of 5 BeanPostProcessor and BeanFactoryPostProcessor classes.


sbrannen:
I was just about to make the same change, so thanks for providing the PR.

sbrannen:

Thanks

PR #26785, opened 2021-04-10
Improve @EventListener documentation in reference manual

hrybs:
1) According to a source code of *ApplicationListenerMethodAdapter* class we can see that return type of a method annotated with *@EventListener* can be not only a *Collection<?>* but also *array*.
2) Method annotated with *@EventListener* can have any visibility (public, protected, package-private and private) and can be final and static.


sbrannen:
This has been merged into `master` in a80c5fcfaf2786ab506d176e4a713ae983f9bfe5 and revised in bcc6171a72b0a7fa3e21919f2c2fc7b3daf36a30.

Thanks

PR #26787, opened 2021-04-11
Avoid exceptions when evaluating validation hints

dreis2211:
Hi,

we noticed in one of our loadtests that a substantial amount of "hidden" `NoSuchMethodException`s is being thrown when using `@javax.validation.Valid` on controller methods.
```
@PostMapping("test")
public void test(@RequestBody @Valid @NotNull TestRequest request) {
     // Some code
}
```

This is caused by calling `AnnotationUtils.getValue()` on this annotation, which has no such property. This PR bypasses the annotation utility call. I also took the freedom to refactor this logic into a common place in `ValidationUtils` to avoid copy-paste throughout the different modules.

An isolated JMH benchmark shows the following results when dealing with `@Valid`:
```
Benchmark                                         Mode  Cnt     Score     Error   Units
MyBenchmark.testNew                               avgt   10     6,253 ±   0,312   ns/op
MyBenchmark.testOld                               avgt   10  1761,995 ±  92,262   ns/op
```

As a workaround you can use `org.springframework.validation.annotation.Validated` of course, but I would prefer not to have such a drawback when using `javax.validation.Valid`.

Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/spring-context/src/main/java/org/springframework/validation/ValidationUtils.java
+++ b/spring-context/src/main/java/org/springframework/validation/ValidationUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-context/src/main/java/org/springframework/validation/ValidationUtils.java
+++ b/spring-context/src/main/java/org/springframework/validation/ValidationUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,13 +16,17 @@
 
 package org.springframework.validation;
 
+import java.lang.annotation.Annotation;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
+import org.springframework.validation.annotation.Validated;
 
 /**
  * Utility class offering convenient methods for invoking a {@link Validator}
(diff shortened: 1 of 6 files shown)


spring-context/src/main/java/org/springframework/validation/ValidationUtils.java
```diff
@@ -255,4 +259,40 @@ public static void rejectIfEmptyOrWhitespace(
...
 
+	/**
+	 * Determine any validation hints by the given annotation.
+	 * <p>This implementation checks for {@code @javax.validation.Valid},
+	 * Spring's {@link org.springframework.validation.annotation.Validated},
+	 * and custom annotations whose name starts with "Valid".
+	 * @param ann the annotation (potentially a validation annotation)
+	 * @return the validation hints to apply (possibly an empty array),
+	 * or {@code null} if this annotation does not trigger any validation
+	 * @since 5.3.6
```
dreis2211, line 270: That would be 5.3.7 now - let me know if I should add that.

jhoeller:
Hey @dreis2211 it's indeed worth handling the common `java.annotation.Valid` case explicitly there, and also worth extracting the evaluation logic. However, it'll have to be a new `org.springframework.validation.annotation.ValidationAnnotationUtils` class in order to avoid a package cycle, since it refers to the `Validated` annotation in the subpackage there. If you could refactor it that way with a `since` marker for 5.3.7, we'd be happy to roll it in...

dreis2211:
@jhoeller Thanks for the review. I just committed the request changes.

spring-context/src/main/java/org/springframework/validation/ValidationUtils.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
```
snicoll, line 2: Nitpicking, this can be restored.

snicoll (approved):

spring-context/src/main/java/org/springframework/validation/ValidationUtils.java
dreis2211: Oh man - missed that. Let me restore that.

dreis2211: Done

snicoll:
@dreis2211 I can see `ModelAttributeMethodProcessorTests` is failing. Can you please have a look?

dreis2211:
@snicoll Should be fixed - I was a bit overly ambitious and wanted to optimize the custom `@Valid` annotations case as well, which broke when the annotation was defined as an inner class.

snicoll:
Thanks for the follow-up Christoph.

PR #26788, opened 2021-04-11
Fix lazy inject for prototype

lsq27:
Everytime a method of a lazy injected prototype bean is invoked, Spring will instantiate the bean once again.
No need to invoke `doResolveDependency()` repeatedly since target is already created.


snicoll:
I am not sure I understand what you're saying. If the bean has a prototype scope, this is exactly what should be happening, the bean should be created again and its dependencies should be resolved.

snicoll:
Closing due to lack of requested feedback. If the feedback is provided we can consider reopening.

PR #26794, opened 2021-04-13
Close jar files after opening a connection

f-cramer:
Before this PR jar files that were accessed via a URLResource, would stay opened indefinitely, and thus prevent the file from being modified. This PR fixes this by closing the JarURLConnection that is opened when calling `exists()`, `contentLength()` or `lastModified()` on `Resource`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
+++ b/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
@@ -20,6 +20,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.JarURLConnection;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLConnection;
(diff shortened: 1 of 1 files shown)


f-cramer:
Is there any chance to get this into one of the next releases? Or might I be able to improve the pull request somehow to make it "mergable"?

sbrannen:
Hi @f-cramer,

We'd like to get a better understanding of your use case.

How are you obtaining the `URLResource` reference? Are you creating it yourself or getting an instance from a mechanism in Spring Framework?

Please note that the proposed changes to `AbstractFileResolvingResource` are likely at the wrong level, since that would then automatically apply to other subclasses such as `ClassPathResource` and `ServletContextResource` which  is likely undesirable. In other words, such a change would likely need to be limited to `URLResource` or custom subclasses of `URLResource`.

Speaking of the latter, would it be an option for you to create a subclass of `URLResource` for your use case?

f-cramer:
Hey @sbrannen 
the PR was opened to provide a fix for a bug in Java 8, which somehow prevented a JarURLConnection from correctly closing the corresponding file handle after usage (sorry, its been some time).
As I can not find the JDK bug report and the problem does no longer seem to be happening in JDK 17, I will close this PR

PR #26797, opened 2021-04-13
Improve ref docs regarding @Bean method visibility constraints

hrybs:
There are 2 main approaches how to use `@Configuration` annotation:

1. When `proxyBeanMethods` attribute is `true` then a *CGLIB* proxy is created. In this case the visibility of a `@Bean` annotated method can be anything except `private`. 
2. When `proxyBeanMethods` attribute is `false` then no *CGLIB* proxy is created. In this case the visibility of a `@Bean` annotated method can be anything including `private`.


sbrannen:

sbrannen:

Thanks

Issue #24611, opened 2020-02-28
Improve Javadoc of URI related options in WebClient.Builder


chenqimiao:
**Affects:** \<spring-webflux 5.1.8.RELEASE>

---

Please see a demo .
```java
public static void main(String args[]){
    DefaultUriBuilderFactory uriBuilderFactory = new DefaultUriBuilderFactory();
    uriBuilderFactory.setEncodingMode(EncodingMode.NONE);
    WebClient webClient = WebClient.builder().uriBuilderFactory(uriBuilderFactory)
        .baseUrl("http://localhots:9090").build();
    webClient.get().uri("/api/v1/query").retrieve().bodyToMono(String.class).block();
  }
```
Maybe you already know what I mean.
I want a webClient implementation that does not perform url encoding and decoding and has a baseurl. But unfortunately my code above seems to be able to meet my needs, but in fact baseurl does not take effect.
I looked through the source code，the root cause is found in the following code.
```java
private UriBuilderFactory initUriBuilderFactory() {
 if (this.uriBuilderFactory != null) {
	return this.uriBuilderFactory;
 }
 DefaultUriBuilderFactory factory = this.baseUrl != null ?
		new DefaultUriBuilderFactory(this.baseUrl) : new DefaultUriBuilderFactory();
 factory.setDefaultUriVariables(this.defaultUriVariables);
 return factory;
}
```
If `uriBuilderFactory` is set, `baseurl` will be ignored.I looked at the specific implementation of `DefaultUriBuilderFactory` again，At this point I understand that the correct way should be to set the `baseurl` through the constructor of `DefaultUriBuilderFactory`.

I think this kind of API design that If `uriBuilderFactory` is set, `baseurl` will be ignored  will cause some confusion.I wonder if you think so.Is it possible to eliminate this confusion by improving the `DefaultUriBuilderFactory` and related code.
Thank you for reading and sincerely look forward to your reply.

rstoyanchev:
They are mutually exclusive options. It's in the Javadoc for `Builder#uriBuilderFactory`:

```java
 * This is an alternative to and effectively overrides the following:
 * <ul>
 * <li>{@link #baseUrl(String)}
 * <li>{@link #defaultUriVariables(Map)}.
 * </ul>
```

The Javadoc on `baseUrl` and `defaultUriVariables` could be improved to mention they are nothing but shortcuts for passing in a `UriBuilderFactory` but you can't use both at the same time.

[...]

PR #26807, opened 2021-04-14
Remove leftover Javadoc from WebClient

mustafau:
This redundant Javadoc is from commit 5f1e4ffc4f986469c9f665e681bb34e8a7dd09e8

See https://github.com/spring-projects/spring-framework/issues/24611


snicoll:
@mustafau good catch. Thank you for making your first contribution to Spring Framework.

Issue #26719, opened 2021-03-24
Add a way to set executeExistingDelayedTasksAfterShutdown from ThreadPoolTaskScheduler


tkgregory:
### Observed behaviour

`ThreadPoolTaskScheduler` is used for methods annotated with `@Scheduled`.  

In the example of a method which is to be run at a specific time e.g. `@Scheduled( cron = "0 0 2 * * *" )` (run at 2 a.m.), the `ScheduledThreadPoolExecutor` within `ThreadPoolTaskScheduler` knows that it must execute this at the future time.

However, when I call `threadPoolTaskScheduler.shutdown()` the shutdown is being blocked until the queued task has been exectued. In the above example this would be, 2 a.m. + the time it takes to execute the task. Note that this behaviour only happens when I have configured a long enough termination period with `threadPoolTaskScheduler.setAwaitTerminationSeconds`. 

One other point. I know the `ThreadPoolTaskScheduler` will get automatically shutdown when the application shuts down, but I am doing it manually for other reasons which I won't go into in this issue.

In summary:
* I have a method scheduled for execution in the future
* my `ThreadPoolTaskScheduler` has a long terimation period (`setAwaitTerminationSeconds`)
* before that time I'm manually calling `threadPoolTaskScheduler.shutdown()`
* shutdown is blocking until the scheduled time

My requirement here is that the `ThreadPoolTaskScheduler` should shut down faster by only waiting for those tasks which are currently being processed. Tasks which are scheduled for a future time will not be executed.

### Current workaround

I found a method on `ScheduledThreadPoolExecutor` named `setExecuteExistingDelayedTasksAfterShutdownPolicy` which seems to do what I need. 

>     * Sets the policy on whether to execute existing delayed
>     * tasks even when this executor has been {@code shutdown}.
>     * In this case, these tasks will only terminate upon
>     * {@code shutdownNow}, or after setting the policy to
>     * {@code false} when already shutdown.
>     * This value is by default {@code true}.

When I set this value to `false` by configuring a custom `ThreadPoolTaskScheduler` I observe the desired behaviour.

```
        final ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler()
        {
            @Override
            protected ExecutorService initializeExecutor(final ThreadFactory threadFactory, final RejectedExecutionHandler rejectedExecutionHandler)
            {
                final ExecutorService executorService = super.initializeExecutor(threadFactory, rejectedExecutionHandler);

[...]

PR #26813, opened 2021-04-15
Easier policy setters for ThreadPoolTaskScheduler.

Aelykto:
Added an easier way to set ExecuteExistingDelayedTasksAfterShutdown and ContinueExistingPeriodicTasksAfterShutdown policies. This closes an existing issue: https://github.com/spring-projects/spring-framework/issues/26719"

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/concurrent/ThreadPoolTaskScheduler.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/concurrent/ThreadPoolTaskScheduler.java
@@ -64,7 +64,7 @@ public class ThreadPoolTaskScheduler extends ExecutorConfigurationSupport
 
 	private volatile boolean removeOnCancelPolicy;
 
-	private volatile boolean executeExistingDelayedTasksAfterShutdownPolicy;
+	private volatile boolean executeExistingDelayedTasksAfterShutdownPolicy = true;
 
 	private volatile boolean continueExistingPeriodicTasksAfterShutdownPolicy;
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@Aelykto Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26813)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26813) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Aelykto Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26813)!

spring-context/src/main/java/org/springframework/scheduling/concurrent/ThreadPoolTaskScheduler.java
```diff
@@ -64,6 +64,10 @@
 
 	private volatile boolean removeOnCancelPolicy;
 
+	private volatile boolean executeExistingDelayedTasksAfterShutdownPolicy;
```
cuspymd, line 67: Default value in `ScheduledThreadPoolExecutor` is `true`. Does it matter if it's different?

Aelykto: I believe you are actually correct and it matters for the getter:
if we don't have a scheduled executor inside our class - we return local mentioned value.

Aelykto: fyi I added a default true value corresponding to scheduled thread pool executor

jhoeller:
As per my comment on #26719, I ended up with a streamlined arrangement that does not include any policy getter methods.

PR #26816, opened 2021-04-16
Fix LocaleEditor description and formatting in reference manual

hrybs:
The official documentation contains incorrect description of *LocaleEditor* class (page 244 of pdf):
1) The language part is missing. The problem lies in incorrect formatting.
2) We can also use spaces as separators instead of underscores. We can see it from source code of *StringUtils* class.


sbrannen:
Good catch on the invalid formatting. 👍 

Thanks

PR #26822, opened 2021-04-18
Use String.startsWith() instead of String.substring() in PatternMatchUtils

stsypanov:
Instead of allocating a new substring at each call we can use `String.startsWith()`


sbrannen:

Thanks

PR #26823, opened 2021-04-19
Remove reference to Commons Collections in ColumnMapRowMapper Javadoc

mdeinum:
Remove the reference to Commons Collections.


sbrannen:
Good catch!

Looks like that should have been done in Spring Framework 3.0. 😉 

Better late than never...

Issue #25736, opened 2020-09-08
Deprecate Joda time support


bclozel:
Since 5.0, Spring Framework requires JDK8 and the current 5.3 generation is meant to be the long term support branch for the 5.x version. This issue is about deprecating our Joda time support entirely and schedule it for complete removal in the next major release.

For reference, Spring Boot removed the dependency management for Joda time since 2.3, see spring-projects/spring-boot#19699

PR #26825, opened 2021-04-19
Remove reference to Joda Time support in @DateTimeFormat Javadoc

hrybs:
According to #25736 and to be compatible with official documentation it would be great to remove mention of Joda Time support.


sbrannen:

Thanks

PR #26827, opened 2021-04-20
WebMvcConfigurer support Order

refeccd:


spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/DelegatingWebMvcConfiguration.java
```diff
@@ -46,10 +47,8 @@
 
 
 	@Autowired(required = false)
-	public void setConfigurers(List<WebMvcConfigurer> configurers) {
-		if (!CollectionUtils.isEmpty(configurers)) {
-			this.configurers.addWebMvcConfigurers(configurers);
-		}
+	public void setConfigurers(ObjectProvider<WebMvcConfigurer> configurers) {
```
snicoll, line 50: Autowiring a list already orders it the same way `ObjectProvider#orderedStream` does. Do you have any indication that's not the case? If so, adding a test that showcases that is necessary.

snicoll (changes requested):
Thanks but I think the current code is fine as it is.

liaozan:
Sorry about this. I have a `WebMvcConfigurer` in my own project, and there is also a `WebMvcConfigure` in the second party package. I want my `WebMvcConfigurer` to be processed at the end,so i added this code,but it connot work
`Autowiring a list already orders it the same way ObjectProvider#orderedStream does`, but if other beans doesn't ordered,will apply the default order `Ordered.LOWEST_PRECEDENCE`

snicoll:
@liaozan I don't think that this is accurate. Injecting a `List` uses the same algorithm that `ObjectProvider#orderedStream` as far as I know. As I've requested, can you please add a unit test that shows that the problem you've described is now fixed?

liaozan:
I mean, using `List` or `ObjectProvider` can not solve the problem, I created this pr without verification, sorry
It seems that there is no way to solve my problem. No matter which sorting method is used, the final result cannot affect the injection order of classes that are not marked and sorted in the third party.

PR #26836, opened 2021-04-21
Fix Kotlin code formatting for Inline Maps section of ref docs

hrybs:
Official documentation contains incorrect Kotlin code example (page 292 of pdf).
There are `Map<*, >` instead of `Map<*, *>` and `Map<, *>` instead of `Map<*, *>`.
The cause of this issue is incorrect formatting.


hrybs:
I see that validation failed because of some internal error. Please check it.

sbrannen:
> Official documentation contains incorrect Kotlin code example (page 292 of pdf).

In the future, please provide a link to the affected section of the reference manual, since the page numbers of the PDF can change -- for example, "in the [Inline Maps](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#expressions-inline-maps) section". If you don't want to provide a link, stating something like "in the Inline Maps section of the SpEL chapter" is also sufficient.

Thanks

sbrannen:

Thanks

PR #26850, opened 2021-04-23
Update ForwardedHeaderTransformer.java

Been24:
when set spring.webflux.base-path  and use Spring's support for handling forwarded headers.
ForwardedHeaderTransformer should combine X-Forwarded-Prefix with base-path in #apply


Been24:
@sbrannen

rstoyanchev:
No, this is meant to work as a replacement. The prefix replaces the current contextPath, if any.

PR #26855, opened 2021-04-24
Use HttpMethod enum constants instead of strings in DispatcherServlet

BryceYangS:
doDispatch method of Dispatcher servlet class uses hard coded strings, such as "GET" and "HEAD".
so, I replaced it with HttpMethod enum.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
@@ -49,6 +49,7 @@ import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.support.PropertiesLoaderUtils;
 import org.springframework.core.log.LogFormatUtils;
+import org.springframework.http.HttpMethod;
 import org.springframework.http.server.RequestPath;
 import org.springframework.http.server.ServletServerHttpRequest;
 import org.springframework.lang.Nullable;
@@ -1046,10 +1047,9 @@ public class DispatcherServlet extends FrameworkServlet {
 
 				// Process last-modified header, if supported by the handler.
 				String method = request.getMethod();
-				boolean isGet = "GET".equals(method);
-				if (isGet || "HEAD".equals(method)) {
+				if (HttpMethod.GET.matches(method) || HttpMethod.HEAD.matches(method)) {
 					long lastModified = ha.getLastModified(request, mappedHandler.getHandler());
-					if (new ServletWebRequest(request, response).checkNotModified(lastModified) && isGet) {
+					if (new ServletWebRequest(request, response).checkNotModified(lastModified) && HttpMethod.GET.matches(method)) {
 						return;
 					}
 				}
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
@@ -1047,9 +1047,10 @@ public class DispatcherServlet extends FrameworkServlet {
 
 				// Process last-modified header, if supported by the handler.
 				String method = request.getMethod();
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@BryceYangS Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26855)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26855) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@BryceYangS Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26855)!

TAKETODAY:
i think it's not a good idea

BryceYangS:
> i think it's not a good idea

@TAKETODAY Can you tell me why you think so? Your feedback will be of great help to my study.

spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
```diff
@@ -1046,10 +1047,9 @@ protected void doDispatch(HttpServletRequest request, HttpServletResponse respon
 
 				// Process last-modified header, if supported by the handler.
 				String method = request.getMethod();
-				boolean isGet = "GET".equals(method);
-				if (isGet || "HEAD".equals(method)) {
+				if (HttpMethod.GET.matches(method) || HttpMethod.HEAD.matches(method)) {
 					long lastModified = ha.getLastModified(request, mappedHandler.getHandler());
-					if (new ServletWebRequest(request, response).checkNotModified(lastModified) && isGet) {
+					if (new ServletWebRequest(request, response).checkNotModified(lastModified) && HttpMethod.GET.matches(method)) {
```
rstoyanchev, line 1052: Can you please update this to avoid matching GET twice by using a local variable as in the current code?

rstoyanchev (changes requested):

spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
BryceYangS: Thanks a lot for the review. I applied it.

dreis2211:
I just stumbled over this PR - I hope you don't mind me commenting. I wonder if we should use `HttpMethod.GET.name().equals()` rather than `HttpMethod.GET.matches()`. I did a small benchmark (on JDK 11) comparing things and the former is expectedly a bit faster because it safes the map access.

```
MyBenchmark.testMatches                         avgt    6   4,476 ±  1,153   ns/op
MyBenchmark.testNameEquals                      avgt    6   2,462 ±  0,114   ns/op
```
Given that we deal with `DispatcherServlet` here and , I wonder if we should strive for the better performing option even though it's definitely a micro-optimization.

jhoeller:
@dreis2211 I suppose we could also fine-tune the `HttpMethod.matches` implementation to `name().equals(method)` then - if known to be faster?

dreis2211:
@jhoeller I was wondering but it actually was like that before - see 3d87718fc6d14a9638682317121f9647e0441cc3. Maybe you remember why?

BryceYangS:
@jhoeller @dreis2211 Thank all of your comments. I thought the problem about a performance for a while. If there is a reason why `HttpMethod.matches ` was changed, How about using `name().equals(method)` in DispatcherServlet?

TAKETODAY:
@BryceYangS Using `name().equals(method)` is not as good as using the original

jhoeller:
No idea why I changed that back then... maybe a vague suspicion that hashing within the `Map` lookup would be faster than a String equality comparison, or maybe just a stylistic notion? In any case, we can change the current `HttpMethod.resolve` implementation back to a name comparison if that is known to be faster, and then rely on that implementation where appropriate.

rstoyanchev (approved):

rstoyanchev:
This has been merged and I've also switched HttpMethod matches to String equals comparison with b76e0c482608c664d5d1d701b995dc4ceb0ed340.

Issue #26542, opened 2021-02-13
Add TimestampGenerator contract for generating timestamps


Gorlah:
I would like to propose a new `TimestampGenerator` functional interface be added as a core utility. I feel this would be a great parallel to the existing [`IdGenerator`](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/IdGenerator.java) functional interface that already exists. The contract would look as follows:

```Java
@FunctionalInterface
public interface TimestampGenerator {

    long generateTimestamp();
}
```

For components that need to generate timestamps, a bean could easily be declared as follows:

```Java
@Configuration
class TimestampConfiguration {

    @Bean
    TimestampGenerator timestampGenerator() {
        return System::currentTimeMillis;
    }
}
```

This comes in handy when mocking dependencies in tests as relying on static methods for generating timestamps makes assertions difficult. In many of my projects I end up defining a similar interface so I would love to see it become part of the core framework.

A simple proof of concept can be seen [here](https://github.com/Gorlah/timestamp-generator/blob/main/src/main/java/com/gorlah/timestampgenerator/TimestampGeneratorApplication.java).

Gorlah:
I have never contributed to the Spring project before. If the feature is deemed worthy I would love to pick this up.

jnizet:
Isn't this what the standard [`java.time.Clock` class](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html) has been designed for?

sbrannen:
@gorlah, thanks for the proposal.

> Isn't this what the standard [`java.time.Clock` class](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html) has been designed for?

Indeed, we recommend the use of `java.time.Clock` for such use cases.

See https://github.com/spring-projects/spring-framework/issues/24884#issuecomment-612887209 and https://github.com/spring-projects/spring-framework/issues/25782#issuecomment-694722314 for related discussions.

In light of that, I am closing this issue.

Issue #25782, opened 2020-09-17
Unable to change Clock for spring framework scheduling


AstralStorm:
**Affects:** All versions, tested with 4.3.10 and verified to still be a problem in tip of master ec97089

---

It seems that Spring task scheduler implementation still uses old java.util.Date for some of its timings (esp. tasks delayed by time) and it is not possible to easily inject a different clock source, such as java.time.Clock bean instance or via SchedulerConfigurer.

Such a feature would be very valuable for unit testing time-dependent functionality, which now requires use of PowerMock or JMockit to directly patch System.currentTimeMillis, Clock.systemUTC or both.

jhoeller:
So you'd like to specify a `java.time.Clock` bean through `SchedulerConfigurer`, expecting it to be picked up in the runtime scheduling setup wherever we're currently calling `System.currentTimeMillis()` for scheduling decisions? That seems feasible, and aligned with where Bean Validation 2.0 went with its `ValidatorFactory.getClockProvider()` method.

I'm not sure we need a system-wide `Clock` bean at the core context level since we're only really using timestamps for logging and monitoring elsewhere, not for time-based decisions, so I suppose we would still prefer the real clock for those purposes even in a test setup.

jhoeller:
A related recent request for a configurable clock bean to be provided for test setups: #24884 - we'll try to revisit the topic for 5.3.

jhoeller:
If you were specifying a custom `Clock` bean for scheduling purposes, what exactly would you expect it to be used for? All time taking, including monitoring and metrics such as actual execution time, or just time-based decisions such as trigger calculations? Would you try to used a fixed clock (always returning the same timestamp) or rather some kind of custom mutable clock?

jhoeller:
I've added `Clock` exposure to `TaskScheduler` and `TriggerContext`, consistently used within our implementation instead of `System.currentTimeMillis()` and `new Date()` calls now, configurable via `setClock` calls on `ThreadPoolTaskScheduler` and `ConcurrentTaskScheduler`.

weiwang107:
Is this really working?

bclozel:
Unless I'm mistaken this is working as designed. The clock is used for calculating the next triggers. A fixed clock is not supposed to "stop time and prevent all executions".

Issue #24884, opened 2020-04-09
Provide a Mutable Clock in Spring Test Context Framework


asaikali:
Testing date based business logic is hard, since Spring is now based on Java 8 its possible to use java.time.Clock and mutable clock from [Three Ten Extra](https://www.threeten.org/threeten-extra/). It would be useful to include such a mutable clock in spring test context framework along with an injectable Clock that returns the default system clock. Below is code that implement this. 

```java
/**
 * A factory for obtaining an instance fo the java.util.clock that will be injected into by spring
 * into any components that require a clock to know the current time.
 */
public interface ClockProvider {
  Clock clock();
}
```

```java
/** Default Clock provider for the application. It returns the same clock used by Instant.now() */
@Component
class DefaultClockProvider implements ClockProvider {

  @Override
  public Clock clock() {
    return Clock.systemUTC();
  }
}
```
notice  that I have to use `@Primary` and `@TestFixtureComponent` on my test fixture implementation that provides a jsr 310 extras mutable clock. 
```java
@Primary
@TestFixtureComponent
class TestClockProvider implements ClockProvider {
  private final Logger logger = LoggerFactory.getLogger(TestClockProvider.class);

  @PostConstruct
  void init() {
    logger.warn("** ++ TestClockProvider with a Mutable Clock In Use ++ **");
  }

  @Override
  public Clock clock() {
    return MutableClock.of(Instant.now(), ZoneOffset.UTC);
  }
}
```

There is also a `@Configuration` class to create a Clock Bean so that my code can just inject a java.util.Clock where it needs to know the current time. This class is in `src/main/java`

```java
@Configuration
class ClockConfig {
  @Bean
  Clock clock(ClockProvider clockProvider) {
    return clockProvider.clock();
  }
}
```

TimeMachine class below can be autowired into test classes and used to move the clock forwards, backwards ... etc. 

``` java 
@TestFixtureComponent
@ConditionalOnBean(TestClockProvider.class)
public class TimeMachine {
  private final MutableClock clock;

  public TimeMachine(Clock clock) {
    this.clock = (MutableClock) clock;
  }

  /**
   * Advanced the clock used by the application by a certain number of seconds.
   *
   * @param seconds the number of seconds to advance the clock by
   */
  public void advanceSeconds(long seconds) {
    this.clock.add(Duration.ofSeconds(seconds));
  }
}
```

[...]

PR #26871, opened 2021-04-28
Add ApplicationEvent constructor for specifying timestamp

medwards:
Unit testing listener behaviour where the time between events is important is a pain because it is impossible to specify the time of the event.

I don't think there is a way to expose this behaviour in only test code, but I can do so if someone points out how.

Lastly, if the constructor is intentionally left out, it might be good to document this.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/ApplicationEvent.java
+++ b/spring-context/src/main/java/org/springframework/context/ApplicationEvent.java
@@ -16,6 +16,7 @@
 
 package org.springframework.context;
 
+import java.time.Clock;
 import java.util.EventObject;
 
 /**
@@ -44,6 +45,17 @@ public abstract class ApplicationEvent extends EventObject {
 	public ApplicationEvent(Object source) {
 		super(source);
 		this.timestamp = System.currentTimeMillis();
+	}
+
+	/**
+	 * Create a new {@code ApplicationEvent} with a fixed timestamp.
+	 * @param source the object on which the event initially occurred or with
+	 * which the event is associated (never {@code null})
+	 * @param clock a clock which will provide the timestamp
+	 */
+	public ApplicationEvent(Object source, Clock clock) {
+		super(source);
+		this.timestamp = clock.millis();
 	}
 
 


sbrannen:
> Unit testing listener behaviour where the time between events is important is a pain because it is impossible to specify the time of the event.

Do you need to manually specify the timestamps?

Or are you looking for a way to consume events fired during the execution of a test in the order in which they were fired?

If the latter applies to you, you may find the [`ApplicationEvents` testing feature](https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html#testcontext-application-events) useful.

sbrannen:
> Do you need to manually specify the timestamps?

As an alternative, would introducing a constructor that accepts a `java.time.Clock` meet your needs?

medwards:
> recording application events published in the ApplicationContext so that assertions can be performed against those events within tests

I don't need to check for published events in this unit test (I mock the `ApplicationEventPublisher` provided to the component)

The component I want to test reacts differently based on the amount of time between events that it listens to (using the event timestamp to determine this).

`Clock` works for me, I can mock it easily. I've updated the branch /w this alternative.

medwards:
@sbrannen : anything else I can provide in order to triage?

spring-context/src/main/java/org/springframework/context/ApplicationEvent.java
```diff
@@ -46,6 +47,17 @@ public ApplicationEvent(Object source) {
...
 
+	/**
+	 * Create a new {@code ApplicationEvent} with a fixed timestamp.
+	 * @param source the object on which the event initially occurred or with
+	 * which the event is associated (never {@code null})
+	 * @param clock a clock which will provide the timestamp
+	 */
+	public ApplicationEvent(Object source, Clock clock) {
+		super(source);
+                this.timestamp = clock.millis();
```
sbrannen, line 58:
```suggestion
		this.timestamp = clock.millis();
```

Tabs instead of spaces, please.

sbrannen:
> @sbrannen : anything else I can provide in order to triage?

Thanks for the ping. 

I'm in favor of the change, since I'm generally interested in making things easier to test.

So, I'll bring it up within the team to decide if we want to introduce this new constructor.

sbrannen:
Related to #24884, #25782, and #26542.

medwards:
Fixed whitespace.

sbrannen:

Thanks for submitting your first contribution to the Spring Framework! 👍

PR #26874, opened 2021-04-29
Improve readability of MessageSourceSupport Javadoc

hrybs:
The Javadoc of  [org.springframework.context.support.MessageSourceSupport](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/MessageSourceSupport.html#setAlwaysUseMessageFormat-boolean-) class contains unreadable piece of text for method *setAlwaysUseMessageFormat(boolean)*.
We need to replace """ with `"''"`.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/MessageSourceSupport.java
+++ b/spring-context/src/main/java/org/springframework/context/support/MessageSourceSupport.java
@@ -64,7 +64,7 @@ public abstract class MessageSourceSupport {
 	 * Set this to "true" to enforce MessageFormat for all messages,
 	 * expecting all message texts to be written with MessageFormat escaping.
 	 * <p>For example, MessageFormat expects a single quote to be escaped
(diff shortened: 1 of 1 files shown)


cuspymd (commented):
It would be nice if "true" and "false" were also expressed as code tags. And `MessageFormat` appears several times during the explanation, and it would be good if its were expressed consistently with code tag.

sbrannen:
Good catch.

The font used in the generated HTML does indeed make it difficult to discern that there are two single quotes. They are there, but the human eye practically cannot detect that.

Thanks

sbrannen:
> It would be nice if "true" and "false" were also expressed as code tags. And `MessageFormat` appears several times during the explanation, and it would be good if its were expressed consistently with code tag.

@cuspymd, I agree and made those changes in 7316dc726ab1fdc06b868cfe99ec55051e9fffd5.

PR #26876, opened 2021-04-29
Fix syntax for RequestEntity headers example in ref docs

kihyuk-sung:
Hi. There is an additional parenthesis in reference document about spring integration - RestTempate Headers part


sbrannen:

PR #26878, opened 2021-04-29
Polish

izeye:
This PR fixes some typos and polishes trivial stuff.


sbrannen:
Thanks for catching these.

I am to blame for the use of `System.err` in those tests, so I'm glad to see those gone. 😉

PR #26893, opened 2021-05-05
Fix MethodMatcher API doc description

hrybs:
Chapter *6. Spring AOP APIs*, section *6.1.1. Concepts* contains incorrect API description for *org.springframework.aop.MethodMatcher* interface. Instead of array there should be varargs.


sbrannen:

Thanks

PR #26900, opened 2021-05-06
fix aop @target bug 

caoxuhao:
**1. Problem Describe**
I found spring will proxy some classes without the designated annotation , when I use @target in the pointcut expression.

For example:

    //this is my annotation
    @Target({ElementType.METHOD, ElementType.TYPE})
    public @interface PrivilegeCheck {
        int[] roleTypes() default {};
    }
    
    //this is my aop 
    @Component
    @Aspect
    public class AopTest {
        @Around("@target(com.cxh.PrivilegeCheck)")
	    public Object around(ProceedingJoinPoint joinPoint) throws Throwable {
		    System.out.println("before aop");
		    return joinPoint.proceed();
	    }
    }
    
    //this is my target class
    @Component
    @PrivilegeCheck
    public class Father {
	    public int id = 1;
	    public int getId() {
		    return id;
	    }
    }
    
    //this is a normal final class without @PrivilegeCheck
    @Component
    public final class FinalClz {
	  public int get(){
		  return 1;
	  }
    }

The FinalClz class with also proxy, when i use the @target(com.cxh.PrivilegeCheck). But there isn't @PrivilegeCheck on the FinalClz class.  If FinalClz is not a final class, the problem just wast a little time. But if the FinalClz class is a final class, as we known the cglib can not proxy final class, so the project will start fail.

**2. My Solution**
    Because of the method findResidueInternal in ThisOrTargetAnnotationPointcut.class set the var  to shadow.getTargetVar() instead of shadow.getTargetAnnotationVar, the method visit(HasAnnotation hasAnn) of the SubtypeSensitiveVarTypeTestVisitor class in RuntimeTestWalker.class in the spring source code has an error.
    So I use the method hasAnnotation(UnresolvedType ofType) of ReferenceType to solve this problem.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/aspectj/RuntimeTestWalker.java
+++ b/spring-aop/src/main/java/org/springframework/aop/aspectj/RuntimeTestWalker.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Field;
 import org.aspectj.weaver.ReferenceType;
 import org.aspectj.weaver.ReferenceTypeDelegate;
 import org.aspectj.weaver.ResolvedType;
+import org.aspectj.weaver.UnresolvedType;
 import org.aspectj.weaver.ast.And;
 import org.aspectj.weaver.ast.Call;
 import org.aspectj.weaver.ast.FieldGetCall;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@caoxuhao Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26900)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26900) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@caoxuhao Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26900)!

PR #26905, opened 2021-05-07
UriComponentsBuilder handles invalid port numbers correctly

glqdlt:
Hello.
Fixed a bug in the'UriComponentBuilder' class.
The bug occurs in'UriComponentBuilder#fromUriString()'.
A bug occurs when the port of the argument is not a number.

This issue occurs with the 'PORT_PATTERN' regular expression.
If port is not a number, the regular expression is not captured.
This is a good regex, but it causes problems.

You can check the issue in the test code below.
```
UriComponents stringPort = UriComponentsBuilder.fromUriString("http://localhost:port/path").build();
assertThat(stringPort.getScheme()).isEqualTo("http");
assertThat(stringPort.getHost()).isEqualTo("localhost");
assertThat(stringPort.getPath()).isEqualTo("/path"); 
// expected: "/path", but was "port/path"
```

'PORT' that is not captured is captured by'PATH_PATTERN'.
`private static final String PATH_PATTERN = "([^?#]*)";`

I have fixed this bug.
Thank you.

PR #26910, opened 2021-05-08
UriComponentsBuilder when with port and query or fragment but no path

nguyensach:
When there is a port, a query, or a fragment but there is no path in URI String, UriComponentsBuilder#fromUriString() will result in a bug.

Example: 
- with `http://localhost:8080?foo=bar` URI, after running UriComponentsBuilder#fromUriString(), the port will be `8080?foo=bar`.
- with `http://localhost:8080#foo` URI, after running UriComponentsBuilder#fromUriString(), the port will be `8080#foo`.

It seems that this change #26905 affected UriComponentsBuilder#fromUriString().


glqdlt:
I have fixed this issue and additional bugs in #26918.
Thank you.

izeye:
This seems to have been resolved via https://github.com/spring-projects/spring-framework/commit/0468ef46ac33e33a6791ae9d05788c07548f0453 already.

rstoyanchev:
Yes indeed, thanks!

Issue #21184, opened 2018-03-26
Provide a mechanism to indicate that a WebFlux request mapping is blocking and should be called on a separate thread [SPR-16643]


spring-issuemaster:
**[Andy Wilkinson](https://jira.spring.io/secure/ViewProfile.jspa?name=awilkinson)** opened **[SPR-16643](https://jira.spring.io/browse/SPR-16643?redirect=false)** and commented

Spring Boot's new Actuator endpoint model allows users to write endpoints with operations that may block. We need to adapt those operations to both WebMVC and WebFlux. When adapting to WebFlux, we identify operations that may block and use `publishOn` so that the work is performed on a different thread. We currently use the operation's return type to determine whether or not it will block, erring on the side of assuming that it will.

If Framework provided a mechanism for indicating whether or not a request mapping would block (an `@Blocking` annotation has been discussed) and automatically did the `publishOn`, we could adopt this approach in Spring Boot's Actuator and provide a consistent experience across Framework and Boot.

---

**Affects:** 5.0.4

rstoyanchev:
**Team Attention:** with the team partially present, the following more current points came up.

After some time has passed, we now have R2DBC progressing, and removing a major reason for blocking methods. Project Loom is also likely to change calculations by making it less problematic to mix blocking and non-blocking. On the other hand at that point it might become less of an issue to automatically switch threads for controller methods with a blocking signature. 

We also have Blockhound to help detect issues with unintended mixing of blocking and non-blocking so we don't need to be as defensive.

For the time being we stick to the current behavior. One thing that we can do is provide some dedicated support to take a snippet of blocking code and run it off the Reactor thread so it's as easy as providing a lambda (e.g. not even having to call `publishOn`). This would also make it possible to do the same consistently in both annotated controllers and functional handlers, since otherwise there would be no answer for this in WebFlux.fn. To be decided if that belongs in Reactor or in the Spring Framework yet.

rstoyanchev:
Closing for now as there is no intention to do this but can be re-opened.

rstoyanchev:
This is now scheduled for 6.1 M2 in #30678.

PR #26912, opened 2021-05-09
Provide option to switch threads for WebFlux controller methods

fugui:
If the controller method return type is not asynchronous, it should be running in Elastic thread pool.

In WebFlux, we should support asynchronous and synchronous both.  if the controller method's return type is Mono/Flux/Future/CompleteableFuture, we should runnint this controller in synchronous(IO) thread, but if they just not want running in Asychronous Mode, they may invoke database or something else (IO operation), they just not return Mono/Flux type, and we throw them into elastic thread pool.

It's very helpful to many projects, to gain asynchronous and synchronous advances.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/BlockingPredicate.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/BlockingPredicate.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.reactive.result.method;
+
+import java.lang.reflect.Method;
+import java.util.function.Predicate;
+
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+
+/**
+ * Help to predicate the controller method blocking or not.
+ * it's default non-blocking in webflux.
+ *
+ *  @author Gui Fu
+ *  @since 5.3.8
+ */
+public class BlockingPredicate implements Predicate<Method> {
+	@Override
+	public boolean test(Method method) {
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@fugui Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26912)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26912) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@fugui Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-framework&pullRequestId=26912)!

rstoyanchev:
WebFlux does support both blocking and non-blocking but by default it assumes non-blocking handling. This is irrespective of the return type actually, and a controller method that returns an actual value is not necessarily blocking. We don't want to force a switch of threads when we don't know if the controller is actually performing a blocking operation or not. There would have to be something more explicit to indicate that.

You can see a longer discussion and decision under #21184.

fugui:
>> For the now the conclusion is that controller methods need to explicitly use .publishOn to switch threads where needed.

Yes, I got it.  but when you are in a big business development team,  it not easy to enable everyone to understand asynchronous , and be professional in reactor.   

And I also think @Async annotation is not a good solution, it's too boring to write everywhere.  

Might I add a global configuration to enable global controller returntype detection?

`		if (isAsyncReturnType(method.getReturnType()) && autoBlockingDetection ) {
			argumentValues.publishOn(Schedulers.boundedElastic());
		}
`

rstoyanchev:
`@Async` does not fit because it expects `void` or `Future` as the return type. It's also confusing to use because in WebFlux the default assumption is "async" so it would be something to label a method or class as "blocking".

As a compromise we could expose a property on `RequestMappingHandlerMapping` with  a `Predicate<HandlerMethod>` to help decide if a method is blocking or not, in which case we would switch threads or not. We'd call this only if the return type is not asynchronous. You could then simply return `true` or use some other strategy to determine based on the controller method.

fugui:
Thanks,  I think this " on RequestMappingHandlerAdapter with a Predicate<HandlerMethod> to help decide" is the best suggestion, it give architects great flexibility to create business frameworks.

rstoyanchev:
On further thought, and team discussion, we'll defer this to 6.0, to consider more comprehensively, and expose as a first-class feature. It is something that affects more broadly all places where we invoke annotated methods on a reactive foundation, which is in WebFlux but also in RSocket. We may consider adding an annotation as well as a global option that enables some strategy to determine methods that may block.

In the mean time, to enable globally, you could extend `RequestMappingHandlerAdapter` and override its `handle` method.

rstoyanchev:
This is now superseded by  #30678.

PR #26914, opened 2021-05-09
Resolve Values from Mapping Path for Mvc Uri Builder

### LINKED ISSUE #26795: MvcUriComponentsBuilder should resolve property placeholders in request mapping paths
I'm using Spring Framework 5.3.4 with Spring Boot 2.4.3.

I have a controller declared in this way:

```java
@Controller
@RequestMapping("${saml.sp.controller-base-mapping}")
public class ServiceProviderController {
  /...
}
```
So, the property `saml.sp.controller-base-mapping` is externalized in `application.properties` file: if its value is `sp`, this should be equivalent to: `@RequestMapping("sp")`. And indeed Spring Web MVC handles this well.

However, if I then use `MvcUriComponentsBuilder` to retrieve the corresponding URL:

```java
MvcUriComponentsBuilder.fromController(ServiceProviderController.class).build().toURL();
```

This returns a `UriComponents` of: `http://localhost:8081/${saml.sp.controller-base-mapping}`

First problem: I would have expected the externalized property to be already resolved (so to get `http://localhost:8081/sp`), because this should be an expansion mechanism that comes BEFORE path variable expansion.

Indeed (and here comes the second problem), if I try to do:

```java
MvcUriComponentsBuilder.fromController(ServiceProviderController.class).buildAndExpand("sp");
```

I then get a `UriComponents` of: `http://localhost:8081/$sp`

This is somewhat expected, since path variable expansion is another story and does not use the "$" marker.

lower-case:
Fixes #26795

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/MvcUriComponentsBuilder.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/MvcUriComponentsBuilder.java
@@ -37,12 +37,14 @@ import org.springframework.cglib.proxy.Callback;
 import org.springframework.cglib.proxy.Enhancer;
 import org.springframework.cglib.proxy.Factory;
 import org.springframework.cglib.proxy.MethodProxy;
+import org.springframework.context.EmbeddedValueResolverAware;
 import org.springframework.core.DefaultParameterNameDiscoverer;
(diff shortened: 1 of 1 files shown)


rstoyanchev (changes requested):
I think the `StringValueResolver` should be an instance field, for use with the instance `withXxx` methods. And it needs test cases.

snicoll:

PR #26917, opened 2021-05-09
Introduce soft assertions for MockMvc

wyhasany:
It happens very often that MockMvc is used in
heavyweight integration tests. It's no use to waste
time to check if another condition has been fixed or
not. Soft assertion could help a lot to check all
conditions at once even if one of them fail.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/ResultMatcher.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/ResultMatcher.java
@@ -16,6 +16,9 @@
 
 package org.springframework.test.web.servlet;
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * A {@code ResultMatcher} matches the result of an executed request against
  * some expectation.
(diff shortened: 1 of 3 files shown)


sbrannen:
Related to #20963

spring-test/src/main/java/org/springframework/test/web/servlet/ResultMatcher.java
```diff
@@ -70,4 +73,32 @@ static ResultMatcher matchAll(ResultMatcher... matchers) {
...
+				catch (AssertionError assertionException) {
+					if (failedMessages == null) {
+						failedMessages = new ArrayList<>();
+					}
+					failedMessages.add(assertionException.getMessage());
+				}
+			}
+			if (failedMessages != null) {
+				throw new AssertionError(String.join("\n", failedMessages));
+			}
```
nguyensach, line 100:
```suggestion
			List<String> failedMessages = new ArrayList<>();
			for (ResultMatcher matcher : matchers) {
				try {
					matcher.match(result);
				}
				catch (AssertionError assertionException) {
					failedMessages.add(assertionException.getMessage());
				}
			}
			if (!failedMessages.isEmpty()) {
				throw new AssertionError(String.join("\n", failedMessages));
			}
```

wyhasany:
@nguyensach I've applied your suggestions. Do you need any action on my side?

nguyensach (approved):
LGTM 🚀

spring-test/src/main/java/org/springframework/test/web/servlet/ResultMatcher.java
```diff
@@ -70,4 +73,29 @@ static ResultMatcher matchAll(ResultMatcher... matchers) {
...
+	 */
+	static ResultMatcher matchAllSoftly(ResultMatcher... matchers) {
+		return result -> {
+			List<String> failedMessages = new ArrayList<>();
+			for (ResultMatcher matcher : matchers) {
+				try {
+					matcher.match(result);
+				}
+				catch (AssertionError assertionException) {
+					failedMessages.add(assertionException.getMessage());
```
kkocel, line 92: to be consistent with WebTest version you could consider adding indices here https://github.com/spring-projects/spring-framework/pull/26969/files#diff-396f20e4c7198aab834e4d234e312f65e9fedf27d32582f336d1e8b0ca0a0328R521

kkocel (changes requested):

spring-test/src/main/java/org/springframework/test/web/servlet/ResultMatcher.java
wyhasany: good idea, fixed

sbrannen:
Current work on this issue can be seen in the following feature branch: https://github.com/spring-projects/spring-framework/compare/main...sbrannen:issues/gh-26917-MockMvc-soft-assertions

sbrannen:
Thanks for the PR, @wyhasany! 👍 

This has been merged into `main` in 35bec8102b014feac18c5b096b39c3592c27b76e, refined in 5f47d3be227e115c57b03e73f5df97993c44fdd8 and cd078eaad83e2c6cb441719086b460e665c20f35, and then fully revised in dd9b99e13d6b30462fb168799e6bbca1456123da.

In the end we decided to introduce a new `andExpectAll()` method in `ResultActions` and to move away from static methods in `ResultMatcher`. In fact, we even deprecated the existing `matchAll()` method in `ResultMatcher`.

PR #26918, opened 2021-05-10
Bug Fix the port regular expression in'UriComponentsBuilder'.

glqdlt:
Sorry
I've checked a bug in a commit I've fixed earlier.(https://github.com/spring-projects/spring-framework/pull/26905)
This bug was reported in #26910.
If there are'Url Query' and'Url Fragments', it is a normal case, but a bug occurs.
I have fix this.


ASarco:
This causes another problem, if by mistake a a uri like this is built: `http://localhost:8085api/something` , the previous regex would still extract the correct port number and path. However with this change, it will extract `8085api` as the port and of course will throw a NumberFormatException.

PR #26927, opened 2021-05-10
Improve support for port numbers in allowedOriginPattern of CorsConfiguration

### LINKED ISSUE #26926: Improve support for port numbers in allowedOriginPattern of CorsConfiguration
Currently it's not possible to support any port or multiple, specific ports with a single pattern. The workaround is to create two origin patterns such as `"https://*.example.com"` and `"https://*.example.com:*"`.

One idea is to enhance the pattern syntax with a comma separated list of ports or a wildcard:
```
https://*.example.com:[8080]
https://*.example.com:[8080,9000]
https://*.example.com:[*]
```

This was originally reported in #26892.

korektur:
This is to support ports in cors origin patterns.
Closes gh-26926

Example:
```
https://*.example.com:[8080]
https://*.example.com:[8080,9000]
https://*.example.com:[*]
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
+++ b/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


cuspymd (commented):
It would be nice if a description of the added patterns was added to the document of the `setAllowedOriginPatterns()` method.

rstoyanchev (changes requested):
Thanks for the PR. This looks good overall. 

One observation. It looks like I can now match an implicit port (e.g. `"https://abc.org"`) or any port (including an implicit port) with `"*"`, but if I want to match both implicit and specific others, then I need two patterns. I suppose if one defines  `"https://abc.org:[8080,8081]"` it's impossible to know just by looking whether it matches the implicit port or not, but it's probably best to err on the side of allowing less.

In summary I agree with the way you've done it. However, we'll need to make sure the syntax and these nuances are covered well in `CorsConfiguration#setAllowedOriginPatterns`. Let me know if you intend to do that or otherwise I can take it from there.

korektur:
@rstoyanchev sure, I can add some documentation on that.

korektur:
@rstoyanchev I've added some documentation, can you review please?

PR #26928, opened 2021-05-11
Optimize some scenes which need not proxy

caoxuhao:
If using @target in the pointcut expression, spring will create proxy for the classed which are in maybe case and will check the aop when user call the functions of proxy class.
If there's a final class in these proxy functions. the project will start failed. 
So I return the target object, for these final class without proxy.


caoxuhao:
If using @target in the pointcut expression, spring will create proxy for the classed which are in maybe case 
and will check the aop when user call the functions of proxy class.
    If there's a final class in these proxy functions. the project will start failed.
    So I return the target object, for these final class without proxy.

PR #26951, opened 2021-05-18
Polish PORT_PATTERN in UriComponentsBuilder

izeye:
This PR polishes `PORT_PATTERN` in `UriComponentsBuilder` by removing some parts that don't seem to be necessary.

I might be missing something. If it's intentional, it would good to have some tests that demonstrate why they are necessary as the current tests including `UriComponentsBuilderTests` pass without them.


rstoyanchev:
Ah yes, this is related to support for the port provided as a URI variable, see 8fbd310b07698d18517cbe3481a4fca7f09ccc9a, including the tests added.

izeye:
@rstoyanchev Thanks for the feedback! Since the changes made in https://github.com/spring-projects/spring-framework/pull/26905, they don't seem to be necessary as all the written tests pass without them. Am I missing something?

rstoyanchev:
You're right on that. We're now matching anything up to the path, query, or fragment and later integer parsing for the port will either succeed or fail.

PR #26952, opened 2021-05-18
Introduce ResponseEntity.internalServerError()

dreis2211:
Hi,

this PR introduces a new convenience method `ResponseEntity.internalServerError()`.

In our code we have this sort of construct once in a while:

```java
		try {
			doSomething();
		} catch (SomethingNotFoundException ex) {
			return ResponseEntity.notFound().build();
		} catch (Exception ex) {
			return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
		}
		return ResponseEntity.noContent().build();
```

That obviously works but I'm always looking for the convenience method to handle the 500 status code, which doesn't exist.

Let me know what you think.

Cheers,
Christoph


sbrannen:

Thanks

PR #26967, opened 2021-05-22
Fix broken Javadoc tags

izeye:
This PR fixes broken Javadoc tags.

### DIFF
--- a/spring-core/src/main/java/org/springframework/asm/ClassWriter.java
+++ b/spring-core/src/main/java/org/springframework/asm/ClassWriter.java
@@ -79,7 +79,7 @@ public class ClassWriter extends ClassVisitor {
 
   /**
    * The access_flags field of the JVMS ClassFile structure. This field can contain ASM specific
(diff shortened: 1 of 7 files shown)


sbrannen:
Out of curiosity, how did you notice all of those broken tags?

----

Regarding the Javadoc for `org.springframework.asm.ClassWriter`, the following is our policy on changes to the ASM code.

> Please refrain from modifying classes under `org.springframework.asm`, `org.springframework.cglib`, and `org.springframework.objenesis`. Those include repackaged forks of the third-party libraries ASM, CGLIB, and Objenesis. Any refactoring to those classes should take place upstream in the originating repository. The Spring Framework will then pick up the changes when syncing with official updates of the forked third-party libraries.

So I'll simply revert that change when merging.

sbrannen:

Thanks

izeye:
@sbrannen Thanks for the review and spotting the accidental change on the ASM module.

> Out of curiosity, how did you notice all of those broken tags?

I just found one and then searched to fix similar occurrences.

izeye:
I created https://gitlab.ow2.org/asm/asm/-/merge_requests/321 to fix it in the upstream module.

PR #26968, opened 2021-05-22
Refactor ConcurrentBeanWrapperTests to use @RepeatedTest

eas5:
Problem:

No problem is found in this test. This is a refactoring suggestion that simplifies test design by eliminating a repetition structure whose increment was not being used during the repetition and can be replaced by an annotation.

Solution:

The `@RepeatedTest` annotation provides the total number of repetitions desired. Each invocation of a repeated test behaves like the execution of a regular `@Test` method with full support for the same lifecycle callbacks and extensions, which improves test design, eases future maintenance activities when needed, and better informs which repetition failed.


sbrannen:
As the author of [`@RepeatedTest`](https://github.com/junit-team/junit5/commit/2199096a62eb549bfb12c0be010324a2735f72e9), it was my pleasure to merge your first PR for the Spring Framework.

Thanks

PR #26969, opened 2021-05-23
Introduce soft assertions for WebTestClient

wyhasany:
## Overview

It happens very often that `WebTestClient` is used in heavyweight integration tests. It's no use to waste time to check if another condition has been fixed or not. Soft assertions could help a lot to check all conditions at once even if one of them fails.

## Proposal

New API would look like as follows:

```java
client.get().uri("/test")
	.exchange()
	.expectAllSoftly(
		exchange -> exchange.expectStatus().isOk(),
		exchange -> exchange.expectBody(String.class)
			.isEqualTo("It works!")
	);
```

## Related Issues

- #26917

### DIFF
--- a/spring-test/src/test/java/org/springframework/test/web/reactive/server/samples/SoftAssertionTests.java
+++ b/spring-test/src/test/java/org/springframework/test/web/reactive/server/samples/SoftAssertionTests.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.test.web.reactive.server.samples;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.test.web.reactive.server.WebTestClient;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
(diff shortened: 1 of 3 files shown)


spring-test/src/test/java/org/springframework/test/web/reactive/server/samples/SoftAssertionTests.java
```diff
@@ -0,0 +1,75 @@
...
+	@Test
+	public void testAllFails() throws Exception {
+		assertThatExceptionOfType(AssertionError.class).isThrownBy(() ->
+				this.client.get().uri("/test")
+						.exchange()
+						.expectAllSoftly(
+								exchange -> exchange.expectStatus().isBadRequest(),
+								exchange -> exchange.expectBody(String.class).isEqualTo("It won't work :(")
+						)
+		).withMessage("Status expected:<400 BAD_REQUEST> but was:<200 OK>\nResponse body expected:<It won't work :(> but was:<It works!>");
```
cuspymd, line 63: If there are many asserts, and some of them fail, it can be difficult to determine which assert failed.
It would be nice if the indices were also included in the error message.

wyhasany: Resolved, pls check the changes in the code

kkocel (approved):

sbrannen:
This PR is blocked until #26917 has been merged into `main`.

sbrannen:
This has been merged into `main` in 25dca404138e85e8864e4bfa4b37c88f236a559e and revised in dd9b99e13d6b30462fb168799e6bbca1456123da.

Thanks for the 2nd _soft assertions_ PR! 👍

rstoyanchev:
Sorry for the late review. Overall looks good, but I've spotted on follow-up issue to be fixed.

Currently individual expectations like `responseSpec.expectStatus()` are internally surrounded with `ExchangeResult#assertWithDiagnostics` which wraps any AssertionError in another `AssertionError` that adds the request and response details. That means when multiple assertions are executed, those details are also printed multiple times.

I'm not immediately sure how to solve this, but we'll need to find something.

sbrannen:
> Currently individual expectations like `responseSpec.expectStatus()` are internally surrounded with `ExchangeResult#assertWithDiagnostics` which logs request and response details. That means when multiple assertions are executed, those details are also logged multiple times.

Good catch.

> I'm not immediately sure how to solve this, but we'll need to find something.

OK. Let's brainstorm.

PR #26980, opened 2021-05-27
Fix typo in code example

hunjipo:
Today, I found this error while browsing the spring web reactive document.


snicoll:
Good catch @hunjipo and thank you for making your first contribution to Spring Framework.

Issue #26474, opened 2021-01-29
Exclude User Destination Messages in Broker Message Handler


ellers-espirit:
**Affects:** v5.3.3

We used WebSockets within Spring Boot 2.3.5 (Spring Framework 5.2.10). After updating to Spring Boot 2.4.2 (Spring Framework 5.3.3), we noticed that our WebSocket mechanism was no longer working.

We looked at #26118, but the discussion there was about duplicate subscriptions and wrongly registered /user destinations. In our case we have only a single subscription per sessionId/subscriptionId.

WebSocket is basically configured like this:

```
@Configuration
@EnableWebSocketMessageBroker
public class WebSocketConfig implements WebSocketMessageBrokerConfigurer {
	@Override
	public void configureMessageBroker(final MessageBrokerRegistry registry) {
		registry.setApplicationDestinationPrefixes("/app");
		registry.setUserDestinationPrefix("/user");
	}
	@Override
	public void registerStompEndpoints(final StompEndpointRegistry registry) {
		registry.addEndpoint("/websocket").withSockJS();
	}
}
```

In our application we use user-specific WebSocket messages as a push mechanism. The client subscribes to _/user/events_, the server sends messages to _/user/{username}/events_ using _SimpMessagingTemplate#convertAndSendToUser_. This worked as expected up to Version 5.3.3.

```
_simpMessagingTemplate.convertAndSendToUser("username", "/events", pojo);
```

After a little bit of debugging, we've noticed that there might now be a problem with user-specific subscriptions due to a change with #24395. _DestinationCache#calculateSubscriptions_ now operates on an exact destination equals-comparison when using non-pattern-based subscriptions.

```
		@NonNull
		private LinkedMultiValueMap<String, String> calculateSubscriptions(String destination) {
			LinkedMultiValueMap<String, String> sessionsToSubscriptions = new LinkedMultiValueMap<>();
			DefaultSubscriptionRegistry.this.subscriptionRegistry.forEachSubscription((sessionId, subscriptionDetail) -> {
				if (subscriptionDetail.isAntPattern()) {
					...
				}
				else if (destination.equals(subscriptionDetail.getDestination())) {
					...
```

[...]

Issue #26118, opened 2020-11-19
Spring Boot 2.4 / Spring Framework 5.3 broken WebSocket subscription using simple message broker 


BenS89:
We used WebSockets within Spring Boot 2.3.5 using a simple message broker. After the update to Spring Boot 2.4 / Spring Framework 5.3, we noticed that our web socket mechanism is not working anymore. We use the following web socket config: 

```java
@Configuration
@EnableWebSocketMessageBroker
public class WebSocketConfig implements WebSocketMessageBrokerConfigurer {

    public void configureMessageBroker(MessageBrokerRegistry config) {
        config.enableSimpleBroker("/topic", "/user");
        config.setApplicationDestinationPrefixes("/app");
        config.setUserDestinationPrefix("/user");
    }

    public void registerStompEndpoints(StompEndpointRegistry registry) {
        registry.addEndpoint("/websocket");
        registry.addEndpoint("/websocket").setAllowedOrigins("http://localhost:8080").withSockJS();
    }

    @Bean(name = "websocketLocaleResolver")
    public SessionLocaleResolver websocketLocaleResolver() {
        SessionLocaleResolver websocketLocaleResolver = new SessionLocaleResolver();
        websocketLocaleResolver.setDefaultLocale(Locale.GERMANY);

        return websocketLocaleResolver;
    }

}
```

A look at the `DefaultSubscriptionRegistry` shows that there changed a lot comparing to Spring Boot 2.3.5. We could nail down the issue because we saw that even though our web socket subscription was successful within the `DefaultSubscriptionRegistry`, the method `this.subscriptionRegistry.findSubscriptions(message)` within the `SimpleBrokerMessageHandler` always returns an empty map (in contrast to Spring Boot 2.3).

```java
protected void sendMessageToSubscribers(@Nullable String destination, Message<?> message) {
		MultiValueMap<String,String> subscriptions = this.subscriptionRegistry.findSubscriptions(message);
...
```

When rolling back to Spring 2.3.5, the same web socket business logic works fine with the old implementations of  `DefaultSubscriptionRegistry` and `SimpleBrokerMessageHandler`.

We also noticed that with Spring Boot 2.3.5 the `DestinationCache` returns a map including a `SessionId` and `SubscriptionId` (expected behavior). Using Spring Boot 2.4.0, the returned LinkedMultiValueMap sessionIdToSubscriptionIds is always empty.

```java
public LinkedMultiValueMap<String, String> getSubscriptions(String destination) {
[...]

PR #26986, opened 2021-05-27
Allow multiple destinations per subscription in DefaultSubscriptionRegistry

alienisty:
Issue #24395 broke "user" destinations because it made subscription-id->destination a one to one association whereas it is a one to many.

This restores such state while keeping the optimisations done in #24395.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
@@ -216,7 +216,7 @@ public class DefaultSubscriptionRegistry extends AbstractSubscriptionRegistry {
 			SessionInfo info = this.sessionRegistry.getSession(sessionId);
 			if (info != null) {
 				for (String subscriptionId : subscriptionIds) {
-					info.getSubscription(subscriptionId).forEach(subscription -> {
+					info.getSubscriptions(subscriptionId).forEach(subscription -> {
 						if (evaluateExpression(subscription.getSelector(), message)) {
 							result.add(sessionId, subscription.getId());
 						}
@@ -420,7 +420,7 @@ public class DefaultSubscriptionRegistry extends AbstractSubscriptionRegistry {
 			return this.subscriptionMap.values().stream().flatMap(Set::stream);
 		}
 
-		public Stream<Subscription> getSubscription(String subscriptionId) {
+		public Stream<Subscription> getSubscriptions(String subscriptionId) {
 			return this.subscriptionMap.getOrDefault(subscriptionId, emptySet()).stream();
 		}
 
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
@@ -16,17 +16,23 @@
 
 package org.springframework.messaging.simp.broker;
 
+import static java.util.Collections.emptySet;
+
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Queue;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiConsumer;
+import java.util.stream.Stream;
 
 import org.springframework.expression.EvaluationContext;
 import org.springframework.expression.Expression;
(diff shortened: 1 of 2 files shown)


spring-messaging/src/main/java/org/springframework/messaging/simp/broker/DefaultSubscriptionRegistry.java
```diff
@@ -411,24 +414,25 @@ public SessionInfo removeSubscriptions(String sessionId) {
...
-			return this.subscriptionMap.values();
+		public Stream<Subscription> getSubscriptions() {
+			return this.subscriptionMap.values().stream().flatMap(Set::stream);
 		}
 
-		@Nullable
-		public Subscription getSubscription(String subscriptionId) {
-			return this.subscriptionMap.get(subscriptionId);
+		public Stream<Subscription> getSubscription(String subscriptionId) {
+			return this.subscriptionMap.getOrDefault(subscriptionId, emptySet()).stream();
```
cuspymd, line 424: It doesn't seem like a function that only returns one item anymore, isn't `getSubscriptions` a more appropriate name?

alienisty: It definitely is.

rstoyanchev (commented):
Subscription id's must be unique. From the [STOMP spec](https://stomp.github.io/stomp-specification-1.2.html#SUBSCRIBE):

> Since a single connection can have multiple open subscriptions with a server, an id header MUST be included in the frame to uniquely identify the subscription. 

This has been discussed at length under #26118 and #26474. I believe this is likely the same issue.

Can you share what your `@EnableWebSocketMessageBroker` config looks like? In particular what, if any, destination prefixes, are configured.

alienisty:
> Subscription id's must be unique. From the [STOMP spec](https://stomp.github.io/stomp-specification-1.2.html#SUBSCRIBE):
> 
> > Since a single connection can have multiple open subscriptions with a server, an id header MUST be included in the frame to uniquely identify the subscription.
> 
> This has been discussed at length under #26118 and #26474. I believe this is likely the same issue.
> 
> Can you share what your `@EnableWebSocketMessageBroker` config looks like? In particular what, if any, destination prefixes, are configured.

We don't use the annotation, we extend DelegatingWebSocketMessageBrokerConfiguration directly:
```java
@Configuration
public static class MessageBrokerConfiguration
    extends DelegatingWebSocketMessageBrokerConfiguration
    implements ApplicationListener<ApplicationPreparedEvent> {

  private final ApplicationProperties.WebSocket.StompBrokerRelay config;

  private final WebSocketMessageBrokerStats webSocketMessageBrokerStats;

  public MessageBrokerConfiguration(
      ApplicationContext context,
      ApplicationProperties properties,
      @Lazy WebSocketMessageBrokerStats webSocketMessageBrokerStats) {
    // We need to ensure that the application context is available when the STOMP sub-protocol bean is built,
    // otherwise the session events won't work. (see WebSocketMessageBrokerConfigurationSupport.stompWebSocketHandlerMapping())
    this.setApplicationContext(context);
    this.config = properties.getWebSocket().getStompBrokerRelay();
    this.webSocketMessageBrokerStats = webSocketMessageBrokerStats;
  }

  @Override
  public void onApplicationEvent(ApplicationPreparedEvent event) {
    long statsLoggingPeriod = config.getStatsLoggingPeriod();
    webSocketMessageBrokerStats.setLoggingPeriod(MINUTES.toMillis(statsLoggingPeriod));
  }
}
```

As you can see there is nothing changing how the protocol is implemented and the superclass creates and registers an instance of SubProtocolWebSocketHandler and UserDestinationMessageHandler into the ExecutorSubscribableChannel. In ExecutorSubscribableChannel.sendInternal(), the same Message, therefore with the same subscription-id header is passed to both handlers. When the message is a "SUBSCRIBE", both hanlders will try to subscribe their destinations with the same subscription-id.

I can appreciate what the STOMP protocol specification says, I would argue, though, that the subscription-id must be unique per connection not per destination, so much so that messages need to specify both subscription-id and destination; if they were a one-to-one association, the destination header would be redundant.
We are also using the latest StompJs library which also claims to be a rewrite which is more adherent to the STOMP spec and it works just fine receiving messages to different destinations but same subscription-id.

At the very least we have a problem with how user destinations are registered.

rstoyanchev:
I think your issue is exactly the same. 

If there aren't any destination prefixes configured for messages to the broker, then the broker ends up having all messages all messages forwarded to it, but "/user" prefixed messages aren't meant to go directly there. Instead, `UserDestinationMessageHandler` is the only thing that should handle such messages [as described in the reference](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#websocket-stomp-user-destination). It transforms the destination to a "/queue" prefixed, unique destination, and that's what's then sent to the broker. So in your case the broker sees both the "/user/" and the transformed "/queue" prefixed messages with the same id, but it should see only the latter.

Can you try configuring prefixes for the messages that should go to the broker? See the discussion and recommendations under #26474.

alienisty:
Hi @rstoyanchev, setting the user destination prefix to "/user" fixes the problem, but I think that either the reference documentation should be changed or such prefix should be "/user" by default. In fact, the referece says:
>An application can send messages that target a specific user, and Spring’s STOMP support recognizes destinations prefixed with /user/ for this purpose. For example, a client might subscribe to the /user/queue/position-updates destination. This destination is handled by the UserDestinationMessageHandler and transformed into a destination unique to the user session.

which is true, but the reference documentation doesn't mention the user destination prefix, and for sending a message using a message template, it only says:
>You can send a message to user destinations from any application component by, for example, injecting the SimpMessagingTemplate created by the Java configuration or the XML namespace. (The bean name is brokerMessagingTemplate if required for qualification with @Qualifier.) The following example shows how to do so:
```java
@Service
public class TradeServiceImpl implements TradeService {

    private final SimpMessagingTemplate messagingTemplate;

    @Autowired
    public TradeServiceImpl(SimpMessagingTemplate messagingTemplate) {
        this.messagingTemplate = messagingTemplate;
    }

    // ...

    public void afterTradeExecuted(Trade trade) {
        this.messagingTemplate.convertAndSendToUser(
                trade.getUserName(), "/queue/position-updates", trade.getResult());
    }
}
```

which now it does not work unless we set the prefix to "/user".

rstoyanchev:
I have added a3655c48581dd3977de02f13dd3a7ecc3dcde6c2 to add some guidance to the reference docs as suggested and will close this issue. 

For 6.0, we could consider having the broker explicitly filter out messages with user destinations when it is itself not configured with any specific prefixes, but that would be better as a separate issue.

PR #26993, opened 2021-05-27
Add extra test case for creating bean

tyc31316:
I've noticed the method
public Object createBean(Class<?> beanClass, int autowireMode, boolean dependencyCheck) throws BeansException at line 256in java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
is not covered in the current test suit, and hence I added a simple unit test in order to improve the coverage.


snicoll:
Thanks for the PR but I can see a number of unrelated changes so I am going to close this one.

PR #27001, opened 2021-05-29
Fix @Transactional examples regarding method visibility

hrybs:
Spring documentation "*Data Access*", paragraph "*1.4.6. Using @Transactional*" contains incorrect description of **@Transactional** annotation that is used on a class level and incorrect corresponding code snippets.
* According to source codes of *AnnotationTransactionAspect* and *ProxyTransactionManagementConfiguration* the **@Transactional** annotation when used on a class level is only applied to public method even if we use AspectJ.
* The are 2 incorrect code snippets that show class level **@Transactional** annotation for package-private methods.

### DIFF
--- a/src/docs/asciidoc/data-access.adoc
+++ b/src/docs/asciidoc/data-access.adoc
@@ -1314,19 +1314,19 @@ Consider the following class definition:
 	@Transactional
 	public class DefaultFooService implements FooService {
 
-		Foo getFoo(String fooName) {
+		public Foo getFoo(String fooName) {
 			// ...
 		}
 
-		Foo getFoo(String fooName, String barName) {
+		public Foo getFoo(String fooName, String barName) {
 			// ...
 		}
 
-		void insertFoo(Foo foo) {
+		public void insertFoo(Foo foo) {
 			// ...
 		}
 
-		void updateFoo(Foo foo) {
+		public void updateFoo(Foo foo) {
 			// ...
 		}
 	}
@@ -1356,7 +1356,7 @@ Consider the following class definition:
 	}
 ----
 
-Used at the class level as above, the annotation indicates a default for all methods
+Used at the class level as above, the annotation indicates a default for all public methods
 of the declaring class (as well as its subclasses). Alternatively, each method can
 get annotated individually. Note that a class-level annotation does not apply to
 ancestor classes up the class hierarchy; in such a scenario, methods need to be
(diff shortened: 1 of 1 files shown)


sbrannen:
Good catch!

Thanks

sbrannen:
> the `@Transactional` annotation when used on a class level is only applied to public method even if we use AspectJ.

On second thought, that's not entirely true: it depends on the `publicMethodsOnly` flag in `AnnotationTransactionAttributeSource`.

When using `@EnableTransactionManagement`, `ProxyTransactionManagementConfiguration.transactionAttributeSource()` uses the default `AnnotationTransactionAttributeSource` constructor which sets `publicMethodsOnly` to `true`.

However, in the Spring TestContext Framework, the `TransactionalTestExecutionListener` sets the `publicMethodsOnly` flag to `false`.

In light of that, I'll modify the wording in the docs accordingly.

sbrannen:
As a proof of concept, I put together the following test case which demonstrates that `@Transactional` -- when declared solely at the class level -- can be configured to apply to all non-private methods in a standard Spring application as well.

```java
import javax.sql.DataSource;

import org.springframework.jdbc.datasource.DataSourceTransactionManager;
import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseBuilder;
import org.springframework.transaction.PlatformTransactionManager;
import org.springframework.transaction.annotation.AnnotationTransactionAttributeSource;
import org.springframework.transaction.annotation.EnableTransactionManagement;
import org.springframework.transaction.annotation.ProxyTransactionManagementConfiguration;
import org.springframework.transaction.annotation.Transactional;
import org.springframework.transaction.interceptor.TransactionAttributeSource;

import static org.springframework.transaction.support.TransactionSynchronizationManager.isActualTransactionActive;

@SpringJUnitConfig
class NonPublicTxMethods {

	@Test
	void test(@Autowired Service service) {
		assertThat(isActualTransactionActive()).isFalse();
		service.doSomething();
	}

	@Configuration
	@EnableTransactionManagement
	static class Config {

		@Bean
		DataSource dataSource() {
			return new EmbeddedDatabaseBuilder().generateUniqueName(true).build();
		}

		@Bean
		PlatformTransactionManager transactionManager(DataSource dataSource) {
			return new DataSourceTransactionManager(dataSource);
		}

		/**
		 * Overrides {@link ProxyTransactionManagementConfiguration#transactionAttributeSource()}.
		 */
		@Bean
		TransactionAttributeSource transactionAttributeSource() {
			return new AnnotationTransactionAttributeSource(false);
		}

		@Bean
		Service service() {
			return new Service();
		}
	}

	@Transactional
	static class Service {

		void doSomething() {
			assertThat(isActualTransactionActive()).isTrue();
		}
	}

}
```

hrybs:
Good explanation! But when we use AspectJ the annotation is only applied to public methods.

sbrannen:
> But when we use AspectJ the annotation is only applied to public methods.

Yes, it's an unfortunate limitation of the predefined pointcuts in `AnnotationTransactionAspect` that limits the application to `public` methods when `@Transactional` is only declared at the class level.

However, non-public methods can at least be directly annotated with `@Transactional` when using AspectJ CTW or LTW.

PR #27015, opened 2021-06-02
Link alternate documentation formats

Buzzardo:
Link the HTML to the PDF and vice-versa.

### DIFF
--- a/src/docs/asciidoc/index.adoc
+++ b/src/docs/asciidoc/index.adoc
@@ -11,6 +11,14 @@ independent of release cadence, are maintained externally on the project's
 https://github.com/spring-projects/spring-framework/wiki[*Github Wiki*].
 ****
 
+ifdef::backend-html5[]
+NOTE: This documentation is also available as {doc-root}/spring-framework/docs/{spring-version}/reference/pdf/index.pdf[PDF].
+endif::[]
+
+ifdef::backend-pdf[]
+NOTE: This documentation is also available as {doc-root}/spring-framework/docs/{spring-version}/reference/html/index.html[HTML].
+endif::[]
+
 [horizontal]
 <<overview.adoc#overview, Overview>> :: history, design philosophy, feedback,
 getting started.


rstoyanchev:
Thanks for the PR!

I'm noticing that more than half the page is taken up by logos, titles, and links to other resources. This is what it looks like in my browser:

image

I realize that most of it is unrelated to this change, but nevertheless it seems a good opportunity to bring this up and see how we can optimize it so that the actual content does not require scrolling.

Buzzardo:
We lose a lot of space in the border around the version (especially) and around the wiki link (though less so). The border around the note for other formats could be thinner, too.

Also, other projects put the content that Spring Framework is keeping in the wiki into a section within the docs (so in a link within the list of links). Aside from this layout issue, another concern I have is that readers who are familiar with our other docs won't think to look for a wiki link (even though it is called out at the top - people do get into ruts).

Consider Spring Boot's layout (https://docs.spring.io/spring-boot/docs/current/reference/html/). The spacing seems much more friendly, because it is considerably tighter and gets more content into the same space. (Note that I would move the author list to the bottom, because very few readers are looking for it.) Could Framework adopt Boot's styling? Boot has a backend project that Framework could use (and that I hope all Spring projects will eventually use). Similarly, can we move the content in the wiki into a topic within the regular contents list?

rstoyanchev:
Thanks for the feedback and pointers.

For the use of the wiki, at some point we moved all cross-version information there. This is quite intentional as it allows us to target multiple generations and versions from one place and it allows us to continuously make updates. 

I've made some additional updates. The link to the wiki is now folded into the main index of links at the end. I've also moved the link to the PDF links below the main index. I don't think it needs to be the first thing to see unless we can fine a more compact way to display it.

For the version, we're looking for a way to display it in a less obtrusive way. It's controlled by the [revision line](https://docs.asciidoctor.org/asciidoc/latest/document/revision-line/) property of Asciidoctor and it doesn't seem easy to customize but we might be able to target it very specifically with the style for `#header .details span:first-child` in the spring.css. @oodamien what do you think?

sbrannen:
> For the version, we're looking for a way to display it in a less obtrusive way. It's controlled by the [revision line](https://docs.asciidoctor.org/asciidoc/latest/document/revision-line/) property of Asciidoctor and it doesn't seem easy to customize but we might be able to target it very specifically with the style for `#header .details span:first-child` in the spring.css. @oodamien what do you think?

We ended up taking a different approach as outlined in the commit message for 71995a90876fd5b0aa39259df35476fcc3b0c5ef.

The upcoming snapshot docs should therefore look like the following.

image

PR #27016, opened 2021-06-03
Adjust expected IdentityHashMap sizes in ClassUtils

aaguilera:
Proposed change aimed to keep coherence between variable declaration 
and static initialization code, which inserts 9 values, not 8, into 
these two variables.


sbrannen:
Although changing the `expectedMaxSize` supplied to the `IdentityHashMap(int)` constructor from 8 to 9 likely will not change the actual initial (and final) capacity of the map, I don't see any harm in aligning the _expected_ maximum size with the known maximum size for an `IdentityHashMap`.

sbrannen:

Thanks for your first contribution to the Spring Framework!

PR #27031, opened 2021-06-07
Document that class-level @ResponseStatus is inherited by @ExceptionHandler methods

hrybs:
The [Javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/ResponseStatus.html) of `@ResponseStatus` doesn't mention that when `@ResponseStatus` is applied to a class then it is inherited also by `@ExceptionHandler` annotated methods.


sbrannen:

Thanks

PR #27045, opened 2021-06-10
fix: Correct the condition about isEnforceReadOnly

nurinamu:
Although isEnforceReadOly is true, Statement might be not READ ONLY if definition.isReadOnly is not true.
So I replaced && to || to use the flag as named "Enforced".


mp911de:
The check deliberately verifies both conditions to be `true`. Enabling read only behavior on the database requires another SQL statement to be issued and in cases where transactions consist of a single statement, issuing another one leads to significant overhead compared to the unit of work. 

Therefore, we provide this optimization as opt-in feature that needs to be activated on the transaction definition and the transaction manager. We have the same arrangement on the JDBC transaction manager variant.

PR #27047, opened 2021-06-10
Support $$ quotes in JDBC ScriptUtils

tyc31316:
This is an attempt to support $$ quote as mentioned in this issue: https://github.com/spring-projects/spring-framework/issues/27018
I tried to add the $$ logic into the method of static void splitSqlScript in spring-jdbc/src/main/java/org/springframework/jdbc/datasource/init/ScriptUtils.java


OrangeDog:
@tyc31316  did you mean to close this?

Issue #26985, opened 2021-05-27
DateTimeFormatterRegistrar should fall back to ISO parsing by default


abccbaandy:
**Affects:** 5.3.7

Currently `DateTimeFormatterRegistrar` behavior is different from default JDK

For example, given `OffsetDateTime.parse(XXX)`, if XXX can be parsed this way, it should be parsed by `DateTimeFormatterRegistrar` too.

sbrannen:
Please note that changing the current behavior would be a breaking change.

The documentation for `DateTimeFormatterRegistrar` states that it does not use ISO formatting by default, but that can be changed via `DateTimeFormatterRegistrar#setUseIsoFormat(true)`.

Can you please provide more information regarding your exact use case?

abccbaandy:
Yes, I know `setUseIsoFormat`. I create this issue to discussion is there any chance to change it to true by default.

The reason I already said in first comment. This option should be true by default.

About my real use case : I use RFC 3339 format in all time field in my rest api. It work well with json(jackson) without any config, the default config is good and standard.
But when use it in query parameter, it fail.

abccbaandy:
Yes, I can do that, but I want make it default without any config.

sbrannen:
Thanks for the feedback, @abccbaandy.

Since changing the default would be a breaking change, we will revisit this topic for Spring Framework 6.

jhoeller:
I've slightly repurposed this issue for revisiting our fallback parsing behavior there. Based on the scenario described above, we should indeed leniently accept the default `parse` format of every `java.time` type.

This is different from switching `useIsoFormat` to `true` since that affects printing as well and generally bypasses other style settings. Also, this new ISO fallback behavior only kicks in if no fallback patterns have been specified, so it's easy to bypass.

jhoeller:
Related enhancements in 6.0: #1215, #19846

PR #27048, opened 2021-06-10
Issue #26985 change date time default formate

Yin-Jui:
Link of issue: https://github.com/spring-projects/spring-framework/issues/26985

Issue: DateTimeFormatterRegistrar should use ISO formats by default

Currently, if the users want to format the time in ISO format, they need to annotate the controller method by @DateTimeFormat(iso = ISO.DATE_TIME). However, the issue is using ISO format should be a default setting without any configuration.
For this issue, I looked into the DefaultFormattingConversionService class and there is an addDefaultFormatters method, which specifies the default handling of date-time values.
To make the ISO format the default setting, I first create a DateTimeFormatterRegistrar instance and then invoke setUseIsoFormat(true). Finally, register this formatter through the FormatterRegistry which is passed in as the argument.


sbrannen:
Thanks for the PR, @Yin-Jui.

Unfortunately, the PR drops support for standard date/time formats as well as Joda Time. In addition, we will need to update tests and documentation in the reference manual. We may need to perform some other actions as well.

That's why issue #26985 is labeled `status: pending-design-work` (for design work to be performed by the core team) and currently slated for 6.0 M1. Thus the team will look into this topic later this year.

In light of that I am closing this PR.

PR #27063, opened 2021-06-15
Avoid unnecessary Map wrapping in DefaultPathSegment

Dunemaster:
Avoid creating unnessary wrapper for empty map in DefaultPathSegment


sbrannen:
Thanks for the PR!

It turns out that your proposal introduces a bug, which in turn made me notice an existing bug.

Namely, any map returned by `PathSegment#parameters()` must be immutable, and this includes the _shared_ empty map (`EMPTY_PARAMS`).

I am therefore repurposing this PR as a bug report and will soon push a fix for the bug.

sbrannen:
Superseded by #27064

PR #27066, opened 2021-06-15
Reducing the level of code nesting for some methods

serjnarbut:
Trivial code formatting - reduce code nesting on conditionals in HttpHeaders

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -703,27 +703,25 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 	 */
 	public List<Charset> getAcceptCharset() {
 		String value = getFirst(ACCEPT_CHARSET);
-		if (value != null) {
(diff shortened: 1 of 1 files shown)


spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -1575,28 +1573,28 @@ public void clearContentHeaders() {
...
-						}
-						else {
-							result.add(matcher.group(1));
-						}
+		if (values == null) {
+			return Collections.emptyList();
+		}
+		List<String> result = new ArrayList<>();
+		for (String value : values) {
+			if (value != null) {
```
cuspymd, line 1581: It would be good to `continue` if `value` is `null`.

serjnarbut: yes, indeed. Fixed

poutsma:
Thank you for the PR but due to these changes the positive case, where the header has a value, is no longer handled first. We feel that it's better to keep the code the current way.

PR #27071, opened 2021-06-16
Add maxAttempts to ExponentialBackOff

garyrussell:
If you wish to stop after a certain number of attempts with an
`ExponentialBackOff` you have to calculate the `maxElapsedTime`
corresponding to the number of attempts.

See https://github.com/spring-projects/spring-kafka/blob/0ee2d49ea525de038d315bc243b62f1004a26dc6/spring-kafka/src/main/java/org/springframework/kafka/support/ExponentialBackOffWithMaxRetries.java#L74-L84

Add a new property to make it more convenient to stop after a
certain number of attempts.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOff.java
+++ b/spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOff.java
@@ -76,6 +76,9 @@ public class ExponentialBackOff implements BackOff {
 	 */
 	public static final long DEFAULT_MAX_ELAPSED_TIME = Long.MAX_VALUE;
 
+	/**
+	 * The default maximum attempts.
+	 */
 	public static final int DEFAULT_MAX_ATTEMPTS = Integer.MAX_VALUE;
 
 	private long initialInterval = DEFAULT_INITIAL_INTERVAL;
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOff.java
```diff
@@ -160,11 +168,33 @@ public void setMaxElapsedTime(long maxElapsedTime) {
...
 	 */
 	public long getMaxElapsedTime() {
 		return this.maxElapsedTime;
 	}
 
+	/**
+	 * The maximum number of attempts after which a call to
+	 * {@link BackOffExecution#nextBackOff()} returns {@link BackOffExecution#STOP}.
+	 * @param maxAttempts the maxAttempts.
+	 * @see #setMaxElapsedTime(long)
```
artembilan, line 182: Probably some `@since` has to be here on public API.
Let's start from the current one and we'll see how this PR going to be approved!

artembilan (changes requested):
A couple Checkstyle issues:

```
[ant:checkstyle] [ERROR] /tmp/build/f663b219/git-repo/spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOff.java:79:9: Missing a Javadoc comment. [JavadocVariable]

[ant:checkstyle] [ERROR] /tmp/build/f663b219/git-repo/spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOff.java:195:24: Reference to instance variable 'maxAttempts' needs "this.". [RequireThis]
```

See Concourse CI report for PR.

PR #27075, opened 2021-06-18
Delegate a call to bean for the RequestMappingHandlerAdapter.getLastModifiedInternal(..) method

omarchenko4j:
Add delegation to the bean for the RequestMappingHandlerAdapter.getLastModifiedInternal(..) method if the bean implements the LastModified interface.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerAdapter.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerAdapter.java
@@ -92,6 +92,7 @@ import org.springframework.web.method.support.InvocableHandlerMethod;
 import org.springframework.web.method.support.ModelAndViewContainer;
 import org.springframework.web.servlet.ModelAndView;
 import org.springframework.web.servlet.View;
+import org.springframework.web.servlet.mvc.LastModified;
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Thanks for the suggestion, but this really needs to be more on the controller method level. Have you seen the [options](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc-caching-etag-lastmodified) for a controller method?

MarchenkoProjects:
Yes, I saw it, but it will be located at the controller level anyway.

Let's imagine that my controller only manages one page, then it is very convenient to implement the logic of calculating the last modification without clogging the logic of page processing.

Take a look at an example:
```java
@Controller
public class TestPageController implements LastModified {

    @GetMapping("/test/{id}")
    public ModelAndView getTest(@PathVariable String id) {
        // Model preparation.
        return new ModelAndView("pages/test_page");
    }

    @Override
    public long getLastModified(HttpServletRequest request) {
        return -1; // Calculate the last modified.
    }
}

```

rstoyanchev:
**Team Decision:** `LastModified` isn't a good fit for annotated controllers with where multiple controller methods can be added at any time. Also it is an outdated mechanism to be deprecated immediately in 5.3.x in favor of using `ServerWebRequest#checkNotModified`, or in a controller method, returning a `ResponseEntity` with eTag and lastModified headers set, which in turn invokes the `checkNotModified`.

 that should no longer be in use, and we are going to deprecate it in 5.3.x.

PR #27081, opened 2021-06-19
Using entrySet iterator in ServletServerHttpRequest#getBodyFromServle…

GungnirLaevatain:
Using entrySet iterator can avoid the Map.get(key) lookup, which is more efficient


jhoeller:
I eventually went with a custom variant that uses `entrySet()` but leaves the `Writer` interaction as-is, avoiding the intermediate `StringBuilder`. Thanks for the pull request in any case, and also for the additional test case!

PR #27083, opened 2021-06-19
Only track DisposableBean targets in ThreadLocalTargetSource

slandrum00:
Currently all target objects are added to the targetSet, this looks to be done so when the application is shutdown destroy can be called on each target object. But if the target is not a DisposableBean, then there is nothing to destroy and the object reference is being held on to for no good reason. On app servers that turn over threads often (jboss/undertow) this can lead to a memory leak. Since the thread/threadlocal is gone but the targetSet has still has it's reference for destroying, but there is nothing to destroy.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/target/ThreadLocalTargetSource.java
+++ b/spring-aop/src/main/java/org/springframework/aop/target/ThreadLocalTargetSource.java
@@ -18,6 +18,7 @@ package org.springframework.aop.target;
 
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.springframework.aop.IntroductionAdvisor;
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @slandrum00,

Thanks for submitting your first pull request ever!

> But if the target is not a DisposableBean, then there is nothing to destroy and the object reference is being held on to for no good reason.

That's unfortunately not true. The `DisposableBean` interface is only one of the mechanisms supported for destroying those target beans. The full list of supported mechanisms is documented in the class-level Javadoc for `org.springframework.beans.factory.support.DisposableBeanAdapter`:

> Adapter that implements the `DisposableBean` and `Runnable` interfaces performing various destruction steps on a given bean instance:
> 
> * DestructionAwareBeanPostProcessors;
> * the bean implementing DisposableBean itself;
> * a custom destroy method specified on the bean definition.

Thus, limiting the support to only `DisposableBean` would be a breaking change.

PR #27108, opened 2021-06-30
Fix cross reference in ref docs for `@Bean` method autowiring

arnovh:
I am still studying spring technology, so perhaps I don't understand this concept. However I think the resolution mechanism is more comparable with constructor mode autowiring than with constructor based dependency injection (argument matching)?

### DIFF
--- a/src/docs/asciidoc/core/core-beans.adoc
+++ b/src/docs/asciidoc/core/core-beans.adoc
@@ -8027,8 +8027,7 @@ parameter, as the following example shows:
 ----
 
 
-The resolution mechanism is pretty much identical to constructor-based dependency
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @arnovh,

Thanks for submitting your first PR!

> However I think the resolution mechanism is more comparable with constructor mode autowiring than with constructor based dependency injection (argument matching)?

I can certainly understand why you would think that, since they are closely related concepts.

However, your proposed change would direct the user to autowiring modes that are only applicable when using Spring's XML configuration. Whereas, for an `@Bean` method the mechanism for autowiring by type for method parameters is effectively the same as for constructors and static factory methods.

In general, an `@Bean` method is a form of a factory method.

PR #27115, opened 2021-07-01
Model.set() Kotlin extension method does not allow null value

gnagy:
This parameter is `@Nullable` in the underlying addAttribute method


snicoll:
Good catch, thanks @gnagy

PR #27122, opened 2021-07-02
Added description for HandlerInterceptor

sangyongchoi:
Added description for HandlerInterceptor

I know that the posthandle is not intended to be executed when an error occurs in the handler .
But the docs didn't show that.
It says that called after HandlerAdapter actually invoked the handler, but it seems not enough.
If this explanation is added, I think it will be a better docs for users.

thank you


rstoyanchev:
I've made a slightly different change. Thanks for the suggested update in any case!

PR #27124, opened 2021-07-03
Optimize HttpStatus value resolution from code

imanushin:
The purpose - optimize function ```HttpStatus.resolve```. Previous version worked with O(N) complexity, new works with O(1) complexity and doesn't require any allocation.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpStatus.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpStatus.java
@@ -420,7 +420,7 @@ public enum HttpStatus {
 	/**
 	 * Index of array value is the same with HttpStatus code.
 	 * For example, CODE_TO_VALUE[404]==HttpStatus.NOT_FOUND
-	 * Missing values are null: = CODE_TO_VALUE[1]==null
+	 * Missing values are null: CODE_TO_VALUE[1]==null
 	 */
 	private static final HttpStatus[] CODE_TO_VALUE = getCodeToValueArray();
 
--- a/spring-web/src/main/java/org/springframework/http/HttpStatus.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpStatus.java
@@ -18,6 +18,8 @@ package org.springframework.http;
 
 import org.springframework.lang.Nullable;
 
+import java.util.Arrays;
+
 /**
  * Enumeration of HTTP status codes.
  *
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/HttpStatus.java
```diff
@@ -557,13 +558,33 @@ public static HttpStatus valueOf(int statusCode) {
...
+				.map(HttpStatus::value)
+				.max(Integer::compareTo)
+				.get();
+
+		final HttpStatus[] result = new HttpStatus[maximumCode + 1];
+
+		for (HttpStatus status : enumValues) {
+			// all deprecated values are strictly after actual, which allows condition below
+			// See also test HttpStatusTests
+			if (result[status.value] == null) {
```
diguage, line 582: Maybe the condition is unnecessary.

bclozel:
I'm not sure this PR is really improving the situation.

A quick micro-benchmark shows that this PR speeds up `valueOf(int)` by a factor 3-4x. It's a nice improvement, but not big enough to make a real difference in our web benchmarks.

On the other hand, this approach consistently allocates an array of 500+ size, for 60+ elements in it. Also, in my opinion, these changes don't help with readability. There would be other ways to speed up the index search for this, but iterating over enum a few values is such a common case that I don't think it's really worth it.

Overall, I'm not sure we should merge these changes.

sbrannen:
> Overall, I'm not sure we should merge these changes.

I am also not convinced that the proposed change would provide much benefit in the big picture.

diguage:
How about giving up the array and using `HashMap<Integer, HttpStatus >` to save memory?

dreis2211:
@bclozel I'm surprised that you see an improvement. On my machine with JDK 11 I see actually the contrary. The proposed solution is slower for the most common case of status code 200.
```
Benchmark            Mode  Cnt  Score   Error  Units
MyBenchmark.testNew  avgt   10  2,459 ± 0,022  ns/op
MyBenchmark.testOld  avgt   10  3,710 ± 0,018  ns/op
```
With random values even more so:
```
Benchmark            Mode  Cnt   Score   Error  Units
MyBenchmark.testNew  avgt   10   8,784 ± 0,621  ns/op
MyBenchmark.testOld  avgt   10  33,358 ± 3,076  ns/op
```

```
@State(Scope.Thread)
public class MyBenchmark {

	@State(Scope.Thread)
	public static class BenchmarkState {
		public int status = 200;
		public Random random = new Random();
	}

	@Benchmark
	public HttpStatus testOld(BenchmarkState state) {
		return HttpStatus.resolveOld(state.random.nextInt(500));
	}

	@Benchmark
	public HttpStatus testNew(BenchmarkState state) {
		return HttpStatus.resolveNew(state.random.nextInt(500));
	}
}
```
Apart from that I share the opinion from @bclozel that the proposed solution doesn't help with readability.

bclozel:
@dreis2211 I'm confused - in the case of the average time mode, isn't it the lower the better? Or am I misreading the old vs. new variants?

Anyway I'm closing this PR because it doesn't seem we're achieving much here, and this part of our codebase was never identified as a bottleneck of any kind, so there's little value in working on other possible optimizations.

dreis2211:
@bclozel No, you're not misreading things, I'm just stupid. I was testing with throughput before on something else.

bclozel:
@dreis2211 No worries - you were brave enough to publish your benchmark, I'm sure mine are flawed in 10 different ways. My initial attempt with throughput was not really conclusive because error margins were really high - there might have been a lot of noise on my laptop at the time.

PR #27127, opened 2021-07-04
Avoid duplicated code by using map

YooInKeun:
I propose to avoid duplicated code by using map. I think it will improve simplicity and readability.
Thank you : )


snicoll:
Duplicate of #27126 and #27125. @YooInKeun please stop creating duplicate PRs as it generates unnecessary noise for the watchers of this repository. To update a PR you can push additional commit on the branch you've used to create the PR.

YooInKeun:
@snicoll I'm sorry. I'll be careful about that. 🙏

bclozel:
I don't think this is really an optimization. This code path can be hot in SpEL and I don't think we should introduce Map lookups for readability purposes. This also doesn't remove any duplication at all.

I'm closing this as a result.

PR #27165, opened 2021-07-13
Allow builder to set constructor detector for object mapper

SatsG:
This is to allow the Jackson Object Mapper Builder to be able to accept a constructor detector. The idea is that it can then be used by [this](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java#L145) to configure it via properties.

see [gh-27178](https://github.com/spring-projects/spring-boot/issues/27178)


snicoll:
@SatsG thank you for your first PR to the spring framework (very complete I should say!). Unfortunately, `ConstructorDetector` is new in Jackson 2.12 and we need to keep runtime compatibility to Jackson 2.9 still.

The issue you referred to can be implemented in any case as the builder exposes a customizer you can use to access the `ObjectMapper` before it is returned, see `postConfigurer`.

PR #27188, opened 2021-07-19
Expand relative URI for RestTemplate if possible

quaff:
```java
RestTemplate template = new RestTemplate();
template.setUriTemplateHandler(new DefaultUriBuilderFactory("http://localhost:8080"));
template.getForObject("/users/{id}", String.class, 1);
template.getForObject("/users/1", String.class);
// following will works as above after this commit
template.getForObject(URI.create("/users/1"), String.class);

template.exchange(RequestEntity.method(HttpMethod.GET, "/users/{id}", 1).build(), String.class);
template.exchange(RequestEntity.method(HttpMethod.GET, "/users/1").build(), String.class);
// following will works as above after this commit
template.exchange(RequestEntity.method(HttpMethod.GET, URI.create("/users/1")).build(), String.class);
```
It will make Spring Boot `TestRestTemplate` much more cleaner, save it from ubiquitous `applyRootUriIfNecessary`
https://github.com/spring-projects/spring-boot/blob/05a64ecb2ca237d963866665d3a0e7fd1920fee6/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/client/TestRestTemplate.java#L942-L948

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
@@ -100,6 +100,7 @@ import org.springframework.web.util.UriTemplateHandler;
  * @author Juergen Hoeller
  * @author Sam Brannen
  * @author Sebastien Deleuze
+ * @author Yanming Zhou
  * @since 3.0
  * @see HttpMessageConverter
  * @see RequestCallback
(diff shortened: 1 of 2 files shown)


quaff:
`RestTemplate` is still relevant, could you review this PR? @rstoyanchev

rstoyanchev (commented):
This will result in double encoding if the URI has encoded characters, e.g. `URI.create("/users/1%202")` results in `http://localhost:8080/users/1%25202`.

Generally with the method variants that take `URI`, we assume the `URI` has been fully initialized externally and we don't touch it further. That way if the URI template doesn't work for you for some reason, you can always take over.

I'm not sure why you want to pass a URI for a relative URL, or how you prepare it, but you could use the same `UriBuilderFactory` to that instead of `URI.create`.

PR #27193, opened 2021-07-20
Allow specifying of CoroutineContext in invokeSuspendingFunction

lukas-krecan:
In case someone needs to specify default `CoroutineContext` for all
WebMvc calls, he can use [this trick](https://stackoverflow.com/questions/68437892/configure-default-kotlin-coroutine-context-in-spring-mvc). If we can specify `CoroutineContext` in `invokeSuspendingFunction`,
we do not have to duplicate so much code.

Default parameter value not used as to not break the Java signature.


lukas-krecan:
Related PR https://github.com/spring-projects/spring-framework/pull/27195

poutsma:
Thanks for submitting a PR, @lukas-krecan. Our apologies for taking so long to resolve it.

PR #27194, opened 2021-07-21
Do not close response for InputStreamResource

quaff:
Closing underlying stream should be responsibility of InputStreamResource
```java
ResponseEntity<InputStreamResource> entity = restTemplate.getForEntity(url, InputStreamResource.class);
try (InputStream is = response.getBody().getInputStream()) {
	// do something
}
```
"java.io.IOException: closed" is fixed after this commit


bclozel:
`RestTemplate` is closing the response by design.
See #21424 and https://github.com/spring-projects/spring-framework/issues/19448#issuecomment-453446935 for more background about this. I'm declining this PR as a result.

Thanks!

PR #27195, opened 2021-07-21
Allow overriding invokeSuspendingFunction

lukas-krecan:
In case someone needs to specify default CoroutineContext for all
WebMvc calls, he can use this [trick](https://stackoverflow.com/questions/68437892/configure-default-kotlin-coroutine-context-in-spring-mvc). If we can override invokeSuspendingFunction, we do not have to
duplicate so much code.


lukas-krecan:
Related PR https://github.com/spring-projects/spring-framework/pull/27193

PR #27208, opened 2021-07-24
Reduce visibility of constructors in ScriptException

xiaoliang-liu:
Reduce the visibility of the constructors in `org.springframework.jdbc.datasource.init.ScriptException`.

There is currently no need to use `public` visibility here.


sbrannen:
Hi @xiaoliang-liu,

Thank you for submitting your first PR to the Spring Framework!

> There is currently no need to use `public` visibility here.

This is technically correct. However, if we were to make this change we would want to make similar changes to the R2DBC variant of `ScriptException`, `DataAccessException`, `NestedRuntimeException`, and other classes across the framework.

Furthermore, this is essentially a matter of coding style for constructors that would otherwise be `public` if the class in question were not `abstract`. As such, those constructors can be considered part of the public API for people extending those classes (even if `protected` would technically serve the same goal).

In light of that, we prefer to keep these constructors `public`.

PR #27215, opened 2021-07-27
Fix JDK proxy causing controller to fail to register

wangkaijin:
Some scenarios require JDK proxy, but the controller cannot be registered if JDK proxy is used. I have seen the problem mentioned in the previous issue, but the solution is to have the user change the configuration to force the cglib proxy to be enabled. I think this is not a good solution.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractHandlerMethodMapping.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractHandlerMethodMapping.java
@@ -266,7 +266,8 @@ public abstract class AbstractHandlerMethodMapping<T> extends AbstractHandlerMap
 			Object bean = null;
 			try {
 				bean = obtainApplicationContext().getBean(beanType);
-			} catch (BeansException ignore) {
+			}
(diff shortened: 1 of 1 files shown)


rstoyanchev:
For JDK proxies, [annotations are expected](https://docs.spring.io/spring-framework/docs/5.3.8/reference/html/web.html#mvc-ann-requestmapping-proxying) on the interface. We have some samples in [JdkProxyControllerTests](https://github.com/spring-projects/spring-framework/blob/42edef0bccb75a09523040b9567866281b574629/spring-webmvc/src/test/java/org/springframework/web/servlet/mvc/annotation/JdkProxyControllerTests.java#L101-L142) that demonstrate that.

wangkaijin:
@rstoyanchev My English is not very good, do you mean that I need to execute the JdkProxyControllerTests test class? I have already passed the JdkProxyControllerTests test locally

rstoyanchev:
What I mean is that this is a feature and not a bug. Please, check the documentation:
https://docs.spring.io/spring-framework/docs/5.3.8/reference/html/web.html#mvc-ann-requestmapping-proxying

PR #27220, opened 2021-07-28
Clarify that ClientRequest.from(..) also copies body

AlexejTimonin:
- Add to javadoc that ```ClientRequest.from(..)``` will also copy body from the given request.
- Add test to show that the body inserter will be copied over.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientRequest.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientRequest.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientRequest.java
@@ -125,8 +125,8 @@ public interface ClientRequest {
 	// Static builder methods
 
 	/**
-	 * Create a builder with the method, URI, headers, and cookies of the given request.
-	 * @param other the request to copy the method, URI, headers, and cookies from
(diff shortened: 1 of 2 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientRequest.java
```diff
@@ -125,8 +125,8 @@
 	// Static builder methods
 
 	/**
-	 * Create a builder with the method, URI, headers, and cookies of the given request.
-	 * @param other the request to copy the method, URI, headers, and cookies from
+	 * Create a builder with the method, URI, headers, cookies and body of the given request.
+	 * @param other the request to copy the method, URI, headers, cookies and body from
```
qiangmao, line 129: 这是一个评论

sbrannen, line 129:
```suggestion
	 * Create a builder with the method, URI, headers, cookies, attributes, and body of the given request.
	 * @param other the request to copy the method, URI, headers, cookies, attributes, and body from
```

sbrannen (changes requested):
I've requested one small change to document that "attributes" are copied as well.

Please also update the copyright date in `ClientRequest` and `DefaultClientRequestBuilderTests`.

spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientRequest.java
AlexejTimonin: Thank you for your review and suggested changes. I will address them this evening.

AlexejTimonin:
@sbrannen I could not see any tests for attributes in ```DefaultClientRequestBuilderTests``` so I updated the ```from()``` test to include attributes.

spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientRequest.java
AlexejTimonin: I don´t understand your review, could you take it in English?

sbrannen (approved):

PR #27223, opened 2021-07-29
Fix reference to Optional.isPresent() in ObjectUtils.isEmpty()

matvs:
`ObjectUtils.isEmpty()` Javadoc mentions `Optional.empty()`, whereas `empty()` returns an empty `Optional` and `Optional.isPresent()` checks if `Optional` indeed is not empty.


sbrannen:

Issue #25801, opened 2020-09-23
Thread-scoped bean creation freezes if dependent bean is retrieved before dependency bean 


nguyenquoc:
**Affects:** 5.2.9

---

It seems that if a dependent thread-scoped bean is retrieved from the `ApplicationContext` before the thread-scoped dependency, then bean creation could lock up in an infinite loop, never to return again.

Regression bug. Works fine with spring-framework 5.2.7, fails with spring-framework 5.2.9. May be a bug in spring-test rather than spring-context. May be hard to reproduce, hoping that this limited information is sufficient to lead to a fix.

`SimpleThreadScope` likely relevant since `SimpleThreadScope` has an infinite loop - the bean creation freezes in:

```java
// org.springframework.context.support.SimpleThreadScope.class, the scope map is empty
public Object get(String name, ObjectFactory<?> objectFactory) {     
  Map<String, Object> scope = (Map)this.threadScope.get();     
    return scope.computeIfAbsent(name, (k) -> {         
  return objectFactory.getObject();     
}); 
```

Standalone test case:

```java
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
import org.springframework.beans.factory.config.Scope;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.context.support.GenericApplicationContext;
import org.springframework.context.support.SimpleThreadScope;
import org.springframework.test.context.ContextConfiguration;
import org.springframework.test.context.TestContext;
import org.springframework.test.context.TestExecutionListeners;
import org.springframework.test.context.junit.jupiter.SpringExtension;
import org.springframework.test.context.support.AbstractTestExecutionListener;
import org.springframework.test.context.support.AnnotationConfigContextLoader;
import org.springframework.test.context.support.DependencyInjectionTestExecutionListener;

@ExtendWith(SpringExtension.class)
@ContextConfiguration(classes={MyTest.MyTestConfig.class},
        loader= AnnotationConfigContextLoader.class)
@TestExecutionListeners({MyTest.MyTestExecutionListener.class,
        DependencyInjectionTestExecutionListener.class})
public class MyTest {
    @Autowired
    private MyView removeNodeStatusScreen;

    @Test
    void justRun() {
        System.out.println("This test should run");
    }

[...]

PR #27234, opened 2021-07-30
Simplify getInternalBeanFactoryForBean

lyxell:
This pull request simplifies `getInternalBeanFactoryForBean` in `AbstractBeanFactoryBasedTargetSourceCreator` and makes its intent more obvious.

The problem in https://github.com/spring-projects/spring-framework/issues/25801#issuecomment-698346498 does not apply to this patch since the mapping function in this patch does not modify/access the map itself.

This patch was generated automatically by the static analysis tool [Logifix](https://github.com/lyxell/logifix) as part of a research project.


sbrannen:

Thanks for your first contribution to the Spring Framework.

PR #27238, opened 2021-08-04
Polish `PrintingResultHandlerIntegrationTests.printMvcResultsToWriterWithFailingGlobalResultMatcher()`

izeye:
This PR polishes `PrintingResultHandlerIntegrationTests.printMvcResultsToWriterWithFailingGlobalResultMatcher()` a bit by adding `Assertions.fail()` on an unexpected path.


snicoll:
Thanks again Johnny.

sbrannen:
Oops! I can't believe I checked in a _broken_ test. 😱 

Good catch, and thanks for fixing it, @izeye! 👍

PR #27239, opened 2021-08-04
Replace XMLReaderFactory with SAXParserFactory

Frederick888:
XMLReaderFactory has been marked as deprecated and without additional
configuration, and it's slower than SAXParserFactory.

Previously `XMLReaderFactory.createXMLReader()` is called upon every
request. This is an anti-pattern as mentioned in [1] and it can be very
slow since it loads the jar service file unless a parser has been
pre-assigned [2] (e.g. by setting org.xml.sax.driver).

SAXParserFactory uses a FactoryFinder [3] instead, which takes advantage
of a thread-local cache provided by ServiceLoader. Developers can still
pre-assign a factory by setting javax.xml.parsers.SAXParserFactory to
make it faster.

[1] https://bugs.openjdk.java.net/browse/JDK-6925410
[2] https://github.com/openjdk/jdk/blob/c8add223a10030e40ccef42e081fd0d8f00e0593/src/java.xml/share/classes/org/xml/sax/helpers/XMLReaderFactory.java#L144-L148
[3] https://github.com/openjdk/jdk/blob/66c653c561b3b5e904579af62e23ff94952bca05/src/java.xml/share/classes/javax/xml/parsers/SAXParserFactory.java#L181-L185

---

Related: #19055

Upstream discussion: https://mail.openjdk.java.net/pipermail/jdk-dev/2021-August/005853.html

### DIFF
--- a/spring-core/src/test/java/org/springframework/util/xml/AbstractStaxHandlerTests.java
+++ b/spring-core/src/test/java/org/springframework/util/xml/AbstractStaxHandlerTests.java
@@ -21,6 +21,8 @@ import java.io.StringWriter;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.SAXParser;
+import javax.xml.parsers.SAXParserFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.transform.Result;
 import javax.xml.transform.dom.DOMResult;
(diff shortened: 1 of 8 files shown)


Frederick888:
Sorry I should've replaced it outside spring-web as well.

snicoll:
@Frederick888 thanks for the PR and sorry it took so long to process it.

With `XMLReaderFactory` being deprecated since 9 already and `SAXParserFactory` being the documented replacement, this looks good to me. With perhaps additional places in the meantime to check. WDYT @jhoeller?

Issue #27266, opened 2021-08-12
ServerResponseResultHandler can NOT handle "Mono<ServerResponse>"


hongjunan:
**Affects:** \ 5.3.8

---

`org.springframework.web.reactive.function.server.support.ServerResponseResultHandler.java` can NOT handle `Mono<ServerResponse> or Flux<ServerResponse>`.

`public boolean supports(HandlerResult result)` always return false.
**because it check a returnValue type of HandlerResult .**
need to replace a return type with a generic type of it.
see below error code.
```
	@Override
	public boolean supports(HandlerResult result) {
		return (result.getReturnValue() instanceof ServerResponse);
	}
```

fix like below
it has to compare with a generic type of returnType
```
	@Override
	public boolean supports(HandlerResult result) {
		return ServerResponse.class.isAssignableFrom(result.getReturnType().getGeneric().toClass());
	}

```

hongjunan:
i make a PR

[https://github.com/spring-projects/spring-framework/pull/27268](url)

poutsma:
I don't really understand what you mean by handling `Flux<ServerResponse>`. Such a stream could contain multiple responses, with multiple response codes, headers, and bodies. We can only use one; what do we do with the rest?

And I am not so sure why supporting `Mono<ServerResponse>` is necessary. In other words: what is the use case here?

spring-projects-issues:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

hongjunan:
Currently, reactive handler  cannot handle webflux streaming.
Because handler are not checking whether the response value is reactive.
Please take a look at the source code I suggested above.
The current program is not working properly because of wrong condition checking.

snicoll:
@hongjunan I am also confused as what you're trying to do and, with a lack of concrete use case, we can't justify spending more time on this issue. If you can provide that information, we can obviously reconsider.

PR #27268, opened 2021-08-12
Bugfix/webflux Fix a support method logic of ServerResponseResultHandler .

hongjunan:
`org.springframework.web.reactive.function.server.support.ServerResponseResultHandler.java` can NOT handle `Mono<ServerResponse> or Flux<ServerResponse>`.

`public boolean supports(HandlerResult result)` always return false.
**because it check a returnValue type of HandlerResult .**
need to replace a return type with a generic type of it.
see below error code.
```
	@Override
	public boolean supports(HandlerResult result) {
		return (result.getReturnValue() instanceof ServerResponse);
	}
```

fix like below
it has to compare with a generic type of returnType
```
	@Override
	public boolean supports(HandlerResult result) {
		return ServerResponse.class.isAssignableFrom(result.getReturnType().getGeneric().toClass());
	}

```

Related: [https://github.com/spring-projects/spring-framework/issues/27266](url)

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/support/ServerResponseResultHandler.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/support/ServerResponseResultHandler.java
@@ -84,7 +84,7 @@ public class ServerResponseResultHandler implements HandlerResultHandler, Initia
 
 	@Override
 	public boolean supports(HandlerResult result) {
-		return result.getReturnType().getGeneric().toClass() == ServerResponse.class;
+		return ServerResponse.class.isAssignableFrom(result.getReturnType().getGeneric().toClass());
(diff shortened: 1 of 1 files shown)


hongjunan:
> should this use `.toClass().isAssignableFrom` or `.equals` rather than class instance identity? (could see situations where this may be problematic in environments that use OSGi, for example).

Totally agree with your comment.
i fixed it. thank you.

[https://github.com/spring-projects/spring-framework/pull/27268/commits/8e2c3f0d44b1613db4453ed599b6f173caddbd33](url)

rstoyanchev:
We don't need both an issue and a PR, and typically, the issue is closed, but in this case the conversation is under way on the issue, and the PR is trivial. It can always be re-opened.

Issue #19852, opened 2017-02-24
Support of MockRestServiceServer for WebClient [SPR-15286]


spring-issuemaster:
**[Stéphane Nicoll](https://jira.spring.io/secure/ViewProfile.jspa?name=snicoll)** opened **[SPR-15286](https://jira.spring.io/browse/SPR-15286?redirect=false)** and commented

Spring Boot has a `@RestClientTest` for pure client-side tests. We currently auto-configure `MockRestServiceServer` and bind any `RestTemplate` created by the `RestTemplateBuilder` to it.

This would be pretty awesome if we could port that feature to `WebClient` transparently if webflux is on the classpath. That would require that `MockRestServiceServer` can be configured with a `WebClient` in addition to a `RestTemplate`

---

**Sub-tasks:**
- #19295 Document how to test WebClient code with a mock server

15 votes, 24 watchers

membersound:
Using `WebClient` for client-side requests is pretty useless if there is no proper mock in spring for testing our code. Having rewritten my `RestTemplate` client code to `WebClient`, I'm disappointed there is no replacement for `MockRestServiceServer` yet...

dsundarraj2:
@membersound Have you used wiremock ?
WireMock can be used for mocking the responses for WebClient. Check this code repo which has examples for the same.
Repo:
https://github.com/code-with-dilip/learn-wiremock/tree/master/learn-wiremock-spring-cloud
TestCases:
https://github.com/code-with-dilip/learn-wiremock/blob/master/learn-wiremock-spring-cloud/src/test/java/com/learnwiremock/service/UserServiceWireMockRuleTest.java.
Cheers!

membersound:
While this might work, I'd prefer relying on some kind of "official" framework, eg something that's included and maintained by spring themselves. Otherwise, if WebClient evolves by time, and WireMock is left behind, that might led to rewrites for all of my projects. As we all know, time is money, so that's too risky...

rstoyanchev:
@membersound I appreciate the point but there is also a lot of effort involved to create, maintain, and evolve something over time and it's hard to justify given the existing alternatives. HTTP provides loose coupling so implementations can exist and evolve independently. In this case we're even talking about mocking a server and shaping its behavior. I should think that switching from one to another shouldn't be as big a deal as it might seem at first. That said MockWebServer has more stars than the Spring Framework does and I don't think it is going to be a problem with longevity.

SFBorland:
[...]

PR #27279, opened 2021-08-14
Implement a MockReactiveRestServiceServer for WebClient testing

ascopes:
I noticed that Spring does not currently provide the same tools for
testing WebClient-based reactive remote calls that the non-reactive
RestTemplate API has, which from experience recently has lead to
a little bit of frustration due to having to use more 
heavyweight/complicated non-Spring APIs to mock WebClient calls.

Since this felt relatively simple to implement, I went ahead and
attempted to draft a reactive counterpart to the existing non-reactive
MockRestServiceServer API.

This implementation is a shim that internally converts all requests
and responses to the non-reactive counterparts. This enables the
reactive API to use all the existing non-reactive request and
response verification components without having to duplicate the
entire API.

WebClientUtils in WebFlux has had to be updated to exclude Error
types from the filter, otherwise AssertionErrors always end up
being filtered out by the ExchangeFunctions conduit. I don't believe
this should be an issue though, as the JavaDoc for Error types
states that "no reasonable application should try to catch these
types", which means that the ExchangeFunctions probably shouldn't
have been allowing this through in the first place. Applied de-Morgans
law to reduce the complexity of the expression a little.

The API itself attempts to be as similar to MockRestServiceServer as
possible. For example:

```kotlin
val server = MockReactiveRestServiceServer.createServer()
val client = server.createWebClient()
val responseHeaders = HttpHeaders()

responseHeaders.add(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)

server
    .expect(requestTo("/actuator/health"))
    .andExpect(method(HttpMethod.GET))
    .andExpect(header(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON_VALUE))
    .andRespond(withSuccess()
        .headers(responseHeaders)
        .body("""{"status": "UP"}"""))

StepVerifier
    .create(client.get().uri("/actuator/health").retrieve().bodyToMono(String.class))
    .assertNext { assertThat(it).isEqualTo("""{"status": "UP"}""") }
    .verify()
```

Hope that contributions such as this are not considered out of place,
or anything. If this is in the wrong place, please let me know!

Likewise if there are any other suggestions or feedback, I'd be happy
to take those into account!

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/MockReactiveRestServiceServer.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/MockReactiveRestServiceServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-test/src/main/java/org/springframework/test/web/client/MockReactiveRestServiceServer.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/MockReactiveRestServiceServer.java
@@ -1,0 +1,322 @@
+/*
+ * Copyright 2002-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


snicoll:
Thanks for the PR but getting to a state where it is feature complete would require quite a bit more work and we've already declining such an option in #19852. For consistency, I am going to close this. Sorry it took so long to process it.

PR #27285, opened 2021-08-16
Introduce class-level execution phases for `@Sql`

### LINKED ISSUE #18929: Introduce class-level execution phases for `@Sql` [SPR-14357]
**[urbanq](https://jira.spring.io/secure/ViewProfile.jspa?name=urbanq)** opened **[SPR-14357](https://jira.spring.io/browse/SPR-14357?redirect=false)** and commented

It would be helpful to introduce class-level support in the `ExecutionPhase` enum used by `@Sql`: `BEFORE_CLASS` and `AFTER_CLASS`.

So when executing integration test we can easily run one single SQL script for all test methods in the test class.

---

**Affects:** 4.2.6

**Reference URL:** https://stackoverflow.com/questions/32871817/using-annotation-sql-is-it-possible-to-execute-scripts-in-class-level-before-m

**Issue Links:**
- #20570 Introduce opt-in support for merging class-level and method-level `@Sql` declarations

4 votes, 5 watchers

aahlenst:
This PR is a proposal to address #18929.

So far, scripts or inline statements declared with `@Sql` run before or after every test method. It is not possible to declare a script or inline statement to be run once before any test method has been run or after all test methods have been run.

This PR makes it possible by adding two additional execution phases, `BEFORE_TEST_CLASS` and `AFTER_TEST_CLASS`. Scripts or inline statements with `BEFORE_TEST_CLASS` will be run once before any test method has been started. Anything marked with `AFTER_TEST_CLASS` will be run after all test methods have been completed.

#1835 added `MergeMode`. `MergeMode` allows method-level `@Sql` annotations to override or extend class-level `@Sql` annotations. In my proposal, `MergeMode` has no effect on `@Sql` annotations with an execution phase of `BEFORE_TEST_CLASS` or `AFTER_TEST_CLASS`. The reason is that any method-level processing would come too late to influence a script or inline statement that runs when the test context is being set up. Furthermore, it would pose the question how to resolve conflicting annotations. If one test method wished to override a class-level `BEFORE_TEST_CLASS` annotation while others would not, honoring that wish would effectively turn `BEFORE_TEST_CLASS` into `BEFORE_TEST_METHOD`. `AFTER_TEST_CLASS` has roughly the same problem, just backwards.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/TestContext.java
+++ b/spring-test/src/main/java/org/springframework/test/context/TestContext.java
@@ -42,6 +42,7 @@ import org.springframework.test.annotation.DirtiesContext.HierarchyMode;
  * override {@link #setMethodInvoker(MethodInvoker)} and {@link #getMethodInvoker()}.
  *
  * @author Sam Brannen
+ * @author Andreas Ahlenstorf
  * @since 2.5
  * @see TestContextManager
  * @see TestExecutionListener
--- a/spring-test/src/main/java/org/springframework/test/context/TestContext.java
+++ b/spring-test/src/main/java/org/springframework/test/context/TestContext.java
@@ -111,6 +111,25 @@ public interface TestContext extends AttributeAccessor, Serializable {
 	Object getTestInstance();
 
 	/**
+	 * Tests whether a test method is part of this test context. Returns
+	 * {@code true} if this context has a current test method, {@code false}
+	 * otherwise.
+	 *
+	 * <p>The default implementation of this method always returns {@code false}.
+	 * Custom {@code TestContext} implementations are therefore highly encouraged
+	 * to override this method with a more meaningful implementation. Note that
+	 * the standard {@code TestContext} implementation in Spring overrides this
+	 * method appropriately.
+	 * @return {@code true} if the test execution has already entered a test
+	 * method
+	 * @since 6.1
+	 * @see #getTestMethod()
+	 */
+	default boolean hasTestMethod() {
+		return false;
+	}
+
+	/**
 	 * Get the current {@linkplain Method test method} for this test context.
 	 * <p>Note: this is a mutable property.
 	 * @return the current test method (never {@code null})
--- a/spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
+++ b/spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
@@ -33,6 +33,11 @@ import org.springframework.core.annotation.AliasFor;
  *
  * <p>Method-level declarations override class-level declarations by default,
  * but this behavior can be configured via {@link SqlMergeMode @SqlMergeMode}.
+ * However, this does not apply to class-level declarations that use
+ * {@link ExecutionPhase#BEFORE_TEST_CLASS} or
+ * {@link ExecutionPhase#AFTER_TEST_CLASS}. Such declarations are retained and
+ * scripts and statements are executed once per class in addition to any
+ * method-level annotations.
  *
  * <p>Script execution is performed by the {@link SqlScriptsTestExecutionListener},
  * which is enabled by default.
@@ -160,6 +165,18 @@ public @interface Sql {
 	 * Enumeration of <em>phases</em> that dictate when SQL scripts are executed.
 	 */
 	enum ExecutionPhase {
+
+		/**
+		 * The configured SQL scripts and statements will be executed
+		 * once <em>before</em> any test method is run.
+		 */
+		BEFORE_TEST_CLASS,
+
+		/**
+		 * The configured SQL scripts and statements will be executed
+		 * once <em>after</em> any test method is run.
+		 */
+		AFTER_TEST_CLASS,
 
 		/**
 		 * The configured SQL scripts and statements will be executed
--- a/spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
+++ b/spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
@@ -66,6 +66,7 @@ import org.springframework.core.annotation.AliasFor;
  * modules as well as their transitive dependencies to be present on the classpath.
  *
  * @author Sam Brannen
+ * @author Andreas Ahlenstorf
  * @since 4.1
  * @see SqlConfig
  * @see SqlMergeMode
--- a/spring-test/src/main/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListener.java
+++ b/spring-test/src/main/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListener.java
@@ -67,10 +67,17 @@ import static org.springframework.util.ResourceUtils.CLASSPATH_URL_PREFIX;
  * {@link Sql#scripts scripts} and inlined {@link Sql#statements statements}
  * configured via the {@link Sql @Sql} annotation.
  *
- * <p>Scripts and inlined statements will be executed {@linkplain #beforeTestMethod(TestContext) before}
- * or {@linkplain #afterTestMethod(TestContext) after} execution of the corresponding
- * {@linkplain java.lang.reflect.Method test method}, depending on the configured
- * value of the {@link Sql#executionPhase executionPhase} flag.
+ * <p>Class-level annotations that are constrained to a class-level execution
+ * phase ({@link ExecutionPhase#BEFORE_TEST_CLASS} or
+ * {@link ExecutionPhase#AFTER_TEST_CLASS}) will be run
+ * {@linkplain #beforeTestClass(TestContext) once before all test methods} or
+ * {@linkplain #afterTestMethod(TestContext) once after all test methods},
+ * respectively. All other scripts and inlined statements will be executed
+ * {@linkplain #beforeTestMethod(TestContext) before} or
+ * {@linkplain #afterTestMethod(TestContext) after} execution of the
+ * corresponding {@linkplain java.lang.reflect.Method test method}, depending
+ * on the configured value of the {@link Sql#executionPhase executionPhase}
+ * flag.
  *
  * <p>Scripts and inlined statements will be executed without a transaction,
  * within an existing Spring-managed transaction, or within an isolated transaction,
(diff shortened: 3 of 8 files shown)


spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
```diff
@@ -148,12 +153,23 @@
...
 	/**
 	 * Enumeration of <em>phases</em> that dictate when SQL scripts are executed.
 	 */
 	enum ExecutionPhase {
 
+		/**
+		 * The configured SQL scripts and statements will be executed
+		 * once <em>before</em> any test method is run.
+		 */
+		BEFORE_TEST_CLASS,
```
snicoll, line 165: It's a bit strange to have a method level annotation that can dictate something that applies to the whole class.

spring-test/src/main/java/org/springframework/test/context/TestContext.java
```diff
@@ -109,6 +109,26 @@ default void publishEvent(Function<TestContext, ? extends ApplicationEvent> even
...
+	 * to override this method with a more meaningful implementation. Note that
+	 * the standard {@code TestContext} implementation in Spring overrides this
+	 * method appropriately.
+	 *
+	 * @return {@code true} if the test execution has already entered a test
+	 * method
+	 * @since 5.3
+	 * @see #getTestMethod()
+	 */
+	default boolean hasTestMethod() {
```
snicoll, line 128: Can you expand why this is needed in the context of this change?

snicoll (commented):
Thanks for the PR. I've added a couple comments for your consideration.

spring-test/src/main/java/org/springframework/test/context/TestContext.java
aahlenst: Because of https://github.com/spring-projects/spring-framework/blob/735b29c101438be8a8f24ca9ba0ec426cb9f4c3b/spring-test/src/main/java/org/springframework/test/context/transaction/TestContextTransactionUtils.java#L256-L261

and 

https://github.com/spring-projects/spring-framework/blob/735b29c101438be8a8f24ca9ba0ec426cb9f4c3b/spring-test/src/main/java/org/springframework/test/context/support/DefaultTestContext.java#L170-L174

Previously, the test context was always a method. Therefore, calling `testContext.getTestMethod()` was no problem, because there is always a test method. But now processing can already happen before the test reaches a test method. In that case, the assertion `Assert.state(testMethod != null, "No test method");` fails.  Because I could not predict the consequences of removing the assertion, I added that method.

spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
aahlenst: I'm aware that it's not optimal. As mentioned in the PR description, there are more consistency issues. My question is: Is this acceptable? Or would it be better to introduce a new annotation like `@SqlPerClass` (might not be the best name)? The upside would be that the problems around `MergeMode` would be gone, too.

aahlenst: [DirtiesContext](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/annotation/DirtiesContext.html) has a similar design, but exposes two elements:`classMode` and `methodMode`. Still, nothing prevents me from annotating a method with `@DirtiesContext(classMode = BEFORE_CLASS)` which has no effect.

sbrannen: Before @aahlenst posted those last two comments, I was also going to point out that there is a precedent for this, namely `@DirtiesContext`.

I agree that it is a bit strange for it to be possible to specify a class-level setting at the method level that simply gets ignored; however, I am not in favor of introducing yet another SQL-related annotation for this purpose. We already have quite a few. And I don't think it necessarily warrants an additional enum and annotation attribute.

As it stands in the proposal, the addition of the two new enum constants in `ExecutionPhase` makes it impossible for a user to configure conflicting phases for a given `@Sql` declaration, and that's a good thing.

FWIW, I have never heard any complaints from users about the `ClassMode` and `MethodMode` dichotomy in `@DirtiesContext`, and the documentation clearly states that setting an inappropriate mode "has no meaning" (i.e., will be ignored).

In light of that, I am OK with the PR as-is.

sbrannen (changes requested):
I think the PR looks pretty good.

Please revert changes that deleted intentional blank lines, and please rebase on `main`.

Thanks!

sbrannen:
> Please revert changes that deleted intentional blank lines, and please rebase on `main`.

FYI: we would like to include this feature in the upcoming Spring Framework 6.1 RC1.

So if you don't have time to rebase on `main`, just let us know, and we'll take over the PR as-is.

spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
snicoll: Can we at lease throw an exception if you're using the class-level mode on a method?

aahlenst:
> FYI: we would like to include this feature in the upcoming Spring Framework 6.1 RC1.

That's awesome. Thanks a lot.

I have updated the PR. Due to its age, it required some rework. I also improved a test. If I should do anything else, like adding the exception as @snicoll suggested, let me know.

spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
sbrannen: Yes, we can definitely throw an exception for inappropriate user configuration.

`IllegalArgumentException` sounds reasonable to me.

sbrannen:
> I have updated the PR. 

Thanks!

> Due to its age, it required some rework. 

I can well imagine.

> I also improved a test. 

OK, I'll check it out in the coming week.

> If I should do anything else, like adding the exception as @snicoll suggested, let me know.

Yes, please throw an `IllegalArgumentException` -- can be achieved via Spring's `Assert` class as in other places in the TestContext framework.

spring-test/src/main/java/org/springframework/test/context/jdbc/Sql.java
aahlenst: We throw now.

Just thinking aloud: If `@Sql` has this, wouldn't it be nice if `@DirtiesContext` had it, too? Or would that be too disruptive years after the introduction?

sbrannen: > We throw now.

Thanks

> Just thinking aloud: If `@Sql` has this, wouldn't it be nice if `@DirtiesContext` had it, too? Or would that be too disruptive years after the introduction?

`@Sql` is different because it has a single enum-based `executionPhase` attribute, which means we can easily detect a user configuration error.

Whereas, `@DirtiesContext` has two mutually exclusive enum-based attributes: `classMode` and `methodMode`, and each of those has a `default` value. Thus, we cannot actually detect a user configuration error.

The only way to be able to detect a user configuration error for `@DirtiesContext` would be to introduce new `DEFAULT` enum constants in `ClassMode` and `MethodMode` and use those as the `default` attribute values going forward.

However, that would be a breaking change for any users who had inadvertently explicitly supplied the previous default values for those attributes.

Plus, it would make things a bit more cumbersome.

In light of the above, I don't think we should make any changes to `@DirtiesContext` in this regard this late in the game.

spring-test/src/main/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListener.java
```diff
@@ -366,7 +413,11 @@ private String[] getScripts(Sql sql, Class<?> testClass, Method testMethod, bool
 	 * Detect a default SQL script by implementing the algorithm defined in
 	 * {@link Sql#scripts}.
 	 */
-	private String detectDefaultScript(Class<?> testClass, Method testMethod, boolean classLevel) {
+	private String detectDefaultScript(Class<?> testClass, @Nullable Method testMethod, boolean classLevel) {
+		if (!classLevel && testMethod == null) {
+			throw new AssertionError("Method-level @Sql requires a testMethod");
+		}
```
sbrannen, line 419: We don't throw `AssertionError` in such scenarios.

Please use `Assert.state(...)` instead.

spring-test/src/main/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListener.java
```diff
@@ -407,4 +458,9 @@ private void registerClasspathResources(String[] paths, RuntimeHints runtimeHint
 				.forEach(runtimeHints.resources()::registerResource);
 	}
 
+	private static boolean isValidMethodLevelPhase(ExecutionPhase executionPhase) {
+		// Class-level phases cannot be used on methods.
+		return executionPhase != ExecutionPhase.BEFORE_TEST_CLASS &&
+				executionPhase != ExecutionPhase.AFTER_TEST_CLASS;
```
sbrannen, line 464: Please check the known acceptable method-level phases instead.

spring-test/src/test/java/org/springframework/test/context/jdbc/AfterTestClassSqlScriptsTests.java
```diff
@@ -0,0 +1,87 @@
...
+
+		@Override
+		public void afterTestClass(TestContext testContext) throws Exception {
+			DataSource dataSource = TestContextTransactionUtils.retrieveDataSource(testContext, null);
+			try {
+				new JdbcTemplate(dataSource).queryForList("SELECT name FROM user", String.class);
+				throw new AssertionError("BadSqlGrammarException should have been thrown.");
+			}
+			catch (BadSqlGrammarException expected) {
+			}
```
sbrannen, line 77: Please replace with an AssertJ assertion along the lines of `assertThatExceptionOfType(BadSqlGrammarException.class).isThrownBy(...)`.

sbrannen (changes requested):
I think this PR is almost ready, though I've requested a few additional minor changes.

aahlenst:
@sbrannen I applied the requested changes.

sbrannen (approved):

sbrannen:
This has been merged into `main` in 5aa2d0515ed89ba0c8945eba7eb5be07afb83133 and revised in 0afcb4dfed4044b202b9efd55767fbd1939e6d5d.

It turns out that `hasTestMethod()` wasn't needed after all. See the commit message for the second commit for details on what I changed.

Thanks for the contribution! 👍

PR #27303, opened 2021-08-20
Fix UrlPathHelper#shouldRemoveSemicolonContent()

### LINKED ISSUE #27256: UrlPathHelper: checkReadOnly() called in read method shouldRemoveSemicolonContent()
**Affects:** 5.3.8

---

Method `checkReadOnly` is called in read method UrlPathHelper#shouldRemoveSemicolonContent, line 149. If we want to know whether we should `removeSemicolonContent` when `this.readOnly` is true, an error will be thrown.
```
public boolean shouldRemoveSemicolonContent() {
    checkReadOnly();
    return this.removeSemicolonContent;
}
```

evpaassen:
Fixes #27256

This PR fixes a bug that renders read-only `UrlPathHelper`s unusable, because `shouldRemoveSemicolonContent()` throws an exception. For example, when using `UrlPathHelper#rawPathInstance`, which is read-only.

The `checkReadOnly()` method should only be called from methods that modify properties to prevent modification of read-only instances.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UrlPathHelper.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UrlPathHelper.java
@@ -147,7 +147,6 @@ public class UrlPathHelper {
 	 * Whether configured to remove ";" (semicolon) content from the request URI.
 	 */
 	public boolean shouldRemoveSemicolonContent() {
-		checkReadOnly();
 		return this.removeSemicolonContent;
 	}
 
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @evpaassen,

Can you please introduce a unit test that fails prior to the change and passes after the change?

evpaassen:
Hi @sbrannen,

I added a test now. I'm not sure it really has added value, because nothing regarding the `readOnly` property in this class was tested before, and the new test is effectively testing a getter method. But on the other hand, it will prevent a regression, I guess.

Please let me know if you're happy with it, or that I should make any changes.

sbrannen:
> I added a test now. I'm not sure it really has added value, because nothing regarding the `readOnly` property in this class was tested before, and the new test is effectively testing a getter method. But on the other hand, it will prevent a regression, I guess.

Yes, my intent was to make sure the problem is fixed and does not return (i.e., to serve as a regression test).

> Please let me know if you're happy with it, or that I should make any changes.

That looks good now.

Thanks

evpaassen:
You're welcome. And thanks for reviewing and merging!

sbrannen:
Note that this was also backported to `5.2.17` in 10a90d32044014386aee5c9923c38320fc9c2d90.

PR #27308, opened 2021-08-21
Propagate `CoroutineContext` in reactive transaction

### LINKED ISSUE #27307: Reactive Transaction when applied to Kotlin Coroutines doesn't pass the `CoroutineContext`
**Affects:** 5.3.8

---

#### Problem:
* Reactive transaction when applied to coroutines doesn't pass the caller's `CoroutineContext`.
```kotlin
// in whatever caller.. start coroutine with custom context.
val user = User(role = "admin") // custom coroutine context
GlobalScope.launch(user) {
    val hasUser1 = transaction1() // return false
    val hasUser2 = transaction2() // return false, too.
}

@Transactional
suspend fun transaction1(): Boolean {
    return coroutineContext[User] != null
}

suspend fun transaction2(): Boolean = transactionalOperator.executeAndAwait {
    coroutineContext[User] != null
} 
```

#### Cause:
* [CoroutinesUtils#invokeSuspendingFunction](https://github.com/spring-projects/spring-framework/blob/v5.3.9/spring-core/kotlin-coroutines/src/main/kotlin/org/springframework/core/CoroutinesUtils.kt#L62) and [TransactionalOperatorExtension#executeAndAwait](https://github.com/spring-projects/spring-framework/blob/v5.3.9/spring-tx/src/main/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensions.kt) overwrites caller's coroutine context while converting coroutines into a `Publisher` by doing...:
```kotlin
mono(Dispatchers.Unconfined) { ... }
```

#### Fix:
* Passing the `CoroutineContext` of the caller may fix the problem.
```kotlin
// in `TransactionalOperatorExtension`
suspend fun <T : Any?> TransactionalOperator.executeAndAwait(f: suspend (ReactiveTransaction) -> T): T = {
    val ctx = Dispatchers.Unconfined + coroutineContext.minusKey(Job) // combine the caller's context
    return execute { status -> mono(ctx) { f(status) } }.map { value -> Optional.of(value!!) }
        .defaultIfEmpty(Optional.empty()).awaitLast().orElse(null)
}

// in `CorotuinesUtils`
fun invokeSuspendingFunction(method: Method, target: Any, vararg args: Any?): Publisher<*> {
    val function = method.kotlinFunction!!
    val ctx = Dispatchers.Unconfined + ((args[args.size - 1] as Continuation<*>?)?.context?.minusKey(Job) ?: EmptyCoroutineContext) // combine the caller's context
    val mono = mono(ctx) {
        function.callSuspend(target, *args.sliceArray(0..(args.size-2))).let { if (it == Unit) null else it }
    }.onErrorMap(InvocationTargetException::class.java) { it.targetException }
    return if (function.returnType.classifier == Flow::class) {
        mono.flatMapMany { (it as Flow<Any>).asFlux() }
    }
    else {
        mono
    }
}
```

ks-yim:
#### Motivation:
* Fixes #27307 

#### Modifications:
* Propagate caller's `CoroutineContext` when converting coroutines into `Mono`.
  - In `CoroutinesUtils#invokeSuspendingFunction`
  - In `TransactionalOperatorExtensions`
    + `Flow<T>.transactional(...)`
    + `TransactionalOperator.executeAndAwait(...)`
* Change `TransactionalOperator.executeAndAwait(...)`'s type parameter upper bound from `Any` to `Any?`
  - For better usability.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
@@ -20,7 +20,11 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Objects;
 
+import org.reactivestreams.Publisher;
+
 import kotlin.Unit;
+import kotlin.coroutines.Continuation;
+import kotlin.coroutines.CoroutineContext;
 import kotlin.jvm.JvmClassMappingKt;
 import kotlin.reflect.KClassifier;
 import kotlin.reflect.KFunction;
@@ -31,10 +35,10 @@ import kotlinx.coroutines.CoroutineStart;
 import kotlinx.coroutines.Deferred;
 import kotlinx.coroutines.Dispatchers;
 import kotlinx.coroutines.GlobalScope;
+import kotlinx.coroutines.Job;
 import kotlinx.coroutines.flow.Flow;
 import kotlinx.coroutines.reactor.MonoKt;
 import kotlinx.coroutines.reactor.ReactorFlowKt;
-import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
@@ -71,7 +75,9 @@ public abstract class CoroutinesUtils {
 	public static Publisher<?> invokeSuspendingFunction(Method method, Object target, Object... args) {
 		KFunction<?> function = Objects.requireNonNull(ReflectJvmMapping.getKotlinFunction(method));
 		KClassifier classifier = function.getReturnType().getClassifier();
-		Mono<Object> mono = MonoKt.mono(Dispatchers.getUnconfined(), (scope, continuation) ->
+		Continuation<?> cont = (Continuation<?>) args[args.length - 1];
+		CoroutineContext ctx = Dispatchers.getUnconfined().plus(cont.getContext().minusKey(Job.Key));
+		Mono<Object> mono = MonoKt.mono(ctx, (scope, continuation) ->
 					KCallables.callSuspend(function, getSuspendedFunctionArgs(target, args), continuation))
 				.filter(result -> !Objects.equals(result, Unit.INSTANCE))
 				.onErrorMap(InvocationTargetException.class, InvocationTargetException::getTargetException);
--- a/spring-tx/src/main/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensions.kt
+++ b/spring-tx/src/main/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensions.kt
@@ -1,13 +1,33 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.springframework.transaction.reactive
 
-import java.util.Optional
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.reactive.asFlow
 import kotlinx.coroutines.reactive.awaitLast
 import kotlinx.coroutines.reactor.asFlux
 import kotlinx.coroutines.reactor.mono
 import org.springframework.transaction.ReactiveTransaction
+import java.util.Optional
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.coroutineContext
 
 /**
  * Coroutines variant of [TransactionalOperator.transactional] as a [Flow] extension.
@@ -15,17 +35,22 @@ import org.springframework.transaction.ReactiveTransaction
  * @author Sebastien Deleuze
  * @since 5.2
  */
-fun <T : Any> Flow<T>.transactional(operator: TransactionalOperator): Flow<T> =
-		operator.transactional(asFlux()).asFlow()
+fun <T : Any> Flow<T>.transactional(
+	operator: TransactionalOperator,
+	context: CoroutineContext = EmptyCoroutineContext
+): Flow<T> =
+	operator.transactional(asFlux(context)).asFlow()
 
 /**
-* Coroutines variant of [TransactionalOperator.execute] with a suspending lambda
-* parameter.
-*
-* @author Sebastien Deleuze
-* @author Mark Paluch
-* @since 5.2
-*/
-suspend fun <T : Any> TransactionalOperator.executeAndAwait(f: suspend (ReactiveTransaction) -> T?): T? =
-		execute { status -> mono(Dispatchers.Unconfined) { f(status) } }.map { value -> Optional.of(value) }
-				.defaultIfEmpty(Optional.empty()).awaitLast().orElse(null)
+ * Coroutines variant of [TransactionalOperator.execute] with a suspending lambda
+ * parameter.
+ *
+ * @author Sebastien Deleuze
+ * @author Mark Paluch
+ * @since 5.2
+ */
+suspend fun <T> TransactionalOperator.executeAndAwait(f: suspend (ReactiveTransaction) -> T): T {
+	val ctx = Dispatchers.Unconfined + coroutineContext.minusKey(Job)
+	return execute { status -> mono(ctx) { f(status) } }.map { value -> Optional.of(value!!) }
+		.defaultIfEmpty(Optional.empty()).awaitLast().orElse(null)
+}
--- a/spring-tx/src/test/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensionsTests.kt
+++ b/spring-tx/src/test/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensionsTests.kt
@@ -16,6 +16,7 @@
 
 package org.springframework.transaction.reactive
 
+import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.toList
@@ -24,6 +25,8 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.fail
 import org.springframework.transaction.support.DefaultTransactionDefinition
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
 
 class TransactionalOperatorExtensionsTests {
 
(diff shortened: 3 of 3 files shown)


sdeleuze (commented):
I will have a deeper look to the implementation, but looks like we have something to fix here indeed.

spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
```diff
@@ -71,7 +75,9 @@
 	public static Publisher<?> invokeSuspendingFunction(Method method, Object target, Object... args) {
 		KFunction<?> function = Objects.requireNonNull(ReflectJvmMapping.getKotlinFunction(method));
 		KClassifier classifier = function.getReturnType().getClassifier();
-		Mono<Object> mono = MonoKt.mono(Dispatchers.getUnconfined(), (scope, continuation) ->
+		Continuation<?> cont = (Continuation<?>) args[args.length - 1];
```
sdeleuze, line 78: I am not super confortable doing this kind of manual handling of the continuation, so I have asked guidance to Kotlin team to see if there is a better way, waiting there feedback.

spring-tx/src/main/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensions.kt
```diff
@@ -1,31 +1,56 @@
...
-		execute { status -> mono(Dispatchers.Unconfined) { f(status) } }.map { value -> Optional.of(value) }
-				.defaultIfEmpty(Optional.empty()).awaitLast().orElse(null)
+ * Coroutines variant of [TransactionalOperator.execute] with a suspending lambda
+ * parameter.
+ *
+ * @author Sebastien Deleuze
+ * @author Mark Paluch
+ * @since 5.2
+ */
+suspend fun <T> TransactionalOperator.executeAndAwait(f: suspend (ReactiveTransaction) -> T): T {
```
sdeleuze, line 52: I am not sure about the principle of this change of null-safety, and in any case it should not be part of this PR.

@mp911de Any take on the change of null-safety requested here?

mp911de: A reactive transaction can complete without emitting a value, so we should keep `T` nullable.

spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
sdeleuze: Feedback from Kotlin team:

> What puzzles me is why there is `Job` in the context.
> When you have one, it means that the lifecycle of the result is just incorrect — it is controlled solely by mono, but not by the Job as the caller might’ve expected.
> It’s hard to tell if it is exactly wrong without knowing why Job is there in the first place

Potentially related to spring-projects/spring-data-commons#2532 cc @mp911de.

sdeleuze:
Moving this issue back to waiting for triage while we are waiting more information on https://github.com/spring-projects/spring-data-commons/issues/2532 side.

spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
```diff
@@ -71,7 +75,9 @@ public static <T> Deferred<T> monoToDeferred(Mono<T> source) {
...
```
1004789224, line 78: This might casuse an NPE, and this code is validated
```java
public static Publisher<?> invokeSuspendingFunction(Method method, Object target, Object... args) {
		KFunction<?> function = Objects.requireNonNull(ReflectJvmMapping.getKotlinFunction(method));
		if (method.isAccessible() && !KCallablesJvm.isAccessible(function)) {
			KCallablesJvm.setAccessible(function, true);
		}
		KClassifier classifier = function.getReturnType().getClassifier();
		Continuation<?> originContinuation = (Continuation<?>) args[args.length - 1];
		CoroutineContext context;
		if (originContinuation == null) {
			context = (CoroutineContext) Dispatchers.getUnconfined();
		} else {
			context=originContinuation.getContext().minusKey(Job.Key);
		}
		Mono<Object> mono = MonoKt.mono(context, (scope, continuation) ->
					KCallables.callSuspend(function, getSuspendedFunctionArgs(target, args), continuation))
				.filter(result -> !Objects.equals(result, Unit.INSTANCE))
				.onErrorMap(InvocationTargetException.class, InvocationTargetException::getTargetException);
		if (classifier != null && classifier.equals(JvmClassMappingKt.getKotlinClass(Flow.class))) {
			return mono.flatMapMany(CoroutinesUtils::asFlux);
		}
		return mono;
	}
```

spring-tx/src/main/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensions.kt
jamesbassett: I ran into this as well. I was expecting the signature to be

 `suspend fun <T> TransactionalOperator.executeAndAwait(f: suspend (ReactiveTransaction) -> T): T {`

 i.e. for `T` to not be nullable and to not have the upper bound on `Any`.

This would make all of the following possible (and for the first scenario to not require a `!!`).

```kotlin
    val test1: Int = transactionalOperator.executeAndAwait { 1 }
    val test2: Int? = transactionalOperator.executeAndAwait { "1".toIntOrNull() }
    val test3: Int? = transactionalOperator.executeAndAwait { null }
```

Am I missing something?

ks-yim: @mp911de Sorry that I didn't give a feedback for this comment for so long.
I ran into the same issue that @jamesbassett has just pointed out.

Any special reason to return `T?` instead of just `T`?
I got to always cast the returned value with `!!` even if the lambda is guaranteed to return a non-null value.

spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
ks-yim: Thanks for pointing this out 👍.

psimonsked (approved):

jamesbassett:
Are there any updates on this issue? It's a bit of a roadblock for us - we rely on the context being propagated so can't use transactions at the moment :(

sdeleuze:
@jamesbassett Any chance you could provide a repro since we have no feedback on that on the Spring Data issue?

JoeMaher:
@sdeleuze I have created a reproducible example here https://github.com/JoeMaher/tx-coroutine-context-loss-example, please let me know if you have any queries/concerns

sdeleuze:
Great, thanks, I will have a look shortly.

jamesbassett:
Hi @sdeleuze did you get a chance to look at @JoeMaher 's reproducer? 🤞

sdeleuze:
Sorry for the delay but I am currently focused on the latest bits of native support for Spring Framework 6 RC1.

jamesbassett:
Hi @sdeleuze just bumping this (now that Spring 6 is released!)

spring-tx/src/main/kotlin/org/springframework/transaction/reactive/TransactionalOperatorExtensions.kt
sdeleuze: Fixed by #29919

spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
sdeleuze: It is indeed the same for WebFlux. I will provide your feedback to the Kotlin team, may be indeed ok for this specific use case.

spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
sdeleuze: I don't think the refined versioned shared below is impacted since we are always in a case where there is a `Continuation` parameter, but feel free to double check.

sdeleuze:
Thanks for your patience all.

I re-worked the PR after fixing the nullability issue via #29919 with [this commit](https://github.com/sdeleuze/spring-framework/commit/gh-27308), I now need to have the confirmation from the Kotlin team this is ok to remove `Job.Key` from the Coroutines context for this specific use case.

If validated, we may backport the fix to `5.3.x`.

spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
sdeleuze: @qwwdfsad Could you please check @ks-yim previous comment, especially this quote below, to help us analyzing if removing `Job.Key` from the `CoroutineContext` is ok for this very specific use case where we intercept the suspending invocation to plug our Reactive transaction management and later invoke `AwaitKt.awaitSingleOrNull` with the original continuation:
> I understand the concerns of Kotlin team worrying that the coroutine's lifecycle is no longer be managed by the caller's Job context, but I thought it's okay to do this for this specific use case since the lifecycle control of the returned Mono goes back into the caller's Job context anyways in the subsequent call for awaitSingleOrNull in [this line](https://github.com/spring-projects/spring-framework/blob/main/spring-tx/src/main/java/org/springframework/transaction/interceptor/TransactionAspectSupport.java#L370-L372).

FWIW our tests are green with this implementation proposed in [my refined commit](https://github.com/sdeleuze/spring-framework/commit/gh-27308).

dkhalanskyjb: (Another guy from the coroutines project here) You are doing very fancy things here! It did take me a while to understand them. If the `CompletableFuture` gets cancelled, it will cancel `awaitSingleOrNull`; if `awaitSingleOrNull` gets cancelled, it will cancel the `mono`; if `mono` completes with an error or a value, it will return that via `awaitSingleOrNull`. So, in effect, you're getting the same job control with extra steps.

The only issue I see is coroutine code that explicitly does something with the `Job`, like `currentCoroutineContext()[Job]!!.invokeOnCompletion { println("ok") }`. If you intercept such code in a `mono`, then, naturally, the `Job` used will be the one created by `mono`, not the `Job` from `future`. I don't see a way around this, though.

sdeleuze:
@mp911de @poutsma Could you please have a look to [this commit](https://github.com/sdeleuze/spring-framework/commit/gh-27308) and [this comment from Kotlin team](https://github.com/spring-projects/spring-framework/pull/27308#discussion_r1095746663) to check if you see something against merging this on `main` and `5.3.x`.

The limitation on `Job` is coming from Coroutines implementation, looks like a pretty unusual use case, and given the fact that this popular use case is broken for months, merging this will a a significant improvement for a lot of users.

poutsma:
> @mp911de @poutsma Could you please have a look to [this commit](https://github.com/sdeleuze/spring-framework/commit/gh-27308) and [this comment from Kotlin team](https://github.com/spring-projects/spring-framework/pull/27308#discussion_r1095746663) to check if you see something against merging this on `main` and `5.3.x`.

I did look, and can follow along with the discussion, but this is too far out of my area of expertise to give any insightful comment. If the PR resolves #27307, and does not break any existing code, than it seems good to me.

sdeleuze:
The fix and tests for both functional and annotation variants are now merged in `main` and will be available as of Spring Framework `6.0.5` and Spring Boot `3.0.3`.

I evaluated if it is possible to merge it on `5.3.x` but this is not straightforward as it would require to backport 2 other changes that are impacting the public APIs, so I chose at least for now to not backport it.

jamesbassett:
That's fantastic news @sdeleuze thank you so much for progressing this 🥳 I'll give it a test drive when Boot 3.0.3 drops (I imagine that's later this month?)

nkonev:
@jamesbassett there is the calendar with Spring Ecosystem releases https://calendar.spring.io/

jamesbassett:
Just dropping in to say thanks @sdeleuze I tested out Boot 3.0.3 today and can confirm that everything now works as expected (we're not losing context when using transactions) and `executeAndAwait()` doesn't require the `!!` (hold-my-beer operator) any more 🥳 Thanks for all your help ❤️

PR #27309, opened 2021-08-21
Support `TimeUnit` in the `@Scheduled` annotation

vbrown37:
Added an option to use `java.util.concurrent.TimeUnit` with `fixedDelay`, `fixedRate` and `initialDelay` in the `@Scheduled` annotation. Using by default milliseconds as before.

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/annotation/Scheduled.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/annotation/Scheduled.java
@@ -16,13 +16,13 @@
 
 package org.springframework.scheduling.annotation;
 
-import java.util.concurrent.TimeUnit;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.concurrent.TimeUnit;
 
 import org.springframework.scheduling.config.ScheduledTaskRegistrar;
 
@@ -120,12 +120,6 @@ public @interface Scheduled {
 	String fixedDelayString() default "";
 
 	/**
-	 * Specify the {@link TimeUnit} to use for the fixedDelay and the fixedDelayString values.
-	 * @return the {@link TimeUnit}, by default milliseconds will be used.
-	 */
-	TimeUnit fixedDelayTimeUnit() default TimeUnit.MILLISECONDS;
-
-	/**
 	 * Execute the annotated method with a fixed period between
 	 * invocations.
 	 * Using milliseconds by default with fixedRateTimeUnit().
@@ -142,12 +136,6 @@ public @interface Scheduled {
 	 * @since 3.2.2
 	 */
 	String fixedRateString() default "";
-
-	/**
-	 * Specify the {@link TimeUnit} to use for the fixedRate and the fixedRateString values.
-	 * @return the {@link TimeUnit}, by default milliseconds will be used.
-	 */
-	TimeUnit fixedRateTimeUnit() default TimeUnit.MILLISECONDS;
 
 	/**
 	 * Number to delay before the first execution of a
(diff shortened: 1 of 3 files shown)


spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
```diff
@@ -398,7 +397,7 @@ protected void processScheduled(Scheduled scheduled, Method method, Object bean)
 			Set<ScheduledTask> tasks = new LinkedHashSet<>(4);
 
 			// Determine initial delay
-			long initialDelay = scheduled.initialDelay();
+			long initialDelay = scheduled.initialDelayTimeUnit().toMillis(1) * scheduled.initialDelay();
```
sbrannen, line 400: Wouldn't it be more idiomatic to use `TimeUnit.MILLISECONDS.convert(scheduled.initialDelay(), scheduled.initialDelayTimeUnit())`?

Axzial: Seems better, fixed !

spring-context/src/main/java/org/springframework/scheduling/annotation/Scheduled.java
```diff
@@ -16,6 +16,7 @@
 
 package org.springframework.scheduling.annotation;
 
+import java.util.concurrent.TimeUnit;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Repeatable;
```
sbrannen, line 22:
```suggestion
import java.lang.annotation.Documented;
import java.lang.annotation.ElementType;
import java.lang.annotation.Repeatable;
import java.util.concurrent.TimeUnit;
```

This [breaks the build](https://ci.spring.io/builds/163051) due to import order.

Please be sure to run `./gradlew check` before pushing commits to a PR.

spring-context/src/main/java/org/springframework/scheduling/annotation/Scheduled.java
```diff
@@ -101,52 +102,76 @@
...
 	 * @since 3.2.2
 	 */
 	String fixedDelayString() default "";
 
 	/**
-	 * Execute the annotated method with a fixed period in milliseconds between
+	 * Specify the {@link TimeUnit} to use for the fixedDelay and the fixedDelayString values.
+	 * @return the {@link TimeUnit}, by default milliseconds will be used.
+	 */
+	TimeUnit fixedDelayTimeUnit() default TimeUnit.MILLISECONDS;
```
sbrannen, line 126: Please introduce a single `TimeUnit timeUnit() default TimeUnit.MILLISECONDS;` attribute and get rid of the other `TimeUnit` attributes.

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
```diff
@@ -75,7 +76,6 @@
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.util.StringValueResolver;
-
```
sbrannen, line 78: Please reinstate this line.

sbrannen (changes requested):
I've requested some minor changes.

Please also introduce some tests that use different `TimeUnit` values.

If you can get these changes in soon, this can be included in Spring Framework 5.3.10.

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
Axzial: Done.

spring-context/src/main/java/org/springframework/scheduling/annotation/Scheduled.java
Axzial: Done.

spring-context/src/main/java/org/springframework/scheduling/annotation/Scheduled.java
Axzial: Should be all good !

sbrannen (approved):
Thanks for making the requested changes!

I'll make a final review of this PR in time for 5.3.10.

PR #27330, opened 2021-08-28
chore: add map params variants to update method

hantsy:


hantsy:
When I used Spring Jdbc in my latest projects, I found the `update(String sql, SqlParameterSource paramSource, KeyHolder generatedKeyHolder, String[] keyColumnNames)` lacks a variant to accept a `Map` instead of the `SqlParameterSource`.

leeseojune53:
Wow nice feature!!

snicoll:
Thanks for the PR @hantsy but when more arguments are required, we fallback on `SqlParameterSource`. You can easily get one from the map by calling `new MapSqlParameterSource(paramMap)`. Given how easy those methods can be called, I don't think the extra signatures are warranted.

PR #27334, opened 2021-08-30
simplify if statement

SeaItFover:
simplify if statement in ResolvableType class


rahimkhanabdul (approved):

rstoyanchev:
Thanks for the suggestion. The changes are possible, as suggested also by the IDE, but not necessarily the best outcome for maintenance.

PR #27335, opened 2021-08-30
Optimize for single TransactionSynchronization

marschall:
Optimize TransactionSynchronizationManager.getSynchronizations() to
avoid sorting and ArrayList creation in case of a single
TransactionSynchronization.


marschall:
Spring Batch registers a single `TransactionSynchronization` in `TaskletStep#doExecute(StepExecution)`. In certain extreme cases the `ArrayList` and `Object[]` creation can become noticeable. This PR optimizes the case of a single `TransactionSynchronization`, avoids the sorting and instead calls `Collections#singletonList(Object)`.

Before:

TransactionSynchronizationManager_getSynchronizations


After:

marschall:
Thank you for merging

PR #27337, opened 2021-08-30
Fix Kotlin example for filtering handler functions

w3-3w:
The Kotlin code example at _Filtering Handler Functions_ seems to be not equivalent to Java.

Suppose

> The `before` filter that adds a custom request header is only applied to the two GET routes.
The `after` filter that logs the response is applied to all routes, including the nested ones.

is correct, `POST("/person", handler::createPerson)` and `after()` should not be inside `nest`, if my understanding to RouterFunction is correct.


sbrannen:

Thanks for submitting your first PR to the Spring Framework! 👍

PR #27349, opened 2021-09-01
Merge AopTestUtils into AopUtils

### LINKED ISSUE #27348: Merge AopTestUtils into AopUtils
Methods in `AopTestUtils` is not only useful for tests.

quaff:
Fix gh-27348

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/support/AopUtils.java
+++ b/spring-aop/src/main/java/org/springframework/aop/support/AopUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


snicoll:
Duplicate #27349

spring-aop/src/main/java/org/springframework/aop/support/AopUtils.java
```diff
@@ -116,6 +118,70 @@ public static boolean isCglibProxy(@Nullable Object object) {
...
+	 * @see #getUltimateTargetObject
+	 * @since 5.3.10
+	 */
+	@SuppressWarnings("unchecked")
+	public static <T> T getTargetObject(Object candidate) {
+		Assert.notNull(candidate, "Candidate must not be null");
+		try {
+			if (AopUtils.isAopProxy(candidate) && candidate instanceof Advised) {
+				Object target = ((Advised) candidate).getTargetSource().getTarget();
+				if (target != null) {
```
Roli24, line 141: Do we need to check instance of condition with a null condition?

jhoeller:
We intentionally separated those utility methods. I'm afraid those AOP test methods are potentially unsafe since they never release the target instance that they take from the `TargetSource` there. That might not matter for an eagerly initialized singleton target in a test but generally speaking such a step is not side effect free. Overall, these methods do not represent idiomatic usage of the AOP framework and are therefore not to be promoted as proper utilities in `AopUtils` itself.

That said, if such a step works for you for a certain purpose, feel free to mimic/copy it into wherever you need it. Just be aware of those `TargetSource` side effects. Internal discovery of a target instance for a certain purpose is easier to justify than doing so in a general public utility, after all.

Issue #26266, opened 2020-12-14
Memory leak with CglibAopProxy$ProxyCallbackFilter


Pancor:
It seems that there is a memory leak with Spring Boot 2.4.1 on my Debian 10.5 with Java 11.0.1 when using `@PreAuthorize("hasRole('ROLE')")` annotation on controller's endpoint. As you can see in the first image, there is constant grow in old generation memory of G1 garbage collector when running tests. The grow stops at max heap size which is 512 MB. I prepared a [sample project](https://github.com/Pancor/leak) where you can reproduce the problem. 

<details>
  <summary>Memory usage 1</summary>
  
![image](https://user-images.githubusercontent.com/24636767/102074708-75f9dd00-3e05-11eb-97ec-927ef3d27a80.png)
  
</details>
<details>
  <summary>Memory usage 2</summary>
  
![image](https://user-images.githubusercontent.com/24636767/102074755-88741680-3e05-11eb-81e6-8d0620f987d1.png)
  
</details>

After importing the project just run this command:

`./gradlew clean test`

Tests will fail because they assume that HTTP response will be 200, but it is 401, but it shouldn't matter.
If you remove `@PreAuthorize` from [DummyController](https://github.com/Pancor/leak/blob/master/src/main/java/pl/pancordev/leak/controllers/DummyController.java) or `prePostEnabled = true` from [SecurityConfig](https://github.com/Pancor/leak/blob/master/src/main/java/pl/pancordev/leak/SecurityConfig.java) then memory usage seems to be ok.
<details>
  <summary>Memory usage 3</summary>
  
![image](https://user-images.githubusercontent.com/24636767/102076516-18b35b00-3e08-11eb-9d64-c4717c6737b0.png)
  
</details>
<details>
  <summary>Memory usage 4</summary>
  
![image](https://user-images.githubusercontent.com/24636767/102076541-236df000-3e08-11eb-98ab-ce1bb912493e.png)
  
</details>

The project uses dummy controller with one endpoint which has `@PreAuthorize("hasRole('ROLE')")` annotation. It has also basic `SecurityConfig` to enable `@PreAuthorize` annotation. Controller uses some dummy services to make sure that for each test Spring will recreate the context. I also set `spring.test.context.cache.maxSize=1` in [build.gradle](https://github.com/Pancor/leak/blob/master/build.gradle) to make sure that the grow of memory does not come from cache.

If you have any more questions about my configuration or anything else, let me know.

[...]

PR #27375, opened 2021-09-09
Fix memory leak on AOP Proxy class definition cache

yokotaso:
Fix memory leak on AOP Proxy class definition cache.

## How to reproduce 

1. Like `@Validated` annotation with Request scope Controller like bellow

```java
// Annnotation definition
@Target(ElementType.TYPE)
@Documented
@Controller
@Scope(WebApplicationContext.SCOPE_REQUEST)
@RequestMapping(method = { RequestMethod.GET, RequestMethod.HEAD })
@Transactional
@Validated
public @interface MyController {
}

@MyController
public class RootController {
    @ModelAttribute("requireJs")
    public boolean isRequireJs() {
        return true;
    }

    @RequestMapping("/search")
    public String search(
        Model model, 
        @RequestParam(value = "space", required = false) @Max(value=100) String space) {
        // ommit
    }
}
```

2. Aop class definition cache won't work well, therefore memory leak

## Why

this bug is intruced by:

- https://github.com/spring-projects/spring-framework/pull/26017/files#diff-8b811560ae60c9f7937f0d378880f1b4366fa67ea7024db018931a26e3b449edL370-L376
- https://github.com/spring-projects/spring-framework/pull/26017/files#diff-dd90aa36a7221a652e04631ac19dcb6693a9c841e43a90d36cc3caa51121630eL945-L947

1. Because `@Validated` annotation added,
`AbstractAdvisingBeanPostProcessor#postProcessAfterInitialization` add advice for `@Validated`

2. Before https://github.com/spring-projects/spring-framework/pull/26017/,
cache in org.springframework.cglib.core.AbstractClassGenerator  works well.
Because org.springframework.aop.framework.AdvisedSupport#getConfigurationOnlyCopy
copy org.springframework.aop.framework.AdvisedSupport#advisorsArray for create copy instance.

But after https://github.com/spring-projects/spring-framework/pull/26017/,
use same reference of org.springframework.aop.framework.AdvisedSupport#advisors for 
create copy instance. and this advitors field update by AbstractAdvisingBeanPostProcessor#postProcessAfterInitialization

3. When Create Aop class instance, try to reuse cache for class definition

Cache key is defined below code:

- org.springframework.cglib.proxy.Enhancer#createClass
- org.springframework.cglib.proxy.Enhancer#createHelper

org.springframework.cglib.proxy.Enhancer#filter is registered at
https://github.com/spring-projects/spring-framework/blob/main/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java#L201-L202

4. cache miss for class definition, memory leak happen
At AbstractAdvisingBeanPostProcessor#postProcessAfterInitialization, advice object
updated and therefore cache key is pollute, cache miss and memory leak happen.
memoryleak

So, When creating copy instance on org.springframework.aop.framework.AdvisedSupport#getConfigurationOnlyCopy,
It is good that create new instance of ArrayList and copy all advices.

This bug reported by @hikoma Thanks!

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
@@ -521,8 +521,8 @@ public class AdvisedSupport extends ProxyConfig implements Advised {
 		copy.copyFrom(this);
 		copy.targetSource = EmptyTargetSource.forClass(getTargetClass(), getTargetSource().isStatic());
 		copy.advisorChainFactory = this.advisorChainFactory;
-		copy.interfaces = this.interfaces;
-		copy.advisors = this.advisors;
+		copy.interfaces = new ArrayList<>(this.interfaces);
+		copy.advisors = new ArrayList<>(this.advisors);
 		return copy;
 	}
 


sbrannen:
Thanks for the PR and the detailed report.

We'll look into it.

sbrannen:
Related issues:

-  #26266

PR #27379, opened 2021-09-10
Migrate CoroutinesUtils from Kotlin to Java

philwebb:
This PR migrates `CoroutinesUtils` from Kotlin code to Java. The primary benefit of the change is that the Kotlin module is no longer required when working on the Java code in an IDE. This is especially helpful for Eclipse users (like myself and @sbrannen) as we no longer need to install Kotlin tooling.


sdeleuze (approved):

sdeleuze:
Ok from my POV when rebased on top of the latest Gradle changes, with extra care in testing snapshots with my Coroutines samples and with Spring Data which makes an extensive usage of Coroutines cc @mp911de.

mp911de:
Thanks for the ping. We're not using `CoroutinesUtils` in Spring Data as we have our own Coroutines to `Publisher` and vice versa translation.

sbrannen:
This is now in `main`.

Thanks, @philwebb!

PR #27387, opened 2021-09-11
Fix misplaced comma in AOP doc

rcrlik:
I think that a comma was misplaced in a sentece in the Core / AOP documetation.


snicoll:
Good catch @crlikcngroup and thank you for making your first contribution to Spring Framework.

PR #27415, opened 2021-09-15
Add HTTP response status to request logging.

selalerercapitolis:


bclozel:
Sorry it took so long for us to review this PR.

`AbstractRequestLoggingFilter` is only about request logging and this should remain the case. If response is interesting to your case, I'd suggest extending this contract with response-related methods. This PR mixes response information with the request logging information, I think this goes against the design of this class. I'm closing this PR as a result.

Thanks!

PR #27417, opened 2021-09-15
JpaTransactionManager datasource is not thread safe

bwilsonPaychex:
If JpaTransactionManager is used in a multitenant setup then the dataSource is not thread safe. This is in contrast to the TransactionSynchronizationMangager which uses ThreadLocals to manage its fields.

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/hibernate5/HibernateTransactionManager.java
+++ b/spring-orm/src/main/java/org/springframework/orm/hibernate5/HibernateTransactionManager.java
@@ -116,7 +116,7 @@ public class HibernateTransactionManager extends AbstractPlatformTransactionMana
 	private SessionFactory sessionFactory;
 
 	@Nullable
-	private DataSource dataSource;
+	private ThreadLocal<DataSource> dataSource = new ThreadLocal<DataSource>();
 
 	private boolean autodetectDataSource = true;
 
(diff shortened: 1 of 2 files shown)


mdeinum:
Wouldn't this apply to the other `PlatformTransactionManager` that operate on a `DataSource` as well? And wouldn't this be a slippery slope? Shouldn't the `DataSource`, when used with multi tenancy, be a proxy/delegate that eventually operates on the actual instance (like the `AbstractRoutingDatasource` for instance).

bwilsonPaychex:
> Wouldn't this apply to the other `PlatformTransactionManager` that operate on a `DataSource` as well? And wouldn't this be a slippery slope? Shouldn't the `DataSource`, when used with multi tenancy, be a proxy/delegate that eventually operates on the actual instance (like the `AbstractRoutingDatasource` for instance).

Are you referring the HibernateTransactionManager? If so, then yes I can make the change there as well. We do not use that one but it would have the same issue.

As per slippery slope, I can tell you much of the code in JpaTransactionManager is assuming the dataSource isn't switched during processing hence why the ThreadLocal was needed. Using the ThreadLocal did completely solve the issue we had

snicoll:
Thanks for the PR @bwilsonPaychex but we can't really consider it without some tests that exercise (and fix) the behavior you've described. Can you please add some?

PR #27466, opened 2021-09-25
Optimize header removal in ForwardedHeaderFilter

GreenRecycleBin:
The current implementation suggests that the request's headers are not
expected to change. Hence, it's not necessary to copy them.
Furthermore, it might be costly to do so if there are many headers.
Instead, cache only the request's header names for method getHeaderNames.

Methods getHeader and getHeaders delegate to the respective methods of
request if the header name is not in FORWARDED_HEADER_NAMES. Otherwise,
they return null or an empty Enumeration respectively.


GreenRecycleBin:
It took a few tries to get `./gradlew :spring-web:checkstyleMain` passing. Is there a IntelliJ IDEA XML file for the project's style? I've looked through the documentation but didn't find any.

PR #27468, opened 2021-09-25
Add missing @Nullable annotations on ResponseCookieBuilder

ghostd:
The ResponseCookie class already has @Nullable annotations for domain and path properties;
but the builder was not synchronised with these declarations.


jhoeller:
This a bit of a subtle topic: On various builder APIs of ours, we tend to enforce non-null values for the calls given there, as a measure of API guidance. For a builder, arguably you shouldn't be calling a particular method if you have no value to provide, and resetting existing values is not part of the scope there. From that perspective, the present nullability mismatch is intentional.

ghostd:
I understand, thanks for the explanation. I'll cose the PR.

PR #27477, opened 2021-09-27
Remove redundant assignment of default values to volatile/atomic fields

stsypanov:
This is a continuation of https://github.com/spring-projects/spring-framework/pull/25261


stsypanov:
Any reviews?

PR #27494, opened 2021-09-29
SimpleCacheResolver resolves templated SpEL expressions

Sam-Kruglov:
Currently, it's expected from the user to provide their own `CacheResolver` if they wish to use SpEL inside `@Cache*(cacheNames="...")`. I think it wouldn't hurt if it was implemented out of the box as templated expressions:
```
"myCache-#{#arg}-#{@bean.something()}" etc.
```
Especially given the fact that `CacheOperationExpressionEvaluator` is already implemented for `key`, `unless`, and `condition` attributes of `@Cache*` annotations.

And it will behave exactly the same as it does now but also can evaluate a templated expression if there is one:
```java
if (!expression.contains(templateParserContext.getExpressionPrefix())) return expression;
```
I decided to put that stuff right into `SimpleCacheResolver` making its factory methods backward-compatible. I don't see a point in creating a second resolver class because it still does the same thing if there's no nested expression in the name, and there's effectively no penalty in simply checking `String#contains` every time.

Example usecase:
```java
@Cacheable(cacheNames = "schemas-#{#vendorId}", key = "#entityId")
public Schema getSchema(String vendorId, String entityId){...}

@CacheEvict(cacheNames = "schemas-#{#vendorId}", allEntries = true)
public void updateVendor(String vendorId, VendorDto newData){...}
```
Here, if I don't have SpEL inside cache names, I can only define a cache named "schemas", but whenever I want to update one vendor, I only want its cache to be evicted, not everyone's. An alternative is to manage the cache programmatically or implement my own cache resolver.

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/CacheOperationExpressionEvaluator.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/CacheOperationExpressionEvaluator.java
@@ -28,6 +28,7 @@ import org.springframework.context.expression.BeanFactoryResolver;
 import org.springframework.context.expression.CachedExpressionEvaluator;
 import org.springframework.expression.EvaluationContext;
 import org.springframework.expression.Expression;
+import org.springframework.expression.common.TemplateParserContext;
 import org.springframework.lang.Nullable;
 
 /**
@@ -41,9 +42,10 @@ import org.springframework.lang.Nullable;
  * @author Phillip Webb
  * @author Sam Brannen
  * @author Stephane Nicoll
+ * @author Sam Kruglov
  * @since 3.1
  */
-class CacheOperationExpressionEvaluator extends CachedExpressionEvaluator {
+public class CacheOperationExpressionEvaluator extends CachedExpressionEvaluator {
 
 	/**
 	 * Indicate that there is no result variable.
@@ -67,6 +69,8 @@ class CacheOperationExpressionEvaluator extends CachedExpressionEvaluator {
 
 	private final Map<ExpressionKey, Expression> unlessCache = new ConcurrentHashMap<>(64);
 
+	private final Map<ExpressionKey, Expression> nameCache = new ConcurrentHashMap<>(64);
+
 
 	/**
 	 * Create an {@link EvaluationContext}.
@@ -79,12 +83,13 @@ class CacheOperationExpressionEvaluator extends CachedExpressionEvaluator {
 	 * {@link #NO_RESULT} if there is no return at this time
 	 * @return the evaluation context
 	 */
-	public EvaluationContext createEvaluationContext(Collection<? extends Cache> caches,
+	public EvaluationContext createEvaluationContext(@Nullable Collection<? extends Cache> caches,
 			Method method, Object[] args, Object target, Class<?> targetClass, Method targetMethod,
 			@Nullable Object result, @Nullable BeanFactory beanFactory) {
 
-		CacheExpressionRootObject rootObject = new CacheExpressionRootObject(
-				caches, method, args, target, targetClass);
+		Object rootObject = caches != null
+				? new CacheExpressionRootObjectWithCaches(caches, method, args, target, targetClass)
+				: new CacheExpressionRootObject(method, args, target, targetClass);
 		CacheEvaluationContext evaluationContext = new CacheEvaluationContext(
 				rootObject, targetMethod, args, getParameterNameDiscoverer());
 		if (result == RESULT_UNAVAILABLE) {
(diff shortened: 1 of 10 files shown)


spring-context/src/main/java/org/springframework/cache/interceptor/CacheOperationExpressionEvaluator.java
```diff
@@ -67,6 +69,8 @@
 
 	private final Map<ExpressionKey, Expression> unlessCache = new ConcurrentHashMap<>(64);
 
+	private final Map<ExpressionKey, Expression> nameCache = new ConcurrentHashMap<>(64);
```
Sam-Kruglov, line 72: Is there any reason why we can't just declare a single `Map<ExpressionKey, Expression>` inside the parent `CachedExpressionEvaluator`?

spring-context/src/test/java/org/springframework/cache/CacheReproTests.java
```diff
@@ -131,7 +132,7 @@ public void spr13081ConfigFailIfCacheResolverReturnsNullCacheName() {
 	}
 
 	@Test
-	public void spr14230AdaptsToOptional() {
+	public void spr14230AdaptsToOptionalAndResolvesCacheWithExpression() {
```
Sam-Kruglov, line 135: I think maybe there's no need to rigorously test this since it's using the already existing logic of `CacheOperationExpressionEvaluator`.

Here's the logs from this run (not the **Computed cache name**):
```java
12:43:04.885 [Test worker] TRACE o.s.c.a.AnnotationCacheOperationSource - Adding cacheable method 'insertItem' with attribute: [Builder[public org.springframework.beans.testfixture.beans.TestBean org.springframework.cache.CacheReproTests$Spr14230Service.insertItem(org.springframework.beans.testfixture.beans.TestBean)] caches=[itemCache-1] | key='#item.name' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless='']
12:43:04.902 [Test worker] TRACE o.s.c.a.AnnotationCacheOperationSource - Adding cacheable method 'findById' with attribute: [Builder[public java.util.Optional org.springframework.cache.CacheReproTests$Spr14230Service.findById(java.lang.String)] caches=[itemCache-#{'1'}] | key='' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless='' | sync='false']
12:43:04.954 [Test worker] TRACE o.s.c.i.CacheInterceptor - Computed cache key 'tb1' for operation Builder[public org.springframework.beans.testfixture.beans.TestBean org.springframework.cache.CacheReproTests$Spr14230Service.insertItem(org.springframework.beans.testfixture.beans.TestBean)] caches=[itemCache-1] | key='#item.name' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless=''
12:43:04.974 [Test worker] TRACE o.s.c.i.SimpleCacheResolver - Computed cache name 'itemCache-1' for operation Builder[public java.util.Optional org.springframework.cache.CacheReproTests$Spr14230Service.findById(java.lang.String)] caches=[itemCache-#{'1'}] | key='' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless='' | sync='false'
12:43:04.974 [Test worker] TRACE o.s.c.i.CacheInterceptor - Computed cache key 'tb1' for operation Builder[public java.util.Optional org.springframework.cache.CacheReproTests$Spr14230Service.findById(java.lang.String)] caches=[itemCache-#{'1'}] | key='' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless='' | sync='false'
12:43:04.975 [Test worker] TRACE o.s.c.i.CacheInterceptor - Cache entry for key 'tb1' found in cache 'itemCache-1'
12:43:05.026 [Test worker] TRACE o.s.c.i.SimpleCacheResolver - Computed cache name 'itemCache-1' for operation Builder[public java.util.Optional org.springframework.cache.CacheReproTests$Spr14230Service.findById(java.lang.String)] caches=[itemCache-#{'1'}] | key='' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless='' | sync='false'
12:43:05.026 [Test worker] TRACE o.s.c.i.CacheInterceptor - Computed cache key 'tb1' for operation Builder[public java.util.Optional org.springframework.cache.CacheReproTests$Spr14230Service.findById(java.lang.String)] caches=[itemCache-#{'1'}] | key='' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless='' | sync='false'
12:43:05.027 [Test worker] TRACE o.s.c.i.CacheInterceptor - No cache entry for key 'tb1' in cache(s) [itemCache-1]
12:43:05.027 [Test worker] TRACE o.s.c.i.CacheInterceptor - Computed cache key 'tb1' for operation Builder[public java.util.Optional org.springframework.cache.CacheReproTests$Spr14230Service.findById(java.lang.String)] caches=[itemCache-#{'1'}] | key='' | keyGenerator='' | cacheManager='' | cacheResolver='' | condition='' | unless='' | sync='false'

```

snicoll:
@Sam-Kruglov thank you for the PR but we're not keen to extend the support and rather recommend to use `CacheResolver`.

PR #27511, opened 2021-10-02
Add ParameterNamesModule to "well known" jackson modules

Sam-Kruglov:
Also, I'm surprised to see that `org.springframework.http.converter.json.Jackson2ObjectMapperBuilder#findModulesViaServiceLoader` is not `true` by default. Why not always just let it scan the classpath, seems like `findWellKnownModules` does the same but manually.

### DIFF
--- a/spring-web/spring-web.gradle
+++ b/spring-web/spring-web.gradle
@@ -59,6 +59,7 @@ dependencies {
 	testImplementation("com.fasterxml.jackson.datatype:jackson-datatype-jdk8")
 	testImplementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310")
 	testImplementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+	testImplementation("com.fasterxml.jackson.module:jackson-module-parameter-names")
 	testImplementation("org.apache.tomcat:tomcat-util")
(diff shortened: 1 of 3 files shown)


sdeleuze (approved):
`ParameterNamesModule` looks useful for common use case, so I will merge those changes.

I will raise your point about enabling service loader instead with the team. Well-know modules can be enabled via this mechanism so if it does not introduce side effects, that could indeed be an alternative.

bclozel:
@floatdrop See #40036 for a case that sounds close to yours.

floatdrop:
> @floatdrop See [spring-projects/spring-boot#40036](https://github.com/spring-projects/spring-boot/issues/40036) for a case that sounds close to yours.

Kind of the same – we have some classes with `@JsonProperty` in constructor as well as `@JsonProperty` on properties _plus_ public getters, that cause name conflicts when ParameterNamesModule is autoinjected in deserializer.

Excluding `jackson-module-parameter-names` from modules temporary fixes the issue.

Sam-Kruglov:
@floatdrop hey, unless you want property names to appear different in json than they are in the pojo, you can just delete those annotations. Jackson can just read constructor parameter names and getters via reflection. Never had a problem like that

Issue #23749, opened 2019-10-02
some methods can be added in StringUtils


rsbeoriginal:
---

To make string to uppercase and lowercase.

sbrannen:
Java has had built-in support for that since Java 1.0.

Please see `java.lang.String.toUpperCase()` and `java.lang.String.toLowerCase()`.

rsbeoriginal:
yes java has it, but calling the method on null object will return Null Pointer. So to prevent that is it possible to have a wrapper function over it ?

sbrannen:
As stated in the class-level Javadoc, `StringUtils` is...

> Mainly for internal use within the framework; consider Apache's Commons Lang for a more comprehensive suite of String utilities.

We typically do not introduce new utility methods unless there is a concrete need for them within the framework itself.

PR #27514, opened 2021-10-03
resolve spring-projects/spring-framework#27432. Added Assert.isFalse …

### LINKED ISSUE #27432: Add isFalse assertion
Hello, everyone!

How about to add `isFalse` assert in `org.springframework.util.Assert` class?
I think it would be nice to avoid negative assertions like that code:
`Assert.isTrue(!expression, "assert error");`

Or tell please, what is your approach to write negative assertions?

Thank you in advance!
Your faithfully,
springframework user

Melancholic:
Added `Assert.isFalse(...)` methods with JavaDoc & Unit Tests.

Resolve spring-projects/spring-framework#27432

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/Assert.java
+++ b/spring-core/src/main/java/org/springframework/util/Assert.java
@@ -57,6 +57,7 @@ import org.springframework.lang.Nullable;
  * @author Sam Brannen
  * @author Colin Sampaleanu
  * @author Rob Harrop
+ * @author Andrey Nagorniy
  * @since 1.1.2
  */
 public abstract class Assert {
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @Melancholic 

Thanks for creating your first PR for the Spring Framework.

Many people have made suggestions to introduce additional convenience methods in `Assert`, `StringUtils`, and similar utilities in the Spring Framework; however, they have typically all been declined along the lines stated in https://github.com/spring-projects/spring-framework/issues/17766#issuecomment-453427274, essentially since these utilities are only intended for use within the framework itself. 

Related Issues:

- #810
- #17508
- #17766
- #23749

In light of that, I am inclined to decline this request; however, I'll leave the issue open to allow the team to decide whether we want to potentially reconsider this for Spring Framework 6.0.

bclozel:
There isn't much demand for this and we tend to not introduce variants unless there is a clear use case for utility methods within Spring Framework. I'm closing this as a result.

PR #27544, opened 2021-10-11
Use Arrays.hashCode() in ByteArrayResource.hashCode()

stsypanov:
Current implementation produces equal hash codes for all instances of `ByteArrayResource` with the same length of underlying array disregarding their contents. This seems to be wrong as hash code should be calculated on the basis of the content of `byte[]`.


sbrannen:
This has been merged into `5.3.x` in 114fa47171e666cfd2214cf436f99c267b6a2d7d and revised in eb07dea7958b0314fc79ea4f4b5e8ce710c169ef.

Thanks

PR #27550, opened 2021-10-12
Include correct keyword in CookieAssertions failure messages

koenpunt:
I think the implementation of these methods was copied from the `secure` one, but the message wasn't updated to reflect the method name.


sbrannen:
Good catch.

Thanks

PR #27554, opened 2021-10-13
Use Charset variants of URLEncoder and URLDecoder methods

dreis2211:
Hi,

as the JDK baseline was raised, we can make use of the methods in `URLEncoder` and `URLDecoder` that take a `Charset`. This should save a few cycles and is also less code in some cases because we do not need to catch `UnsupportedEncodingException`s.

Cheers,
Christoph


sbrannen:
That's a very nice simplification of the status quo.

Thanks for noticing those new method variants and for the PR. 👍

Issue #23632, opened 2019-09-12
Spring "SchedulerFactoryBean" overrides the Quartz "jobstore" properties values 


ms3eed:
If you are using Spring's SchedulerFactoryBean, it overrides the configured value from the properties file.
So if you tried to use JobStoreTx it is always overridden by the LocalDataSourceJobStore from spring

Code snippet below shows the part from the SchedulerFactoryBean. I have overcome it by using a customizer. 

```CollectionUtils.mergePropertiesIntoMap(this.quartzProperties, mergedProps);
		if (this.dataSource != null) {
			mergedProps.setProperty(StdSchedulerFactory.PROP_JOB_STORE_CLASS, LocalDataSourceJobStore.class.getName());
		}
```
What if I wan to extend the existing LocalDataSourceJobStore?

PR #27560, opened 2021-10-14
Use LocalDataSourceJobStore only if one is not specified via Quartz properties

p-daniil:
Faced with the problem that I can't set my JobStore via quartz properties:
`spring.quartz.properties.org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreTX`

Fixes

-  #23632


p-daniil:
Hi @sbrannen,
I have added test.
Without my fix it results with:
```
expected: org.quartz.impl.jdbcjobstore.JobStoreTX
 but was: org.springframework.scheduling.quartz.LocalDataSourceJobStore
```

PR #27562, opened 2021-10-15
mark function resolveType as private

takeaction21:
function resolveType() is a internal function usually act as a callee, it is called by other function in ResolveType.java under conditional precondition. it is not a general function for users, we should mark this function as private explicitly, to prevent users from making mistakes


takeaction21:
should keep default package access

PR #27565, opened 2021-10-16
use URLEncoder/URLDecoder new method to simplify the code. Fix #27554

diguage:
Fix #27554


bclozel:
#27554 is already a PR and contributes those changes.

PR #27569, opened 2021-10-18
DefaultResponseErrorHandler update javadoc comment

smilep:
Error message truncation was removed as part of https://github.com/spring-projects/spring-framework/issues/27552, updating the javadoc comment of the method to reflect that.


rstoyanchev:
Indeed, thanks for catching this.

PR #27586, opened 2021-10-21
Improve example in Javadoc for HttpEntity

no-brand:
There is a minor typo in java doc.
To be specific, provided code snippet in `HttpEntity` does not work because of lack of quotation marks.

```java
HttpHeaders headers = new HttpHeaders();
headers.setContentType(MediaType.TEXT_PLAIN);
HttpEntity<String> entity = new HttpEntity<String>(helloWorld, headers);
URI location = template.postForLocation("https://example.com", entity);
```

**Affects:** \<5.3.10>


sbrannen:
Hi @no-brand,

Thanks for creating your first pull request for the Spring Framework! 👍 

It is assumed that `helloWorld` is a variable of type `String` defined elsewhere, but you're right that the example can be improved. So we'll improve it.

PR #27605, opened 2021-10-25
Update AbstractAdvisingBeanPostProcessor.java

baker-yuan:
resolve @async circular dependencies problem


snicoll:
@baker-yuan thanks for adding the test. I am not keen to bring this extra complexity for support cyclic dependencies. There are ways to break it using `Optional` or `ObjectProvider` and this should be the preferred option anyway.

PR #27606, opened 2021-10-25
Support comma-delimited list of origin patterns in @CrossOrigin

### LINKED ISSUE #24982: Cannot use SpEL expression to inject an array into @CrossOrigin's value attribute
**Affects:** \5.2.2

we usually inject an array use comma like this：
```properties
key=A,B,C
```
```java
@Value("${key}")
String[] keys;
```

but can't inject an array like below
```properties
origin=http://domainA.com,http://domainB.com
```
only key have one element, below SpEL can work fine, if key have more than 2 elements, it won't work
```java
@CrossOrigin(value = "${origin}")
```
Or
```java
@CrossOrigin(value = "#{'${origin}'.split(',')}")
```

Can provide any advice?

funky-eyes:
example

```java
@CrossOrigin(originPatterns = "${originPatterns:http://a.b.c:[8088,8089],http://d.e.f}", methods = {POST, OPTIONS})
```

I found that when using the above example configuration, it is not possible to configure multiple source addresses through properties, hope to support this function, thank you very much

fixed #24982


a364176773:
@rstoyanchev PTAL

rstoyanchev (changes requested):
Sorry for the slow reply. 

I think that `CorsConfiguration` should remain simple, and should not surprise in cases where input is collected in other ways, that could lead to a potential surprise. So, at the risk of some duplication (between spring-webmvc and spring-webflux), I think the support for splitting should be done in `RequestMappingHandlerMapping`, at the point of collection, and be documented as a feature on the respective `@CrossOrigin` attributes. Currently the annotation does not have any mention of the capbaility but it's where the feature needs to apply and be advertised as such.

Second, for the implementation, I think we should stick to using "," i.e. the common way of separating, but obviously the parsing will need to be smarter to iterate over the full string and ignore any port-related commas, i.e. within square brackets.

a364176773:
> Sorry for the slow reply.
> 
> I think that `CorsConfiguration` should remain simple, and should not surprise in cases where input is collected in other ways, that could lead to a potential surprise. So, at the risk of some duplication (between spring-webmvc and spring-webflux), I think the support for splitting should be done in `RequestMappingHandlerMapping`, at the point of collection, and be documented as a feature on the respective `@CrossOrigin` attributes. Currently the annotation does not have any mention of the capbaility but it's where the feature needs to apply and be advertised as such.
> 
> Second, for the implementation, I think we should stick to using "," i.e. the common way of separating, but obviously the parsing will need to be smarter to iterate over the full string and ignore any port-related commas, i.e. within square brackets.

thx, I'm going to improve that logic

a364176773:
@rstoyanchev I have finished the modification  PTAL

a364176773:
@sbrannen @rstoyanchev  There is a timeout error in concourse- CI, I do not know what the specific error is, please help me to check

sbrannen:
> There is a timeout error in concourse- CI, I do not know what the specific error is, please help me to check

A timeout on the CI server typically means that the CI server was under too much load at that point in time.

It's more important that you have successful local builds on your machine before pushing commits to your PR.

For example, I see that you are using `org.junit.jupiter.api.Assertions` which is forbidden in Spring Framework's Checkstyle rules. For assertions in tests, we [use AssertJ](https://github.com/spring-projects/spring-framework/wiki/Code-Style#assertions) exclusively.

Please make sure you get a passing build using `./gradlew build` locally before  pushing commits, and please push a commit that fixes any current issues breaking the build locally for you.

a364176773:
> > There is a timeout error in concourse- CI, I do not know what the specific error is, please help me to check
> 
> A timeout on the CI server typically means that the CI server was under too much load at that point in time.
> 
> It's more important that you have successful local builds on your machine before pushing commits to your PR.
> 
> For example, I see that you are using `org.junit.jupiter.api.Assertions` which is forbidden in Spring Framework's Checkstyle rules. For assertions in tests, we [use AssertJ](https://github.com/spring-projects/spring-framework/wiki/Code-Style#assertions) exclusively.
> 
> Please make sure you get a passing build using `./gradlew build` locally before pushing commits, and please push a commit that fixes any current issues breaking the build locally for you.

ok, thank you

rstoyanchev (commented):
As this is purely a web concern, there should be no changes in spring-beans, nor in spring-context. The parsing of an origin can be simplified a little and re-used differently (without involving spring-context). 

I'm also not sure we should support expressions that split into an array. It should be sufficient to support coma-separated origin values.

rstoyanchev:
I've gone for a slightly different implementation. Thanks for the PR in any case!

a364176773:
@rstoyanchev Sorry I'm just now seeing your message, thanks, and hopefully I'll have the opportunity to contribute again in the future

a364176773:
https://github.com/spring-projects/spring-framework/commit/badba7c3e976666536522102c9a90f8098be6e0f
This commit doesn't seem to carry support for SpEL expressions
@rstoyanchev

rstoyanchev:
The reporter tried using SpEL and suggested support for it, but my read of the issue is that the main goal was to inject an array through an external property. So we've added support for a coma-delimited value, which takes care of that goal.

PR #27623, opened 2021-10-29
Improve the efficiency of UrlPathHelper.getSanitizedPath()

happyWilliam0:
Hi,

Tiny optimization is possible using the new method (time complexity : O(N)) instead of the old one (time complexity : O(N^2)).

Cheers,
William


rstoyanchev:
The change helps more, the more duplicate slashes there are. However, for the happy scenario, which should be the most common case, it adds object allocations. I've updated the change to restore the initial check for the presence of duplicate slashes, and then continue from there if necessary.

PR #27642, opened 2021-11-05
fix ColumnMapRowMapper.java #apply Method Result KeyValue Not corresponding

016K:
add getColumnValue(Row row, String columnName)  Method
change apply(Row row, RowMetadata rowMetadata) code # 60 # 63 # 64 

Reason for submission:
# code 57 (Collection<String> columns = rowMetadata.getColumnNames();)
# method apply(Row row, RowMetadata rowMetadata) 

The order of the key is not the same as the order of the value, causing the result to be an error.
I sent you an email if you can see it. From admin@maxmc.cn


snicoll:
@mp911de can you please have a look to this one?

mp911de:
`RowMetadata.getColumnMetadatas()` requires an iteration order that corresponds with row column order. Iterating the resulting `List<? extends ColumnMetadata>` by counter must return the requested column value when calling `Row.get(index)`.

That being said, please let us know which driver is affected and either please report a bug into the affected driver project or we can do so.

snicoll:
Thanks Mark. @016K please follow-up with the request above if necessary.

Issue #27637, opened 2021-11-04
Add method to ClientResponse that returns Mono terminating with createException


jwChung:
[The method `ClientResponse.createException()` returns the `Mono<WebClientResponseException>` type.](https://github.com/spring-projects/spring-framework/blob/main/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ClientResponse.java#L182) `Mono<Exception>` feels like `Either<Exception, Exception>`. I think that the `Mono<RESULT>` type is actually needed. So, what if there is the `createError` method that returns the `Mono<RESULT>` type in the `ClientResponse` interface?

poutsma:
I can see the point for having a different exception, but can't you just `map` the `WebClientResponseException` to the type you'd like?

jwChung:
@poutsma Isn't `Mono<T>.onErrorMap/onErrorResume` for that? 

In many cases, a returning value of `createException()` is changed to `Mono<RESULT>` in the following way.

```java
Mono<RESULT> mono = clientResponse.createException().flatMap(Mono::error);
```

I wish I could just write it down as follows.

```java
Mono<RESULT> mono = clientResponse.createError();
```

rstoyanchev:
There is a related example in #27645 that doesn't even involve changing the exception. If you use `createException` from `exchangeToMono` and `exchangeToFlux` we have it documented as returning `Mono.error(response.createException())` but actually you need `response.createException().flatMap(Mono::error)` only to match the generic type of the result. It could also be `response.createException().cast(...)` but either way it's inconvenient.

It would make sense to align `createException` with `Mono#error` in terms of being able to cast to anything. After all an error signal switches from some type `T` to an `Exception` so this is to be expected. WDYT @poutsma?

poutsma:
Personally I prefer the explicitness that the `Mono<WebClientResponseException>` return type offers, as opposed to a generic type. In the latter case, reading the javadoc of the method is pretty much required to see what it does, and that is not true for the former, which pretty much does what you expect from `createException`.

That said, by looking at #27645, I realise that the current signature is not ideal either and requires the use of a `flatMap`, which is not obvious. There is no reason against having both `createException` and `createError`, so I will add it.

PR #27645, opened 2021-11-05
Fix incorrect example of error handling in WebClient Javadoc

filiphr:


rstoyanchev:
I've updated this slightly differently to just `createException` since in this example the left-hand side is `Mono<Object>` but otherwise it could also be cast with the `cast` operator. Note also that for 6.0 there is a new `createErrror` method that casts to any type, similar to Mono#error, see #27637.

filiphr:
Thanks for doing the changes @rstoyanchev. I do understand the reasoning for your adaptation. However, I have to say that the example will still be confusing to people. Most of the time I do not have a `Mono<Object>` as a return type. My main goal for this came from the fact that I wanted to not have exceptions thrown, but rather handle the error myself.

Nevertheless, hopefully people that are more familiar with the reactive stack will not be confused and will know that they need to do the flat map.

Great about the new `createError` method.

rstoyanchev:
Fair enough. I've updated 5.3.x to show a more representative sample and in main it shows use of `createError` already.

filiphr:
Thanks a lot @rstoyanchev, really appreciate that you are taking into consideration my feedback

PR #27646, opened 2021-11-06
Use 'toString(Charset)' instead of 'toString(String)' for encodings

SungMinHong:
As Spring Framework 6 uses JDK17 for its baseline, 
we can make use of 'toString(Charset)' which was introduced in JDK 10.


jhoeller:
Thanks for the pull request and for raising the topic so timely! I'll follow it up with a related commit of my own (#27664).

PR #27654, opened 2021-11-08
RFC: resttemplate builder

tomasbjerre:
I created [a library](https://github.com/tomasbjerre/spring-resttemplate-client) and would like to ask if it can be merged into Spring? I only added the production code in this PR to show what would need to be changed. The test cases are still in [the library](https://github.com/tomasbjerre/spring-resttemplate-client).

Would you like me to continue work on this?

-----
Dynamically create Spring RestTemplate (proxy class) client from annotated interface.

Given `MyApiInterface` is a Spring-annotated Java interface.

```java
final MyApiInterface myClient = SpringRestTemplateClientBuilder
  .create(MyApiInterface.class)
  .setUrl(this.getMockUrl())
  .setRestTemplate(restTemplate)         // Optional
  .setHeader("header-name", "the value") // Optional
  .setHeaders(HttpHeaders)               // Optional
  .build();
```

Here `myClient` is a dynamically created [Java Proxy object](https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/Proxy.html).

```java
final ResponseEntity<MyDTO> response = myClient.getMyDto();
```

The method invocation (`getMyDto()`) on that object (`myClient`) will be translated to a HTTP request using the annotations of that method in that interface. And `response` will be the result of that HTTP request.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/builder/SpringRestTemplateClientBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/client/builder/SpringRestTemplateClientBuilder.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


JWThewes:
Isn't this what feign is supposed to do? I dont't see a big difference here?

tomasbjerre:
The difference is the dependencies needes. It is nice not to have Feign and all its transitives on classpath.

bclozel:
Sorry we didn't get to this earlier, but we now have [dedicated support for this in Spring Framework](https://docs.spring.io/spring-framework/docs/6.0.0-SNAPSHOT/reference/html/integration.html#rest-http-interface). I'm closing this issue as a result.

Thanks!

PR #27670, opened 2021-11-11
Improve toString representation of complex parameters in SimpleKey

asgh:
Arrays.deep* is being used in all the other functions, use it in toString as well - this avoids debugging output that just looks like [Ljava.lang.Object;@123456

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/SimpleKey.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/SimpleKey.java
@@ -23,7 +23,6 @@ import java.util.Arrays;
 
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
 
 /**
  * A simple key as returned from the {@link SimpleKeyGenerator}.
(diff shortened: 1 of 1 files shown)


snicoll:
Very nice @asgh. Thank you for making your first contribution to Spring Framework.

MitrichDa:
@asgh It changed keys generated in @Cacheble. It can affect projects with multiple services and distributed cache

asgh:
> @asgh It changed keys generated in @cacheble. It can affect projects with multiple services and distributed cache

You have multiple servers running apps using different version of spring? Seems risky to me, but still wouldn't it just mean some things are cached twice? Or are you doing complex cache invalidation?

Are you really sure that `@cachable` is using the `toString()` version of a key to determine equality for caching? Something tells me that's not what's happening.

MitrichDa:
@asgh Microservices with redis distributed cache, yes, and not all of them update spring version simultaneously. 

@Cacheable(cacheNames = "test-cache")
public String myMethod(String test1, String test2) {
    // Method logic
} 
myMethod("test1", "test2");

will result with key in redis test-cache::SimpleKey [test1, test2] or test-cache::SimpleKey [test1,test2] depending on version

bclozel:
@MitrichDa I'm afraid it is too late to consider reverting that change. [Spring Framework 6.1 has been out for a while and is almost out of open source support](https://spring.io/projects/spring-framework#support). Spring Framework 6.2 is also using the same algorithm, so changing it back will disrupt you more.

I would suggest updating your cache keys to using a more predictable value if you wish to use a cache with various Spring Framework minor versions.

PR #27671, opened 2021-11-11
Change configureAndRefreshWebApplicationContext change to requireNonNullElseGet if use jdk 17

xixingya:
configureAndRefreshWebApplicationContext change to requireNonNullElseGet if use jdk 17


jhoeller:
Generally speaking, this new `Objects.requireNonNullElseGet` method is indeed useful for such scenarios. However, in our framework codebase, I'd rather leave such simple `if` checks in place with dedicated if/else code blocks, allowing for inline documentation of each block and making the structure easy to grasp visually. Also, a regular `if` construct is slightly more efficient than a lambda-provided supplier at runtime.

PR #27674, opened 2021-11-11
fix: the basePackage do not have META-INF/spring.component then switc…

qxo:
If the basePackage not defined in META-INF/spring.component,
then using old `scanCandidateComponents(basePackage)`;

Otherwise the basePackage not in META-INF/spring.component will not load
if the env classpath has any META-INF/spring.component file.

the issue poc code: https://github.com/qxo/spring-components-index-bug-poc


snicoll:
@qxo thanks for the PR but the component indexer has been deprecated so we won't make further change to it, see #30431

PR #27679, opened 2021-11-12
fix :  the same method was cached twice in the "attributeCache", changed to once

androiderLee:
**in  AbstractFallbackTransactionAttributeSource**
```
public TransactionAttribute getTransactionAttribute(Method method, @Nullable Class<?> targetClass) {
	// ...
	Object cacheKey = getCacheKey(method, targetClass);
	TransactionAttribute cached = this.attributeCache.get(cacheKey);
	if (cached != null) {
		//...
	}
	else {
		// We need to work it out.
		TransactionAttribute txAttr = computeTransactionAttribute(method, targetClass);
		//...
	}
}

```
**When using JDK dynamic proxy：**
```
@Configuration
@EnableTransactionManagement(mode = AdviceMode.PROXY )
public class EnableTransactionManagementBean {
   //...
}
```
**For example：**
```
public interface YqleeService {
    void  testMethod();
}
```

```
@Service
public class YqleeServiceImpl implements YqleeService {

    @Transactional
    @Override
    public void testMethod() {
        //TODO do something
    }
}
```

when calling before generating the JDK dynamic proxy, the parameter "method" belongs to the implementation class(YqleeServiceImpl )：

before

but, when calling after generating the JDK dynamic proxy, the parameter "method" belongs to the interface(YqleeService)：

after

**The method "getTransactionAttribute"  is called twice, but the "attributeCache" does not work very well because the "cacheKey" is different.because the parameter "method" belongs to a different class.**

```
// First, see if we have a cached value.
Object cacheKey = getCacheKey(method, targetClass);
```
It might be better to change to this：

`Object cacheKey = getCacheKey(AopUtils.getMostSpecificMethod(method,targetClass), targetClass);`

**as same as AbstractFallbackCacheOperationSource**

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/AbstractFallbackCacheOperationSource.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/AbstractFallbackCacheOperationSource.java
@@ -89,7 +89,7 @@ public abstract class AbstractFallbackCacheOperationSource implements CacheOpera
 			return null;
 		}
 
-		Object cacheKey = getCacheKey(method, targetClass);
+		Object cacheKey = getCacheKey(AopUtils.getMostSpecificMethod(method,targetClass), targetClass);
 		Collection<CacheOperation> cached = this.attributeCache.get(cacheKey);
 
 		if (cached != null) {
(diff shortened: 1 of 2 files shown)


androiderLee:
@rstoyanchev 
I have updated the description, please review when you have time, thank you!

snicoll:
@androiderLee thank you for the PR but `getMostSpecificMethod` is not a cheap operation so that change will most probably have the opposite effect of what you intended. It's fine if we have a duplicate entry in the cache for this, compared to paying the cost of calling this method every time.

PR #27680, opened 2021-11-12
Use `Integer.parseInt` without substring method

KimDoubleB:
As Spring Framework 6 uses JDK17 for its baseline, `parseInt` method can be used without using `substring` method.
- use `parseInt(CharSequence s, int beginIndex, int endIndex, int radix)` - JDK 9

I think this modification can prevent to create unnecessary string object.


poutsma:
Merged in 804b343

Thank you for submitting a PR!

PR #27702, opened 2021-11-19
Use InputStream's new methods instead of StreamUtils

yaboong:
As Spring Framework 6 uses JDK17 for its baseline, we can make use of the methods `transferTo(OutputStream out)` and `readAllBytes()` supported by `InputStream` in JDK9 instead of `StreamUtils`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StreamUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StreamUtils.java
@@ -140,6 +140,29 @@ public abstract class StreamUtils {
 	}
 
 	/**
+	 * Copy the contents of the given InputStream to the given OutputStream.
+	 * <p>Leaves both streams open when done.
+	 * @param in the InputStream to copy from
+	 * @param out the OutputStream to copy to
+	 * @return the number of bytes copied
+	 * @throws IOException in case of I/O errors
+	 */
+	public static int copy(InputStream in, OutputStream out) throws IOException {
+		Assert.notNull(in, "No InputStream specified");
+		Assert.notNull(out, "No OutputStream specified");
+
+		int byteCount = 0;
+		byte[] buffer = new byte[BUFFER_SIZE];
+		int bytesRead;
+		while ((bytesRead = in.read(buffer)) != -1) {
+			out.write(buffer, 0, bytesRead);
+			byteCount += bytesRead;
+		}
+		out.flush();
+		return byteCount;
+	}
+
+	/**
 	 * Copy a range of content of the given InputStream to the given OutputStream.
 	 * <p>If the specified range exceeds the length of the InputStream, this copies
 	 * up to the end of the stream and returns the actual number of copied bytes.
--- a/spring-core/src/main/java/org/springframework/util/StreamUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StreamUtils.java
@@ -65,9 +65,7 @@ public abstract class StreamUtils {
 			return new byte[0];
 		}
 
-		ByteArrayOutputStream out = new ByteArrayOutputStream(BUFFER_SIZE);
-		copy(in, out);
-		return out.toByteArray();
+		return in.readAllBytes();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


spring-core/src/main/java/org/springframework/util/StreamUtils.java
```diff
@@ -141,29 +139,6 @@ public static void copy(String in, Charset charset, OutputStream out) throws IOE
...
 
-	/**
-	 * Copy the contents of the given InputStream to the given OutputStream.
-	 * <p>Leaves both streams open when done.
-	 * @param in the InputStream to copy from
-	 * @param out the OutputStream to copy to
-	 * @return the number of bytes copied
-	 * @throws IOException in case of I/O errors
-	 */
-	public static int copy(InputStream in, OutputStream out) throws IOException {
```
AntonLGVS, line 152: Don't delete it. We have many legacy projects that still use this functionality. I propose to do this deprecate.

yaboong: @AntonLGVS  Restored the method I deleted, thanks.
If it's still needed to support legacy projects, how about making this method use `transferTo()` in `InputStream`?
like below
```java
public static int copy(InputStream in, OutputStream out) throws IOException {
    Assert.notNull(in, "No InputStream specified");
    Assert.notNull(out, "No OutputStream specified");

    int byteCount = (int) in.transferTo(out);
    out.flush();
    return byteCount;
}
```

bananayong: I'm not sure Spring should keep StreamUtils#copy method. For Utils such as copy method, It looks like more proper to use apache commons or guava libraries.

AntonLGVS: @yaboong Thanks. It's a good idea to use `transferTo()` within the `copy` method.

AntonLGVS: @bananayong Why use an external library when Spring provides a basic set of methods. If you want to delete, you must first make a deprecate to give time for the transition. Otherwise, updating the spring package will break everything.

spring-core/src/main/java/org/springframework/util/StreamUtils.java
```diff
@@ -65,9 +65,7 @@
 			return new byte[0];
 		}
 
-		ByteArrayOutputStream out = new ByteArrayOutputStream(BUFFER_SIZE);
-		copy(in, out);
-		return out.toByteArray();
+		return in.readAllBytes();
```
poutsma, line 68: This changes the behavior of this method. Formally, a maximum of `BUFFER_SIZE` (4096) was read, whereas `InputStream::readAllBytes()` reads `Integer.MAX_VALUE`.

spring-web/src/main/java/org/springframework/http/converter/ByteArrayHttpMessageConverter.java
```diff
@@ -53,11 +52,7 @@ public boolean supports(Class<?> clazz) {
 
 	@Override
 	public byte[] readInternal(Class<? extends byte[]> clazz, HttpInputMessage inputMessage) throws IOException {
-		long contentLength = inputMessage.getHeaders().getContentLength();
-		ByteArrayOutputStream bos =
-				new ByteArrayOutputStream(contentLength >= 0 ? (int) contentLength : StreamUtils.BUFFER_SIZE);
-		StreamUtils.copy(inputMessage.getBody(), bos);
-		return bos.toByteArray();
+		return inputMessage.getBody().readAllBytes();
```
poutsma, line 55: The same comment made above applies here, but this commit also removes the `contentLength` optimisation. I think using `InputStream::readNBytes(contentLength >= 0 ? (int) contentLength : StreamUtils.BUFFER_SIZE)` makes more sense.

spring-core/src/main/java/org/springframework/util/StreamUtils.java
bananayong: `BUFFER_SIZE` looks only for buffer. After all, all of bytes from `InputStream` will be read. `copy` method is really copy all bytes from `InputStream` to `OutputStream`

If replaced with `InputStream::readNBytes(BUFFER_SIZE)` then only `BUFFER_SIZE` size of all bytes will be read.

PR #27716, opened 2021-11-22
Support SQUARE brackets [] around NAMED parameter

AntonLGVS:
At the moment, if sql statement has a variable in square brackets, then the name of the variable will be incorrect.
For example:
`SELECT ARRAY[:ext];`
We'll get "**ext]**" instead of "**ext**".

But:
`SELECT ARRAY[:ext   ]; --with white space`
We'll get "**ext**".


schauder:
I agree with the OP. This makes sense for the array syntax at least of [Postgres](https://www.postgresql.org/docs/9.1/arrays.html) and [DB2](https://www.ibm.com/docs/en/db2-for-zos/11?topic=type-arrays-in-sql-statements). It might even be the standard, but I don't know.

We should probably also make sure that expansion of collections into a comma separated list happens, so that this works when `ext` isn't just a singular argument but a collection of things, which would make sense for arrays.

An alternative workaround is to pass in a `java.sql.Array` directly.

PR #27735, opened 2021-11-26
DataAccessUtils result accessors with Optional return value

### LINKED ISSUE #27728: DataAccessUtils result accessors with Optional return value
Related to #724, `DataAccessUtils` should provide an `optionalResult` method for `Collection` as well as `Stream` arguments (and also for `Iterator` because we need that internally anyway), returning a corresponding `Optional` wrapper or `Optional.empty()`.

In addition, a `singleResult` overload for a `Stream` argument (and for `Iterator`) would complete the overall arrangement in `DataAccessUtils`.

KaidosGH:
Closes #27728
(https://github.com/spring-projects/spring-framework/issues/27728)

Hi everyone, this is my first PR in any project here on github with the goal of improving my understanding of java. Hope i understood the enhancement request correctly.

### DIFF
--- a/spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
+++ b/spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
@@ -74,11 +74,13 @@ public abstract class DataAccessUtils {
 		if (results == null) {
 			return null;
 		}
-		List<T> resultList = results.limit(2).toList();
-		if (resultList.size() > 1) {
-			throw new IncorrectResultSizeDataAccessException(1, resultList.size());
+		try (results) {
+			List<T> resultList = results.limit(2).toList();
+			if (resultList.size() > 1) {
+				throw new IncorrectResultSizeDataAccessException(1, resultList.size());
+			}
+			return CollectionUtils.isEmpty(resultList) ? null : resultList.get(0);
 		}
-		return CollectionUtils.isEmpty(resultList) ? null : resultList.get(0);
 	}
 
 	/**
@@ -134,11 +136,13 @@ public abstract class DataAccessUtils {
 		if (results == null) {
 			return Optional.empty();
 		}
-		List<T> resultList = results.limit(2).toList();
-		if (resultList.size() > 1) {
-			throw new IncorrectResultSizeDataAccessException(1, resultList.size());
+		try (results) {
+			List<T> resultList = results.limit(2).toList();
+			if (resultList.size() > 1) {
+				throw new IncorrectResultSizeDataAccessException(1, resultList.size());
+			}
+			return CollectionUtils.isEmpty(resultList) ? Optional.empty() : Optional.of(resultList.get(0));
 		}
-		return CollectionUtils.isEmpty(resultList) ? Optional.empty() : Optional.of(resultList.get(0));
 	}
 
 	/**
--- a/spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
+++ b/spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
@@ -141,14 +141,7 @@ public abstract class DataAccessUtils {
 	 * element has been found in the given Iterator
 	 */
 	public static <T> Optional<T> optionalResult(@Nullable Iterator<T> results) throws IncorrectResultSizeDataAccessException {
-		if (results == null) {
-			return Optional.empty();
-		}
-		T result = results.hasNext() ? results.next() : null;
-		if (results.hasNext()) {
-			throw new IncorrectResultSizeDataAccessException(1, 2);
-		}
-		return Optional.ofNullable(result);
+		return Optional.ofNullable(singleResult(results));
 	}
 
 	/**
--- a/spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
+++ b/spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
@@ -21,7 +21,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
 
 import org.springframework.dao.DataAccessException;
 import org.springframework.dao.EmptyResultDataAccessException;
@@ -75,11 +74,11 @@ public abstract class DataAccessUtils {
 		if (results == null) {
 			return null;
 		}
-		List<T> resultList = results.toList();
+		List<T> resultList = results.limit(2).toList();
 		if (resultList.size() > 1) {
 			throw new IncorrectResultSizeDataAccessException(1, resultList.size());
 		}
-		return resultList.stream().findFirst().orElse(null);
+		return CollectionUtils.isEmpty(resultList) ? null : resultList.get(0);
 	}
 
 	/**
@@ -96,12 +95,11 @@ public abstract class DataAccessUtils {
 		if (results == null) {
 			return null;
 		}
-		Iterable<T> iterable = () -> results;
-		List<T> resultList = StreamSupport.stream(iterable.spliterator(), false).toList();
-		if (resultList.size() > 1) {
-			throw new IncorrectResultSizeDataAccessException(1, resultList.size());
+		T result = results.hasNext() ? results.next() : null;
+		if (results.hasNext()) {
+			throw new IncorrectResultSizeDataAccessException(1, 2);
 		}
-		return resultList.stream().findFirst().orElse(null);
+		return result;
 	}
 
 	/**
@@ -136,11 +134,11 @@ public abstract class DataAccessUtils {
 		if (results == null) {
 			return Optional.empty();
 		}
-		List<T> resultList = results.toList();
+		List<T> resultList = results.limit(2).toList();
 		if (resultList.size() > 1) {
 			throw new IncorrectResultSizeDataAccessException(1, resultList.size());
 		}
-		return resultList.stream().findFirst();
+		return CollectionUtils.isEmpty(resultList) ? Optional.empty() : Optional.of(resultList.get(0));
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


BrucessKING:
Using the '.*' form of import should be avoided - java.util.*.

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +61,109 @@
...
+	 * @throws IncorrectResultSizeDataAccessException if more than one
+	 * element has been found in the given Iterator
+	 */
+	@Nullable
+	public static <T> T singleResult(@Nullable Iterator<T> results) throws IncorrectResultSizeDataAccessException {
+		if (results == null) {
+			return null;
+		}
+		Iterable<T> iterable = () -> results;
+		List<T> resultList = StreamSupport.stream(iterable.spliterator(), false).toList();
```
marschall, line 100: Why a stream instead of just

    results.next()

KaidosGH:
> Using the * form of import should be avoided - java.util.*.

Thanks for the tip, i adjusted that.

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +61,109 @@
...
+	 * @return the single result object, or {@code null} if none
+	 * @throws IncorrectResultSizeDataAccessException if more than one
+	 * element has been found in the given Stream
+	 */
+	@Nullable
+	public static <T> T singleResult(@Nullable Stream<T> results) throws IncorrectResultSizeDataAccessException {
+		if (results == null) {
+			return null;
+		}
+		List<T> resultList = results.toList();
```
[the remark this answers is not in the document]
KaidosGH: Hi @marschall,

thanks for reviewing, sounds good, i refactored this. Does it make more sense now?

`throw new IncorrectResultSizeDataAccessException(1, resultList.size());`
as we limited the resultList size with .limit, this would not report the original passed Stream results size though but instead a maximum of 2... is that an issue?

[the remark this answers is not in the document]
marschall: Yes, looks better to me.

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +60,107 @@
...
+	 * @return the single result object, or {@code null} if none
+	 * @throws IncorrectResultSizeDataAccessException if more than one
+	 * element has been found in the given Stream
+	 */
+	@Nullable
+	public static <T> T singleResult(@Nullable Stream<T> results) throws IncorrectResultSizeDataAccessException {
+		if (results == null) {
+			return null;
+		}
+		List<T> resultList = results.limit(2).toList();
```
marschall, line 77: There is a trade off here and in `#optionalResult` whether we should call `#close` on the stream:

On one hand Java convention is the code who creates a "resource" is responsible for disposing it, in this case the caller of this method would be responsible to call `#close` on the streams where required.
On the other hand failing to call `#close` on some streams, like the ones returned by `JdbcOperations#queryForStream`, can lead to resource leaks. Having the caller put streams in a try-with-resources block is both cumbersome and error prone.

KaidosGH: Makes sense. The passed stream is now closed in both methods. Also e.g. hibernate's Query.stream() documentation suggests the stream should be closed afterwards, so not so far fetched 👍

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +60,111 @@
...
+	 * element has been found in the given Collection
+	 */
+	public static <T> Optional<T> optionalResult(@Nullable Collection<T> results) throws IncorrectResultSizeDataAccessException {
+		if (CollectionUtils.isEmpty(results)) {
+			return Optional.empty();
+		}
+		if (results.size() > 1) {
+			throw new IncorrectResultSizeDataAccessException(1, results.size());
+		}
+		return results.stream().findFirst();
```
rkvigneswaran, line 123: Would it be better if we delegate this to original `singleResult()` method

```suggestion
		return Optional.ofNullable(singleResult(results));
```

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +60,111 @@
...
+	public static <T> Optional<T> optionalResult(@Nullable Stream<T> results) throws IncorrectResultSizeDataAccessException {
+		if (results == null) {
+			return Optional.empty();
+		}
+		try (results) {
+			List<T> resultList = results.limit(2).toList();
+			if (resultList.size() > 1) {
+				throw new IncorrectResultSizeDataAccessException(1, resultList.size());
+			}
+			return CollectionUtils.isEmpty(resultList) ? Optional.empty() : Optional.of(resultList.get(0));
```
rkvigneswaran, line 144: Would it be better if we delegate this to newly added singleResult(Stream<T>) method?
```suggestion
		return Optional.ofNullable(singleResult(results));
```

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +60,111 @@
...
+	 */
+	public static <T> Optional<T> optionalResult(@Nullable Iterator<T> results) throws IncorrectResultSizeDataAccessException {
+		if (results == null) {
+			return Optional.empty();
+		}
+		T result = results.hasNext() ? results.next() : null;
+		if (results.hasNext()) {
+			throw new IncorrectResultSizeDataAccessException(1, 2);
+		}
+		return Optional.ofNullable(result);
```
rkvigneswaran, line 165: Would it be better if we delegate this to newly added `singleResult(Iterator<T>)` method?
```suggestion
		return Optional.ofNullable(singleResult(results));
```

rkvigneswaran (changes requested):
Hi,

This is my first review comments in any open source project. Kindly let me know if my comments are invalid.

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
KaidosGH: Hi @rkvigneswaran,

that is actually pretty smart. Cuts out a lot of redundant code! 👍

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +60,89 @@
...
+	 */
+	@Nullable
+	public static <T> T singleResult(@Nullable Stream<T> results) throws IncorrectResultSizeDataAccessException {
+		if (results == null) {
+			return null;
+		}
+		try (results) {
+			List<T> resultList = results.limit(2).toList();
+			if (resultList.size() > 1) {
+				throw new IncorrectResultSizeDataAccessException(1, resultList.size());
```
bananayong, line 80: I think IncorrectResultSizeDataAccessException should contain the size of the results stream.
if `resultList.size()` will be used, actual size of IncorrectResultSizeDataAccessException is `<=2` always unlike actual size of results.

If prefer stream way, How about utilze `teeing`?
```java
		results.collect(teeing(counting(), reducing((first, second) -> first), (count, result) -> {
			if (count > 1) {
				throw new IncorrectResultSizeDataAccessException(1, count.intValue());
			}
			return result.orElse(null);
		}))
```

rkvigneswaran: Just a suggestion. There are multiple constructors provided by `IncorrectResultSizeDataAccessException` class. 

* One of the constructors doesn't take `actualSize` argument and assigns the value `-1` if the size if unknown. You can consider using it
  * `public IncorrectResultSizeDataAccessException(int expectedSize)` - [Javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/dao/IncorrectResultSizeDataAccessException.html#IncorrectResultSizeDataAccessException-int-)
* Another one explicitly suggests setting the value to `-1` if size is unknown
  * `public IncorrectResultSizeDataAccessException(int expectedSize, int actualSize)` - [Javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/dao/IncorrectResultSizeDataAccessException.html#IncorrectResultSizeDataAccessException-int-int-)

Both of them would be suitable for this scenario since size is unknown

bananayong: It makes sense.

rkvigneswaran:
```suggestion
				throw new IncorrectResultSizeDataAccessException(1);
```

Adding as suggestion for convenience

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
```diff
@@ -56,6 +60,89 @@
...
+	 * element has been found in the given Iterator
+	 */
+	@Nullable
+	public static <T> T singleResult(@Nullable Iterator<T> results) throws IncorrectResultSizeDataAccessException {
+		if (results == null) {
+			return null;
+		}
+		T result = results.hasNext() ? results.next() : null;
+		if (results.hasNext()) {
+			throw new IncorrectResultSizeDataAccessException(1, 2);
```
rkvigneswaran, line 102: Same suggestion as above : [Link](https://github.com/spring-projects/spring-framework/pull/27735/files#r757849057)
```suggestion
			throw new IncorrectResultSizeDataAccessException(1);
```
Since `actualSize` of the iterator isn't known beforehand, we can use `public IncorrectResultSizeDataAccessException(int expectedSize)` constructor and set the `actualSize`  as `-1`

rkvigneswaran (changes requested):
Modifying the exception to make it clear that size of the underlying Stream/Iterator is not known

spring-tx/src/main/java/org/springframework/dao/support/DataAccessUtils.java
KaidosGH: yeah let's use the `public IncorrectResultSizeDataAccessException(int expectedSize)` for the iterator and stream method. 
What do you mean by "UT"?

PR #27754, opened 2021-12-01
Fix transitive dependency on  Java EE 8 JAX-B API in spring-webmvc test compile classpath

### LINKED ISSUE #27685: spring-webmvc has the Java EE 8 JAX-B API on its test compile classpath (through dom4j)
It's a transitive dependency of `org.dom4j:dom4j`: https://ge.spring.io/s/rhkh754ui6xpc/dependencies?dependencies=jaxb-api&expandAll&focusedDependency=WzIzLDExMiw1OTg1LFswLDMsWzNdXV0&focusedDependencyView=versions.

xixingya:
spring-webmvc has the Java EE 8 JAX-B API on its test compile classpath (through dom4j) 
see https://github.com/spring-projects/spring-framework/issues/27685

### DIFF
--- a/spring-webmvc/spring-webmvc.gradle
+++ b/spring-webmvc/spring-webmvc.gradle
@@ -45,7 +45,9 @@ dependencies {
 	testImplementation("org.apache.httpcomponents:httpclient")
 	testImplementation("commons-io:commons-io")
 	testImplementation("org.mozilla:rhino")
-	testImplementation("org.dom4j:dom4j")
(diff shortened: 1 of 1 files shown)


rstoyanchev:
We'll need to have a closer look at this anyway to verify that it addresses the problem, so sending a one line change like this doesn't help.

bclozel:
Reopening to globally exclude the transitive dependency

PR #27764, opened 2021-12-04
Update spring-webmvc RequestBodyAdvice.supports method

selfcoders:
I found that when the RequestBodyAdvice is working, the supports method will be executed twice to determine whether it is applicable to the beforeBodyRead and afterBodyRead methods, but sometimes such a judgment is not necessary, so I think it should be split for different judgments, so I put RequestBodyAdvice The .supports method is divided into beforeBodySupport and afterBodySupport methods

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/JsonViewRequestBodyAdvice.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/JsonViewRequestBodyAdvice.java
@@ -57,7 +57,7 @@ public class JsonViewRequestBodyAdvice extends RequestBodyAdviceAdapter {
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


snicoll:
@selfcoders thanks for the PR but please stop creating duplicates PRs for the same issue. It is possible to amend your PR if you need to.

selfcoders:
> @selfcoders感谢 PR，但请停止为同一问题创建重复的 PR。如果需要，可以修改您的 PR。

Sorry this is my mistake

selfcoders:
The first push modification was just a method split, and the second push modification added new parameters to the original method.

bclozel:
This change breaks compatibility and doesn't help with performance - it actually makes things worse by calling the same method twice where instead we had a single call previously. This change might improve things for a custom implementation of yours, but I don't think that breaking the contract and hinder performance for most Spring applications out there is an acceptable tradeoff.

I'm declining this PR as a result. Thanks for your contribution!

PR #27768, opened 2021-12-06
Expose headers from native client request after request is committed

sokomishalov:
Fixes issues like [that](https://github.com/zalando/logbook/pull/1150#issuecomment-986223607), when public API does not expose headers that have been set by connector implementation.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/AbstractClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/AbstractClientHttpRequest.java
@@ -74,14 +74,13 @@ public abstract class AbstractClientHttpRequest implements ClientHttpRequest {
 		this.cookies = new LinkedMultiValueMap<>();
 	}
 
-
 	@Override
 	public HttpHeaders getHeaders() {
 		if (this.readOnlyHeaders != null) {
 			return this.readOnlyHeaders;
 		}
 		else if (State.COMMITTED.equals(this.state.get())) {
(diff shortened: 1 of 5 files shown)


rstoyanchev (changes requested):
The change might not be implemented in exactly this way. For example we don't use protected fields. 

Rather than providing a solution, please update the description of the pull request in order to describe the actual problem, and not just reference some external issue which requires additional context.

sokomishalov:
@rstoyanchev Fair enough
The problem is that `ClientHttpRequest::getHeaders` contain only headers that have been applied via `WebClient` API and don't contain headers applied by an underlying connector, (i.e. "user-agent").
Regarding implementation with restriction on protected fields - should this logic be duplicated to the `ReactorClientHttpRequest`?

rstoyanchev (changes requested):
Thanks for elaborating. 

We have a `NettyHeadersAdapter` that's used to expose the underlying headers on the response. We can enable the same on the request side. So, rather than exposing all the protected fields, I would suggest a protected `initReadOnlyHeaders` method that can be overridden in sub-classes. For Reactor Netty, it would be:
```java
@Override
protected HttpHeaders initReadOnlyHeaders() {
	return HttpHeaders.readOnlyHttpHeaders(new NettyHeadersAdapter(request.requestHeaders()));
}
```
Let me know if you plan to update the PR, or otherwise I'll take it from here.

sokomishalov:
@rstoyanchev totally makes sense, I've reworked it according to your recommendations (netty, apache http 5, jetty).
p.s. I've noticed you marked these changes to the 6.x milestone - is there any chance this enhancement to be backported on 5.x ?

PR #27792, opened 2021-12-09
Feature/add methods validate collections

muriloalvesdev:
To keep using !true to negate conditions I believe is something not as clear as a method that contains a name that describes what is being done. With that in mind, I decided to create these methods and open this PR.


snicoll:
@muriloalvesdev thanks for the PR but we're not keen to add invert of existing methods. You can easily do so in your own arrangement though.

PR #27801, opened 2021-12-11
 Polish ExceptionHandlerMethodResolver mapping

Drezir:
Refactor exception resolver cache get/null/put opration into single call.


snicoll:
Thanks for making your first contribution to Spring Framework.

PR #27804, opened 2021-12-12
Fix javadoc reference to ThrowsAdvice

aoyvx:
In the process of learning, I found a little flaw in Javadoc

Maybe we can change {@link org.springframework.aop.MethodBeforeAdvice} to {@link org.springframework.aop.ThrowsAdvice}


snicoll:
@aoyvx thank you for making your first contribution to Spring Framework.

PR #27807, opened 2021-12-13
Remove unnecessary @SuppressWarnings

quaff:
RequiredAnnotationBeanPostProcessor is removed 
see https://github.com/spring-projects/spring-framework/commit/4750a9430cdae9156d1e7fc32cec2c11ba2b8514


snicoll:
Good catch, thanks again @quaff.

PR #27822, opened 2021-12-16
Replace use of StringBuffer with StringBuilder

mdeinum:


sbrannen:
I was thinking we'd already gotten rid of `StringBuffer` usage in favor of `StringBuilder`, but those two methods were still forced to stick with `StringBuffer` on JDK 8.

With JDK 9, `java.util.regex.Matcher` received overloaded variants of `appendReplacement()` and `appendTail` that accept a `StringBuilder`. 👍 

Good catch!

sbrannen:

Thanks

PR #27823, opened 2021-12-16
Use try-with-resources for AutoClosables

mdeinum:
Noticed a couple of places that could benefit from try-with-resources.

### DIFF
--- a/spring-context-indexer/src/main/java/org/springframework/context/index/processor/MetadataStore.java
+++ b/spring-context-indexer/src/main/java/org/springframework/context/index/processor/MetadataStore.java
@@ -62,11 +62,8 @@ class MetadataStore {
 
 
 	private CandidateComponentsMetadata readMetadata(InputStream in) throws IOException {
-		try {
+		try (in){
 			return PropertiesMarshaller.read(in);
-		}
-		finally {
-			in.close();
 		}
 	}
 
--- a/spring-core/src/main/java/org/springframework/core/LocalVariableTableParameterNameDiscoverer.java
+++ b/spring-core/src/main/java/org/springframework/core/LocalVariableTableParameterNameDiscoverer.java
@@ -101,6 +101,8 @@ public class LocalVariableTableParameterNameDiscoverer implements ParameterNameD
 			}
 			return NO_DEBUG_INFO_MAP;
 		}
+		// We cannot use try-with-resources here as, potential, exception upon closing
+		// would still bubble up the stack
 		try {
 			ClassReader classReader = new ClassReader(is);
 			Map<Executable, String[]> map = new ConcurrentHashMap<>(32);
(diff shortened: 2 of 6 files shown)


spring-context-indexer/src/main/java/org/springframework/context/index/processor/MetadataStore.java
```diff
@@ -62,12 +62,9 @@ public void writeMetadata(CandidateComponentsMetadata metadata) throws IOExcepti
 
 
 	private CandidateComponentsMetadata readMetadata(InputStream in) throws IOException {
-		try {
+		try (in) {
```
sbrannen, line 65: At a quick glance, this change is the only change in this PR that does not break existing behavior.

Can you please revisit your changes and ensure that existing behavior is not altered?

For example, a try-with-resources block will close the `AutoCloseable` object, but it will not swallow exceptions thrown by the invocation of `close()`. So anywhere that we were previously intentionally swallowing exceptions such as `IOException`, we would still have to swallow those exceptions.

In addition, we would have to be able to disambiguate between an `IOException` thrown from with the try-block vs. one thrown by the invocation of `close()`.

In light of that, I don't think it makes sense to use try-with-resources in such use cases, but I'm happy to be proven wrong.

sbrannen (changes requested):

spring-context-indexer/src/main/java/org/springframework/context/index/processor/MetadataStore.java
mdeinum: You are right, I mistakenly thought that exceptions in a try-with-resource were going to be swallowed. They aren't and still bubble up. Lets remove and forget about this pull request :).

sbrannen: Actually, I'd appreciate it if you could keep the one valid change and add inline documentation to the other locations to point out why try-with-resources should not be used there.

That will provide a big help to the team to make sure we don't accidentally switch to try-with-resources for those use cases in the future.

sbrannen: >Actually, I'd appreciate it if you could keep the one valid change and add inline documentation to the other locations to point out why try-with-resources should not be used there.

Of course if you'd prefer to simply close this PR, that's also fine.

I'll leave it up to you.

mdeinum: Done. Left the one in place, placed a remark on the others.

sbrannen (approved):

spring-context-indexer/src/main/java/org/springframework/context/index/processor/MetadataStore.java
sbrannen: Thanks for making the changes. That looks good now.

sbrannen:

Thanks

PR #27824, opened 2021-12-16
Use Collection factory methods

mdeinum:
Use the collection factory methods when applicable
and available. Replaces the use of
Collections.unmodifiableList/Set/Map with the `of`
factory method.


sbrannen:

Thanks

Issue #27788, opened 2021-12-09
Add CacheErrorHandler implementation that logs exceptions rather than rethrowing them


bananayong:
Hello.

I would like to suggest that the Spring framework provide another `CacheErrorHandler` Implementation.
Whenever I build a spring application that uses an external cache, I do write a class that implements CacheErrorHandler for just error logging rather than throwing an exception.

Maybe most developers don't want to fail a method invocation when the cache is not available because the original method can return results.

So, If Spring provides another CacheErrorHandler implementation for just logging, will be very convinient.

Thank you!

---

Drezir:
I think that this is why we have `org.springframework.cache.annotation.CachingConfigurer` mechanism. But the implementation could be done, but the question is which logging level to use? Should it be configurable or not?

bananayong:
If logging level is configurable It could be very useful, but I think It's ok just WARN.

snicoll:
Closing in favor of PR #27826  - Thanks @Drezir!

PR #27826, opened 2021-12-16
Add CacheErrorHandler implementation that logs exceptions rather than rethrowing them

Drezir:
issue #27788

I have some more configurable version prepared but I think it is too much (see attachment)
[configurable-version.zip](https://github.com/spring-projects/spring-framework/files/7726399/configurable-version.zip)
e

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cache.interceptor;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.cache.Cache;
+import org.springframework.lang.Nullable;
+
+/**
+ * A {@link CacheErrorHandler} implementation that simply logs exception.
+ *
+ * @author Adam Ostrožlík
+ * @since 5.3
+ */
+public class LoggingCacheErrorHandler implements CacheErrorHandler {
+
+	protected static final Log logger = LogFactory.getLog(LoggingCacheErrorHandler.class);
+	private final boolean includeStacktrace;
+
+	/**
+	 * Construct new {@link LoggingCacheErrorHandler} that does not log stacktraces.
+	 */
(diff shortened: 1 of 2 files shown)


spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * A {@link CacheErrorHandler} implementation that simply logs exception.
+ *
+ * @author Adam Ostrožlík
+ * @since 5.3
+ */
+public class LoggingCacheErrorHandler implements CacheErrorHandler {
+
+	protected static final Log logger = LogFactory.getLog(LoggingCacheErrorHandler.class);
```
snicoll, line 33: I don't think this should be `protected`.

spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
```diff
@@ -0,0 +1,83 @@
...
+	 */
+	public LoggingCacheErrorHandler() {
+		this.includeStacktrace = false;
+	}
+
+	/**
+	 * Construct new {@link LoggingCacheErrorHandler} that may log stacktraces.
+	 * @param includeStacktrace whether to log or not log stacktraces.
+	 */
+	public LoggingCacheErrorHandler(boolean includeStacktrace) {
```
snicoll, line 47: It would be nice to have a constructor that takes a `Logger`. This way you can easily provide a mock and test the behavior.

spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
```diff
@@ -0,0 +1,83 @@
...
+		this.includeStacktrace = includeStacktrace;
+	}
+
+	@Override
+	public void handleCacheGetError(RuntimeException exception, Cache cache, Object key) {
+		logCacheError("Cache get error for key " + key + " and cache " + cache.getName(), exception);
```
snicoll, line 53: Perhaps putting the name of the cache first? Something like

`Cache 'test' failed to get entry with key 'my-key'`

snicoll (changes requested):
Thanks for the PR. It would be nice to have some tests to exercise the new code. Would you have the time to look at it? If you don't, no worries and let us know so that we can take over.

Drezir:
> Thanks for the PR. It would be nice to have some tests to exercise the new code. Would you have the time to look at it? If you don't, no worries and let us know so that we can take over.

Sure, I will do it.

spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
Drezir: Solved

spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
Drezir: Reformatted

spring-context/src/main/java/org/springframework/cache/interceptor/LoggingCacheErrorHandler.java
Drezir: I have added builder pattern and two test cases. Should tests cover each handler method or does this suffice?

snicoll:
@Drezir thanks for the follow-up.

Issue #27819, opened 2021-12-15
javax.servelet.http.Part field cannot be resolved with StandardMultipartHttpServletRequest.


binchoo:
**Affects: 5.3.2** 
**Related issue: #15220**

Hello, I'm trying to upload a file via multipart/form-data request. The handler method requires one object parameter, which contains a field of type `javax.servelet.http.Part`. As `MultipartFile` typed field can be resolved with parts, I've expected the same from `Part`.

```java
  @Setter
  @Getter
  @AllArgsConstructor
  @NoArgsConstructor
  public class UserHavingPartVO {
      private String name;
      private int age; 
      private Part file;
  }

  @Controller
  public class FileUploadController {

      @PostMapping("/vopart")
      public ResponseEntity bindVOHavingPart(UserHavingPartVO userVo){
          if (userVo.getFile() == null)
              return ResponseEntity.noContent().build();
          return ResponseEntity.ok().build();
      }
  }
```

As described in **senario 7** (from related issue https://github.com/spring-projects/spring-framework/issues/15220), 
`Part` field could not be bind with values, as `StandardMultipartHttpServletRequest` wraps `Part` into private class `StandardMultipartHttpServletRequest.StandardMultipartFile`. 
In turn, conversion from `StandardMultipartFile` to `Part` had occured but failed.

Test and generated logs are below. I hope available approaches are found for this scenario.
```java
@SpringJUnitWebConfig(classes = {WebConfig.class})
public class FileUploadControllerTests {

    MockMvc mvc;

    FileUploadController controller = new FileUploadController();

    MockPart mockPart = new MockPart("file", "filename.png", "file".getBytes());

    @BeforeEach
    public void setup() {
        this.mvc = MockMvcBuilders.standaloneSetup(controller).build();
    }

    @DisplayName("params + Part -> VO Having Part : 400 Error")
    @Test
    void bindVOHavingPart() throws Exception {
        MultipartHttpServletRequest request = (MultipartHttpServletRequest)
                mvc.perform(multipart("/vopart")
                        .part(mockPart)
                        .param("name", "jaebin-joo")
                        .param("age", "11"))
                    .andExpect(status().is4xxClientError()) // StandardMultipartFile 타입을 Part로 변환할 수 없기 때문에
                    .andReturn().getRequest();

        assertThat(request.getParts().size()).isEqualTo(1);
        assertThat(request.getMultiFileMap().size()).isEqualTo(0);
    }
}
```
```
[...]

PR #27830, opened 2021-12-17
Improve tests and Javadoc on binding to a property of type javax.servlet.Part

binchoo:
When a controller obtains files from multipart requests,
there is an inconsistency when `jakarta.servlet.http.Part` type is an attribute of an object or an argument of a handler method.

If a method has an object argument, and expects files to be saved to attributes of that argument, the attribute type should not be `Part`, but it would rather be `MultipartFile`.

This is because `StandardMultipartHttpServletRequest` wraps its `Part`s, whose filename exists, with `StandardMultipartFile`.
`StandardMultipartFile` is an private class that implements `MultipartFile` class. So `Part` type attributes generate conversion errors.

In contrast, if the method has an argument type of `Part`, it can normally receive multipart files from the framework.

My idea is simply modifying `StandardMultipartFile` class to implement both `MultipartFile` and `Part`, so preventing conversion issues. Likewise I added new `MockStandardMultipartFile` class that replaces `MockMultipartFile` class. 

**Issues:** https://github.com/spring-projects/spring-framework/issues/27819

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockMultipartHttpServletRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockMultipartHttpServletRequestBuilder.java
@@ -19,6 +19,7 @@ package org.springframework.test.web.servlet.request;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.Serializable;
 import java.net.URI;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
@@ -154,7 +155,7 @@ public class MockMultipartHttpServletRequestBuilder extends MockHttpServletReque
 				String filename = part.getSubmittedFileName();
 				InputStream is = part.getInputStream();
 				if (filename != null) {
-					request.addFile(new MockMultipartFile(name, filename, part.getContentType(), is));
+					request.addFile(new MockStandardMultipartFile(part, filename));
 				}
 				else {
 					InputStreamReader reader = new InputStreamReader(is, getCharsetOrDefault(part, defaultCharset));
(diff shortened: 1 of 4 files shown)


rstoyanchev:
In the description, you refer to scenario 7 from #15220, but that issue and those scenarios specifically were addressed with PR #370. There is a test class [WebRequestDataBinderIntegrationTests](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/test/java/org/springframework/web/bind/support/WebRequestDataBinderIntegrationTests.java) for those scenarios. I'm wondering what is the difference between that and what you're running into?

rstoyanchev:
Thanks for elaborating.

We do support `Part` for data binding, but only in fallback mode, when there is no `MultipartResolver` (this is in `ServletRequestDataBinder`). If you don't want to use `MultipartFile` you need to make sure that a `MultipartResolver` is not declared.

binchoo:
@rstoyanchev 
Thanks! Your comment gave me a lot of hints to deep-dive into multipart bindings.

As I reviewed,
- DispatcherServlet::checkMultipart
- ServletRequestDataBinder::bind

when a request is a `MultipartRequest` that has been resolved by `MultipartResolver`, then
- `Part` in this request can be resolved for arguments type of `Part` and `MultipartFile`.
- `Part` in this request can be bound to attributes that are `MultipartFile` only.

So I concluded that `MockMultipartHttpServletRequestBuilder` is doing ok because it is a mock for `MultipartRequest` and follows the rules above.

> If you don't want to use MultipartFile you need to make sure that a MultipartResolver is not declared.

Yes I could test this by registering a multipart resolver that purposely returns false in `isMultiPart()`. 
```java
@Bean
MyMultipartResolver multipartResolver() {
    return new MyMultipartResolver();
}

class MyMultipartResolver extends StandardServletMultipartResolver {
    @Override
    public boolean isMultipart(HttpServletRequest request) {
        return false;
    }
}
```

binchoo:
c3ea427 Added tests covering these behaviors.

> When a request is a MultipartRequest that has been resolved by MultipartResolver, then
> - Part in this request can be resolved for arguments type of Part and MultipartFile.
> - Part in this request can be bound to attributes that are MultipartFile only.

rstoyanchev:
Thanks for the updates. I'll process this. 

Note that if you're in Boot application you can use `spring.servlet.multipart.enabled=false` to turn off the `MultipartResolver` config and then declare a bean of type `javax.servlet.MultipartConfigElement` which enables multipart support at the Servlet container level.

Issue #25768, opened 2020-09-14
WebClient/MockMVC issue with jquery ajax post


guisimon28:
Hello,

i am using htmlunit with spring test in order to test all ihm interface from my web application. It works fine with html form (post and get) and with ajax get but i have a problem with ajax post request.

The controller don't received the request. If i replace post by get the junit test case works fine.

this the html view

```html
<!DOCTYPE html>
<html lang="fr" xmlns="http://www.w3.org/1999/xhtml" 
	xmlns:th="http://www.thymeleaf.org">
	<head>
		<title>Spring Html Unit</title>
        <meta charset="utf-8" />
		<meta name="format-detection" content="telephone=no">
		<meta name="viewport" content="width=device-width, initial-scale=1.0">
		<script src="https://code.jquery.com/jquery-3.5.1.min.js" integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0=" crossorigin="anonymous"></script>
		<script>
			function postTest() {
				$.ajax({
			        type:"post",
					data: {
						'subject' : 'subject test',
						'message' : 'message test'
					},
			        url:"[[@{/test/post}]]",
			        success: function(data, textStatus, jqXHR){
			        	$("#result").text(data);
			        }
			    });
			}		
		</script>
	</head>
	<body>
		<h4 th:text="'Hello to Thymeleaf '+${myParam}">Hello from Thymeleaf</h4>
		
		<button type="button" onClick="postTest()">Test post</button>
		<div>
			<span>result :</span><span id="result"></span>
		</div>
		
	</body>
</html>
```

and the controller

```java
@Controller
public class WelcomeController {

	@GetMapping("/")
	public String init(Model model) {
		model.addAttribute("myParam", "Guillaume");
		return "welcome";
	}

	@PostMapping("/test/post")
	public @ResponseBody String post(@RequestParam String subject, @RequestParam String message, Model model) {
		return subject + " " + message;
	}
}
```

you can also find the complete code on my github https://github.com/guisimon28/spring-test-htmlunit

Can you help me to find if there is some missing configuration or if its a htmlunig bug or pull request ?

Thanks for All

Guillaume

bclozel:
[...]

PR #27832, opened 2021-12-17
Working POST ajax requests sent by HtmlUnit

jcaillabet:
As I understood when I analysed the behaviour of a POST request in a classic web application (by a form or an ajax request), the HttpServletRequest object must have both his **content** and **parameters** fields filled.

In HtmlUnitRequestBuilder.java, when converting the webRequest object into the MockHttpServletRequest the **content** field is filled only for an ajax request, and the **parameters** field are filled only for a form request. It seems there's a confusion about the meaning of the term _parameters_.

My correction is intended to have the **parameters** field filled for both the form and the ajax request when using an UI/UX unit test (as it is done when using a classic application).

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
@@ -25,6 +25,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
@@ -242,6 +243,14 @@ final class HtmlUnitRequestBuilder implements RequestBuilder, Mergeable {
 		if (requestBody == null) {
 			return;
 		}
+
+		Arrays.asList(requestBody.split("&")).stream()
+			.filter(param -> param.contains("="))
+			.forEach( param -> {
+			String[] entry = param.split("=", 2);
+			request.addParameter(urlDecode(entry[0]), urlDecode(entry[1]));
+		});
+
 		request.setContent(requestBody.getBytes(charset));
 	}
 


bclozel:
Is this related to #25758 ?

jcaillabet:
Hello

Yes, it is related to #25768. I work with guisimon28, and he suggested me to analyse the issue.

bclozel:
I think we need to review the entire situation again. Especially if the intent on the HtmlUnit API is a bit clearer now.
In the meantime I'm not sure this or really adresses all the situations (like array params and more).

bclozel:
I've just updated [my repro project](https://github.com/bclozel/htmlunitrepro) and the issue is still there.
I think we'll need to deal with the fact that `WebRequest` is an unreliable API and that we might need to parse the request body on the Spring Framework side. The problem with that approach is that since we have no understanding of the behavior and that this inconsistency won't be solved on the HtmlUnit side, we might run into more issues in the future.

I'll have a deeper look into your PR, but I can't guarantee a target milestone for now. For some reason, HtmlUnit/htmlunit#223 is still opened - so I'm wondering if a PR there would be safer - although I don't understand the link with the fetch API support in the last comment.

thuri:
I don't think the PR is a good solution.
The code uses the requestBody property of the WebRequest, which may not have been filled by htmlunit (which actually is the main problem of HtmlUnit/hmlunit#223 ).

But regardless of that this PR would only work if the body is application/x-www-form-urlencoded and it would change the actual request as parameters that the client sends in the body are pushed to the URL when they are added to the WebRequest#setRequestParameters.

Spring (on the server side) will make params from the url and from the body available to the Controller methods via `@RequestParam` but that doesn't mean that we should change the request the client has made via htmlunit/mockmvc

rstoyanchev:
Indeed, @jcaillabet a key question missing from your analysis, is why does HtmlUnit behave in this way, i.e. exposing what looks like inconsistent behavior for Ajax vs regular form requests? And why should that be then addressed on the Spring side, and potentially in other places that use or build on HtmlUnit?

jcaillabet:
This is a good question that could be maybe aborded in [HtmlUnit/htmlunit#223](https://github.com/HtmlUnit/htmlunit/issues/223).

What I analysed is that HtmlUnit fill a WebRequest object differently if we do an ajax or a regular form request. For an ajax request the requestBody_ property is filled, and for a regular form request the requestParameters_ property will be filled. And the HtmlUnit javadoc explain that only one of then can be filled.

The problem I had is that in Spring we only read the requestParameters_ property, whatever the request type is. However I have tested a correction in HtmlUnit made by @thuri, that is correct for me. So finally I think that a correction in HtmlUnit is conceivable.

PR #27834, opened 2021-12-17
Use String.repeat() and pattern matching instanceof

salatmaster:
Using various new features from Java 9-17 such as `String.repeat()` and pattern matching instanceof.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/AbstractDataFieldMaxValueIncrementer.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/AbstractDataFieldMaxValueIncrementer.java
@@ -133,12 +133,7 @@ public abstract class AbstractDataFieldMaxValueIncrementer implements DataFieldM
 		String s = Long.toString(getNextKey());
 		int len = s.length();
 		if (len < this.paddingLength) {
-			StringBuilder sb = new StringBuilder(this.paddingLength);
-			for (int i = 0; i < this.paddingLength - len; i++) {
-				sb.append('0');
-			}
-			sb.append(s);
-			s = sb.toString();
+			s = "0".repeat(this.paddingLength - len) + s;
 		}
 		return s;
 	}
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/ArgumentTypePreparedStatementSetter.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/ArgumentTypePreparedStatementSetter.java
@@ -61,8 +61,7 @@ public class ArgumentTypePreparedStatementSetter implements PreparedStatementSet
 		if (this.args != null && this.argTypes != null) {
 			for (int i = 0; i < this.args.length; i++) {
 				Object arg = this.args[i];
-				if (arg instanceof Collection && this.argTypes[i] != Types.ARRAY) {
-					Collection<?> entries = (Collection<?>) arg;
+				if (arg instanceof Collection<?> entries && this.argTypes[i] != Types.ARRAY) {
 					for (Object entry : entries) {
 						if (entry instanceof Object[] valueArray) {
 							for (Object argValue : valueArray) {
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/PreparedStatementCreatorFactory.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/PreparedStatementCreatorFactory.java
@@ -265,8 +265,7 @@ public class PreparedStatementCreatorFactory {
 					}
 					declaredParameter = declaredParameters.get(i);
 				}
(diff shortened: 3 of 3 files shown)


spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLErrorCodeSQLExceptionTranslator.java
```diff
@@ -360,38 +360,44 @@ protected DataAccessException createCustomException(
 			// invoke constructor
 			Constructor<?> exceptionConstructor;
 			switch (constructorType) {
```
JarvisCraft, line 362: Guess that should be replaced with single `return switch` with `yield`s inside.

spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/AbstractDataFieldMaxValueIncrementer.java
```diff
@@ -133,12 +133,7 @@ public String nextStringValue() throws DataAccessException {
 		String s = Long.toString(getNextKey());
 		int len = s.length();
 		if (len < this.paddingLength) {
-			StringBuilder sb = new StringBuilder(this.paddingLength);
-			for (int i = 0; i < this.paddingLength - len; i++) {
-				sb.append('0');
-			}
-			sb.append(s);
-			s = sb.toString();
+			s = "0".repeat(this.paddingLength - len) + s;
```
JarvisCraft, line 136: This may be less efficient due to the need for allocation of a temporary `String`.

JarvisCraft (changes requested):

spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/AbstractDataFieldMaxValueIncrementer.java
AbstractCoderX: JMH says otherwise 
изображение

spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLErrorCodeSQLExceptionTranslator.java
AbstractCoderX: done

spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/AbstractDataFieldMaxValueIncrementer.java
JarvisCraft: Consider attaching benchmark's code

sbrannen:
@AbstractCoderX, have you run a full local build?

`./gradlew check`

Just curious, because the last time I checked.... we could not use switch expressions due to a limitation with Checkstyle (specifically [Spring Java Format](https://github.com/spring-io/spring-javaformat/issues/300)).

AbstractCoderX:
> Just curious, because the last time I checked.... we could not use switch expressions due to a limitation with Checkstyle (specifically [Spring Java Format](https://github.com/spring-io/spring-javaformat/issues/300)).

As far as I can see, this has already been fixed, why does the codestyle check fail?

sbrannen:
> > Just curious, because the last time I checked.... we could not use switch expressions due to a limitation with Checkstyle (specifically [Spring Java Format](https://github.com/spring-io/spring-javaformat/issues/300)).
> 
> As far as I can see, this has already been fixed, why does the codestyle check fail?

The fix is in `spring-javaformat` 0.0.30 which has not been released. We are therefore still using `spring-javaformat` 0.0.29.

https://github.com/spring-projects/spring-framework/blob/b24cbc5e9791c87e033d1779b4a1a88f90b1e372/build.gradle#L323

sbrannen:
> The fix is in `spring-javaformat` 0.0.30 which has not been released. We are therefore still using `spring-javaformat` 0.0.29.
> 
> https://github.com/spring-projects/spring-framework/blob/b24cbc5e9791c87e033d1779b4a1a88f90b1e372/build.gradle#L323

In light of that, @AbstractCoderX feel free to move the "switch expression" changes to a separate PR so that we can merge in the other changes independent of that.

AbstractCoderX:
> > The fix is in `spring-javaformat` 0.0.30 which has not been released. We are therefore still using `spring-javaformat` 0.0.29.
> > https://github.com/spring-projects/spring-framework/blob/b24cbc5e9791c87e033d1779b4a1a88f90b1e372/build.gradle#L323
> 
> In light of that, @AbstractCoderX feel free to move the "switch expression" changes to a separate PR so that we can merge in the other changes independent of that.

done

sbrannen:
@AbstractCoderX, are you planning on submitting a PR with the "switch expression" changes?

AbstractCoderX:
@sbrannen https://github.com/spring-projects/spring-framework/pull/28014

PR #27836, opened 2021-12-18
Fix the issue#27683

### LINKED ISSUE #27683: AntPathMatcher with a custom separator doesn't match correctly a specific String
**Affects:** 5.3.10

---

**Background:**

I assume the [AntPathMatcher](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html) matches (among othersú a path against a pattern if all below apply:
- `pattern` and `path` are equal
- they both contain only alphanumeric characters (no Ant-specific characters such as `*`, `?` or `{` with `}` and no slashes `/` and `\`)

**Reproducible sample:**

The following code reproduces a strange behavior:

```
new AntPathMatcher().match("consent", "consent");             // true
new AntPathMatcher().match("consentreg", "consentreg");       // true

new AntPathMatcher("\\t").match("consent", "consent");        // true
new AntPathMatcher("\\t").match("consentreg", "consentreg");  // false
```

Do I assume the behavior wrongly? If so, would you describe the behavior and how to fix my use case?

**Workaround:** 

As of now, I must check whether the `pattern` is not Ant pattern and check the equality first:

```
String string = "consentreg";
String pattern = "consentreg";
AntPathMatcher matcher = new AntPathMatcher("\\t");
boolean result = (!matcher.isPattern(pattern) && pattern.equals(string)) || getAntPathMatcher().match(pattern, string);
// true
```

aoyvx:
fix #27683 .
`pathSeparator` should be a complete separator, But in `StringUtils#tokenizeToStringArray`,each of the characters is individually considered as a delimiter. So I try to replace `StringUtils#tokenizeToStringArray` with `StringUtils#delimitedListToStringArray`.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/BeanFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/BeanFactory.java
@@ -232,7 +232,7 @@ public interface BeanFactory {
 	 * specific type, specify the actual bean type as an argument here and subsequently
 	 * use {@link ObjectProvider#orderedStream()} or its lazy streaming/iteration options.
 	 * <p>Also, generics matching is strict here, as per the Java assignment rules.
-	 * For lenient fallback matching with unchecked semantics (similar to the ´unchecked´
+	 * For lenient fallback matching with unchecked semantics (similar to the `unchecked`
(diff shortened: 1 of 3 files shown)


rstoyanchev:
Closing along with the linked issue.

Issue #23219, opened 2019-07-01
MockMvc no longer handles UTF-8 characters


momega:
I wrote simple controller:

```java
@RestController
public class TestController {

    @GetMapping("/api")
    public Map<String, String> simpleTest() {
        // here is the text with special characters, all in UTF-8
        return Collections.singletonMap("test", "Příliš žluťoučký kůň úpěl ďábelské ódy");
    }
}
```

Also very simple tutorial-like test:

```java
    @Before
    public void setup() {
        this.mockMvc = MockMvcBuilders.webAppContextSetup(this.wac).build();
    }

    @Test
    public void test() throws Exception {
        this.mockMvc.perform(get("/api"))
                .andDo(print())
                .andExpect(jsonPath("$.test").value("Příliš žluťoučký kůň úpěl ďábelské ódy"));
    }
```

The test fails in version 5.2.0.M3 now. Test passes in the previous milestone version 5.2.0.M2 and earlier versions.

The issue is somehow related to deprecation of `org.springframework.http.MediaType#APPLICATION_JSON_UTF8_VALUE`.

Of course it is possible to fix the test by adding _accept_ but I rather do not do that, because `APPLICATION_JSON_UTF8_VALUE` is deprecated now.

```java
.accept(MediaType.APPLICATION_JSON_UTF8_VALUE)
```

I have also prepared simple maven project to simulate the problem.
[springutf8.zip](https://github.com/spring-projects/spring-framework/files/3345177/springutf8.zip)

It seems to me, that `MockMvc` client does not handle UTF-8 characters well, like other modern browsers.

sbrannen:
@sdeleuze, it appears that gh-22788 introduced a regression in `MockMvc`. Do you have time to look into this?

sdeleuze:
@sbrannen @rstoyanchev The situation here is the following.

As expected, the JSON data is written as UTF-8 by Jackson converter/encoder regardless of the `ISO-8859-1` default Servlet encoding which is only used when the `String` serialization is performed by the Servlet container itself.

On `MockMVC` side, by default `MockHttpServletResponse#characterEncoding` is set to `ISO-8859-1` and used for writing and reading `String`. For writing text, I think that's relevant since it will mimic the Servlet Container behavior. But for reading, I think current behavior can lead to surprising results like the one raised in this issue.

In order to get a more predictable behavior, I think my proposal would be to change:

```java
public String getContentAsString() throws UnsupportedEncodingException {
[...]

PR #27846, opened 2021-12-23
Use UTF-8 for application/json in MockHttpServletResponse

izeye:
This PR changes to use UTF-8 for `application/json` in `MockHttpServletResponse`.

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletResponse.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletResponse.java
@@ -24,6 +24,7 @@ import java.io.PrintWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
@@ -314,6 +315,11 @@ public class MockHttpServletResponse implements HttpServletResponse {
 				MediaType mediaType = MediaType.parseMediaType(contentType);
 				if (mediaType.getCharset() != null) {
 					setExplicitCharacterEncoding(mediaType.getCharset().name());
+				}
+				else {
+					if (contentType.equals(MediaType.APPLICATION_JSON_VALUE)) {
+						this.characterEncoding = StandardCharsets.UTF_8.name();
+					}
 				}
 			}
 			catch (Exception ex) {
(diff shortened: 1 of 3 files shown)


sbrannen:
Hi @izeye,

Is there an existing issue related to this change? If not, can you please briefly explain the rationale for the change (even if it seems self explanatory to some)?

Also, please note that you only changed the code in the test fixture. Thus, you would also need to update `org.springframework.mock.web.MockHttpServletResponse`, `MockHttpServletResponseTests`, etc.

izeye:
@sbrannen Thanks for the feedback!

> Is there an existing issue related to this change? If not, can you please briefly explain the rationale for the change (even if it seems self explanatory to some)?

I noticed that `MockMvcResultHandlers.print()` with UTF-8 doesn't work with the `MediaType.APPLICATION_JSON` (`application/json`) while it works with the deprecated `MediaType.APPLICATION_JSON_UTF8` (`application/json;charset=UTF-8`). See https://github.com/izeye/spring-boot-throwaway-branches/blob/minimal-rest/src/test/java/com/izeye/throwaway/HomeControllerTests.java#L67-L70 for an example.

> Also, please note that you only changed the code in the test fixture. Thus, you would also need to update `org.springframework.mock.web.MockHttpServletResponse`, `MockHttpServletResponseTests`, etc.

Sorry. I should have looked into it closely, but it seems that I just searched and picked a wrong one. Thanks for guiding me to the right direction. I updated accordingly. Please let me know if there's anything missing or worng.

sbrannen:
> Sorry. I should have looked into it closely, but it seems that I just searched and picked a wrong one. 

No problem. Even core maintainers (like me) often forget to update _both_ versions of those mocks.

> Thanks for guiding me to the right direction. 

You're welcome.

> I updated accordingly. Please let me know if there's anything missing or worng.

I think that looks pretty good now; however, having looked at the issue in more detail I'm starting to wonder if that is the correct "fix".

As far as I know, we don't set the character encoding for the response to UTF-8 in such scenarios in production code (in `spring-webmvc`). Instead, we let the client correctly interpret the response as UTF-8.

For the particular test scenario you've described, I see the `print()` functionality as the "client" in Mock MVC. 

In other words, I don't think we should make this change to `MockHttpServletResponse.setContentType(String)`. Instead, I think we should make a change to `org.springframework.test.web.servlet.result.PrintingResultHandler.printResponse(MockHttpServletResponse)` to infer UTF-8 if the `Accept` header is set to `application/json`.

@rstoyanchev and @sdeleuze, what are your thoughts on the matter?

rstoyanchev:
There is some history. The corresponding MediaType was deprecated in #22788 and removed from use throughout because it is no longer defined for application/json. Subsequently, a similar change was considered for MockMvc in #23219 but resulted in a more limited change. I do not think we should revive the charset parameter on application/json anywhere but if there is a specific problem to solve somewhere we can try to be more lenient where it is present or needs to be implied.

sbrannen:
Hi @izeye,

Thanks again for the PR.

In the end, we decided to address this differently (see comments above).

I am therefore closing this as superseded by #27926.

PR #27854, opened 2021-12-24
Optimize initialization of HttpMessageConverter's

qxo:
In WebMvcConfigurationSupport will setMessageConverters after
ExceptionHandlerExceptionResolver and RequestMappingHandlerAdapter creation
so there no need init messageConverters in its constructor.


rstoyanchev (changes requested):
This sets a field that is not marked `@Nullable` to `null` and in turn the getter may return `null` even though it's not marked as `@Nullable`. 

A better way to do this would be to mark the `messageConverters` field `@Nullable` and provide default initialization in `afterPropertiesSet`, while the getter could return an empty list while not yet initialized.

Issue #27856, opened 2021-12-25
Conditional judgment can be optimized


1725636955:
https://github.com/spring-projects/spring-framework/blob/544e9bb4036965cc4054fba72f2a5690539e44bf/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassUtils.java#L126-L128

Can be optimized

The following code，Optimal condition judgment，Increase readability
`if (config != null && Boolean.TRUE.equals(config.get("proxyBeanMethods"))) {
			beanDef.setAttribute(CONFIGURATION_CLASS_ATTRIBUTE, CONFIGURATION_CLASS_FULL);
}`

snicoll:
@1725636955 thanks but there's no need to create both an issue and a PR as the latter supersedes the former.

PR #27857, opened 2021-12-25
perf: Conditional Judgment Modification

spirit-unpredictable:
Increase readability，Optimal condition judgment
#27856


rstoyanchev:
Readability is subjective. In this case the flag is true by default, so "not false" means, as long as the default hasn't been changed.

PR #27859, opened 2021-12-27
Make task executors implement ExecutorService or ScheduledExecutorService

qian0817:
Make `ThreadPoolTaskExecutor` implements `ExecutorService` and make `ThreadPoolTaskScheduler` implement `ScheduledExecutorService` in order to make these classes more convenient to use.


snicoll:
@qian0817 Thanks for the PR but I don't think the extra delegation is worth it. `ThreadPoolTaskExecutor` has a `getThreadPoolExecutor` that provides access to the executor. And same for `ThreadPoolTaskScheduler`.

PR #27862, opened 2021-12-28
Remove unnecessary check in isBridgedCandidateFor()

aoyvx:
In BridgeMethodResolver#isBridgedCandidateFor, candidateMethod is never not bridged, so it seems unnecessary to judge whether candidateMethod and bridgeMethod are the same.


rstoyanchev:
Seems like a valid point that if they are the same method, then if `isBridge` is false on one, it's false on both. This might be more about explicitness, what do you think, @jhoeller?

jhoeller:
Good point, in the given context that this internal method is called within, they are never going to be equal indeed if the candidate is not a bridge.

PR #27865, opened 2021-12-28
Add countOccurrencesIgnoreCase method to StringUtils

mipo256:


snicoll:
@Mikhail2048 thanks for the PR but we'd usually add such a method if used itself within the framework (which we aren't as far as I am aware). For that reason,  I am going to close this PR.

PR #27870, opened 2022-01-01
Make `H2SequenceMaxValueIncrementer` compatible with H2 database 2.0.x

hpoettker:
Spring JDBC is currently not compatible with 2.0.x of H2. This creates problems for users that want to upgrade. See e.g.

- https://github.com/spring-cloud/spring-cloud-task/issues/821
- https://github.com/spring-projects/spring-boot/issues/29034

I've adjusted `H2SequenceMaxValueIncrementer` such that it works with both 1.4 and 2.0 and adjusted some test DDL scripts such that they work with both H2 versions and HSQL.

I've also written a test for the incrementer as it is currently seems not to be covered by any test. I was uncertain whether it should be added to  `DataFieldMaxValueIncrementerTests` as that class uses mocks for all data sources instead of embedded databases.

The complete build is successful for the currently used H2 version 1.4.200. With version 2.0.204, the tests in `spring-jdbc` and `spring-test` also pass. However, the tests in `spring-r2dbc` fail with 2.0.204. But that is not related to the changes in the test DDL scripts.


hpoettker:
The build failed with
```
OrderedMessageSendingIntegrationTests > exceedTimeLimit() FAILED
    java.lang.AssertionError at OrderedMessageSendingIntegrationTests.java:171
```
I don't think this is related to my changes.

spring-jdbc/src/main/java/org/springframework/jdbc/support/incrementer/H2SequenceMaxValueIncrementer.java
```diff
@@ -47,7 +47,7 @@ public H2SequenceMaxValueIncrementer(DataSource dataSource, String incrementerNa
 
 	@Override
 	protected String getSequenceQuery() {
-		return "select " + getIncrementerName() + ".nextval from dual";
+		return "select next value for " + getIncrementerName() + " from dual";
```
katzyn, line 50: Please, use standard-compliant `VALUES NEXT VALUE FOR sequenceName`. All other constructions aren't guaranteed to work in all compatibility modes of all versions of H2.

`SELECT … FROM DUAL` is also an Oracle compatibility feature that may be rejected by H2 or interpreted in some unexpected way.

hpoettker: @katzyn Thanks! Changed.

sbrannen:
Hi @hpoettker,

Thanks for submitting your first PR to the Spring Framework, and thanks for the detailed description! 👍 

> I've adjusted `H2SequenceMaxValueIncrementer` such that it works with both 1.4 and 2.0 and adjusted some test DDL scripts such that they work with both H2 versions and HSQL.

Thanks for checking both versions.

> I've also written a test for the incrementer as it is currently seems not to be covered by any test. I was uncertain whether it should be added to `DataFieldMaxValueIncrementerTests` as that class uses mocks for all data sources instead of embedded databases.

It's better to have a dedicated integration test using a real, embedded database for this purpose.

----

The H2 documentation for [Sequence value expression](http://www.h2database.com/html/grammar.html#sequence_value_expression) indeed states that we should be using `NEXT VALUE FOR <sequence>` instead of `<sequence>.nextval from dual`. So thanks for bringing that to our attention, @katzyn!

As mentioned elsewhere, the syntax `<sequence>.nextval from dual` can be used with H2 when using the `Legacy` or `Oracle` [compatibility modes](http://www.h2database.com/html/features.html#compatibility).

sbrannen:
> However, the tests in `spring-r2dbc` fail with 2.0.204. But that is not related to the changes in the test DDL scripts.

This will be addressed in the following PR for `r2dbc-h2`:

- https://github.com/r2dbc/r2dbc-h2/pull/204

So, for the time being we cannot upgrade to H2 2.x in the build.

sbrannen:
This has been merged into `5.3.x` and `main` in ed4e2289cdcc72e9bc8e061f208c4dde721d4d1b and further refined in 18781e58ac8ece4e2d0b8511f047307d5b2677a7 and 75e18114700b8e8cc0b95a3732ffaa2ceedf0a48.

Thanks

sbrannen:
See also:

- #27902

PR #27884, opened 2022-01-04
Replace use of StringBuffer with StringBuilder

xixingya:


sbrannen:
This is effectively a follow up to #27822

sbrannen:

Thanks

Issue #22269, opened 2019-01-16
Use try-with-resources in Spring 5 documentations


pashazadeh:
Spring 5 supports JDK 8+, so it is much better to change the sample codes to Java 7 try-with-resources instead of old fashion style.

For example in [4.4.2. Implementing DAOs Based on JPA: EntityManagerFactory and EntityManager](https://docs.spring.io/spring/docs/5.1.3.RELEASE/spring-framework-reference/data-access.html#orm-jpa-dao) there exists following code

```
EntityManager em = this.emf.createEntityManager();
try {
    Query query = em.createQuery("from Product as p where p.category = ?1");
    query.setParameter(1, category);
    return query.getResultList();
} finally {
    if (em != null) {
        em.close();
    }
}
```
which could be replaced with
```
try(EntityManager em = this.emf.createEntityManager()) {
    Query query = em.createQuery("from Product as p where p.category = ?1");
    query.setParameter(1, category);
    return query.getResultList();
}
```
The same thing applies to OXM samples, and many more.

PR #27886, opened 2022-01-04
Correct not compiling example code in Data Access docs

jjank:
I've noticed, that there is a piece of sample code that does not compile with Spring 5.x, because the the EntityManager interface [does not implement AutoCloseable](https://github.com/eclipse-ee4j/jpa-api/blob/2.2-2.2.3/api/src/main/java/javax/persistence/EntityManager.java#L54) until JPA 3.x ([relevant commit in jpa-api](https://github.com/eclipse-ee4j/jpa-api/commit/1ae01ae3793d129bf921e9b1fe893bed744065d8))

The sample should be fine for Spring 6.x - so I'm not sure if targeting `main` is the correct workflow here?

* Partially reverts 189e1afc6ee6bd40b
* Releates to https://github.com/spring-projects/spring-framework/issues/22269

Thanks for looking at this!

### DIFF
--- a/src/docs/asciidoc/data-access.adoc
+++ b/src/docs/asciidoc/data-access.adoc
@@ -2444,9 +2444,9 @@ the `TransactionOperator` resembles the next example:
 		}
 
 		public Mono<Object> someServiceMethod() {
-			
+
 			// the code in this method runs in a transactional context
-			
+
             Mono<Object> update = updateOperation1();
 
 			return update.then(resultOfUpdateOperation2).as(transactionalOperator::transactional);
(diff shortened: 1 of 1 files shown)


sbrannen:
Good catch!

And congratulations on submitting your first PR for the Spring Framework.

> The sample should be fine for Spring 6.x

We're actually still using JPA 3.0 (not yet on 3.1) on `main`. So the example should also be changed for 6.0 for the time being.

> so I'm not sure if targeting `main` is the correct workflow here?

I'll apply your changes to `5.3.x` and forward merge to `main`.

sbrannen:

Thanks

PR #27888, opened 2022-01-04
Clear cache when clearing argument resolvers

demkom58:
Getting a resolver for a parameter does not work correctly after clearing because the cache is not cleared.


rstoyanchev:
You're correct, although I'm curious how you came across it? The `clear()` method is used only in the case of the non-reactive, messaging variant, and that's from the setter for argument resolvers, which is most often called during initializations, probably before the cache fills.

demkom58:
> You're correct, although I'm curious how you came across it? The `clear()` method is used only in the case of the non-reactive, messaging variant, and that's from the setter for argument resolvers, which is most often called during initializations, probably before the cache fills.

Oh, I was just researching how everything works internally.

PR #27896, opened 2022-01-06
Use try-with-resources where feasible

quaff:
Revert and polishing https://github.com/spring-projects/spring-framework/commit/c04fa858fc3344b8d8a9b01b3b07ce0d00d51676


sbrannen:
Hi @quaff,

I had the same idea but decided against it.

Although your proposal is technically correct, I don't think it warrants a change to the existing code just for the sake of using try-with-resources. In other words, I don't see much gain in readability or maintainability with the change. In any case, try-with-resources is used in that code block on `main` for 6.0+.

PR #27905, opened 2022-01-08
Fix SpEL propertyReadWriteWithRootObject() test

justlikeliuen:
Partially modify PropertyAccessTests#propertyReadWriteWithRootObject.
The test case is for method 'withRootObject' but actually it's copied
for the previous case.


sbrannen:
Good catch. That test method is indeed not testing what it's intended to.

And congratulations on submitting your first PR to the Spring Framework! 👍

sbrannen:

Thanks

As a side note, your fix made me notice a similar issue in another test method in that class, which I fixed in 59c6b7e44543d3fdadfb78ae3c6a7c76de2fce91.

Issue #25310, opened 2020-06-24
Determine why transferTo tests fail for Undertow


poutsma:
With the introduction of the `DefaultMultipartParser`, two tests started failing when running against Undertow:

1. `org.springframework.web.reactive.function.MultipartIntegrationTests#transferTo` and
2. `org.springframework.web.reactive.result.method.annotation.MultipartIntegrationTests#transferTo`.

As temporary workaround the tests were disabled for 5.3 M1, but this needs to be resolved by M2.

gregw:
I'm working on a [Jetty Core](https://github.com/spring-projects/spring-framework/pull/32097/files) integration that is similar to undertow in that it is a fully asynchronous integration without servlets.  We are also failing `MultipartWebClientIntegrationTests#transferTo`, with the data received not being the correct value.   

However, if we copy the bytebuffers as we create the DataBuffer wrappers around the jetty buffer abstraction.  We see no leaks of either DataBuffer references, nor the underlying jetty RetainableByteBuffer.   Thus our current best theory is that this test is accessing the databuffer after it has been released.

So maybe this is a bad test??  How is undertow failing this test?

gregw:
Strangely enough we are also failing `org.springframework.web.reactive.function.MultipartRouterFunctionIntegrationTests#proxy`, which undertow also is failing.  We fail this one with a leak and read after the channel is complete.  Is there an issue for this test as well?

PR #27908, opened 2022-01-09
Simplify debugging of Undertow transferTo issue #25310

JiangTChen:
Since undertowHttpServer conflicts with other HTTPServers. Temporarily bypass with transferToForUndertow. It's better than return and pass.

### DIFF
--- a/spring-webflux/src/test/java/org/springframework/web/reactive/function/MultipartIntegrationTests.java
+++ b/spring-webflux/src/test/java/org/springframework/web/reactive/function/MultipartIntegrationTests.java
@@ -23,6 +23,8 @@ import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.Map;
 
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
 import reactor.test.StepVerifier;
@@ -101,6 +103,16 @@ class MultipartIntegrationTests extends AbstractRouterFunctionIntegrationTests {
 		// TODO Determine why Undertow fails: https://github.com/spring-projects/spring-framework/issues/25310
 		assumeFalse(httpServer instanceof UndertowHttpServer, "Undertow currently fails with transferTo");
 
+		verifyTransferTo(httpServer);
+	}
+
+	@Disabled("Unstable on Undertow,Use @RepeatedTest(100) for verify")
+	@Test
+	void transferToWithUndertow() throws Exception {
+		verifyTransferTo(new UndertowHttpServer());
+	}
+
+	private void verifyTransferTo(HttpServer httpServer) throws Exception {
 		startServer(httpServer);
 
 		Mono<String> result = webClient


sbrannen:
Hi @JiangTChen,

> Since undertowHttpServer conflicts with other HTTPServers. Temporarily bypass with transferToForUndertow.

Can you briefly expound on the first statement and explain how your workaround (the second statement) solves the problem?

sbrannen:
Please note that your PR [fails the build](https://ci.spring.io/builds/175139):

```
MultipartIntegrationTests > transferToWithUndertow() FAILED
    java.lang.AssertionError at MessageFormatter.java:115
```

In light of that, I'm marking this as a draft.

JiangTChen:
@sbrannen Thanks for your review.
I found that the pass rate was not 100%. 
There are three kinds of exceptions. I think the root cause is about "Connection prematurely closed BEFORE response".
It's very tricky, I'm not sure if it's a configuration of AbstractHttpHandlerIntegrationTests#startServer
 or undertow bug.
I will keep digging.
Should we add @ignore on transferToWithUndertow() first, to instead of return pass. 
It can alter others and help to fix.

Thanks

spring-webflux/src/test/java/org/springframework/web/reactive/function/MultipartIntegrationTests.java
```diff
@@ -95,12 +103,29 @@ void parts(HttpServer httpServer) throws Exception {
...
+		);
+	}
+
+	@ParameterizedTest(name = "[{index}] {0}")
+	@MethodSource("httpServersWithoutUndertow")
 	void transferTo(HttpServer httpServer) throws Exception {
 		// TODO: check why Undertow fails
-		if (httpServer instanceof UndertowHttpServer) {
-			return;
-		}
```
sbrannen, line 103: Instead of recreating all of the parameterized test infrastructure, please change the above part that you removed to an assumption that fails when the server is Undertow -- for example, with JUnit Jupiter's `assumeFalse(...)` (if Checkstyle permits that) or AssertJ's `Assumptions` class. For an example, see `org.springframework.http.server.reactive.ZeroCopyIntegrationTests.zeroCopy(HttpServer)`.

Once you're done that, please revert all of the unnecessary changes in this test class.

spring-webflux/src/test/java/org/springframework/web/reactive/function/MultipartIntegrationTests.java
```diff
@@ -95,12 +103,29 @@ void parts(HttpServer httpServer) throws Exception {
...
+		//  UndertowHttpServer is unstable, so independently use transferToWithUndertow to test
+		verifyTransferTo(httpServer);
+	}
+
+	@Disabled("Unstable on Undertow,Use @RepeatedTest(100) for verify")
+	@Test
+	void transferToWithUndertow() throws Exception {
```
sbrannen, line 124: I agree that extracting the test for Undertow to a dedicated method makes it easier to seek a solution for the time being.

But please make sure to shutdown the server at the end of the test method, preferably in a `finally` block.

sbrannen (changes requested):

sbrannen:
> I found that the pass rate was not 100%. You can use @RepeatedTest(100) instead of @test to verify it. There are three kinds of exceptions. I think the root cause is about "Connection prematurely closed BEFORE response". It's very tricky, I'm not sure if it's a configuration of AbstractHttpHandlerIntegrationTests#startServer or undertow bug. I will keep digging. 

Thanks for the explanation.

> Should we add `@Disable` on transferToWithUndertow() first, to instead of return pass. It can alter others and help to fix.

I originally thought you were proposing a fix for Undertow, but since that's not the case, yes adding `@Disabled` to the new test method is a must.

And please do continue investigating the root cause of the Undertow issue. We haven't had the time to do that ourselves, and we'd be grateful for a fix from the community.

JiangTChen:
Hi sbrannen,
Thanks for you advice. I have added assumeTrue in the test, and deleted transferToWithUndertow. Now it looks better.
Thanks

JiangTChen:
Hi @sbrannen,
Please help to review again. 
Thanks.

spring-webflux/src/test/java/org/springframework/web/reactive/function/MultipartIntegrationTests.java
sbrannen: Using `assumeTrue` in a catch-block is not what I had in mind, since that will hide all failures (for all server types), which is unacceptable.

In https://github.com/spring-projects/spring-framework/commit/a247b83cd9c9aefd3c329d493c5ce7cd11d0cdfa you can see what I meant: effectively we want to _abort_ the test only if the current server is Undertow.

Thus, please rebase your work on `main` and limit the changes to the new `transferToWithUndertow()` test method and the extracted `verifyTransferTo(HttpServer)` method (without the try-catch).

`transferToWithUndertow()` will need to be annotated with `@Disabled` for the time being.

In addition, please add `See gh-25310` to your commit message so that it links to #25310.

sbrannen (changes requested):

spring-webflux/src/test/java/org/springframework/web/reactive/function/MultipartIntegrationTests.java
JiangTChen: Hi @sbrannen ,
I found the parent class "AbstractHttpHandlerIntegrationTests" use `@AfterEach` to shutdown the server.  Do you sure I need add `finally` block?
Thanks

sbrannen: You're right. You're still invoking `startServer(httpServer)`, so `stopServer()` will still get invoked. So there's no need for the extra `finally` block.

sbrannen:

Feel free to continue to investigate the cause of the issue and report your findings in #25310.

Thanks

PR #27927, opened 2022-01-12
fix:Spring fails to determine XML is XSD-based if DOCTYPE appears in a comment

### LINKED ISSUE #27915: Spring fails to determine XML is XSD-based if DOCTYPE appears in a comment
**Affects:** 5.3.14 and earlier
---

If you want to load a DTD configuration file, ` <! --` Different positions from ` --> ` will affect the loading of spring.

If my profile is written like this:
```xml

<!--

DOCTYPE -->  <!--  -->
            http://www.springframework.org/schema/beans/spring-beans-2.5.xsd">

</beans>
```

Then, when Spring is used to load the configuration file, an exception will appear. 

This is because there is a bug when determining the profile validation pattern.

In `org.springframework.util.xml.XmlValidationModeDetector`,Spring analyzes programs one line at a time.

However, each analysis focuses only on the first `<!--` in each line. Spring will put `<!--` The previous content is not considered as a comment.
Thus, the situation in the above example is ignored.

I think that can result in `org.springframework.util.xml.XmlValidationModeDetector#consumeCommentTokens(String line)` return to add the following code before:

```java
@Nullable
private String consumeCommentTokens(String line) {
    int indexOfStartComment = line.indexOf(START_COMMENT);
    if (indexOfStartComment == -1 && !line.contains(END_COMMENT)) {
        return line;
    }

    String result = "";
    String currLine = line;
    if (indexOfStartComment >= 0) {
        result = line.substring(0, indexOfStartComment);
        currLine = line.substring(indexOfStartComment);
    }

    while ((currLine = consume(currLine)) != null) {
        if (!this.inComment && !currLine.trim().startsWith(START_COMMENT)) {
            int index = result.indexOf(END_COMMENT);
            if(index != -1) result = result.substring(index + END_COMMENT.length());
            return result + currLine;
        }
    }
    return null;
}
```

It can be removed in front of '-->'.

shooye:
It solves the problem that `XmlValidationModeDetector` cannot determine whether XML is based on XSD due to improper parsing of comments spanning multiple lines in the original XML configuration file.

Fixes #27915

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/xml/XmlValidationModeDetector.java
+++ b/spring-core/src/main/java/org/springframework/util/xml/XmlValidationModeDetector.java
@@ -77,7 +77,8 @@ public class XmlValidationModeDetector {
 	/**
 	 * Indicates whether or not the current parse position is inside an XML comment.
 	 */
-	private boolean inComment;
+	private boolean nextInComment;
+	private boolean nowInComment;
 
 
 	/**
@@ -95,7 +96,7 @@ public class XmlValidationModeDetector {
 			String content;
 			while ((content = reader.readLine()) != null) {
 				content = consumeCommentTokens(content);
-				if (this.inComment || !StringUtils.hasText(content)) {
+				if (this.nowInComment || !StringUtils.hasText(content)) {
 					continue;
 				}
 				if (hasDoctype(content)) {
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @shooye,

We really appreciate all the effort you put in to provide a fix!

In the end, we decided to go with a simpler solution that relies on recursion in order to reuse the existing comment parsing logic. If you're interested in checking it out, you'll find it in https://github.com/spring-projects/spring-framework/commit/4b1b25496bfd72c288c3af07c741183d0d90567a#diff-d2b71d5dbef74357add917ba7662a8f3d376fe9e965f82cd08407587b55bf833.

Thanks again for reporting this set of bugs.

PR #1848, opened 2018-06-03
Prevent instantiation of AnnotatedElementUtils

cheese10yun:
I think. The Util class should not be able to create objects.
So I added the abstract keyword to the AnnotatedElementUtils class

PR #27933, opened 2022-01-14
Ensure utility classes cannot be instantiated.

scruel:
According to pull request #1848, utility classes are noninstantiable, but abstract is not enough to prevent from creating objects by subclasses.
This change modified almost all utility classes, but I do think it is necessary, and the effects should small and limited.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotatedElementUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotatedElementUtils.java
@@ -90,7 +90,11 @@ import org.springframework.util.MultiValueMap;
  * @see AnnotationUtils
  * @see BridgeMethodResolver
  */
-public abstract class AnnotatedElementUtils {
+public class AnnotatedElementUtils {
+
+	private AnnotatedElementUtils() {
+		throw new AssertionError();
+	}
 
 	/**
 	 * Build an adapted {@link AnnotatedElement} for the given annotations,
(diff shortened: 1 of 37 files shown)


scruel:
> @scruel thanks for the PR but we'd rather keep things as they are.

Fine… but it must be a bad decision, and could you explain based on what reason so that you will not change this?

snicoll:
`abstract` is a fine enough signal for us. Yes you can instantiate public utils class still but I don't think this is worth the extra noise.

scruel:
> `abstract` is a fine enough signal for us. Yes you can instantiate public utils classes still but I don't think this is worth the extra noise.

Fine enough signal for you? But could you consider that spring is using by so many people around the world, how can you confirm others will not treat `abstract` as a keyword as it is: you should extend this and create an object?
Maybe you should read *Effective Java, item 4*.
The "extra noise" you mentioned are already exists in this project, you could find out some utils classes have private constructor and most of them are not.

scruel:
@snicoll Fine enough signal for you? But could you consider that spring is using by so many people all around the world, how can you to be sure that others will not treat `abstract` as a keyword as it is: you should extend this and create an object?
I recommend you read Effective Java, item 4.
The "extra noise" you mentioned are already exists in this project, you could find out some utils classes which have the private constructor(most of them does not have), and I do think `abstract` should be treated as the "noise", because we should not use it for preventing to instantiate the object.
Thanks.

scruel:
@snicoll Fine enough signal for you is not acceptable, could you consider that spring is using by so many people all around the world, no one can make sure that others will not treat `abstract` as a keyword as it is: you should extend this and create an object?
By the way, the "extra noise" that you mentioned are already exists in this project, you could find some utils classes which have the private constructor(most of them does not have), and I do think `abstract` should be treated as the "noise", because we should not use it for preventing to instantiate the object.
Also, I recommend you read *Effective Java, item 4*.
Thanks.

PR #27936, opened 2022-01-15
Instrument webmvc

jonatan-ivanov:
Questions/todo items:
- Error handling might be more tricky than this
- Controller class/method tag is missing from the span
- Integrating with Micrometer's new http tag provider
- Consolidating tags (e.g.: we have `method` and `http.method`)
- Add tests
- Do we want to keep LongtaskTimer here?

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
+++ b/spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
@@ -42,7 +42,7 @@ public interface AutoTimer {
 	 * An {@link AutoTimer} implementation that is enabled but applies no additional
 	 * customizations.
 	 */
-	AutoTimer ENABLED = (builder) -> {
+	AutoTimer ENABLED = builder -> {
 	};
 
 	/**
--- a/spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
+++ b/spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
@@ -20,9 +20,9 @@ import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
-import io.micrometer.core.annotation.Timed;
-import io.micrometer.core.instrument.Timer;
-import io.micrometer.core.instrument.Timer.Builder;
+import io.micrometer.api.annotation.Timed;
+import io.micrometer.api.instrument.Timer;
+import io.micrometer.api.instrument.Timer.Builder;
 
 import org.springframework.util.CollectionUtils;
 
--- a/spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
+++ b/spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2021 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
```diff
@@ -0,0 +1,114 @@
...
+ * Strategy that can be used to apply {@link Timer Timers} automatically instead of using
+ * {@link Timed @Timed}.
+ *
+ * @author Tadaya Tsuyukubo
+ * @author Stephane Nicoll
+ * @author Phillip Webb
+ * @since 6.0.0
+ */
+@FunctionalInterface
+public interface AutoTimer {
```
philwebb, line 39: Would this class be better in `io.micrometer`? Other than `CollectionUtils` there's no direct Spring connection.

spring-webmvc/src/main/java/org/springframework/web/servlet/observability/HttpServletRequestWrapper.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * An {@link HttpServerRequest} that wraps/delegates to {@link HttpServletRequest}.
+ *
+ * @author Marcin Grzejszczak
+ * @author Jonatan Ivanov
+ *
+ * @since 6.0.0
+ */
+public final class HttpServletRequestWrapper implements HttpServerRequest {
```
philwebb, line 35: I wonder if this can be package private? It looks like it's not pluggable. Possible also worth renaming since it's specifically for MVC.

philwebb: Couple of related Boot issues if we're redesigning this:
https://github.com/spring-projects/spring-boot/issues/29303
https://github.com/spring-projects/spring-boot/issues/13064

spring-webmvc/src/main/java/org/springframework/web/servlet/observability/HttpServletResponseWrapper.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * A {@link HttpServerResponse} that wraps and delegates to {@link HttpServletResponse}.
+ *
+ * @author Marcin Grzejszczak
+ * @author Jonatan Ivanov
+ *
+ * @since 6.0.0
+ */
+public final class HttpServletResponseWrapper implements HttpServerResponse {
```
philwebb, line 33: As above. I wonder if this can be package private.

spring-webmvc/src/main/java/org/springframework/web/servlet/observability/HttpServletRequestWrapper.java
```diff
@@ -0,0 +1,111 @@
...
+	private HttpServletRequestWrapper(HttpServletRequest request) {
+		this.request = request;
+	}
+
+	/**
+	 * Static factory method to create an instance.
+	 * @param request the request to wrap
+	 * @return an {@link HttpServletRequestWrapper} instance that uses the provided request.
+	 */
+	static HttpServletRequestWrapper wrap(HttpServletRequest request) {
```
philwebb, line 47: Not sure about framework, but boot often keeps static factory methods like this at the bottom of the file.

spring-core/src/main/java/org/springframework/core/observability/AutoTimer.java
jonatan-ivanov: This is something we need to figure out. I think based on what we have right now, it would be but we need to check if we can/want to provide a different disable mechanism (for `Meter`s this mechanism is provided by `MeterFilter`s but it won't work in this use-case).

bclozel:
Superseded by #28880

PR #27941, opened 2022-01-17
Fix regression in BeanPropertyRowMapper regarding underscore name

### LINKED ISSUE #27929: BeanPropertyRowMapper underscoreName prepends redundant underscore when field descriptor begins in uppercase
When pojo contains eMail field and Lombok generates setEMail accessor, then BeanPropertyRowMappers underscoreName is _e_mail, which breaks database integration. Symptom appeared with Spring 5.3 from commit https://github.com/spring-projects/spring-framework/commit/6316a353bb074b118ab71ad89fe60553c4d073c9

mdeinum:
With 6316a35 I introduced a regression for properties starting with an uppercase letter.

This commit fixes that regression including an additional test to cover this case.

Fixes: gh-27929


sbrannen:

Thanks for fixing your own regression! 😄

PR #27943, opened 2022-01-17
Add a devcontainer for VSCode and Github Codespaces

dsyer:
With this change user can start a container locally with
VSCode remote, or open an IDE from Github in the browser.

VSCode doesn't support .aj sources yet, so it's not a
completely smooth experience, but you can make changes and
build from the command line.


snicoll:
Thanks for the PR @dsyer but we're not confident in our ability to maintain this in the long run. We don't use a cloud setup on a day-to-day basis, so only newcomers are likely to find out about issues.

PR #27948, opened 2022-01-18
`MissingServletRequestPartException` extends `ServletRequestBindingException`

### LINKED ISSUE #27910: `MissingServletRequestPartException` should extend `ServletRequestBindingException` directly
Unlike similar exceptions, [`MissingServletRequestPartException`](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/multipart/support/MissingServletRequestPartException.java) directly extends `ServletException`; whereas, the following extend `MissingRequestValueException` which extends `ServletRequestBindingException` which indirectly extends `ServletException`.

- `MissingServletRequestParameterException`
- `MissingRequestHeaderException`
- ...

How about changing it if there is no good reason? 🤔

heowc:
Closes gh-27910


rstoyanchev:
Thanks for the suggestion and pull request.

heowc:
Thanks for the quick check 😁

PR #27953, opened 2022-01-19
Adds customization option for Executors coming from AsyncConfigurers

marcingrzejszczak:
Related to https://github.com/spring-projects/spring-boot/issues/29151
Required by https://github.com/spring-cloud/spring-cloud-sleuth/issues/2100

cc @snicoll @wilkinsona

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/annotation/AbstractAsyncConfiguration.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/annotation/AbstractAsyncConfiguration.java
@@ -70,7 +70,7 @@ public abstract class AbstractAsyncConfiguration implements ImportAware {
 	 * Collect any {@link AsyncConfigurer} beans through autowiring.
 	 */
 	@Autowired
-	void setConfigurers(ObjectProvider<AsyncConfigurer> configurers, ObjectProvider<AsyncConfigurerExecutorCustomizer> customizers) {
+	void setConfigurers(ObjectProvider<AsyncConfigurer> configurers, ObjectProvider<AsyncConfigurerExecutorCustomizer> customizer) {
 		Supplier<AsyncConfigurer> configurer = SingletonSupplier.of(() -> {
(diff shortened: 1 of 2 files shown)


spring-context/src/main/java/org/springframework/scheduling/annotation/AbstractAsyncConfiguration.java
```diff
@@ -81,7 +81,8 @@ void setConfigurers(ObjectProvider<AsyncConfigurer> configurers) {
 			}
 			return candidates.get(0);
 		});
-		this.executor = adapt(configurer, AsyncConfigurer::getAsyncExecutor);
+		Supplier<Executor> adaptedExecutor = adapt(configurer, AsyncConfigurer::getAsyncExecutor);
+		customizers.ifAvailable(customizer -> this.executor = () -> customizer.customize(adaptedExecutor.get()));
```
nkonev, line 85: `1. this.executor` won't be set if customizers aren't available. Is this desired ?
2. customizers has plural form but in fact it's a single customizer, not a collection

marcingrzejszczak: Nope, good catch

nkonev (approved):

maradanasai:
This is happening for us when we are migrating to latest. Here is the sample reproducible code https://github.com/maradanasai/otel-async-exper.git . Using spring boot 2.7.14 and spring cloud 2021.0.8 version with jdk 17. It is blocking migrating to latest versions. Please fix this

snicoll:
Sorry for the delay. Unfortunately, we can't add such an interface in 5.3.x at this time, in particular as it is not required in the next generation of micrometer tracing and the framework. I think Sleuth should rather reintroduce the BPP that used to swap the executor, or document how it can be done. This also brings more consistency with what we have to do in the next major release.

PR #27960, opened 2022-01-20
Modify status code 103 to ("Early Hints")

waisink:
According to rfc8297(https://tools.ietf.org/html/rfc8297), the reason
phase of status code 103 should be "Early Hints", not "Checkpoint".

This commit modifies the status code 103 reason phase from ("Checkpoint")
to ("Early Hints").

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/result/StatusResultMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/result/StatusResultMatchers.java
@@ -145,9 +145,9 @@ public class StatusResultMatchers {
 	}
 
 	/**
-	 * Assert the response status code is {@code HttpStatus.CHECKPOINT} (103).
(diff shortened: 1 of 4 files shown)


sbrannen (changes requested):
We unfortunately cannot accept the current proposal since it introduces breaking changes -- by changing the names of existing methods and constants.

The correct way to achieve this would be to introduce the new constants and methods and simultaneously deprecate the existing ones (pointing to the new variants).

waisink:
Thank you for replying and suggesting.

However i am wondering that if preserve the old constants, is that ok when use old constants value to call the method `HttpStatus.valueOf` as the following expression return **false**

`HttpStatus.CHECKPOINT.equals(HttpStatus.valueOf(HttpStatus.CHECKPOINT.value()))`

simonbasle:
@waisink I implemented a change with the same goal as yours but going down the deprecation road in #29816, with an additional method to compare arbitrary status codes' value

PR #27967, opened 2022-01-21
Fix CaffeineCacheManager configuration in the documentation

arey:
With Spring Framework 6.0.0-M2, the `caches` property is not valid and may be replaced by `cacheNames`


snicoll:
Thanks for the PR.

> With Spring Framework 6.0.0-M2, the caches property is not valid

As far as I can see, it never was.

PR #27984, opened 2022-01-28
Fix ServletUriComponentsBuilder examples in ref docs

wkwkhautbois:
Fix `ServletUriComponentsBuilder` examples in reference docs.

Fixed two problems.

- Type mismatch
- Comments are inaccurate

before

```java
// Re-uses host, port and context path...

ServletUriComponentsBuilder ucb = ServletUriComponentsBuilder.fromContextPath(request)
        .path("/accounts").build()
```

after

```java
// Re-uses host, port, scheme and context path...

URI uri = ServletUriComponentsBuilder.fromContextPath(request)
        .path("/accounts")
        .build()
        .toUri();
```

### DIFF
--- a/src/docs/asciidoc/web/webmvc.adoc
+++ b/src/docs/asciidoc/web/webmvc.adoc
@@ -4104,10 +4104,9 @@ as the following example shows:
 
 	// Re-uses host, scheme, port, path and query string...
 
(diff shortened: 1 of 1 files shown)


sbrannen:
Good catch!

Thanks for making your first contribution to the Spring Framework. 👍

PR #27993, opened 2022-02-01
Add support for strict JSON comparison in WebTestClient

elgleidson:
Support for `expectedBody().json(<expected strict json>, true);`

Part of the code base already has support - via `JsonExpectationsHelper` - but `WebTestClient` has no method to use the strict version. The existing method uses `lenient`, which is not always desired.

Example: 
Let's say `/persons/1/` returns the following response:

```json
{
  "name":"John",
  "surname":"Smith"
}
```

The following test works fine:

```java
webTestClient.get()
    .uri("/persons/1")
    .accept(MediaType.APPLICATION_JSON)
    .exchange()
    .expectStatus().isOk()
    .expectBody().json("{\"name\":\"John\"}");
```

I'm just adding a new method so that we can use strict comparison when appropriate.

Example:

This test fails because the response has more fields than expected and we're using strict mode: 

```java
webTestClient.get()
    .uri("/persons/1")
    .accept(MediaType.APPLICATION_JSON)
    .exchange()
    .expectStatus().isOk()
    .expectBody().json("{\"name\":\"John\"}", true);
```

But the following test passes because we're expecting all fields:

```java
webTestClient.get()
    .uri("/persons/1")
    .accept(MediaType.APPLICATION_JSON)
    .exchange()
    .expectStatus().isOk()
    .expectBody().json("{\"name\":\"John\",\"surname\":\"Smith\"}", true);
```

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/reactive/server/DefaultWebTestClient.java
+++ b/spring-test/src/main/java/org/springframework/test/web/reactive/server/DefaultWebTestClient.java
@@ -659,10 +659,10 @@ class DefaultWebTestClient implements WebTestClient {
 		}
 
 		@Override
-		public BodyContentSpec json(String json) {
+		public BodyContentSpec json(String json, boolean strict) {
 			this.result.assertWithDiagnostics(() -> {
 				try {
-					new JsonExpectationsHelper().assertJsonEqual(json, getBodyAsString());
+					new JsonExpectationsHelper().assertJsonEqual(json, getBodyAsString(), strict);
 				}
 				catch (Exception ex) {
 					throw new AssertionError("JSON parsing error", ex);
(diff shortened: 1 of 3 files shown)


sbrannen:
It looks like the code examples for `The following test works fine` and `This test fails` are identical. Can you please update this PR's description?

elgleidson:
> It looks like the code examples for `The following test works fine` and `This test fails` are identical. Can you please update this PR's description?

Done.

sbrannen:
This has been merged into `5.3.x` and `main` in 920be8e1b201fa947b07941eeace62ecf0f431a9 and slightly revised in eb84c843736aaa555d9078a544280bd70b714fbb.

Thanks for making your first contribution to the Spring Framework! 👍

PR #27994, opened 2022-02-01
Improve SimpleJdbcCall for Oracle database

loiclefevre:
Improves SimpleJdbcCall when targeting the Oracle database:

- Improve support of Oracle PL/SQL packages for Spring JDBC: Oracle
database stored procedures or functions can be overloaded. Hence,
the possibility to find several matching signatures by just using
the schema name, catalog name and procedure/function name is real.
This PR adds columns data type checking to properly find the desired
stored procedure or function.

- Ease debugging for DBMS_ packages if SYS not specified: explicitly
checks the catalog name being used and if starting with DBMS_% and no
matching signature was found then report in the log it could be
useful to use SYS as the schema name

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/CallMetaDataContext.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/CallMetaDataContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


snicoll:
@loiclefevre please refrain from opening duplicate PRs. If you need to amend your existing PR, pushing to an existing branch will update your existing PR. As you've created the PR from main, it might close itself if you upgrade the remote but you can reopen the PR and push force it.

loiclefevre:
Hello, do you have any news regarding this PR? Thanks

snicoll:
I am afraid not. The team is busy on other themes lately and we didn't get a chance to look at it yet.

snicoll:
Unfortunately, I don't think we can consider this PR due to the amount of breaking change it brings. Would you be willing to provide an update that doesn't break existing signatures?

PR #27997, opened 2022-02-03
Polish codes in spring-webmvc

1chz:
This PR polishes trivial things in `DispatcherServlet` in spring-webmvc.


rstoyanchev:
Thanks for the suggestion but we'll keep the code as it is.

Issue #27971, opened 2022-01-24
Spring AOP cannot generate proxy for lambda on Java 16+


Ferioney:
I'm migrating to Java 17 and found an issue with proxy generation for 'lambda beans'.

- Spring Boot: 2.6.3
- Maven: 3.8.4
- Java:
```
openjdk version "17.0.2" 2022-01-18 LTS
OpenJDK Runtime Environment Corretto-17.0.2.8.1 (build 17.0.2+8-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.2.8.1 (build 17.0.2+8-LTS, mixed mode, sharing)
```

Spring AOP cannot generate a proxy for beans like:

```java
    @Bean
    public Supplier<String> stringSupplier() {
        return () -> "lambda supplier value";
    }
```

Errors:
```
Caused by: org.springframework.aop.framework.AopConfigException: 
Could not generate CGLIB subclass of class com.example.issue.aopissue.service.IssueConfiguration$$Lambda$638/0x0000000800e796e0: 
Common causes of this problem include using a final class or a non-visible class; 
nested exception is org.springframework.cglib.core.CodeGenerationException: java.lang.reflect.InaccessibleObjectException--
>Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain) 
throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @5ecddf8f

```

But if I use class implementation it works fine:

```java
public class TestSupplierService implements Supplier<String> {

    @Override
    public String get() {
        return "class supplier value";
    }
}
```

I prepared demo project [aop-issue](https://github.com/Ferioney/aop-issue) to show the issue:

- [Application Configuration](https://github.com/Ferioney/aop-issue/blob/main/src/main/java/com/example/issue/aopissue/service/IssueConfiguration.java)
- [Aspect Configuration](https://github.com/Ferioney/aop-issue/blob/main/src/main/java/com/example/issue/aopissue/aspect/AspectConfig.java)
- [Test with lambda bean, context doesn't start](https://github.com/Ferioney/aop-issue/blob/main/src/test/java/com/example/issue/aopissue/AopLambdaImplementedTests.java)
- [Test with clas implementation, works as expected](https://github.com/Ferioney/aop-issue/blob/main/src/test/java/com/example/issue/aopissue/AopClassImplementedTests.java)

[...]

PR #28010, opened 2022-02-05
Fix Javadoc since for AopProxyUtils.isLambda()

izeye:
This PR fixes Javadoc `@since` tag for `AopProxyUtils.isLambda()`.

See gh-27971


snicoll:
Good catch, thanks again @izeye.

sbrannen:
Indeed... thanks for catching AND fixing my typo. 👍

Issue #22609, opened 2019-03-17
Prevent @Bean method overloading by default (avoiding accidental overloading and condition mismatches)


rfelgent:
Hello poeple,

I lost some hours configuring a bean with same id but different properties in java config, as the raised error message was not very helpful.

The bean in question must be created, as it is required via declarative `@DependsOn` configuration.

The error

```
A component required a bean named 'postgresqlContainer' that could not be found.
	- Bean method 'postgresqlContainer' in 'PersistenceConfig.DbServersConfig' not loaded because @ConditionalOnProperty (app.persistence.servers.postgresql.enabled=true) found different value in property 'enabled'
	- Bean method 'postgresqlContainer' in 'PersistenceConfig.DbServersConfig' not loaded because @ConditionalOnProperty (app.persistence.servers.postgresql.enabled=true) found different value in property 'enabled'
```

This config fails:

```java
@Configuration
public class PersistenceConfig {

  @Configuration
  public class DbServersConfig {

    @Bean(value = "postgresqlContainer", initMethod = "start", destroyMethod = "stop")
    @ConditionalOnProperty(prefix = "app.persistence.servers.postgresql", name = "enabled", havingValue = "true")
    public PostgreSQLContainer postgresqlContainer(TCPostgresqlProperties postgresqlProperties) {
        Instance retVal = <do_your_logic>
        return retVal;
    }

    @Bean("postgresqlContainer")
    @ConditionalOnProperty(prefix = "app.persistence.servers.postgresql", name = "enabled", havingValue = "false", matchIfMissing = true)
    public PostgreSQLContainer postgresqlContainer() {
      return null;
    }
  }
}
```

This config works:

```java
@Configuration
public class PersistenceConfig {

  @Configuration
  public class DbServersConfig {

    @Bean(value = "postgresqlContainer", initMethod = "start", destroyMethod = "stop")
    @ConditionalOnProperty(prefix = "app.persistence.servers.postgresql", name = "enabled", havingValue = "true")
    public PostgreSQLContainer  postgresqlContainer(TCPostgresqlProperties postgresqlProperties) {
        Instance retVal = <do_your_logic>
        return retVal;
    }

    @Bean("postgresqlContainer")
    @ConditionalOnProperty(prefix = "app.persistence.servers.postgresql", name = "enabled", havingValue = "false", matchIfMissing = true)
    public PostgreSQLContainer postgresqlContainer2() {
      return null;
    }
  }
}
```

[...]

PR #28019, opened 2022-02-08
Enhance conditional bean creation when using method overloading

Hakky54:
This pull request closes the issue mentioned here: https://github.com/spring-projects/spring-framework/issues/22609

Short summary:
A bean won't be created when it is being skipped by a condition for example when using `@ConditionalOnProperty` or `@ConditionalOnExpression` because the condition is evaluated to false. However a second bean with exactly the same name with different parameters may evaluate to true because the condition has met, so this bean should be created but that does not happen.

An example code snippet:
```java
class Config {

    @Bean
    @ConditionalOnExpression("${some-property-enabled} == true and ${some-different-property} == false")
    public ExampleBean baz() {
        return new ExampleBean();
    }

    @Bean
    @ConditionalOnExpression("${some-property-enabled} == false and ${some-different-property} == true")
    public ExampleBean baz(Object foo) {
        return new ExampleBean();
    }

}
```

When `some-property-enabled` is false and some-different-property is true the second method should be called and the bean should be created, however this does not happen. The bean name is already in the to be skipped beans list.

The pull request contains the changes to reconsider a skipped bean when it encounters a new bean definition having a different condition. If it should not be skipped it will be created. The PR also takes into inheritance into consideration (for backwards compatibility, see https://github.com/spring-projects/spring-framework/blob/c263cbfbe486d7712b9e46b80ed151c1eda98c74/spring-context/src/test/java/org/springframework/context/annotation/ConfigurationClassWithConditionTests.java#L139)

In that specific use case the parent class defines a bean which should be created, however it is getting overridden by a subclass which defines a condition on the bean, so it will be created conditionally and may not even be created when it evaluates to false.

This PR uses a wrapper for the MethodMetadata as a helper class for resolving the inheritance with the condition on a subclass when it still needs to be skipped.

### My use case
In my use case I wanted to refactor my code base. I have the bean configuration below:

[github/mutual-tls-ssl/SSLConfig.java](https://github.com/Hakky54/mutual-tls-ssl/blob/ca34a84fc4d6cd62fc675ec148e52d1fedd800ff/client/src/main/java/nl/altindag/client/SSLConfig.java#L24)

```java
@Component
public class SSLConfig {

    @Bean
    @Scope("prototype")
    public SSLFactory sslFactory(
            @Value("${client.ssl.one-way-authentication-enabled:false}") boolean oneWayAuthenticationEnabled,
            @Value("${client.ssl.two-way-authentication-enabled:false}") boolean twoWayAuthenticationEnabled,
            @Value("${client.ssl.key-store:}") String keyStorePath,
            @Value("${client.ssl.key-store-password:}") char[] keyStorePassword,
            @Value("${client.ssl.trust-store:}") String trustStorePath,
            @Value("${client.ssl.trust-store-password:}") char[] trustStorePassword) {
        SSLFactory sslFactory = null;

        if (oneWayAuthenticationEnabled) {
            sslFactory = SSLFactory.builder()
                    .withTrustMaterial(trustStorePath, trustStorePassword)
                    .withProtocols("TLSv1.3")
                    .build();
        }

        if (twoWayAuthenticationEnabled) {
            sslFactory = SSLFactory.builder()
                    .withIdentityMaterial(keyStorePath, keyStorePassword)
                    .withTrustMaterial(trustStorePath, trustStorePassword)
                    .withProtocols("TLSv1.3")
                    .build();
        }

        return sslFactory;
    }

}
```

And I wanted to refactor it to the following snippet:
```java
@Component
public class SSLConfig {

    @Bean
    @Scope("prototype")
    @ConditionalOnExpression("${client.ssl.one-way-authentication-enabled} == true and ${client.ssl.two-way-authentication-enabled} == false")
    public SSLFactory sslFactory(@Value("${client.ssl.trust-store:}") String trustStorePath,
                                 @Value("${client.ssl.trust-store-password:}") char[] trustStorePassword) {

        return SSLFactory.builder()
                .withTrustMaterial(trustStorePath, trustStorePassword)
                .withProtocols("TLSv1.3")
                .build();
    }


    @Bean
    @Scope("prototype")
    @ConditionalOnExpression("${client.ssl.two-way-authentication-enabled} == true and ${client.ssl.one-way-authentication-enabled} == false")
    public SSLFactory sslFactory(@Value("${client.ssl.key-store:}") String keyStorePath,
                                 @Value("${client.ssl.key-store-password:}") char[] keyStorePassword,
                                 @Value("${client.ssl.trust-store:}") String trustStorePath,
                                 @Value("${client.ssl.trust-store-password:}") char[] trustStorePassword) {

        return SSLFactory.builder()
                .withIdentityMaterial(keyStorePath, keyStorePassword)
                .withTrustMaterial(trustStorePath, trustStorePassword)
                .withProtocols("TLSv1.3")
                .build();
    }

}
```

However it fails. But with the code changes of the Pull request it will create one of the beans when the conditions is true.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClass.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClass.java
@@ -16,7 +16,7 @@
 
 package org.springframework.context.annotation;
 
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
@@ -65,8 +65,7 @@ final class ConfigurationClass {
 	private final Map<ImportBeanDefinitionRegistrar, AnnotationMetadata> importBeanDefinitionRegistrars =
 			new LinkedHashMap<>();
 
-	final Set<String> skippedBeanMethods = new HashSet<>();
-
+	final Map<String, MethodMetadataWrapper> skippedBeanMethods = new HashMap<>();
 
 	/**
 	 * Create a new {@link ConfigurationClass} with the given name.
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassBeanDefinitionReader.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassBeanDefinitionReader.java
@@ -69,6 +69,7 @@ import org.springframework.util.StringUtils;
  * @author Phillip Webb
  * @author Sam Brannen
  * @author Sebastien Deleuze
+ * @author Hakan Altindag
  * @since 3.0
  * @see ConfigurationClassParser
  */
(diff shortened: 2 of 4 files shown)


jhoeller:
Thanks for the PR and the use cases raised, however, we intend to proceed in a different direction and disallow `@Bean` method overloading completely. See https://github.com/spring-projects/spring-framework/issues/22609#issuecomment-1040354124 for an extensive explanation.

PR #28037, opened 2022-02-11
Adds observation support for R2DBC

marcingrzejszczak:
# What is it about?

Adds observation support for R2DBC.

- adds micrometer BOM
- adds micrometer tracing BOM
- adds micrometer & micrometer tracing optional deps to R2DBC module
- adds tracing support for R2DBC (metrics come out of the box)

# To discuss

- I have left a commented out section in the listener where in Sleuth we needed to check if the context is ready. The listener might be called early in the lifecycle of Spring and certain beans might not have been initialized (micrometer or tracing related ones)
- I would like to add the automated asciidoctor documentation generation from sources (check [this project out](https://github.com/micrometer-metrics/micrometer-docs))
- I've a test annotated as `@Disabled` that is typically used for manual assertion of spans in Zipkin or Wavefront. Do we want to leave it there in the code or are there some dedicated samples that we can reuse for that?

cc @mp911de

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,8 +39,8 @@ configure(allprojects) { project ->
 			mavenBom "org.junit:junit-bom:5.8.2"
 
 			// observability
-			mavenBom "io.micrometer:micrometer-bom:2.0.0-SNAPSHOT"
-			mavenBom "io.micrometer:micrometer-tracing-bom:1.0.0-SNAPSHOT"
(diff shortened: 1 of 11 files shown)


bclozel:
I think this PR has been superseded by [the r2dbc-proxy project](http://r2dbc.io/r2dbc-proxy/docs/current-snapshot/docs/html/) which provides metrics, tracing and more.

PR #28041, opened 2022-02-11
Support "*" catch-all mappings for MockMvc Filter registrations

### LINKED ISSUE #28017: `PatternMappingFilterProxy#addUrlPattern` not support `*` url pattern
**Affects:** 5.4.15

---
Does not work in MockMvc when `FilterRegistrationBean` url pattern is `*`.

See [demo](https://github.com/Deycoesr/SPR28017)

funky-eyes:
I believe most business developers think * is equivalent to /*

fixes #28017

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/setup/PatternMappingFilterProxy.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/setup/PatternMappingFilterProxy.java
@@ -41,9 +41,11 @@ import org.springframework.web.util.UrlPathHelper;
  */
 final class PatternMappingFilterProxy implements Filter {
 
-	private static final String EXTENSION_MAPPING_PATTERN = "*.";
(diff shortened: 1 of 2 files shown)


bclozel:
The intent behind #28017 so far is documentation only. This PR changes the behavior and we didn't come to that conclusion yet. Would you like to amend this PR address the original issue?

a364176773:
> 

Could you please tell me exactly how to do that?

a364176773:
> The intent behind #28017 so far is documentation only. This PR changes the behavior and we didn't come to that conclusion yet. Would you like to amend this PR address the original issue?

I think the #28017 solution is more than just documentation, it should be supported at the code level

a364176773:
@bclozel When can you review the PR, please?

bclozel:
Thanks @funky-eyes for your contribution, this is now merged.

Issue #28038, opened 2022-02-11
CronExpression fails to calculate next execution on the day of daylight saving time


guilroux:
Affects: Spring Framework 5.3.15

Looks like the fix for #26744 doesn't cover the following case.

Problem: `CronExpression` fails to calculate properly next execution when running on the day of daylight saving time, just before DST is applied.

```java
ZonedDateTime current = ZonedDateTime.of(LocalDateTime.parse("2021-03-28T01:00:00"), ZoneId.of("Europe/Amsterdam"));
CronExpression cronExpression = CronExpression.parse("0 5 0 * * *");
ZonedDateTime next = cronExpression.next(current);
```

In this case the value returned by `cronExpression.next(current)` is `2021-03-30T00:05` instead of the expected `2021-03-29T00:05`.

To be more precise any input date between `2021-03-28T00:05+01:00` and `2021-03-28T01:59+01:00` will calculate next execution to `2021-03-30T00:05` instead of the expected `2021-03-29T00:05`.

After DST, next execution is successfully calculated i.e. `2021-03-29T00:05`.

poutsma:
Thanks for the PR,  @vikeychen

PR #28044, opened 2022-02-12
Fix CronExpression fails to calculate next execution on the day of daylight saving time

vikeychen:
Hello!

This commit fixes an issue with CronExpression fails to calculate next execution on the day of daylight saving time.

See gh-28038


poutsma:
Merged in 7276752e7cbc87aead678ae84ec8a83e7603a141, thanks for submitting a PR!

Issue #27954, opened 2022-01-19
Deprecate SpringFactoriesLoader#loadFactoryNames


bclozel:
As seen in #27753, the `SpringFactoriesLoader#loadFactoryNames` (introduced in #15158 ) has been very useful over the years but is too permissive for our AOT and native work. Some existing use cases can use reflection in a way that's not predictable for AOT processors.

Here are three example use cases in Spring Boot:
* the [ConfigDataLoaders](https://github.com/spring-projects/spring-boot/blob/1e4e4f168d3956177d8dc81bdaba964bc54cf1ae/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataLoaders.java) are loading `ConfigDataLoader` class names and are calling non-default constructors using a set of dependencies maintained locally
* the [ImportAutoConfigurationImportSelector](https://github.com/spring-projects/spring-boot/blob/1e4e4f168d3956177d8dc81bdaba964bc54cf1ae/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ImportAutoConfigurationImportSelector.java) is loading auto-configuration class names.
* the [FailureAnalyzers](https://github.com/spring-projects/spring-boot/blob/197203968aaa579e5fa7dac73b90a826fe7c6908/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/diagnostics/FailureAnalyzers.java#L83) is using class names to create objects directly and continue in failure scenarios.

Instead of Strings, the replacement should return a different type that allows for the actual type inspection and a Supplier to get an actual instance. Maybe something similar to `ServiceLoader.Provider`. This proposal will probably not solve the two use cases mentioned above, so we'll need to find proper replacements in Spring Boot or provide new features in Spring Framework to support them.

snicoll:
Actually the related issue doesn't deprecate the method so let's reopen this one.

snicoll:
Moving to the next milestone since  we just introduce the infrastructure and we'd need to give some time for team to use it before deprecating `loadFactoryByNames`.

snicoll:
"Blocked" by #29699

snicoll:
"Blocked" by #29699

sbrannen:
If we deprecate `loadFactoryNames()`, we should keep in mind that it is still actively used in `spring-test` in `AbstractTestContextBootstrapper.getDefaultTestExecutionListenerClassNames()`.

[...]

PR #28057, opened 2022-02-16
Improve SpringFactories to customize how arguments and instantiation failures are handled

philwebb:
Two commits that lay the groundwork for deprecating `loadFactoryNames`.

See #27954


snicoll:
For the record the AOT work in the core container may also use this. In Spring Native, the components that take part of the processing of the `ApplicationContext` are discovered in `spring.factories`. It was convenient to provide support for the *Aware interface but if we can replace this by a list of well-defined types, that would be more idiomatic as these components are not beans.

PR #28075, opened 2022-02-18
Deprecate SerializationUtils#deserialize

ledoyen:
As based on Java's serialization mechanism, it can be the source of Remote Code Execution vulnerabilities.

Today this utility is part of the `public` API and can be naively used to _convert_ from object to text and vice versa.
However a naive use can lead to RCE vulnerability if user-input data (like files, cookies, etc.) is transfered using this utility.

I think it should be nice to at least warn the user about the use of this tool (with `@Deprecated`) and later on remove it totally from the public API as this sole use in Spring code is to clone exceptions in `org.springframework.cache.jcache.interceptor.CacheResultInterceptor`.

I am not sure on how it can (or should) be handled.
Let me know if you need me to adapt the code of this PR.

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/jcache/interceptor/CacheResultInterceptor.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/jcache/interceptor/CacheResultInterceptor.java
@@ -150,7 +150,7 @@ class CacheResultInterceptor extends AbstractKeyCacheInterceptor<CacheResultOper
 	@Nullable
 	private static <T extends Throwable> T cloneException(T exception) {
 		try {
-			return (T) SerializationUtils.deserialize(SerializationUtils.serialize(exception));
+			return SerializationUtils.clone(exception);
 		}
 		catch (Exception ex) {
 			return null;  // exception parameter cannot be cloned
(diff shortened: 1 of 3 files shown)


ttddyy:
I recently fixed a code using `SerializationUtils#deserialize` because it was flagged by a penetration test.
So, it would be nice to have such a notion, then users would be aware of the implication of using the method.

sbrannen:
This has been merged into `main` in 7f7fb58dd0dae86d22268a4b59ac7c72a6c22529 and polished in c8d0146bccf6c9089aed020fb629d3a1f36aea60.

The "warning" without official deprecation has also been backported to `5.3.x` (see #28246).

Thanks

Tomator01:
when report  this cve?

ledoyen:
@Tomator01 This is not a CVE per se.

Using this tool to handle user input data can lead to a CVE.
However using it internally as `CacheResultInterceptor` was will not result in a CVE.

sbrannen:
This is not a CVE in the core Spring Framework.

The purpose of this change is to inform anyone who had previously been using `SerializationUtils#deserialize` that it is dangerous to deserialize objects from untrusted sources.

The core Spring Framework does not use `SerializationUtils` to deserialize objects from untrusted sources.

If you believe you have discovered a security issue, please report it responsibly with the dedicated page: https://spring.io/security-policy

And please refrain from posting any additional comments to this commit.

Thank you

PR #28084, opened 2022-02-20
Introduce DurationEditor for java.time.Duration support

dfa1:
Recently, in a project that is using spring-framework (no spring-boot), I saw a `@Configuration` class with 
a lot of configuration about timeouts injected as plain `long` (most likely milliseconds but it is not explicitly documented, as often it happens):

```
@Value("${connect.timeout}")
long connectTimeout;
```

With this small patch, it would be possible to directly inject a `Duration`:

```
@Value("${connect.timeout}")
Duration connectTimeout;
```

In my opinion, this is less error prone, as usually these configuration values must be adjusted with some calculations (i.e. seconds to milliseconds).

This patch allows two formats:
- ISO 8601 format with PT prefix, e.g. `PT10s`, `PT1m30s` (as `Duration.parse`);
- simplified format without PT prefix, e.g. `10s`, `1m`, `2h` (IMHO this format is very readable for properties files).

Comments and feedback is welcome. I'm not sure about using a property editor, perhaps a `Converter` is a better choice?
And I'm not sure if this should be mentioned somewhere in the spring-reference... please advice :-)

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/PropertyEditorRegistrySupport.java
+++ b/spring-beans/src/main/java/org/springframework/beans/PropertyEditorRegistrySupport.java
@@ -26,6 +26,7 @@ import java.net.URI;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


dfa1:
is there something I can do to merge this PR?

dfa1:
@rstoyanchev just reworked a bit the commit, can you please have a review?

snicoll:
@dfa1 thaks for the PR and sorry for the delay. We've been working on an extensive review of duration support in #30396. I unfortunately am going to close this PR in favor of ours.

dfa1:
@snicoll no worries... the new PR is much better and complete! 😇 

thanks!

Issue #28025, opened 2022-02-10
Improve CollectionFactory to allow for single statement collection creation


snicoll:
Right now `CollectionFactory` allows to create a collection type, trying to find the most suitable implementation based on an concrete instance. 

We have some use cases in AOT where we'd like to be able to streamline the creation of a collection in a single statement. We've tried something along those lines:

```java
Stream.of(new RuntimeBeanReference("myCommandHandler")).collect(Collectors.toCollection(ManagedList::new))
```

Unfortunately the indirection with the collector can easily confuse the compiler. A more direct use where the `collectionFactory` would be provided with a bunch of elements could be nice.

sbrannen:
@snicoll, I've pushed a proof of concept to the following branch.

https://github.com/sbrannen/spring-framework/commits/issues/gh-28025-CollectionFactory-collectionOf

I think the new `createCollection(Class<C>, Class<E>, E...)` may suit your needs, but I'd like to get your feedback before pushing anything to `main`.

Thanks!

snicoll:
This is no longer needed as the AOT use cases became irrelevant.

PR #28094, opened 2022-02-22
CollectionFactory add createSingletonCollection method

xixingya:
Hello,this pr may solve this https://github.com/spring-projects/spring-framework/issues/28025
if I misundertand it please tell me, thank you! 
why use Collectors can easily confuse the compiler


snicoll:
@xixingya thank you for the PR but that doesn't address #28025. What we're trying to do there is en equivalent of `List.of()` (or `Set.of()`)  where we could provide the collection implementation that we want to use, potentially reusing the existing algorithm to find out the closest one based on an existing instance.

Issue #28090, opened 2022-02-22
When SQL init, 'PostgresqlBadGrammarException' is thrown by improperly splitting a valid statement that ';' containing


wickedev:
## Issue

The syntax below is valid in PostgreSQL, but ';' characters cause improper separation and execution.

```sql
CREATE OR REPLACE RULE insert_docents AS ON INSERT TO docents DO INSTEAD (
    INSERT INTO users
    VALUES (default,
            new.email,
            new.name,
            new.hash_salt,
            new.roles)
    ON CONFLICT (id) DO NOTHING;
    INSERT
    INTO docent_users
    VALUES (currval('users_id_seq'),
            new.introduce_short,
            new.introduce_long,
            new.video,
            new.images)
    );
```

```
Caused by: org.springframework.r2dbc.connection.init.ScriptStatementFailedException: Failed to execute SQL script statement #4 of class path resource [db/schema.sql]: CREATE OR REPLACE RULE insert_docents AS ON INSERT TO docents DO INSTEAD ( INSERT INTO users VALUES (default, new.email, new.name, new.hash_salt, new.roles) ON CONFLICT (id) DO NOTHING; nested exception is io.r2dbc.postgresql.ExceptionFactory$PostgresqlBadGrammarException: [42601] syntax error at end of input
```

another example

```sql
CREATE OR REPLACE FUNCTION update_emp_view() RETURNS TRIGGER AS $$
    BEGIN
        --
        -- Perform the required operation on emp, and create a row in emp_audit
        -- to reflect the change made to emp.
        --
        IF (TG_OP = 'DELETE') THEN
            DELETE FROM emp WHERE empname = OLD.empname;
            IF NOT FOUND THEN RETURN NULL; END IF;

            OLD.last_updated = now();
            INSERT INTO emp_audit VALUES('D', user, OLD.*);
            RETURN OLD;
        ELSIF (TG_OP = 'UPDATE') THEN
            UPDATE emp SET salary = NEW.salary WHERE empname = OLD.empname;
            IF NOT FOUND THEN RETURN NULL; END IF;

            NEW.last_updated = now();
            INSERT INTO emp_audit VALUES('U', user, NEW.*);
            RETURN NEW;
        ELSIF (TG_OP = 'INSERT') THEN
            INSERT INTO emp VALUES(NEW.empname, NEW.salary);

            NEW.last_updated = now();
            INSERT INTO emp_audit VALUES('I', user, NEW.*);
            RETURN NEW;
[...]

PR #28096, opened 2022-02-22
Support dollar sign quoting and parentheses in `ScriptUtils`

wickedev:
- spring-r2dbc `ScriptUtils.splitSqlScript`: Change not to use ; character between $$ ... $$ or ( ... ) as statement separator
- [Existing Issue](https://github.com/spring-projects/spring-framework/issues/28090)
- Reopen PR with branch issue. sorry. [Existing PR](https://github.com/spring-projects/spring-framework/pull/28091)

### DIFF
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/connection/init/ScriptUtils.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/connection/init/ScriptUtils.java
@@ -22,6 +22,7 @@ import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Stack;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


wickedev:
@sbrannen Currently, the change is applied only to the spring-r2dbc side, is it good to reflect the same change to the spring-jdbc side as well?

spring-r2dbc/src/main/java/org/springframework/r2dbc/connection/init/ScriptUtils.java
```diff
@@ -484,8 +488,19 @@ else if (script.startsWith(blockCommentStartDelimiter, i)) {
 			else if (!inSingleQuote && (c == '"')) {
 				inDoubleQuote = !inDoubleQuote;
 			}
+			else if (matchOnDollarQuote(c, nc)) {
+				inDollarQuote = !inDollarQuote;
+			}
+			else if (c == '(') {
+				brackets.push(i); // open bracket
```
Roli24, line 495: Can we use switch in spite of multiple if-else calls

Issue #28029, opened 2022-02-10
Make BeanDefinitionValueResolver public


snicoll:
`BeanDefinitionValueResolver` is package private at the moment and the AOT engine could use calling it directly to resolve constructur arguments typically. 

We could also improve the current usage where a default `TypeConverter` can be provided by default as the current pattern uses a protected method of the bean factory to initialize it.

PR #28097, opened 2022-02-23
Update Javadoc since for BeanDefinitionValueResolver

izeye:
This PR updates the Javadoc `@since` tag for the `BeanDefinitionValueResolver`.

See gh-28029


snicoll:
Spring Framework doesn't use the same convention (of not having a `@since` and then adding it if the class becomes public). I need someone else to double check if I am right (@jhoeller ?).

klasikcorp30 (approved):

snicoll:
It's back to being private so this PR is obsolete.

PR #28100, opened 2022-02-23
Fix library name for JavaPoet in src/docs/dist/license.txt

izeye:
This PR fixes library name for JavaPoet in `src/docs/dist/license.txt`.

This PR also updates Objenesis version in the file to align with [the version that Spring Framework is using](https://github.com/spring-projects/spring-framework/blob/main/spring-core/spring-core.gradle#L13).


snicoll:
Thanks again Johnny! I have backported the Objenesis version update in `5.3.x` as well, see 453c6d41f71acc54

PR #28105, opened 2022-02-25
Support custom HTTP status in client-side REST testing support

AzZureman:
## TL; DR;
Add supporting for custom http statuses in mocked responses
I copied logic from `org.springframework.http.ResponseEntity`

## Foreword
In my company we use "custom not found status" in our REST API's to prevent mixing network error (404 "not found" for unknown routes) with rest responses (when entity with given id not found).
There are good example for rest controller tests.
Suppose we have a rest endpoint for car entity: `GET /car/{id}`. 
Let's try to write test for case when entity with given id not found. Something like that:
```java
mockMvc.perform(get("/car/{id}/", nonExistsId.toString()))
        .andExpect(status().is(404));
```
And if our code in rest controller have bug, then test will fail with something like "500 internal server error"
But if we also make mistake in test in request's path, like that:
```java
mockMvc.perform(get("/carS/{id}/", nonExistsId.toString()))
        .andExpect(status().is(404));
```
then test will pass and we don't catch bugs in cases when entity with given id is non-exists.
So, for this reason we used "454" as http status for "not found" case.

## Problem
In our clients code we also write tests and want to cover this case with custom http status, but we cant because MockRestResponseCreators and MockClientHttpResponse not support custom http codes:
```java
  mockRestServiceServer.expect(requestTo(serverURI)
          .andExpect(method(HttpMethod.GET))
          .andRespond(withStatus(454)); // <-- will not work cause support only HttpStatus 
```

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/http/client/MockClientHttpResponse.java
+++ b/spring-test/src/main/java/org/springframework/mock/http/client/MockClientHttpResponse.java
@@ -32,7 +32,7 @@ import org.springframework.util.Assert;
  */
 public class MockClientHttpResponse extends MockHttpInputMessage implements ClientHttpResponse {
 
-	private final HttpStatus status;
+	private final Object status;
 
 
 	/**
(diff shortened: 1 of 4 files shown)


AzZureman:
Can i do something to move PR forward?

PR #28108, opened 2022-02-27
Use equals() instead of contains() in RollbackRuleAttribute

### LINKED ISSUE #28098: Use exception type information instead of pattern matching for transaction rollback rules
Source code in question：

```java
private int getDepth(Class<?> exceptionClass, int depth) {
	if (exceptionClass.getName().contains(this.exceptionName)) {
		// Found it!
		return depth;
	}
	// If we've gone as far as we can go and haven't found it...
	if (exceptionClass == Throwable.class) {
		return -1;
	}
	return getDepth(exceptionClass.getSuperclass(), depth + 1);
}
```

test code:

```java
public class CustomException extends Exception {
}

public class CustomExceptionX extends Exception {
}

@Override
@Transactional(rollbackFor = CustomException.class)
public void testTransaction() throws Exception {
    taskMapper.softDeleteById(1, "test");
    if (1 == 1) {
        throw new CustomExceptionX();
    }
}
```

lzj960515:
Closes gh-28098

### DIFF
--- a/spring-tx/src/main/java/org/springframework/transaction/interceptor/RollbackRuleAttribute.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/interceptor/RollbackRuleAttribute.java
@@ -112,7 +112,7 @@ public class RollbackRuleAttribute implements Serializable{
 
 
 	private int getDepth(Class<?> exceptionClass, int depth) {
-		if (exceptionClass.getName().contains(this.exceptionName)) {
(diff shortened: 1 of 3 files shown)


sbrannen:
Please note that this breaks the existing `RollbackRuleTests.foundEnclosedExceptionWithEnclosingException()` test which means that the proposal is a breaking change.

PR #28113, opened 2022-02-28
Avoid multiple invocations of private lifecycle method

### LINKED ISSUE #28083: Private init/destroy method may be invoked twice
Imagine, you have a bean with an `init` method, annotated with `@PostConstruct` and a configuration class, referencing the same `init` method on its own.

```java
public class MyBean {

    @PostConstruct
    public void init() {
        ...
    }

}
```

```java
@Configuration
public class MyConfiguration {

    @Bean(initMethod = "init")
    public MyBean myBean() {
        return new MyBean();
    }
}
```

The `init` method will be invoked only once (what is pretty well). 

This is true for all method modifiers except `private`.
As soon as you change the method modifier to `private` the  `init` method is invoked twice. 

```diff
public class MyBean {

    @PostConstruct
-    public void init() {
+    private void init() {
        ...
    }
}
```

vikeychen:
See gh-28083

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
@@ -1830,6 +1830,12 @@ public abstract class AbstractAutowireCapableBeanFactory extends AbstractBeanFac
 			}
 		}
 
+		if (Modifier.isPrivate(initMethod.getModifiers())) {
+			if (mbd.isExternallyManagedInitMethod(ClassUtils.getQualifiedMethodName(initMethod))) {
+				return;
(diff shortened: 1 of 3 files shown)


sbrannen:
Thanks for the PR, but we have decided to approach the fix differently, and I have therefore assigned #28083 to myself.

sbrannen:
@vikeychen, I merged your tests in af14eea1ef76576acd07ba90cd0a656bd6b31969, since there was no need to rewrite the work you'd already done there. 

And it ended up being a bit nostalgic for me, since `Spr3775InitDestroyLifecycleTests` was a test class I wrote way back in 2007. 😉 

Thanks!

vikeychen:
@sbrannen, wow it's been a long time! I'm glad I could be a little help. 🤓

PR #28127, opened 2022-03-02
Return empty LinkedHashMap instead of `Collections.emptyMap()` in PathVariableMapMethodArgumentResolver

dugenkui03:
I think it would be better that return empty `LinkedHashMap` instead of `Collections.emptyMap()` in `PathVariableMapMethodArgumentResolver`, because they do't have same support for `put` operation.  And I think that the supported operatioin of `Map` annotated with `@PathVariable` should not be changed according the value(empty or not).

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/PathVariableMapMethodArgumentResolver.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/PathVariableMapMethodArgumentResolver.java
@@ -16,7 +16,7 @@
 
 package org.springframework.web.reactive.result.method.annotation;
 
-import java.util.Collections;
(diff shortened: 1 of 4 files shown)


bclozel:
Those are not meant to mutable; in several places we're wrapping the underlying collections with `Collections.unmodifiableMap`. So using `Collections.emptyMap()` is consistent with that approach.

I'm declining this PR as a result. Thanks!

dugenkui03:
@bclozel Thanks for your response.

In `mvc.method.annotation.PathVariableMapMethodArgumentResolver`, the original Map maybe immutable, but the returned value is mutable .
image
Do you think is it nessary that return unmodifiable view of copy even the Map is not empty, for keep same support for `put` operation.

bclozel:
For consistency we could indeed change that to make it immutable.

Issue #27991, opened 2022-01-31
Add GraalVM native JSON configuration generation


sdeleuze:
As a follow-up of #27829, we need to implement GraalVM native JSON configuration generation for [reflection](https://www.graalvm.org/22.0/reference-manual/native-image/Reflection/#manual-configuration), [JDK proxies](https://www.graalvm.org/22.0/reference-manual/native-image/DynamicProxy/#manual-configuration), [resources](https://www.graalvm.org/22.0/reference-manual/native-image/Resources/) and serialization as one of the possible outputs.

Other use cases could be generating HTML reports for example, so the APIs should be designed for other use cases, GraalVM native JSON configuration generation being just one use case among others.

snicoll:
Superseded by PR #28131

PR #28131, opened 2022-03-03
Add GraalVM native JSON configuration generation

sdeleuze:
Here is a PR intended to be a proposal to implement #27991.

The related commit implements 4 package private Jackson `JsonSerializer` for `JavaSerializationHints`, `ProxyHints`, `ReflectionHints` and `ResourceHints` to serialize GraalVM native JSON configuration as documented in https://www.graalvm.org/22.0/reference-manual/native-image/BuildConfiguration/.

It exposes the related functionality via `NativeConfigurationGenerator` which allows to generate the relevant files on the filesystem.

The generated `*-config.json` files has been validated working with GraalVM 22.0.

Looking forward to your feedback @jhoeller.

### DIFF
--- a/spring-core/spring-core.gradle
+++ b/spring-core/spring-core.gradle
@@ -67,6 +67,7 @@ dependencies {
 	testImplementation("org.xmlunit:xmlunit-matchers")
 	testImplementation("io.projectreactor:reactor-test")
 	testImplementation("io.projectreactor.tools:blockhound")
+	testImplementation("org.skyscreamer:jsonassert")
 	testFixturesImplementation("com.google.code.findbugs:jsr305")
 	testFixturesImplementation("org.junit.platform:junit-platform-launcher")
 	testFixturesImplementation("org.junit.jupiter:junit-jupiter-api")
--- a/spring-core/src/main/java/org/springframework/aot/nativex/FileNativeConfigurationGenerator.java
+++ b/spring-core/src/main/java/org/springframework/aot/nativex/FileNativeConfigurationGenerator.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.aot.nativex;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Path;
+
+import org.springframework.aot.hint.JavaSerializationHints;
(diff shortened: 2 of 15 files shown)


sdeleuze:
I think I am going to update that PR to use `RuntimeHints` in `NativeConfigurationGenerator`.

sdeleuze:
I have updated the PR to expose a single `public void generate(RuntimeHints hints, Path targetDirectory)` method in `NativeConfigurationGenerator`.

bclozel:
I've just discussed this with @sdeleuze because I remembered a specific point about the AOT processing in Spring Native: this processing phase is triggered by the build plugins as they fork a JVM process using the application runtime classpath. This last point is necessary since all the classloader detections need to happen in a consistent fashion with the actual runtime behavior of the application.

Using Jackson as a required dependency means that build plugins would have to artificially add Jackson as a dependency to the runtime classpath, thus invalidating Jackson auto-detection during the AOT processing. This is why in Spring Native we chose to use/shade the vaadin "android-json" dependency to manipulate JSON. We had more specific needs there, such as reading/merging/writing JSON documents. In this case, since we're only writing a document out, maybe we should use a `StringBuffer` approach and avoid this problem completely?

sdeleuze:
As discussed with @bclozel, using Jackson is not ok since during the AOT processing that will modify th classpath and impact the active beans via conditions. I will rework the PR with manual serialization of the JSON files.

spring-core/src/main/java/org/springframework/aot/hint/nativex/NativeConfigurationGenerator.java
```diff
@@ -0,0 +1,160 @@
...
+		module.addSerializer(ResourceHints.class, new ResourceHintsSerializer());
+		this.objectMapper.registerModule(module);
+	}
+
+	/**
+	 * Generate the native configuration files for the provided hints.
+	 * @param hints the hints to serialize
+	 * @param targetDirectory directory where the {@code /META-INF/native-image/*-config.json} files should be generated
+	 */
+	public void generate(RuntimeHints hints, Path targetDirectory) {
```
snicoll, line 71: I wonder if a strategy interface here would be more appropriate. The generator is focusing on file system but that should be an implementation detail. It would be nice also if the [notion of namespacing](https://www.graalvm.org/22.0/reference-manual/native-image/BuildConfiguration/#embedding-a-configuration-file) was a bit more evident. 

For instance, an implementation could take a groupId, an artifactId and then generate the `META-INF/native-image/{groupId}/{artifactId}` structure based on a base path.

sdeleuze:
@bclozel @snicoll Could you please review the updated PR and let me know your advice?

spring-context/src/main/java/org/springframework/context/nativex/NativeConfigurationGenerator.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.context.nativex;
+
+import org.springframework.aot.hint.RuntimeHints;
+
+public interface NativeConfigurationGenerator {
```
sbrannen, line 21: Add class-level Javadoc.

sdeleuze:
The application used to validate it works as expected is https://github.com/sdeleuze/generate-native-config.

sdeleuze:
Merged via 77e0100f42253ab26947cf2258ee9653e2ac811d.

PR #28132, opened 2022-03-03
Batch update return generated keys

ctailor2:
Add `NamedParameterJdbcTemplate#batchUpdate` methods taking `KeyHolder` and `String[] keyColumnNames` as counterparts to the `NamedParameterJdbcTemplate#update` methods of the same form.

This is in service of an enhancement to [`spring-data-jdbc`](https://github.com/spring-projects/spring-data-relational/pull/1191) to perform batch inserts to persist collections of entities referenced by an aggregate, when adequately supported by the database. The proposed changes to `spring-data-jdbc` currently include a [copy of the changes](https://github.com/spring-projects/spring-data-relational/pull/1191/files#diff-2526db37d52bc8ab9d5cb280c184ac07d28c0f1bfea856ea228653584374c04b) being proposed here, and could be refactored to use the respective `NamedParameterJdbcTemplate#batchUpdate` once accepted and merged.

Related to spring-projects/spring-framework#6530.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
@@ -1563,7 +1563,8 @@ public class JdbcTemplate extends JdbcAccessor implements JdbcOperations {
 					int[] results = ps.executeBatch();
 					afterUpdateCallback.doInPreparedStatement(ps);
 					return results;
-				} else {
+				}
+				else {
 					List<Integer> rowsAffected = new ArrayList<>();
 					for (int i = 0; i < batchSize; i++) {
 						pss.setValues(ps, i);
@@ -1579,7 +1580,8 @@ public class JdbcTemplate extends JdbcAccessor implements JdbcOperations {
 					}
 					return rowsAffectedArray;
 				}
-			} finally {
+			}
+			finally {
 				if (pss instanceof ParameterDisposer) {
 					((ParameterDisposer) pss).cleanupParameters();
 				}
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
@@ -1026,49 +1026,20 @@ public class JdbcTemplate extends JdbcAccessor implements JdbcOperations {
 	}
 
 	@Override
+	public int[] batchUpdate(PreparedStatementCreator psc, final BatchPreparedStatementSetter pss) throws DataAccessException {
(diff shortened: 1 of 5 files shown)


trisberg:
I'm getting some checkstyle violations trying to build. Could you fix them so I can try this out? Thanks.

ctailor2:
@trisberg I resolved the checkstyle violations and made some other updates.

pworoniecki:
Any update on this topic? Is it abandoned or is something still to be considered?
Would be great to have this code merged (though I guess it needs some adjustment to the current state of the code).

sabomichal:
Would be nice to have this merged already.

spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
```diff
@@ -1567,6 +1527,72 @@ private static int updateCount(@Nullable Integer result) {
...
+						new RowMapperResultSetExtractor<>(getColumnMapRowMapper(), rowsExpected);
+				generatedKeys.addAll(result(rse.extractData(keys)));
+			}
+			finally {
+				JdbcUtils.closeResultSet(keys);
+			}
+		}
+	}
+
+	private PreparedStatementCallback<int[]> getPreparedStatementCallback(BatchPreparedStatementSetter pss, @Nullable KeyHolder generatedKeyHolder) {
```
ePaul, line 1545: I would suggest a more descriptive name:
```suggestion
	private PreparedStatementCallback<int[]> getPreparedStatementCallbackForBatchUpdate(BatchPreparedStatementSetter pss, @Nullable KeyHolder generatedKeyHolder) {
```
(Of course, also adjust this where it's used.)

ePaul:
I just spent some hours trying to figure out how to implement this myself (or rather, how to generalize [this Stack Overflow answer](https://stackoverflow.com/a/15942268/600500) into something which works for my use case).

Only then I had the idea of checking whether someone already did something here, and it looks good.

snicoll:
@ctailor2 thanks for the PR, it misses some tests for the changes in `NamedParameterJdbcTemplate` though.

@schauder can you share your thoughts? I can see something in spring data jdbc was merged with a reference to a core framework class and I am a bit puzzled by that.

schauder:
@snicoll not sure what kind of thoughts you are looking for. 

The changes in Spring Data JDBC are creating a copy of frameworks `NamedParameterJdbcTemplate` with the changes of this PR included, because we really needed the changes. We would be glad to go back to the official `NamedParameterJdbcTemplate` once this PR is merged.

snicoll:
@ctailor2 and all, thanks for your patience and please use `NamedParameterJdbcTemplate` proper again.

PR #28156, opened 2022-03-10
Initial version with TX, Context Propagation and Observation

marcingrzejszczak:
@mp911de


spring-tx/src/main/java/org/springframework/transaction/support/DefaultTransactionTagsProvider.java
```diff
@@ -0,0 +1,96 @@
...
+		}
+		TransactionDefinition transactionDefinition = context.getTransactionDefinition();
+		Tags tags = Tags.empty();
+		if (transactionDefinition.getTimeout() > 0) {
+			tags = tags.and(TransactionObservation.LowCardinalityTags.TIMEOUT.of(String.valueOf(transactionDefinition.getTimeout())));
+		}
+		if (StringUtils.hasText(transactionDefinition.getName())) {
+			tags = tags.and(TransactionObservation.LowCardinalityTags.NAME.of(transactionDefinition.getName()));
+		}
+		return tags.and(TransactionObservation.LowCardinalityTags.TRANSACTION_MANAGER.of(ClassUtils.getQualifiedName(context.getTransactionManagerClass())),
```
marcingrzejszczak, line 47: `TransactionManagerClass ` can be `null` ?

spring-tx/src/main/java/org/springframework/transaction/support/ObservationPlatformTransactionManager.java
```diff
@@ -0,0 +1,129 @@
...
+	@Override
+	public TransactionStatus getTransaction(TransactionDefinition definition) throws TransactionException {
+		Observation obs = this.observationRegistry.getCurrentObservation();
+		if (obs == null) {
+			obs = TransactionObservation.TX_OBSERVATION.observation(this.observationRegistry, this.context)
+					.tagsProvider(this.transactionTagsProvider)
+					.start();
+		}
+		this.observation = obs;
+		this.scope = this.observation.openScope();
```
marcingrzejszczak, line 67: if there was a scope `obs != null` DO  NOT open a new scope

spring-tx/src/main/java/org/springframework/transaction/support/TransactionObservation.java
```diff
@@ -0,0 +1,114 @@
...
+			@Override
+			public String getKey() {
+				return "spring.tx.timeout";
+			}
+		},
+
+		/**
+		 * Transaction name.
+		 */
+		NAME {
```
marcingrzejszczak, line 105: This is high cardinality cause name by default is a new UUID

bclozel:
We have implemented lots of observability features already and I'm not sure we should go further ahead here with an observation per transaction call. Maybe this is similar to #30089 in the sense that we don't really need metrics/traces for this part? I'm not sure we need context propagation anyway as the transaction guarantee some context usage already.

I'm closing this issue for now and we'll revisit depending on the community requests.

PR #28183, opened 2022-03-15
Improve documentation of Caching annotations

luvarqpp:
There was missing basic information for `conditional` attribute of `Cacheable` annotation. Perhaps something similar should be added also to `unless` attribute.

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/annotation/Cacheable.java
+++ b/spring-context/src/main/java/org/springframework/cache/annotation/Cacheable.java
@@ -123,7 +123,8 @@ public @interface Cacheable {
 
 	/**
 	 * Spring Expression Language (SpEL) expression used for making the method
-	 * caching conditional.
+	 * caching conditional. If condition is evaluated to {@code true}, result
+	 * is cached.
 	 * <p>Default is {@code ""}, meaning the method result is always cached.
 	 * <p>The SpEL expression evaluates against a dedicated context that provides the
 	 * following meta-data:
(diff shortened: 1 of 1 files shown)


spring-context/src/main/java/org/springframework/cache/annotation/Cacheable.java
```diff
@@ -137,6 +137,7 @@
 	 * can be accessed via {@code #root.args[1]}, {@code #p1} or {@code #a1}. Arguments
 	 * can also be accessed by name if that information is available.</li>
 	 * </ul>
+	 * If condition is evaluated to {@code true}, result is cached.
```
snicoll, line 140: I think this would be better as an opening statements, rather than at the end. We should also update the paragraph about the default to mention that if no expression is set, the method result is always cached.

snicoll (commented):
Thanks for the PR. Making it obvious that the expression should return a `Boolean` and what the value should be to cache the item is a nice improvement to the Javadoc.

I've made a comment of moving this as an opening statement, rather than at the end. Would you be willing to upgrade your PR in that direction?

spring-context/src/main/java/org/springframework/cache/annotation/Cacheable.java
luvarqpp: Nice words for me, thanks. I am on it.

luvarqpp: Is it better now?
Should I squash / rebase?

snicoll:
@luvarqpp thank you for making your first contribution to Spring Framework.

PR #28188, opened 2022-03-16
Apply "instanceof pattern matching" in spring-core

diguage:


Roli24 (commented):
Can we not make common utility Method to check the instance of method ?

rstoyanchev:
I'm not convinced this needs to be applied throughout. Especially where variables are used once.

sbrannen (changes requested):
Please rebase your changes on `main`.

Also, for checks like `value instanceof Class clazz` you'll need to add the generic wildcard like `value instanceof Class<?> clazz`. So please make the appropriate changes. In other words, we don't want to have to suppress "raw types" warnings in such cases.

diguage:
> Please rebase your changes on `main`.
> 
> Also, for checks like `value instanceof Class clazz` you'll need to add the generic wildcard like `value instanceof Class<?> clazz`. So please make the appropriate changes. In other words, we don't want to have to suppress "raw types" warnings in such cases.

OK , let me update the codes.

spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
```diff
@@ -572,8 +572,8 @@ public static Mono<DataBuffer> join(Publisher<? extends DataBuffer> dataBuffers)
 	public static Mono<DataBuffer> join(Publisher<? extends DataBuffer> buffers, int maxByteCount) {
 		Assert.notNull(buffers, "'dataBuffers' must not be null");
 
-		if (buffers instanceof Mono) {
-			return (Mono<DataBuffer>) buffers;
+		if (buffers instanceof Mono mono) {
```
diguage, line 575: I want to add the generic wildcard to the `Mono`, but the latter return statement shows an error.
Should I add the generic wildcard to the `Mono`?

sbrannen: >  Should I add the generic wildcard to the Mono?

I think we just have to leave that one as-is: a generic wildcard is not always possible.

In any case, I'll review it when I merge your work.

spring-core/src/main/java/org/springframework/util/xml/DomContentHandler.java
```diff
@@ -50,8 +50,8 @@ class DomContentHandler implements ContentHandler {
 	 */
 	DomContentHandler(Node node) {
 		this.node = node;
-		if (node instanceof Document) {
-			this.document = (Document) node;
+		if (node instanceof Document document) {
```
sbrannen, line 53: This causes the build to fail due to a Checkstyle bug.

I've addressed it locally on my machine and will create an issue to track the Checkstyle bug.

sbrannen:
This has been merged into `main` in bbaf7578b2ca66ec1a4cff114c3e2bc4d7d3b462 and revised in 43dd813ee64e7c62aeeaf24cdd52230457336d5b and 16f65e59b49ebdc7813819c137203fccccf8f4f2.

Thanks! 👍

sbrannen:
It turns out that the Checkstyle issue is a known limitation.

- https://github.com/checkstyle/checkstyle/issues/10969

PR #28197, opened 2022-03-18
Add Node.js file extensions to javascript

dsyer:
Node.js recognizes .mjs as an ES6 module and .cjs as a CommonJS module. Other web servers e.g. in other languages recognize them as well. This change makes the resource handlers in Spring Web work the same with *.mjs as *.js.

Example with Python:

```
$ curl -v localhost:8000/hello.mjs > /dev/null
...
< Content-type: application/javascript
< Content-Length: 825
< Last-Modified: Tue, 15 Mar 2022 11:00:54 GMT
< 
```

### DIFF
--- a/spring-web/src/main/resources/org/springframework/http/mime.types
+++ b/spring-web/src/main/resources/org/springframework/http/mime.types
@@ -146,7 +146,7 @@ application/ipfix				ipfix
 application/java-archive			jar
 application/java-serialized-object		ser
 application/java-vm				class
-application/javascript				js
+application/javascript				js mjs cjs
 # application/jose
 # application/jose+json
 # application/jrd+json


bclozel:
It looks like the Python dev server is [only registering the `*.mjs` extension](https://github.com/python/cpython/blob/d27af88c1bcf3c6c185f81218b7bf86b38128ea9/Lib/mimetypes.py#L431) and is relying on [known files](https://github.com/python/cpython/blob/d27af88c1bcf3c6c185f81218b7bf86b38128ea9/Lib/mimetypes.py#L48-L58) for the rest. This server is meant for development only and not for production usage. Well-known servers such as [nginx](https://github.com/nginx/nginx/blob/828fb94e1dbe1c433edd39147ba085c4622c99ed/conf/mime.types#L8), [httpd](https://svn.apache.org/viewvc/httpd/httpd/trunk/docs/conf/mime.types?view=markup#l145) or [Tomcat](https://github.com/apache/tomcat/blob/76b2641ae9704822b9446426e3f330ab45672a9f/java/org/apache/catalina/startup/MimeTypeMappings.properties#L385) do not support those extensions.

Given that [registering additional extensions is quite easy](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc-config-content-negotiation), I don't think we should add those entries here. The main goal here is to remain in sync with those well-known servers, so I'm declining this PR as a result.

Thanks!

Issue #10789, opened 2009-09-16
Dependency injection of @Configurable objects should work across test suites [SPR-6121]


spring-issuemaster:
**[Marc Ludwig](https://jira.spring.io/secure/ViewProfile.jspa?name=baronludwig)** opened **[SPR-6121](https://jira.spring.io/browse/SPR-6121?redirect=false)** and commented

#### Overview

We have a large number of unit/integration tests that assert behavior within our system and rely upon injection of dependencies into configurable domain objects.

These tests all work when executed individually; however, when executed within a suite (either through the IDE or Ant) certain tests fail as dependencies have not been injected into the `@Configurable` objects. We have also seen dependencies that were configured for test _X_ being injected into test _Y_ rather than the dependencies for test _Y_; but I have no test case for this.

---

#### Steps to Reproduce

I have created a set of three tests -- tests 1 and 3 are basically identical. If these are executed in a suite the third test will fail as the dependency is not injected into the `@Configurable` object even though it is available to the test. This only occurs if test 2 is a Spring test.

See attached zip file.

---

#### Further Resources

- [Spring Test Context Caching + AspectJ @Transactional + Ehcache pain](http://www.javacodegeeks.com/2014/04/spring-test-context-caching-aspectj-transactional-ehcache-pain.html) blog by Java Code Geeks

---

**Affects:** 2.5.6, 3.0.5

**Reference URL:** http://forum.springsource.org/showthread.php?t=77980

**Attachments:**
- [springtest.zip](https://jira.spring.io/secure/attachment/15724/springtest.zip) (_3.15 MB_)

**Issue Links:**
- #15989 SpringContexts not properly closed after test-class is finished (_**"is duplicated by"**_)
- #11019 TestContext framework should support one AspectJ instance per ApplicationContext
- #17123 AnnotationTransactionAspect retains reference to JpaTransactionManager from closed context

11 votes, 14 watchers

sbrannen:
Superseded by #11019

PR #28201, opened 2022-03-19
@Async aspect should clear cached executors when different BeanFactory set

cdalexndr:
When using AspectJ and multi-context (ex: testing), the bean factory needs to be changed on the singleton aspect.
This fix clears cached method executors when changing bean factory.
If the first context was previously closed with the executors, then an error will be thrown when trying to reuse the closed executor.


spring-aop/src/main/java/org/springframework/aop/interceptor/AsyncExecutionAspectSupport.java
```diff
@@ -152,7 +160,10 @@ public void setExceptionHandler(AsyncUncaughtExceptionHandler exceptionHandler)
 	 */
 	@Override
 	public void setBeanFactory(BeanFactory beanFactory) {
-		this.beanFactory = beanFactory;
+		if (!Objects.equals( this.beanFactory, beanFactory )) {
```
andrei-ivanov, line 163: is this passing the style checks?

cdalexndr: Task :spring-aop:checkstyleMain reports no error

sbrannen: In any case, please remove empty space after an opening parenthesis and before a closing parenthesis, since that is the standard convention within the Spring Framework codebase.

And please make that change consistently within this PR.

sbrannen (changes requested):

sbrannen:
@cdalexndr, is there an existing issue associated with this "fix"?

If not, can you please add more information to this PR's description to explain the rationale and what error cases the proposed change fixes?

cdalexndr:
@sbrannen updated description

sbrannen:
## Related Issues

- #11019
- #11897
- #17123
- #10789
- #12619

jhoeller:
While I do not see the need for a `DisposableBean` implementation as the PR suggests, we can simply clear the cached executors in `setBeanFactory` directly (for scenarios where the instance is reused). I'll resolve it along those lines.

Issue #25986, opened 2020-10-28
Inefficient request handling inside ServletRequestDataBinder


jengebr:
**Affects:** Spring web 4.2+, 5.x (confirmed, possibly others)

---
A large, publicly-facing application receives more than 100 request parameters (don't ask) and the growing runtime of ServletRequestDataBinder has exposed an inefficiency: for each parameter, AbstractNestablePropertyAccessor.processLocalProperty() throws an exception, which is caught by AbstractPropertyAccessor.setPropertyValues().  This exception is then discarded because "ignoreUnknown" is true, and life goes on.

Throwing exceptions is an expensive operation because of the need to fill in stack traces (Throwable.fillInStackTrace) and this generates significant wasted effort.  Modifying ServletRequestDataBinder.bind() to mark each PropertyValue as optional eliminates this penalty while maintaining the same functional behavior.  Custom benchmarking shows a 93% reduction in cpu and clock time (21 vs. 350).

Workarounds are difficult because ServletRequestDataBinder is accessed via a chain of hard-coded references, plus our application uses the subclass ExtendedServletRequestDataBinder.  The best option I've found so far is to use a custom bean factory to override the requestMappingHandlerAdapter bean (declared in WebMvcConfigurationSupport.requestMappingHandlerAdapter()) and make a subclass that overrides createDataBinderFactory, etc.  "Convoluted" and "ugly" are good descriptions of those solution.

jengebr:
@jhoeller  thank you for the fast turnaround!  Is it possible to backport this to 4.3.x?

jhoeller:
I'll see what we can do in terms of backports. Depending on the eventual shape of the `Abstract(Nestable)PropertyAccessor` revision, we may be able to easily backport it all the way down to 4.3.x indeed.

For a start, we got immediate 5.3.1 and 5.2.11 releases coming on Nov 10, so those will be the first where such a revision can be released. 5.1.20, 5.0.20 and 4.3.20 relases are subsequently scheduled for December.

jengebr:
Excellent, thank you!  I appreciate anything you can do, and I understand your priorities.  Good luck with the releases!

jhoeller:
[...]

PR #28224, opened 2022-03-23
Fix property binding issue for spring beans when default value is null

dillonm79:
This PR introduced a bug an does not allow us to upgrade passed 5.3.0: https://github.com/spring-projects/spring-framework/issues/25986 . Before, `AbstractNestablePropertyAccessor#processLocalProperty` threw `createNotWritablePropertyException`  when our class being deserialized did not have a setter on a property. This was then swallowed by `AbstractPropertyAccessor#setPropertyValues(org.springframework.beans.PropertyValues, boolean, boolean, boolean)` and binding of the next property was attempted. This is the wanted outcome. Now if this exception is not thrown (caught by the guard clause), we don't have default values on these fields so AbstractNestablePropertyAccessor#setDefaultValue throws an assertion which is not swallowed. Setting `ignoreInvalid` to false does cause `createNotWritablePropertyException` to be thrown, but it does not swallow the exception as before. This PR keeps the exact same functionality as now, but allows not having a default value for a property to be swallowed, in the same way it is in the other cases if desired. It also is a better way to handle the error and then throwing an assertion.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
@@ -16,8 +16,25 @@
 
 package org.springframework.beans;
 
+import java.beans.PropertyChangeEvent;
+import java.lang.reflect.Array;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.UndeclaredThrowableException;
+import java.security.PrivilegedActionException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+
 import org.springframework.core.CollectionFactory;
 import org.springframework.core.ResolvableType;
 import org.springframework.core.convert.ConversionException;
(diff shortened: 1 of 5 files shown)


mikejhill:
This is a breaking issue for us. The patch seems to fix the problem. It looks like the optimization in #25986 didn't take into account cases where it would prevent a default value from being set. Although, frankly, I'm not sure how the optimization ever would have worked when non-writable property errors are suppressed:

```java
private Object setDefaultValue(PropertyTokenHolder tokens) {
    PropertyValue pv = createDefaultPropertyValue(tokens);
    /*
     * #setPropertyValue is effectively a no-op if the property is non-writable and non-writable exceptions are
     * suppressed (suppressNotWritablePropertyException).
     *
     * Before the optimization (#25986), exceptions would bubble up here and *later* be ignored if suppressed, causing
     * execution to escape without running any remaining instructions in this method. After the optimization, remaining
     * instructions do get executed, causing a different exception to be thrown (see PR description and comments below).
     */
    setPropertyValue(tokens, pv);
    /*
     * #getPropertyValue returns the value of the property read method for the backing instance. This is null if
     * #setPropertyValue was not completed (e.g., in the case that the property is not writable) and if the field does
     * not have a pre-existing value (e.g., private Object myField = null).
     */
    Object defaultValue = getPropertyValue(tokens);
    /*
     * This is presumably attempting to assert that the default value was set properly, but the default value never gets
     * set in the above cases, so an IllegalStateException is always thrown.
     */
    Assert.state(defaultValue != null, "Default value must not be null");
    return defaultValue;
}
```

The only preconditions I can find to trigger this are as follows:
- A property is included for binding that is non-writable (no setter)
- Non-writable exceptions are suppressed on the binder (`suppressNotWritablePropertyException`)
- The property has no default value at instantiation time (e.g., `private Object myField = null` with no default value provided by its getter)

I'm not familiar enough to know whether this is the right fix. In particular, I'm not certain that any of the other `setDefaultValue` callers are set up to properly handle null return values. But the patch definitely seems to fix the main issue for us.

jhoeller:
The `setDefaultValue` method is only really called within auto-growing of non-existent properties. I'm having trouble reproducing this, could you please provide a unit test that fails before and passes after the PR?

Generally speaking, I don't see a need for yet another ignoreXxx property here. If there is a specific regression to be found here, we should make it work within the existing arrangement and the with existing ignore flags.

bclozel:
Closing as we didn't get the request feedback. We can reopen this PR if we get a test that reproduces a failure with the existing codebase.

PR #28242, opened 2022-03-28
Property name consistent with the one in the code

lifei-zhang-awx:
In XML file, the property name is 'integerProperty' while the field in the code is 'i'.


sbrannen:
Hi @lifei-zhang-awx,

Thanks for submitting your first PR to the Spring Framework.



> In XML file, the property name is 'integerProperty' while the field in the code is 'i'.

Setter injection in XML configuration is based on JavaBeans property naming conventions.

The property name is the name in the setter method, not the name of the field.

```java
    public void setIntegerProperty(int i) {
        this.i = i;
    }
```

Thus, the example XML configuration is correct as it is now.

In light of that that, I am closing this PR.

PR #28251, opened 2022-03-30
Fix JDK >= 9 classLoader bypass

raylax:


spring-beans/src/main/java/org/springframework/beans/CachedIntrospectionResults.java
```diff
@@ -286,7 +286,7 @@ private CachedIntrospectionResults(Class<?> beanClass) throws BeansException {
 			// This call is slow so we do it once.
 			PropertyDescriptor[] pds = this.beanInfo.getPropertyDescriptors();
 			for (PropertyDescriptor pd : pds) {
-				if (Class.class == beanClass &&
+				if ((Class.class == beanClass || beanClass.getName().equals("java.lang.Module")) &&
```
luozhenyu, line 289: Consider using the Module.class instead of a String? The minimal JDK of Spring is 17 now.

rstoyanchev:
@raylax when creating an issue, please read the policy on the issue template.

Issue #28260, opened 2022-03-31
Backports for PropertyDescriptor filtering 


magJ:
**Affects:** Major and minor versions other than 5.3.x

---

Can we please have the PropertyDescriptor filtering commit https://github.com/spring-projects/spring-framework/commit/002546b3e4b8d791ea6acccb81eb3168f51abb15 backported to 4.x branches?

rstoyanchev:
@magJ are you on JDK 9+?

luozhenyu:
> @magJ are you on JDK 9+?

Spring framework 4.3.x is widely used for years and some people upgrade JDK 8 to 11 to get better performance.

magJ:
@rstoyanchev Yes, we have at least some projects that are using spring 4.x running on JDK 9+/

Artur-:
If this is not backported then we and and/or somebody else will be forced to fork Spring 4 and release a custom version. Does not sound tempting

jhoeller:
Please note that Spring Framework 4.x was *never* officially supported on a JDK higher than 8, not in open source and not for commercial support arrangements either. The official support range is JDK 6-8: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions - running it on a newer JDK version is technically possible for certain scenarios, but anyone doing so has always been on their own in terms of support.

Also, Spring Framework 4.3.30 (the latest release before 4.x EOL) is on a December 2020 security patch level. Continued usage of that version misses out on 15 months of vulnerability patches and defensiveness measures that we applied to 5.x in the meantime. A patched version for the present CachedIntrospectionResults change would cover this particular attack vector on the unsupported JDK 9+ but might still keep you exposed to other vulnerabilities even in supported areas of the framework.

For older Spring versions, we strategically suggest an upgrade to 5.3.x, or tactically a workaround as suggested in our blog post: https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement

luozhenyu:
> Please note that Spring Framework 4.x was _never_ officially supported on a JDK higher than 8, not in open source and not for commercial support arrangements either. The official support range is JDK 6-8: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions - running it on a newer JDK version is technically possible for certain scenarios, but anyone doing so has always been on their own in terms of support.
> 
[...]

PR #28263, opened 2022-03-31
Refine PropertyDescriptor filtering

luozhenyu:
Restrict property paths under `Class` and properties of types
`ClassLoader` or `ProtectionDomain`.

Closes gh-28260


bclozel:
The team takes care of backporting issues. Thanks for the offer though!

PR #28265, opened 2022-03-31
add properties for allow/deny on ConfigurableWebBindingInitializer

tubbynl:
to eventually implement them as configurable properties in WebProperties and into WebMvcAutoConfiguration and it's flux sibling if i can find it

https://github.com/spring-projects/spring-boot/issues/30490


rstoyanchev:
This will not serve the goals stated in https://github.com/spring-projects/spring-boot/issues/30490 because the same properties can be set on the DataBinder from an `@InitBinder` method in the controller or in the a `@ControllerAdvice` and that would override the values initially set.

tubbynl:
The intention of this pull request is to be able to configure these things without code changes. 

This in combination with AutoConfiguration will enable that

rstoyanchev:
I understand the intention, but in https://github.com/spring-projects/spring-boot/issues/30490 the goal is described as being able to react to a security issue. However, the proposed solution assumes the application does not explicitly configure allowed/disallowed fields in code, so it is not guaranteed to work.

PR #28322, opened 2022-04-11
Avoid duplicate `ApplicationListener` firing (proxy vs. target)

### LINKED ISSUE #28283: Using AOP and DI itself causes Spring ApplicationListener to be fired twice
## Software versions

- Spring Version  5.3.18 and earlier 
- JDK Version 1.8.0_202

## Overview

When I use Spring ApplicationListener, in order to prevent transaction invalidation, my ApplicationListener implementation class writes the following code (of course, the code can be written differently to avoid this problem), which will cause my listener to trigger twice after the event is published. I think it's not normal, but not sure if it's a bug, so I want to ask everyone's opinion.

```java
@Component
public class EventDemoListener implements ApplicationListener<EventDemo> {
    @Autowired
    DemoService1 demoService1;
    @Autowired
    DemoService2 demoService2;
    @Autowired
    EventDemoListener eventDemoListener;

    @Override
    public void onApplicationEvent(EventDemo event) {
        eventDemoListener.testTransaction();
        System.out.println("receiver " + event.getMessage());
    }

    @Transactional(rollbackFor = Exception.class)
    public void testTransaction() {
        demoService1.doService();
        demoService2.doService();
    }
}
```
Through this demo project, this problem can be reproduced. Please read the README.md document before running.    
[https://github.com/ZiFeng-Wu/spring-study](https://github.com/ZiFeng-Wu/spring-study)

## Analysis

1. After analysis, because here DI itself , When `EventDemoListener` is created, property filling will trigger `DefaultSingletonBeanRegistry#getSingleton(String, boolean)` in advance.

2. Then `singletonFactory.getObject()` executed in `getSingleton()` will cause the unproxyed `EventDemoListener` object to be put into `AbstractAutoProxyCreator#earlyProxyReferences`.

3. After the properties are filled, calling `AbstractAutowireCapableBeanFactory#initializeBean(String, Object, RootBeanDefinition)` and executing `ApplicationListenerDetector#postProcessAfterInitialization(Object, String)` will cause the unproxyed `EventDemoListener` object to be put into the `AbstractApplicationEventMulticaster.DefaultListenerRetriever#applicationListeners` container.

    [![enter image description here][1]][1]

4. Then when the event is published, execute `AbstractApplicationEventMulticaster.DefaultListenerRetriever#getApplicationListeners()`  and use `ApplicationListener<?> listener =beanFactory.getBean(listenerBeanName, ApplicationListener.class)` to obtain the listener is the proxied `EventDemoListener` object.

5. At this time, there are only unproxyed `EventDemoListener` object in the `applicationListeners` container, so the proxied `EventDemoListener` object will be added to the final returned allListeners collection, as shown in the figure below, which will eventually cause the listener to be triggered twice.

    [![enter image description here][2]][2]

  [1]: https://i.stack.imgur.com/lz51c.png
  [2]: https://i.stack.imgur.com/oZ8OA.png

kriegaex:
In `AbstractApplicationEventMulticaster.retrieveApplicationListeners`, despite best efforts to avoid it, unwrapped proxies (singleton targets) can end up in the list of programmatically registered listeners. In order to avoid duplicates, we need to find and replace them by their proxy counterparts, because if both a proxy and its target end up in `allListeners`, listeners will fire twice.

Fixes #28283.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/event/AbstractApplicationEventMulticaster.java
+++ b/spring-context/src/main/java/org/springframework/context/event/AbstractApplicationEventMulticaster.java
@@ -262,6 +262,24 @@ public abstract class AbstractApplicationEventMulticaster
 					if (supportsEvent(beanFactory, listenerBeanName, eventType)) {
 						ApplicationListener<?> listener =
 								beanFactory.getBean(listenerBeanName, ApplicationListener.class);
+
+						// Despite best efforts to avoid it, unwrapped proxies (singleton targets) can end up in the
+						// list of programmatically registered listeners. In order to avoid duplicates, we need to find
+						// and replace them by their proxy counterparts, because if both a proxy and its target end up
+						// in 'allListeners', listeners will fire twice.
+						ApplicationListener<?> unwrappedListener =
+								(ApplicationListener<?>) AopProxyUtils.getSingletonTarget(listener);
+						if (listener != unwrappedListener) {
+							if (filteredListeners != null && filteredListeners.contains(unwrappedListener)) {
+								filteredListeners.remove(unwrappedListener);
+								filteredListeners.add(listener);
+							}
+							if (allListeners.contains(unwrappedListener)) {
+								allListeners.remove(unwrappedListener);
+								allListeners.add(listener);
+							}
+						}
+
 						if (!allListeners.contains(listener) && supportsEvent(listener, eventType, sourceType)) {
 							if (retriever != null) {
 								if (beanFactory.isSingleton(listenerBeanName)) {
--- a/spring-context/src/test/java/org/springframework/context/event/ApplicationContextEventTests.java
+++ b/spring-context/src/test/java/org/springframework/context/event/ApplicationContextEventTests.java
@@ -39,6 +39,11 @@ import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.context.ApplicationListener;
 import org.springframework.context.PayloadApplicationEvent;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.event.test.self_inject.MyApplication;
+import org.springframework.context.event.test.self_inject.MyEventListener;
+import org.springframework.context.event.test.self_inject.MyEventPublisher;
+import org.springframework.context.support.AbstractApplicationContext;
 import org.springframework.context.support.GenericApplicationContext;
 import org.springframework.context.support.StaticApplicationContext;
 import org.springframework.context.support.StaticMessageSource;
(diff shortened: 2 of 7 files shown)


kriegaex:
Hello. I am not sure who is going to be able to look into this, maybe @jhoeller or @sbrannen? Is there anything I can or need to do in order to facilitate this being reviewed? I know there are many open bugs in Spring, but only maybe 10% of them have actual PRs attached. Maybe there is a way to expedite this and get it off the table. If I just should be waiting because work overload or Easter holidays, feel free to scold me, I would not mind. At least I would know that this was noticed and I just need to wait. 🙂

AgataKowal:
Hi @kriegaex 
Is there a chance for this PR to be ever merged? Are you aware if this problem has been addressed in any other PRs? It was not easy to find this problem and I was happy to see the solution... I have bumped into the same problem and it causes indexing issues (the event that should always be received only once was received twice). I am wondering how to handle it externally, without changes in the spring core code.

kriegaex:
@AgataKowal, I am not a committer, just a contributor. Therefore, I have no idea why this is still "in triage" and has never been commented on, not to mention merged. I even directly addressed two developers who previously committed code in this area. There is not much more I can do, only wait and hope.

sbrannen (changes requested):
Hi @kriegaex,

Apologies for the belated review. This one slipped through the cracks.

Can you please introduce a test that fails before the change and passes afterwards?

Thanks

kriegaex:
@sbrannen, I will be travelling a lot in the next few days, but try to see what I can do. It was long ago, but my comments on the related bug should give me enough information to reproduce it. Can you please point me to a directory with similar tests, so I can learn what your usual tests look like. I guess we are talking about an integration test, don't we?

sbrannen:
> Can you please point me to a directory with similar tests, so I can learn what your usual tests look like. I guess we are talking about an integration test, don't we?

Please take a look at `org.springframework.context.event.AbstractApplicationEventListenerTests` and the other test classes in the same package in the `spring-context` module. If that doesn't help, let us know.

kriegaex:
@sbrannen: I just rebased on the main branch locally. Now, IDEA cannot even import the Gradle build anymore. Unfortunately, I am a Maven person and do not speak Gradle. How can I build this thing without JDK 21?

kriegaex:
@sbrannen, unfortunately, I am not an active Spring user, just an AOP extert trying to help where I can. My debugging skills were enough to fix this issue, but I really do not comprehend how those tests work and how I can add my test case to the test bed in the right way. Hoewever, I hope that you can. 

Please, take a look at temporary commit https://github.com/spring-projects/spring-framework/pull/28322/commits/294be230123c1e9e36862882234d4485bdbdb547. If you run `org.springframework.context.event.i28283.DemoApplication` from the test directory with VM option `-ea` or `-enableassertions` on the JVM command line, the assertion will fail before and pass after the commit fixing the problem. That should give you an idea about where and how to add a regression test to one of the existing classes. Feel free to check out and push to this PR. After the test has been implemented, we can drop the commit containing the reproducer.

kriegaex:
OK, I took some more time to convert my reproducer into a simple test which fails before the fix and passes afterwards, see commit https://github.com/spring-projects/spring-framework/pull/28322/commits/0c0b93a58e81bf7a6c1224768deaf4ac099e3e63. @sbrannen, I think, you can review the PR now.

kriegaex:
@sbrannen, can we please get this off the table after 1.5 years and before I forget the whole context again and have to think myself into the problem one more time? Like this time, when trying to remember what it was all about, how I debugged it, and spending more time finding out how to write a test for it than fixing the actual bug? I know you must be busy, but please let us put this behind us, and please also port it back to all other versions which still get updates. Thank you.

sbrannen (approved):
Thanks for adding the failing test. That helps a lot!

kriegaex:
@jhoeller, is anything stopping you from porting back this fix to the 6.0.x and 5.3.x branches? Even in 5.2.x , I saw a commit as recent as July, so maybe that would be a candidate, too. But the former two definitely are, I guess.

PR #28332, opened 2022-04-13
Use "instanceof pattern matching" in sample in reference docs

jjank:


sbrannen:
Although this is a very minor change, I've merged it into `main`.

I was also surprised that there's only one example using `instanceof` (aside from the SpEL examples).

Thanks

Issue #25152, opened 2020-05-28
Make use of custom types configurable in YamlProcessor


sbrannen:
## Overview

Since SnakeYaml allows for custom types to be loaded when they are specified inside the YAML document, we should make it possible to configure the types that can be constructed.

## Proposal

Provide a mechanism in `YamlProcessor` for configuring the supported custom types for the `Yaml` instance created in `org.springframework.beans.factory.config.YamlProcessor.createYaml()`.

For example, a new `void setSupportedCustomTypes(Class<?>... types)` method could be added to `YamlProcessor`. If no supported types are configured, the `Yaml` instance would be created as it is currently, without any restrictions. If supported types have been configured, the `Yaml` instance would be created with restrictions in place. The [`TypeLimitedConstructor`](https://github.com/spring-projects/spring-boot/blob/a84b7bccf1d95b2e017be445f4cfcda438737d09/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/YamlJsonParser.java#L53-L69) recently introduced in Spring Boot can serve as inspiration for the latter.

## Related Issues

- https://github.com/spring-projects/spring-boot/issues/21596

michaldo:
I would like to know purpose of this PR.
I'm going to replace snakeyaml suporting YAML 1.1 with snakeyaml-engine supporting YAML 1.2. But API is different and `setSupportedCustomTypes` is not easy replaceable.

I see that Spring Boot defines `OriginTrackedYamlLoader`, a subclass of `YamlProcessor`. It means clients can modify  YamlProcessor - especially supported types - without explicite API.

PR #28349, opened 2022-04-15
Switch to Yaml 1.2 with snakeyaml-engine

michaldo:
This PR replace snakeyaml with snakeyaml-engine in order to support YAML 1.2
Version 1.2 solves among others Norway problem https://hitchdev.com/strictyaml/why/implicit-typing-removed, mentioned in https://github.com/spring-projects/spring-boot/issues/17113

Please note that to mitigate the Norway problem, Spring Boot team decided to use apostrophe in every documentation snippet: https://github.com/spring-projects/spring-boot/issues/28709/. That decreases documenation readability.

The following page compares versions 1.1 and 1.2: https://yaml.readthedocs.io/en/latest/pyyaml.html

I drop off 'supportedTypes' support. The feature was added in https://github.com/spring-projects/spring-framework/issues/25152. I asked question for purpose, but I didn't get answer.

I decide to drop because:
1. Supported types are not directly provided by snakeyaml-engine
2. Clients - for example Spring Boot OriginTrackedYamlLoader - still may modify yaml setup and implement supported types
3. Supported types are not documented
4. I think that supported types is wrong idea. Parsing YAML should be simple, conversion should be implemented in application code.


linghengqian:
YAML 1.1 and YAML 1.2 are two different sets of specifications, which is why SnakeYAML separates the two components. Is there an issue to discuss such a big change in advance?

michaldo:
Main reason to change in ambigous conversion unquoted Yes, No, On, Off.
I do not think the change is big. Contrary, I think the change will be transparent to almost all users.

linghengqian:
Overall, there are too many unit tests to fix – a change that completely destroys downstream use.

michaldo:
I see. I will fix failing test

wt., 7 cze 2022 o 04:59 Ling Hengqian ***@***.***> napisał(a):

michaldo:
@linghengqian I merged main branch to PR and failing tests disappeared

linghengqian:
CC @sbrannen

asomov:
@michaldo I tried to move to SnakeYAML Engine also, but I found a dependency with Spring Boot. I think it should be taken together.

PR #28351, opened 2022-04-16
Make `BeanThatListens`, `TestApplicationListener`, `ResourceUrlProvider` threadsafe

dugenkui03:
The listener could be invoked concurrently by multiple thread, we should make them threadsafe.

https://github.com/spring-projects/spring-framework/blob/f40a391916ed6c1f9e1130638a0bf19479e514dd/spring-context/src/main/java/org/springframework/context/event/SimpleApplicationEventMulticaster.java#L40
>By default, all listeners are invoked in the calling thread.

### DIFF
--- a/spring-context/src/testFixtures/java/org/springframework/context/testfixture/beans/BeanThatListens.java
+++ b/spring-context/src/testFixtures/java/org/springframework/context/testfixture/beans/BeanThatListens.java
@@ -17,6 +17,7 @@
 package org.springframework.context.testfixture.beans;
 
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


jhoeller:
While you got a point that it is generally a good idea to make listener implementations thread-safe (in terms of expecting concurrent invocations), two of the three classes affected here are test classes that are used in synchronous test setups. The third one is a `ContextRefreshedEvent` listener which is guaranteed to be invoked within a strong startup/shutdown monitor, so the listener does not benefit from its own internal synchronization either.

Issue #28305, opened 2022-04-07
Mariadb needs to be treated as an independent db type.


ben-enfuse-io:
Currently Spring Framework treats mariadb as a version of MySql. It needs to treat mariadb as its own type. I found this while working an identical issue in Spring Cloud Task.

snicoll:
Closing in favor of PR #28355

Issue #25811, opened 2020-09-24
Add MariaDB to the supported database products for procedures and functions


queueseven:
After switching the database driver from `mysql-connector-java` to `mariadb-java-client` in one of our applications, stored procedures called with `SimpleJdbcCall` no longer worked and failed with an exception like `java.sql.SQLSyntaxErrorException: Incorrect number of arguments for PROCEDURE foo.bar; expected 12, got 0`.

After some investigation I noticed that the `CallMetaDataProviderFactory` has a list called `supportedDatabaseProductsForProcedures` (which contains `MySQL`, but not `MariaDB`) so `accessProcedureColumnMetaData` was set to `false` and `initializeWithProcedureColumnMetaData` was not called on the `CallMetaDataProvider`; thus no parameters where added to the procedure call, resulting in the `SQLSyntaxErrorException`.

So I think `MariaDB` should be added to the `supportedDatabaseProductsForProcedures`.

jhoeller:
As with #24443, we're going to add this to 5.2.x (5.2.10) and also backport it to 5.1.x (5.1.19). Thanks for raising it!

jhoeller:
As pointed out in https://github.com/spring-projects/spring-framework/commit/392ad0999076ed1158ba9139a0a71e5a298474de, we should add MariaDB to `supportedDatabaseProductsForFunctions` (since MySQL is also there).

PR #28355, opened 2022-04-18
Treat MariaDB as an independent database type

ben-enfuse-io:
…b.com/spring-projects/spring-framework/issues/28305

In this commit we added MariaDB as an independent database type as per #28305 

Thanks for taking time to look at it!


snicoll:
This looks like an oversight of #25811 - what do you think @jhoeller

snicoll:
@ben-enfuse-io thank you for making your first contribution to Spring Framework.

snicoll:
@fmbenhassine please create a separate issue where we can have that discussion.

Issue #28271, opened 2022-04-01
Add application/graphql+json Media type and MIME type constants


bclozel:
While `MimeTypeUtils` and `MediaType` are not meant to thoroughly list all types, or even commonly used types - those classes should list types that are supported by Spring Framework or central to projects in the Spring portfolio.

Given spring-projects/spring-graphql#108, this issue is about adding the `"application/graphql+json"` type.

PR #28369, opened 2022-04-23
Add Javadoc since tags for GraphQL constants

izeye:
This PR adds Javadoc `@since` tag for GraphQL constants.

See gh-28271


snicoll:
Thanks again @izeye

PR #28374, opened 2022-04-24
Fix find RootBeanDefinition strategy in class FactoryAwareOrderSourceProvider

tommas1988:
Consider ancestor bean factories for finding `RootBeanDefinition` in class `FactoryAwareOrderSourceProvider`.

Currently the code will only find `RootBeanDefinition` for a bean name in current bean factory, and `null` if no bean definition is found for the searching bean name. Not consider any ancestor bean factories. This is the cause of the bug of [issues 2466 spring-cloud-gateway](https://github.com/spring-cloud/spring-cloud-gateway/issues/2466)


tommas1988:
rebuild

jhoeller:
Your submitted test case works fine against the latest state of the framework, so I assume something else must have been in the way there. Note that `getMergedBeanDefinition` (as called by `FactoryAwareOrderSourceProvider`) is aware of parent factories already, so such a fallback should be transparent, with no need to perform explicit traversal in `FactoryAwareOrderSourceProvider` itself.

I'll add a test along those lines to make sure we do not cause any regressions here.

Issue #26842, opened 2021-04-21
HttpStatus.resolve allocates HttpStatus.values() once per invocation


richardstartin:
I just noticed looking at profiles of some reactor based benchmarks we run to measure Datadog tracer overhead that `HttpStatus.resolve` allocates an `HttpStatus[]` once per invocation, so once per response. In a very modest throughput benchmark (~600rps) this is allocating 1MB/s just to resolve the integer status code to the `HttpStatus` enum value.
<img width="997" alt="Screenshot 2021-04-21 at 22 35 38" src="https://user-images.githubusercontent.com/16439049/115624009-e81d8980-a2f1-11eb-8e95-024611d0009d.png">

I tracked the spring framework code down [here](https://github.com/spring-projects/spring-framework/blob/5b1ab31559798df83f1e8d54d2b754f12c69c14e/spring-web/src/main/java/org/springframework/http/HttpStatus.java#L553).

This could be fixed by caching `HttpStatus.values()` in an array, which could be iterated over as many times as one likes without any further allocation.

poutsma:
I think we can do better than a cached array, and might as well start using a `Map<Integer, HttpStatus>`.

richardstartin:
Boxing HTTP status codes can be problematic because the common ones all fall outside the range of the `Integer` cache. We use [this](https://github.com/DataDog/dd-trace-java/blob/master/internal-api/src/main/java/datadog/trace/api/cache/RadixTreeCache.java) data structure for representing mappings between small, sparse finite sets of integers (such as status codes or ports) and arbitrary things, which works very well for tracing, and something similar may be applicable here.

poutsma:
Thank you for help.

I am not sure if we want to introduce a new data structure into the framework to resolve this. I'll make a copy of the array instead, as originally suggested.

richardstartin:
That makes sense - we don't record any time spent in `HttpStatus.resolve` at all, it only shows up for the allocations.

PR #28404, opened 2022-05-01
Caching HttpStatus/Series instances by int key

dfa1:
Reusing same solution of org.springframework.web.socket.sockjs.transport.TransportType to avoid a loop and allocations of Enum.values().

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpMethod.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpMethod.java
@@ -17,7 +17,7 @@
 package org.springframework.http;
 
 import java.io.Serializable;
-import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


dfa1:
@poutsma hello, just noticed that you did a nice commit to avoid some allocations in HttpStatus (7f1062159ee9926d5abed7cadc2b36b6b7fc242e). This doing same for HttpStatus.Series + some other minor fixes. What do you think?

poutsma:
I had the same idea of using a `Map<Integer, HttpStatus>`, but was convinced by @richardstartin that this was a bad idea because the common status codes fall outside the range of the `Integer` cache. See https://github.com/spring-projects/spring-framework/issues/26842#issuecomment-824717247.

So I am afraid I am going to have to decline this PR, as all the other changes seem to be cosmetic.

poutsma:
I had the same idea of using a `Map<Integer, HttpStatus>`, but was convinced by @richardstartin that this was a bad idea because the common status codes fall outside the range of the `Integer` cache. See https://github.com/spring-projects/spring-framework/issues/26842#issuecomment-824717247.

So I am afraid I am going to have to decline this PR, as all the other changes seem to be cosmetic.

PR #28422, opened 2022-05-05
Fix BindingResult error when ModelAttribute has custom name in WebFlux

bougar:
I guess this part of the code is trying to get the `ModelAttribute` annotation from the `BindingResult` parameter itself instead of the previous one as it should be. This is causing `ann` variable to be always `null` and therefore ignoring the value of the `@ModelAttribute` annotation.


sbrannen:
Good catch!

Thanks

PR #28426, opened 2022-05-08
Polish Javadoc and inline comments

yuhangbin:
Fix some typos and polish trivial stuff.


spring-test/src/main/java/org/springframework/test/context/ContextConfiguration.java
```diff
@@ -89,7 +89,7 @@
  * @see SmartContextLoader
  * @see ContextConfigurationAttributes
  * @see MergedContextConfiguration
- * @see org.springframework.context.ApplicationContext ApplicationContext
+ * @see org.springframework.context.ApplicationContext
```
sbrannen, line 92: This is intentional. Please revert that change.

spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
```diff
@@ -755,7 +755,7 @@ protected void invokeBeanFactoryPostProcessors(ConfigurableListableBeanFactory b
 		PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(beanFactory, getBeanFactoryPostProcessors());
 
 		// Detect a LoadTimeWeaver and prepare for weaving, if found in the meantime
-		// (e.g. through an @Bean method registered by ConfigurationClassPostProcessor)
+		// (e.g. through a @Bean method registered by ConfigurationClassPostProcessor)
```
sbrannen, line 758: When read as "at Bean", the use of "an" is correct.

Thus, please revert this change.

spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
```diff
@@ -112,7 +112,7 @@
  *
  * <p>Implements resource loading by extending
  * {@link org.springframework.core.io.DefaultResourceLoader}.
- * Consequently treats non-URL resource paths as class path resources
+ * Consequently, treats non-URL resource paths as class path resources
```
sbrannen, line 115: The prose here uses verbal phrases. Thus "consequently" is an adverb for "treats" and does not need a comma.

So please revert this change.

spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
```diff
@@ -284,7 +284,7 @@ public String getApplicationName() {
 
 	/**
 	 * Set a friendly name for this context.
-	 * Typically done during initialization of concrete context implementations.
+	 * Typically, done during initialization of concrete context implementations.
```
sbrannen, line 287: Please revert this change.

sbrannen (changes requested):

spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
yuhangbin: I will revert.

spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java
yuhangbin: Thank you for telling me. I will revert

yuhangbin:
Already revert.

sbrannen (approved):

sbrannen:

PR #28427, opened 2022-05-08
Use Arrays.toString for initMethodNames and destroyMethodNames arrays in AbstractBeanDefinition#toString

missingdays:
Default array `toString` implementation will not produce a readable result, `Arrays.toString` should be used instead


jhoeller:
Good point! This remained unnoticed in the course of a refactoring from one method name to a set of method names...

Issue #28409, opened 2022-05-03
Documentation needs to be updated with instructions for STOMP Client


linghengqian:
- In https://docs.spring.io/spring-framework/docs/6.0.0-M3/reference/html/web.html#websocket-stomp-enable , the documentation mentions that `JSteunou/webstomp-client` is `jmesnil/stomp-websocket` , which library's most active and maintained development successor. But actually `JSteunou/webstomp-client` stopped updating in 2018 and archived the Git repository. This means that `JSteunou/webstomp-client` will no longer be maintained, I think it should be marked as `JSteunou/webstomp-client` and `jmesnil/stomp-websocket` is no longer maintained.
- ![image](https://user-images.githubusercontent.com/20187731/166458366-7ce3e9e9-a022-443a-aafc-becf4ec31a43.png)
- ![image](https://user-images.githubusercontent.com/20187731/166458975-e04152c0-8b7f-48a2-81a4-64044f00c86c.png)

PR #28428, opened 2022-05-09
Documentation updated with instructions for STOMP Client

xixingya:
Hello, this pr may update the issue https://github.com/spring-projects/spring-framework/issues/28409


linghengqian:
An additional topic is, should we point to another client being maintained?

xixingya:
> An additional topic is, should we point to another client being maintained?

sure,but I am not found another client,if you find,Welcome to point out

linghengqian:
> > An additional topic is, should we point to another client being maintained?
> 
> sure,but I am not found another client,if you find,Welcome to point out
> 

- I revisited https://stomp.github.io/implementations.html and noticed `stomp-js/rx-stomp` (refer to https://github.com/stomp-js/rx-stomp/ ) and `stomp-js/stompjs` (refer to https://github.com/stomp-js/stompjs/ ) is actively maintained under the same org. 

- I'm not really sure what `RxJS` means, but if it's the same context as `RxJava`, then obviously `stomp-js/stompjs` is not as well maintained as `stomp-js/rx-stomp`.

snicoll:
@xixingya thanks for the PR but we'd prefer to revisit this part of the doc in more details.

PR #28433, opened 2022-05-09
Make inner classes static when feasible

missingdays:
A static nested class does not keep an implicit reference to its enclosing instance.

This prevents a common cause of memory leaks and uses less memory per instance of the class.


jhoeller:
Half of the sources affected only show up in main, so this PR seems more like a 6.0 M4 candidate. Maybe we could manually backport a few of the inner class definitions as they exist in 5.3.x, @sbrannen ?

sbrannen:
I've already merged it locally for `5.3.x` (excluding nonexistent classes) and `main`.

jhoeller:
Sounds good, it's definitely worth having in both branches for everything affected. There were just quite a few new AOT code spots in that PR, not too many spots that affect 5.3.x as well. We generally strive for static inner classes where possible already.

sbrannen:
This has been merged into `5.3.x` (selectively) and `main`.

Thanks

Issue #28371, opened 2022-04-23
Unable to disable decoding Path Parameter in RestController


taoj-action:
Using Spring-Boot 2.3.12, we followed the solution on https://github.com/spring-projects/spring-framework/issues/20211 and add 

`server.undertow.decode-url: false`

to disable request decoding. The request mapping is:

`/foo/{id}/bar/{value}`

After the following GET request, the request reach to REST controller.

`GET /foo/83115320/bar/test+%2B-%2F%3A%3D_`

We would expect the value inside the controller is kept as-is due to disabled decoding:
`test+%2B-%2F%3A%3D_`

However, we actually get something like the following which has been decoded:
`test++-/:=_`

I debugged into the code and the decoding happens around org.springframework.web.method.annotation:
`	Object arg = resolveName(resolvedName.toString(), nestedParameter, webRequest);`

the request URI inside webRequest has not changed.

taojwmware:
When running a unit test using @WebMvcTest(), the value is not decoded. Auto-decode only happens when running the application.

quaff:
Spring Boot use `tomcat`  by default, please ensure you are using `undertow`.

taojwmware:
Yes, we are using undertow.
```
dependencies {
    implementation 'org.springframework.boot:spring-boot-starter-actuator'
    implementation 'org.springframework.boot:spring-boot-starter-undertow'
    implementation('org.springframework.boot:spring-boot-starter-web') {
        exclude group: 'org.springframework.boot', module: 'spring-boot-starter-tomcat'
    }
```

taojwmware:
I found a workaround by customizing the UrlPathHelper. 
```
    @Override
    public void configurePathMatch(PathMatchConfigurer configurer) {
        UrlPathHelper urlPathHelper = new UrlPathHelper() {
            @Override
            public Map<String, String> decodePathVariables(HttpServletRequest request, Map<String, String> vars) {
                return vars;
            }
        };
        urlPathHelper.setUrlDecode(false);
        configurer.setUrlPathHelper(urlPathHelper);
    }
```
Although it is ideal, it works. I Hope the Spring community can come out with a better solution! Thanks!

rstoyanchev:
[...]

PR #28450, opened 2022-05-12
jdbc related new features and test cases

yinpeng0528:
The original issue can be refer as: https://github.com/spring-projects/spring-framework/issues/28371
Since there is no way to keep any of the path variable values passing through As-Is without decoding at all, just create a new function to return the original path variable values.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UrlPathHelper.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UrlPathHelper.java
@@ -669,6 +669,20 @@ public class UrlPathHelper {
 	}
 
 	/**
+	 * Decode the given URI path variables without {@link #decodeRequestString}
+	 * it is assumed the URL path from which the variables were extracted is already decoded
+	 * through a call to {@link #getLookupPathForRequest(HttpServletRequest)}. Under certain
+	 * circumstance such as passing special character, it is the only way to keep any path
+	 * variable values passed as is.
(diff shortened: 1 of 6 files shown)


yinpeng0528:
Add test cases for JdbcTemplateQueryTests

spring-web/src/main/java/org/springframework/web/util/UrlPathHelper.java
```diff
@@ -668,6 +668,20 @@ public Map<String, String> decodePathVariables(HttpServletRequest request, Map<S
...
+	 * through a call to {@link #getLookupPathForRequest(HttpServletRequest)}. Under certain
+	 * circumstance such as passing special character, it is the only way to keep any path
+	 * variable values passed as is.
+	 * @param vars the URI variables extracted from the URL path
+	 * @return the same Map
+	 */
+	public Map<String, String> decodePathVariablesForSpecialCharacter( Map<String, String> vars) {
+		this.setUrlDecode(false);
+		return vars;
+	}
```
rstoyanchev, line 683: `UrlPathHelper` is a singleton and is used to determine the lookup path for all incoming requests. Changing the `urlDecode` here would affect all requests, and potentially break handling mappings.

rstoyanchev (commented):
I don't see how the JDBC changes are related to the decoding or URI path variables, but in any case, #28371 is now closed, and the changes here are not how we would address that issue.

PR #28453, opened 2022-05-12
Improve collectionFactory which allows single statement collection

BinGuo0717:
The original issue can be refer as: [#28025](https://github.com/spring-projects/spring-framework/issues/28025)
I improve CollectionFactory to allow for single statement collection creation by adding a new function with collection and collectionType.


sbrannen:
The corresponding issue is already assigned to me. 

In addition, please do not copy other people’s work and claim it as your own.

PR #28459, opened 2022-05-14
Add assertions to the test case of nullReturningBeanPostProcessor

nealshan:
Hey Guys,

I would like to update the test code of nullReturningBeanPostProcessor.
1. Remove the second bean BeanPostProcessor since it does not help to understand the test case.
2. Add an assertion to tell, if a BeanPostProcessor returns null, the existing bean will be returned.

### DIFF
--- a/spring-context/src/test/java/org/springframework/context/annotation/AnnotationConfigApplicationContextTests.java
+++ b/spring-context/src/test/java/org/springframework/context/annotation/AnnotationConfigApplicationContextTests.java
@@ -154,19 +154,9 @@ class AnnotationConfigApplicationContextTests {
 				return (bean instanceof TestBean ? null : bean);
 			}
 		});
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @nealshan,

Thanks for the proposal!


> 1. Remove the second bean BeanPostProcessor since it does not help to understand the test case.

The 2nd BPP was actually meant to demonstrate the expected behavior, but I suspect the original developer forgot to complete the implementation.

In light of that, I've updated the test method slightly differently in f2fe7f335c3b55cb2a2f9bd4cada56dc2ad35633.

> 2. Add an assertion to tell, if a BeanPostProcessor returns null, the existing bean will be returned.

That is definitely useful (and was missing), and I've included it in my commit.

Thanks for bringing this to our attention.

PR #28472, opened 2022-05-18
Improve test coverage of RdbmsOperation

JiayiLi1999:
1. getSql()
2. tooManyMapParameters()


mahadevshindhe (approved):

snicoll:
Thank you for making your first contribution to Spring Framework, @JiayiLi1999.

I didn't include the SQL test as it's testing the getter and we prefer to test the outcome of setting the property. I've also noticed that some tests were not testing what they were supposed to (including the test you added) as the `DataSource` was not set (so it was failing, but not for the reason that the test was expecting).

PR #28478, opened 2022-05-19
Testcases

binguo07:
Add two new test cases for jdbc part


mdeinum:
Contains code related to https://github.com/spring-projects/spring-framework/pull/28454 and not only 2 JDBC tests... You might want to clean that up.

PR #28480, opened 2022-05-19
add two more test cases for jdbc

binguo07:
1.  I add test case for getting ket with different key rows in KeyHolderTests.java
2. I add test case for testing query for list in JdbcTemplateQueryTests.java

### DIFF
--- a/spring-jdbc/src/test/java/org/springframework/jdbc/core/JdbcTemplateQueryTests.java
+++ b/spring-jdbc/src/test/java/org/springframework/jdbc/core/JdbcTemplateQueryTests.java
@@ -278,6 +278,21 @@ public class JdbcTemplateQueryTests {
 	}
 
 	@Test
+	public void testQueryForListWithMultipleElements() throws Exception {
+		String sql = "SELECT AGE FROM CUSTMR WHERE ID < 4";
+		given(this.resultSet.next()).willReturn(true, true,true, false);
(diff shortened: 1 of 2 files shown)


spring-jdbc/src/test/java/org/springframework/jdbc/core/JdbcTemplateQueryTests.java
```diff
@@ -277,6 +277,21 @@ public void testQueryForLongPrimitive() throws Exception {
 		verify(this.statement).close();
 	}
 
+	@Test
+	public void testQueryForListWithMultipleElements() throws Exception {
```
snicoll, line 281: This is already tested by `doTestQueryForListWithArgs`

spring-jdbc/src/test/java/org/springframework/jdbc/support/KeyHolderTests.java
```diff
@@ -113,4 +113,17 @@ void getKeysWithMultipleKeyRows() {
 			.withMessageStartingWith("The getKeys method should only be used when keys for a single row are returned.");
 	}
 
+	@Test
```
snicoll, line 116: Unfortunately, I don't think it's exercising a different code path than `getKeysWithMultipleKeyRows`

snicoll (commented):
Thanks for submitting your first PR to the Spring Framework. Unfortunately, I don't think we should merge this, see below.

PR #28488, opened 2022-05-19
Clarify docs on JNDI properties in Servlet environment

arend-von-reinersdorff:
- Clarified that JndiPropertySource is always enabled. Previous statement "can optionally enable" might have been before #13189
- Mentioned global off switch JndiLocatorDelegate.IGNORE_JNDI_PROPERTY_NAME in StandardServletEnvironment


snicoll:
Thanks again @arend-von-reinersdorff

arend-von-reinersdorff:
Thanks for merging @snicoll :-)

You restored the word "including" in this phrase:

> is populated with additional default property sources including servlet config, servlet context parameters, and a JndiPropertySource

I find "including" a bit misleading: It sounds like more than the 3 mentioned property sources could be added by StandardServletEnvironment. But this is not the case.

PR #28493, opened 2022-05-20
Correctly detect Optional return type for `@HttpExchange` methods

wonwoo:


sbrannen:

Good catch!

PR #28498, opened 2022-05-21
Fix attribute return types in `@DeleteExchange`, `@PatchExchange`, and `@PutExchange`

eddumelendez:
`value` and `url` doesn't match with `HttpExchange` return type.


sbrannen:

Thanks!

p.s. I fixed `@PutExchange` in ea9b1fccbd06c8ddc672cbb2a5917c2105ef3b4c, too.

PR #28503, opened 2022-05-23
Support coroutines suspending in @HttpExchange interface

wonwoo:
Usage example:

```kotlin
interface SuspendService {

	@GetExchange
	suspend fun execute()

}
```


rstoyanchev (changes requested):
Thanks for the pull request. However, there are a couple of return value types, `Flux<T>` and `Mono<Flux<T>>`,  that are missing from the tests, to make this change complete. Would you like to add that as well?

PR #28515, opened 2022-05-24
Support Kotlin suspending functions with @Scheduled

wine-area:
This commit provides a scheduled task to support coroutines, but I'm not sure how to do a full unit test


sdeleuze:
Thanks for this contribution. After reviewing this PR with @simonbasle, I think that's better to reimplement it as a commit part of #29924. So I will close this PR, and you can track #29924 for the availability of Coroutines support for `@Scheduled`.

Issue #28516, opened 2022-05-24
Bean definition contribution includes attributes that are not used at runtime


snicoll:
A typical `@Configuration`-annotated bean include a `org.springframework.context.annotation.ConfigurationClassPostProcessor.configurationClass` attribute in generated code. This attribute is useless at runtime and should not be contributed.

In Spring Native, we took the decision of not writing those attributes and that didn't seem to matter so I don't know how much of those attributes are required.

cc @philwebb

snicoll:
Another option would be to find a way to opt-in for certain attributes. I think that keeping the feature and letting him custom implementations drive what they need is interesting.

I am not too sure if we need to exclude an attribute or include a specific attribute (with, perhaps a list of well-known ones we want to include/exclude). In Spring Native we don't use this at all and it's been going ok so far but I am sure we'll hit a problem at some point.

snicoll:
Moving to M6 as such an optimization of code generation is not a blocker for M5, even if it would have been nice to finalize this part of the API.

snicoll:
We don't really know what is mandatory and what isn't. It would be nice to make a decision so that teams can test their integration based on what we intend to do for 6.0

snicoll:
We've decided to stop writing those attributes and get feedback on actual use of the attributes, if any. Based on that we can reconsider our support before GA.

PR #28538, opened 2022-05-29
Bean definition contribution includes attributes that are not used at…

rilmay:
Hello! This commit fixes issue #28516 by removing mechanism for adding attributes to dynamically compiled beans. I think that attributes in compiled beans are not relevant, so this solution may do the thing. Other solution may be implementing filter to automatically filter attributes with "PostProcessor" string in it as they are used internally and don't really contribute in runtime. What do you think?


snicoll:
Thanks for the PR. I don't think that removing the feature altogether is what we should aiming at. We're still discussing in the issue so we're not really ready to accept a contribution for this. Thanks anyway!

PR #28540, opened 2022-05-30
Undeprecate PathMatchConfigurer::isUseTrailingSlashMatch

vpavic:
This commit removes deprecation from `PathMatchConfigurer::isUseTrailingSlashMatch` that was accidentally introduced in 5225a574.


vpavic:
Hmm, my observation that `PathMatchConfigurer::isUseTrailingSlashMatch` deprecation was accidental is based on its usage in `WebMvcConfigurationSupport` and the fact that there's no apparent replacement.

So you're saying it was deprecated with the intent to reduce the visibility to package private (and not replacing the method with something else)? In that case my PR is indeed invalid. But if you're reducing the visibility I think you still likely need to drop to deprecation.

rstoyanchev:
Yes that was the intent, to stop exposing them as a public API. Once they are protected, the deprecations can be be dropped and they would continue to be used internally.

bclozel:
Closing as this change was intended.

PR #27818, opened 2021-12-15
Allow `@Async` qualifier to be declared with a placeholder or SpEL expression

terminux:
Hi, I submitted this PR to solve #27775 based on @sbrannen ’s [comment](https://github.com/spring-projects/spring-framework/issues/27775#issuecomment-987670815).

Closes gh-27775

PR #28549, opened 2022-05-31
Optimize SpEL and property placeholder support for @Async qualifiers

dawn9117:
optimize spel in AsyncExecutionAspectSupport


dawn9117:
> spel is already supported.

Okay, but such as @Async (" ${executor.name:} ") doesn't work  when executor.name not configured.
Additionally EmbeddedValueResolver can be reused.

sbrannen:
### Related Issues

-  #27818

spring-aop/src/main/java/org/springframework/aop/interceptor/AsyncExecutionAspectSupport.java
```diff
@@ -167,6 +174,9 @@ protected AsyncTaskExecutor determineAsyncExecutor(Method method) {
 		if (executor == null) {
 			Executor targetExecutor;
 			String qualifier = getExecutorQualifier(method);
+			if (stringValueResolver != null && StringUtils.hasLength(qualifier)) {
+				qualifier = stringValueResolver.resolveStringValue(qualifier);
```
sbrannen, line 178: This change currently [fails the build](https://ci.spring.io/builds/190047#L62728b6d:254:256).

```suggestion
			if (this.stringValueResolver != null && StringUtils.hasLength(qualifier)) {
				qualifier = this.stringValueResolver.resolveStringValue(qualifier);
```

sbrannen (changes requested):
Please make the requested change and introduce a test that fails before this PR and passes after this PR.

dawn9117:
> Please make the requested change and introduce a test that fails before this PR and passes after this PR.

I've already added it. Please review, thanks.

sbrannen:
> I've already added it. Please review, thanks.

Thanks for making the requested changes. I'll see about getting it merged in.

sbrannen (approved):

sbrannen:

Thanks 👍

PR #28550, opened 2022-06-01
Always construct new exception on error in DefaultWebClient

pokab:
Otherwise memory leak may occur due to repeated use of singleton exception. This is due to the ability for Throwable to contain a list of suppressed exceptions (and therefore being mutable) and the list to grow indefinitely.


sbrannen:
Good catch.

Thanks for submitting your first PR to the Spring Framework! 👍

PR #28559, opened 2022-06-03
Improve AssertJ usage in cache config tests

neiser:
Just some code improvements I've found while studying code in that area. It boils down to using more AssertJ features instead of hand-written code and removes `public` at some places (unnecessary with JUnit 5).

### DIFF
--- a/spring-context/src/test/java/org/springframework/cache/config/CacheAdviceParserTests.java
+++ b/spring-context/src/test/java/org/springframework/cache/config/CacheAdviceParserTests.java
@@ -17,23 +17,22 @@
 package org.springframework.cache.config;
 
 import org.junit.jupiter.api.Test;
-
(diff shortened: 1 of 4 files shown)


snicoll:
@neiser thanks for this. Going forward, please refrain from reformating the code. Running the build locally should also reveal checkstile violations. I've polished things up in [baebf71](https://github.com/spring-projects/spring-framework/commit/baebf71ed7356812820a2a31f971d60ddd168617)

neiser:
@snicoll Thanks for reformatting. I hoped that the `.editorconfig` would have setup my IntelliJ correctly, but I'll run the build next time locally before pushing!

snicoll:
>  I hoped that the .editorconfig would have setup my IntelliJ correctly

It can't catch everything. If you look at your commit, it is obvious that import statements have moved and new lines have been removed, etc.

PR #28572, opened 2022-06-06
Update reference docs to use PropertySourcesPlaceholderConfigurer

gorisanson:
`PropertyPlaceholderConfigurer` has been deprecated as of Spring 5.2, in favor of `PropertySourcesPlaceholderConfigurer`, according to the doc comments following:

https://github.com/spring-projects/spring-framework/blob/305055d6b1a42c7795891b7b389936ed80270505/spring-beans/src/main/java/org/springframework/beans/factory/config/PropertyPlaceholderConfigurer.java#L52-L54


snicoll:
Thanks again @gorisanson

PR #28583, opened 2022-06-08
Expose ThreadPoolTaskExecutor queue size and capacity for metrics

rguihard:
We're using grafana to monitor our app through java jmx exporter. We planned to configure some alerts based on those metrics, and we think it could be interesting to have at least the currentQueueCapacity for this purpose  since the queue size directly affects the app memory load.
(Having the queueCapacity seems also interesting to set up triggers whose trigger value are calculated based on the maximum capacity of the queue.)


sbrannen:
This has been merged into `5.3.x` and `main` in e386bdb82c9a6fb7e740e4321167ac4afafa1d07 and revised in 8478e8e70aafb795c7ee0491204915ee5cea08e0.

Note that there were build failures due to Checkstyle violations. I fixed those for you, but please make sure to run `./gradlew check` before pushing commits in the future to ensure that the build passes. 

Thanks for contributing your first PR to the Spring Framework! 👍

PR #28646, opened 2022-06-17
Use putAll in WebFlux for possibly better performance

vikeychen:


bclozel:
@sbrannen Could you elaborate on why this is in the triage queue? Is there something to be discussed or clarified?
Is there a reason we don't merge this right now?

venkateshvictory:
Sorry I am learning I did it only

venkateshvictory:
Can you tell me about entitymanager and about dao

rstoyanchev:
I think this could be merged indeed. There might also be a further opportunity to improve `ExtendedWebExchangeDataBinder` which overrides and repeats the base class implementation of `getValuesToBind`. I'll have a closer look.

sbrannen:
> @sbrannen Could you elaborate on why this is in the triage queue? Is there something to be discussed or clarified? Is there a reason we don't merge this right now?

I put it in the triage queue because I believed that it probably could be merged as soon as someone got a chance to review the PR in detail. When I noticed the PR I didn't have time to review it in detail, so I thought the triage queue was the easiest way to signal to the team that this is likely some low hanging fruit.

Issue #28645, opened 2022-06-17
Introduce supports() method in BeanPostProcessor


huifer:
When writing an implementation class for the `BeanPostProcessor` interface, you usually have the following code:

```java
@Override
public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
    if (bean instanceof People) {
        // do something
    }
    return bean;
}
```

The `BeanPostProcessor` interface implementation class needs to enter this code each time it passes through each if block, and some of these entries don't do anything to the bean object, but return it directly.

Maybe the `BeanPostProcessor` interface can have a `supports()` method to judge whether to support processing in the `postProcessBeforeInitialization` method.

The improved code is as follows.

```java
public interface BeanPostProcessor  {
    @Nullable
    default Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
        return bean;
    }

    @Nullable
    default Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
        return bean;
    }

    default boolean supports(Object bean, String beanName) {
        return true;
    }
}
```

sbrannen:
### Related Issues

- https://github.com/spring-projects/spring-framework/issues/21362#issuecomment-634603682

huifer:
Some of my thoughts
> https://github.com/spring-projects/spring-framework/issues/21362#issuecomment-1158850666

snicoll:
Thanks for the suggestion but `BeanPostProcessor` is a low-level interface and we'd rather not make it more complex at this time. Some improvements are scheduled in a separate issue, see https://github.com/spring-projects/spring-framework/issues/21362#issuecomment-634618632.

Issue #21362, opened 2018-05-14
Provide a mechanism to be notified that a bean has been created [SPR-16822]


spring-projects-issues:
**[Andy Wilkinson](https://jira.spring.io/secure/ViewProfile.jspa?name=awilkinson)** opened **[SPR-16822](https://jira.spring.io/browse/SPR-16822?redirect=false)** and commented

We have some code in Spring Boot that's intended to perform initialisation of a `DataSource` just before it's first made available to application code.  It currently uses a `BeanPostProcessor` and uses the post-processing of any `DataSource` as the trigger for initializing a `DataSource`. For reasons that escape me, this `DataSource` may not be the same `DataSource` as the one that is being post-processed so it performs a lookup of a bean while another bean is being post-processed. That has caused several problems.

I think we could simplify the above-described logic so that we only perform initialisation of the `DataSource` bean that is being post-processed, thereby avoiding the problematic lookup. However, this action that's targeted to a specific bean doesn't feel like a perfect fit for the broader contract of a `BeanPostProcessor`. A callback that can be registered for a specific bean would be a better fit for our needs.

We'd like any callback that may be introduced to happen after bean post-processing. The crucial thing is that the callback is invoked before the bean can be used by application code. When I was trying to figure out what to do about the referenced Boot issue, a callback at around the point where `AbstractBeanFactory.afterPrototypeCreation(String)` or `DefaultSingletonBeanRegistry.afterSingletonCreation(String)` is called seemed to be a good fit. However, this is only based on a visual inspection of the code. I haven't verified that it would work as hoped.

---

**Affects:** 5.0.6

**Reference URL:** https://github.com/spring-projects/spring-boot/issues/13042

2 votes, 8 watchers

snicoll:
@jhoeller following the recent brainstorm we had on the subject, a bean created event for top-level singleton would help us solve several issue in Spring Boot. I am tentatively assigning this one to 5.2 RC2.

jhoeller:
Any update/reminder about the state of affairs in Boot land? I suppose the workaround is still in place, and a bean-created event would be desirable towards a replacement where no extra BPP is needed in Boot?

Implementing this would require some delegation between the generic bean creation algorithm and the event facility at context level... that's not super-elegant but doable. The main remaining question for me is how to control the publication of the event since we would probably prefer to avoid it if there is no actual use for it.

snicoll:
[...]

PR #28653, opened 2022-06-18
issues-28645

huifer:
fix https://github.com/spring-projects/spring-framework/issues/28645


sbrannen:
Please refrain from submitting PRs with new features without first discussing with the team if the feature should be implemented and how it should be implemented.

With regard to #28645 and #21362, the team has not yet decided if we will even introduce such support.

In addition, this PR contains numerous changes to formatting and code element order that do not belong in this PR and would be rejected anyway.

PR #28669, opened 2022-06-21
Fix DataSourceUtils inconsistent exception handling

### LINKED ISSUE #28258: DataSourceUtils discards IllegalStateException stacktrace
https://github.com/spring-projects/spring-framework/blob/03e695ad5ff998d8da9b89d163ec4f04747bd570/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceUtils.java#L86

I'm unsure why `SQLException` is retained as a cause but `IllegalStateException` is discarded. It makes things like this very hard to debug:
```
org.springframework.jdbc.CannotGetJdbcConnectionException: Failed to obtain JDBC Connection: Session/EntityManager is closed
    at org.springframework.jdbc.datasource.DataSourceUtils.getConnection(DataSourceUtils.java:86)
    at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:330)
    at org.springframework.boot.actuate.jdbc.DataSourceHealthIndicator.getProduct(DataSourceHealthIndicator.java:122)
    at org.springframework.boot.actuate.jdbc.DataSourceHealthIndicator.doDataSourceHealthCheck(DataSourceHealthIndicator.java:105)
    ...
```

kevin0x90:
Align IllegalStateException with SQLException handling since IllegalState is only handled internally and the public facing Exception remains unchanged in either way.

Closes gh-28258

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceUtils.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -83,7 +83,7 @@ public abstract class DataSourceUtils {
 			throw new CannotGetJdbcConnectionException("Failed to obtain JDBC Connection", ex);
 		}
 		catch (IllegalStateException ex) {
-			throw new CannotGetJdbcConnectionException("Failed to obtain JDBC Connection: " + ex.getMessage());
+			throw new CannotGetJdbcConnectionException("Failed to obtain JDBC Connection", ex);
 		}
 	}
 
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/CannotGetJdbcConnectionException.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/CannotGetJdbcConnectionException.java
@@ -48,4 +48,7 @@ public class CannotGetJdbcConnectionException extends DataAccessResourceFailureE
 		super(msg, ex);
 	}
 
+	public CannotGetJdbcConnectionException(String msg, IllegalStateException ex) {
+		super(msg, ex);
+	}
 }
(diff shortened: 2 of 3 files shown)


spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceUtils.java
```diff
@@ -151,14 +148,13 @@ public static Connection doGetConnection(DataSource dataSource) throws SQLExcept
...
-	 * @throws SQLException if thrown by JDBC methods
-	 * @throws IllegalStateException if the DataSource returned a null value
+	 * @throws SQLException if thrown by JDBC methods or the DataSource returned a null value
 	 * @see DataSource#getConnection()
 	 */
 	private static Connection fetchConnection(DataSource dataSource) throws SQLException {
 		Connection con = dataSource.getConnection();
 		if (con == null) {
-			throw new IllegalStateException("DataSource returned null from getConnection(): " + dataSource);
+			throw new SQLException("DataSource returned null from getConnection(): " + dataSource);
```
snicoll, line 157: That change doesn't sound right to me. We shouldn't be throwing `SQLException` here.

snicoll (commented):
Thanks for the PR but I don't think we should be changing the nested cause the way you suggest.

spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceUtils.java
kevin0x90: I did this change because the `CannotGetJdbcConnectionException` only accepts SQLException as cause parameter. 
So what do you think of relaxing the type constraint in `CannotGetJdbcConnectionException` a bit so that also IllegalStateException could be provided either via a constructor overload or a more relaxed type specified.

kevin0x90: So i would interpret this like the PR could be just closed.

kevin0x90: Ok i will adjust the code accordingly.

snicoll:
Good stuff, thanks @kevin0x90 and @Suthman.

kevin0x90:
Thanks for accepting our small contribution @snicoll

PR #28670, opened 2022-06-21
Simplify creation of LoggingCacheErrorHandler with logged stacktrace

vpavic:
At present, creating `LoggingCacheErrorHandler` that logs stacktrace also requires supplying the logger to be used. This might be inconvenient for some users, as it requires usage of Commons Logging API.

This commit simplifies creation of such as `LoggingCacheErrorHandler` instance by adding a constructor that only accepts a boolean flag indicating whether to log stacktrace.

---

This was originally a part of:

- #28648

For consideration:


sbrannen:

Thanks

PR #28715, opened 2022-06-27
Expose headers from STOMP RECEIPT frame to registered callbacks

schnapster:
My use case for wanting access to the headers in receipt callbacks is to communicate additional information about the result of processing of a message back to the client.

Relevant parts from the STOMP docs: https://stomp.github.io/stomp-specification-1.2.html#RECEIPT
```
RECEIPT
A RECEIPT frame is sent from the server to the client once a server has successfully
processed a client frame that requests a receipt. A RECEIPT frame MUST include
the header receipt-id, where the value is the value of the receipt header in the frame
which this is a receipt for.

RECEIPT
receipt-id:message-12345

^@
A RECEIPT frame is an acknowledgment that the corresponding client frame has
been processed by the server. Since STOMP is stream based, the receipt is also
a cumulative acknowledgment that all the previous frames have been received by
the server. However, these previous frames may not yet be fully processed. If the
client disconnects, previously received frames SHOULD continue to get
processed by the server.
```

Please let me know if I can improve anything or if you think this requires more discussion up front =)

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/DefaultStompSession.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/DefaultStompSession.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


napstr:
Ahoy, I'd like to get this into the 5.x version of the framework. Happy to do the backporting, however, is it ok to have the deprecation hit an upcoming 5.3.x release or would that be a problem?

rstoyanchev (changes requested):
It's fine to add an extra receipt callback with headers, in case you need it, but if you don't, I see no reason to deprecate the existing one without the headers. Could you revise the PR along those lines?

Issue #26227, opened 2020-12-07
Avoid CGLIB proxies in websocket/messaging configuration


sdeleuze:
In order to allow support in native executables and for consistency with the changes done in the whole Spring portfolio, websocket and messaging configurations should be switched to `@Configuration(proxyBeanMethods = false)` in order to not require CGLIB proxies.

sdeleuze:
Draft commit: https://github.com/sdeleuze/spring-framework/commit/9e792a4844438a8d17f956f14f6b0cc08eb1f639

PR #28719, opened 2022-06-28
Fix issue autowiring channelExecutors and taskScheduler in spring messaging & websocket

CodeInDreams:
There are several beans require `TaskExecutor` or `TaskScheduler` without `@Qualifier`, which means inject wrong executors if we declare some executor bean `@Primary`.
I found the problem by changing taskExecutor in `WebSocketMessageBrokerConfigurer#configureClientInboundChannel`
```java
@Override
public void configureClientInboundChannel(@NonNull ChannelRegistration registration) {
    registration.interceptors(authInterceptor);
    registration.taskExecutor(myExecutor());
}
```

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/config/AbstractMessageBrokerConfiguration.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/config/AbstractMessageBrokerConfiguration.java
@@ -149,8 +149,8 @@ public abstract class AbstractMessageBrokerConfiguration implements ApplicationC
 
 
 	@Bean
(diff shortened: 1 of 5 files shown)


CodeInDreams:
> Hi @CodeInDreams,
> 
> Congratulations on submitting your first PR ever on GitHub!
> 
> Unfortunately, this is a breaking change. Furthermore, it reverts the intentional changes made in conjunction with #26227.
> 
> In light of that, I am closing this PR.

Then we should either use `@Qualifier` or not to use autowiring. @sbrannen If agree with me, which do you prefer?

PR #28736, opened 2022-06-30
Ensure channelExecutors and taskScheduler in STOMP WebSocket config are qualified

CodeInDreams:
There are several beans require TaskExecutor or TaskScheduler without `@Qualifier`, which means inject wrong executors if we declare some executor bean `@Primary`.

As previously discussed at #28719, I use `@Qualifier` instead to avoid cglib proxies.


sbrannen:
Related Issues:

- #26227

PR #28742, opened 2022-07-01
Remove code duplication in AopUtils.isAopProxy() to improve readability

waahhh:


sbrannen:
Thanks for the suggestion.

However, the current arrangement is intentional and optimized for performance.

In light of that, we will keep the limited code duplication in place.

PR #28751, opened 2022-07-03
Simplify @EneabledIfRuntimeHintsAgent support

sbrannen:
Since the `RuntimeHintsAgentCondition` is based solely on the result of invoking a `boolean` static method, there is no need to implement a custom `ExecutionCondition`. For such use cases, the `@EnabledIf` support in JUnit Jupiter is sufficient.

This commit therefore replaces the custom `RuntimeHintsAgentCondition` with use of `@EnabledIf` as a meta-annotation.

Please note that this PR also removes the `@Tag("RuntimeHintsTests")` declaration. If that is still needed for particular use cases, that can be retained.


sbrannen:
@bclozel, since you just recently introduced `RuntimeHintsAgentCondition`, I wanted to run this by you (in the form of a PR) before pushing to `main`.

sbrannen:
By the way, if you're concerned about reduced diagnostics as a result of this change, I just pushed https://github.com/junit-team/junit5/commit/73d868ce0f09ddb225882062afb174a263dfa7c7 to address that.

spring-core-test/src/main/java/org/springframework/aot/test/agent/EnabledIfRuntimeHintsAgent.java
```diff
@@ -51,13 +50,13 @@
...
  *
  * @author Brian Clozel
+ * @author Sam Brannen
  * @since 6.0
  */
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@ExtendWith(RuntimeHintsAgentCondition.class)
-@Tag("RuntimeHintsTests")
```
bclozel, line 60: Why remove the tag here?

sbrannen: > Sorry I missed that bit; yes, this tag is used in [our test suite configuration](https://github.com/spring-projects/spring-framework/blob/db43425bf92431d8fb4c73c4b5de3c3f75db47e8/buildSrc/src/main/java/org/springframework/build/hint/RuntimeHintsAgentPlugin.java#L44): we want to run those tests in a separate task, but avoid re-running all other tests.

OK. That makes sense then.

> This annotation is opinionated and the condition implementation was meant to be reused by other Spring projects, with a different setup. 

The annotation remains with the same semantics and effective behavior -- or at least that was my intention.

How would other Spring projects use the custom condition with a different setup?

> I see that the condition has been made package private in [279f55f](https://github.com/spring-projects/spring-framework/commit/279f55fdfa1659897086331cb960f31c627affe3)

Yes, that's the idiomatic way to implement custom conditions that are transparently configured for the user via a custom `public` annotation like `@EnabledIfRuntimeHintsAgent`.

It's rarely the case that users need to access the custom condition class itself, but if you have a use case for that please let me know.

> (and yet marked as `@since 6.0`?). 

It was introduced in 6.0, so that makes sense. `@since` tags are not only for public APIs.

> Should other Spring projects copy our annotation definition in their own project if they want a different setup? Is there a better way to achieve this?

I'm not quite sure what you're asking, but perhaps my previous comments already answer that. If not, please expound.

crazycodeon:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

bclozel:
Enjoy your holiday @crazycodeon 👋

PR #28759, opened 2022-07-05
Correctly identify MaxUploadSizeExceededException through keywords in message from Jetty 9.4.x

kacperkrzyzak:
MaxUploadSizeExceededException is recognized based on the exception message containing the words "size" and "exceed". Jetty server throws a message  ["Request exceeds maxRequestSize...".](https://github.com/eclipse/jetty.project/blob/064682b4ce57282e49a80a64b6d7a7a66fb47b28/jetty-util/src/main/java/org/eclipse/jetty/util/MultiPartInputStreamParser.java#L641)  Shouldn't such message also be treated as a MaxUploadSizeExceededException? 

The proposed change addresses the issue regarding handling errors for too large file uploads when using the Jetty 9.4.x server. Now it is possible that MultipartException is thrown instead of MaxUploadSizeExceededException when the file exceeds the size limit.


sbrannen:

Thanks, and congratulations on submitting your first PR to the Spring Framework! 👍

PR #28774, opened 2022-07-08
Support `@Value` for record injection

### LINKED ISSUE #28770: Support `@Value` on records
For context see https://github.com/spring-projects/spring-boot/issues/31603

Basically, I want to be able to do this
```
	@Component
	public record Test(@Value("${username}") String user) {}
```
right now that code fails, even tho the bean is created with the right value, because the AutowiredAnnotationBeanPostProcessor attempts to populate the record via field injection (even tho the values have already been set).

wine-area:
Add `@Value` support for record injection by disabling `populateBean()` for records, since records are immutable.

Addresses #28770

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
@@ -1352,6 +1352,10 @@ public abstract class AbstractAutowireCapableBeanFactory extends AbstractBeanFac
 	 */
 	@SuppressWarnings("deprecation")  // for postProcessPropertyValues
 	protected void populateBean(String beanName, RootBeanDefinition mbd, @Nullable BeanWrapper bw) {
+		// record has not set methods
+		if (mbd.getBeanClass().isRecord()) {
+			return;
+		}
 		if (bw == null) {
 			if (mbd.hasPropertyValues()) {
 				throw new BeanCreationException(
--- a/spring-context/src/test/java/org/springframework/context/annotation/configuration/AutowiredConfigurationTests.java
+++ b/spring-context/src/test/java/org/springframework/context/annotation/configuration/AutowiredConfigurationTests.java
@@ -43,6 +43,7 @@ import org.springframework.context.support.GenericApplicationContext;
 import org.springframework.core.annotation.AliasFor;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.Resource;
+import org.springframework.core.testfixture.stereotype.Component;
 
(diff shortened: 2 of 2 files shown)


rstoyanchev:
This aims to fix #28770.

spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
```diff
@@ -1352,6 +1352,10 @@ protected BeanWrapper autowireConstructor(
 	 */
 	@SuppressWarnings("deprecation")  // for postProcessPropertyValues
 	protected void populateBean(String beanName, RootBeanDefinition mbd, @Nullable BeanWrapper bw) {
+		// record has not set methods
+		if (mbd.getBeanClass().isRecord()) {
```
sbrannen, line 1356: This check is insufficient, since the bean definition may not have a "bean class".

In any case, I will address this after merging this PR.

So there is no need to modify this PR.

spring-context/src/test/java/org/springframework/context/annotation/configuration/AutowiredConfigurationTests.java
```diff
@@ -506,4 +516,8 @@ public TestBean testBean() throws IOException {
 		}
 	}
 
+
+	@Component
+	static record RecordBean(@Value("recordBeanName") String name) {
```
sbrannen, line 521: To access the `recordBeanName` system property, this would need to be changed to `@Value("${recordBeanName}")` or `@Value("#{systemProperties[recordBeanName]}")`.

I will address this after merging this PR, so there is no need to modify this PR.

spring-context/src/test/java/org/springframework/context/annotation/configuration/AutowiredConfigurationTests.java
```diff
@@ -233,6 +234,15 @@ void testCustomPropertiesWithGenericContext() throws IOException {
 		context.close();
 	}
 
+	@Test
+	void testValueInjectionWithRecord() {
+		System.setProperty("recordBeanName", "recordBean");
+		GenericApplicationContext context = new AnnotationConfigApplicationContext(RecordBean.class);
+		context.refresh();
```
sbrannen, line 241: FYI: a context cannot be refreshed multiple times.

This results in:

> java.lang.IllegalStateException: GenericApplicationContext does not support multiple refresh attempts: just call 'refresh' once

I will address this after merging the PR.

sbrannen:
This has been merged into `main` in 7c9fc575ff9625b21b39cf34e2a524dc8efe0ca9 and revised in 00c2c1d2a1d342eea4c6ffd1430d7d3928271e33.

Thanks

PR #27609, opened 2021-10-25
 Simplify default locale/timezone resolution in cookie/session locale resolvers

vpavic:
At present, the customization of the default locale and timezone resolution in `CookieLocaleResolver` and `SessionLocaleResolver` requires subclassing them and overriding `determineDefaultLocale` and/or `determineDefaultTimeZone` methods.

This PR simplifies resolution of the default locale and timezone resolution by introducing dedicated functions for these purposes, thus allowing the customization without needing to resort to subclassing the locale resolvers.

Additionally, there are also 2 small commits that improve some aspects of the `LocaleResolver` hierarchy:
-  Update `AcceptHeaderLocaleResolver` to extend `AbstractLocaleResolver`:
This commit updates AcceptHeaderLocaleResolver to extend AbstractLocaleResolver, which allows the removal of defaultLocale managing code in AcceptHeaderLocaleResolver.

- Update `LocaleContextResolver` to implement `LocaleResolver`:
This commit updates `LocaleContextResolver` to implement `LocaleResolver` using default methods, which simplifies `AbstractLocaleContextResolver` and aligns it more closely with `AbstractLocaleResolver`.

PR #28779, opened 2022-07-08
Revise `CookieLocaleResolver` to use `ResponseCookie`

vpavic:
This PR provides two commits that improve `CookieLocaleResolver`, as a follow-up on #27609:

1. **Simplify creation of `CookieLocaleResolver` with custom cookie name**

At present, creating `CookieLocaleResolver` with custom cookie name requires creating an instance using the default constructor and then using setter to configure the desired cookie name.

This commit introduces `CookieLocaleResolver` constructor that accepts cookie name thus allowing for a simpler creation of an instance with the desired cookie name.

2. **Refactor `CookieLocaleResolver`**

At present, `CookieLocaleResolver` extends `CookieGenerator` instead of `AbstractLocale(Context)Resolver` like other `LocaleResolver` implementations. This means it duplicates some common aspects of `LocaleResolver` hierarchy while also exposing some `CookieGenerator` operations, such as `#addCookie` and `#removeCookie`.

Additionally, `CookieGenerator`'s support for writing cookies is based on Servlet support which at current baseline doesn't support SameSite directive.

This commit refactors `CookieLocaleResolver` to make it extend `AbstractLocaleContextResolver` and also replaces `CookieGenerator`'s cookie writing support with newer and more capable `ResponseCookie`.

### DIFF
--- a/spring-webmvc/src/test/java/org/springframework/web/servlet/i18n/LocaleResolverTests.java
+++ b/spring-webmvc/src/test/java/org/springframework/web/servlet/i18n/LocaleResolverTests.java
@@ -99,6 +99,8 @@ class LocaleResolverTests {
 
 			if (localeContextResolver instanceof AbstractLocaleContextResolver) {
 				((AbstractLocaleContextResolver) localeContextResolver).setDefaultTimeZone(TimeZone.getTimeZone("GMT+1"));
+				request.removeAttribute(CookieLocaleResolver.LOCALE_REQUEST_ATTRIBUTE_NAME);
+				localeContextResolver.resolveLocaleContext(request);
 				assertThat(TimeZone.getTimeZone("GMT+1")).isEqualTo(((TimeZoneAwareLocaleContext) localeContext).getTimeZone());
 			}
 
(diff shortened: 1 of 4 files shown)


spring-webmvc/src/test/java/org/springframework/web/servlet/i18n/LocaleResolverTests.java
```diff
@@ -99,6 +99,8 @@ private void doTest(LocaleResolver localeResolver, boolean shouldSet) {
 
 			if (localeContextResolver instanceof AbstractLocaleContextResolver) {
 				((AbstractLocaleContextResolver) localeContextResolver).setDefaultTimeZone(TimeZone.getTimeZone("GMT+1"));
+				request.removeAttribute(CookieLocaleResolver.LOCALE_REQUEST_ATTRIBUTE_NAME);
+				localeContextResolver.resolveLocaleContext(request);
```
vpavic, line 103: Extending `AbstractLocaleContextResolver` uncovered some test scenarios that weren't asserted for `CookieLocaleResolver`. This workaround isn't pretty, but is needed since `CookieLocaleResolver` consumes default locale while resolving locale context.

rstoyanchev:
I've also added some improvements in `ResponseCookie` to make it more friendly for use in `CookieLocalResolver` so that it doesn't have to replicate all of its fields.

sabomichal:
It seems that introducing `cookie` as an internal property, you made the whole class non thread safe. You are mutating the internal state even when cookie value is set. Is this an intended behavior?

sbrannen:
@sabomichal, this issue was closed almost 2 years ago.

If you believe you have encountered a bug or something worth discussing, please create a new issue.

Issue #26377, opened 2021-01-12
Avoid synchronization in AbstractAspectJAdvice#calculateArgumentBindings


hnxy:
I see `AbstractAspectJAdvice#calculateArgumentBindings` is synchronized, but if first time the Advice method calculates the argument binding, the next time it can return directly without lock, I think it can prevent some lock competition. 

such as:

```java
public final void calculateArgumentBindings() {
    // The simple case... nothing to bind.
    if (this.argumentsIntrospected || this.parameterTypes.length == 0) {
	    return;
    }
    synchronized(this) {
      if (this.argumentsIntrospected || this.parameterTypes.length == 0) {
        return;
      }
      int numUnboundArgs = this.parameterTypes.length;
      Class<?>[] parameterTypes = this.aspectJAdviceMethod.getParameterTypes();
      if (maybeBindJoinPoint(parameterTypes[0]) || maybeBindProceedingJoinPoint(parameterTypes[0])) {
        numUnboundArgs--;
      }
      else if (maybeBindJoinPointStaticPart(parameterTypes[0])) {
        numUnboundArgs--;
      }
      
      if (numUnboundArgs > 0) {
      // need to bind arguments by name as returned from the pointcut match
        bindArgumentsByName(numUnboundArgs);
      }
  
       this.argumentsIntrospected = true;
    }		
}
```

leexgone:
We had the same problem！ This led to serious performance issues in our system.
I had to modify the source code locally to temporarily fix it.
We hope this problem can be fixed as soon as possible.

sbrannen:
* superseded by #28783

jhoeller:
Reopened for a simpler approach as discussed in my closing comment on #28783: We should be able to drop the synchronization completely since `ReflectiveAspectJAdvisorFactory` always performs an early call to `calculateArgumentBindings` anyway, covering our standard scenario. I'll resolve this in time for the 5.3.22 release.

leexgone:
> Reopened for a simpler approach as discussed in my closing comment on #28783: We should be able to drop the synchronization completely since `ReflectiveAspectJAdvisorFactory` always performs an early call to `calculateArgumentBindings` anyway, covering our standard scenario. I'll resolve this in time for the 5.3.22 release.

That's a great idea! I'm look forward to the new version.

PR #28783, opened 2022-07-11
Reduce synchronization scope in AbstractAspectJAdvice#calculateArgumentBindings

flyhenanren:
fix synchronized problem, issues #26377

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
+++ b/spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
@@ -367,25 +367,29 @@ public abstract class AbstractAspectJAdvice implements Advice, AspectJPrecedence
 	 * to which argument name. There are multiple strategies for determining
 	 * this binding, which are arranged in a ChainOfResponsibility.
 	 */
-	public final synchronized void calculateArgumentBindings() {
+	public final void calculateArgumentBindings() {
 		// The simple case... nothing to bind.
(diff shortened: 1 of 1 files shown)


jeooica:
We need it in 5.3.x branch.

ryl2018:
good fix

leexgone:
Great!

jhoeller:
After reviewing our arrangement there, it turns out that the `calculateArgumentBindings` synchronization is effectively unnecessary these days: We always create AspectJ advice instances with an early call to `calculateArgumentBindings` in `ReflectiveAspectJAdvisorFactory`, not requiring any synchronization for that standard scenario. The implicit lazy calculation of argument bindings is only really there for test scenarios or other custom arrangements, and there are no guarantees for thread safety defined there. Given the commonality of the standard scenario, I'll optimize for that, recommending custom scenarios to also perform an early `calculateArgumentBindings` call if necessary.

From that perspective, the approach taken in this PR won't be needed anymore. I'll reopen #26377 accordingly. Thanks for the PR, in any case!

For a review, even if we are not using that approach: The implementation in the PR seems incomplete in terms of its memory barriers. At the very least, `argumentsIntrospected` would have to be `volatile` for such a nested synchronization approach to work. Even the data structures populated by `bindArgumentsByName` would have to be guarded for thread-safe publication through the use of concurrent data structures, since otherwise it is not guaranteed that their state is visible to other threads when `argumentsIntrospected` is `true` in that early check before the synchronized block.

PR #28794, opened 2022-07-12
Fix and improve Javadoc in spring-tx

marcwrobel:


marcwrobel:
@sbrannen, thank you for you reviews. For the time being I create a PR for each module. But if it is easier for you I can make bigger PRs (with 2-3 modules in each).

sbrannen:
> But if it is easier for you I can make bigger PRs (with 2-3 modules in each).

2-3 modules per PR would help. Thanks.

PR #28802, opened 2022-07-12
Fix and improve Javadoc in spring-core and spring-context

marcwrobel:


spring-core/src/main/java/org/springframework/asm/AnnotationVisitor.java
```diff
@@ -131,7 +131,7 @@ public AnnotationVisitor visitAnnotation(final String name, final String descrip
   /**
    * Visits an array value of the annotation. Note that arrays of primitive values (such as byte,
    * boolean, short, char, int, long, float or double) can be passed as value to {@link #visit
-   * visit}. This is what {@link ClassReader} does for non empty arrays of primitive values.
+   * visit}. This is what {@link ClassReader} does for non-empty arrays of primitive values.
```
sbrannen, line 134: Please refrain from modifying classes under `org.springframework.asm`, `org.springframework.cglib`, and `org.springframework.objenesis`. Those include repackaged forks of the third-party libraries ASM, CGLIB, and Objenesis. Any refactoring to those classes should take place upstream in the originating repository. The Spring Framework will then pick up the changes when syncing with official updates of the forked third-party libraries.

----

The above is our standard policy, but don't worry about updating this PR. I'll revert all ASM changes while merging.

sbrannen:
This has been merged.

Thanks

PR #28803, opened 2022-07-12
Fix and improve Javadoc in spring-beans and spring-aop

marcwrobel:


spring-aop/src/main/java/org/springframework/aop/framework/ProxyFactoryBean.java
```diff
@@ -60,7 +60,7 @@
  * to use the "targetName"/"target"/"targetSource" properties instead.
  *
  * <p>Global interceptors and advisors can be added at the factory level. The specified
- * ones are expanded in an interceptor list where an "xxx*" entry is included in the
+ * ones are expanded in an interceptor list where a "xxx*" entry is included in the
```
sbrannen, line 63: "an" is actually correct since "xxx*" is pronounced "ex...".

sbrannen:
This has been merged.

Thanks

PR #28817, opened 2022-07-14
Fix broken link in rsocket documentation

jupiterhub:


sdeleuze:
The typo fix commit is merged since 2020, the fork is not up to date, so only the RSocket link fix is relevant, so I bring back the original title.

PR #28819, opened 2022-07-14
Replace use of the <tt> HTML tag in Javadoc

marcwrobel:
The `<tt>` HTML tag was used in HTML 4 to define teletype text. It is not supported in HTML5 and its use is discouraged (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt).


sbrannen:

Thanks

PR #28821, opened 2022-07-14
Use `{@code}` instead of HTML entities in Javadoc

marcwrobel:
Special characters, like < or >, don't need to be escaped in `{@code}` Javadoc tag (https://reflectoring.io/howto-format-code-snippets-in-javadoc/). Using `{@code}` improves readability of the Javadoc in code.


marcwrobel:
Note that this is only possible in inline code blocks because indentation and line breaks are lost with `{@code]`.

marcwrobel:
It seems checkstyle does not like tags in `{@code}` blocks, such as `{@code <async-supported>true</async-supported>}`. This PR is all about that, so I wonder if it should not be abandonned.

PR #28834, opened 2022-07-18
Make ResourcePatternHints.Builder.build public

mhalbritter:
Makes the `ResourcePatternHints.Builder.build()` method public. I assume it's an oversight that it's not public, as the class and all other methods are.


snicoll:
That's by design I am afraid. The `build` methods are package private as you get a customizer of the builder in various callbacks method and being able to call `build` at that stage is a bit odd. What are you trying to do?

mhalbritter:
Testing hints. Thanks for pointing me to `RuntimeHintsPredicates`. This PR is no longer needed.

Issue #28107, opened 2022-02-25
A request to empty file in runnable jar cause 404


dcsuibian:
Same as this issue: https://github.com/spring-projects/spring-boot/issues/16034

`org.springframework.core.io.AbstractFileResolvingResource` will consider an empty file in runnable jar is unreadable.

<img width="621" alt="image-20220225162730187" src="https://user-images.githubusercontent.com/54771493/155682660-c890a557-c936-41a6-8888-55b95bc20b88.png">

It cause different and confusing behavior after `mvn package`.

I think empty folders can indeed be considered unreadable, but files should not be the same.

Do you have any plans to fix it?

rstoyanchev:
Please, don't use images to point to code that is already on GitHub. While the image shows more, most often, other surrounding code needs to be examined too. So this is just not helpful and bloats the issue description.

rstoyanchev:
An argument about how an empty file should be treated, could be made on both sides. Considering the present behavior is now long standing, and the implementation issues described in https://github.com/spring-projects/spring-framework/issues/21372#issuecomment-453471835, it makes no sense to change this, and disrupt others who already depend on the present behavior.

I'm closing this but it can be re-opened if there is enough interest. Feel free to leave a comment with more details on the actual use case of how this happens.

PR #28850, opened 2022-07-21
Improve resource handling for empty files contained in jars

larsgrefer:
We noticed a strange behavior in WebMVC when working with empty files:

We had some empty files in a classpath location configured in `org.springframework.web.servlet.config.annotation.WebMvcConfigurer#addResourceHandlers`

When running the application locally using the IDE, the files could be _"successfully"_ accessed by the browser (it got a HTTP 200 response with content-length 0)

After packaging the application, the browser suddenly got 404 responses for the empty files.

The difference seems to be in `org.springframework.core.io.AbstractFileResolvingResource#checkReadable`.
As long as the empty files are actual files on the file system (`ResourceUtils.isFileURL(url) == true`), the corresponding ClasspathResource is considered both existent and readable.
After packaging into a jar file, the ClasspathResource is still existent, but no longer considered readable.

I know that serving empty files as static resources is not a typical use case, but I think it should still be (a) possible and (b) consistent, regardless of whether the ClasspathResource in question is an actual File or a JarFileEntry

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
+++ b/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
@@ -20,6 +20,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.JarURLConnection;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLConnection;
@@ -27,6 +28,7 @@ import java.nio.channels.FileChannel;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.StandardOpenOption;
+import java.util.jar.JarEntry;
 
 import org.springframework.util.ResourceUtils;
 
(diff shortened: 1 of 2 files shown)


JanStureNielsen:
It seems like `checkReadable` should be consistently handling empty-content, or not at all, regardless of packaging, and this inconsistency feels like a bug; so, what motivated the comment ["empty file or directory -> not considered readable..."](https://github.com/spring-projects/spring-framework/blob/55a8bdc86c14e988d31768f76b42ea0a0f8f89b4/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java#L134)? 

If empty-content is okay for `ResourceUtil`, `HttpURLConnection`, and `JarURLConnection` branches, why not okay for the rest?

bclozel:
Indeed, serving an empty resource over HTTP should not result in a 404 response.

The current behavior can be tracked to #21372, where we tried to avoid serving empty files for directory entries. In the course of fixing that problem (in 69f14a20380e8 and 616a40adb64a09), it appeared that [a long-standing JVM bug prevents us from considering folders and empty resources separately](https://github.com/spring-projects/spring-framework/issues/21372#issuecomment-453471832). @JanStureNielsen this should answer your comment.

This behavior has been in place since Spring Framework 5.1 and we've been reluctant to change things more in this area, because of risks of regression or performance issues. In the meantime, this problem has been reported multiple times in Spring Boot and Spring Framework (see #28107 for the latest instance).

I'm inclined to consider this change, which should solve the JAR case and improve consistency with the file system case. Now we're quite late in the 5.3.x cycle, so maybe pushing this to 6.0.0 would make more sense. What do you think @jhoeller @rstoyanchev ?

PR #28852, opened 2022-07-21
added missing @Nullable annotations to methods that may return null and are overridden by methods already using @Nullable

jensdietrich:
**1/4 Added `@Nullable` to return type of 
`org.springframework.asm.AnnotationVisitor::visitAnnotation`**

_Reasons:_ 
1. On inspection, the method may return null
2. The method is overridden by methods with nullable annotation (violating LSP indicating that return types should not be generalised in overridden methods, assuming that most static null checkers will use the nonull-by-default assumption):

- org.springframework.core.type.classreading.MergedAnnotationReadingVisitor::visitAnnotation
- org.springframework.core.type.classreading.MergedAnnotationReadingVisitor$ArrayVisitor::visitAnnotation

**2/4 Added `@Nullable` to return type of 
`org.springframework.asm.ClassVisitor::visitMethod`**

_Reasons:_ 
1. On inspection, the method may return null
2. The method is overridden by methods with nullable annotation (violating LSP indicating that return types should not be generalised in overridden methods, assuming that most static null checkers will use the nonull-by-default assumption):

- org.springframework.core.type.classreading.SimpleAnnotationMetadataReadingVisitor::visitMethod
- org.springframework.core.LocalVariableTableParameterNameDiscoverer$ParameterNameDiscoveringVisitor::visitMethod

**3/4 Added `@Nullable` to return type of 
`org.springframework.asm.ClassVisitor::visitAnnotation`**

_Reasons:_ 
1. On inspection, the method may return null
2. The method is overridden by methods with nullable annotation (violating LSP indicating that return types should not be generalised in overridden methods, assuming that most static null checkers will use the nonull-by-default assumption):

- org.springframework.core.type.classreading.SimpleAnnotationMetadataReadingVisitor::visitAnnotation

**4/4 Added `@Nullable` to return type of 
`org.springframework.asm.MethodVisitor::visitAnnotation`**

_Reasons:_ 
1. On inspection, the method may return null
2. The method is overridden by methods with nullable annotation (violating LSP indicating that return types should not be generalised in overridden methods, assuming that most static null checkers will use the nonull-by-default assumption):

- org.springframework.core.type.classreading.SimpleMethodMetadataReadingVisitor::visitAnnotation

Note that there are similar methods in those classes that could also be annotated as `@Nullable` based on code inspection. This PR is conservative as it only adds `@Nullable` to methods where there are multiple reasons. If this gets accepted, I can create a separate PR for those.

### DIFF
--- a/spring-core/src/main/java/org/springframework/asm/AnnotationVisitor.java
+++ b/spring-core/src/main/java/org/springframework/asm/AnnotationVisitor.java
@@ -27,6 +27,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 package org.springframework.asm;
 
+import org.springframework.lang.Nullable;
+
 /**
  * A visitor to visit a Java annotation. The methods of this class must be called in the following
  * order: ( {@code visit} | {@code visitEnum} | {@code visitAnnotation} | {@code visitArray} )*
@@ -121,6 +123,7 @@ public abstract class AnnotationVisitor {
    *     visitor is not interested in visiting this nested annotation. <i>The nested annotation
    *     value must be fully visited before calling other methods on this annotation visitor</i>.
    */
+  @Nullable
   public AnnotationVisitor visitAnnotation(final String name, final String descriptor) {
     if (av != null) {
       return av.visitAnnotation(name, descriptor);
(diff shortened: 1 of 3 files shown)


snicoll:
@jensdietrich thanks for the PR but  ASM is a 3rd party lib we shade and we don't want to modify them.

jensdietrich:
Makes sense, I should have checked for shading.

PR #28857, opened 2022-07-22
spring-orm mvp conversion to hibernate6

hduerkop:


snicoll:
@hduerkop thanks for the PR but we have no plan short term to do this so this PR is not actionable. Going forward, please create an issue first to discuss the change to avoid wasted efforts.

PR #28888, opened 2022-07-29
Fix memory leak with CglibAopProxy$ProxyCallbackFilter

vikeychen:
## Bug report
Bug report and reason is explained very well here #26266.

In the cglib side, this issue [#80](https://github.com/cglib/cglib/issues/80#issue-160623315) is the consequence of trying to solve `CallbackFilter` leak.
As vlsi said
> "weakly reachable CallbackFilters" and "classloader reuse" do not play together well.

## Solution
* This solution is to try to reduce the leak by make `advisors` and `advisorChainFactory` WeakReference , so that there should have no chance to leak 'big object'. Think about class created by cglib is also 'leaked', I think that should be acceptable.

* Another quick solution is to make `AbstractAutoProxyCreator` implements `DisposableBean`, clear `CallbackFilter` field in generated class.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
@@ -512,6 +512,20 @@ public class AdvisedSupport extends ProxyConfig implements Advised {
 	}
 
 	/**
+	 * Build a configuration-only of AdvisedSupport.
+	 */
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks for the PR and sorry it got overlooked but https://github.com/spring-projects/spring-framework/issues/26266 has been fixed in the meantime. Please give it a try if you can.

PR #28937, opened 2022-08-08
Add reflection runtime hints for OptionalValidatorFactoryBean.

OlgaMaciaszek:


sdeleuze:
@snicoll is currently trying to check if we can remove the reflection usage here and for similar patterns, so I restored the `waiting-for-triage` label.

snicoll:
Thank you very much for reporting the problem and the PR @OlgaMaciaszek. I've removed the need for reflection so the additional hint shouldn't be necessary. See #28939

PR #28958, opened 2022-08-12
Use URL and HTTP method in DefaultResponseErrorHandler

jerzykrlk:
Good evening team,

This is a followup of a previous PR from January 2021: https://github.com/spring-projects/spring-framework/pull/26480 .
Previously it was closed with a conclusion that the change bypasses a protected method:

> Given the change of protected methods I'd rather avoid this change at this stage of the 5.3.x branch. 

Now that Spring 6 is coming, do you think there's a chance to add this change?

Below is the original description of the PR:

======

The REST Client exception contains a preview of the body, which is great, and it helps a lot to quickly diagnose errors.

Now, the stacktrace looks like this:
```
Caused by: org.springframework.web.client.HttpServerErrorException$InternalServerError: 500 Server Error: [no body]
	at org.springframework.web.client.HttpServerErrorException.create(HttpServerErrorException.java:101)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:238)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:147)
```

I'd like to add the HTTP method and URL to the error, so that the stacktrace looks like this:
```
Caused by: org.springframework.web.client.HttpServerErrorException$InternalServerError: 500 Server Error after GET http://localhost:49709/status/server : [no body]
	at org.springframework.web.client.HttpServerErrorException.create(HttpServerErrorException.java:101)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:238)
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:147)
```

======

Thank you!

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java
@@ -19,6 +19,7 @@ package org.springframework.web.client;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
@@ -28,6 +29,7 @@ import java.util.function.Function;
 import org.springframework.core.ResolvableType;
 import org.springframework.core.log.LogFormatUtils;
 import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.HttpStatusCode;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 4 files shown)


rstoyanchev (commented):
Apologies this fell through again, but already too late for 6.1. We have looked into it however and discussed it. We plan to add it in 6.2.

One thing to keep in mind, the exception message should not include the full URI, in particular the query which may reveal sensitive info. 

Also, now that we have both `RestClient` and `RestTemplate`, for extra context it's worth mentioning that while `RestClient` makes it possible to use `ResponseErrorHandler`, it also exposes its own `ErrorHandler` interface that takes the request and response.

snicoll:
Thanks very much @jerzykrlk! I've polished things up a bit to streamline thing. I've also changed the structure of the error message to be more compliant with other places in the framework. It now reads:

```
500 Server Error on GET request for "http://localhost:49709/status/server": [no body]

jerzykrlk:
Hello Stéphane - thanks for accepting this!

PR #28961, opened 2022-08-13
Update StreamUtils drain and emptyInput to use JDK builtins

kilink:
Update StreamUtils.drain to use InputStream.transferTo with a null OutputStream. This avoids allocating buffers for cases where the supplied InputStream has an optimized transferTo method (e.g., ByteArrayInputStream and FileInputStream).

Additionally, update StreamUtils.emptyInput to simply call InputStream.nullInputStream.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StreamUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StreamUtils.java
@@ -16,7 +16,6 @@
 
 package org.springframework.util;
 
-import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.FilterInputStream;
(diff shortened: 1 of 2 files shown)


j3graham:
There are a few more similar opportunities in StreamUtils:
- `copyToByteArray` could use `in.readAllBytes()` instead of going through a `ByteArrayOutputStream`
-  `copy(InputStream in, OutputStream out)` could also use `transferTo`

kilink:
> There are a few more similar opportunities in StreamUtils:
> 
> * `copyToByteArray` could use `in.readAllBytes()` instead of going through a `ByteArrayOutputStream`
> * `copy(InputStream in, OutputStream out)` could also use `transferTo`

There's already a PR for that here: #27702

PR #28964, opened 2022-08-15
Adding JCStress support for ConcurrentLruCache

jabrena:
The PR improve current testing support for the class:
- https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/util/ConcurrentLruCache.java
- https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/test/java/org/springframework/util/ConcurrentLruCacheTests.java

Adding JCStress support for the class `ConcurrentLruCache`:
https://github.com/openjdk/jcstress

How to test it:
```
./gradlew clean jcstress --tests "ConcurrentLruCacheJCStressTests" -Pmodule=spring-core
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ plugins {
 	id 'com.github.johnrengelman.shadow' version '7.1.2' apply false
 	id 'de.undercouch.download' version '5.1.0'
 	id 'me.champeau.jmh' version '0.6.6' apply false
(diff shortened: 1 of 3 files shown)


snicoll:
Thanks for the PR but we prefer to avoid depending on too many external plugins in our build, especially those that are flagged as experimental.

jabrena:
Hi @snicoll, 

in reactor project & rsocket, JCstress is used, consider it :)

PR #28965, opened 2022-08-15
Increase StreamUtils.BUFFER_SIZE to 8192

kilink:
This aligns the buffer size used in StreamUtils with the buffer sized
used throughout the JDK (see InputStream, BufferedInputStream, Files).

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StreamUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StreamUtils.java
@@ -33,7 +33,7 @@ import org.springframework.lang.Nullable;
 /**
  * Simple utility methods for dealing with streams. The copy methods of this class are
  * similar to those defined in {@link FileCopyUtils} except that all affected streams are
- * left open when done. All copy methods use a block size of 4096 bytes.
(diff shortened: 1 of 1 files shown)


kilink:
FYI, it's all the buffer size used in Guava and Apache Commons.

spring-core/src/main/java/org/springframework/util/StreamUtils.java
```diff
@@ -83,7 +83,7 @@ public static String copyToString(@Nullable InputStream in, Charset charset) thr
 			return "";
 		}
 
-		StringBuilder out = new StringBuilder(BUFFER_SIZE);
+		StringBuilder out = new StringBuilder();
```
kilink, line 86: My reasoning for using the default here, is that for any output less than BUFFER_SIZE, we end up over-allocating by specifying this value. If we use the default, and the first read fills the buffer, we will resize to BUFFER_SIZE immediately anyway when we write to the StringBuilder.

bclozel:
Thanks for your first contribution @kilink , this will be in the next 6.0.0 milestone.

PR #28984, opened 2022-08-21
Apply minor polishing in `spring-aop`

Drezir:


spring-aop/src/main/java/org/springframework/aop/scope/ScopedProxyUtils.java
```diff
@@ -110,16 +110,20 @@ public static String getTargetBeanName(String originalBeanName) {
...
 	 * to the target of a scoped proxy
 	 * @since 5.1.10
 	 * @see #getTargetBeanName(String)
 	 * @see #isScopedTarget(String)
 	 */
+	@Nullable
 	public static String getOriginalBeanName(@Nullable String targetBeanName) {
 		Assert.isTrue(isScopedTarget(targetBeanName), () -> "bean name '" +
 				targetBeanName + "' does not refer to the target of a scoped proxy");
+		if (targetBeanName == null) {
```
r331, line 124: Doesn't work because this method org.springframework.aop.scope.ScopedProxyUtils#isScopedTarget

spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
```diff
@@ -268,18 +268,16 @@ public void setArgumentNamesFromStringArray(String... args) {
 						this.argumentNames[i] + "' that is not a valid Java identifier");
 			}
 		}
-		if (this.argumentNames != null) {
```
r331, line 271: We should check it
`@Nullable
	private String[] argumentNames;`

r331 (changes requested):

spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
Drezir: It looks like it cannot be null, it is always a string array

Drezir (commented):
Remove unnecessary null check

spring-aop/src/main/java/org/springframework/aop/scope/ScopedProxyUtils.java
Drezir: You are right, fixed

sbrannen:
Unfortunately many of the `apply instanceof pattern matching` changes had already been made on `main` before I merged this PR. 

I have therefore repurposed this PR for the other polishing changes it contains.

Issue #28985, opened 2022-08-21
Provide access to all configured interceptors in AbstractHandlerMapping


lizongbo:
**Versions**
Spring Version 5.3.22
**Overview**

I wanted to get all the HandlerInterceptor that had been registered in Spring MVC after the application was successfully launched, but I found that this method was protected, so it could not be obtained directly, and I had to use the reflection method to call, so it was recommended to set it to a public method.
My code was:
```java
            
RequestMappingHandlerMapping requestMappingHandlerMapping = configurableApplicationContext.getBean(RequestMappingHandlerMapping.class);

Field adaptedInterceptorsField = ReflectionUtils.findField(RequestMappingHandlerMapping.class, "adaptedInterceptors");

adaptedInterceptorsField.setAccessible(true);

List<HandlerInterceptor> hiList = (List<HandlerInterceptor>) adaptedInterceptorsField.get(requestMappingHandlerMapping);
```

I also suggest that Springboot's actuator provide endpoints to get that information。

rstoyanchev:
Please, elaborate on your reason for this request. Generally, the interceptors are configured and invoked within the `HandlerMapping`, and in that sense they are for internal use.

lizongbo:
I'm implementing our development framework based on Springboot, and since we have hundreds of business function developers, and we maintain hundreds of systems, and their Java development capabilities are high and low, I have to collect information while the program is running and automatically analyze the differences in changes in production materials. So I used a unified main class in the framework, called the start method of applicationContext in the main method, and then triggered the information collection operation in the start event, and the collected information will be written to a json file.
Then, through the devops tool, the json file in the docker container is collected and the information is recorded into the database. Before each deployment of materials to the production environment, through the devops tool to compare the current branch to the material and the materials that have been published production branches, you can quickly find the differences in changes such as filters and interceptors, and timely find possible error changes. This enables a quick review of changes to hundreds of systems.

Here's some of my sample code:
1. main class:
`public static void main(String[] args) {
ConfigurableApplicationContext configurableApplicationContext =  SpringApplication.run(XXXXServerApplication.class, args);
 configurableApplicationContext.start();
}`
2. start event call back
`
  @Order(Ordered.HIGHEST_PRECEDENCE + 6)
    @EventListener
    public void onContextStartedEvent(ContextStartedEvent event) {
 dumpAppConfigInfo(event.getApplicationContext());
}`

4. dump logic:


[...]

Issue #28134, opened 2022-03-04
LocalDataSourceJobStore for Quartz should not invoke setDontSetAutoCommitFalse(true)


lizongbo:
Quartz locks SQL using `select xxxx for update` when `useDBLocks` is `true`.
(https://github.com/quartz-scheduler/quartz/blob/master/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/StdRowLockSemaphore.java)

And `dontSetAutoCommitFalse` is default use false in `JobStoreSupport` class.
(https://github.com/quartz-scheduler/quartz/blob/master/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java)

But Spring Framework invokes `setDontSetAutoCommitFalse(true)` with `springTxDataSource`.

https://github.com/spring-projects/spring-framework/blob/68757073b0cf69c41d5c17e8abdffbc388cdbabe/spring-context-support/src/main/java/org/springframework/scheduling/quartz/LocalDataSourceJobStore.java#L101-L103

and Quartz code is:

```java
            if (!isDontSetAutoCommitFalse()) {
                conn.setAutoCommit(false);
            }
```

(code in https://github.com/quartz-scheduler/quartz/blob/master/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java, line 803)

the `executeInLock` method code is 

```java
    @Override
    protected Object executeInLock(
            String lockName, 
            TransactionCallback txCallback) throws JobPersistenceException {
        boolean transOwner = false;
        Connection conn = null;
        try {
            if (lockName != null) {
                // If we aren't using db locks, then delay getting DB connection 
                // until after acquiring the lock since it isn't needed.
                if (getLockHandler().requiresConnection()) {
                    conn = getConnection();
                }
                
                transOwner = getLockHandler().obtainLock(conn, lockName);
            }

            if (conn == null) {
                conn = getConnection();
            }

            return txCallback.execute(conn);
        } finally {
            try {
                releaseLock(lockName, transOwner);
            } finally {
                cleanupConnection(conn);
            }
        }
    }
```

(code in https://github.com/quartz-scheduler/quartz/blob/master/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreCMT.java, line 225)

In MySQL, `select for update ` need run in a transaction; see:
https://dev.mysql.com/doc/refman/5.7/en/innodb-locking-reads.html

[...]

PR #28986, opened 2022-08-21
change method to public and use quartz's default value to eanable transactionable

lizongbo:
change getMappedInterceptors and getAdaptedInterceptors to public method

#28985

use quartz's default value to eanable transaction
#28134


snicoll:
@lizongbo this PR is changing two completely unrelated things so it's very unlikely we'd accept it in its current form. What's the reason for changing Quartz's default?

lizongbo:
> @lizongbo this PR is changing two completely unrelated things so it's very unlikely we'd accept it in its current form. What's the reason for changing Quartz's default?

It was two commit, and since I wasn't familiar with mentioning Pull Requests separately, I wanted to submit the second commit as a separate pull request, but on the github‘s web page there was only a button show “view pull Requests”, and I couldn't create a new Pull Request. 
The default value of quartz is false, and after Spring is set to true, quartz is not running in the transaction when acquiring and unlocking locks, and there will be a very low probability of causing bugs when concurrently acquiring locks.
the detail info is in Issue #28134。

snicoll:
OK, let's continue the discussion on the issue please.

Issue #26381, opened 2021-01-13
Allow MockRestServiceServer to validate requests but actually obtain response via HTTP call


nightswimmings:
For the cases when using MockRestServiceServer on an integration test, I think we could benefit having some options to make it bypass certain requests that are available in the local environment (acting like a Spy rather than a Stub).

For instance, imagine that we are executing an Integration Test on an app with an endpoint that internally performs 3 requests:
  - 1 request to another endpoint of the same app (architecture considerations aside)
  - 1 request to an external endpoint that is being already mocked with a docker container according to the testing @Profile
  - 1 request to a third-party endpoint that we want to mock

If we autoconfigure MockRestServiceServer and RestTemplate with the simplest setup, we will end up with the requirement of 
mocking all requests. Thus, it would be nice to allow certain requests bypassing the expectation or at least the need for a mocked response and let them pass through as real calls.

By doing 

`mockRestServiceServer.expect(requestTo(call)); `

with no response definition, one would expect mockRestServiceServer to account the expectation but let the request pass through, but it fails instead complaining that no mocked response has been defined.

At least, IMO this should be the default behaviour when the application (not the test) performs a request against itself (localhost/127.0.0.0), or even more accurately, against a Controller of the same app

My need is described here: https://stackoverflow.com/questions/65693954/is-it-possible-to-make-some-of-my-springboottest-app-requests-to-bypass-spy-in

rstoyanchev:
`MockRestServiceServer` is a custom `ClientHttpRequestFactory` for the `RestTemplate` and when a call is made to the `RestTemplate` it needs to provide a response. So I'm not sure I understand the scenario.

BTW if you could please edit your description and provide a complete description. As per our guidelines, it's okay to refer to a SO thread, but an issue created here needs to have a self-sufficient description.

nightswimmings:
I tried a better description

rstoyanchev:
Thanks for updating the description! 

When `MockRestServiceServer` instruments the `RestTemplate` with a custom `ClientHttpRequestFactory`, the resulting `ClientHttpRequest` needs to return some kind of response. By "let the request pass through" you mean that it should execute an actual HTTP call by using whatever request factory was there before?

To do this, a `ResponseCreator` would need to re-create the `ClientHttpRequest` with the original request factory, re-populate it, and use it instead. This should be possible in an application, for example:

```java
RestTemplate restTemplate = new RestTemplate();
[...]

PR #28987, opened 2022-08-21
Introduce andPerformRequest() method in ResponseActions

edyda99:
Prior to this commit it wasn't possible to achieve real
client-side Rest using a restTemplate that was bound by MockRestServiceServer

This commit introduces andPerformRequest() method in
which it supplies responseCreator with a real Rest call provided by copying the clientHttpRequest and performing a http call using the default ClientHttpRequest used by restTemplate: SimpleClientHttpRequestFactory

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/DefaultRequestExpectation.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/DefaultRequestExpectation.java
@@ -16,12 +16,14 @@
 
 package org.springframework.test.web.client;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


edyda99:
For testing I used to assert that the exception contains "Connection refused: connect" but the pipeline was failing,
I guess because the port is opened, it's returning connection timeout instead of connection refused.
When someone reviews, I need some suggestions on the best way to test this feature.

rstoyanchev:
Apologies for not noticing this sooner, but this is a duplicate of #26381 and was done, slightly differently, in #29721. Thanks for the pull request in any case!

edyda99:
Sorry to hear that, Can you suggest any ticket for me?

Issue #29008, opened 2022-08-24
Request to improve KotlinSerializationJsonHttpMessageConverter logic in RestTemplate


meloning:
**Affects:** 5.3.22 (below 5.3.22)

---

Hello,
I was using the Kotlinx.serialization module well because of Spring registered `KotlinSerializationJsonHttpMessageConverter` through `WebMvcConfigurationSupport.addDefaultHttpMessageConverters`

However, even though I added the Kotlinx.Serialization module, it was not available in RestTemplate.
```Groovy
implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.4.0")
```

After analyzing why, only Jackson, which is the default serialization, was registered due to the **`if-else if`** logic as shown below, so Kotlinx.Serialization could not be used.

![image](https://user-images.githubusercontent.com/50299857/186432443-928f589b-1fa0-418a-a719-4cbd836a1a08.png)

I can temporarily solve it like the configuration below.
![image](https://user-images.githubusercontent.com/50299857/186434334-ccfd9f73-a7d3-472c-92af-16649dbf9b66.png)
![image](https://user-images.githubusercontent.com/50299857/186434563-d304e620-d776-4bea-ae27-fb700efc4106.png)

I solved it, but I think it is very inconvenient to have to add KotlinSerializationJsonHttpMessageConverter to use kotlinx.serialization.

I as with WebMvcConfigurationSupport.addDefaultHttpMessageConverters, when the Kotlinx.Serialization module does not exist, it is hoped that Jackson2 Serialization will be used by default.
![image](https://user-images.githubusercontent.com/50299857/186436699-eb8dd559-b0cc-41a6-a037-4391d58efcef.png)

I'm going to post a PR on this issue.

please check.
thank you

meloning:
Additionally, I checked the comments on RestTemplate to consider using WebClient.
```
NOTE: As of 5.0 this class is in maintenance mode, with only minor requests for changes and bugs to be accepted going forward. Please, consider using the org.springframework.web.reactive.client.WebClient which has a more modern API and supports sync, async, and streaming scenarios.
```

I checked the KotlinSerialization operation through the `BaseDefaultCodecs.initObjectReaders` method.
![image](https://user-images.githubusercontent.com/50299857/186439185-5cc1d3e9-4404-4ac5-96e5-5f4f39aa1d53.png)
[...]

PR #29009, opened 2022-08-24
Improve so that KotlinSerialization can be applied before Jackson2

meloning:
Reference Issue: [Issue-29008](https://github.com/spring-projects/spring-framework/issues/29008)

I expect to need test code, but I can't find the previously written test code. Can you give me a guide on writing test code?

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/config/AbstractMessageBrokerConfiguration.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/config/AbstractMessageBrokerConfiguration.java
@@ -465,6 +465,9 @@ public abstract class AbstractMessageBrokerConfiguration implements ApplicationC
 		if (registerDefaults) {
 			converters.add(new StringMessageConverter());
 			converters.add(new ByteArrayMessageConverter());
+			if (kotlinSerializationJsonPresent) {
+				converters.add(new KotlinSerializationJsonMessageConverter());
+			}
 			if (jackson2Present) {
 				converters.add(createJacksonConverter());
 			}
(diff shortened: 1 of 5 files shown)


sdeleuze:
That would be consistent with what we do in `WebMvcConfigurationSupport` [here](https://github.com/spring-projects/spring-framework/blob/d16fa82888681e5de1753387a1b499dbe35a9cfe/spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/WebMvcConfigurationSupport.java#L916-L918) which does that because Kotlin serialization require an explicit signal (the class should be annotated with `@Serializable`) otherwise serialization is delegated to Jackson.

@rstoyanchev Are you ok with doing that on `5.3.x` even if we are late in the patch release cycle? It will be a slight change of behavior but should impact very few people and I think it could be considered as a bug and a lack of consistency.

meloning:
Also, parts [AbstractMessageBrokerConfiguration](https://github.com/spring-projects/spring-framework/blob/d16fa82888681e5de1753387a1b499dbe35a9cfe/spring-messaging/src/main/java/org/springframework/messaging/simp/config/AbstractMessageBrokerConfiguration.java#L477) and [AllEncompassingFormHttpMessageConverter](https://github.com/spring-projects/spring-framework/blob/d16fa82888681e5de1753387a1b499dbe35a9cfe/spring-web/src/main/java/org/springframework/http/converter/support/AllEncompassingFormHttpMessageConverter.java#L99) are in the same situation, can I fix it in the PR and upload it?

meloning:
Hello @sdeleuze @rstoyanchev 
I added kotlinSerialization test case in RestTemplateTests and fixed the case where MessageBrokerConfigurationTests that test AbstractMessageBrokerConfiguration fails.

Please check and review.

meloning:
Hi @rstoyanchev @sdeleuze 
There is no answer yet, so I'm leaving a comment again.
Please let me know if there are any errors or if there are any additional considerations.

Please check.
Thank you

sdeleuze:
Merged a refined version via bb3ada4a673c689b9d5c854ec021b8e96e32c447, thanks!

Issue #28931, opened 2022-08-05
Cannot build another project using --include-build on Spring Framework sources


snicoll:
I am trying to build Spring Boot `main` against a local change in Spring Framework `main`. This leads to the following exception:

```
 $ ./gradlew build --include-build ../../spring-framework/main

> Task :buildSrc:compileJava
Note: /Users/snicoll/workspace/work/spring-boot/main/buildSrc/src/main/java/org/springframework/boot/build/context/properties/ConfigurationPropertiesPlugin.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':spring-boot-project:spring-boot:checkTestCompileClasspathForProhibitedDependencies'.
> Could not resolve all task dependencies for configuration ':spring-boot-project:spring-boot:testCompileClasspath'.
   > Could not find com.thoughtworks.qdox:qdox:.
     Required by:
         project :spring-boot-project:spring-boot > project :main:spring-core-test

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

BUILD FAILED in 1m 27s
48 actionable tasks: 40 executed, 3 from cache, 5 up-to-date
```

wilkinsona:
If the team are interested, I would be happy to contribute the necessary changes to move from the dependency management plugin to Gradle's built-in platform support.

snicoll:
Yes please, thanks for the offer!

bclozel:
Closing in favor of #29045

izeye:
It still doesn't seem to work with the same failure in the opening comment.

snicoll:
@izeye thanks for the feedback. I am unable to reproduce the problem. Can you share the command you're using?

wilkinsona:
I thought I had tested this and it worked, but I can reproduce it:

```
./gradlew build --include-build ../../spring-framework     

> Task :buildSrc:compileJava
Note: /Users/awilkinson/dev/spring-projects/spring-boot/main/buildSrc/src/main/java/org/springframework/boot/build/context/properties/ConfigurationPropertiesPlugin.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

FAILURE: Build failed with an exception.

* What went wrong:
[...]

PR #29045, opened 2022-08-30
Replace dependency management plugin with built-in platform support

wilkinsona:
This PR should address https://github.com/spring-projects/spring-framework/issues/28931. Some exclusion declarations that were having no discernible effect have been intentionally removed as part of the migration. `./gradlew build` succeeds locally.


bclozel:
Thanks very much Andy!

PR #29057, opened 2022-09-02
Update CacheAspectSupport.java

ms100:
There is no need to save the rejected results to a list.


snicoll:
@MS100 thanks for the PR but returning true in the catch block makes me feel a little bit uneasy. I'd rather stick with processing them all.

PR #29065, opened 2022-09-02
Fix Transaction already completed issue

Vladislav-Zolotaryov:
I have encountered an issue when using **Reactive Transactions** with **PostgreSQL** in **Serialized Isolation** mode, that when commit fails with **ConcurrencyFailureException**, **TransactionalOperator** actually tries to call rollback on the transaction.

The issue is that in **ReactiveTransactionManager#rollback** method, it is stated in case if commit threw an exception - you should not call the rollback method.

Currently when **ReactiveTransactionManager#commit** fails, the **TransactionalOperator** actually calls the **rollbackOnException** and it causes the rollback method to throw **IllegalTransactionStateException** with message about the transaction already being complete.

My proposal to fix this, is to wrap and exceptions that rise from commit phase in **Flux/Mono.whenUsing** into a special **TransactionCommitException** that allows us the decide weather to rollback or no, and also propagate it upstream so that client could decide to retry on such an occasion.

### DIFF
--- a/spring-r2dbc/src/test/java/org/springframework/r2dbc/connection/R2dbcTransactionManagerUnitTests.java
+++ b/spring-r2dbc/src/test/java/org/springframework/r2dbc/connection/R2dbcTransactionManagerUnitTests.java
@@ -27,12 +27,12 @@ import io.r2dbc.spi.Statement;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
-import org.springframework.transaction.TransactionCommitException;
 import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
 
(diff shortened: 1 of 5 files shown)


snicoll:
@mp911de can you please look at this one when you get a chance? Thanks!

Vladislav-Zolotaryov:
It seems that since I reported it almost a year ago, @sbrannen fixed this issue few months ago, by unwrapping the exception produced by `Flux#usingWhen` in `TransactionalOperatorImpl#unwrapIfResourceCleanupFailure`

So even in case when commit fails with something like `ConcurrencyFailureException`, we should be able to properly write retry logic

Closing the issue, thank you @sbrannen

sbrannen:
Thanks for the feedback, @Vladislav-Zolotaryov.

- Superseded by #27572

PR #29072, opened 2022-09-04
Add `application/zip` mime type

donno2048:


bclozel:
Thanks for the contribution, but these days we try to only add enum values there for technologies that are actively supported by Spring, meaning the enum value would have a legitimate purpose in our own codebase. This is not the case as far as I understand, so I'm declining this PR.

Thanks!

PR #29083, opened 2022-09-05
Introduce registerResource(Resource) in ResourceHints

sbrannen:
This PR simplifies the registration of hints for `classpath:` resources.

### DIFF
--- a/spring-core/src/main/java/org/springframework/aot/hint/support/RuntimeHintsUtils.java
+++ b/spring-core/src/main/java/org/springframework/aot/hint/support/RuntimeHintsUtils.java
@@ -16,14 +16,19 @@
 
 package org.springframework.aot.hint.support;
 
+import java.util.Arrays;
 import java.util.function.Consumer;
 
 import org.springframework.aot.hint.MemberCategory;
+import org.springframework.aot.hint.ResourceHints;
 import org.springframework.aot.hint.RuntimeHints;
 import org.springframework.aot.hint.TypeHint;
 import org.springframework.aot.hint.TypeHint.Builder;
 import org.springframework.core.annotation.AliasFor;
 import org.springframework.core.annotation.MergedAnnotation;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.Resource;
+import org.springframework.util.Assert;
 
 /**
  * Utility methods for runtime hints support code.
@@ -90,6 +95,27 @@ public abstract class RuntimeHintsUtils {
 		if (annotation.isSynthesizable()) {
 			registerSynthesizedAnnotation(hints, annotation.getType());
 		}
+	}
+
+	/**
+	 * Register that the supplied resource should be made available at runtime.
+	 * <p>If the supplied resource is not a {@link ClassPathResource}, it will
+	 * not be registered.
+	 * @param hints the {@link RuntimeHints} instance to use
+	 * @param resource the resource to register
+	 * @throws IllegalArgumentException if the supplied resource does not
+	 * {@linkplain Resource#exists() exist}
+	 * @see ResourceHints#registerPattern(String)
+	 */
+	public static void registerResource(RuntimeHints hints, Resource resource) {
+		if (resource instanceof ClassPathResource classPathResource) {
+			Assert.isTrue(resource.exists(), () -> "Resource does not exist: " + resource);
+			hints.resources().registerPattern(classPathResource.getPath());
+		}
+	}
+
+	public static void registerResources(RuntimeHints hints, Resource... resources) {
+		Arrays.stream(resources).forEach(resource -> registerResource(hints, resource));
 	}
 
 }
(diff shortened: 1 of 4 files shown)


spring-core/src/main/java/org/springframework/aot/hint/support/RuntimeHintsUtils.java
```diff
@@ -92,4 +97,25 @@ public static void registerAnnotationIfNecessary(RuntimeHints hints, MergedAnnot
...
+	 * not be registered.
+	 * @param hints the {@link RuntimeHints} instance to use
+	 * @param resource the resource to register
+	 * @throws IllegalArgumentException if the supplied resource does not
+	 * {@linkplain Resource#exists() exist}
+	 * @see ResourceHints#registerPattern(String)
+	 */
+	public static void registerResource(RuntimeHints hints, Resource resource) {
+		if (resource instanceof ClassPathResource classPathResource) {
+			Assert.isTrue(resource.exists(), () -> "Resource does not exist: " + resource);
```
snicoll, line 112: Couldn't this be `registerResourceIfNecessary` and then not register the resource if it's not found. That would be consistent with other methods in the same category.

spring-core/src/main/java/org/springframework/aot/hint/support/RuntimeHintsUtils.java
```diff
@@ -92,4 +97,25 @@ public static void registerAnnotationIfNecessary(RuntimeHints hints, MergedAnnot
...
+			hints.resources().registerPattern(classPathResource.getPath());
+		}
+	}
+
+	public static void registerResources(RuntimeHints hints, Resource... resources) {
```
snicoll, line 117: We didn't have a `registerAnnotations` either. I am not strongly opposed of adding this, but rather than dong this, you could return a `Consumer<ResourcePatternHints.Builder>` and use the existing `registerPattern`.

spring-test/src/main/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListener.java
```diff
@@ -390,19 +392,10 @@ private Stream<Method> getSqlMethods(Class<?> testClass) {
...
-	private String cleanClasspathResource(String location) {
-		location = location.substring(CLASSPATH_URL_PREFIX.length());
-		if (!location.startsWith(SLASH)) {
-			location = SLASH + location;
-		}
-		return location;
+	private void registerClasspathResources(RuntimeHints runtimeHints, ClassLoader classLoader, String... locations) {
+		DefaultResourceLoader resourceLoader = new DefaultResourceLoader(classLoader);
+		RuntimeHintsUtils.registerResources(runtimeHints,
+			TestContextResourceUtils.convertToResources(resourceLoader, locations));
```
snicoll, line 398: For the above to be useful in a more broader context, something that lets you convert locations to `Resources` outside of `spring-test` would be needed.

snicoll (commented):
thanks for the proposal. I've added a few comments.

spring-core/src/main/java/org/springframework/aot/hint/support/RuntimeHintsUtils.java
sbrannen: Yes, that's a good idea. 👍

spring-core/src/main/java/org/springframework/aot/hint/support/RuntimeHintsUtils.java
sbrannen: > you could return a Consumer<ResourcePatternHints.Builder> and use the existing registerPattern.

Which code is the "you" in this context?

In any case, I may omit the `registerResources()` method initially. It's a nice "convenience method" to have but not essential.

spring-test/src/main/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListener.java
sbrannen: If we leave out `registerResources()` (plural), this becomes a moot point.

I'm not sure if such a utility would be used very often outside `spring-test`. Nobody has requested it so far, so maybe it's better to leave it as-is for the time being.

spring-core/src/main/java/org/springframework/aot/hint/support/RuntimeHintsUtils.java
```diff
@@ -92,4 +97,25 @@ public static void registerAnnotationIfNecessary(RuntimeHints hints, MergedAnnot
 		}
 	}
 
+	/**
+	 * Register that the supplied resource should be made available at runtime.
+	 * <p>If the supplied resource is not a {@link ClassPathResource}, it will
```
snicoll, line 102: "Or does not exist"?

sbrannen: The current Javadoc reads like this:

> Determine if the supplied resource is a `ClassPathResource` that _exists_ and register the resource for run-time availability accordingly.

sbrannen:
Reopening to switch back to the original method name and semantics.

PR #29085, opened 2022-09-06
Add support for providing JNI runtime hints

### LINKED ISSUE #29007: Support jni-config.json generation
While not needed by framework itself it'd be beneficial to create `jni` config with aot hints when application needs it so that you don't need to hand craft these config files.

jvalkeal:
This commit adds support generating graalvm `jni-config.json` file.

Configuration format for jni/reflection in graalvm is documented
to be exactly same so we can re-use facilities for reflection hints
which should be relatively clean for a user as also graalvm uses same
classes for both jni/reflection.

Closes gh-29007


sdeleuze:
Merged via 135f9070c5de85a3fbe5aee68ba1c48d413a23cc, thanks.

Issue #28868, opened 2022-07-25
Deprecate support for themes


vpavic:
These days, theme switching is purely CSS based and does not really benefit from some special support on the server side. See [this guide](https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/) as an example.

With that in mind, Framework's [support for themes](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc-themeresolver) is quite obsolete and 6.0 might be a good opportunity to deprecate it.

bclozel:
Closing in favor of #29114

PR #29114, opened 2022-09-09
Deprecate support for theme

Nheyll:
This PR resolves the following issue #28868
I tried to apply the practices used in prior deprecation (ex : PropertyPlaceholderConfigurer, FailureCallback) which are : 

- Deprecate all classes fully related to theming
- Update javadoc with `@deprecated as of 6.0, with no direct replacement`
- Remove imports in other classes
- Deprecate methods and classes related to theming in other classes
- Update javadoc in other classes with `Theme support is deprecated as of 6.0`
- Suppress the deprecation warnings in test classes


bclozel:
Thanks @Nheyll , this will be in the next milestone!

PR #29132, opened 2022-09-11
 Introduce TestSocketUtils as a replacement for SocketUtils

### LINKED ISSUE #28210: Introduce TestSocketUtils as a replacement for SocketUtils
`SocketUtils` was officially deprecated in 5.3.16 (#28052) and removed in 6.0 M3 (#28054); however, there is still need for a subset of this functionality in integration tests for testing scenarios in which it is not possible for the system under test to select its own random port (or rely on the operating system to provide an ephemeral port).

We will therefore introduce a scaled down version in `spring-test` called `TestSocketUtils`, supporting retrieval of a single TCP port, and potentially with a simple retry mechanism that allows test code using the available port to be retried a given number of times if an exception is thrown.

onobc:
This PR introduces `TestSocketUtils` as a replacement for the deprecated `SocketUtils`. It was borrowed from https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/testFixtures/java/org/springframework/core/testfixture/net/TestSocketUtils.java.

Fixes #28210

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/util/TestSocketUtils.java
+++ b/spring-test/src/main/java/org/springframework/test/util/TestSocketUtils.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.test.util;
+
+import java.net.InetAddress;
+import java.net.ServerSocket;
+import java.util.Random;
+
+import javax.net.ServerSocketFactory;
+
+/**
+ * Simple utility methods for finding available ports on {@code localhost} for
+ * use in integration testing scenarios.
+ *
+ * <p>This is a limited form of {@code SocketUtils} which is deprecated in Spring
+ * Framework 5.3 and removed in Spring Framework 6.0.
+ *
+ * <p>{@code SocketUtils} was introduced in Spring Framework 4.0, primarily to
+ * assist in writing integration tests which start an external server on an
+ * available random port. However, these utilities make no guarantee about the
+ * subsequent availability of a given port and are therefore unreliable (the reason
+ * for deprecation and removal).
+ *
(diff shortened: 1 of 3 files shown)


spring-test/src/test/java/org/springframework/test/util/TestSocketUtilsTests.java
```diff
@@ -0,0 +1,55 @@
...
+	@Test
+	void findAvailableTcpPort() {
+		int port = TestSocketUtils.findAvailableTcpPort();
+		assertThat(port >= 1024).as("port [" + port + "] >= " + 1024).isTrue();
+		assertThat(port <= 65535).as("port [" + port + "] <= " + 65535).isTrue();
+	}
+
+	@Test
+	void findAvailableTcpPortWhenNoAvailablePortFoundInMaxAttempts()  {
+		try (MockedStatic<ServerSocketFactory> mockedServerSocketFactory = Mockito.mockStatic(ServerSocketFactory.class)) {
```
onobc, line 46: I am expecting this to get balked at for a couple of reasons...
1. "controversial" inline mock maker 
2. mocking statics in general

On **1)**, I am not overly concerned w/ using the [inline mock maker](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#39) as I do not see any of the scenarios in `spring-test` that is has known issues with. Albeit, it might be considered a bit intrusive for only adding a single test case. 

On **2)**, I always have mixed feelings on mocking statics.

I figured I would throw this out here and see what the team suggests.

spring-test/src/main/java/org/springframework/test/util/TestSocketUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+		return candidatePort;
+	}
+
+	/**
+	 * Determine if the specified TCP port is currently available on {@code localhost}.
+	 */
+	private static boolean isPortAvailable(int port) {
+		try {
+			ServerSocket serverSocket = ServerSocketFactory.getDefault().createServerSocket(
```
r331, line 99: May be 
`ServerSocketFactory.getDefault()
            .createServerSocket(port, 1, InetAddress.getByName("localhost"))
            .close();`

sbrannen: That's not essential, but thanks for the proposal.

spring-test/src/test/java/org/springframework/test/util/TestSocketUtilsTests.java
sbrannen: Indeed, we don't have any plans to mock statics within Framework.

I have therefore revised the approach in a local commit that I will push as a follow up to merging your PR.

sbrannen (approved):

sbrannen:
This has been merged into `5.3.x` and `main` in ee51dab1f348361f26c477ad64ab730ad7359d6e and revised in 698f8995f72239d11052e1319c6b2802a34ba3e8.

Thanks

PR #29136, opened 2022-09-12
Add runtime hints for WebFlux's WebSocket upgrade strategies

wilkinsona:
Tested using the `webflux-netty` smoke test and a local build of these changes: https://ge.spring.io/s/nvq4vhd5i2gh4.


sdeleuze:
Thanks for this PR, but I implemented that fix differently in #29146 in order to reuse the detection logic already implemented and only register the needed reflection entry.

PR #29139, opened 2022-09-13
Fix SimpleMailMessage nullability annotations

stgerhardt:
Fixes the nullability annotations in `SimpleMailMessage` so the `set*(...)` methods have their parameters marked as `@Nullable`.

These method parameters had no annotations and when `SimpleMailMessage` is used from Kotlin with `-Xjsr305=strict` enabled, the `set*(...)` methods were treated as being marked `@NonNull`. This also meant Kotlin would not see matched pairs of get/set methods as `var` properties, so calling `message.text = "Text"` would not compile in Kotlin.

Adds tests in `SimpleMailMessageJsr305ComplianceTests.kt` that verify the nullability annotations for `SimpleMailMessage` are correct. This file will actually fail to compile if the nullability annotations are missing, causing the tests to fail.

Also modifies `spring-context-support.gradle` to force the Kotlin test code to compile with `-Xjsr305=strict`, enabling the test code to work properly.

### DIFF
--- a/spring-context-support/spring-context-support.gradle
+++ b/spring-context-support/spring-context-support.gradle
@@ -1,5 +1,7 @@
 description = "Spring Context Support"
 
+apply plugin: "kotlin"
+
 dependencies {
 	api(project(":spring-beans"))
 	api(project(":spring-context"))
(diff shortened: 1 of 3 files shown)


stgerhardt:
Apologies for the force push on my fork - forgot to set my git command line email properly and wanted to fix it.

sdeleuze:
Merged via fded79d807b7d6b69e01ecc4e241399a438ee220 without the `testFunction` method. Thanks for the Kotlin tests, but for the sake of consistency and since we intend to introduce later globally null-safety validation during the build, I did not introduce them.

PR #29155, opened 2022-09-14
Shutdown tasks annotated with @Scheduled when waitForTasksToCompleteOnShutdown is true

### LINKED ISSUE #26482: waitForTasksToCompleteOnShutdown has no effect on tasks created with @Scheduled annotation
**Affects:** 2.4.2 (and likely several previous releases)

**Steps to reproduce:**

1. create a Spring Boot App with Spring Initializr
2. `@EnableScheduling` at application class
3. create `@Scheduled` method  with `Thread.sleep(60000)`
4. set `spring.task.scheduling.shutdown.await-termination=true` in application.properties
5. start the application
6. shutdown the application as soon as it is up

**Expected Behavior**
The task should not be interrupted, since `spring.task.scheduling.shutdown.await-termination` sets `waitForTasksToCompleteOnShutdown` to `true` on the TaskScheduler. 

**Actual Behavior**
The task is interrupted the InterruptetException is logged. 

**More Details**

- When I manually schedule a task by something like `scheduler.scheduleWithFixedDelay()`, everything works as expected.
- Debugging shows, that my `@Scheduled` task is interrupted by `ScheduledAnnotationBeanPostProcessor.destroy()`
- I also noticed, that this might be intentionally. At least it was in 2016. See for instance https://github.com/spring-projects/spring-framework/issues/19633.
-  However, it is at least very surprising, that waitForTasksToCompleteOnShutdown has not the excpected effect on tasks that are created in the somehow "most prominent way", the `@Scheduled`-Annotation. So IMHO it should either be fixed or documented.
- Moreover, the behavior is inconsistent with `@Async` : if I create an `@Async` method and let it sleep for 60000ms while having `spring.task.execution.shutdown.await-termination=true` ,  everythinq works as expected: The async task is not interrupted on shutdown.

Note: When I say "shutdown" I mean to stop the application in Spring Tool Suite. In the log I see "Application shutdown requested.", so I assume I am testing an orderly shutdown here and not some corner case.

savinov:
Shutdown tasks annotated with @Scheduled when waitForTasksToCompleteOnShutdown is true

1. ScheduledTaskRegistrar: shutdown taskScheduler of ExecutorConfigurationSupport type in destroy()
2. ScheduledAnnotationBeanPostProcessor: do not cancel tasks in destroy(), taskRegistrar cancel them

Closes gh-26482

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
@@ -593,16 +593,10 @@ public class ScheduledAnnotationBeanPostProcessor
 
 	@Override
 	public void destroy() {
+		this.registrar.destroy();
 		synchronized (this.scheduledTasks) {
-			Collection<Set<ScheduledTask>> allTasks = this.scheduledTasks.values();
-			for (Set<ScheduledTask> tasks : allTasks) {
-				for (ScheduledTask task : tasks) {
-					task.cancel();
-				}
-			}
 			this.scheduledTasks.clear();
 		}
-		this.registrar.destroy();
 	}
 
 }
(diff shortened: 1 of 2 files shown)


jhoeller:
This is effectively superseded by #31019 and the wider lifecycle revision in 6.1 in general.

PR #29171, opened 2022-09-19
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/backport-bot.yml
+++ b/.github/workflows/backport-bot.yml
@@ -6,8 +6,13 @@ on:
   push:
     branches:
       - '*.x'
+permissions:
+  contents: read
 jobs:
   build:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+      issues: write # to create and close issues
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


bclozel:
Thanks, but this is already taken care of.
See https://github.com/spring-projects/spring-framework/pull/29104#issuecomment-1240355349

sashashura:
I respectfully disagree. By setting the default repository setting to Read-Only you potentially breaking the backport-bot step, because it may need the permission to create/close issues.
The repo or org level read only setting is good until you actually need write permissions.

bclozel:
Good catch!

PR #29184, opened 2022-09-21
Update snakeyaml version because it contains a vulnerability

katamotokosuke:
Hello.
I found that the snakeyaml vulnarability is in the spring-framework project.
https://nvd.nist.gov/vuln/detail/CVE-2022-25857

I updated to a non-vlunarable version.

Thank you.


bclozel:
Thanks but we're only using this dependency as an optional dependency, we're not shipping this dependency version with our POMs. As a result, compiling against this version doesn't mean Spring Framework is vulnerable in any way. We'll upgrade our optional dependencies in 5.3.x and merge forward in main.

PR #29200, opened 2022-09-25
Add support for java.util.Optional to GenericHttpMessageConverter

### LINKED ISSUE #24498: Support Optional in Jackson 2 message converters

nlmenezes:


rstoyanchev:
@poutsma the issue that this fixes #24498 seems to have been fixed recently. Can you review this PR which fixes the same to see if it is needed?

poutsma:
This seems to be a clone of #24498, including the exact same `@author` tags. Closing as well.

PR #29201, opened 2022-09-25
Fix MockMvc sample setup

inabajunmr:


snicoll:
@inabajunmr thank you for making your first contribution to Spring Framework.

PR #29203, opened 2022-09-26
Consistent use of Stream#toList()

Drezir:


spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/ControllerMethodResolver.java
```diff
@@ -373,17 +372,7 @@ public InvocableHandlerMethod getExceptionHandlerMethod(Throwable ex, HandlerMet
...
-			synchronized (this.sessionAttributesHandlerCache) {
-				result = this.sessionAttributesHandlerCache.get(handlerType);
-				if (result == null) {
-					result = new SessionAttributesHandler(handlerType);
-					this.sessionAttributesHandlerCache.put(handlerType, result);
-				}
-			}
-		}
-		return result;
+		return this.sessionAttributesHandlerCache.computeIfAbsent(handlerType, SessionAttributesHandler::new);
```
bclozel, line 375: Could you remove this change and make sure that this PR is only about switching to `toList()`?
This removes the `synchronized` check and I don't think this call covers this.

Drezir: Removed

bclozel:
Thanks @Drezir for your contribution. I've polished your changes in 9c400ed since they were breaking the build.
`Stream#toList()` returns an immutable list, unlike the original code. I've reverted some of those changes where a mutable list was expected.

Drezir:
> Thanks @Drezir for your contribution. I've polished your changes in [9c400ed](https://github.com/spring-projects/spring-framework/commit/9c400ed384ce50f17c3bc0dadb7773973b07e7a1) since they were breaking the build. `Stream#toList()` returns an immutable list, unlike the original code. I've reverted some of those changes where a mutable list was expected.

Thank you very much

PR #29204, opened 2022-09-26
update License Copyright 2002-2019 to Copyright 2002-2022

handy-git:
update License Copyright 2002-2019 to Copyright 2002-2022


snicoll:
@handy-git thanks for the proposal but the copyright is changed once we effectively update the file.

PR #29208, opened 2022-09-26
Remove unnecessary synchronized block

Drezir:


bclozel:
Sorry, I missed the fact that the Map was already concurrent friendly in #29203. Thanks!

PR #29220, opened 2022-09-29
Set reason for WebSocket `CloseStatus.SESSION_NOT_RELIABLE`

zhmaeff:
Providing a reason in `CloseStatus` gives a better understanding of the root cause on the WebSocket client side.

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/CloseStatus.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/CloseStatus.java
@@ -145,7 +145,8 @@ public final class CloseStatus implements Serializable {
 	 * client that may be done during normal shutdown.
 	 * @since 4.0.3
 	 */
-	public static final CloseStatus SESSION_NOT_RELIABLE = new CloseStatus(4500);
+	public static final CloseStatus SESSION_NOT_RELIABLE = new CloseStatus(4500)
+			.withReason("Failed to send message within the configured send limit");
 
 
 	private final int code;


zhmaeff:
Hi @rstoyanchev , could you pls review it?

rstoyanchev (commented):
Yes, providing a reason would be useful. My one concern would be to expose information to external clients that reveals the server being used. In that sense the response status isn't meant to be for debugging but more of a hint for what happened. Perhaps we could choose a a fixed reason for this case that is a bit more helpful than just the status code?

zhmaeff:
> Yes, providing a reason would be useful. My one concern would be to expose information to external clients that reveals the server being used. In that sense the response status isn't meant to be for debugging but more of a hint for what happened. Perhaps we could choose a a fixed reason for this case that is a bit more helpful than just the status code?

Hi @rstoyanchev , do you mean the same reason for both send time limit exceeded AND the buffer size limit exceeded cases?

Current impl contains the following possible reasons:
`Send time x (ms) for session 'y' exceeded the allowed limit z`
`Buffer size x bytes for session 'y' exceeds the allowed limit z`

Would reason like this be acceptable? It hides session ids, current config values but still hints possible issues.
`Send time or buffer size limit exceeded`

rstoyanchev:
Something like that, perhaps we can say `"Failed to send message within the configured send limit"`. After all, it's hard to say whether it took too long because the network is slow, or because too many messages are sent, or combination of both. We just know that the send throughput cannot be sustained.

zhmaeff:
@rstoyanchev updated according to your suggestion

PR #29221, opened 2022-09-29
Remove unnecessary synchronization blocks on maps and refactor

Drezir:


valituguran (commented):
Nice refactor.
I have one question: why are the synchronizations not needed?

snicoll:
Thanks but I'd rather keep the code as it is now, an in particular for anything that's not a straightforward object creation within the lambda bloc.

PR #29223, opened 2022-09-30
Optimization Determine Bean Name From Annotation

CodeMan-cmd:
Optimize Logic Code to Reduce Code Complexity


snicoll:
@CodeMan-cmd I am not sure what you mean by "reduce code complexity" but this check is there for a reason. Can you clarify?

CodeMan-cmd:
> @CodeMan-cmd I am not sure what you mean by "reduce code complexity" but this check is there for a reason. Can you clarify?

This judgment is invalid, personally, I think some unused codes can be cleared out.

snicoll:
Why do you think it is invalid?

CodeMan-cmd:
> Why do you think it is invalid?

if (beanName != null && !strVal.equals(beanName)) {} this beanNmae

CodeMan-cmd:
> Why do you think it is invalid?

I thought it was defensive programming at first. think about it, think it should be a bug.

snicoll:
It's checking that the suggested component name matches the current bean name, if any. It isn't a bug and is defensive in case of misconfiguration indeed.

PR #29228, opened 2022-09-30
Fix typo in WebSocket reference doc regarding subscription header

sangmin7648:
must've been confused between `message-id` header and `subscription` header


sbrannen:

Thanks

PR #29235, opened 2022-09-30
Polish observability changes

izeye:
This PR polishes observability changes a bit.


bclozel:
Thanks for having a look @izeye , this really helps the team before our RC1!

PR #29237, opened 2022-09-30
small refactoring + extract large code blocks into methods

Drezir:


bclozel:
Sorry @Drezir but we tend to not accept large PRs with lots of unrelated changes in different packages. Those are hard to review and potential side-effects are hard to track down. This also [broke the build](https://ci.spring.io/builds/204308) meaning you did not run the build locally before submitting the PR.
I'm declining this PR as a result.

PR #29238, opened 2022-09-30
small refactoring of spring-context

Drezir:


snicoll:
Thanks for the PR but we prefer to do that sort of changes ourselve. You've removed the parenthesis around  a ternary operator that we add explicitly so I am afraid the time spent on this isn't worth it.

PR #29310, opened 2022-10-12
Enable memory leak detection for tests with Netty 5 Buffer

violetagg:
Fix several memory leaks in tests.


violetagg:
@rstoyanchev @poutsma PTAL

violetagg:
@rstoyanchev PTAL
I moved the ownership change from `Netty5WebSocketSessionSupport#toMessage` to `WebSocketMessage#retain`

violetagg:
@rstoyanchev Thanks

PR #29319, opened 2022-10-13
Small polish of code in spring-core

divcon:
This PR is to polish the code a bit.


quaff:
The code originate from `asm` project not `springframework` self.

divcon:
> The code originate from `asm` project not `springframework` self.

You mean that this pr is not valid in spring framework project?

quaff:
> > The code originate from `asm` project not `springframework` self.
> 
> Does it mean that this pr is not valid in spring framework project?

I think so.

divcon:
> > > The code originate from `asm` project not `springframework` self.
> > 
> > 
> > Does it mean that this pr is not valid in spring framework project?
> 
> I think so.

Thanks for your comment.
Then I will close this issue.

PR #29321, opened 2022-10-14
Replace use of `Collections.unmodifiable*()` methods where appropriate

divcon:
There seems to be a better place to use the Collections.of() or Stream API than the Collections.unmodifiableXXX() method.
Therefore, I suggest an alternative method in the following cases:

1. If you needs information that is copied rather than the original
  In some cases, a new collection object was created within a function and the created object was returned through an UnmodifiableXXX() method.
  Removed unnecessary object creation through Collection.copyOf() method.

2. When a static block is used to create an immutable collection upon initialization
  Removed unnecessary static blocks by using collection.of() or Stream api.

3. When creating a new Collection object and putting content through an unnecessary for statement
  Remove unnecessary object creation and for statement through Stream api

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/support/AbstractCacheManager.java
+++ b/spring-context/src/main/java/org/springframework/cache/support/AbstractCacheManager.java
@@ -22,6 +22,7 @@ import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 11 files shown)


sbrannen:
This has been merged into `main` in ba136dcf40123445aa959786cdc006112d35774a and revised (and partially reverted in 95f3337bb526e639f3362d5158a15a4c6ca36958).

For details on what needed to be reverted and why, please see the latter commit.

Thanks for the contribution

PR #29323, opened 2022-10-14
Replace StringBuffer with StringBuilder in CGLIB fork

izeye:
This PR replaces `StringBuffer` with `StringBuilder` in the CGLIB fork.

I'm aware that any changes on the CGLIB fork are not accepted, but the CGLIB project itself is "unmaintained", so I'm not sure where this sort of change should go.


jhoeller:
Actually, now that we have a complete fork there just like with ASM, we can apply some general enhancements. We have patched CGLIB significantly in the meantime, and effectively have to maintain it ourselves now. So thanks for the PR, I'll merge it right away!

PR #29342, opened 2022-10-18
Use KeyValue.NONE_VALUE where possible

izeye:
This PR changes to use `KeyValue.NONE_VALUE` where possible.

See https://github.com/micrometer-metrics/micrometer/issues/3458


bclozel:
Thanks @izeye !

PR #29371, opened 2022-10-24
SharedEntityManagerCreator is oblivious to the presence of multiple TransactionManager

### LINKED ISSUE #29365: SharedEntityManagerCreator is oblivious to the presence of multiple TransactionManager
**Affects:** 5.3.22

I am trying to add a secondary DataSource to my application. 
Therefore, I have made sure I have 2 instances of the following beans:
- DataSource
- EntityManagerFactory
- PlatformTransactionManager

I created this little test:
```java
@SpringBootTest
class ReplicaTest {
  @Inject private PlatformTransactionManager masterTransactionManager;
  @Inject private EntityManager masterEntityManager;

  @Qualifier(ReplicaDataSourceBeanNames.TRANSACTION_MANAGER)
  @Inject
  private PlatformTransactionManager replicaTransactionManager;
  @Qualifier(ReplicaDataSourceBeanNames.ENTITY_MANAGER_FACTORY)
  @Inject
  private EntityManager replicaEntityManager;

  @Test
  void test1() {
    masterTransactionTemplate.execute(
            s -> {
              Foo newFoo = new Foo();
              replicaEntityManager.persist(newFoo);
              return newFoo;
            });
  }

 @Test
  void test2() {
    replicaEntityManager.persist(newFoo);
  }
}
```

What happens: `test1` does not fail while `test2` fails with `javax.persistence.TransactionRequiredException: No EntityManager with actual transaction available for current thread - cannot reliably process 'persist' call`.
What I expect: bot tests should fail since `replicaEntityManager` has been linked to `replicaTransactionManager` during the configuration phase.

`SharedEntityManagerCreator` [relies](https://github.com/spring-projects/spring-framework/blob/5e808ad0183a63944ab08017a13cbc7ed75bc581/spring-orm/src/main/java/org/springframework/orm/jpa/SharedEntityManagerCreator.java#L292) on [TransactionSynchronizationManager.isActualTransactionActive()](https://github.com/spring-projects/spring-framework/blob/5e808ad0183a63944ab08017a13cbc7ed75bc581/spring-tx/src/main/java/org/springframework/transaction/support/TransactionSynchronizationManager.java#L436) to decide wheter a transaction is active. But it does not discriminate the transaction by name.

Spring allows to qualitfy the request `TransactionManager` in multiple places:
- https://github.com/spring-projects/spring-framework/blob/5e808ad0183a63944ab08017a13cbc7ed75bc581/spring-tx/src/main/java/org/springframework/transaction/annotation/Transactional.java#L129
- https://github.com/spring-projects/spring-data-jpa/blob/88dffa8e90caa1cb39345decd6d697318210f5a8/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/config/EnableJpaRepositories.java#L138

But in the same time relies heavily on [TransactionSynchronizationManager.isActualTransactionActive()](https://github.com/spring-projects/spring-framework/blob/5e808ad0183a63944ab08017a13cbc7ed75bc581/spring-tx/src/main/java/org/springframework/transaction/support/TransactionSynchronizationManager.java#L436) without carring about the transaction name.

So looks like a bug to me.

reda-alaoui:
Fix #29365

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/EntityManagerFactoryUtils.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/EntityManagerFactoryUtils.java
@@ -63,6 +63,7 @@ import org.springframework.util.StringUtils;
  * <p>Mainly intended for internal use within the framework.
  *
  * @author Juergen Hoeller
+ * @author Réda Housni Alaoui
  * @since 2.0
  */
 public abstract class EntityManagerFactoryUtils {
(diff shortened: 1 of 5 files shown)


reda-alaoui:
:wave:

jhoeller:
Note that `SharedEntityManagerCreator` needs to be able to participate in `JpaTransactionManager` transactions as well as `JtaTransactionManager` transactions. In the latter case, it needs to lazily create an `EntityManager` and synchronize it with the ongoing transaction, assuming that the transaction manager setup and the persistence provider setup are able to participate in such an arrangement. This is the code path that executes for a multi-`JpaTransactionManager` scenario as well, it just tries to best-effort participate in the existing transaction. We cannot reliably determine whether such participation is intended and leading to correct results, so we leave that up to the application.

In terms of the effect of that `isActualTransactionActive` check, `SharedEntityManagerCreator` just asserts against obvious misuse of common write-related methods at an early point there. This is not meant to be a strict arrangement with consistent `TransactionRequiredException` throwing on our side. If we move on there and propagate the call to the target EntityManager, we effectively leave it up to the persistence provider - which might still reject that call. And if no write operation is attempted at all, we will effectively just reuse it as another transaction-scoped EntityManager instance for reading, side by side with the primary EntityManager that the original transaction operates on.

PR #29379, opened 2022-10-25
Introduce DataFieldMaxValueIncrementer for MariaDB

### LINKED ISSUE #29306: Add DataFieldMaxValueIncrementer for MariaDB
Spring Framework started treating MariaDB as an independent database here: https://github.com/spring-projects/spring-framework/pull/28355. However, it does not provide a dedicated `DataFieldMaxValueIncrementer` for it.

To give a bit of context, we have the same request in Spring Batch (to treat MariaDB as independent database, see https://github.com/spring-projects/spring-batch/issues/3891), but due to the lack of a dedicated incrementer, we would have to use the MySQL one, which feels odd given the intent in the first place. Do you agree?

fmbenhassine:
Closes gh-29306


sbrannen:

Thanks

PR #29389, opened 2022-10-27
add build folders to .gitignore

diguage:


sdeleuze:
Those are former modules now removed, could you elaborate on why they should be included in `.gitignore`?

snicoll:
@diguage remove those two folders from your local copy and you'll be fine. No need to update the ignore list.

diguage:
Sorry. I got it wrong. I thought the two folders were some compiled result.

PR #29396, opened 2022-10-28
Fix typos regarding include/exclude terminology

diguage:


sbrannen:
The technical use of the terms "include" and "exclude" in these passages is specific to include/exclude features that allow one to specify something that is included or excluded.

In other words, it is not the _pattern_ or _path_ itself that is included or excluded. For example, an "`include` pattern" is used to determine what will be "included". So the thing that will be "included" is the artifact that matches the "include pattern".

I hope that makes sense.

PR #29399, opened 2022-10-29
Always logging stack trace on bean dispose/close failed.

kazuki-ma:
Motivation
====

I noticed that warn log on bean close faield is not enough for identify root cause in some case.

The actual log I received is as follows,

```
[WARN] (DisposableBeanAdapter.java:248) Invocation of close method failed on bean with name 'webHookEventDecatonClient': java.lang.NoSuchMethodError: 'void org.apache.kafka.clients.producer.Producer.close(long, java.util.concurrent.TimeUnit)'
```

Without stack trace, I can't identify what is wrong. I think printing stack trace is always good idea when logging warn based on exception.

Also, it is not clear why the contents of the Warn Log change depending on whether the Debug Flag is enabled or not.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/DisposableBeanAdapter.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/DisposableBeanAdapter.java
@@ -200,13 +200,7 @@ class DisposableBeanAdapter implements DisposableBean, Runnable, Serializable {
 				((DisposableBean) this.bean).destroy();
 			}
 			catch (Throwable ex) {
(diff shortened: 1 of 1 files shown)


snicoll:
> Also, it is not clear why the contents of the Warn Log change depending on whether the Debug Flag is enabled or not.

That's key to why we have to decline the proposal. We do only log the stacktrace at debug level on purpose as shutting down the container may lead to resource cleanup issues that are benign at that stage of the application lifecycle. If you want to know more  details (like for any other scenarios where you need more information) increasing the log level of `org.springframework.beans.factory.support.DisposableBeanAdapter` is what you should be doing. 

PR #29412, opened 2022-11-01
Reduce allocations caused by producible media types

dreis2211:
Hi,

this PR reduces some (repeated) allocations of `LinkedHashSet`s and corresponding Nodes caused by `ProducesRequestCondition.getProducibleMediaTypes()`.

I noticed this in a fairly vanilla app where this causes somewhat around 2-3% of the allocations during runtime, so not too huge, but noticeable.

Cheers,
Christoph


sdeleuze:
Merged, thanks!

PR #29423, opened 2022-11-02
Allow TransactionAwareDataSourceProxy to eagerly fetch the connection

### LINKED ISSUE #29418: TransactionAwareDataSourceProxy.getConnection let connections leak after commit while DataSourceUtils.getConnection does not
**Affects:** 5.3.22

I have a `DataSource` of type `TransactionAwareDataSourceProxy` . My `PlatformTransactionManager` is `JpaTransactionManager`.

I have the following tests:

```java
@SpringBootTest
class DataSourceTest {

  @Inject private PlatformTransactionManager transactionManager;
  @Inject private DataSource dataSource;

  @Test
  void test1() throws SQLException {
    TransactionStatus transaction = transactionManager.getTransaction(null);
    Connection conn = DataSourceUtils.getConnection(dataSource);
    transactionManager.commit(transaction);
    conn.createStatement().execute("select 1");
  }

  @Test
  void test2() throws SQLException {
    TransactionStatus transaction = transactionManager.getTransaction(null);
    Connection conn = dataSource.getConnection();
    transactionManager.commit(transaction);
    conn.createStatement().execute("select 1");
  }
}
```

As you can see, `test1` takes the `Connection` via `DataSourceUtils.getConnection` while `test2` takes it directly on the `DataSource`.

`test1` fails on `createStatement()` because the connection is closed.
`test2` does not fail.

Note that I didn't find any documentation telling to avoid using `DataSourceUtils` on a `TransactionAwareDataSourceProxy`.

Therefore:
- IMO, this is a bug because tests should have the same output
- I don't know what should be the correct output? Is that normal that a connection "leaks" after closing a transaction like in test2 ?

reda-alaoui:
Fix #29418 

Adds an `obtainTransactionalConnectionEagerly` flag allowing to retrieve the underlying connection during `getConnection` call. I think this flag should be always true to make sure the effective connection is fetched during the transaction covering `getConnection`. But I thought you would want to avoid breaking the current behaviour, so the flag is false by default.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/TransactionAwareDataSourceProxy.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/TransactionAwareDataSourceProxy.java
@@ -67,6 +67,7 @@ import org.springframework.transaction.support.TransactionSynchronizationManager
  * properly. Use {@link Connection#unwrap} to retrieve the native JDBC Connection.
  *
  * @author Juergen Hoeller
+ * @author Réda Housni Alaoui
  * @since 1.1
  * @see javax.sql.DataSource#getConnection()
  * @see java.sql.Connection#close()
@@ -77,6 +78,7 @@ import org.springframework.transaction.support.TransactionSynchronizationManager
 public class TransactionAwareDataSourceProxy extends DelegatingDataSource {
 
 	private boolean reobtainTransactionalConnections = false;
+	private boolean obtainTransactionalConnectionEagerly = false;
 
 
 	/**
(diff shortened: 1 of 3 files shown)


reda-alaoui:
Ping?

reda-alaoui:
:wave:

jhoeller:
This would be a way to enforce consistent behavior for `TransactionAwareDataSourceProxy` next to other transaction-aware resource proxies. I'm inclined to implement this via a `lazyTransactionalConnections` flag which is `true` by default but can be turned to `false`, which seems like a better semantic fit next to `reobtainTransactionalConnections` (since asking for lazy+reobtain is semantically ok, saying lazy=false but reobtain=true is not effective but has no semantic conflict either, whereas saying eager=true + reobtain=true would be semantically odd).

reda-alaoui:
Thank you @jhoeller

Issue #29421, opened 2022-11-02
Investigate suboptimal score on TechEmpower benchmarks 


dreis2211:
Hi 👋 

I'm not a particular huge fan of the [TechEmpower benchmarks](https://www.techempower.com/benchmarks/#section=data-r21), because they barely reflect real use-cases, but (novice) programmers who maybe look for some resources that might help them with their questions of what framework to choose for their next projects might look into these benchmarks and make their decision (partly) based on these numbers.

I've compared Quarkus and Spring(-Boot) here to compare two popular frameworks that have similar goals and deliver similar functionality. There is also a comparison over at Baeldung https://www.baeldung.com/spring-boot-vs-quarkus that compares these two.

<img width="1171" alt="image" src="https://user-images.githubusercontent.com/6304496/199477054-052cb0d9-2f68-4a3b-9327-a01f21a1f14e.png">

Imho, this particular comparison isn't exactly great news for the Spring ecosystem. While there are some benchmarks on par with Quarkus, the majority is "won" by Quarkus - and by a greater margin one would expect. I've recently overheard a conversation why one should use Spring if Quarkus delivers everything as well, but with better performance. Now, we all know that there is more to this story and application A might be better suited for Quarkus while application B might be better suited for Spring - for whatever reasons. There is no silver bullet, after all. But this conversation kept me thinking. Obviously, I'm biased towards Spring and I love the ecosystem, but I'm wondering if we can do something about the performance gap (in these benchmarks).

I know for example that recent changes to the Quarkus benchmarks have been done by people involved with Quarkus. The Spring benchmarks are somewhat driven infrequently by community members (I think I have done something on these ages ago myself as well).

I'm also happy to help with this and at least update the Spring(-Boot) one to latest versions but I wonder if you have comparable machines at hand that you could run these benchmarks against and find out why these numbers are so comparably "bad". (I do have the suspicion that the numbers are actually much closer in reality). I'm also happy to help with the investigations, if you want me to 😉  

Cheers,
Christoph

P.S.: The point of this should be not to overtake Quarkus or establish a Spring vs Quarkus mentality, but to simply improve the Spring numbers. Quarkus is just a good comparison data point imho.

P.S.S: I'm reporting this here, because the majority of logic relevant to the benchmarks is inside Spring-Framework and not in Spring-Boot.

dreis2211:
Addendum 1:

[...]

PR #29428, opened 2022-11-04
Avoid String allocations in MediaType.checkParameters

dreis2211:
Hi,

while investigating things for #29421 I noticed that a tiny, but noticeable, chunk of allocations is coming from `MediaType.checkParameters`.

image

Together with the actual String allocation, this accounts for 1.5-2% of allocations in the benchmarks. So not huge, but avoidable if you ask me.

Cheers,
Christoph


sbrannen:

Thanks

PR #29430, opened 2022-11-04
EMFU#doGetTransactionalEntityManager to check for actual transaction

ettavolt:
… rather than just a synchronization. Otherwise, at least SharedEntityManagerCreator will create a 'lingering' EM inside NEVER/NOT_SUPPORTED propagation, which is closed only after exiting from inside that boundary.

I did look at some other usage of `isSynchronizationActive`, and I'm not sure if it is more proper to use `isActualTransactionActive` there. I suspect this might be true for a couple of places.

`isActualTransactionActive` and `doGetTransactionalEntityManager` seem to be [around](https://github.com/spring-projects/spring-framework/blame/a3d7dc09ef7af361705704ca9f457b4901815663/spring-orm/src/main/java/org/springframework/orm/jpa/EntityManagerFactoryUtils.java) [here](https://github.com/spring-projects/spring-framework/blame/main/spring-tx/src/main/java/org/springframework/transaction/support/TransactionSynchronizationManager.java) since before the recorded history, so we can't say if `isActualTransactionActive` was added after or before `doGetTransactionalEntityManager`.

### DIFF
--- a/spring-test/src/test/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListenerTests.java
+++ b/spring-test/src/test/java/org/springframework/test/context/jdbc/SqlScriptsTestExecutionListenerTests.java
@@ -22,7 +22,6 @@ import org.mockito.BDDMockito;
 import org.springframework.context.ApplicationContext;
 import org.springframework.core.annotation.AnnotationConfigurationException;
 import org.springframework.test.context.TestContext;
-import org.springframework.test.context.jdbc.SqlConfig.TransactionMode;
 
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
@@ -32,6 +31,7 @@ import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.mock;
 import static org.springframework.test.context.jdbc.Sql.ExecutionPhase.AFTER_TEST_CLASS;
 import static org.springframework.test.context.jdbc.Sql.ExecutionPhase.BEFORE_TEST_CLASS;
+import static org.springframework.test.context.jdbc.SqlConfig.TransactionMode.ISOLATED;
 
 /**
  * Unit tests for {@link SqlScriptsTestExecutionListener}.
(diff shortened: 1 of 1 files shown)


spring-orm/src/test/java/org/springframework/orm/jpa/JpaTransactionManagerTests.java
```diff
@@ -424,7 +425,7 @@ public void testPropagationSupportsAndRequiresNewAndEarlyAccess() {
 		assertThat(!TransactionSynchronizationManager.isSynchronizationActive()).isTrue();
 
 		Object result = tt.execute(status -> {
-			EntityManagerFactoryUtils.getTransactionalEntityManager(factory);
```
ettavolt, line 427: PROPAGATION_SUPPORTS says, that if there is no existing transaction, the code will be executed non-transactionally, and, thus it makes sense for get _Transactional_ EntityManager to return null.

snicoll:
can you please rebase and squash to a single commit, removing all the unrelated changes that you've done in test assertions? It makes it harder to understand what changes you've actually introduced in unit tests.

ettavolt:
> can you please rebase

Sure!

> and squash to a single commit, removing all the unrelated changes that you've done in test assertions? It makes it harder to understand what changes you've actually introduced in unit tests.

That's why I've split the changes in three separate commits: 1) auto-conversion (1beb7068f62d0d75683e2d701d5cbfa9c2459ee8) clean up, 2) actual changes, 3) additional test. I think the latter two could be merged together without damaging comprehensibility, but I'd prefer them being swapped, so that one can see the topic test failing before the fix.

What do you think?

snicoll:
I’d like the commit that updates the assertions to be removed please. It’s unrelated to this PR and, as such, should not be handled here.

ettavolt:
@snicoll , I've restarted the PR as #31364, so that when I update it next time it does create close-reopen noise.

snicoll:
I don't know what you mean by create-close. We would have preferred the history to be in a single place but that's too late now. In the future, please push force on your existing branch and that'll update the existing PR.

Issue #25328, opened 2020-06-29
RestTemplate reads JSON body as ISO-8859-1 instead of UTF-8 when +json suffix content-type


pbetkier:
**Affects:** 5.2.0 and later

`RestTemplate` reads JSON body as `ISO-8859-1` instead of `UTF-8` when content type has `+json` suffix, e.g. `application/vnd.api.v1+json`, and `responseType` is set to `String`. Note that `UTF-8` is chosen for `application/json` content-type.

This behaviour exists in Spring for a long time, but has now larger impact after Spring MVC applications stopped setting `;charset=utf-8` (#22788). We started observing it in multiple places of our ~500 Spring Boot microservices ecosystem after upgrades to Spring Boot 2.2.0 and later.

## Details
Let's consider a controller:

```java
@RestController
class Controller {

    @GetMapping(path = "/hello", produces = "application/vnd.api.v1+json")
    Hello hello() {
        return new Hello("zażółć gęślą jaźń");
    }

    static class Hello {
        private final String value;

        @JsonCreator
        Hello(@JsonProperty("value") String value) {
            this.value = value;
        }

        public String getValue() {
            return value;
        }
    }
}
```

Now let's call this controller in 2 different ways:
```java
public static void main(String[] args) {
    new RestTemplate().exchange(
        "http://localhost:8080/hello", HttpMethod.GET, entityWithAcceptJsonSuffix(), String.class
    ).getBody();  // 1

    new RestTemplate().exchange(
        "http://localhost:8080/hello", HttpMethod.GET, entityWithAcceptJsonSuffix(), Hello.class
    ).getBody();  // 2
}

private static HttpEntity<?> entityWithAcceptJsonSuffix() {
        HttpHeaders headers = new HttpHeaders();
        headers.set("accept", "application/vnd.api.v1+json");
        return new HttpEntity<>(headers);
}
```

### Actual behaviour
The body read in the setup above is:

1. `{"value":"zaÅ¼Ã³ÅÄ gÄÅlÄ jaÅºÅ"}`
2. `{"value":"zażółć gęślą jaźń"}`

### Expected behaviour
Both of the above should be read as `{"value":"zażółć gęślą jaźń"}`.

From what I understand `+json` suffix content types, e.g. `application/vnd.api.v1+json`, should be interpreted as `UTF-8` by default ([RFC6839](https://tools.ietf.org/html/rfc6839#section-3.1)).

poutsma:
This is now fixed.

[...]

PR #29437, opened 2022-11-06
Change StringHttpMessageConverter default encoding to UTF-8

chenyhd:
I think it's time to change the default encoding to UTF-8.

There some modules already use UTF-8 as the default encoding.

```
# Json convert
org.springframework.http.converter.json.AbstractJsonHttpMessageConverter#DEFAULT_CHARSET

# From this MediaType think UTF-8 is already the default encoding
org.springframework.http.MediaType#APPLICATION_JSON_UTF8_VALUE
org.springframework.http.MediaType#APPLICATION_JSON_UTF8
```


bclozel:
@chenyhd Isn't `ISO-8859-1` still the default encoding in the Servlet spec? Wouldn't this change break lots of applications as a result?

chenyhd:
Hi @bclozel, good point. 

I was using version 5.0.4 and faced the same issue as #25328, fixed in 5.2.x


Other things spring web:
1. Use UTF-8 as the default encoding for 'text/plain'.and declarative, you will get 'text/plain;charset=UTF-8' 
```
org.springframework.http.converter.FormHttpMessageConverter#DEFAULT_CHARSET
org.springframework.http.converter.FormHttpMessageConverter#FormHttpMessageConverter
org.springframework.http.converter.FormHttpMessageConverter#applyDefaultCharset
```

2. Use UTF-8 as the default encoding for 'application/json', but not declarative,you will get 'application/json' 
```
org.springframework.http.converter.json.MappingJackson2HttpMessageConverter
```

Should I continue? or close PR?

chenyhd:
`org.springframework.http.MediaType#APPLICATION_JSON_UTF8` marked as `@Deprecated` and explain major browser use `UTF-8` not servlet change to 'UTF-8' as default encoding

chenyhd:
I checked out more information,  `org.springframework.http.MediaType#APPLICATION_JSON_UTF8`  marked as `@Deprecated` because of [RFC 8259 Charset Encoding](https://www.rfc-editor.org/rfc/rfc8259#page-9:~:text=Issues%0A%0A8.1.-,Character%20Encoding,-JSON%20text%20exchanged), But this only for `application/json`, I didn't found RFC for `text/plain` charset encoding.

Anyway
- `application/json` from RFC
- `text/plain` from Servlet

The current code is already compatible with both cases.

bclozel:
Looking at the current situation with form data, plain text and JSON, I think we're in the right place. The Jakarta Servlet spec version 6.0 stills states the following:

> If the servlet does not specify a character encoding before the getWriter method of the ServletResponse interface is called or the response is committed, the default ISO-8859-1 is used.

I'm declining this PR as a result. Thanks for looking into it!

PR #29438, opened 2022-11-06
Avoid `instanceof GenericHttpMessageConverter` checks

sdeleuze:
As found by @dreis2211, `instanceof GenericHttpMessageConverter` checks slowdown significantly the iteration on converters in Spring MVC.

This commit introduces a default boolean `isGeneric()` method in `HttpMessageConverter` and a related override in `GenericHttpMessageConverter` in order to avoid this check.

This change increases the throughput of Techempower JSON benchmark by 6.38% and of a benchmark performed with hey (10s after 60s of warmup) by 5.5%.

@jhoeller @rstoyanchev @poutsma The gains are significant so I would would like to propose the inclusion of theses changes in Spring Framework 6. Not sure if that would be ok on `5.3.x` or not. Please let me know what you think about it.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/HttpMessageConverterExtractor.java
+++ b/spring-web/src/main/java/org/springframework/web/client/HttpMessageConverterExtractor.java
@@ -18,7 +18,9 @@ package org.springframework.web.client;
 
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -39,6 +41,7 @@ import org.springframework.util.FileCopyUtils;
  *
  * @author Arjen Poutsma
  * @author Sam Brannen
+ * @author Sebastien Deleuze
  * @since 3.0
  * @param <T> the data type
  * @see RestTemplate
@@ -51,6 +54,8 @@ public class HttpMessageConverterExtractor<T> implements ResponseExtractor<T> {
 	private final Class<T> responseClass;
 
 	private final List<HttpMessageConverter<?>> messageConverters;
+
+	private final Set<Class<?>> genericConverterClasses = new HashSet<>();
 
 	private final Log logger;
 
@@ -80,8 +85,14 @@ public class HttpMessageConverterExtractor<T> implements ResponseExtractor<T> {
 		this.responseClass = (responseType instanceof Class ? (Class<T>) responseType : null);
 		this.messageConverters = messageConverters;
 		this.logger = logger;
-	}
 
+		// Precompute costly instanceof GenericHttpMessageConverter checks
+		for (HttpMessageConverter<?> messageConverter : messageConverters) {
+			if (messageConverter instanceof GenericHttpMessageConverter<?>) {
+				this.genericConverterClasses.add(messageConverter.getClass());
+			}
+		}
+	}
 
 	@Override
 	@SuppressWarnings({"unchecked", "rawtypes", "resource"})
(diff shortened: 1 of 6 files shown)


sdeleuze:
Additional data points from my `hey` benchmarks:
 - Original throughput with default converters : 142169 req/sec
 - Throughput with default converters and `produces = MediaType.APPLICATION_JSON_VALUE` : 145944 req/sec
 - Throughput with a single converter configured : 150544 req/sec
 - Throughput with default converters and this PR: 149985 req/sec
So those figure seems to confirm that this change improves significantly the efficiency of the iteration on converters.

sdeleuze:
After discussing with @jhoeller it seems we could even consider inclusion in `5.3.x`.

dreis2211:
Without confirming the improvements yet, the type pollution agent is still producing this output, because of the checkcast.

```
--------------------------
Type Pollution Statistics:
--------------------------
1:	org.springframework.http.converter.json.MappingJackson2HttpMessageConverter
Count:	192143
Types:
	org.springframework.http.converter.GenericHttpMessageConverter
	org.springframework.http.converter.HttpMessageConverter
Traces:
	org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodProcessor.getProducibleMediaTypes(AbstractMessageConverterMethodProcessor.java:387)
		class: org.springframework.http.converter.GenericHttpMessageConverter
		count: 96087
	org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodProcessor.getProducibleMediaTypes(AbstractMessageConverterMethodProcessor.java:385)
		class: org.springframework.http.converter.HttpMessageConverter
		count: 96056
```
So from that point of view, nothing really changed and **potentially** there could be still a scaling problem caused by https://bugs.openjdk.org/browse/JDK-8180450. But given that a simple boolean check is faster than an **instanceof** and that less activity is happening on the internal cache **potentially**, I'm not surprised to see some improvements here.

sdeleuze:
Notice we need to find a way to fix the broken test `RequestResponseBodyMethodProcessorTests#resolveArgumentTypeVariableWithNonGenericConverter` :
```
class jdk.proxy3.$Proxy29 cannot be cast to class org.springframework.http.converter.GenericHttpMessageConverter (jdk.proxy3.$Proxy29 is in module jdk.proxy3 of loader 'app'; org.springframework.http.converter.GenericHttpMessageConverter is in unnamed module of loader 'app')
java.lang.ClassCastException: class jdk.proxy3.$Proxy29 cannot be cast to class org.springframework.http.converter.GenericHttpMessageConverter (jdk.proxy3.$Proxy29 is in module jdk.proxy3 of loader 'app'; org.springframework.http.converter.GenericHttpMessageConverter is in unnamed module of loader 'app')
	at org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodArgumentResolver.readWithMessageConverters(AbstractMessageConverterMethodArgumentResolver.java:176)
	at org.springframework.web.servlet.mvc.method.annotation.RequestResponseBodyMethodProcessor.readWithMessageConverters(RequestResponseBodyMethodProcessor.java:163)
	at org.springframework.web.servlet.mvc.method.annotation.RequestResponseBodyMethodProcessor.resolveArgument(RequestResponseBodyMethodProcessor.java:136)
```
Calling `ProxyFactory#getProxy(java.lang.Class<T>, org.springframework.aop.TargetSource)` on an interface with a default method seems not working at this point.

sdeleuze:
Test fixed via [this change](https://github.com/spring-projects/spring-framework/pull/29438/commits/f43892693fbd17d459b693282a459a1363a793e8#diff-6f035c3adc1e6f86d471a8ea716fe861cfc7f02adc32e912609a5a65ff3f2eb1R281) which looks ok to me based on what was discussed in #29438.

sdeleuze:
> Without confirming the improvements yet, the type pollution agent is still producing this output, because of the checkcast.

@dreis2211 Notice I initially tried a more elegant/effective fix that would use casting only in the constructor by adding a new `List<GenericHttpMessageConverter<?>> genericMessageConverters` field in addition to the existing `List<HttpMessageConverter<?>> messageConverters` one, but I stopped because the order of the converters matters so we can't iterate first on the `GenericHttpMessageConverter` and after on the non generic `HttpMessageConverter`.

Given the good figures of the changes of this PR, I think my proposal is in practice effective enough, if there is a reasonable way to avoid per request casting, feel free to share.

sdeleuze:
Based on @poutsma feedback, I am testing various alternative implementations without impact on the public APIs.

sdeleuze:
After increasing significantly (10x) the number of runs of my benchmarks with both Techempower and `hey`, I am not able to reproduce a significant difference between `main` and the `isGeneric` property variant anymore, so I will just close this PR unmerged.

PR #29439, opened 2022-11-06
Use ObservationThreadLocalAccessor.KEY where possible

izeye:
This PR changes to use `ObservationThreadLocalAccessor.KEY` where possible.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/filter/reactive/ServerHttpObservationFilter.java
+++ b/spring-web/src/main/java/org/springframework/web/filter/reactive/ServerHttpObservationFilter.java
@@ -21,6 +21,7 @@ import java.util.Set;
 
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
+import io.micrometer.observation.contextpropagation.ObservationThreadLocalAccessor;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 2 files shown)


bclozel:
@marcingrzejszczak Can we safely use `io.micrometer.observation.contextpropagation.ObservationThreadLocalAccessor` from "micrometer-observation"? Your [comment in this class](https://github.com/spring-projects/spring-framework/blob/19cf5035343922111e7c03b60933bf178170f9d4/spring-web/src/main/java/org/springframework/web/filter/reactive/ServerHttpObservationFilter.java#L62-L65) seems to avoid a dependency on purpose.

Or is the key supposed to be maintained in "[context-propagation](https://github.com/micrometer-metrics/context-propagation)"?

marcingrzejszczak:
Context Propagation is typically an optional dependency. If we make it a compile time dependency then we can use the `ObservationThreadLocalAccessor` safely. Otherwise we might reference a class that implements an interface that is not on the classpath

izeye:
I'm not sure if I fully understand the current arrangement as the previous code comment was saying `micrometer-core`, and @marcingrzejszczak seems to be saying `context-propagation`, but actually the constant (`ObservationThreadLocalAccessor.KEY`) is in `micrometer-observation`: https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-observation/src/main/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessor.java#L37

marcingrzejszczak:
I made a mistake with the code comment :facepalm:

PR #29443, opened 2022-11-07
Avoid unnecessary allocations in StompDecoder#unescape

dreis2211:
Hi,

I noticed in one of our applications that a tiny amount (0,5-1%) of allocations is caused by `StompDecoder#unescape`. The unfortunate bit is that there is actually no unescaping needed in most of the cases. This PR avoids unnecessary allocations if the String to unescape is not containing any `\` character.

Cheers,
Christoph


sbrannen:

Thanks 👍

Issue #29411, opened 2022-11-01
GET query parameters containing comma are not properly URL decoded


ionel-sirbu-crunch:
**Affects:** Spring Boot 2.7.5

---

Hello,

I recently realised that one of my `GET` endpoints has issues if the values passed in the query params contain commas (`,`).
Basically, I expect to be able to embed a comma into a single value being passed into a query param, however, the receiving end must define the data type as `List<String>` & Spring splits the value into 2 rather than keeping it in one piece.

To give an example, assume you have this endpoint:

```java
@GetMapping
public ResponseEntity<List<String>> getSomeResource(@RequestParam("res") List<String> resources) {
    return ResponseEntity.ok(resources);
}
```

If you hit it with this request:

```bash
curl http://localhost:8080/test?res=xx%2Cyy
```

then you get back `["xx","yy"]`, which is exactly the same result as requesting `curl http://localhost:8080/test?res=xx,yy`, but I expect to get back a list with a single element instead, `["xx,yy"]`.

Demo app can be found [here](https://github.com/ionel-sirbu-crunch/spring-decode-comma).

Thank you very much!

ionel-sirbu-crunch:
Probably in the same functional area, spaces (` `) embedded in query params don't get decoded properly either, they are left as pluses (`+`). I've added a test to reflect that too in the demo app.

dmytro-zhelieznyi:
Hello, It simply can be fixed by small changes for this code snipe. 
org.springframework.core.convert.support.StringToCollectionConverter class
But the question is will it affect the existing code if code base will be moved to this spring version or higher?
Related changes https://github.com/spring-projects/spring-framework/pull/29451

ionel-sirbu-crunch:
The way I see it, at a very high level, the parameter resolution needs to be processed in 2 rather distinct initial stages:

1. extraction of raw query parameters from the query string - this is where we account for commas (`?q=what,where`) & repeated params (`?q=what&q=where`);
2. URL decoding of the extracted values.

That way commas, equal signs & ampersands encoded into the values (stage 2) are not mixed with the ones that are actually part of the URL syntax (stage 1).

Obviously, further processing will take place after these initial steps, e.g. data conversion.

sbrannen:
Potentially related to:

- #23820

rstoyanchev:
Thanks for logging the issue indeed, and indeed that is how the framework gets better. 

[...]

PR #29451, opened 2022-11-08
Added comma checker for string to collection converter

dmytro-zhelieznyi:
Related to https://github.com/spring-projects/spring-framework/issues/29411

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/StringToCollectionConverter.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/StringToCollectionConverter.java
@@ -65,7 +65,8 @@ final class StringToCollectionConverter implements ConditionalGenericConverter {
 		}
 		String string = (String) source;
 
-		String[] fields = StringUtils.commaDelimitedListToStringArray(string);
+		String[] fields = {string};
+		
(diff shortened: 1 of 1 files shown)


mdeinum:
This is a breaking change as when a comma is detected it will now keep it as is with 1 comma and still parse it when there are more then 1.

dmytro-zhelieznyi:
Hello, thank you for pointing out the error. I have made changed, so now "comma" will not be detected as separator for creating array. In this case if @RequestParam receive string like this "?res=aaa,bbb,ccc" it will be consumed as single string and assigned as 1st element of collection. For separating parameters we should separate it in the following way "?res=aaa&res=bbb&res=ccc". In this case spring will use org.springframework.core.convert.support.ArrayToCollectionConverter.

mdeinum:
Which still is a breaking change especially with such a converter which is used in other places as well (like binding variables with `@Value`) which would fail now as well.

sbrannen:
Hi @dmytro-zhelieznyi,

Thanks for submitting your first PR to the Spring Framework.

Unfortunately, as pointed out by @mdeinum, this constitutes a breaking change to the contract of `StringToCollectionConverter`.

PR #29456, opened 2022-11-09
Fix a syntax error in an XML listing in core-validation.adoc

davidcostanzo:
This is a simple fix to an XML listing in the core-validation doc (it adds a missing double-quote at the end of an attribute).
Fixing the syntax error also fixes the rendered syntax coloring.


sbrannen:

Thanks

PR #29461, opened 2022-11-10
Use unquoted value in assertion message in MediaType.checkParameters()

izeye:
Prior to 2a853aea6754341a069191999a53d2065f49cf34, unquoted values were used there, but it seems that double quotes could be duplicated now. This PR tries to restore the previous behavior.


jhoeller:
Just realized that the original commit got backported to 5.3.x as well. Will fix it there with a cherry-pick.

PR #29471, opened 2022-11-11
Removed unnecessary function call in Mono.switchIfEmpty()

divcon:
In this code, the 'switchIfEmpty()' method always calls createWebSession().
If this is unintentional, it needs to be fix.
 
``` java
return Mono.defer(() -> retrieveSession(exchange)
		.switchIfEmpty(createWebSession())
		.doOnNext(session -> exchange.getResponse().beforeCommit(() -> save(exchange, session))));
```

Please review and let me know if I misunderstood.


rstoyanchev (commented):
The `WebSessionStore#createWebSession` is itself deferred, so it's okay to call the local private method which only aims to add a DEBUG log message. We are also using the call to `createWebSession` as a trigger for checking expired sessions, so it does serve a purpose.

PR #29489, opened 2022-11-15
Replace else-if with if 

wizard4wu:


snicoll:
@Encyclopedias there are now three pull requests for the same thing. Please refrain from creating more as each sends a notification to a lot of watchers. If you need to update your PR, you can push additional commits on your existing branch.

Thanks for the PR but the code as it is now reads fine to me and removing the `else` does not add value IMO.

Encyclopedias:
@snicoll sorry, I will not do this
next time. In this PR, there is a merge commit, the commit is produce after i update the code from spring main branch to my fork spring main branch. Does it impact the next pr?

PR #29499, opened 2022-11-16
Polish `AbstractAutowireCapableBeanFactory` and apply `instanceof` pattern matching

wizard4wu:
1. `hasInstantiationAwareBeanPostProcessors()`: IMO, It is not necessary to call twice.
   1. Before initialization BeanPostProcessorCache, if the cache is not null, the first and second method calls always return directly.
   2. Before initialization BeanPostProcessorCache, if the cache is null, the first method call will initialize the cache, and the second method call will still return directly.
2. `needsDepCheck`
    we need to get the result when we need to use. There is a return result in the for cycle. So it is possible that we get the result but we do not use it;
3. use `instanceof` pattern matching


spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
```diff
@@ -598,12 +598,12 @@ protected Object doCreateBean(String beanName, RootBeanDefinition mbd, @Nullable
 			populateBean(beanName, mbd, instanceWrapper);
 			exposedObject = initializeBean(beanName, exposedObject, mbd);
 		}
-		catch (Throwable ex) {
-			if (ex instanceof BeanCreationException && beanName.equals(((BeanCreationException) ex).getBeanName())) {
-				throw (BeanCreationException) ex;
+		catch (Throwable th) {
```
snicoll, line 601: Please do not make stylistic change like that. We name our exception `ex` regardless of the type.

snicoll (changes requested):
Thanks for the PR.

spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
sbrannen: Following up on what @snicoll said, we'd likely rewrite that similar to the following:

```java
if (ex instanceof BeanCreationException bce && beanName.equals(bce.getBeanName())) {
	throw bce;
}
```

Encyclopedias: @snicoll @sbrannen  done

sbrannen:
Thanks for the contribution!

This has been merged into `main` in e9555d482554c3f4c7b33fbf1f38591002a94857 and revised in cefd74555ada1285ecd8dc174f6dfc68115807f8.

Regarding double invocation of `hasInstantiationAwareBeanPostProcessors()`, I decided to leave that in place since the state of the bean factory can change during the act of resolving autowiring candidates.

Encyclopedias:
@sbrannen 
You think the bean factory can change during the act of resolving autowiring candidates. And i see the add the data to 
 instantiationAware list only in the getBeanPostProcessorCache() method. When the beanPostProcessorCache is null, the code can init the BeanPostProcessorCache. So the data will add to instantiationAware list. And the instantiationAware list is only changed in the getBeanPostProcessorCache() method.  When the method(hasInstantiationAwareBeanPostProcessors) is called twice, the beanPostProcessorCache must not null, because the cache has already initialized, it does not add the data to instantiationAware list. So the list does not change during the process.
 
 If there is any misunderstanding, correct me please. I hope to receive you reply. Thanks.

PR #29503, opened 2022-11-16
Polish code a bit

divcon:
1. Remove final keyword in front of method name in final class.
- ReadSubscription class is declared as final class. So I think final keyword is not needed with methods in ReadSubscription class.


snicoll:
@divcon I am afraid I am not keen to process this PR. It mixes style change (such as removing `final` keywords) with Java 17 API usage. `k` and `v` are not really variable names we'd use. In short, it'd take us more time to polish this than doing this ourselves. Thanks anyway.

divcon:
@snicoll Thanks for your comment.

PR #29506, opened 2022-11-17
Update `LogAdapter` to allow build-time code removal

sdeleuze:
Allow for example to remove those classes and 90 related methods when Logback is used:
- `org.apache.commons.logging.LogAdapter$JavaUtilAdapter`
-  `org.apache.commons.logging.LogAdapter$JavaUtilLog`
- `org.apache.commons.logging.LogAdapter$LocationResolvingLogRecord`
- `org.apache.commons.logging.LogAdapter$Log4jAdapter`
-  `org.apache.commons.logging.LogAdapter$Log4jLog`
- `org.apache.commons.logging.LogAdapter$LogApi`
- `org.apache.logging.log4j.message.ObjectMessage`
- `org.apache.logging.log4j.message.ReusableObjectMessage org.apache.logging.log4j.simple.SimpleLoggerContext`
- `org.apache.logging.log4j.simple.SimpleLoggerContextFactory`

`org.apache.logging.slf4j.SLF4JProvider` extends `org.apache.logging.log4j.spi.Provider`, so a bunch of Log4j classes are still reachable, but looks like a useful improvement.

### DIFF
--- a/spring-core/src/main/java/org/springframework/aot/nativex/feature/PreComputeFieldFeature.java
+++ b/spring-core/src/main/java/org/springframework/aot/nativex/feature/PreComputeFieldFeature.java
@@ -36,7 +36,8 @@ class PreComputeFieldFeature implements Feature {
 			Pattern.compile(Pattern.quote("org.springframework.core.NativeDetector#imageCode")),
 			Pattern.compile(Pattern.quote("org.springframework.") + ".*#.*Present"),
 			Pattern.compile(Pattern.quote("org.springframework.") + ".*#.*PRESENT"),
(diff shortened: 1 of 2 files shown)


jhoeller (approved):
Looks good to me, also the use of `Function` there instead of the enum-based delegation arrangement.

wilkinsona:
I think it's `NoClassDefFoundError` that needs to be caught in addition to the existing catch of `ClassNotFoundException`.

sdeleuze:
I think this is now fixed, sorry for the regression, this one was pretty hard to anticipate. I think previously this potential bug was possible but hidden by the fact the classpath checks were done lazily in the static block. Thanks for catching this with Boot tests.

Issue #29513, opened 2022-11-18
Link to Spring WebFlux section is broken


kenduraghav:
**Affects:** \<Spring Framework version>
5.2.5.RELEASE to 5.2.22.RELEASE, 5.3.24, Current(6.0.0)
---

 [Under Other Web Frameworks](https://docs.spring.io/spring-framework/docs/5.2.5.RELEASE/spring-framework-reference/web.html#web-integration) Spring Webflux link is mapped wrongly which navigates to 404 error page

![image](https://user-images.githubusercontent.com/45258718/202598925-90f650f2-5377-4e51-ab79-86d10c684a21.png)
![image](https://user-images.githubusercontent.com/45258718/202599007-c79f847a-2257-40e0-95b9-684a341a97b9.png)

Would like to contribute and fix the issue.

Regards
Raghav

PR #29525, opened 2022-11-19
Fix link to WebFlux section in reference manual

divcon:
In [Web on Reactive Stact](https://docs.spring.io/spring-framework/docs/5.2.5.RELEASE/spring-framework-reference/web-reactive.html), the link to the **webflux** chapter in the firxt paragraph is broken.
This link will take you to 404 page.
(This is similar to #29513)

So I Fix the link in this PR.
Please review and let me know if I misunderstood.

### DIFF
--- a/framework-docs/src/docs/asciidoc/web-reactive.adoc
+++ b/framework-docs/src/docs/asciidoc/web-reactive.adoc
@@ -6,7 +6,7 @@ include::page-layout.adoc[]
 This part of the documentation covers support for reactive-stack web applications built
 on a https://www.reactive-streams.org/[Reactive Streams] API to run on non-blocking
 servers, such as Netty, Undertow, and Servlet containers. Individual chapters cover
-the <<webflux.adoc#webflux, Spring WebFlux>> framework,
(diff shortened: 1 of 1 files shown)


framework-docs/src/docs/asciidoc/web-reactive.adoc
```diff
@@ -6,7 +6,7 @@ include::page-layout.adoc[]
 This part of the documentation covers support for reactive-stack web applications built
 on a https://www.reactive-streams.org/[Reactive Streams] API to run on non-blocking
 servers, such as Netty, Undertow, and Servlet containers. Individual chapters cover
-the <<webflux.adoc#webflux, Spring WebFlux>> framework,
+the <<web/webflux.adoc#webflux, Spring WebFlux>> framework,
```
sbrannen, line 9:
```suggestion
the <<webflux, Spring WebFlux>> framework,
```

Your fix works; however, when cross referencing a section within the current file (after includes have been applied), there is no need to use the folder and file name.

I'll simplify that when merging the PR.

divcon: Thanks for your review.

PR #29531, opened 2022-11-20
Deprecate `LocalVariableTableParameterNameDiscoverer` completely (avoiding its exposure in native images)

sdeleuze:
`DefaultParameterNameDiscoverer` should be updated to not use `LocalVariableTableParameterNameDiscoverer` when running as a native image.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/DefaultParameterNameDiscoverer.java
+++ b/spring-core/src/main/java/org/springframework/core/DefaultParameterNameDiscoverer.java
@@ -16,11 +16,13 @@
 
 package org.springframework.core;
 
+import org.springframework.aot.AotDetector;
+
 /**
  * Default implementation of the {@link ParameterNameDiscoverer} strategy interface,
  * using the Java 8 standard reflection mechanism (if available), and falling back
- * to the ASM-based {@link LocalVariableTableParameterNameDiscoverer} for checking
- * debug information in the class file.
+ * to the ASM-based {@link LocalVariableTableParameterNameDiscoverer} (when not using
+ * AOT-processed optimizations) for checking debug information in the class file.
  *
  * <p>If a Kotlin reflection implementation is present,
  * {@link KotlinReflectionParameterNameDiscoverer} is added first in the list and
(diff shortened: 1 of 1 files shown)


snicoll:
I think this should behave the same way on the JVM. the check should be on AOT not native.

sdeleuze:
Could be great for consistency indeed even if technically on JVM + AOT we have to the capability to use it. Let's validate the choice in today's Framework meeting.

sdeleuze:
I updated the PR accordingly.

jhoeller:
An interesting case in terms of automatic exclusion indeed.

We certainly don't want to support that parameter name discovery strategy in a native image (where the underlying class files are generally not available), and it's not recommended in any other scenario either... since you can always compile with `-parameters` instead, with no need to parse class files then. In such a recommended setup, `StandardReflectionParameterNameDiscoverer` will always be able to resolve the parameter names first, so `LocalVariableTableParameterNameDiscoverer` will never actually be reached.

From that perspective, for a setup following Java 8+ recommendations, `LocalVariableTableParameterNameDiscoverer` could even be removed completely. It's only really there for backwards compatibility with older setups that got migrated without `-parameters`. Baking that assumption into our AOT arrangement - namely that you need to compile with `-parameters` when you are optimizing for AOT, completely avoiding unnecessary class file parsing - seems sensible.

jhoeller (approved):

jhoeller:
Reopening this one after a team discussion: It is ultimately preferable to deprecate `LocalVariableTableParameterNameDiscoverer` completely, not using it by default in any setup anymore, since that class file parsing strategy has been long superseded by the Java 8 `-parameters` flag on `javac` already.

We have considered doing this before and meant to address the native image impact of it in 6.0 but somehow missed this last week. So let's fix this glitch now, in time for the Boot 3.0 GA release, with a note in the upgrade wiki page.

jhoeller:
As a lenient measure for the transition period, we'll keep `LocalVariableTableParameterNameDiscoverer` active on the JVM but log a warning for each successful parameter name resolution. This will only be logged when parameter names actually need to be inspected, and only when `StandardReflectionParameterNameDiscoverer` did not return anything, suggesting that compilation with `-parameters` has been missed somewhere.

PR #29534, opened 2022-11-21
Fix link to Bean Utils Light Library in BeanUtils Javadoc

mdeinum:
The URL to the BULL library has changed (not sure when, probably way back). This updates it to the correct location.


sbrannen:

Thanks

PR #29557, opened 2022-11-23
Restrict forwards in MockMvcWebConnection to 100

### LINKED ISSUE #29483: Possible infinite forward loop with MockMvcWebConnection
When a filter is configured to conditionally forward, and it is configured to handle FORWARD dispatches as well, and it prevents infinite forward loops by either extending `OncePerRequestFilter` or otherwise using request attributes, this can result in infinite forward loops in `WebClient` tests using `MockMvcWebConnection`. Because in this case request attributes are not propagated from the original request to the forwards.

Example `RequireSettingFilter`:
```java
@Component
public class RequireSettingFilter extends OncePerRequestFilter {

	@Override
	protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
		if (System.getProperty("mysetting") == null) {
			request.getRequestDispatcher("/error/missingsetting").forward(request, response);
			return;
		}
		filterChain.doFilter(request, response);
	}
}
```

`RedirectTest`:
```java
@WebMvcTest
public class RedirectTest {
	@Autowired
	WebClient webClient;

	@Test
	void testForwardToError() throws FailingHttpStatusCodeException, MalformedURLException, IOException {
		HtmlPage page = webClient.getPage("/demo.html");
		assertEquals("A demo", page.asNormalizedText());
	}
}
```

Suggested fix: In `org.springframework.test.web.servlet.htmlunit.MockMvcWebConnection.getResponse(WebRequest)` forwards are handled. Limit this to e.g. 100 forwards and afterwards throw an exception that the page is not forwarding properly. Infinite loops are very bad, because they can make the build system hang.

Note that just copying request attributes from the original request to the new one wouldn't help for the case of `OncePerRequestFilter`, because it clears the attributes when exiting `doFilterInternal(...)`.

**Affects:** 5.3.23

manthanb:
When a filter is configured to conditionally forward, and it is configured to handle FORWARD dispatches as well, and it prevents infinite forward loops by either extending OncePerRequestFilter or otherwise using request attributes, this can result in infinite forward loops in WebClient tests using MockMvcWebConnection. This change will restrict the maximum number of forwards in MockMvcWebConnection to 100.

Closes gh-29483

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockMvcWebConnection.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockMvcWebConnection.java
@@ -142,6 +142,9 @@ public final class MockMvcWebConnection implements WebConnection {
 			forwardedUrl = httpServletResponse.getForwardedUrl();
 			forwards += 1;
 		}
+		if (forwards == MAX_FORWARDS) {
+			throw new IllegalStateException("Forwarded more than " + forwards + " times in a row, potential infinite forward loop");
+		}
 		storeCookies(webRequest, httpServletResponse.getCookies());
 
 		return new MockWebResponseBuilder(startTime, webRequest, httpServletResponse).build();
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockMvcWebConnection.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockMvcWebConnection.java
@@ -68,6 +68,8 @@ public final class MockMvcWebConnection implements WebConnection {
 
 	private WebClient webClient;
 
+	private static int MAX_FORWARDS = 100;
+
 
 	/**
 	 * Create a new instance that assumes the context path of the application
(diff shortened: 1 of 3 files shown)


manthanb:
Hey @sbrannen,
I am not sure if you are the right person to reach out to, and I am very sorry if you are not.
I was wondering whenever you have some time, if you can take a quick look at this request and see if it fixes the issue of potential infinite loop with MockMvcWebConnection - https://github.com/spring-projects/spring-framework/issues/29483.

simonbasle (changes requested):
@manthanb thanks for the contribution. I think that throwing an exception as proposed in the original issue would be better, wdyt? The message can mention that there is a probable infinite loop and that it forwarded 100 times.

simonbasle:
@manthanb are you still able to work on this? (cf review above)

manthanb:
Hi @simonbasle Very sorry I missed your last comment. Yes, I will make the changes and commit today.

simonbasle:
(note: to be backported in 5.3.x, as tracked by gh-29866)

simonbasle (approved):
I went ahead and made the adjustments to throw rather than ignore

PR #29587, opened 2022-11-27
Change default encoding in WebClientResponseException

ooeunz:
Occasionally, response body that are not in the ISO-8859-1 range cause encoding problems.


poutsma:
Unfortunately, ISO-8859-1 is the [default charset defined in HTTP 1.1](https://www.rfc-editor.org/rfc/rfc2616#section-3.7.1), and we cannot change that to UTF-8 one-sidedly. So I afraid we cannot merge this PR.

PR #29589, opened 2022-11-27
improve performance of getHandlerInternal

ShijunDeng:
In some scenarios, we need to get `HandlerMethod` ahead of [org.springframework.web.reactive.DispatcherHandler](https://github.com/spring-projects/spring-framework/blob/f4e23fe204588a744b111b8c7f6bbd1dbeda97b0/spring-webflux/src/main/java/org/springframework/web/reactive/DispatcherHandler.java#L142), such as an implement of authentication by [WebFilter](https://github.com/spring-projects/spring-framework/blob/f4e23fe204588a744b111b8c7f6bbd1dbeda97b0/spring-web/src/main/java/org/springframework/web/server/WebFilter.java#L29) 

```java
public class AuthFilter implements WebFilter, Ordered {

    @Autowired
    private RequestMappingHandlerMapping handlerMapping;

    @Override
    public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
        return handlerMapping.getHandler(exchange).switchIfEmpty(chain.filter(exchange))
            .flatMap(handler -> {
                    if (handler instanceof HandlerMethod) {
                        HandlerMethod methodHandle = (HandlerMethod) handler;
                        CheckPermission permission = methodHandle.getMethodAnnotation(CheckPermission.class);
                        if (Objects.isNull(permission)) {
                            permission = AnnotationUtils.findAnnotation(methodHandle.getBeanType(), CheckPermission.class);
                        }

                        if (Objects.nonNull(permission)) {
                            // TODO  do something..
                        }
                    }
                    return chain.filter(exchange);
                }
            );
    }
}
```
`HandlerMethod` is necessary but there is no alternative way to get it withon WebFilter, which limited the capabilities of WebFilter. so, we need call `getHandler`, however this may lead to repeated call of lookupHandlerMethod  as [issue](https://github.com/spring-projects/spring-framework/issues/29591):
 
https://github.com/spring-projects/spring-framework/blob/00da70e26b9cb5ece2a9499e7c1b449e856b4aea/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/AbstractHandlerMethodMapping.java#L293
 
To avoid duplicate search for `HandlerMethod` in  [`mapping.getHandler`](https://github.com/spring-projects/spring-framework/blob/f4e23fe204588a744b111b8c7f6bbd1dbeda97b0/spring-webflux/src/main/java/org/springframework/web/reactive/HandlerMapping.java#L98) , I suggest implementing a improved getHandlerInternal, in which `HandlerMethod` would be searched in attributes at first.

```java
Object handlerMethod = exchange.getAttributes().get(BEST_MATCHING_HANDLER_ATTRIBUTE);
if (handlerMethod instanceof HandlerMethod) {
	return Mono.just((HandlerMethod) handlerMethod);
}
```

related issue：https://github.com/spring-projects/spring-framework/issues/29591

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/RequestMappingInfoHandlerMapping.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/RequestMappingInfoHandlerMapping.java
@@ -98,6 +98,10 @@ public abstract class RequestMappingInfoHandlerMapping extends AbstractHandlerMe
 
 	@Override
 	public Mono<HandlerMethod> getHandlerInternal(ServerWebExchange exchange) {
+		Object handlerMethod = exchange.getAttributes().get(BEST_MATCHING_HANDLER_ATTRIBUTE);
+		if (handlerMethod instanceof HandlerMethod) {
+			return Mono.just((HandlerMethod) handlerMethod);
+		}
 		exchange.getAttributes().remove(HandlerMapping.PRODUCIBLE_MEDIA_TYPES_ATTRIBUTE);
 		return super.getHandlerInternal(exchange)
 				.doOnTerminate(() -> ProducesRequestCondition.clearMediaTypesAttribute(exchange));


rstoyanchev:
`WebFilter`s are called first and are not meant to know what the `DispatcherHandler` will do or what controller method it will choose. You can of course call a `HandlerMapping` from a `WebFilter`, but that is not something that we recommend or support.

Issue #29567, opened 2022-11-24
ResponseStatusException not working correctly


ydemartino:
**Affects:** 6.0

The javadoc for the ResponseStatusException is
```
 * Subclass of {@link ErrorResponseException} that accepts a "reason" and maps
 * it to the "detail" property of {@link org.springframework.http.ProblemDetail}.
```

But when I set a reason, the detail is always null. It seems the class is missing `getBody().setDetail(getReason());`.
Or maybe this is an issue in the javadoc?
Or am I missing something?

---

rstoyanchev:
This was the case, but [was removed](https://github.com/spring-projects/spring-framework/commit/b045e5baef0445e4feb16a5fc548bc4a7bf5555b#diff-63747db89909f3050833efe38439d4f5ba79d66baf952d1c16d6651652719d0fL80) when adding tests for all sub-classes, with the idea that the reason isn't always what should be present to the outside world. 

What I think it should say, and work, is that it uses the reason as the detail by default, but sub-classes can set the detail to something else if needed.

PR #29608, opened 2022-11-29
Set detail from reason in both constructors of ResponseStatusException

singhbaljit:
- Leftover item that was not fully resolved in a27f2e994b632a63ca1bbd6f3de5d9b60246ccaa (#29567). Basically, `setDetail()` was added to one constructor, but not the other.
- Also some minor clean up I noticed while browsing the tests for the related classes

This fixes the issue I'm seeing where the detail is not present when using `spring.webflux.problemdetails.enabled=true` in Spring Boot.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/ResponseStatusException.java
+++ b/spring-web/src/main/java/org/springframework/web/server/ResponseStatusException.java
@@ -76,8 +76,7 @@ public class ResponseStatusException extends ErrorResponseException {
 	 * @param cause a nested exception (optional)
 	 */
 	public ResponseStatusException(HttpStatusCode status, @Nullable String reason, @Nullable Throwable cause) {
-		super(status, cause);
-		this.reason = reason;
+		this(status, reason, cause, null, null);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


singhbaljit:
@rstoyanchev Where would I write the integration tests for this?

singhbaljit:
This fixes the issue I'm seeing where the `detail` is not present when using `spring.webflux.problemdetails.enabled=true` in Spring Boot.

spring-web/src/main/java/org/springframework/web/server/ResponseStatusException.java
```diff
@@ -78,6 +78,7 @@ public ResponseStatusException(int rawStatusCode, @Nullable String reason, @Null
 	public ResponseStatusException(HttpStatusCode status, @Nullable String reason, @Nullable Throwable cause) {
 		super(status, cause);
 		this.reason = reason;
+		setDetail(reason);
 	}
```
rstoyanchev, line 82: I would make this delegate to the protected constructor in the same class:
```java
public ResponseStatusException(HttpStatusCode status, @Nullable String reason, @Nullable Throwable cause) {
	this(status, reason, cause, null, null);
}
```
That becomes more clear and is already covered by tests.

rstoyanchev (changes requested):

spring-web/src/main/java/org/springframework/web/server/ResponseStatusException.java
singhbaljit: Done

singhbaljit:
That test seems flaky, its passing locally.

singhbaljit:
I cant seem to retrigger to the CI pipeline.

rstoyanchev:
This is merged now. Thanks for the pull request!

PR #2055, opened 2018-12-16
Support W3C Web Annotation Protocol Media Type

vedatakses:
**According to the document:**
- [**W3C Web Annotation Protocol**](https://www.w3.org/TR/annotation-protocol/) 
    
**The following media type can be supported by the framework:**
 
    application/ld+json; profile="http://www.w3.org/ns/anno.jsonld"

PR #29613, opened 2022-11-30
Add text csv value to MediaType

bozzelliandrea:
added csv support to the MediaType enum.


bclozel:
Thanks for the suggestion, but these days we're trying to keep the number of values low here - even if the media type is well known, we tend to add values only if Spring uses it for building a specific feature or support.

See #26087 #25491 #2055 #1589 #1054 and more

PR #29622, opened 2022-12-02
Improve Netty code sample

mdeinum:
The previous code read 

```
HttpServer.create().host(host).port(port).handle(adapter).bind().block();
```
Which doesn't block and terminates after the `block` The proper way to block is to register on the dispose callback.  See https://projectreactor.io/docs/netty/release/reference/index.html#_starting_and_stopping

```
HttpServer.create().host(host).port(port).handle(adapter).bindNow().onDispose().block();
```

### DIFF
--- a/framework-docs/src/docs/asciidoc/web/webflux.adoc
+++ b/framework-docs/src/docs/asciidoc/web/webflux.adoc
@@ -387,7 +387,7 @@ The code snippets below show using the `HttpHandler` adapters with each server A
 ----
 	HttpHandler handler = ...
 	ReactorHttpHandlerAdapter adapter = new ReactorHttpHandlerAdapter(handler);
-	HttpServer.create().host(host).port(port).handle(adapter).bind().block();
+	HttpServer.create().host(host).port(port).handle(adapter).bindNow().onDispose().block();
 ----
 [source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
 .Kotlin


rstoyanchev:
The main purpose of those snippets is to show how to plug `HttpHandler` into each server, and that includes starting the server. It doesn't aim to show anything more, so in that sense, the Reactor Netty sample as it is currently is not wrong. I can, however, change it to use `bindNow()` which is a bit shorter.

mdeinum:
It isn't the `bindNow`. The current sample doesn't block and thus nothing waits for an incoming request. It runs and stops, no wait, no possibility to fire a request and have it handled. So it does start and immediately quits, in earlier versions of Netty this wasn't the case (if remembered correctly). 

Hence the update in the code sample.

rstoyanchev:
I'm not sure I follow. We use the [same code](https://github.com/spring-projects/spring-framework/blob/63d841664c4b7b11451c25104dc0d66c220ba24f/spring-web/src/testFixtures/java/org/springframework/web/testfixture/http/server/reactive/bootstrap/ReactorHttpServer.java#L51) as what the sample shows, as far as I can see, and all our WebFlux integration tests run with that.

mdeinum:
Now that is weird. I have the same code in a `main` and it just passes through, no waiting. 

```
public class ReactorNettyBootstrap {

		var context = new AnnotationConfigApplicationContext(WebFluxConfiguration.class);
		var handler = WebHttpHandlerBuilder.applicationContext(context).build();

		var adapter = new ReactorHttpHandlerAdapter(handler);
		HttpServer.create().host("localhost").port(8090).handle(adapter).bind().block();
	}
}
```
This just runs and finishes. No waiting, no possibility to serve something. 

I do recall this working in the past, for some reason, for me, it doesn't work anymore. However if this is working for Spring it must be me, so then please close this PR (I probably need to do more digging then, but I wouldn't know where to look anymore?).

rstoyanchev:
`bind` waits just long enough for the connection to be established and returns immediately after. In your case, the `main` method does nothing else to do, and exits.

PR #29628, opened 2022-12-03
Make `getSingleton(String)` dependent on `allowCircularReferences` flag in `AbstractAutowireCapableBeanFactory`

wizard4wu:
When the allowCircularReferences is false, the singletonFactories is not be used. And the earlySingletonObjects is based on the singletonFactories, so the earlySingletonObjects is not be used. 
And the old code is always to get bean in the singletonFactories and earlySingletonObjects, even the allowCircularReferences is false.
I think we should get beans by the circular reference flag. So i provide the method in the abstract class, it is only for the bean with the allowCircularReferences flag.


Encyclopedias:
@sbrannen Can you check this？ Thanks.

wizard4wu:
Hi @sbrannen 
 
Can you help me confirm whether there is a problem?
If there is a problem, can you explain it? It's been a long time Thank you

PR #29635, opened 2022-12-04
Apply 'instanceof pattern matching' in spring-webflux

divcon:
Most variable names are written in the manner typically used in Java. (ForExample -> forExample)

### DIFF
--- a/spring-webflux/src/test/java/org/springframework/web/reactive/protobuf/Msg.java
+++ b/spring-webflux/src/test/java/org/springframework/web/reactive/protobuf/Msg.java
@@ -144,8 +144,8 @@ public  final class Msg extends
    */
   public String getFoo() {
     Object ref = foo_;
-    if (ref instanceof String) {
-      return (String) ref;
+    if (ref instanceof String stringRef) {
+      return stringRef;
     } else {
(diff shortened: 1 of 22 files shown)


spring-webflux/src/test/java/org/springframework/web/reactive/protobuf/Msg.java
```diff
@@ -144,8 +144,8 @@ public boolean hasFoo() {
    */
   public String getFoo() {
     Object ref = foo_;
-    if (ref instanceof String) {
-      return (String) ref;
+    if (ref instanceof String stringRef) {
+      return stringRef;
```
sbrannen, line 148:
```java
// Generated by the protocol buffer compiler.  DO NOT EDIT!
// source: sample.proto
```

For future reference, please do not modify generated code since the changes will be lost if the code is regenerated.

In light of that, I'll revert the changes to generated code when I merge the PR.

divcon: I'm sorry for my carelessness.

divcon:
Thanks for your review.

Issue #29630, opened 2022-12-03
Align multipart codecs on client and server


CoderYellow:
According to this [check](https://github.com/spring-projects/spring-framework/blob/b50415062b3d9243c770ac7c59875a6076681bd1/spring-web/src/main/java/org/springframework/http/codec/multipart/MultipartWriterSupport.java#L88), the response type for controller method should be `Mono<MultiValueMap>`, but when I use this:
```kotlin
    @GetMapping("multi-part-map", produces = [MULTIPART_FORM_DATA_VALUE])
    fun multipartMap(): Mono<MultiValueMap<String, HttpEntity<*>>> {
        val builder = MultipartBodyBuilder()
        builder.part("fieldPart", "fieldValue")
        builder.part("filePart1", FileSystemResource("...logo.png"))
        builder.part("jsonPart", Person("Jason"))
        builder.part(
            "myPart",
            DefaultFormFieldPart(HttpHeaders(), "myPartName", "myPartData")
        ) // Part from a server request

        val parts: MultiValueMap<String, HttpEntity<*>> = builder.build()
        return Mono.just(parts)
    }
```
I got this error:
```java
java.lang.ClassCastException: class org.springframework.util.LinkedMultiValueMap cannot be cast to class org.springframework.http.codec.multipart.Part (org.springframework.util.LinkedMultiValueMap and org.springframework.http.codec.multipart.Part are in unnamed module of loader 'app')
	at reactor.core.publisher.FluxFlatMap.trySubscribeScalarMap(FluxFlatMap.java:152) ~[reactor-core-3.5.0.jar:3.5.0]
```
Because [here](https://github.com/spring-projects/spring-framework/blob/917c41fd52f655fc4a7c4448e1164afd7e77d21a/spring-web/src/main/java/org/springframework/http/codec/multipart/PartHttpMessageWriter.java#L53) expects `Flux<Part>` response type, so I have to implement something like this:
```kotlin
class DefaultFormFieldPart(
    private val headers: HttpHeaders,
    private val name: String,
    private val value: String
) : Part,
    MultiValueMapAdapter<String, String>(emptyMap()) {

    override fun content(): Flux<DataBuffer> {
        return Flux.defer {
            val bytes = this.value.toByteArray(UTF_8)
            Flux.just(
                DefaultDataBufferFactory.sharedInstance.wrap(
                    bytes
                )
            )
        }
    }
    
    override fun name(): String {
        return name
    }

[...]

PR #29639, opened 2022-12-05
Expose part generate api

CoderYellow:
ehance issue #29630


rstoyanchev:
There is no need for a separate PR for this. A decision on this can be made either way as part of #29630.

PR #29676, opened 2022-12-12
Use consistent visibility for ResponseEntityExceptionHandler.getMessageSource()

izeye:
This PR changes to use consistent visibility for `ResponseEntityExceptionHandler.getMessageSource()`.

See 6c8fb6c


sbrannen:
Good catch!

Thanks

PR #29691, opened 2022-12-15
Provide method with `ContextView` instead of `Context` in `ServerWebExchangeContextFilter`

singhbaljit:
Use `ContextView` in `ServerWebExchangeContextFilter#get`.

 * `Context` is a subclass of `ContextView`; hence, this is a non-breaking change.
 * It makes the convenience method more convenient because `Mono#transformDeferredContextual` and `Mono#deferContextual` provide `ContentView` instance. This avoids the need for a clunky cast or the unnecessary `.map(Context::of)`.


singhbaljit:
@rstoyanchev

rstoyanchev (changes requested):
It is a reasonable and necessary change, and although it is source compatible, it is not binary compatible. We can instead introduce a new method `getExchange(ContextView)` and deprecate the existing one, which would provide a path to migrate.

Would you like to make this adjustment, or otherwise we can also take it from here?

PR #29693, opened 2022-12-15
Fix manipulating property sources example in Javadoc for `ConfigurableEnvironment`

quim3ra:
The Manipulating property sources example is stating the MutablePropertySources expect a `Map<String,String>` whereas is a `Map<String,Object>`


sbrannen:
Good catch.

Thanks

PR #29694, opened 2022-12-15
Small fixes and improvements to the reference documentation

simonbasle:
This PR fixes a few obvious typos and small issues in the reference documentation:
- Remove mention that users could be on Java < 5
- Fix typo 'can can'
- Fix subsection style in WebFlux Concurrency Model
- Fix typo: Inn -> In
- Fix links to Jakarta Mail 1.6
- Remove ref to JOTM, inactive since 2009

It makes 2 larger changes to improve cross-linking consistency between the MVC and WebFlux chapters:
1. Turn plain 'WebFlux' links to 'See equivalent in the Reactive stack' and turn plain 'WebMVC' links to 'See equivalent in the Servlet stack'
> I found the links between chapters to be a little bit too terse, and there were some inconsistencies
2. Rework linking to Sprinv MVC Async support vs WebFlux section
> These links point to a section in the same chapter, but could be confused with the "WebFlux" links from above

The PR has been split into meaningful commits for ease of review, but these can be squashed upon merging.


framework-docs/src/docs/asciidoc/integration/email.adoc
```diff
@@ -8,11 +8,12 @@ This section describes how to send email with the Spring Framework.
 The following JAR needs to be on the classpath of your application in order to use
 the Spring Framework's email library:
 
-* The https://eclipse-ee4j.github.io/mail/[JavaMail / Jakarta Mail 1.6] library
+* The https://jakartaee.github.io/mail-api/[JavaMail / Jakarta Mail 1.6] library
 
 This library is freely available on the web -- for example, in Maven Central as
 `com.sun.mail:jakarta.mail`. Please make sure to use the latest 1.6.x version
 rather than Jakarta Mail 2.0 (which comes with a different package namespace).
+See the `v1.x` code in the https://github.com/jakartaee/mail-api/tree/v1.x[Jakarta Mail-Api repository].
```
sbrannen, line 16: Unless I'm missing something, this whole section is wrong regarding the versions.

We currently compile against `jakarta.mail:jakarta.mail-api:2.0.1` for our e-mail support (pulling in dependencies on things like `jakarta.mail.Message`). So users of Spring Framework 6.0 must use Jakarta Mail 2.0+.

sbrannen (changes requested):

framework-docs/src/docs/asciidoc/integration/email.adoc
simonbasle: eh, I made this change on the basis that the 1.6.x mention was correct, primarily because the _link_ was 404. perhaps this needs to be evaluated / reworked separately then ? (with a specific backport to 5.3.x that looks like the edits I made)

simonbasle: I've pushed a commit that removes that github link. So the content referring 1.6 is still probably wrong, but at least the Jakarta Mail link is fixed.

sbrannen: > I made this change on the basis that the 1.6.x mention was correct, 

Yes, I assumed that was the case. This is all coming from the [5.3.x version of the docs](https://docs.spring.io/spring-framework/docs/5.3.x/reference/html/integration.html#mail) and was just an oversight (that it hasn't yet been updated in the 6.0.x docs).

I didn't mean to imply that you made an error with the versions. Rather, I just happened to notice it, because you modified that section. 😉 

> perhaps this needs to be evaluated / reworked separately then ? (with a specific backport to 5.3.x that looks like the edits I made)

Yes. 👍 

Would you like to handle that? If not, I can take care of it.

sbrannen (approved):

sbrannen:
This work has been merged into `main` (with partial commit squashing, reordering, etc.).

Thanks

framework-docs/src/docs/asciidoc/integration/email.adoc
sbrannen: Actually, I'll just go ahead and address that.

PR #29696, opened 2022-12-16
Improve debug logging in `RestTemplate#handleResponse` if `getStatusCode` raises `IOException`

yuezk:
After enabling the DEBUG log level, the following exception was swallowed in our production environment.

```
org.springframework.web.client.ResourceAccessException: I/O error on GET request for "https://xxx.oktapreview.com/.well-known/openid-configuration": Permission denied: connect; nested exception is java.net.SocketException: Permission denied: connect
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
@@ -908,7 +908,7 @@ public class RestTemplate extends InterceptingHttpAccessor implements RestOperat
 				logger.debug("Response " + statusCode);
 			}
 			catch (IOException ex) {
-				// ignore
+				logger.debug("Failed to get response status code", ex);
 			}
 		}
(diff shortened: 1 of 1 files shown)


sbrannen:
@rstoyanchev, this seems reasonable to me, but I wanted to run it by you first before merging since you may have had a specific reason for not logging the exception.

yuezk:
Hi guys, any update on this?

rstoyanchev:
Looks reasonable to add this change as we are already within an `isDebugEnabled` section, and makes sense to log any exception from getting the status code at the same log level. 

This seems to be very long standing behavior (10+years) where an `IOException` can come mainly from `HttpURLConnection`, while for other client libraries it's a simple getter. That's probably why it was ignored, but it should actually be logged.

yuezk:
Shall we back port it to 5.3.x? Thanks.

PR #29697, opened 2022-12-16
Extract ResourceEntityResolver HTTPS schema resolution fallback

### LINKED ISSUE #29662: ResourceEntityResolver defaults to remote access
This is definitely true on Spring 6.0.2, I suspect it's equally true for 5.3.x and likely everything going back many years...

I would not characterize this as a security issue, but it seems like a poor default decision to allow the org.springframework.beans.factory.xml.ResourceEntityResolver class to fall through to remote http(s) resolution of XML entites.

I would think at minimum that ought to be off by default, or certainly an optional behavior at least, but there's no obvious way I saw to override that behavior save for duplicating the class and injecting the replacement into our custom bean defintion reader.

There's also a nasty bug in Java's XML parser that causes it to handle null coming back from an EntityResolver by just doing its own resolution, but throwing an exception out of the resolveEntity call does prevent that. So I think it would be a good improvement to either default throw, or at least provide some kind of option to do so.

simonbasle:
This commit extracts the DTD/XSD remote lookup fallback from the
resolveEntity method into a protected method.

A warn logging statement is added to the extracted fallback in order to
make it clear that remote lookup happened.

Overriding the protected method would allow users to avoid this fallback
entirely if it isn't desirable, without the need to duplicate the local
resolution code.

Fixes gh-29662.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
@@ -118,13 +118,20 @@ public class ResourceEntityResolver extends DelegatingEntityResolver {
 	}
 
 	/**
-	 * A fallback method for {@link #resolveEntity(String, String)} used when a "schema" entity (DTD or XSD) couldn't
-	 * be resolved as a local resource. The default behavior is to perform a remote resolution over https.
-	 * Override this method if this behavior is not desirable.
-	 * @param publicId the public identifier of the external entity being referenced, or null if none was supplied.
-	 * @param systemId the system identifier of the external entity being referenced.
-	 * @return an InputSource object describing the new input source, or null to request that the
-	 * parser open a regular URI connection to the system identifier.
+	 * A fallback method for {@link #resolveEntity(String, String)} that is used when a
+	 * "schema" entity (DTD or XSD) cannot be resolved as a local resource. The default
+	 * behavior is to perform a remote resolution over HTTPS.
+	 * <p>Subclasses can override this method to change the default behavior.
+	 * <ul>
+	 * <li>Return {@code null} to fall back to the parser's
+	 * {@linkplain org.xml.sax.EntityResolver#resolveEntity(String, String) default behavior}.</li>
+	 * <li>Throw an exception to prevent remote resolution of the XSD or DTD.</li>
+	 * </ul>
+	 * @param publicId the public identifier of the external entity being referenced,
+	 * or null if none was supplied
+	 * @param systemId the system identifier of the external entity being referenced
+	 * @return an InputSource object describing the new input source, or null to request
+	 * that the parser open a regular URI connection to the system identifier.
 	 */
 	@Nullable
 	protected InputSource resolveEntityFallbackForSchemas(@Nullable String publicId, String systemId) {
--- a/spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
@@ -110,26 +110,45 @@ public class ResourceEntityResolver extends DelegatingEntityResolver {
 				}
 			}
 			else if (systemId.endsWith(DTD_SUFFIX) || systemId.endsWith(XSD_SUFFIX)) {
-				// External dtd/xsd lookup via https even for canonical http declaration
-				String url = systemId;
-				if (url.startsWith("http:")) {
-					url = "https:" + url.substring(5);
-				}
-				try {
-					source = new InputSource(ResourceUtils.toURL(url).openStream());
-					source.setPublicId(publicId);
-					source.setSystemId(systemId);
-				}
-				catch (IOException ex) {
-					if (logger.isDebugEnabled()) {
-						logger.debug("Could not resolve XML entity [" + systemId + "] through URL [" + url + "]", ex);
-					}
-					// Fall back to the parser's default behavior.
-					source = null;
-				}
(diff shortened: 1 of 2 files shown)


simonbasle:
wip because I'm wondering about the extracted method's javadoc: it should probably list returning `null`and throwing as override possibilities, but I'm not sure how to explain what each of these options does in a nutshell. would appreciate suggestions here

scantor:
> `draft` because I'm wondering about the extracted method's javadoc: it should probably list returning `null`and throwing as override possibilities, but I'm not sure how to explain what each of these options does in a nutshell. would appreciate suggestions here

Refer to the EntityResolver's resolveEntity semantics as a reference perhaps? That's ultimately what you're triggering.

spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
```diff
@@ -110,27 +110,46 @@ public InputSource resolveEntity(@Nullable String publicId, @Nullable String sys
...
-				}
+				source = resolveEntityFallbackForSchemas(publicId, systemId);
 			}
 		}
 
 		return source;
 	}
 
+	/**
+	 * A fallback method for {@link #resolveEntity(String, String)} used when a "schema" entity (DTD or XSD) couldn't
```
sbrannen, line 121:
```suggestion
	 * A fallback method for {@link #resolveEntity(String, String)} that is used when a "schema" entity (DTD or XSD) cannot
```

spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
```diff
@@ -110,27 +110,46 @@ public InputSource resolveEntity(@Nullable String publicId, @Nullable String sys
...
+	 * be resolved as a local resource. The default behavior is to perform a remote resolution over https.
+	 * Override this method if this behavior is not desirable.
```
sbrannen, line 123:
```suggestion
	 * <p>Subclasses can override this method the change the default behavior.
	 * <ul>
	 * <li>Return {@code null} to fall back to the parser's
	 * {@linkplain org.xml.sax.EntityResolver#resolveEntity(String, String) default behavior}.</li>
	 * <li>Throw an exception to prevent remote resolution of the XSD or DTD.</li>
	 * </ul>
```

spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
```diff
@@ -110,27 +110,46 @@ public InputSource resolveEntity(@Nullable String publicId, @Nullable String sys
...
+				source = resolveEntityFallbackForSchemas(publicId, systemId);
 			}
 		}
 
 		return source;
 	}
 
+	/**
+	 * A fallback method for {@link #resolveEntity(String, String)} used when a "schema" entity (DTD or XSD) couldn't
+	 * be resolved as a local resource. The default behavior is to perform a remote resolution over https.
```
sbrannen, line 122:
```suggestion
	 * be resolved as a local resource. The default behavior is to perform remote resolution over HTTPS.
```

spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
```diff
@@ -110,27 +110,46 @@ public InputSource resolveEntity(@Nullable String publicId, @Nullable String sys
...
+	 * Override this method if this behavior is not desirable.
+	 * @param publicId the public identifier of the external entity being referenced, or null if none was supplied.
+	 * @param systemId the system identifier of the external entity being referenced.
+	 * @return an InputSource object describing the new input source, or null to request that the
+	 * parser open a regular URI connection to the system identifier.
+	 */
```
sbrannen, line 128: In Spring Framework, we attempt to wrap Javadoc around 90 characters, and we never end a `@param` or `@return` description with a period `.` unless the description includes sentences (which we try avoid).

spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
```diff
@@ -110,27 +110,46 @@ public InputSource resolveEntity(@Nullable String publicId, @Nullable String sys
...
+	@Nullable
+	protected InputSource resolveEntityFallbackForSchemas(@Nullable String publicId, String systemId) {
```
sbrannen, line 130:
```suggestion
	protected InputSource resolveSchemaEntity(@Nullable String publicId, String systemId) {
```

sbrannen (changes requested):

spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
simonbasle: done, and integrated your code suggestions in the change

spring-beans/src/main/java/org/springframework/beans/factory/xml/ResourceEntityResolver.java
simonbasle: renamed the method and added tests

sbrannen: Thanks! 👍

sbrannen (approved):

sbrannen:
This has been merged into `main` in 5965917d1655819379236bba0d3098b70bc6eb80 and revised in 8df4f927fd5d584ae1c9306cb2cd47c2e5bf6d2c.

Thanks, and... congratulations on having your first PR merged into the code base! 👍

PR #29710, opened 2022-12-18
Apply "instanceof pattern matching"

diguage:


sbrannen:

In the future, please include larger change sets in your PRs.

There are certainly other places in the code base that can benefit from "instance of pattern matching", and we would prefer to address them on a per-module basis (for example, the entire `spring-aop` module).

Thanks

diguage:
OK. Thank you for the tips.

I had checked all the codes of `spring-aop`, and only this one needed to be modified.

sbrannen:
> I had checked all the codes of `spring-aop`, and only this one needed to be modified.

I'm not sure how you checked, because I found several other places that needed to be modified in `spring-aop`. See commit b71db12c43a1a09fa2fb49bf88ce8f00e017511b.

PR #29711, opened 2022-12-18
Minor refactoring for YAML usage

asomov:
Refactor YAML usage to be flexible in Spring Boot (the PR to Spring Boot will come after this PR)


asomov:
@rstoyanchev any progress? Do you expect any action from me ?

snicoll:
@asomov we can't processed with this PR without tests that excercise the change and with the current description. What you call minor refactoring introduces a new protected method which, I suppose, you want to utilize in Spring Boot but you haven't shared why.

PR #29721, opened 2022-12-20
Add ExecutingResponseCreator to delegate request and response

### LINKED ISSUE #26381: Allow MockRestServiceServer to validate requests but actually obtain response via HTTP call

simonbasle:
This commit adds a new convenience `ResponseCreator` implementation that
delegates to a `ClientHttpRequestFactory` to perform a request and
use the response.

For further convenience, a factory method based on a `RestTemplate` is
also added: `MockRestResponseCreators#byExecutingRequestUsing`. This
acts as a pseudo-DSL for fluently setting the ResponseCreator in a
`ResponseActions`.

The primary purpose of this implementation is to perform an actual
request to an external service and use the response in MockMVC.

Closes gh-26381

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/response/ExecutingResponseCreator.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/response/ExecutingResponseCreator.java
@@ -29,6 +29,8 @@ import org.springframework.util.StreamUtils;
 /**
  * A {@code ResponseCreator} which delegates to a {@link ClientHttpRequestFactory}
  * to perform the request and return the associated response.
+ * This is notably useful when testing code that calls multiple remote services, some
+ * of which need to be actually called rather than further mocked.
  * <p>Note that the input request is asserted to be a {@code MockClientHttpRequest} and
  * the URI, method, headers and body are copied.
  * <p>The factory can typically be obtained from a {@code RestTemplate} but in case this
--- a/spring-test/src/main/java/org/springframework/test/web/client/response/ExecutingResponseCreator.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/response/ExecutingResponseCreator.java
@@ -25,15 +25,22 @@ import org.springframework.mock.http.client.MockClientHttpRequest;
 import org.springframework.test.web.client.ResponseCreator;
 import org.springframework.util.Assert;
 import org.springframework.util.StreamUtils;
-import org.springframework.web.client.RestTemplate;
 
 /**
  * A {@code ResponseCreator} which delegates to a {@link ClientHttpRequestFactory}
  * to perform the request and return the associated response.
  * <p>Note that the input request is asserted to be a {@code MockClientHttpRequest} and
  * the URI, method, headers and body are copied.
- * <p>The {@link MockRestResponseCreators#byExecutingRequestUsing(RestTemplate)} method
- * can be used as a pseudo-DSL for creating an instance of this class.
+ * <p>The factory can typically be obtained from a {@code RestTemplate} but in case this
+ * is used with e.g. {@code MockRestServiceServer}, make sure to capture the factory early
+ * before binding the mock server to the RestTemplate (as it replaces the factory):
+ * <pre><code>
+ * ResponseCreator withActualResponse = new ExecutingResponseCreator(restTemplate);
+ * MockRestServiceServer server = MockRestServiceServer.bindTo(restTemplate).build();
+ * //...
+ * server.expect(requestTo("/foo")).andRespond(withSuccess());
+ * server.expect(requestTo("/bar")).andRespond(withActualResponse);
+ * </code></pre>
  *
  * @since 6.0.4
  */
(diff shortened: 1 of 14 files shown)


rstoyanchev (changes requested):
Thanks for putting this together based on the discussion under #26381. Looking at the result, I'm wondering if it wouldn't be more idiomatic to simply provide a `ResponseCreator` implementation instead:
```java
public class ExecutingResponseCreator implements ResponseCreator {

	private final ClientHttpRequestFactory requestFactory;


	public ExecutingResponseCreator(RestTemplate restTemplate) {
		this.requestFactory = restTemplate.getRequestFactory();
	}


	@Override
	public ClientHttpResponse createResponse(ClientHttpRequest request) throws IOException {
		Assert.state(request instanceof MockClientHttpRequest, "Expected mock request");
		MockClientHttpRequest mockRequest = (MockClientHttpRequest) request;
		ClientHttpRequest newRequest = this.requestFactory.createRequest(mockRequest.getURI(), mockRequest.getMethod());
		newRequest.getHeaders().putAll(mockRequest.getHeaders());
		StreamUtils.copy(mockRequest.getBodyAsBytes(), newRequest.getBody());
		return newRequest.execute();
	}
}
```

which is then used:
```java
ResponseCreator executingResponseCreator = new ExecutingResponseCreator(restTemplate);
MockRestServiceServer server = MockRestServiceServer.bindTo(restTemplate).build();
...
server.expect(requestTo("/foo")).andRespond(withSuccess());
server.expect(requestTo("/bar")).andRespond(executingResponseCreator);
```

The rest of the code, `MockRestServiceServer`, `ResponseActions`, `RequestExpectationManager` implementations would remain unaware of any of this, which simplifies the implementation, and is generally easier to understand how it fits.

It would provide extra flexibility too because it becomes just another `ResponseCreator` that's no different from the rest. I'm thinking, we could provide an option for setting a default `ResponseCreator` in the `MockRestServiceServer` builder. That's independently useful, but if set to the `ExecutingResponseCreator` it would execute requests by default when `andResponse` isn't called. That's the effect that was sought in the original comment under #26381.

simonbasle:
@rstoyanchev I have reverted the previous implementation and introduced `ExecutingResponseCreator` instead per your suggestion.

simonbasle:
I was wondering how to make that feature more discoverable, hence the additional factory method in `MockRestResponseCreators`. Perhaps some mention in documentation is needed as well?

simonbasle:
mmh I realize just now that the factory method in `MockRestResponseCreators` doesn't make much sense, because the `ExecutingResponseCreator` MUST be instantiated early on, before the `RestTemplate`'s factory is changed...

I wonder how to improve the discoverability of the feature 🤔

spring-test/src/main/java/org/springframework/test/web/client/response/ExecutingResponseCreator.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.http.client.ClientHttpRequestFactory;
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.mock.http.client.MockClientHttpRequest;
+import org.springframework.test.web.client.ResponseCreator;
+import org.springframework.util.Assert;
+import org.springframework.util.StreamUtils;
+
+/**
+ * A {@code ResponseCreator} which delegates to a {@link ClientHttpRequestFactory}
+ * to perform the request and return the associated response.
```
rstoyanchev, line 31: I think it would be good to have a second sentence that starts with "This is useful when...". As far as I can tell it's about testing code that calls multiple remote service, where some need should be called rather than mocked.

rstoyanchev (changes requested):
Thanks that looks very nice and simple.

For the visibility, we could:
-  add a paragraph in the Javadoc for `MockRestResponseCreator` that links to the alternative `ExecutingResponseCreator` along with an `@see` reference.
- add something in the reference docs under [this section](https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html#spring-mvc-test-client).
- yet another option is to add a sample test to the ones [mentioned here](https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html#spring-mvc-test-client-resources), and that could use `MockWebServer` like in `WebClientIntegrationTests` for example

spring-test/src/main/java/org/springframework/test/web/client/response/ExecutingResponseCreator.java
simonbasle: thanks for the feedback. I've addressed the documentation and discoverability suggestions.

Issue #29314, opened 2022-10-13
Creating bean with field annotated with `@Value("${test.date}")` fails


Linda-pan:
**Affects:** 5.3.21

---

## Background

TimeZone is  Asia/Singapore
snakeyml 1.31
spring-boot 2.6.9

## Code

```java
import java.util.Date;
import lombok.Data;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.context.annotation.Configuration;

@Configuration
@Data
public class TestDateTime {
    @Value("${test.date:2022-10-13 00:00:00}")
    private Date test;   

}
```

## application.yml

```yaml
     test:
         date: 2022-10-13 00:00:00
```

## Error

<details>

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'testDateTime': Unsatisfied dependency expressed through field 'test'; nested exception is org.springframework.beans.TypeMismatchException: Failed to convert value of type 'java.lang.String' to required type 'java.util.Date'; nested exception is org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [@org.springframework.beans.factory.annotation.Value java.util.Date] for value 'Thu Oct 13 08:00:00 SGT 2022'; nested exception is java.lang.IllegalArgumentException
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.resolveFieldValue(AutowiredAnnotationBeanPostProcessor.java:659) ~[spring-beans-5.3.21.jar:5.3.21]
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:639) ~[spring-beans-5.3.21.jar:5.3.21]
	at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:119) ~[spring-beans-5.3.21.jar:5.3.21]
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessProperties(AutowiredAnnotationBeanPostProcessor.java:399) ~[spring-beans-5.3.21.jar:5.3.21]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1431) ~[spring-beans-5.3.21.jar:5.3.21]
[...]

PR #29731, opened 2022-12-23
Issues 29314

huifer:
https://github.com/spring-projects/spring-framework/issues/29314


snicoll:
Thanks @huifer, I am going to close this. Sample repro with tests should be created separately from the main repository.

PR #29732, opened 2022-12-23
Support ordering of AOP advice annotations at the method level

weixiaobaoL:
Make the advice methods support order annotation to sorting.


jhoeller:
This does not look like an ideal with the `@Order` annotation to me since that annotation is primarily meant to distinguish between independent components, typically at the class level or factory method level. The semantics of ordering often only appear in resolution results where multiple such components end up. So while e.g. aspect classes can be ordered as part of a larger interceptor chain that way, advice methods within the same aspect class should rather rely on method name ordering or even get collapsed into combined methods. It's all within the same class there anyway, so there is arguably no need to establish a declarative order relationship between such closely related methods within the very same class.

PR #29738, opened 2022-12-24
Provide a CSV decoder and encoder

markusheiden:
For CSV Spring does not yet provide out-of-the-box WebClient streaming support. 
Because it is a commonly used and efficient data exchange format it would be nice to have built-in support.

This MR provides a Jackson-based CSV decoder and encoder.

Because Jackson does not yet provide CSV streaming support the decoder uses some rows lookahead to avoid blocking situations.


bclozel:
Thanks @markusheiden for this PR and all your efforts, and sorry for this very late response.
We haven't seen a lot of demand for CSV + streaming support in Spring WebFlux over the last months. Unfortunately, we don't feel that the added maintenance work is worth it. The main intent here is to provide must have support and reference implementations directly in Spring Framework; very popular codecs (like fastjson) and templating engine (thymeleaf) build and maintain that support independently.

This code could be contributed instead as an independent library. We would be happy to provide support in Spring Framework (codec or testing utilities) to make that easier.

I'm declining this PR as a result.

PR #29741, opened 2022-12-25
Add CharBufferDecoder

markusheiden:
This provides the optimization mentioned in a comment in the Jackson2JsonDecoder.


markusheiden:
https://github.com/spring-projects/spring-framework/blob/42b16591ec9978f9e317035bad998e617e79bb7e/spring-web/src/main/java/org/springframework/http/codec/json/Jackson2JsonDecoder.java#L76

poutsma:
It took us a while to resolve, but thank you for submitting a PR!

In order to reduce code duplication between `StringDecoder` and `CharBufferDecoder`, I created an abstract base class that is extended by both.

PR #29742, opened 2022-12-26
Avoid wasted memory on empty maps and sets

Dunemaster:
Empty LinkedHashMaps waste quite a lot of memory (250kb) for a medium size project

The PR introduces a simple fix

A screen from yourkit profiler:

image

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
@@ -107,15 +107,20 @@ public class InjectionMetadata {
 	}
 
 	public void checkConfigMembers(RootBeanDefinition beanDefinition) {
-		Set<InjectedElement> checkedElements = new LinkedHashSet<>(this.injectedElements.size());
-		for (InjectedElement element : this.injectedElements) {
-			Member member = element.getMember();
-			if (!beanDefinition.isExternallyManagedConfigMember(member)) {
-				beanDefinition.registerExternallyManagedConfigMember(member);
-				checkedElements.add(element);
-			}
+		if (this.injectedElements.isEmpty()) {
+			this.checkedElements = Set.of();
 		}
-		this.checkedElements = checkedElements;
+		else {
+			Set<InjectedElement> checkedElements = new LinkedHashSet<>(this.injectedElements.size());
+			for (InjectedElement element : this.injectedElements) {
+				Member member = element.getMember();
+				if (!beanDefinition.isExternallyManagedConfigMember(member)) {
+					beanDefinition.registerExternallyManagedConfigMember(member);
+					checkedElements.add(element);
+				}
+			}
+			this.checkedElements = checkedElements;
+		}
 	}
 
 	public void inject(Object target, @Nullable String beanName, @Nullable PropertyValues pvs) throws Throwable {
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
@@ -108,10 +108,10 @@ public class InjectionMetadata {
 
 	public void checkConfigMembers(RootBeanDefinition beanDefinition) {
 		if (this.injectedElements.isEmpty()) {
-			this.checkedElements = Set.of();
+			this.checkedElements = Collections.emptySet();
 		}
 		else {
-			Set<InjectedElement> checkedElements = new LinkedHashSet<>(this.injectedElements.size()*4/3);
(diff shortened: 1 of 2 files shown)


spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
```diff
@@ -107,15 +107,20 @@ protected boolean needsRefresh(Class<?> clazz) {
...
-		for (InjectedElement element : this.injectedElements) {
-			Member member = element.getMember();
-			if (!beanDefinition.isExternallyManagedConfigMember(member)) {
-				beanDefinition.registerExternallyManagedConfigMember(member);
-				checkedElements.add(element);
+		if (this.injectedElements.isEmpty()) {
+			this.checkedElements = Set.of();
+		}
+		else {
+			Set<InjectedElement> checkedElements = new LinkedHashSet<>(this.injectedElements.size());
```
vlsi, line 114: This should be size*4/3 to avoid hash resizes in the for loop below

vlsi: See https://richardstartin.github.io/posts/5-java-mundane-performance-tricks

Dunemaster: Fixed

jhoeller:
Any specific reason to use `Set/Map.of()` over `Collections.emptySet/Map()`? The latter seems even more efficient implementation-wise.

Dunemaster:
> Any specific reason to use `Set/Map.of()` over `Collections.emptySet/Map()`? The latter seems even more efficient implementation-wise.

Would you be so kind to explain the difference? Both methods return a static instance of an empty map/set

I used Set/Map.of() for the sake of brevity, and do not see any performance benefit is usiing either Set/Map.of() or Collections.emptySet/Map()

vlsi:
The difference between `Map.of` and `Collections.emptyMap` is that `Map.of().contains(null)` would throw exception while `Collections.emptyMap().contains(null)` would return false.

See https://stackoverflow.com/a/46406203/1261287

Dunemaster:
> The difference between `Map.of` and `Collections.emptyMap` is that `Map.of().contains(null)` would throw exception while `Collections.emptyMap().contains(null)` would return false.
> 
> See https://stackoverflow.com/a/46406203/1261287

wow, thanks.
 I will change to Collections.emptyMap, if you think that is better

spring-beans/src/main/java/org/springframework/beans/factory/annotation/InjectionMetadata.java
```diff
@@ -107,15 +107,20 @@ protected boolean needsRefresh(Class<?> clazz) {
...
-		for (InjectedElement element : this.injectedElements) {
-			Member member = element.getMember();
-			if (!beanDefinition.isExternallyManagedConfigMember(member)) {
-				beanDefinition.registerExternallyManagedConfigMember(member);
-				checkedElements.add(element);
+		if (this.injectedElements.isEmpty()) {
+			this.checkedElements = Set.of();
+		}
+		else {
+			Set<InjectedElement> checkedElements = new LinkedHashSet<>(this.injectedElements.size()*4/3);
```
mjd507, line 114: **size * 4 / 3 + 1** should be better for handling table resize corner cases. (like if the original size is 100, the capacity should be 134 instead of 133).

and for this PR, I don't think we need to specifically handle the empty list case as well as specify the capacity, it makes the code a little bit dirty.

Dunemaster: >and for this PR, I don't think we need to specifically handle the empty list case as well as specify the capacity, it makes the code a little bit dirty.

The data from memory dump shows otherwise.  Memory is wasted on numerous empty lists

jhoeller:
Aside from the `contains(null)` benefit, the footprint of `Collections.emptySet/Map()` should also be slightly better since those implementations are actually designed emptiness (not just initialized without any elements as with `Set/Map.of()`).

Dunemaster:
Replaced Set.of() and Map.of with Collections.emptyXXX in order to avoid possible compatibility breaks

vlsi (approved):
LGTM, thanks.

Frankly speaking, I incline that adding early return in `checkConfigMembers` would make the code slightly easier to follow.

mjd507 (commented):
Looks good for me (I’m just not quite sure about the impact of the code changes from high level, so let someone else to review and approve)

sbrannen:
This has been merged into `main` in 3738a45658426d7caa976021f01f94c125d9e4ef and revised in d5fb5d029bdc32fa00fa45660d07438a41b57206.

Thanks

Dunemaster:
Should this change be ported to 5.3?

sbrannen:
> Should this change be ported to 5.3?

We do not intend to backport minor enhancements such as this to `5.3.x`.

PR #29747, opened 2022-12-28
Use new Java features (switch expressions, text blocks, new JDK methods)

krzyk:
Converted few cases where new switch statements/expressions make the code more readable.
Few cases of multi line strings where converted to text blocks (unfortunately there is an issue of tabs/spaces because checkstyle doesn't know it is in textblock).

Also uses new methods to read strings directly.

### DIFF
--- a/spring-web/src/test/java/org/springframework/web/multipart/support/StandardMultipartHttpServletRequestTests.java
+++ b/spring-web/src/test/java/org/springframework/web/multipart/support/StandardMultipartHttpServletRequestTests.java
@@ -84,11 +84,13 @@ public class StandardMultipartHttpServletRequestTests {
 		new FormHttpMessageConverter().write(map, null, output);
 
 		assertThat(output.getBodyAsString(StandardCharsets.UTF_8)).contains(
-				"Content-Disposition: form-data; name=\"file\"; filename=\"myFile.txt\"\r\n" +
-						"Content-Type: text/plain\r\n" +
-						"Content-Length: 6\r\n" +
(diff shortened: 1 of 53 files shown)


spring-web/src/test/java/org/springframework/web/multipart/support/StandardMultipartHttpServletRequestTests.java
```diff
@@ -84,11 +84,13 @@ public void multipartFileResource() throws IOException {
...
-						"Content-Length: 6\r\n" +
-						"\r\n" +
-						"myBody\r\n");
+				"""
+						Content-Disposition: form-data; name="file"; filename="myFile.txt"\r
+						Content-Type: text/plain\r
+						Content-Length: 6\r
+						\r
+						myBody\r
+						""");
```
sbrannen, line 93: Instead of manually appending `\r` to every line in a text block, we prefer to write text blocks as "plain text" and then adjust the line endings only if necessary.

For example, in this particular case we would remove all `\r` entries and apply `.replace("\n", "\r\n")` to the text block.

NO need to make these changes in this PR though. I have already applied such changes locally which will be pushed to `main` along with your PR.

sbrannen:
This has been merged into `main` in afb8a0d1b18b711f5c27a3662058372e2e3cbbee and slightly revised in 0415975dd1124d4c667fa4ef88e33c9701ff6f23 and a4956dfe26f61b7eb1daa2a4b8c688fa0cb7c644.

Thanks

PR #29748, opened 2022-12-28
Fix wrong asserted code in SQLExceptionSubclassTranslatorTests

izeye:
It seems to be a copy-and-paste error.

See a644245


simonbasle (approved):
indeed looks like a typo. the test was still green due to the blanket acceptability of any `23505` combination

izeye:
This is not a user-facing enhancement, so the "type: task" label seems to be better for this.

PR #29750, opened 2022-12-28
JMS SingleConnectionFactory cached connection usage counter misbehavior

nicomz:
The SingleConnectionFactory wraps the connection inside a ProxyClass. 
As the connection starts a counter gets increased.  
When the proxied connection gets stopped the counter is evaluated to see 
if there is any other usage of that connection. 
In case of being the last one, connection#stop is called. 
If connection#stop is called and an exception is thrown, 
the counter decrement is skipped and then it's never called again.


jhoeller:
The current behavior is arguably by design since a failed stop attempt is not guaranteed to have stopped the connection, so we still treat it as started. This also allows for another application-level stop call to try stopping again (we would otherwise ignore such a subsequent stop call).

PR #29753, opened 2022-12-29
NoUniqueBeanDefinitionException should make sure beansNameFound is serializable

perlun:
Seen in Spring 5.3.18. `LinkedHashMap.keySet()` tends to return a `java.util.LinkedHashMap$LinkedKeySet` instance, which is not serializable. This becomes a problem then when you e.g. try to propagate the `NoUniqueBeanDefinitionException` error across a network boundary (our use case).

Wrapping these in `HashSet`s seem like the simplest way around. The only other call site was passing in an `Arrays.asList()` object, and `asList` is explicitly defined to return a serializable list.

(If this change is considered good/safe, it would perhaps be worth back-porting this to the Spring 5 branch as well. It does indeed change the semantics, but it could also arguably be considered a back-port safe bug fix.)

Thanks for a great project! :pray:

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/config/DependencyDescriptor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/config/DependencyDescriptor.java
@@ -21,6 +21,7 @@ import java.io.ObjectInputStream;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
+import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


perlun:
(Test failure seems unrelated to me. :thinking:)

snicoll:
Thanks again @perlun and that's a good catch. Rather than doing this at the caller level, I've fixed the exception itself so that it creates a copy of the  collection. This is similar to what we do for arrays.

PR #29781, opened 2023-01-07
Increase number of TOC levels in generated PDF documentation

vikeychen:
This was introduced by 4b22a4a0d.

This is single pdf file now.
image

This is splitted file before.
image

### DIFF
--- a/framework-docs/src/docs/asciidoc/page-layout.adoc
+++ b/framework-docs/src/docs/asciidoc/page-layout.adoc
@@ -1,4 +1,4 @@
 :toc: left
-:toclevels: 4
+:toclevels: 5
 :tabsize: 4
 :docinfo1:
--- a/framework-docs/src/docs/asciidoc/spring-framework.adocbook
+++ b/framework-docs/src/docs/asciidoc/spring-framework.adocbook
@@ -1,6 +1,7 @@
 :noheader:
 :toc:
 include::attributes.adoc[]
+include::page-layout.adoc[]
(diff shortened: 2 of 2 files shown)


framework-docs/src/docs/asciidoc/page-layout.adoc
```diff
@@ -1,4 +1,4 @@
 :toc: left
-:toclevels: 4
+:toclevels: 5
```
sbrannen, line 2: We explicitly selected 4 TOC levels and do not wish to change that to 5.

framework-docs/src/docs/asciidoc/spring-framework.adocbook
```diff
@@ -1,6 +1,7 @@
 :noheader:
 :toc:
 include::attributes.adoc[]
+include::page-layout.adoc[]
```
sbrannen, line 4: `page-layout.adoc` is specific to the HTML5 backend and as such should not be applied for the generated PDF.

For the PDF, we explicitly apply document configuration local to the `spring-framework.adocbook` file -- as can be seen by the existing `:noheader:` and `:toc:` entries.

sbrannen (commented):
Thanks for submitting the PR.

As I explained in my comments, to achieve the goal of increasing the number of TOC levels in the PDF we would prefer to modify the `spring-framework.adocbook` directly.

In light of that, I have updated the title of this PR and will make the necessary changes.

PR #29784, opened 2023-01-09
Add MediaType.TEXT_CSS

IAmLuisJ:
Add new MediaType "text/css" and TEXT_CSS_VALUE constant to more easily define a CSS resource in spring web. Text/CSS is a standard [MIME_Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textcss) and [MediaType](https://www.iana.org/go/rfc2318)


rstoyanchev:
Thanks for the suggestion, but we don't aim to expose all standard MIME types as constants via `MediaType`. You can create your own constant.

IAmLuisJ:
I agree it wouldn't make sense to cover all MIME Types but text/css has to be a pretty popular one for spring web wouldn't it? It currently supports text/html. 

I appreciate you reviewing the suggestion!

PR #29786, opened 2023-01-09
Allow to ignore null values in `BeanUtils.copyProperties()`

kslex:
I try to use `BeanUtils.copyProperties` in my local project and looking for ways to avoid copying null value. I saw [this close pull request](https://github.com/spring-projects/spring-framework/pull/25027).  I try to make it more compatible.


snicoll:
Thanks for the PR but, as stated in its javadoc, `BeanUtils` is primarily for our own use and such feature is not needed at the moment. For more advanced support, please consider the options [described in the javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/BeanUtils.html).

PR #29790, opened 2023-01-10
Improve generated default name for `@JmsListener` subscription

### LINKED ISSUE #29763: Wrong default name of the JMS subscription
I have a Spring application that uses annotation driven JMS listeners, i.e. I have some classes with methods annotated with `@JmsListener`. In this annotation, I do not set the attribute `subscription`.

However, since I want to work with topics (rather than queues) and use durable subsribers, I have configured a `JmsListenerContainerFactory` and set its property `subscriptionDurable` to `true`.

Now I expect all the JmsListeners to be durable and have sensible subscription names.

However, in the JMS admin tool, I see that all the subscriptions have the same name: `org.springframework.jms.listener.adapter.MessagingMessageListenerAdapter`.

I'd expect the default subscription name to be something like `<className>.<methodName>` (where "className" is the class that contains the method annotated with `@JmsListener`), but not always the same. This can even lead to an error where several JmsListeners will overwrite each other's subscriptions.

fml2:
This should fix #29763.

### DIFF
--- a/spring-jms/src/main/java/org/springframework/jms/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-jms/src/main/java/org/springframework/jms/listener/adapter/MessagingMessageListenerAdapter.java
@@ -73,7 +73,8 @@ public class MessagingMessageListenerAdapter extends AbstractAdaptableMessageLis
 	public String getSubscriptionName() {
 		if (this.handlerMethod != null) {
 			return this.handlerMethod.getBeanType().getName() + "#" + this.handlerMethod.getMethod().getName();
-		} else {
+		}
+		else {
 			return this.getClass().getName();
 		}
 	}
--- a/spring-jms/src/main/java/org/springframework/jms/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-jms/src/main/java/org/springframework/jms/listener/adapter/MessagingMessageListenerAdapter.java
@@ -20,6 +20,7 @@ import jakarta.jms.JMSException;
 import jakarta.jms.Session;
 
 import org.springframework.core.MethodParameter;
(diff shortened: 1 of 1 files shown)


fml2:
Could this also be backported to the 5.x line?

sbrannen:
As a proof of concept, I pushed changes to https://github.com/spring-projects/spring-framework/compare/main...sbrannen:spring-framework:issues/gh-29790-default-jms-subscription-name that include the method signature in the generated subscription name (in order to generate unique subscription names for overloaded `@JmsListener` methods).

However, before pushing those changes I decided to verify the requirements of a subscription name in the [JMS 3.0 spec](https://jakarta.ee/specifications/messaging/3.0/jakarta-messaging-spec-3.0.pdf). That points out that characters such as `(`, `)`, `,`, and `#` are not guaranteed to be supported. In addition, there's only a guarantee of support for up to 128 characters.

In light of that, I have decided to use a `<FQCN>.<method name>` (FQCN: fully qualified class name) format instead of the `<FQCN>#<method name>` format in this PR.

sbrannen:
> Could this also be backported to the 5.x line?

Yes, we will backport this to `5.3.x`.

fml2:
> In light of that, I have decided to use a `<FQCN>.<method name>` (FQCN: fully qualified class name) format instead of the `<FQCN>#<method name>` format in this PR.

This is perfectly OK for me. The main thing is that the name is meaningful and not fixed and is equal to the name of a rather internal Spring class.

sbrannen:
This has been merged into `main` in ad4e0d9ad75b011aae5a7ca3eb34240148e7802c and revised in 04321d0577aca893913fd9db7619d9b836ae058b.

PR #29796, opened 2023-01-10
Preserve existing persistence context property value during AOT processing

christophstrobl:
The generated `CodeBlock` that takes care of injecting the `PersistenceContext` may override an existing property value. This PR checks existing values and potentially omits creating the code block.

Related to: spring-projects/spring-data-jpa#2753 & spring-projects/spring-data-commons#2760

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/support/PersistenceAnnotationBeanPostProcessor.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/support/PersistenceAnnotationBeanPostProcessor.java
@@ -181,6 +181,7 @@ import org.springframework.util.StringUtils;
  * @author Juergen Hoeller
  * @author Stephane Nicoll
  * @author Phillip Webb
+ * @author Christoph Strobl
  * @since 2.0
  * @see jakarta.persistence.PersistenceUnit
  * @see jakarta.persistence.PersistenceContext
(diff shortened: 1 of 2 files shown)


sdeleuze:
Worth to review with @snicoll when he is back from PTO.

@christophstrobl Is this PR ready for review for potential merge or do you plan further refinement?

christophstrobl:
@sdeleuze you can review it already. Happy to modify if necessary.

snicoll:
So it looks like something has post-processed the persistence context already and the core goes in the way. I think we need to fix this differently. 

If I understand correctly, Spring Data has knowledge of something that we don't. Why isn't that reflected in the `BeanDefinition`. If it is, why is the skip necessary?

christophstrobl:
There's an `EntityManagerBeanDefinitionRegistrarPostProcessor` in data-jpa that post processes `EntityManagerFactory`. @odrotbohm can you please help us find the right point for the changes.

mp911de:
Spring Data is setting the property value for `entityManager` within the `BeanDefinition`. In JVM mode, the container skips `@PersistenceContext` post-processing while AOT mode works differently.

snicoll:
`InjectionMetadata#inject` has a `checkPropertySkipping` that is meant to determine if injection should take place. I missed that while building the AOT contribution for the persistence context. It should attempt to run the same code so that no contribution takes place if it should be skipped.

snicoll:
Thanks for the PR @christophstrobl but I think we should tackle this one differently. I've created #30476

Issue #29806, opened 2023-01-12
There should be more proper documentation for the the file AbstractMergedAnnotation.java


deepak1067:
**Affects:** \<Spring Framework version>

---

bclozel:
Closing in favor of #29807

PR #29807, opened 2023-01-12
Documentaion updated in AbstractMergedAnnotation.java file 

deepak1067:
Issue solved for #29806


bclozel:
The public methods are already documented in the parent type. As for the private methods, we don't document them unless there is a specific need to it. Thanks, but I'm declining this as a result.

PR #29809, opened 2023-01-12
Fix integer multiplications by implicitly cast to long

alidandach:
The `Thread.sleep()` takes long value as a parameter, so instead of integer value we should use a long value


sbrannen:
Unless I'm missing something, the changes you have made are unnecessary since the Java compiler will automatically perform the appropriate widening conversion from `int` to `long`.

So, what is the rationale for this PR?

alidandach:
Dear @sbrannen , 
yes, you are right. it's a small change.
But, I believe that it's better to emphasize the data type instead of throwing it on the compiler.

Thank you for the quick answer.

sbrannen:
Thanks for the response, @alidandach.

We likely have numerous places across the code base where we rely on the compiler to perform necessary widening conversions, and we do not plan to make changes across the code base to address that. The same applies to auto-boxing and auto-unboxing: we do not explicitly perform the boxing/unboxing manually.

PR #29811, opened 2023-01-13
Add hasSmartInstantiationAwareBeanPostProcessors method to clarify semantics 

supalle:
The `instantiationAware` and `smartInstantiationAware` of `BeanPostProcessorCache` are not exactly equal. Using `hasSmartInstantiationAwareBeanPostProcessors()` to replace `hasInstantiationAwareBeanPostProcessors()` before using `smartInstantiationAware` can not only optimize the code, but also clarify the semantics.


snicoll:
@supalle it's really hard to make up our mind based on the proposal as it doesn't have any test that illustrates what you mean by clarifying the semantics. Can you please clarify?

supalle:
有点久远了，我重新看了一下，问题不大，无伤大雅。

PR #29816, opened 2023-01-13
Deprecate HttpStatus 103 CHECKPOINT in favor of new EARLY_HINTS

simonbasle:
This commit takes rfc8297 into account and introduces a newer code 103
HttpStatus value which uses `Early Hints` as the more correct reason
phrase, deprecating the outdated `CHECKPOINT` enum value for 103.

Additionally:
  - `HttpStatus.valueOf(103)` will return the new enum value
  - `HttpStatusCode#isSameCodeAs(HttpStatusCode)` is introduced to ease
  comparison of deprecated enums vs their newer counterparts (or any
  instance of a more generic `HttpStatusCode`) by comparing the integer
  `value()`
  - `HttpStatusTests` covers the new deprecation as well as the three
  previously deprecated codes, including a check with the above new
  method to ensure they have comparable integer values

Supersedes and Closes gh-27960

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/result/StatusResultMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/result/StatusResultMatchers.java
@@ -148,7 +148,7 @@ public class StatusResultMatchers {
 	 * @see #isEarlyHints()
 	 * @deprecated in favor of {@link #isEarlyHints()}
 	 */
-	@Deprecated
+	@Deprecated(since = "6.0")
 	public ResultMatcher isCheckpoint() {
 		return isEarlyHints();
 	}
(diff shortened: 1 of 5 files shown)


bclozel:
LGTM!
I think we tend to align all deprecations to look like the following: `@Deprecated(since = "6.0", forRemoval = true)`.
Usage sites require `@SuppressWarnings("removal")`.

simonbasle:
Good to know. I wonder if that should be a candidate for backport. I guess if we decided in favor of backport, the `since` portion of the `@Deprecated` in java would reflect that?

bclozel:
Sticking to 6.0 is fine I think. Backporting deprecations that late in the 5.x generation is not useful.

simonbasle:
Perfect. I've updated the `@Deprecated` to add `since` info.

I see a lot of occurrences of the annotation without any params (351 vs 182), and only 39 matches with `forRemoval`, so I haven't used the later (as I don't think there is any particular urge to remove the deprecated enum value)

PR #29837, opened 2023-01-17
Fix unnecessary 'public' constructor from abstract class `DecoratingClassLoader`

lauvsong:
Constructors of abstract classes can only be called in constructors of their subclasses.
So there is no point in making them public.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/DecoratingClassLoader.java
+++ b/spring-core/src/main/java/org/springframework/core/DecoratingClassLoader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @lauvsong,

Thanks for submitting your first pull request for the Spring Framework! 👍 

Regarding the proposed changes, we actually have numerous places across the code base where methods or constructors are declared as `public` in `abstract` or non-public types.

Thus, although the rationale for the proposed changes is technically correct, we would prefer not to accept PRs that make such small changes, in order to discourage the community from submitting similar PRs for which we unfortunately do not have the resources to process.

If you would like to make contributions to the Spring Framework, consider looking for issues with documentation or see if there are any issues that have been assigned the [ideal-for-contribution](https://github.com/spring-projects/spring-framework/labels/status%3A%20ideal-for-contribution) label.

lauvsong:
I can relate to. Thank you for the detail feedback 👍

Issue #23499, opened 2019-08-22
Bean validation doesn't work with Kotlin coroutines controller


yuki-teraoka:
**Affects:**  5.2.0.RC1

when using Kotlin suspend function and Validated annotation, like this.

    package example
    
    import org.springframework.http.HttpStatus
    import org.springframework.http.ResponseEntity
    import org.springframework.validation.annotation.Validated
    import org.springframework.web.bind.annotation.RequestMapping
    import org.springframework.web.bind.annotation.RestController

    @Validated
    @RestController
    class Foo {
        @RequestMapping("/foo")
        suspend fun foo() = ResponseEntity("foo", HttpStatus.OK)
    }

occured exception.

    java.lang.ArrayIndexOutOfBoundsException: 0
            at java.util.Arrays$ArrayList.get(Arrays.java:3841) ~[na:1.8.0_102]
            Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException:
    Error has been observed at the following site(s):
            |_ checkpoint ⇢ Handler example.Foo#foo(Continuation) [DispatcherHandler]
            |_ checkpoint ⇢ org.springframework.security.web.server.authorization.AuthorizationWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.web.server.authorization.ExceptionTranslationWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.web.server.authentication.logout.LogoutWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.web.server.savedrequest.ServerRequestCacheWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.web.server.context.SecurityContextServerWebExchangeWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.web.server.context.ReactorContextWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.web.server.header.HttpHeaderWriterWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.config.web.server.ServerHttpSecurity$ServerWebExchangeReactorContextWebFilter [DefaultWebFilterChain]
            |_ checkpoint ⇢ org.springframework.security.web.server.WebFilterChainProxy [DefaultWebFilterChain]
[...]

PR #29840, opened 2023-01-18
fix the kotlin suspend function validation issure 23499.

cjdxhjj:
the issure is https://github.com/spring-projects/spring-framework/issues/29793


sdeleuze:
Thanks for this PR, I don't think I would fix it this way, but this allowed me to achieve I think a similar outcome just by removing `KotlinReflectionParameterNameDiscoverer` usage in `LocalValidatorFactoryBean`, see https://github.com/spring-projects/spring-framework/issues/23499#issuecomment-1386704975. So please try my change instead and provide your feedback as a comment in #23499. Since we will likely use another implementation of the same idea, I will close this PR unmerged.

PR #29851, opened 2023-01-18
Add support for JSR-330 and Jakarta `@Inject` for autowiring test constructors

FlorianLehmann:
## Overview

Spring 3.0 added support for JSR-330. However, the `spring-test` module is only supporting `@Autowired` for dependency injection.

This PR intends to add `@Inject` support for `spring-test`. Since Spring Framework 6 still supports `@javax.inject.Inject` in addition to `@jakarta.inject.Inject`, this PR reintroduces support for `@javax.inject.Inject` in `spring-test` as well.

## Examples

The following test passes:

```java
package com.example.demo;

@SpringJUnitConfig
public class FooTests {

	private final String foo;

	@Autowired
	public FooTests(String foo) {
		this.foo = foo;
	}

	@Test
	public void beanInjected() {
		assertThat(this.foo).isEqualTo("foo");
	}

	@Configuration
	static class Config {

		@Bean
		String foo() {
			return "foo";
		}

	}
}
```

However, if `@Autowired` is replaced with `@javax.inject.Inject` the test fails:

```java
package com.example.demo;

import jakarta.inject.Inject;

@SpringJUnitConfig
public class FooTests {

	private final String foo;

	@Inject
	public FooTests(String foo) {
		this.foo = foo;
	}

	@Test
	public void beanInjected() {
		assertThat(this.foo).isEqualTo("foo");
	}

	@Configuration
	static class Config {

		@Bean
		String foo() {
			return "foo";
		}

	}
}
```

## Error

> org.junit.jupiter.api.extension.ParameterResolutionException: No ParameterResolver registered for parameter [java.lang.String foo] in constructor [public com.example.demo.FooTests(java.lang.String)].

### DIFF
--- a/spring-test/spring-test.gradle
+++ b/spring-test/spring-test.gradle
@@ -76,6 +76,7 @@ dependencies {
 	}
 	testImplementation("io.projectreactor.netty:reactor-netty-http")
 	testImplementation("de.bechte.junit:junit-hierarchicalcontextrunner")
+	testImplementation("javax.inject:javax.inject:1")
 	testRuntimeOnly("org.junit.vintage:junit-vintage-engine") {
 		exclude group: "junit", module: "junit"
 	}
--- a/spring-test/src/main/java/org/springframework/test/context/support/TestConstructorUtils.java
+++ b/spring-test/src/main/java/org/springframework/test/context/support/TestConstructorUtils.java
@@ -16,8 +16,11 @@
 
 package org.springframework.test.context.support;
 
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.SpringProperties;
@@ -26,6 +29,7 @@ import org.springframework.lang.Nullable;
 import org.springframework.test.context.TestConstructor;
 import org.springframework.test.context.TestConstructor.AutowireMode;
 import org.springframework.test.context.TestContextAnnotationUtils;
+import org.springframework.util.ClassUtils;
 
 /**
  * Utility methods for working with {@link TestConstructor @TestConstructor}.
@@ -33,6 +37,7 @@ import org.springframework.test.context.TestContextAnnotationUtils;
  * <p>Primarily intended for use within the framework.
  *
  * @author Sam Brannen
+ * @author Florian Lehmann
  * @since 5.2
  * @see TestConstructor
  */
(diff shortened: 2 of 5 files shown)


sbrannen:

sbrannen:
Hi @FlorianLehmann,

> However, the `spring-test` module is only supporting `@Autowired` for dependency injection.

That's not entirely true.

The Spring TestContext Framework supports whatever DI annotations are supported by `AutowiredAnnotationBeanPostProcessor`, which includes `@Autowired` and `@Value` as well as `@jakarta.inject.Inject` and `@javax.inject.Inject` (if the latter two are available on the classpath).

Thus, you are free to use `@Inject` for fields and methods in your test classes.

The only restriction regarding `@Autowired` vs `@Inject` is that `@Inject` cannot be used for autowiring test class constructors when using JUnit Jupiter (but `@Autowired` can).

In addition, `@Inject` cannot be used to autowire a particular constructor or method argument simply because `@Inject` is configured with `@Target({ METHOD, CONSTRUCTOR, FIELD })`. So that limitation is inherent to `@Inject` and therefore out of Spring's control.

If I recall correctly, the latter is the reason that we decided not to support `@Inject` for JUnit Jupiter features: we wanted to support `@Autowired` which can be used consistently for test class constructors as well as individual constructor/method arguments.

However, we will consider adding support for `@Inject` for test class constructors, and I have changed the title of this issue to reflect that.

FlorianLehmann:
Hi @sbrannen,

> I've edited your comment to improve the formatting. You might want to check out this [Mastering Markdown guide](https://guides.github.com/features/mastering-markdown/) for future reference.

Thank you I'll definitely have a look.


> In addition, `@Inject` cannot be used to autowire a particular constructor or method argument simply because `@Inject` is configured with `@Target({ METHOD, CONSTRUCTOR, FIELD })`. So that limitation is inherent to `@Inject` and therefore out of Spring's control.

I believe the same limitation applies for "normal" constructor or method argument. The idea of this PR is to remove the discrepancy between tests and "normal" code when autowiring constructor using @Inject.


> However, we will consider adding support for `@Inject` for test class constructors, and I have changed the title of this issue to reflect that.

:+1: 

Thank you again for your comments.

sbrannen (approved):
I think the proposal looks good in general, though I might revise it a bit after merging into `main`.

sbrannen:
This has been merged into `main` in 8dd857a84d7f5d8c7903f445303906c887c70281 and revised in dfea3d05aa4abd0493a2714c0a77edd13a4a2f18, for inclusion in Spring Framework 6.1.

Thanks for your contribution, @FlorianLehmann! 🍃

PR #29853, opened 2023-01-19
Optimize CorsConfiguration::checkHeaders

ThomHurks:
In the checkHeaders method, if allowedHeaders contains * the result size can be requestHeaders.size(), but if this is not the case then the result size can only be as big as the minimum value of requestHeaders.size() and allowedHeaders.size(). requestHeaders.size() can be potentially big (user input) and allowedHeaders.size() could be small, so this saves memory in that case.


ThomHurks:
Thanks :heart:

PR #29854, opened 2023-01-19
Remove extra empty line in reactive `DefaultCorsProcessor`

ThomHurks:


sbrannen:
Thanks for the PR.

We normally prefer PRs with a larger scope than this one. In the future, if you could across something as minor as this, feel free to just raise an issue and one of the maintainers can then include the change in a larger commit.

In any case, I've merged this PR as a `task`.

ThomHurks:
Ah, will do, thanks!

PR #29875, opened 2023-01-23
use constants BEARER and BASIC instead of String literals

shin-mallang:
I think using constant BEARER instead of "Bearer ", and constant BASIC instead of "Basic " is better to reduce the risk of mistakes and increase reusability.


bclozel:
`"Bearer "` and `"Basic "` are not HTTP header names, so it doesn't make sense to expose them as public constants on `HttpHeaders`. I'm declining this PR as a result. Thanks!

PR #29882, opened 2023-01-25
Update Javadoc on async request timeout value in HandlerFunctionAdapter

srivatsa-cfp:
- Validate the timeout value passed to setAsyncRequestTimeout is positive value.
- Avoid setting negative value. If negative value is passed then set it to default timeout.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/function/support/HandlerFunctionAdapter.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/function/support/HandlerFunctionAdapter.java
@@ -73,12 +73,14 @@ public class HandlerFunctionAdapter implements HandlerAdapter, Ordered {
 	 * should time out. In Servlet 3, the timeout begins after the main request
 	 * processing thread has exited and ends when the request is dispatched again
 	 * for further processing of the concurrently produced result.
-	 * <p>If this value is not set, the default timeout of the underlying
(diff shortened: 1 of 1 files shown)


srivatsa-cfp:
> [The spec says](https://github.com/jakartaee/servlet/blob/master/spec/src/main/asciidoc/servlet-spec-body.adoc) the value should be 0 (never time out) or greater:
> 
> > Sets the time out for the asynchronous processing in milliseconds. A call to this method overrides the time out set by the container. If the time out is not specified via the call to setTimeout, 30000 is used as the default. A value of 0 or less indicates that the asynchronous operation will never time out. The time out applies to the AsyncContext once the container-initiated dispatch, during which one of the ServletRequest.startAsync methods was called, has returned to the container. It is illegal to call this method after the container-initiated dispatch on which the asynchronous cycle was started has returned to the container and will result in an IllegalStateException.
> 
> From a quick look, it does seem like the Servlet containers ignore a negative value. From that perspective, the proposed change doesn't change anything, but a Javadoc update could be useful.

 Thanks for sharing the information. Do you want to route the proposal to updating the Javadoc?

srivatsa-cfp:
Updated the JavaDoc

- Update the Javadoc -  the setting 0 or negative value is indicates that the asynchronous operation will never time out.
- Helps the developer and maintainer to better understand the implementation.

PR #29889, opened 2023-01-27
StringDecoder causes direct allocation of ByteBuffer

xavier-b:
Since my switch to Spring Boot 3, I have a lot of direct allocations in my metrics.
Indeed, my app use `StringDecoder` via, for example, `@RequestBody String body`

In `org.springframework.core.codec.StringDecoder`, when dataBuffer is backed by an NIO direct buffer, `decode` method use `dataBuffer.toByteBuffer()` which will call `java.nio.ByteBuffer.allocateDirect(int)`.

In `java.nio.ByteBuffer` javadoc we can read :
> The buffers returned by this method typically have somewhat higher allocation and deallocation costs than non-direct buffers. [...]
> It is therefore recommended that direct buffers be allocated primarily for large,
> long-lived buffers that are subject to the underlying system's native I/O operations.
> In general it is best to allocate direct buffers only when they yield a measurable
> gain in program performance.

Goal of this code in `StringDecoder` was to decode with a charset, so maybe it's better to use `dataBuffer.toString(charset)` which doesn't do direct allocation.
I see the same problem in `org.springframework.http.codec.FormHttpMessageReader`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/codec/StringDecoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/StringDecoder.java
@@ -189,9 +189,8 @@ public final class StringDecoder extends AbstractDataBufferDecoder<String> {
 			@Nullable MimeType mimeType, @Nullable Map<String, Object> hints) {
 
 		Charset charset = getCharset(mimeType);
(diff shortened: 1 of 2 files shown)


poutsma:
Merged, thanks for spotting this!

PR #29890, opened 2023-01-27
`Validator` factory methods with `BiConsumer`

making:
This commit introduces `of` method in `Validator` to provide a way to create a validator for the specific type `<T>` using `BiConsumer<T, Errors>` and define the validator in a functional way.

```java
Validator passwordEqualsValidator = Validator.of(PasswordResetForm.class)
    .apply((form, errors) -> {
        if (!Objects.equals(form.getPassword(), form.getConfirmPassword())) {
            errors.rejectValue("confirmPassword",
                "PasswordEqualsValidator.passwordResetForm.password",
                "password and confirm password must be same.");
        }
    });
```

This also eliminates the boilerplate for implementing the `supports` method.

BONUS: Supporting BiConsumer allows us to magically adapt 3rd party Validators (like [YAVI](https://yavi.ik.am/#obtaining-a-bivalidator)) that return BiConsumer. 
(I'm the creator of YAVI.)

(original proposal)
```java
Validator validator = Validator.of(CartItem.class)
    .apply(ValidatorBuilder.<CartItem>of()
        .constraint(CartItem::itemId, "itemId", c -> c.notEmpty())
        .constraint(CartItem::amount, "amount", c -> c.greaterThanOrEqual(0))
        .constraint(CartItem::price, "price", c -> c.greaterThanOrEqual(0))
        .build(Errors::rejectValue));
```

(revised version)
```java
Validator validator = Validator.forInstanceOf(CartItem.class, ValidatorBuilder.<CartItem>of()
    .constraint(CartItem::itemId, "itemId", c -> c.notEmpty())
    .constraint(CartItem::amount, "amount", c -> c.greaterThanOrEqual(0))
    .constraint(CartItem::price, "price", c -> c.greaterThanOrEqual(0))
    .build(Errors::rejectValue));
```

### DIFF
--- a/spring-context/src/main/java/org/springframework/validation/Validator.java
+++ b/spring-context/src/main/java/org/springframework/validation/Validator.java
@@ -114,11 +114,10 @@ public interface Validator {
 	 *             "password and confirm password must be same.");
 	 *       }
 	 *     });</pre>
-	 *
 	 * @param targetClass  the class of the object that is to be validated
-	 * @return The {@link Function} that takes the {@link BiConsumer} containing the validation logic and
-	 * returns the {@link Validator} instance
 	 * @param <T> the type of the object that is to be validated
+	 * @return the {@link Function} that takes the {@link BiConsumer} containing the validation logic and
+	 * returns the {@link Validator} instance
 	 */
 	static <T> Function<BiConsumer<T, Errors>, Validator> of(Class<T> targetClass) {
 		Assert.notNull(targetClass, "'targetClass' must not be null.");
--- a/spring-context/src/main/java/org/springframework/validation/Validator.java
+++ b/spring-context/src/main/java/org/springframework/validation/Validator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,11 @@
  */
 
 package org.springframework.validation;
+
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+import org.springframework.util.Assert;
 
 /**
  * A validator for application-specific objects.
@@ -59,6 +64,7 @@ package org.springframework.validation;
  * application.
  *
  * @author Rod Johnson
+ * @author Toshiaki Maki
  * @see SmartValidator
  * @see Errors
  * @see ValidationUtils
@@ -92,4 +98,40 @@ public interface Validator {
 	 */
 	void validate(Object target, Errors errors);
 
+	/**
+	 * Returns the {@link Function} that takes the {@link BiConsumer} containing validation logic for the specific type
+	 * <code>&lt;T&gt;</code> and returns the {@link Validator} instance.<br>
+	 * This validator implements the <i>typical</i> {@link #supports(Class)} method
+	 * for the given <code>&lt;T&gt;</code>.<br>
+	 *
+	 * By using this {@link #of(Class)} method, a {@link Validator}  can be implemented as follows:
+	 *
+	 * <pre class="code">Validator passwordEqualsValidator = Validator.of(PasswordResetForm.class)
+	 *     .apply((form, errors) -> {
+	 *       if (!Objects.equals(form.getPassword(), form.getConfirmPassword())) {
+	 *         errors.rejectValue("confirmPassword",
+	 *             "PasswordEqualsValidator.passwordResetForm.password",
+	 *             "password and confirm password must be same.");
+	 *       }
+	 *     });</pre>
+	 *
+	 * @param targetClass  the class of the object that is to be validated
+	 * @return The {@link Function} that takes the {@link BiConsumer} containing the validation logic and
+	 * returns the {@link Validator} instance
+	 * @param <T> the type of the object that is to be validated
+	 */
+	static <T> Function<BiConsumer<T, Errors>, Validator> of(Class<T> targetClass) {
+		Assert.notNull(targetClass, "'targetClass' must not be null.");
+		return validator -> new Validator() {
+			@Override
+			public boolean supports(Class<?> clazz) {
+				return targetClass.isAssignableFrom(clazz);
+			}
+
+			@Override
+			public void validate(Object target, Errors errors) {
+				validator.accept(targetClass.cast(target), errors);
+			}
+		};
+	}
 }
--- a/spring-context/src/main/java/org/springframework/validation/Validator.java
+++ b/spring-context/src/main/java/org/springframework/validation/Validator.java
@@ -17,7 +17,6 @@
 package org.springframework.validation;
 
 import java.util.function.BiConsumer;
-import java.util.function.Function;
 
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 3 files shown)


spring-context/src/main/java/org/springframework/validation/Validator.java
```diff
@@ -92,4 +98,39 @@ public interface Validator {
...
+	 *       }
+	 *     });</pre>
+	 * @param targetClass  the class of the object that is to be validated
+	 * @param <T> the type of the object that is to be validated
+	 * @return the {@link Function} that takes the {@link BiConsumer} containing the validation logic and
+	 * returns the {@link Validator} instance
+	 */
+	static <T> Function<BiConsumer<T, Errors>, Validator> of(Class<T> targetClass) {
+		Assert.notNull(targetClass, "'targetClass' must not be null.");
+		return validator -> new Validator() {
```
quaff, line 124: Variable name `validator` is a bit confused here since it's not type of `Validator` but `BiConsumer`.

sbrannen: Indeed, the naming is a bit tricky.

That `BiConsumer` effectively implements the `validate(Object, Errors)` method from `Validator`. So, it's more like "validation logic" or "validation delegate". 

Anyone have any good suggestions?

spring-context/src/test/java/org/springframework/validation/DataBinderTests.java
```diff
@@ -82,6 +82,17 @@
  */
 class DataBinderTests {
 
+	Validator spouseValidator = Validator.of(TestBean.class)
+			.apply((tb, errors) -> {
```
sbrannen, line 86: Seeing `apply()` in this "DSL" seems counterintuitive to me.

I'm wondering if we should introduce a dedicated functional interface for this purpose (with a meaningful type name and meaningful method name) instead of reusing `BiConsumer`.

poutsma: The `apply` comes from `Function`; not `BiConsumer` (where it is called `accept`).

spring-context/src/main/java/org/springframework/validation/Validator.java
```diff
@@ -92,4 +98,39 @@ public interface Validator {
...
+	 *             "PasswordEqualsValidator.passwordResetForm.password",
+	 *             "password and confirm password must be same.");
+	 *       }
+	 *     });</pre>
+	 * @param targetClass  the class of the object that is to be validated
+	 * @param <T> the type of the object that is to be validated
+	 * @return the {@link Function} that takes the {@link BiConsumer} containing the validation logic and
+	 * returns the {@link Validator} instance
+	 */
+	static <T> Function<BiConsumer<T, Errors>, Validator> of(Class<T> targetClass) {
```
poutsma, line 122: I would expect `Validator::of` to return a `Validator`; not a function.

poutsma:
A nice proposal to modernise `Validator`, thanks @making!

I do find the usage of `Function` (and the resulting `apply`) a bit complicated, and I don't believe it's necessary. Instead, you can have the `BiConsumer` as a second parameter for the `of` method, like so:

```java
static <T> Validator of(Class<T> targetClass, BiConsumer<T, Errors> validator) {
	return new Validator() {
		@Override
		public boolean supports(Class<?> clazz) {
			return targetClass.isAssignableFrom(clazz);
		}

		@Override
		public void validate(Object target, Errors errors) {
			validator.accept(targetClass.cast(target), errors);
		}
	};
}
```

poutsma:
@making, We would like to make this part of Spring Framework 6.1. Are you ok with changing the PR to reflect our suggestions, or would you like me to pick it up from here?

sdeleuze:
+1 for such functional approach for `Validator`.

making:
@poutsma Updated the pr!

spring-context/src/main/java/org/springframework/validation/Validator.java
making: renamed to `delegate`

spring-context/src/test/java/org/springframework/validation/DataBinderTests.java
sbrannen: Indeed. I got the interfaces confused. In any case, the point I was making is the same.

sbrannen:
> Instead, you can have the `BiConsumer` as a second parameter for the `of` method

That's a big improvement. 👍

poutsma:
> @poutsma Updated the pr!

Looking good!

A couple of improvements that our team discussed before it can be merged in 6.1.
- It would be nice to have a non-opinionated variant of the factory method that takes just a predicate and biconsumer, something like:
`static <T> Validator of(Predicate<Class<?>> supports, BiConsumer<T, Errors> delegate)`
 This allows users to have their own `supports` logic in a predicate.
- I think it's cleaner to move the anonymous `Validator` implementation into its own package-protected class named `DefaultValidator` (or something similar). This helps keep the `Validator` interface cleaner, and also improves the readability of any potential stacktrace.
- Sometimes you want to specify an exact class match, instead of allowing all subclasses. So instead of using `targetClass.isAssignableFrom`, you'd use `targetClass.equals`. @rstoyanchev suggested to have two separate methods: 
  - `forInstanceOf(Class<T>, BiConsumer<T, Errors>)`  (using `isAssignableFrom`, similar to the current PR), and
  - `forType(Class<T>, BiConsumer<T, Errors>)` (using `equals`).
 Note that `forInstanceOf` and `forType` can be implemented using the predicate-based `of` method suggested above (i.e. `of(targetClass::isAssignableFrom, delegate)` and `of(targetClass::equals, delegate)`).

Let me know what you think about these suggestions.

making:
@poutsma 
I'm in favor of having both `forInstanceOf` and `forType`, but I'm not sure it makes sense to have the `of(Predicate<Class<?>> supports, BiConsumer<T, Errors> delegate)` method.
Since the `validate` methods accepts any object, to delegate validation logic to a `BiConsumer<T, Errors>`, the `supports` method must ensure that the object is instance of the `T` class.

I think it would make more sense to have a class like bellow, what do you think?

```java
class TypedValidator<T> implements Validator {

	private final Class<T> targetClass;

	private final BiPredicate<Class<T>, Class<?>> typeVerifier;

	private final BiConsumer<T, Errors> delegate;

	public TypedValidator(Class<T> targetClass, BiPredicate<Class<T>, Class<?>> classVerifier, BiConsumer<T, Errors> delegate) {
		this.targetClass = targetClass;
		this.typeVerifier = classVerifier;
		this.delegate = delegate;
	}

	@Override
	public boolean supports(Class<?> clazz) {
		return this.typeVerifier.test(this.targetClass, clazz);
	}

	@Override
	public void validate(Object target, Errors errors) {
		this.delegate.accept(this.targetClass.cast(target), errors);
	}
}
```

and factory methods in `Validator`

```java
	static <T> Validator forInstanceOf(Class<T> targetClass, BiConsumer<T, Errors> delegate) {
		return new TypedValidator<>(targetClass, Class::isAssignableFrom, delegate);
	}

	static <T> Validator forType(Class<T> targetClass, BiConsumer<T, Errors> delegate) {
		return new TypedValidator<>(targetClass, Objects::equals, delegate);
	}
```

poutsma:
@making Sorry for not responding earlier.

I'm liking that `TypedValidator`, but if I find the `BiPredicate<Class<T>, Class<?>>` a bit confusing. I think it would be simpler and clearer to have a `Predicate<Class<?>>` instead, and in `forInstanceOf` and `forType` use the `targetClass` parameter instead of the field.

making:
@poutsma 
Sorry for the late reply. Did you mean

```java
package org.springframework.validation;

import java.util.function.BiConsumer;
import java.util.function.Predicate;

class TypedValidator<T> implements Validator {

	private final Class<T> targetClass;

	private final Predicate<Class<?>> typeVerifier;

	private final BiConsumer<T, Errors> delegate;

	public TypedValidator(Class<T> targetClass, Predicate<Class<?>> typeVerifier, BiConsumer<T, Errors> delegate) {
		this.targetClass = targetClass;
		this.typeVerifier = typeVerifier;
		this.delegate = delegate;
	}

	@Override
	public boolean supports(Class<?> clazz) {
		return this.typeVerifier.test(clazz);
	}

	@Override
	public void validate(Object target, Errors errors) {
		this.delegate.accept(this.targetClass.cast(target), errors);
	}
}
```

and

```java
	static <T> Validator forType(Class<T> targetClass, BiConsumer<T, Errors> delegate) {
		return new TypedValidator<>(targetClass, clazz -> Objects.equals(targetClass, clazz), delegate);
	}

	static <T> Validator forInstanceOf(Class<T> targetClass, BiConsumer<T, Errors> delegate) {
		return new TypedValidator<>(targetClass, targetClass::isAssignableFrom, delegate);
	}
```

?

poutsma:
Thank you, @making, for submitting this PR. It has now been merged , see dd9f03d9b9c858d44bfc4b64be06e064ce9a0334.

I made some changes, including the `TypedValidator` that was mentioned above. Hopefully it is now clear what I meant.

Feel free to comment or create subsequent PRs if you have any further suggestions regarding this piece of functionality.

Issue #26344, opened 2021-01-04
Sending empty Flux to RSocket channel causes DecodingException: JSON decoding error: No content to map due to end-of-input


MastaP:
Spring Framework Version：5.3.2

Problem background:
The application needs to send a result of a db query over RSocket channel. In some cases there are no elements in the result set and empty Flux is sent. This is normal from the business logic point of view.

I had a discussion about this problem with @OlegDokuka and he suggested to file the issue here (https://gitter.im/rsocket/rsocket-java?at=5fe36654ac9d8e7463c97042)

Application (server) entrypoint:

```java
    @MessageMapping("objectChannel.{test}")
    Flux<Instant> rsocketObjectChannel(@DestinationVariable String test, Flux<Instant> flux) {
        return flux
                .map(i -> i.plus(1, DAYS))
                .doOnNext(s -> System.out.println("rsocketObjectChannel.next: " + s))
                .doOnComplete(() -> System.out.println("rsocketObjectChannel complete"));
    }
```

Test (client) side:

```java
    @Test
    void testEmptyObjectChannel() {
        Flux<Instant> data = Flux.empty();
        Flux<Instant> result = requester.route("objectChannel.{test}", "route-var1")
                .data(data)
                .retrieveFlux(Instant.class)
                .doOnComplete(() -> System.out.println("testObjectChannel complete"));
        StepVerifier
                .create(result)
                .verifyComplete();
    }
```

Result:

```
org.springframework.messaging.handler.invocation.MethodArgumentResolutionException: Could not resolve method parameter at index 1 in reactor.core.publisher.Flux<java.time.Instant> com.gt.example.reactivedemo.RSocketServerHandler.rsocketObjectChannel(java.lang.String,reactor.core.publisher.Flux<java.time.Instant>): Failed to read HTTP message; nested exception is org.springframework.core.codec.DecodingException: JSON decoding error: No content to map due to end-of-input; nested exception is com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (io.netty.buffer.ByteBufInputStream); line: -1, column: 0]
	at org.springframework.messaging.handler.annotation.reactive.PayloadMethodArgumentResolver.handleReadError(PayloadMethodArgumentResolver.java:266) ~[spring-messaging-5.3.2.jar:5.3.2]
[...]

PR #29903, opened 2023-01-30
Support decoding empty DataBuffers for Decoders that support it

### LINKED ISSUE #29898: Messaging: PayloadMethodArgumentResolver should call certain Decoders even if the message is empty
Affects: Spring-Messaging 6.0.3

I am currently implementing an application, that uses Protobuf over RSocket for communication.
In Protobuf, I can define a message sent to my application like this:

```
message ProductsRequest {
  optional string id = 1;
}
```

My MessageHandler is then defined like this:
```java
@MessageMapping("products")
public Mono<ProductsResponse> handle(ProductsRequest request) {
    // implementaion
}
```

The request may contain an id, but it may be omitted.
In case the id is omitted, the message is encoded as an empty buffer (as per Protobuf spec).

However, in the class `org.springframework.messaging.handler.annotation.reactive.PayloadMethodArgumentResolver`, empty buffers are never passed to the decoder, and instead immediately filtered out. (via `.filter(this::nonEmptyDataBuffer)`, which then results in an error because the MessageHandler's argument cannot be resolved.

I understand that this behavior is reasonable for a lot of decoders, however in the case of Protobuf is a nuisance.

I think the best way to implement this is to add a property to the decoder ("can decode empty") and then filter or not, based on this property. I would gladly spend some time on this and come up with a pull request, but I wanted to have some guidance first, to not sent a crappy pull request. 

So, roughly, how should this be implemented?

spheenik:
Some decoders support decoding a proper payload from empty DataBuffers. One good example is the ProtobufDecoder.

This patch adds a method canDecodeEmptyDataBuffer() to the Decoder interface, allowing the Decoder to signal that it is able to do that.

In this case, the PayloadMethodArgumentResolver should pass the empty DataBuffer to the Decoder, and in all other cases behave like before.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/codec/ByteArrayDecoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ByteArrayDecoder.java
@@ -45,7 +45,7 @@ public class ByteArrayDecoder extends AbstractDataBufferDecoder<byte[]> {
 	}
 
 	@Override
-	public boolean canDecodeEmptyDataBuffer() {
+	public boolean canDecodeEmptyMessage() {
 		return true;
 	}
 
--- a/spring-core/src/main/java/org/springframework/core/codec/ByteArrayDecoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ByteArrayDecoder.java
@@ -45,6 +45,11 @@ public class ByteArrayDecoder extends AbstractDataBufferDecoder<byte[]> {
 	}
 
 	@Override
+	public boolean canDecodeEmptyDataBuffer() {
+		return true;
+	}
+
+	@Override
 	public byte[] decode(DataBuffer dataBuffer, ResolvableType elementType,
 			@Nullable MimeType mimeType, @Nullable Map<String, Object> hints) {
 
--- a/spring-core/src/main/java/org/springframework/core/codec/ByteBufferDecoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ByteBufferDecoder.java
@@ -48,6 +48,11 @@ public class ByteBufferDecoder extends AbstractDataBufferDecoder<ByteBuffer> {
 	}
 
 	@Override
+	public boolean canDecodeEmptyDataBuffer() {
+		return true;
+	}
+
+	@Override
 	public ByteBuffer decode(DataBuffer dataBuffer, ResolvableType elementType,
 			@Nullable MimeType mimeType, @Nullable Map<String, Object> hints) {
 
--- a/spring-core/src/main/java/org/springframework/core/codec/ByteBufferDecoder.java
+++ b/spring-core/src/main/java/org/springframework/core/codec/ByteBufferDecoder.java
@@ -48,7 +48,7 @@ public class ByteBufferDecoder extends AbstractDataBufferDecoder<ByteBuffer> {
 	}
 
 	@Override
-	public boolean canDecodeEmptyDataBuffer() {
+	public boolean canDecodeEmptyMessage() {
 		return true;
 	}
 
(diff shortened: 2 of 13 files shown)


spheenik:
In response to the following issue: https://github.com/spring-projects/spring-framework/issues/29898

simonbasle (changes requested):
this sounds reasonable. Can you please at least add some small test coverage. notably there's a `ProtobufDecoderTest` that could have a `testDecodeEmptyBuffer` method for this purpose

spheenik:
I updated the remaining Decoders that are capable to decode empty buffers to also return `true` on `canDecodeEmptyDataBuffers()`.
I also added tests for the `decode()` and the `decodeToMono()` scenarios, and made sure all the marked Decoders support it properly.

rstoyanchev (commented):
Thanks for raising this issue, and submitting a PR.

A `Decoder` has 3 kinds of methods, decoding to `Flux`, `Mono`, or a value. The `Flux` and `Mono` take a `Publisher` of buffers each of which could be a chunk of data, and not necessarily a complete message (e.g. parsing HTTP request body). In that case, empty data buffers (chunks) are okay, and does not mean the aggregated message in the end will be empty too. Moreover, since `Flux` and `Mono` methods aggregate chunks internally, they also deal with empty messages, perhaps producing a value, or completing empty, or with an error.

It's only the `decode` to a value method which takes a single `DataBuffer` and where we can apply a check before calling it. For `Flux` and `Mono`, we can do the same only when each `DataBuffer` is a complete message, but at that point you're processing each `DataBuffer` individually anyway, and might as well call the `decode` to value method, which is what we do for RSocket.

This is why it's quite confusing to call the method `canDecodeEmptyDataBuffer`. It would make more sense to call it `canDecodeEmptyMessage` and document very clearly that it's referring to an empty complete message and not a chunk of data, which is what each `DataBuffer` could be, so it really depends on the context in which the `Decoder` is called.

The exact semantics should also be made clear. I'm not too sure does `true` mean the `Decoder` will produce a non-`null` value, or that it will produce `null` without raising an error, or in the case of `Flux` and `Mono` that translates to producing a value, or filtering empty values, or completing with an error?

spheenik:
Thank you for the detailed explanation of the toMono-Case, and your other remarks. Seems I definitively had a misunderstanding what it is used for.

As a consequence, I renamed the method to `canDecodeEmptyMessage`, as suggested. I also improved on it's documentation, stating that, if the decoder returns true here, the contract will be that it will never return null when decoding an empty Message. I did not mention the error case though, which of course can happen.

I was wondering whether or not to leave the test in the toMono case in, and decided to do so, and improved on it by giving it a Flux of two empty DataBuffers as an input, to make sure a supporting decoder gives only one non-null result back.

simonbasle (changes requested):
I'm concerned about the potential unforeseen side-effects of `canDecodeEmptyMessage() == true` for a lot of low-level decoders (`String`, `Resource`, buffers...). I would reduce the scope and keep most of these decoders with the default value of `false`, and just add support for the `ProtobufDecoder` for now. @rstoyanchev what do you think?

spheenik:
@simonbasle I was also thinking about this, and I can certainly see why you are worried. I came to the conclusion that as long as the decoder makes sure to always return something non null, it should be fine. All the decodeToMono-Scenarios of the ByteArray-style decoders act in this way anyway, not even looking at the flag.
For the `StringDecoder`, it even includes a proper fix, instead of the workaround in gh-26344.
Of course, this is only with my limited knowledge, so I am eagerly awaiting @rstoyanchev 's insight.

spheenik:
It's been a while, nothing happened. To make sure, this doesn't get lost, question: Is there anything more I can do?

snicoll:
@spheenik thank you for following up. We need to take the time to review it one more time to answer your questions.

PR #29918, opened 2023-02-02
Fix RFC-7807 content negotiation

### LINKED ISSUE #29588: Prefer problem detail media type independent of ordering in Accept header
While replacing a custom implementation of Problem Details by the new one that came with Spring 6 I was puzzled by how content negotiation is implemented. I'm currently using Spring Boot 3.0.0 with Spring 6.0.2.

I'm getting a Content-Type `application/problem+json` as expected when sending no accept header or `Accept: */*` or `Accept: application/problem+json, application/json`.

I also expected to receive a Problem Details JSON with Content-Type `application/problem+json` when sending `Accept: application/json`. But that might be correct - it doesn't seem to be clearly defined by [RFC 7807](https://www.rfc-editor.org/rfc/rfc7807.html).

Clearly a bug, seems that I'm still getting Content-Type `application/json` when sending `Accept: application/json, application/problem+json`. See also:

> Message converters and encoders indicate a preference for `application/problem+json` when ProblemType is serialized. This ensures `application/problem+json` is preferred when the client is flexible or has no preference.

from: https://github.com/spring-projects/spring-framework/issues/28189#issuecomment-1121378986

ueberfuhr:
This PR fixes #29588, where we discussed the intended behavior concerning content negotiation with RFC-7807 content types.
This is a solution that

- prefers RFC-7807 content types in case of returning a ProblemDetail (error response)
- excludes RFC-7807 content types in case of returning all other types (success response)

The test simply allows to add further test cases. Currently, there's no support for ProblemDetail serialized with JAXB (`application/problem+xml`), but if it is added in the future, the test adds test cases automatically.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/AbstractMessageConverterMethodProcessor.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/AbstractMessageConverterMethodProcessor.java
@@ -19,13 +19,15 @@ package org.springframework.web.servlet.mvc.method.annotation;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import jakarta.servlet.ServletRequest;
 import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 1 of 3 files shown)


ueberfuhr:
Those are the test cases (=requirements):

image

ralf-ueberfuhr-ars:
I found another issue while fixing the tests: There's one test case throwing an exception and expecting the MVC layer to create a response with an `application/problem+json` content type and a JSON object, that does not match the requirements of this content type, because it does not conform RFC 7807. (see `org.springframework.web.servlet.mvc.method.annotation.ServletAnnotationControllerHandlerMethodTests#produces(boolean)`) 

Is it reasonable to wrap the currently returned body into an RFC-7807-compliant ProblemDetails object instead?

rstoyanchev:
There is now a fix for #29588, so I'm going to close this. Thanks for the PR in any case.

Issue #23533, opened 2019-08-28
Make @Scheduled's fixedDelay work with methods returning Mono


DavidDuwaer:
### Context
Consider this method
```java
@Scheduled(fixedDelay = 1000)
void job()
{
    // some syncronous work
}
```
The `job()` method can be counted upon to never be called until the previous run of `job()` is finished.

Now consider this method
```java
@Scheduled(fixedDelay = 1000)
void job()
{
    someAsyncMethodReturningAMono()
       .subscribe()
}
```
Here, `job()` would be called 1000ms after the previous run of `job()` returns, _not_ when the the `Mono` returned by `someAsyncMethodReturningAMono()` would terminate. If the asynchronous work takes just over a bit longer than 1000ms, it may run at the same time as previous runs. This can be mitigated by using the `.block()` statement to subscribe to `someAsyncMethodReturningAMono()`, effectively making `job()` syncronous again, but it would be better to have a non-blocking answer to this.

### Proposal
To let `@Scheduled`'s `fixedDelay` property, when the annotated method returns a `Mono` or a `Flux`, start the delay count from the moment the returned `Mono` or `Flux` terminates instead of when the method returns the `Mono`, so that one can write this:
```java
@Scheduled(fixedDelay = 1000)
Mono job()
{
    return someAsyncMethodReturningAMono();
}
```
This code would resonate with WebFlux users nicely.

s50600822:
It does not make sense to me when someone wrote an async method then try to make it run in sync manner. Why couple that into the scheduling.

DavidDuwaer:
Someone has a choice to return a Mono/Flux from the method. Returning it would denote they want Spring to consider its asynchrony completion. The user can opt not to return the mono.

Furthermore I'm only proposing an API, not an implementation. The implementation could either be truly async or cause a thread to block; that is independent of my proposal.

Your argument applies the same on WebFlux asynchrony REST controller methods. Do you also think those do not make sense?

simonbasle:
Currently investigating this. There are several hurdles to overcome which I will below. For reference here is the code that would emulate `fixedDelay` and `fixedRate` in the reactive world (based on Reactor operators).

This assumes that the annotated method returns a `Publisher<T>` (e.g. a `Mono<Void>`) and that unlike in the current arrangement, that particular kind of return value _is used_ by the annotation processor.

### fixedDelay
[...]

Issue #29883, opened 2023-01-25
Instrument Scheduled-annotated methods for observability


bclozel:
We should consider whether we can safely instrument `@Scheduled`-annotated methods with the `Observation` API and report relevant metrics and traces.

edyda99:
Hi there! I noticed this issue and I'm interested in working on it, but I wanted to ask if it's okay if I take some time to get up to speed on the codebase and the Observation API before I start making changes. I'm a new contributor and I want to make sure that I understand everything before I start making modifications. Please let me know if that's okay, and I'll be sure to keep you updated on my progress as I work on the issue.

bclozel:
@edyda99 I don't think that this issue is a good first issue for contributing to Spring Framework. Maybe consider contributing on a different one?

edyda99:
Thanks for letting me know. I'll look for another issue. I'd still like to learn from your contributions on this issue, so I'll do my research and keep an eye on your pull requests. Thanks!!

skshyamal:
Is there any plan to have this issue fixes soon?

jonatan-ivanov:
@skshyamal Hope this helps until this is implemented: right now, as a workaround, you can add an `@Observerved` annotation on your `@Scheduled` method or create an `Observation` manually.

wojciechcwek:
@jonatan-ivanov could you please show an example of the workaround you described?

jonatan-ivanov:
Example with `@Observed`: https://micrometer.io/docs/observation#_using_annotations_with_observed
Example with `.observe(...)`: https://micrometer.io/docs/observation#_introduction

bclozel:
I have an initial implementation for this feature and I've scheduled it for 6.1.0-M2 as M1 is about to be released and I'd like to gather some feedback before releasing this in a milestone.

I have worked on several designs and will share some thoughts about those here.

One solution was to build infrastructure directly on the `TaskScheduler` and `TaskExecutor`. The problem is, there is not much opinion we can build there as contracts are based on `Runnable` and we have no clue about the task being run/scheduled. A more practical way of observing such tasks would be to instrument them directly and use a custom name and a custom convention:

```java
@Service
public class MyService {

    private final ObservationRegistry observationRegistry;
    private final TaskScheduler taskScheduler;

    public void indexBlogEntries() {
        Observation observation = Observation.createNotStarted("index.blogs", this.observationRegistry);
        this.taskScheduler.schedule(
                observation.wrap(() -> {
[...]

PR #29924, opened 2023-02-03
Support `@Scheduled` on Reactive methods and Kotlin suspending functions

simonbasle:
This commit adds support for `@Scheduled` annotation on reactive methods
and on Kotlin suspending functions (provided a bridge to Reactive Streams
is available at runtime).

Reactive methods are methods that return an instance of Publisher or an
instance of a class that the ReactiveAdapterRegistry can adapt to a
Publisher with deferred support.

Kotlin suspending functions are converted to Publisher as well, using
`CoroutineUtils` and the support of the `kotlinx.coroutines.reactor`
bridge.

The bean method that produces the `Publisher` is only called once, but
there can be multiple subscriptions to that instance as configured by
the annotation. The usual task-based infrastructure for synchronous code
is reused for the purpose of scheduling the subscriptions.

One special case is when a `fixedDelay` is used, as the non-blocking
nature of subscribing to a Publisher makes it harder to adhere to these
semantics. As a result, for that particular case only, the subscription
is done in a blocking fashion inside the scheduled task.

Publisher onNext events are ignored. Active subscriptions are tracked
by the processor so that long-running publishers and infinite publishers
are also supported, allowing for cancellation if the associated bean is
destroyed or if the context is stopped.

See gh-23533
Closes gh-29924

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
@@ -143,7 +143,6 @@ public class ScheduledAnnotationBeanPostProcessor
 	private final Set<Class<?>> nonAnnotatedClasses = Collections.newSetFromMap(new ConcurrentHashMap<>(64));
 
 	private final Map<Object, Set<ScheduledTask>> scheduledTasks = new IdentityHashMap<>(16);
-	private final Map<Object, Set<ScheduledAnnotationReactiveSupport.ReactiveTask>> scheduledReactiveTasks = new IdentityHashMap<>(16);
 
 
 	/**
@@ -318,15 +317,6 @@ public class ScheduledAnnotationBeanPostProcessor
 		}
 
 		this.registrar.afterPropertiesSet();
-		// Start the reactive tasks (we synchronize on the common scheduledTasks on purpose)
-		synchronized (this.scheduledTasks) {
-			for (Set<ScheduledAnnotationReactiveSupport.ReactiveTask> reactiveTasks : this.scheduledReactiveTasks.values()) {
-				for (ScheduledAnnotationReactiveSupport.ReactiveTask reactiveTask : reactiveTasks) {
-					reactiveTask.subscribe();
-				}
-			}
-		}
-
 	}
 
 	private <T> T resolveSchedulerBean(BeanFactory beanFactory, Class<T> schedulerType, boolean byName) {
@@ -397,37 +387,32 @@ public class ScheduledAnnotationBeanPostProcessor
 
 	/**
 	 * Process the given {@code @Scheduled} method declaration on the given bean,
-	 * attempting to distinguish {@link #processScheduledReactive(Scheduled, Method, Object) reactive}
-	 * method from {@link #processScheduledSync(Scheduled, Method, Object) synchronous} methods.
+	 * attempting to distinguish {@link #processScheduledAsync(Scheduled, Method, Object) reactive}
+	 * methods from {@link #processScheduledSync(Scheduled, Method, Object) synchronous} methods.
 	 * @param scheduled the {@code @Scheduled} annotation
 	 * @param method the method that the annotation has been declared on
 	 * @param bean the target bean instance
 	 * @see #processScheduledSync(Scheduled, Method, Object)
-	 * @see #processScheduledReactive(Scheduled, Method, Object)
+	 * @see #processScheduledAsync(Scheduled, Method, Object)
 	 */
 	protected void processScheduled(Scheduled scheduled, Method method, Object bean) {
 		// Is method a Kotlin suspending function? Throws if true but reactor bridge isn't on the classpath.
-		// Is method returning a Publisher instance? Throws if true but Reactor isn't on the classpath.
+		// Is method returning a reactive type? Throws if true, but it isn't a deferred Publisher type.
 		if (ScheduledAnnotationReactiveSupport.isReactive(method)) {
-			processScheduledReactive(scheduled, method, bean);
+			processScheduledAsync(scheduled, method, bean);
 			return;
 		}
 		processScheduledSync(scheduled, method, bean);
 	}
 
 	/**
-	 * Process the given {@code @Scheduled} method declaration on the given bean,
-	 * as a synchronous method. The method MUST take no arguments. Its return value
-	 * is ignored (if any) and the scheduled invocations of the method take place
-	 * using the underlying {@link TaskScheduler} infrastructure.
-	 * @param scheduled the {@code @Scheduled} annotation
-	 * @param method the method that the annotation has been declared on
-	 * @param bean the target bean instance
-	 * @see #createRunnable(Object, Method)
+	 * Parse the {@code Scheduled} annotation and schedule the provided {@code Runnable}
+	 * accordingly. The Runnable can represent either a synchronous method invocation
+	 * (see {@link #processScheduledSync(Scheduled, Method, Object)}) or an asynchronous
+	 * one (see {@link #processScheduledAsync(Scheduled, Method, Object)}).
 	 */
-	protected void processScheduledSync(Scheduled scheduled, Method method, Object bean) {
+	protected void processScheduledTask(Scheduled scheduled, Runnable runnable, Method method, Object bean) {
 		try {
-			Runnable runnable = createRunnable(bean, method);
 			boolean processedSchedule = false;
 			String errorMessage =
 					"Exactly one of the 'cron', 'fixedDelay(String)', or 'fixedRate(String)' attributes is required";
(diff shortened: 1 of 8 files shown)


sdeleuze:
We should probably review #28515 at the same time and have a common approach.

sdeleuze:
As discussed, let's implement the Coroutines support as part of this PR via an invocation of `.CoroutinesUtils#invokeSuspendingFunction(Method, Object, Object...)` when `KotlinDetector#isSuspendingFunction` detect a method as suspending, like we do for the Web support + related tests in a `KotlinXxx.kt` reusing the same logic from Java tests with suspending function instead of `Mono`, and if that make sense `Flow` instead of Flux.

sdeleuze (commented):
What if we would keep `ScheduledAnnotationReactiveSupport.isReactive(method)` and `processScheduledReactive(scheduled, method, bean)` API as they are and we just adapt the implementations to support suspending functions?

simonbasle:
> What if we would keep `ScheduledAnnotationReactiveSupport.isReactive(method)` and `processScheduledReactive(scheduled, method, bean)` API as they are and we just adapt the implementations to support suspending functions?

Done :) As stated in the latest commit, anything `Publisher` is still in `ScheduledAnnotationReactiveSupport`, but now there is only one check to perform (and the `getPublisherFor` method used by `ReactiveTask` automatically detect the correct case).

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationReactiveSupport.java
```diff
@@ -0,0 +1,209 @@
...
+			}
+			else {
+				throw new IllegalStateException("Kotlin suspending functions may only be annotated with @Scheduled"
+						+ "if Reactor and the Reactor-Coroutine bridge (kotlinx.coroutines.reactor) are present at runtime");
+			}
+		}
+		if (!publisherPresent) {
+			return false;
+		}
+		if (!Publisher.class.isAssignableFrom(method.getReturnType())) {
```
sdeleuze, line 81: May trigger a NPE if `Publisher` is not present in the classpath I think.

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationReactiveSupport.java
```diff
@@ -0,0 +1,209 @@
...
+	 * these criteria then this check returns {@code false}.
+	 * <p>For reactive scheduling, Reactor MUST be present at runtime. In the case of
+	 * Kotlin, the Coroutine {@code kotlinx.coroutines.reactor} bridge MUST also be
+	 * present at runtime (in order to invoke suspending functions as a {@code Mono}).
+	 * Provided that is the case, this method returns {@code true}. Otherwise, it throws
+	 * an {@code IllegalStateException}.
+	 * @throws IllegalStateException if the method is reactive but Reactor and/or the
+	 * Kotlin coroutines bridge are not present at runtime
+	 */
+	static boolean isReactive(Method method) {
```
sdeleuze, line 68: Couldn't we use `ReactiveAdapterRegistry#getSharedInstance` to support various other non `Publisher` reactive types?

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
```diff
@@ -386,14 +397,24 @@ public Object postProcessAfterInitialization(Object bean, String beanName) {
 		return bean;
 	}
```
sdeleuze, line 399: Maybe copy the `processScheduled` original Javadoc here?

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
```diff
@@ -386,14 +397,24 @@ public Object postProcessAfterInitialization(Object bean, String beanName) {
...
+		// Is method returning a Publisher instance? Throws if true but Reactor isn't on the classpath.
+		if (ScheduledAnnotationReactiveSupport.isReactive(method)) {
+			processScheduledReactive(scheduled, method, bean);
+			return;
+		}
+		processScheduledSync(scheduled, method, bean);
+	}
+
 	/**
 	 * Process the given {@code @Scheduled} method declaration on the given bean.
```
sdeleuze, line 411: Maybe adapt the Javadoc to mention this is the sync variant?

sdeleuze (changes requested):

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationReactiveSupport.java
simonbasle: I was under the impression that the canonical `ReactiveAdapterRegistry` isn't necessarily the `getSharedInstance` one but a bean one. When I started on this PR I initially tried to use a `ReactiveAdapterRegistry` from the context, but at the time the processing ran it wasn't available (which makes it difficult to detect arbitrary "reactive-convertible" types).

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationReactiveSupport.java
simonbasle: the `if (!publisherPresent)` check above should prevent that part of the code to be reached and thus avoid a class-loading exception

spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationReactiveSupport.java
sdeleuze: Given the fact that this feature needs to be implemented at `BeanPostProcessor` level and the fact that `ReactiveAdapterRegistry#getSharedInstance()` is pretty useful and flexible out of the box via its classpath detection mechanism, I think for this use case it is reasonable to have the limitation that it will not be customizable from the application context. Maybe worth documenting both at Javadoc and reference documentation level.

sdeleuze (approved):
Looks good, I just mentioned 2 possibles syles improvements in tests. Nice PR!

bclozel:
Thanks for tackling this - it's definitely the right time to address this as we'll work on observability in #29883.

rstoyanchev (commented):
It's good to see this planned. 

My initial comment is to discuss the choice to use Reactor for scheduling. It means a more limited usage style, excluding cron expressions. Also, limits the usefulness of `ReactorAdapterRegistry` to a degree by requiring Reactor for the scheduling. I'm also wondering if `ReactiveTask` is going to cause classpath issues for sync methods but that's a more secondary, implementation concern.

Could the scheduling infrastructure used for sync also be used for reactive methods too, with some additional logic to subscribe to the Publisher every time a task is scheduled?

simonbasle:
I have attempted to take that approach which reuses a maximum of the existing infrastructure by using `Runnable` which subscribe to the Publishers. In a second step, I've attempted to reintroduce some tracking of active subscriptions. That way, `Publisher` subscriptions that don't terminate before the originating bean is destroyed will have a chance to be cancelled  (including infinite publishers).

simonbasle:
I will now rebase on top of `main` to take the Antora doc changes into account, and will update the PR body to make it a relevant commit message for the current state of the PR.

bclozel:
I'm currently working on the observability support of scheduled tasks in #29883. It looks like the current proposal for reactive methods support has a difference of behavior. Let's use two variants of scheduled methods, one blocking and the other reactive, both throwing exceptions during their processing:

```java
@Component
public class ScheduledComponent {

    private static final Logger logger = LoggerFactory.getLogger(ScheduledComponent.class);

    @Scheduled(cron = "0,10,20,30,40,50 * * * * *")
    public void blocking() {
        logger.info("Executing 'blocking' @Scheduled method");
        throw new IllegalStateException("Blocking method failed");
    }

    @Scheduled(cron = "2,12,22,32,42,52 * * * * *")
    public Mono<Void> reactive() {
        return Mono.error(() -> new IllegalStateException("Reactive method failed"))
                .doOnError(exc -> logger.info("Executing 'reactive' @Scheduled method"))
                .then();
    }
}
```

The blocking variant throws the exception from the generated runnable, and it's caught by the `org.springframework.util.ErrorHandler` infrastructure (this can be customized on `TaskScheduler` implementations):

```
2023-06-05T10:44:30.004+02:00  INFO 17840 --- [   scheduling-1] c.example.scheduling.ScheduledComponent  : Executing 'blocking' @Scheduled method
2023-06-05T10:44:30.004+02:00 ERROR 17840 --- [   scheduling-1] o.s.s.s.TaskUtils$LoggingErrorHandler    : Unexpected error occurred in scheduled task

java.lang.IllegalStateException: Blocking method failed
	at com.example.scheduling.ScheduledComponent.blocking(ScheduledComponent.java:17) ~[main/:na]
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:na]
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77) ~[na:na]
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:na]
	at java.base/java.lang.reflect.Method.invoke(Method.java:568) ~[na:na]
	at org.springframework.scheduling.support.ScheduledMethodRunnable.run(ScheduledMethodRunnable.java:84) ~[spring-context-6.1.0-SNAPSHOT.jar:6.1.0-SNAPSHOT]
	at org.springframework.scheduling.support.DelegatingErrorHandlingRunnable.run(DelegatingErrorHandlingRunnable.java:54) ~[spring-context-6.1.0-SNAPSHOT.jar:6.1.0-SNAPSHOT]
	at org.springframework.scheduling.concurrent.ReschedulingRunnable.run(ReschedulingRunnable.java:96) ~[spring-context-6.1.0-SNAPSHOT.jar:6.1.0-SNAPSHOT]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) ~[na:na]
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[na:na]
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:833) ~[na:na]
```

With the current proposal for reactive support, exceptions are never thrown and the error handling contract is not used:

```
2023-06-05T10:44:32.019+02:00  INFO 17840 --- [   scheduling-1] c.example.scheduling.ScheduledComponent  : Executing 'reactive' @Scheduled method
2023-06-05T10:44:32.023+02:00  WARN 17840 --- [   scheduling-1] s.s.a.ScheduledAnnotationReactiveSupport : Unexpected error occurred in scheduled reactive task

java.lang.IllegalStateException: Reactive method failed
	at com.example.scheduling.ScheduledComponent.lambda$reactive$0(ScheduledComponent.java:22) ~[main/:na]
	*__checkpoint ⇢ @Scheduled 'reactive()' in bean 'com.example.scheduling.ScheduledComponent'
Original Stack Trace:
		at com.example.scheduling.ScheduledComponent.lambda$reactive$0(ScheduledComponent.java:22) ~[main/:na]
		at reactor.core.publisher.MonoErrorSupplied.subscribe(MonoErrorSupplied.java:55) ~[reactor-core-3.5.6.jar:3.5.6]
		at reactor.core.publisher.Flux.subscribe(Flux.java:8671) ~[reactor-core-3.5.6.jar:3.5.6]
		at org.springframework.scheduling.annotation.ScheduledAnnotationReactiveSupport$SubscribingRunnable.run(ScheduledAnnotationReactiveSupport.java:194) ~[spring-context-6.1.0-SNAPSHOT.jar:6.1.0-SNAPSHOT]
		at org.springframework.scheduling.support.DelegatingErrorHandlingRunnable.run(DelegatingErrorHandlingRunnable.java:54) ~[spring-context-6.1.0-SNAPSHOT.jar:6.1.0-SNAPSHOT]
		at org.springframework.scheduling.concurrent.ReschedulingRunnable.run(ReschedulingRunnable.java:96) ~[spring-context-6.1.0-SNAPSHOT.jar:6.1.0-SNAPSHOT]
		at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) ~[na:na]
		at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[na:na]
		at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) ~[na:na]
		at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
		at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
		at java.base/java.lang.Thread.run(Thread.java:833) ~[na:na]

2023-06-05T10:44:32.027+02:00 ERROR 17840 --- [   scheduling-1] reactor.core.publisher.Operators         : Operator called default onErrorDropped
```

I don't see this as a limitation, but really as a key difference between imperative and async in general. This behavior is already documented in this PR, but maybe we can mention that the `ErrorHandler` will not be invoked. From an observability perspective, I think we can probably still record failures with the observations.

bclozel:
Closed with 35052f2113274be

PR #29932, opened 2023-02-05
Add an option to disable LazyConnectionDataSourceProxy eager checkDefaultConnectionProperties call

### LINKED ISSUE #29931: Add an option to disable LazyConnectionDataSourceProxy eager checkDefaultConnectionProperties call
**Affects:** 6.0.2

On `afterPropertiesSet`, `LazyConnectionDataSourceProxy` opens a connection to determine `defaultAutoCommit` and `defaultTransactionIsolation`. In case of exception, the latter is trapped, leading `LazyConnectionDataSourceProxy` to determine those properties when the connection is effectively requested by a consumer.

We use TestContainer to test our database interactions. The database startup takes a few seconds. To reduce test startup time, we start the database in a separate thread while the Spring context is initializing. `LazyConnectionDataSourceProxy#checkDefaultConnectionProperties` is the first consumer that makes the Spring Context initialization wait on the database startup thread.

It would be nice to have an option allowing to disable this `eager check`.

reda-alaoui:
Fix #29931

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/LazyConnectionDataSourceProxy.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/LazyConnectionDataSourceProxy.java
@@ -75,6 +75,7 @@ import org.springframework.lang.Nullable;
  * to retrieve the native JDBC Connection.
  *
  * @author Juergen Hoeller
+ * @author Réda Housni Alaoui
  * @since 1.1.4
  * @see DataSourceTransactionManager
  */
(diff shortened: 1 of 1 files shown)


reda-alaoui:
:wave:

jhoeller:
After some back and forth, I ended up dropping the eager connection properties check completely, in favor of a late check on the first `getConnection()` call (but before exposing the Connection proxy, so still exposing consistent behavior for the consumer of the Connection proxy). As a consequence, there is no need for a configuration flag. If a startup-time check should be enforced (which we don't actually recommend these days), the new public `checkDefaultConnectionProperties()` method can be called. Even better would be to explicitly specify the `defaultAutoCommit` and `defaultTransactionIsolation` configuration settings.

I'll use the original ticket #29931 for that revision.

PR #29937, opened 2023-02-07
 Update documentation according to Testcontainers example

### LINKED ISSUE #29936: Update documentation according to testcontainer example
According to the testcontainer example the container should be started first. Otherwise it could lead to container port mapping problems.
See: https://github.com/testcontainers/testcontainers-java/blob/main/examples/spring-boot/src/test/java/com/example/AbstractIntegrationTest.java

samuelstein:
Update documentation according to testcontainer example. Fix #29936

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/DynamicPropertySource.java
+++ b/spring-test/src/main/java/org/springframework/test/context/DynamicPropertySource.java
@@ -76,6 +76,7 @@ import java.lang.annotation.Target;
  *
  *     &#064;DynamicPropertySource
  *     static void redisProperties(DynamicPropertyRegistry registry) {
+ *         redis.start();
  *         registry.add("redis.host", redis::getHost);
  *         registry.add("redis.port", redis::getMappedPort);
  *     }


sbrannen:
The container in the example does not need to be manually started since the JUnit Jupiter extension behind the `@Testcontainers` annotation handles that for you. In light of that, I am closing this PR.

However, this PR did inspire me to review the examples which resulted in #29940.

Thanks,

Sam

samuelstein:
The thing is, that if you don't start the container explicitly in the DynamicPropertySource hook it's possible that you run in an timing issue when the hook will be called earlier than the container is started by the JUnit extension and this will result in an "Mapped port can only be obtained after the container is started" error.
The start method of the container will check anyway if the container is already started.
So calling the start method explicit will only make the process more robust and has no negative consequences i would say.
What do you think?

sbrannen:
@eddumelendez, what are your thoughts on the matter?

eddumelendez:
Hi, the javadoc is using `Testcontainers` and `Container` annotations. Those are part of the junit-jupiter integration and the container will start without calling `start()` method. So, the change doesn't apply.

sbrannen:
Thanks for confirming that, @eddumelendez.

PR #29946, opened 2023-02-08
Extension Base64Util utility class

Karosown:
Extension Base64Util utility class


sdeleuze:
Thanks for your contribution, but `Base64Util` is deprecated, so I don't think it makes sense to add a new method to it, that's why I will close this PR unmerged.

PR #29953, opened 2023-02-09
Allow MockRest to match header/queryParam value list with one Matcher

### LINKED ISSUE #28660: Add support in `MockRestServiceServer` to assert ALL `header` and `queryParam` values with a single `Matcher`
**Affects:** 5.3.21

---

The current API for `MockRestServiceServer` and `MockRestRequestMatchers` does not allow a test to validate that all values of a certain header key satisfy a matcher. Trying to do

```java
.andExpect(header("Accept", endsWith("json")))
```

does not work because it passes if the first value for the header ends with `"json"`. I would like to be able to do something like this:

```java
.andExpect(header("Accept", allMatch(endsWith("json"))))
```

or something to that effect.

simonbasle:
This commit adds a `header` variant and a `queryParam` variant to the
`MockRestRequestMatchers` API which take a single `Matcher` over the
list of values.

Contrary to the vararg variants, the whole list is evaluated and the
caller can choose the desired semantics using readily-available iterable
matchers like `everyItem`, `hasItems`, `hasSize`, `contains` or
`containsInAnyOrder`...

The fact that the previous variants don't strictly check the size of the
actual list == the number of provided matchers or expected values is
now documented in their respective javadocs.

Closes gh-28660

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
@@ -117,8 +117,8 @@ public abstract class MockRestRequestMatchers {
 	 * Assert request query parameter values with the given Hamcrest matcher(s).
 	 * <p> Note that if the queryParam value list is larger than the number of provided
 	 * {@code matchers}, extra values are considered acceptable.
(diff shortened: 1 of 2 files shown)


rstoyanchev (approved):
Looks good, just minor formatting comment. For Javadoc we [wrap around 90](https://github.com/spring-projects/spring-framework/wiki/Code-Style#line-wrapping), or even 80.

PR #29955, opened 2023-02-10
Add RFC-8246 "immutable" attribute to `CacheControl`

making:
This commit adds the "[immutable](https://tools.ietf.org/html/rfc8246)" Cache-Control directives in `CacheControl`.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/CacheControl.java
+++ b/spring-web/src/main/java/org/springframework/http/CacheControl.java
@@ -77,6 +77,8 @@ public class CacheControl {
 	@Nullable
 	private Duration sMaxAge;
 
+	private boolean immutable = false;
+
 
 	/**
 	 * Create an empty CacheControl instance.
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the PR @making !
The `CacheControl` API tries to guide developers towards HTTP best practices for "Cache-Control" header values, with static factory methods and instance methods.

In this case, the popular use case uses both a max-age and the immutable directive, like so: `"Cache-Control: max-age=31536000, immutable"`. The goal here is to instruct the browser that the response should not be revalidated as long as it's not stale, giving a long "max-age" value.

We could have enforced this by requiring a "max-age" value if the freshness directive is used. This is not technically enforced by the RFC. Without such a value, browsers are likely to fall back to [heuristics caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_caching), effectively caching the resource for `(current date - last modified date) / 10`.

[At the time of this comment, numerous browsers aren't supporting this directive](https://caniuse.com/mdn-http_headers_cache-control_immutable): Edge, Chrome, Android browsers, Opera. Because of that, most implementations should use a `max-age` value in all cases for compatibility.

We won't enforce the need for `max-age`, but we should probably guide developers there with the Javadoc.

PR #29958, opened 2023-02-11
Propagate WebClient attributes into underlying HTTP client request where possible

### LINKED ISSUE #26208: Passing attributes from `WebClient` to underlying HTTP library
Hello,

We are trying to enable communication logging (low level HTTP logging) for requests and responses processed with `WebClient`. There is a good tutorial on how to approach the simplest case which I've followed to get the initial results - [Link to Baeldung](https://www.baeldung.com/spring-log-webclient-calls#1-logging-with-jetty-httpclient).

However, HTTP logging in our company is a bit more complex simply just calling SLF4J. In certain cases we need to provide extra context paired to request/response, so that other tooling can correctly process those logs (some internal IDs). We've identified the `attributes` feature as potentially the cleanest way to pass additional context to the underlying library, seeing how they are present on `WebClient` API as well as both Netty and Jetty libraries. However, it seems that the attributes are not copied from `WebClient` model to the Netty/Jetty model.

As a workaround, we are currently wrapping the Request/Response spec to add additional behavior on `WebClient` side.

So the question is - are attributes the right way of solving cases like this and how would one go about adding support for copying attributes from `WebClient` request to underlying library request?

I've found no relevant discussions and this seems to be a useful enhancement.

Crustack:
This PR Closes #26208 
As proposed by @tstavinoha (https://github.com/spring-projects/spring-framework/issues/26208#issuecomment-739817950) and @rstoyanchev (https://github.com/spring-projects/spring-framework/issues/26208#issuecomment-783324482), this adds the option `applyAttributes` to the `WebClient.Builder`. If it is set to `true`, the `ClientRequest.attributes()` will be copied over to the underlying http-client library HttpRequest. For this there are different implementations for the various `ClientHttpConnector` implementations:
* `JettyClientHttpRequest`: Attributes are set with the `org.eclipse.jetty.client.api.Request#attribute(String, Object)` method one by one
* `ReactorClientHttpRequest`: A single `Map<String, Object>` containing all request attributes is set to the Attribute `attributes` in the `reactor.netty.channel.ChannelOperations#channel()` (as proposed by https://github.com/spring-projects/spring-framework/issues/26208#issuecomment-783324482)
* `ReactorNetty2ClientHttpRequest`: Same as in `ReactorClientHttpRequest`, but with the `netty5` classes
* `HttpComponentsClientHttpRequest`: Attributes are set with the `org.apache.hc.client5.http.protocol.HttpClientContext#setAttribute(String, Object)` method one by one
* `JdkClientHttpRequest`: I still need some input on this one. **As far as I can see there is no possibility** to set request attributes to the `java.net.http.HttpRequest`? 

As suggested by https://github.com/spring-projects/spring-framework/issues/26208#issuecomment-739817950, `applyAttributes` defaults to `true` and the user can opt-out when using the Builder to create the Connector.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
@@ -114,7 +114,7 @@ class JdkClientHttpRequest extends AbstractClientHttpRequest {
 	}
 
 	/**
-	 * Not implemented, since {@link HttpRequest} does not offer any possibility to add request attributes
+	 * Not implemented, since {@link HttpRequest} does not offer any possibility to add request attributes.
 	 */
 	@Override
 	protected void applyAttributes() {
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
@@ -56,7 +56,8 @@ class JdkClientHttpRequest extends AbstractClientHttpRequest {
 	private final HttpRequest.Builder builder;
 
 
-	public JdkClientHttpRequest(HttpMethod httpMethod, URI uri, DataBufferFactory bufferFactory) {
+	public JdkClientHttpRequest(HttpMethod httpMethod, URI uri, DataBufferFactory bufferFactory, boolean applyAttributes) {
+		super(applyAttributes);
 		Assert.notNull(httpMethod, "HttpMethod is required");
 		Assert.notNull(uri, "URI is required");
 		Assert.notNull(bufferFactory, "DataBufferFactory is required");
@@ -110,6 +111,15 @@ class JdkClientHttpRequest extends AbstractClientHttpRequest {
 	protected void applyCookies() {
 		this.builder.header(HttpHeaders.COOKIE, getCookies().values().stream()
 				.flatMap(List::stream).map(HttpCookie::toString).collect(Collectors.joining(";")));
+	}
+
+	/**
+	 * Not implemented, since {@link HttpRequest} does not offer any possibility to add request attributes
+	 */
+	@Override
+	protected void applyAttributes() {
+		// TODO
+		throw new RuntimeException(String.format("Using attributes is not available for %s", HttpRequest.class.getName()));
 	}
 
 	@Override
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpRequest.java
@@ -19,13 +19,17 @@ package org.springframework.http.client.reactive;
 import java.net.URI;
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.Map;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.handler.codec.http.cookie.DefaultCookie;
+import io.netty.util.AttributeKey;
+import org.eclipse.jetty.client.api.Request;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.netty.NettyOutbound;
+import reactor.netty.channel.ChannelOperations;
 import reactor.netty.http.client.HttpClientRequest;
 
 import org.springframework.core.io.buffer.DataBuffer;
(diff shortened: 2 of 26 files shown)


spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
```diff
@@ -112,6 +113,15 @@ protected void applyCookies() {
 				.flatMap(List::stream).map(HttpCookie::toString).collect(Collectors.joining(";")));
 	}
 
+	/**
+	 * Not implemented, since {@link HttpRequest} does not offer any possibility to add request attributes
+	 */
+	@Override
+	protected void applyAttributes() {
+		// TODO
+		throw new RuntimeException(String.format("Using attributes is not available for %s", HttpRequest.class.getName()));
```
PhilKes, line 122: Can somebody please verify if there is an option to set request attributes for the `java.net.http.HttpRequest`?
If there is not, what should be the proper behaviour, should an Exception be thrown to the user if `applyAttributes` is set to `true`? Or should it simply always default to `false`?

spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientIntegrationTests.java
```diff
@@ -187,6 +187,64 @@ void retrieveJson(ClientHttpConnector connector) {
...
+				assertThat(attributes.get()).isNull();
+			}
+		} else if (nativeRequest.get() instanceof Request nativeReq) {
+			if (expectAttributesApplied) {
+				assertThat(nativeReq.getAttributes()).containsEntry("foo", "bar");
+			} else {
+				assertThat(nativeReq.getAttributes()).doesNotContainEntry("foo", "bar");
+			}
+		} else if (nativeRequest.get() instanceof org.apache.hc.core5.http.HttpRequest nativeReq) {
+			// TODO get attributes from HttpClientContext
```
PhilKes, line 242: I could not find out how to verify the request attributes were added to the `org.apache.hc.client5.http.protocol.HttpClientContext` in the Test here, can somebody help out with this?

spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClientBuilder.java
```diff
@@ -335,16 +344,18 @@ public WebClient build() {
...
+			return new ReactorNetty2ClientHttpConnector(this.applyAttributes);
 		}
 		else if (jettyClientPresent) {
-			return new JettyClientHttpConnector();
+			return new JettyClientHttpConnector(this.applyAttributes);
 		}
 		else if (httpComponentsClientPresent) {
-			return new HttpComponentsClientHttpConnector();
+			HttpComponentsClientHttpConnector httpComponentsClientHttpConnector = new HttpComponentsClientHttpConnector();
+			httpComponentsClientHttpConnector.setApplyAttributes(this.applyAttributes);
```
PhilKes, line 357: For reasons unknow to me I was not able to call the new constructor with `applyAttributes` parameter. The compile Task threw an error saying:
```
 Task :spring-webflux:compileJava
/tmp/build/f663b219/git-repo/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClientBuilder.java:356: error: cannot access CloseableHttpAsyncClient
			return new HttpComponentsClientHttpConnector(applyAttributes);
			       ^
  class file for org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient not found
``` 
(see [pipelines/builds/1101](https://ci.spring.io/teams/spring-framework/pipelines/spring-framework-6.0.x/jobs/build-pull-requests/builds/1101)).
Therefore I used the default constructor and the setter.

PhilKes:
@rstoyanchev would love some feedback if you have the time.

spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpRequest.java
```diff
@@ -45,6 +48,8 @@
  */
 class ReactorClientHttpRequest extends AbstractClientHttpRequest implements ZeroCopyHttpOutputMessage {
 
+	public static final String ATTRIBUTES_CHANNEL_KEY = "attributes";
```
rstoyanchev, line 51: The attribute name should be a little more qualified. Typically we would use something like: 
```java
ReactorClientHttpRequest.class.getName() + ".attributes"
```

spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
rstoyanchev: If the boolean flag is removed as suggested above, this should become a no-op.

rstoyanchev (changes requested):
Thanks for the pull request!

On second thought I don't think we need to make this opt-in and can drop the config option from `WebClient.Builder` and from `ClientHttpConnector`s. It does make sense to make attributes go further where we can, and treat them as any other part of the request. For Netty we are also adding the full map as a single attribute, so not likely to be a performance issue.

Issue #23551, opened 2019-08-30
Entry set of read-only HttpHeaders loses original headers' ordering


wilkinsona:
**Affects:** 5.1.9.RELEASE

The entry set returned from the result of calling `HttpHeaders.readOnlyHttpHeaders(HttpHeaders)` does not maintain the original headers' ordering. This is a regression from 5.1.0 that I think was introduced in 5.1.1 in https://github.com/spring-projects/spring-framework/commit/ce7278aaf4f20348862267c2081c20dc5bd77128. I believe the set is unordered due to the use of `Collectors.toSet()` which creats an unordered `Set`:

https://github.com/spring-projects/spring-framework/blob/1ea6ce72bb3b1139abc370f6dc32c51fc27ae90b/spring-web/src/main/java/org/springframework/http/ReadOnlyHttpHeaders.java#L146

sbrannen:
Thanks for bringing this to our attention.

Your analysis is spot on, and the fix is as easy as `.collect(Collectors.toCollection(LinkedHashSet::new)))`.

I'll push that to `5.1.x` and `master` soon.

PR #29972, opened 2023-02-15
Optimize MultiValueMap iteration operations

yuzawa-san:
I was doing some flame graph analysis and found that ReadOnlyHttpHeaders.entrySet() was a small hotspot in my codebase. I looked into its implementation and found that some care was taken in the past to fully copy over the entrySet to into a new and [properly ordered](https://github.com/spring-projects/spring-framework/issues/23551) set. The main hotspot in my specific codebase was in BodyInserterRequest.writeTo() which was called every time I used the reactive WebClient. I found that the entrySet() could be avoided and replaced with a forEach with putIfAbsent. The putIfAbsent is particularly useful since in the primary implementation I found it has to hash/normalize keys only once instead of twice like it was doing currently:
https://github.com/spring-projects/spring-framework/blob/b23316302d578694c6915d5fc9ef24b571110ffc/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java#L209-L223
In order to make sure the fast path is taken, I had to make sure the whole call chain avoided the Map class' default implementations (which use the slow entrySet() or a naive putIfAbsent). this allowed us to get to LinkedCaseInsensitiveMap's implementations. The forEach is optimized in LinkedHashMap as well. This foregoes the need to create entrySets and iterators (all with immutability) which saves memory allocations. Here is a stack trace from a breakpoint in LinkedCaseInsensitiveMap.putIfAbsent() (note how the fast path is taken):
image

Some other minor optimizations were included: presizing list in MultiValueMapAdapter.addAll,  use forEach in implementation of MultiValueMapAdapter.addAll (this avoids that slow entrySet path)

ASIDE: I think I found a bug in the ReadOnlyHttpHeaders.entrySet(). It appears that it leaks a mutable value list. See the following code:

```java
HttpHeaders src = new HttpHeaders();
src.set("foo", "bar");
src = HttpHeaders.readOnlyHttpHeaders(src);
src.entrySet().iterator().next().getValue().add("blah") // this should throw since ["bar"] is should be unmodifiable.
assertThat(src.get("foo").size()).isEqualTo(2); // should really be 1?
```
That code currently works, but should it? I have a fix, but I'll save that for a separate PR if you want.

I can also look into adding forEach implementations in things like NettyHeadersAdapter later.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
+++ b/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
@@ -27,6 +27,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.Spliterator;
+import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -284,6 +285,11 @@ public class LinkedCaseInsensitiveMap<V> implements Map<String, V>, Serializable
 			this.entrySet = entrySet;
 		}
 		return entrySet;
+	}
+
+	@Override
+	public void forEach(BiConsumer<? super String, ? super V> action) {
+		this.targetMap.forEach(action);
 	}
 
 	@Override
--- a/spring-core/src/main/java/org/springframework/util/MultiValueMapAdapter.java
+++ b/spring-core/src/main/java/org/springframework/util/MultiValueMapAdapter.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiConsumer;
 
 import org.springframework.lang.Nullable;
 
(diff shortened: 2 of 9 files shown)


yuzawa-san:
Here are the results of [my stress test](https://gist.github.com/yuzawa-san/5102be26fb9e4472419c23b39c4524b2). This is a portion of BodyInserterRequest.writeTo(), but the savings can be realized all over the place: DefaultRequestBodyUriSpec.init*(), AbstractServerResponse.copy(), ReactorClientHttpRequest.apply*(), and everywhere else in user code where putAll or addAll is used on the headers.

before memory:
image


after memory:
image


before cpu:
image

after cpu:
image

yuzawa-san:
@rstoyanchev would it be possible to triage this? let me know if anything else is needed.

bclozel:
Thanks for this PR @yuzawa-san and sorry about the delay.

PR #29980, opened 2023-02-15
reactive ControllerMethodResolver - introduce InvocableHandlerMethod cache

### LINKED ISSUE #29951: Introduce global cache for reactive argument resolvers
Hi team.
It seems that in the current flow of invoking a RequestMapping endpoint,
`org.springframework.web.reactive.result.method.annotation.RequestMappingHandlerAdapter#handle` is creating a new `org.springframework.web.reactive.result.method.InvocableHandlerMethod` for each new request (might be worth caching?), which then creates a new `org.springframework.web.reactive.result.method.HandlerMethodArgumentResolverComposite` that seems to cache the argument resolvers, but is only valid for this single invocation.

My suggestion is to make the `HandlerMethodArgumentResolver#argumentResolverCache` inside `HandlerMethodArgumentResolverComposite` static to allow caching the resolvers between invocations, as the `supportsParameter()` seems like a pure function, as well as most implementations are using reflections to get the methods/arguments annotations which is not cheap.

Thanks

ohadgur:
Currently, a new `InvocableHandlerMethod` is created on each controller request, and in turn, creates a new `HandlerMethodArgumentResolverComposite` and new `argumentResolverCache` which makes each request calculate the argument resolver of this method.

This PR changes this behavior and caches the `InvocableHandlerMethod` by its `HandlerMethod`, which should reduce the number of resolvers lookups, which shouldn't change between invocations.

Closes https://github.com/spring-projects/spring-framework/issues/29951


ohadgur:
Not sure why `org.springframework.web.reactive.function.MultipartRouterFunctionIntegrationTests.proxy` failed because of it, and I can't reproduce locally..

ohadgur:
hi @rstoyanchev , I just merged from main to re-run the build, can you please check?
thanks

rstoyanchev (commented):
I'm afraid we can't do this since a controller may be a prototype bean, and hence the `bean` in `HandlerMethod`, a different instance on each request.

rstoyanchev:
To elaborate further, we would not want to re-use a cached `HandlerMethod` when the bean (controller) instance is not the same. The cache is based on `HandlerMethod#equals`, which calls `bean#equals`, and if the controller implemented it, it wouldn't be a same instance check, which is what we want. We could create a cache key specific for this, but then if prototype beans are used, it would make the cache grow. 

In short, solving this would require more extensive changes, but I'm not sure there is sufficient gain to justify it.

ohadgur:
@rstoyanchev thanks for taking the time to explain. The reason I started this issue is that this code path (especially the argument resolving) comes up in our profiling with  ~2-3% of the total CPU usage, which isn't negligible. Attaching a screenshot of the flamegraph.
image
Do you think a cache key such as (bean type, method/method name), be good enough?

PR #29985, opened 2023-02-16
Introduce caching for functional routing

JakobFels:
Currently the functional API does not support any kind of caching for resources. I was trying to replicate an nginx setup with multiple virtual hosts, each having their own assets folder.

So my code would contain something like:
```java
route.nest(
  RequestPredicates.headers(headers -> "www.exmple.com".equals(headers.host().getHostString())),
  builder -> builder
    .add(ResourcesHandler.resources("/**", getHostSpecificResourceDir())
```
With this setup, I saw no option to define any kind of caching for these resources, because `ResourceHandlerFunction` was returning an `EntityResponse` which could not be modified.

This PR introduces caching support for the `.resources` instruction. Implementing the `ResourceCacheLookupStrategy` interface allows for either defining a general caching rule (e.g. max-age=60 for all resources), or to define custom cache behaviour per file (e.g. max-age=3600 if resource is an xml file).

The default behaviour, if no strategy is provided, is to not cache at all, preserving the status quo.

I have signed the CLA

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/DefaultResourceCacheLookupStrategy.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/DefaultResourceCacheLookupStrategy.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2002-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


JakobFels:
Just wanted to know if anything is missing to get this moving

poutsma:
Thank you for submitting a PR. 

I decided to take an alternative approach (see 307a2c7d7be1fccd97e233bec065c28cdc653730) that allows you to change any headers for server resources, not just `Cache-Control`.

PR #29988, opened 2023-02-16
Added examples for containsWhitespace and trimAllWhitespace methods

esivakumar18:
Added examples for `containsWhitespace` and `trimAllWhitespace` methods in `StringUtils` class


rstoyanchev (commented):
Same here. Thanks fro the suggestion, but I don't think the samples add any value.

PR #29990, opened 2023-02-16
Added examples for capitalize & uncapitalize methods

esivakumar18:
Added examples for `capitalize` &` uncapitalize` methods in `StringUtils` class

PR #29991, opened 2023-02-16
Added examples for startsWithIgnoreCase and endsWithIgnoreCase

esivakumar18:
Added examples for `startsWithIgnoreCase` and `endsWithIgnoreCase` in `StringUtils` class


rstoyanchev (commented):
Also here as with #29990, I think it would be better to improve the description and refer to `String#regionMatches` for a more comprehensive understanding.

PR #29994, opened 2023-02-19
Apply "instance of patten matching" in `spring-context` and `spring-messaging`

divcon:
apply "instance of patten matching".


sbrannen:

Thanks

PR #29998, opened 2023-02-20
Fix `@Aspect` `perthis` syntax in reference manual

kriegaex:
In the Spring AOP manual, the sample code falsely reads in both Java and Kotlin examples as follows:

```java
@Aspect("perthis(com.xyz.myapp.CommonPointcuts.businessService())")
```

But inside `perthis` must be a valid pointcut, not just a method name, as I explained in [this StackOverflow answer](https://stackoverflow.com/a/75506704/1082681).

Thus, the example code should be changed to something like:

```java
@Aspect("perthis(execution(* com.xyz.myapp.CommonPointcuts.businessService()))")
```

@sbrannen, maybe you are the right person to review and merge this issue.

### DIFF
--- a/framework-docs/src/docs/asciidoc/core/core-aop.adoc
+++ b/framework-docs/src/docs/asciidoc/core/core-aop.adoc
@@ -1778,7 +1778,7 @@ annotation. Consider the following example:
 [source,java,indent=0,subs="verbatim,quotes",role="primary"]
 .Java
 ----
-	@Aspect("perthis(com.xyz.myapp.CommonPointcuts.businessService())")
+	@Aspect("perthis(execution(* com.xyz.myapp.CommonPointcuts.businessService()))")
 	public class MyAspect {
 
 		private int someState;
@@ -1792,7 +1792,7 @@ annotation. Consider the following example:
 [source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
 .Kotlin
 ----
-	@Aspect("perthis(com.xyz.myapp.CommonPointcuts.businessService())")
+	@Aspect("perthis(execution(* com.xyz.myapp.CommonPointcuts.businessService()))")
 	class MyAspect {
 
 		private val someState: Int = 0


sbrannen:
Hi @kriegaex,

> In the Spring AOP manual, the sample code falsely reads in both Java and Kotlin examples as follows:
> 
> ```java
> @Aspect("perthis(com.xyz.myapp.CommonPointcuts.businessService())")
> ```

It turns out that the examples are correct as-is. `CommonPointcuts` defines shared pointcuts as explained earlier in the documentation in the [Sharing Common Pointcut Definitions](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#aop-common-pointcuts) section.

To ensure that this works as expected, I pushed some changes to the tests in commit b437b7be3463dc60e3fa49a648d85743182d6f74.

However, I can understand the confusion if one jumps to a particular example in the AOP chapter without having read the _Sharing Common Pointcut Definitions_ section. So we may possibly reconsider our use of shared pointcuts in subsequent examples or possibly cross reference the _Sharing Common Pointcut Definitions_ section.

As a side note, your PR made me realize that our `package` statements are no longer displayed for those AOP/AspectJ examples and that inspired me to create #30001.

So thanks for bringing that to my attention! 👍

kriegaex:
@sbrannen, I am sorry to have missed that. I even think, it happened for the second time within two or so years. Actually, in my own experience it is a rather rare case to refer to pointcuts defined in a class other than the current aspect, even though perfectly fine as a development practice. Unfortunately, it makes reading the manual section in question misleading. When just looking for how to use `this()` and `target()` without also having read the section about sharintg pointcuts the very same day, it is utterly non-intuitive. I am not an AOP newbie. In fact, I am the current maintainer of AspectJ and still missed that. Would you mind me modifying the PR to mention the fact that you are referring to a pointcut defined in another class here, and maybe also provide a brief example of an inline pointcut?

PR #30010, opened 2023-02-22
Use `readNBytes` in ByteArrayHttpMessageConverter when contentLength is available

kilink:
If the content length is available, pass it to readNBytes when reading the message body. When the content length is less than the internal buffer size in InputStream (8192), this avoids a copy, as readNBytes will return the buffer directly. When the content length is greater than the buffer size used in InputStream, passing the content length at least avoids over-allocating the final buffer (e.g., if the content length were 8193 bytes, 1 byte more than the default buffer size).

If the content length isn't present or is too large to represent as an integer, fall back to the default behavior of readAllBytes by passing in Integer.MAX_VALUE.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/ByteArrayHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/ByteArrayHttpMessageConverter.java
@@ -52,7 +52,15 @@ public class ByteArrayHttpMessageConverter extends AbstractHttpMessageConverter<
 
 	@Override
 	public byte[] readInternal(Class<? extends byte[]> clazz, HttpInputMessage inputMessage) throws IOException {
-		return inputMessage.getBody().readAllBytes();
+		long contentLength = inputMessage.getHeaders().getContentLength();
+		final int len;
+		if (contentLength >= 0 && contentLength <= Integer.MAX_VALUE) {
+			len = (int) contentLength;
+		}
+		else {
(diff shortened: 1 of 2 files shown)


rstoyanchev (commented):
Thanks for the suggestion. I think we could do it in the `main` branch, since `readNBytes` was added in JDK 11.

One minor comment on the implementation. Given that `readAllBytes` internally is `readNBytes(Integer.MAX_VALUE)`, couldn't the logic then be such that if there is a suitable `contentLength` call `readNBytes`, or otherwise fall back on `readAllBytes`.

spring-web/src/main/java/org/springframework/http/converter/ByteArrayHttpMessageConverter.java
```diff
@@ -52,7 +52,15 @@ public boolean supports(Class<?> clazz) {
...
-		return inputMessage.getBody().readAllBytes();
+		long contentLength = inputMessage.getHeaders().getContentLength();
+		final int len;
+		if (contentLength >= 0 && contentLength <= Integer.MAX_VALUE) {
+			len = (int) contentLength;
+		}
+		else {
+			len = Integer.MAX_VALUE;
+		}
+		return inputMessage.getBody().readNBytes(len);
```
r331, line 63: Regarding readAllBytes, we read with chunks with max len 8192 (default_buffer_size). 
Do we really get better performance with this code?

PR #30014, opened 2023-02-22
Update comment in Javadoc of ServletRequestPathFilter DispatcherServlet relating to DispatcherServlet

sergiuprdn:
The following behavior documented by `org.springframework.web.filter.ServletRequestPathFilter` does not hold true with current version:

> DispatcherServlet will parse and cache the `RequestPath` if it detects that parsed `PathPatterns` are enabled for any `HandlerMapping` but it will skip doing that if `ServletRequestPathUtils#PATH_ATTRIBUTE` already exists.

This PR adds the logic to skip parse and cache `RequestPath` if `ServletRequestPathUtils.PATH_ATTRIBUTE` already exists. 

The exception where multiple parsing must happen is for Forward requests.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/filter/ServletRequestPathFilter.java
+++ b/spring-web/src/main/java/org/springframework/web/filter/ServletRequestPathFilter.java
@@ -36,10 +36,9 @@ import org.springframework.web.util.ServletRequestPathUtils;
  * {@link org.springframework.web.util.pattern.PathPattern}s are in use anywhere
  * in an application instead of String pattern matching with
  * {@link org.springframework.util.PathMatcher}.
- * <p>Note that in Spring MVC, the {@code DispatcherServlet} will also parse and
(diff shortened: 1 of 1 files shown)


rstoyanchev (commented):
Thanks for bringing this up. You're right that the Javadoc reflects original intent rather than current behavior. 

For a FORWARD, the filter would also get involved, so technically the `DispatcherServlet` can always decide to stick with the existing one. The problem is that when there is no `Filter`, on FORWARD we would find a previous value, and the path would need to be parsed again. This is why in the end `DispatcherServlet` ended up with always doing its own parsing, while the `Filter` is there mostly if anything in the Filter chain relies on the parsed path. 

There are also other scenarios with ERROR and ASYNC dispatches, and I'm not sure it's worth doing anything more here. We might just have to correct the Javadoc of the filter accordingly.

PR #30015, opened 2023-02-22
Add ClasspathFileUtil for easy getting files from resource directory

divinenickname:
We have some tests which needed a json example for reproduction. I think it would be better if we can ability to easy load them or other files from resource directory instead of defining them into class. It'll makes code more clarity and readable.

For example `val body = """
			{
				"bytes": [
					1,
					2
				],
				"array": [
					"Foo",
					"Bar"
				],
				"number": 42,
				"string": "Foo",
				"bool": true,
				"fraction": 42
			}
		""".trimIndent()`

can be changed to ` val body = ResourceClasspathFileUtil("org/springframework/request.json").readText()`

### DIFF
--- a/spring-core/src/main/kotlin/org/springframework/util/ResourceClasspathFileUtil.kt
+++ b/spring-core/src/main/kotlin/org/springframework/util/ResourceClasspathFileUtil.kt
@@ -1,0 +1,13 @@
+package org.springframework.util
+
+import java.io.File
+
+/**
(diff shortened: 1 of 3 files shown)


bclozel:
I think this is already covered by `new ClassPathResource ("some/resource.json")`; this has been made even easier with #24651. I'm declining this PR as a result.

Thanks for your contribution!

PR #30019, opened 2023-02-23
Apply "instanceof pattern matching" in `spring-tx`

diguage:


spring-tx/src/main/java/org/springframework/transaction/jta/JtaTransactionManager.java
```diff
@@ -547,8 +547,8 @@ protected void initTransactionSynchronizationRegistry() {
 	 * @return a corresponding UserTransaction handle
 	 */
 	protected UserTransaction buildUserTransaction(TransactionManager transactionManager) {
-		if (transactionManager instanceof UserTransaction) {
-			return (UserTransaction) transactionManager;
+		if (transactionManager instanceof UserTransaction userTransaction) {
```
sbrannen, line 550: This fails the build due to a Checkstyle limitation that prevents Checkstyle from differentiating between the location `userTransaction` and `this.userTransaction`.

spring-tx/src/main/java/org/springframework/transaction/jta/JtaTransactionManager.java
```diff
@@ -702,11 +702,11 @@ protected UserTransaction findUserTransaction() {
 	 */
 	@Nullable
 	protected TransactionManager findTransactionManager(@Nullable UserTransaction ut) {
-		if (ut instanceof TransactionManager) {
+		if (ut instanceof TransactionManager transactionManager) {
```
sbrannen, line 705: This also fails the build due to the aforementioned Checkstyle limitation.

sbrannen (approved):
Thanks for the PR.

I noticed two issues that break the build.

In addition, there are several other candidates for "instance pattern matching" within the `spring-tx` module.

I have already addressed all of these issues locally, so there is no need to modify this PR.

sbrannen:
This has been merged into `main` in 375114defa711b33babc0aef9b4f918f232e315f and revised in 34e5ce9360ed6c0fe83678a904d5e70064765db6 and b1cf832c28473518240a91afed17bd51f95fbf9a.

Thanks

PR #30020, opened 2023-02-23
Add test support to record async events, with JUnit Jupiter caveat

### LINKED ISSUE #29827: `@RecordApplicationEvents` should record events from spawned threads, too
The infrastructure behind `@RecordApplicationEvents` currently uses a `ThreadLocal` to capture application events published by the current thread. Unfortunately, this rules out events published by asynchronous event listeners in turn. It would be nice if events would be captured that are published on threads spawned by the current execution thread.

simonbasle:
This commit modifies the way the `@RecordApplicationEvents` annotation
works in tests, allowing for capture of events from threads other than
the main test thread (async events) and for the assertion of captured
event from a separate thread (e.g. when using `Awaitility`).

This is done by switching the `ApplicationEventsHolder` to use an
`InheritedThreadLocal`.

There is a mutual exclusion between support of asynchronous events vs
support of JUnit5 parallel tests with the `@TestInstance(PER_CLASS)`
mode. As a result, we favor the former and now `SpringExtension` will
invalidate a test class that is annotated (or meta-annotated, or
enclosed-annotated) with `@RecordApplicationEvents` AND
`@TestInstance(PER_CLASS)` AND `@Execution(CONCURRENT)`.

Closes gh-29827

### DIFF
--- a/spring-test/spring-test.gradle
+++ b/spring-test/spring-test.gradle
@@ -76,6 +76,7 @@ dependencies {
 	}
 	testImplementation("io.projectreactor.netty:reactor-netty-http")
 	testImplementation("de.bechte.junit:junit-hierarchicalcontextrunner")
+	testImplementation("org.awaitility:awaitility")
 	testRuntimeOnly("org.junit.vintage:junit-vintage-engine") {
 		exclude group: "junit", module: "junit"
 	}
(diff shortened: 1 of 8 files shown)


simonbasle:
cc @odrotbohm

draft as this should probably be slated for 6.1.0

PR #30028, opened 2023-02-24
Improve attribute handling in RequestPredicates

yuzawa-san:
I was doing some memory allocation flame chart analysis and found that there was a lot of copying of the attributes map in the AND/OR/NOT predicates which I believe can be avoided. The intent appeared to be to support the PathPatternPredicate which is the only implementation that updates the request state. Given that the PathPatternPredicate can be embedded within any number of boolean operators the save, evaluate, and restore operation was added in those predicates.

The solution I had was to introduce some additional classes which allow for the predicate to make changes. This is the `Evaluation` record which has the test result and whatever state changes should be done if the caller deems it so.

Open question: the PathPatternPredicate appeared to be the only implementation which updated the request, but do folks have state-modifying predicates out in the wild? It would appears physically possible, but at the same time against the `@FunctionalInterface` nature of the definition of RequestPredicate. So this _could_ break extant use cases, but those use cases are not necessarily _philosophically_ correct. Given PathPatternPredicate is internal (and "rule breaking" by changing state), I supposed internally we could support its rule breaking.

The existing RequestPredicateAttributesTests all pass still.

~I only did the changes in webflux, but could port these changes over to the servlet functions as well later if this is accepted.~

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/RequestPredicates.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/RequestPredicates.java
@@ -21,7 +21,6 @@ import java.net.URI;
 import java.security.Principal;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -295,11 +294,6 @@ public abstract class RequestPredicates {
 		}
 	}
 
-	private static void restoreAttributes(ServerRequest request, Map<String, Object> attributes) {
-		request.attributes().clear();
-		request.attributes().putAll(attributes);
-	}
-
 	private static Map<String, String> mergePathVariables(Map<String, String> oldVariables,
 			Map<String, String> newVariables) {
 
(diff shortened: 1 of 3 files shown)


poutsma:
Interesting. I will take a closer look for 6.1 M1.

yuzawa-san:
@poutsma Thanks, here is some background. I have a few dozen routes registered like so:
```java
RouterFunctions.Builder builder = RouterFunctions.route();
builder.add(RouterFunctions.route(RequestPredicates.GET("/foo"), fooHandler));
builder.add(RouterFunctions.route(RequestPredicates.GET("/bar"), barHandler));
....
return builder.build();
```

and I collected this memory icicle graph:
image

and this cpu icicle graph:
image

The saving and restoring uses the majority of the memory and cpu relative to the actual evaluation of the HttpMethodPredicate and PathPatternPredicate

My handlers dont use much cpu or memory, so this was constituting an non-negligible percentage of the applications overall footprint.

yuzawa-san:
Here are some more extensive icicle graphs generated from [my stress test](https://gist.github.com/yuzawa-san/5102be26fb9e4472419c23b39c4524b2).

memory before (the selected item is basically all of the "memory after" graph):
image

memory after:
image


cpu before:
image

cpu after:
image

and i have two more optimizations within that which should cut down even more: https://github.com/spring-projects/spring-framework/pull/30138 and https://github.com/spring-projects/spring-framework/pull/30139

poutsma:
@yuzawa-san Great work, thanks!

PR #30045, opened 2023-02-27
Add SpEL support for registered MethodHandles

### LINKED ISSUE #27099: SpEL FunctionReference could support MethodHandle
A SpEL `FunctionReference` is currently limited to a **static** `Method` object, limiting what it can do drastically.

Oddly enough, the source code seems to imply that it could reference something else than a static Method, but if I'm not mistaken, there is only one code path, failing early with an exception if the method is not a static method:

https://github.com/spring-projects/spring-framework/blob/a2ef6badc4c76790128910851fdde0df55ec15f9/spring-expression/src/main/java/org/springframework/expression/spel/ast/FunctionReference.java#L73-L77

then later:
https://github.com/spring-projects/spring-framework/blob/a2ef6badc4c76790128910851fdde0df55ec15f9/spring-expression/src/main/java/org/springframework/expression/spel/ast/FunctionReference.java#L105-L108

Anyways, it would be nice if `FunctionReference` could actually reference a `MethodHandle`, thus capturing a receiver (or even more context) that wouldn't have to leak in the signature to the "function".

Are there any show stoppers for this? AFAIU, compilation to bytecode is optional, so even that shouldn't be a problem (but maybe it is actually possible). Happy to work on a PR for this if deemed acceptable.

simonbasle:
This commit adds support for MethodHandles in SpEL, using the same
syntax as user-defined functions (which also covers reflective Methods).

The most benefit is expected with handles that capture a static method
with no arguments, or with fully bound handles (where all the arguments
have been bound, including a target instance as first bound argument
if necessary). Partially bound MethodHandle should also be supported.

A best effort approach is taken to detect varargs as there is no API
support to determine if an argument is a vararg or an explicit array,
unlike with Method. Argument conversions are also applied. Finally,
array repacking is not always necessary with varargs so it is only
performed when the vararg is the sole argument to the invoked method.

Closes gh-27099
Closes gh-30045

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
@@ -16,6 +16,8 @@
 
 package org.springframework.expression.spel.support;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodType;
 import java.lang.reflect.Array;
 import java.lang.reflect.Executable;
 import java.lang.reflect.Method;
@@ -23,6 +25,7 @@ import java.util.List;
 import java.util.Optional;
 
 import org.springframework.core.MethodParameter;
+import org.springframework.core.ResolvableType;
 import org.springframework.core.convert.TypeDescriptor;
 import org.springframework.expression.EvaluationException;
 import org.springframework.expression.TypeConverter;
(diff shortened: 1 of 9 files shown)


spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
```diff
@@ -330,6 +333,90 @@ else if (!sourceType.equals(targetType.getElementTypeDescriptor())) {
...
+	 */
+	public static boolean convertAllMethodHandleArguments(TypeConverter converter, Object[] arguments,
+			MethodHandle methodHandle, @Nullable Integer varargsPosition) throws EvaluationException {
+		boolean conversionOccurred = false;
+		final MethodType methodHandleArgumentTypes = methodHandle.type();
+		if (varargsPosition == null) {
+			for (int i = 0; i < arguments.length; i++) {
+				Class<?> argumentClass = methodHandleArgumentTypes.parameterType(i);
+				ResolvableType resolvableType = ResolvableType.forClass(argumentClass);
+				TypeDescriptor targetType = new TypeDescriptor(resolvableType, argumentClass, null);
```
simonbasle, line 355: here the subtle difference compared to the older `convertAllMethodArguments` above is that we get `TypeDescriptor` out of a raw `Class` only (vs  a `MethodParam` obtained via reflection with `MethodParameter.forExecutable`)

ericbottard:
A quick glance at the docs and test tells me this would have been useful. We went with another route (and that leaked in the exposed API sadly), but glad to see this got supported eventually. Thanks!

simonbasle:
Now rebased on top of latest reference doc changes (Antora) and squashed into a single commit.

fanthos:
Is it possible to backport this feature to 5.x? Java 17 is not ready in many environments.

simonbasle:
> Is it possible to backport this feature to 5.x? Java 17 is not ready in many environments.

@fanthos this is very unlikely. We don't backport purely new features to the 5.3.x line, and now that the work has started on 6.1.x, we tend to avoid doing so in 6.0.x as well. Furthermore, `MethodHandle` was introduced in JDK 9 but 5.3.x has a baseline of JDK 8, making backporting even less unlikely.

PR #30046, opened 2023-02-27
Assert non-null arguments in `DefaultServerRequestBuilder` methods

srivatsa-cfp:
- Add Null check for Http Header Name , https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpHeaders.html#add(java.lang.String,java.lang.String). As per doc only header values can be nullable
- Add Null check for Http HeaderConsumer
- Add Null check for Cookie Name 
- Add Null check for CookierConsumer

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/DefaultServerRequestBuilder.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/server/DefaultServerRequestBuilder.java
@@ -122,6 +122,7 @@ class DefaultServerRequestBuilder implements ServerRequest.Builder {
 
 	@Override
 	public ServerRequest.Builder header(String headerName, String... headerValues) {
+		Assert.notNull(headerName, "Header Name must not be null");
 		for (String headerValue : headerValues) {
 			this.headers.add(headerName, headerValue);
 		}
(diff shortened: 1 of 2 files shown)


sbrannen:
Note that `@NonNullApi` is specified at the package level in `package-info.java`. Thus, none of those arguments are allowed to be `null`.

However, we do have inconsistent state in both `DefaultServerRequestBuilder` variants with regard to explicit non-null assertions.

If we introduce additional non-null assertions for those arguments, we would want to do that consistently for all such methods _and_ consistently across both `DefaultServerRequestBuilder` variants (MVC and WebFlux).

srivatsa-cfp:
> 
@sbrannen Yes specification says it to be not null, but no implementation might cause an inconsistent. I will verify it to be consistent across MVC & WebFlux. Do you any further concerns on non-null assertions?

srivatsa-cfp:
Added the assert non-null checks in DefaultServerRequestBuilder in spring-webflux

sbrannen:
@srivatsa-cfp,

We may decide to introduce non-null assertions in those builder methods, _or_ we may decide to remove all of the non-null assertions and to rely on null-safety annotations instead.

So, please hold off on making further changes until the team has reached a decision on this topic.

poutsma:
@srivatsa-cfp We have decided that null assertions would be useful to have in builder methods, so please continue with your PR and let me know when it's done.

srivatsa-cfp:
@poutsma Added all the required null assertions in DefaultSertverRequestBuilder across webflux and web-mvc.

Issue #30033, opened 2023-02-25
Server request URL with spring-webflux 6.0.5 is in resolved IP6 format


intosoft:
After upgrade to spring boot 3.0.3, **Swagger authorization** stopped working on `http://localhost:8080`.

![image](https://user-images.githubusercontent.com/5756904/221354661-2474985f-fec3-46b6-9c49-5941e4acaf0f.png)

The issue is caused by changed method:

```java
private static URI resolveBaseUrl(HttpServerRequest request) throws URISyntaxException {
		String scheme = getScheme(request);

		InetSocketAddress hostAddress = request.hostAddress();
		...
}
```

From:

```java
private static URI resolveBaseUrl(HttpServerRequest request) throws URISyntaxException {
		String scheme = getScheme(request);
		String header = request.requestHeaders().get(HttpHeaderNames.HOST);
		...
}
```

The root cause of issue is that instead of resolving host to `http://localhost`, it resolves `http://[0:0:0:0:0:0:0:1]`.

The main functionality of swagger works with this IP address, but **authorization fails** because it redirects to:

![image](https://user-images.githubusercontent.com/5756904/221354738-6f282ff7-180f-4cfa-bf95-fc18c0f7a3dc.png)

To fix that issue I can set `server.address`, but then it breaks accessing swagger by external IP.

I found the issue after checking swagger, but rest of application works fine. Should openapi be fixed or `resolveBaseUrl` method?

Even enforcing IP4 by `-Djava.net.preferIPv4Stack=true` does not solve issue.

Downgrading to spring-web 6.0.4 solves the issue. Also using IP instead of DSN does not generate such issues.

intosoft:
@bclozel - thanks for handling this.

wilkinsona:
We suspect that https://github.com/spring-projects/spring-boot/issues/34395 is another symptom of the same underlying problem.

rstoyanchev:
Indeed, both this issue and https://github.com/spring-projects/spring-boot/issues/34395 are linked to the change in #28601. 

[...]

PR #30047, opened 2023-02-28
Avoid duplicate URI construction in ReactorServerHttpRequest

yuzawa-san:
It appeared that two URI's were being constructed. The first one was then manually created then converted to a string and then concatenated with the path and then back to a URI. This change only does a single phase of parsing. Additionally, ipv6 InetSocketAddress instances will properly be converted to hostnames with the proper bracket format. I did the change in current and future reactor netty implementations.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorNetty2ServerHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorNetty2ServerHttpRequest.java
@@ -27,6 +27,7 @@ import io.netty5.channel.Channel;
 import io.netty5.handler.codec.http.HttpHeaderNames;
 import io.netty5.handler.codec.http.headers.HttpCookiePair;
 import io.netty5.handler.ssl.SslHandler;
+import io.netty5.util.NetUtil;
 import org.apache.commons.logging.Log;
 import reactor.core.publisher.Flux;
 import reactor.netty5.ChannelOperationsId;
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/server/reactive/ReactorNetty2ServerHttpRequest.java
```diff
@@ -75,48 +76,23 @@ public ReactorNetty2ServerHttpRequest(HttpServerRequest request, Netty5DataBuffe
...
-				catch (NumberFormatException ex) {
-					throw new URISyntaxException(header, "Unable to parse port", portIndex);
-				}
-			}
-			else {
-				return new URI(scheme, header, null, null);
-			}
+		CharSequence header = request.requestHeaders().get(HttpHeaderNames.HOST);
+		if (header != null) {
+			return header;
```
rstoyanchev, line 90: The "Host" header can contain host and port, see for example https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host. Also, parsing needs to take into account that the host maybe an IPv6 address. This is what the removed code [89-110] does, and that needs to remain.

spring-web/src/main/java/org/springframework/http/server/reactive/ReactorNetty2ServerHttpRequest.java
```diff
@@ -75,48 +76,23 @@ public ReactorNetty2ServerHttpRequest(HttpServerRequest request, Netty5DataBuffe
...
 	}
 
-	private static URI resolveBaseUrl(HttpServerRequest request) throws URISyntaxException {
-		String scheme = getScheme(request);
-
+	private static CharSequence resolveHost(HttpServerRequest request) {
 		InetSocketAddress hostAddress = request.hostAddress();
 		if (hostAddress != null) {
-			return new URI(scheme, null, hostAddress.getHostString(), hostAddress.getPort(), null, null, null);
+			return NetUtil.toSocketAddressString(hostAddress);
```
rstoyanchev, line 85: Before this change, the URI constructor that accepts individual URI components would validate the host and port. After this change, the value returned from `resolveHost` is not validated, the URI constructor with a String (called on line 79) doesn't either.

When I tested whether this change would help with https://github.com/spring-projects/spring-boot/issues/34395 where X-Forwarded-HOST is "localhost:3000" and X-Forwarded-Port is "3000", I saw that `resolveHost` returns "localhost:3000:3000", and the resulting URI was "http://localhost:3000:3000/hello". When I test without this change, there is an exception from the URI constructor that was previously used here because "localhost:3000" is not a valid host.

rstoyanchev (commented):
I don't mind eliminating one additional URI construction conceptually, but the performance improvement is negligible, and at the same time, as per my comments below, there is a level of validation lost, and some functionality is removed that should not be.

Thanks for the pull request, but overall I'm not inclined to accept the change.

yuzawa-san:
ok i'll put the validations back, but i'm still going to try to do the URI parse once.

image

yuzawa-san:
i'm going to close this until https://github.com/spring-projects/spring-framework/issues/30033 is resolved

PR #30062, opened 2023-03-02
Prefer request hostName and hostPort in ReactorServerHttpRequest

### LINKED ISSUE #30033: Server request URL with spring-webflux 6.0.5 is in resolved IP6 format

violetagg:
- Prefer request `hostName` and `hostPort` in `ReactorServerHttpRequest#resolveBaseUrl`
- The request `hostName` and `hostPort` are derived from the `Host`/`X-Forwarded-*`/`Forwarded` header associated with this request
- Do not add the port when it is the default one

Closes gh-30033

### DIFF
--- a/framework-platform/framework-platform.gradle
+++ b/framework-platform/framework-platform.gradle
@@ -11,7 +11,7 @@ dependencies {
 	api(platform("io.micrometer:micrometer-bom:1.10.4"))
 	api(platform("io.netty:netty-bom:4.1.89.Final"))
 	api(platform("io.netty:netty5-bom:5.0.0.Alpha5"))
-	api(platform("io.projectreactor:reactor-bom:2022.0.3"))
+	api(platform("io.projectreactor:reactor-bom:2022.0.4-SNAPSHOT"))
 	api(platform("io.rsocket:rsocket-bom:1.1.3"))
 	api(platform("org.apache.groovy:groovy-bom:4.0.8"))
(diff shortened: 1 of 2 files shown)


violetagg:
@rstoyanchev PTAL

PR #30068, opened 2023-03-05
Update `Pointcuts` to match getter methods by also excluding `void` return type

enimiste:
A getter method should always have a return type other than `void`.


rstoyanchev:
Thanks for the suggestion. Do you come across an actual method that begins with "get" and returns void?

enimiste:
Hi, @rstoyanchev 
No, but as there non garanties from the compiler that no one can type a getter method with void return, i think the framework should filter out this case.

A getter with a return type different from void and without any arguments is juste a convention from JavaBean and not a compilation error.

Thanks,

vishalsingh2972:
@enimiste @rstoyanchev is this issue still open?

snicoll:
Given how strict the setter `match` implementation is, I think this is a good chance that improves the consistency.

PR #30069, opened 2023-03-06
Ensure all packages declare `package-info.java` with null-safety annotations

### LINKED ISSUE #30056: Ensure all packages declare `package-info.java` with null-safety annotations
## Overview

Since Spring Framework 5.0 (see #20099), we annotate packages in `package-info.java` files with null-safety annotations such as `@NonNullApi`; however, not all packages contain `package-info.java` files which prevents us from enforcing non-null semantics by default for those packages.

## Scope

This applies to all packages within `src/main`. Conversely, this does not apply to packages in `src/test`.

## Deliverables

- [ ] Configure the [`JavadocPackage`](https://checkstyle.sourceforge.io/config_javadoc.html#JavadocPackage) Checkstyle module to require `package-info.java` files for all packages under `src/main`.
- [ ] Determine if it is possible to require that `package-info.java` files include null-safety annotations – for example, via Checkstyle.
- [ ] If it's possible to require that `package-info.java` files include null-safety annotations, configure the necessary infrastructure.
- [ ] Introduce missing `package-info.java` files with package-level Javadoc and null-safety annotations.

edyda99:
## Overview

This commit adds the `JavadocPackage` Checkstyle module to the project's configuration file (`checkstyle.xml`). The module is configured to emit a warning message when a `package-info.java` file is missing, rather than throwing an exception. In addition, suppressions have been added to prevent the `JavadocPackage` checks from being applied to packages outside of the `src/main` directory. These suppressions ensure that the module only checks the packages that are relevant to the project and prevents false positives from being reported.

## Deliverables

- [x] Configure the [`JavadocPackage`](https://checkstyle.sourceforge.io/config_javadoc.html#JavadocPackage) Checkstyle module to require `package-info.java` files for all packages under `src/main`.
- [x] Determine if it is possible to require that `package-info.java` files include null-safety annotations – for example, via Checkstyle.
- [x] If it's possible to require that `package-info.java` files include null-safety annotations, configure the necessary infrastructure.
- [x] Introduce missing `package-info.java` files with package-level Javadoc and null-safety annotations.

### DIFF
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -240,4 +240,7 @@
 		<module name="io.spring.javaformat.checkstyle.check.SpringJavadocCheck"/>
 		<module name="io.spring.javaformat.checkstyle.check.SpringJUnit5Check"/>
  	</module>
+
+	<!--package-info checker -->
+	<module name="JavadocPackage"/>
 </module>
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -239,6 +239,24 @@
 		<module name="io.spring.javaformat.checkstyle.check.SpringCatchCheck"/>
 		<module name="io.spring.javaformat.checkstyle.check.SpringJavadocCheck"/>
 		<module name="io.spring.javaformat.checkstyle.check.SpringJUnit5Check"/>
+		<!--package-info should contain null-safety annotations-->
+		<!--These two modules will fail to detect multiline annotations-->
+		<module name="RegexpSinglelineJavaCheck">
+			<property name="id" value="javaDocPackageNonNullFieldsAnnotation"/>
+			<property name="format" value="(@NonNullFields|@org\.springframework\.lang\.NonNullFields)"/>
+			<property name="minimum" value="1"/>
+			<property  name="maximum" value="1"/>
+			<property name="message" value="package-info.java is missing required null-safety annotation @NonNullFields."/>
+			<property name="ignoreComments" value="true"/>
+		</module>
+		<module name="RegexpSinglelineJavaCheck">
+			<property name="id" value="javaDocPackageNonNullApiAnnotation"/>
+			<property name="format" value="(@NonNullApi|@org\.springframework\.lang\.NonNullApi)"/>
+			<property name="minimum" value="1"/>
+			<property  name="maximum" value="1"/>
+			<property name="message" value="package-info.java is missing required null-safety annotation @NonNullApi."/>
+			<property name="ignoreComments" value="true"/>
+		</module>
  	</module>
 
 	<!--package-info checker -->
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -246,6 +246,7 @@
 			<property name="format" value="(@NonNullFields|@org\.springframework\.lang\.NonNullFields)"/>
 			<property name="minimum" value="1"/>
 			<property  name="maximum" value="1"/>
+			<property name="severity" value="warning"/>
 			<property name="message" value="package-info.java is missing required null-safety annotation @NonNullFields."/>
 			<property name="ignoreComments" value="true"/>
 		</module>
@@ -254,6 +255,7 @@
 			<property name="format" value="(@NonNullApi|@org\.springframework\.lang\.NonNullApi)"/>
 			<property name="minimum" value="1"/>
 			<property  name="maximum" value="1"/>
+			<property name="severity" value="warning"/>
 			<property name="message" value="package-info.java is missing required null-safety annotation @NonNullApi."/>
 			<property name="ignoreComments" value="true"/>
 		</module>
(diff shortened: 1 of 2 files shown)


simonbasle (commented):
thanks for the suggestion. Can you provide more details about the choices you made for checkstyle modules?

simonbasle:
(didn't see the issue reference, which probably provides some context. my bad)

simonbasle:
@edyda99 I think the `RegexpJavaSingleLine` variant does the job (unless you've spotted occurrences of multilines nullability annotations?) The other one is also harder to read and has its own caveats with comments, so I'd go with the first one.

edyda99:
Hi @simonbasle, I will go with RegexpJavaSingleLine.
No, I didn't spot any occurrences of multilines nullability, but it was about preventing it from happening.
So should I remove the whitespace warning? ( I would suggest keeping it and increasing the severity to "**error**", so whenever someone writes `@\n<annotation_name>` an error will be thrown)

simonbasle:
> ( I would suggest keeping it and increasing the severity to "**error**", so whenever someone writes `@\n<annotation_name>` an error will be thrown)

Perhaps a simpler one line regexp like `@\s*$` (at then any number of whitespace then the end of the line) would fit the bill? That said, I think this PR can focus on the actual null-safety annotations.

simonbasle:
>     * Do you suggest doing micro commits for each missing package-info.java?

Not necessarily, if that's the only thing in the commit.

>     * What do I need to rely on before adding null-safety annotations on a newly added package-info?

It depends on how many package-info are missing. I would locally prepare the change and run some static analysis trying to find null-safety issues raising from that change. If the build doesn't break, I'm not sure I would make it a goal of this PR to fix the null-safety warnings though. Wdyt @sbrannen (also cc @sdeleuze)

sbrannen:
> > * Do you suggest doing micro commits for each missing package-info.java?
> 
> Not necessarily, if that's the only thing in the commit.

I agree with @simonbasle.

Ideally the final result will be 3-4 commits.

- enforce that `package-info.java` files exist and contain Javadoc
  - build should fail
- introduce missing `package-info.java` files
  - build should pass
- enforce that `package-info.java` files declare `@NonNullApi` and `@NonNullFields`
  - build should fail
- introduce missing `@NonNullApi` and `@NonNullFields` declarations
  - build should pass

Of course, we may squash all commits in the PR into a single commit if that turns out to be easier.

> > * What do I need to rely on before adding null-safety annotations on a newly added package-info?

I don't understand the question. Can you please expound?

> It depends on how many package-info are missing. I would locally prepare the change and run some static analysis trying to find null-safety issues raising from that change. If the build doesn't break, I'm not sure I would make it a goal of this PR to fix the null-safety warnings though. Wdyt @sbrannen (also cc @sdeleuze)

Ensuring that proper null-safety semantics are enforced _within_ code in newly annotated packages is beyond the scope of this PR/issue. I'll create a follow-up issue to address that.

src/checkstyle/checkstyle.xml
```diff
@@ -239,5 +239,21 @@
...
- 	</module>
+		<!--package-info should contain null-safety annotations-->
+		<!--This module will fail to detect multiline annotations-->
+		<module name="RegexpSinglelineJavaCheck">
+			<property name="id" value="javaDocPackageAnnotations"/>
+			<property name="format" value="(@NonNull|@org\.springframework\.lang\.NonNull)"/>
+			<property name="severity" value="warning"/>
+			<property name="minimum" value="1"/>
+			<property  name="maximum" value="5"/>
+			<property name="message" value="package-info.java is missing null-safety annotations  or has null-safe annotations made up of multiple lines. Please check."/>
```
sbrannen, line 250:
```suggestion
			<property name="message" value="package-info.java is missing required null-safety annotations @NonNullApi and @NonNullFields."/>
```

Spring's `@NonNullApi` and `@NonNullFields` annotations should never span multiple lines.

sbrannen (changes requested):
Thanks for working on this!

It's coming along nicely.

After that, all that's left is ensuring that the build fails for violations and that all missing `package-info.java` files have been created (so that the build passes again).

src/checkstyle/checkstyle.xml
```diff
@@ -239,5 +239,21 @@
...
+			<property name="ignoreComments" value="true"/>
+		</module>
+	</module>
+
+	<!--package-info checker -->
+	<module name="JavadocPackage">
+		<property name="severity" value="warning"/>
```
sbrannen, line 257: We also want these violations to fail the build.

sbrannen:
You also need to suppress violations in the `framework-docs` module, since that source code is only used in the reference manual.

edyda99:
Hi again,

First sorry for the continuous Force pushes, but I wanted to reduce my commits.
I reduced them to 4 micro commits, with the last one containing the full commit message.
All comments are resolved.

The merge request is ready for the final review

Thank you again!

edyda99:
@sbrannen Ok Great
Could you suggest something else for me?

sbrannen:
This has been merged into `main` in 268e3fec9921653de17ab2f495039239f6b6152e and 7e32f504b021015ca8a7eaed7280453fad724b0f and revised and completed in 99e54fec3ac8dfedd2537fde5f19fbcc4f1f6202.

The latter turned out to be a bit more work than expected. 👀 

Thanks again for the contribution, @edyda99!

PR #30074, opened 2023-03-06
RequestContextFilter: for async filter chains, defer completing request attributes until async processing completes

### LINKED ISSUE #30073: RequestContextFilter is not compatible with async request filters
`RequestContextFilter` sets some thread-locals in holders to bind the request scope to the current servlet request.
Upon returning from `doFilterInternal`, it takes care to both clear the thread locals as well as mark the request attributes as completed.
After the request attributes are completed, no further access to the attributes is possible - it throws an exception.

We upgraded one of our Servlet filters further down the chain from RequestContextFilter from a classic synchronous filter to a new asynchronous filter. This means that when our `doFilter` returns, it returns all the way up through the `RequestContextFilter` back to the container. However, the request is not yet completed - later on, our filter calls `resume`. Our filter takes care to save and restore the `RequestContextHolder` appropriately, to propagate the Spring request scope. From there our Servlet gets called, it tries to access the request scope, and encounters a fatal error, because it is marked as complete.

It seems that in the case of async filters, the request context filter should not finish the request attributes immediately. Instead it should defer marking it as complete until the async processing finishes.

stevenschlansker:
RequestContextFilter: for async filter chains, defer completing request attributes until async processing completes

Fixes #30073

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/filter/RequestContextFilter.java
+++ b/spring-web/src/main/java/org/springframework/web/filter/RequestContextFilter.java
@@ -18,6 +18,8 @@ package org.springframework.web.filter;
 
 import java.io.IOException;
 
+import jakarta.servlet.AsyncEvent;
+import jakarta.servlet.AsyncListener;
 import jakarta.servlet.FilterChain;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Closing as per https://github.com/spring-projects/spring-framework/issues/30073#issuecomment-1464061139.

Issue #24963, opened 2020-04-23
GenericTypeResolver.resolveType() should resolve ParameterizedType recursively


quaff:
```java
import java.lang.reflect.Method;
import java.lang.reflect.Type;
import java.util.List;
import org.springframework.core.GenericTypeResolver;

public class Main {
	public interface Parent<T> {
		T m1();
		List<T> m2();
		List<List<T>> m3();
	}
	public interface Child extends Parent<String> {
	}
	public static void main(String[] args) {
		for (Method m : Parent.class.getMethods()) {
			Type type = m.getGenericReturnType();
			type = GenericTypeResolver.resolveType(type, Child.class);
			System.out.println("method " + m.getName() + "() will return " + type);
		}
	}
}
```
will print
```
method m3() will return java.util.List<java.util.List<T>>
method m1() will return class java.lang.String
method m2() will return java.util.List<java.lang.String>
```
The expected type for m3 should be `java.util.List<java.util.List<String>>`

quaff:
```java
Type type = m.getGenericReturnType(); // List<List<T>>
ResolvableType.forType(type).hasUnresolvableGenerics(); //should be true but false
```

quaff:
Here is my workaround
```java
	public static Type resolve(Type type, Class<?> contextClass) {
		if (type instanceof ParameterizedType) {
			ParameterizedType pt = (ParameterizedType) type;
			ResolvableType[] generics = new ResolvableType[pt.getActualTypeArguments().length];
			int i = 0;
			for (Type arg : pt.getActualTypeArguments()) {
				generics[i++] = ResolvableType.forType(resolve(arg, contextClass));
			}
			return ResolvableType.forClassWithGenerics((Class<?>) pt.getRawType(), generics).getType();
		}
		return GenericTypeResolver.resolveType(type, contextClass);
	}
```

quaff:
Any update?

quaff:
Superseded by https://github.com/spring-projects/spring-framework/pull/30079

PR #30079, opened 2023-03-08
Resolve type variable recursively in `GenericTypeResolver`

quaff:
- Addresses #24963

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
+++ b/spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
@@ -39,6 +39,7 @@ import org.springframework.util.ConcurrentReferenceHashMap;
  * @author Rob Harrop
  * @author Sam Brannen
  * @author Phillip Webb
+ * @author Yanming Zhou
  * @since 2.5.2
  */
 public final class GenericTypeResolver {
@@ -167,7 +168,7 @@ public final class GenericTypeResolver {
 			else if (genericType instanceof ParameterizedType parameterizedType) {
 				ResolvableType resolvedType = ResolvableType.forType(genericType);
 				if (resolvedType.hasUnresolvableGenerics()) {
-					Class<?>[] generics = new Class<?>[parameterizedType.getActualTypeArguments().length];
+					ResolvableType[] generics = new ResolvableType[parameterizedType.getActualTypeArguments().length];
 					Type[] typeArguments = parameterizedType.getActualTypeArguments();
 					ResolvableType contextType = ResolvableType.forClass(contextClass);
 					for (int i = 0; i < typeArguments.length; i++) {
@@ -175,14 +176,17 @@ public final class GenericTypeResolver {
 						if (typeArgument instanceof TypeVariable<?> typeVariable) {
 							ResolvableType resolvedTypeArgument = resolveVariable(typeVariable, contextType);
 							if (resolvedTypeArgument != ResolvableType.NONE) {
-								generics[i] = resolvedTypeArgument.resolve();
+								generics[i] = resolvedTypeArgument;
 							}
 							else {
-								generics[i] = ResolvableType.forType(typeArgument).resolve();
+								generics[i] = ResolvableType.forType(typeArgument);
 							}
 						}
+						else if (typeArgument instanceof ParameterizedType) {
+							generics[i] = ResolvableType.forType(resolveType(typeArgument, contextClass));
+						}
 						else {
-							generics[i] = ResolvableType.forType(typeArgument).resolve();
+							generics[i] = ResolvableType.forType(typeArgument);
 						}
 					}
 					Class<?> rawClass = resolvedType.getRawClass();
--- a/spring-context/src/main/java/org/springframework/context/event/ApplicationListenerMethodAdapter.java
+++ b/spring-context/src/main/java/org/springframework/context/event/ApplicationListenerMethodAdapter.java
@@ -67,6 +67,7 @@ import org.springframework.util.StringUtils;
  * @author Juergen Hoeller
  * @author Sam Brannen
  * @author Sebastien Deleuze
+ * @author Yanming Zhou
  * @since 4.2
  */
 public class ApplicationListenerMethodAdapter implements GenericApplicationListener {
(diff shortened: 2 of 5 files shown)


quaff:
Rebased to main, ready for review.

spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
```diff
@@ -167,25 +168,32 @@ public static Type resolveType(Type genericType, @Nullable Class<?> contextClass
...
+							generics[i] = ResolvableType.forType(typeArgument);
 						}
 					}
+					else if (typeArgument instanceof ParameterizedType) {
+						generics[i] = ResolvableType.forType(resolveType(typeArgument, contextClass));
+					}
 					else if (typeArgument instanceof WildcardType wildcardType) {
-						generics[i] = resolveWildcard(wildcardType, contextType).resolve();
+						ResolvableType result = resolveWildcard(wildcardType, contextType);
+						if (result.resolve() == null) {
```
snicoll, line 190: I wonder if that logic couldn't be applied in `resolveWildcard` since that's the only use of it. This would allow this to be a one liner like the other cases.

snicoll (changes requested):
Thanks for the PR @quaff - This looks promising as it fixes another issue that was reported against generics handling.

Unfortunately, the build is broken with those changes, see https://ge.spring.io/s/kktsevdqhmgqc/tests/overview?outcome=FAILED.

spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
quaff: > I wonder if that logic couldn't be applied in `resolveWildcard` since that's the only use of it. This would allow this to be a one liner like the other cases.

This change makes `GenericTypeResolverTests.resolveWildcardTypeWithUnbounded()` pass.

quaff:
> Thanks for the PR @quaff - This looks promising as it fixes another issue that was reported against generics handling.
> 
> Unfortunately, the build is broken with those changes, see https://ge.spring.io/s/kktsevdqhmgqc/tests/overview?outcome=FAILED.

The failed tests should adapt changes on [ResolvableType.java](https://github.com/spring-projects/spring-framework/pull/30079/files#diff-58e4c9d4400aa89a869df68b7c8974fe646147a93e41dc7c1535306996e5f550), I've updated the commit.

spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
snicoll: OK I am not saying removing. I am thinking about moving `result.resolve()` to the `resolveWildcard` method.

quaff: OK, updated as you suggested.

snicoll:
> The failed tests should adapt changes on [ResolvableType.java](https://github.com/spring-projects/spring-framework/pull/30079/files#diff-58e4c9d4400aa89a869df68b7c8974fe646147a93e41dc7c1535306996e5f550), I've updated the commit.

With those changes in place, you had to change some assertions that looks invalid to me now. For instance, event listening that accepts a `String` as input would be considered if the type is `EntityWrapper<?>`. Same for an event that accepts an `EntityWrapper<?>` would now accept a `List<?>`. 

This may be something that wrong elsewhere that the fix reveals but changing the assertion isn't what I had in mind.

Given the impact this has so late, I've moved this one for 6.2. If you have time to review the above, that would be much appreciated.

quaff:
> > The failed tests should adapt changes on [ResolvableType.java](https://github.com/spring-projects/spring-framework/pull/30079/files#diff-58e4c9d4400aa89a869df68b7c8974fe646147a93e41dc7c1535306996e5f550), I've updated the commit.
> 
> With those changes in place, you had to change some assertions that looks invalid to me now. For instance, event listening that accepts a `String` as input would be considered if the type is `EntityWrapper<?>`. Same for an event that accepts an `EntityWrapper<?>` would now accept a `List<?>`.
> 
> This may be something that wrong elsewhere that the fix reveals but changing the assertion isn't what I had in mind.
> 
> Given the impact this has so late, I've moved this one for 6.2. If you have time to review the above, that would be much appreciated.

I've fixed that, please review. @snicoll

quaff:
Any chance to bring this to 6.1? @jhoeller @snicoll

snicoll:
@quaff please refrain from ping like this. You've already ping me once and trust that we're doing our best to handle it in time if we can.

snicoll:
@quaff with https://github.com/spring-projects/spring-framework/issues/31690#issuecomment-1831867944 some code that we didn't intend to have has been reverted. If you have time to rebase, it would be much appreciated.

quaff:
> @quaff with [#31690 (comment)](https://github.com/spring-projects/spring-framework/issues/31690#issuecomment-1831867944) some code that we didn't intend to have has been reverted. If you have time to rebase, it would be much appreciated.

Done.

snicoll (approved):

kilink:
We happened to be testing our apps on the latest Spring snapshot, and I believe this change has introduced an infinite recursion bug which causes a StackOverFlowError. When we start several of our Spring Boot apps, we end up with the following error:
```
Exception in thread "main" java.lang.StackOverflowError
	at org.springframework.util.ObjectUtils.nullSafeHashCode(ObjectUtils.java:451)
	at org.springframework.core.ResolvableType.calculateHashCode(ResolvableType.java:1012)
	at org.springframework.core.ResolvableType.hashCode(ResolvableType.java:1000)
	at org.springframework.util.ObjectUtils.nullSafeHashCode(ObjectUtils.java:451)
	at org.springframework.core.ResolvableType.calculateHashCode(ResolvableType.java:1012)
	at org.springframework.core.ResolvableType.hashCode(ResolvableType.java:1000)
	at org.springframework.util.ObjectUtils.nullSafeHashCode(ObjectUtils.java:451)
```

I narrowed it down to a specific ApplicationEvent which triggers it. The following will reproduce the problem:

```java
AvailabilityChangeEvent<LivenessState> event = new AvailabilityChangeEvent<>(new Object(), LivenessState.CORRECT);
ResolvableType resolvableType = ResolvableType.forInstance(event);
resolvableType.hasUnresolvableGenerics();
```
[AvailabilityChangeEvent](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/availability/AvailabilityChangeEvent.java) and [LivenessState](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/availability/LivenessState.java) are both classes from Spring Boot.

kilink:
Looks like it can be reproduced directly with just an Enum:

```java
enum Foo { BAR }
ResolvableType.forType(Foo.class.getGenericSuperclass()).hasUnresolvableGenerics();
// or
ResolvableType.forClassWithGenerics(Enum.class, Foo.class).hasUnresolvableGenerics();
```

kilink:
Seems like it's self-bounding generics that are the problem, the following triggers the bug as well:

```java
class Foo<T extends Foo<T>> {}
class Bar extends Foo<Bar> {}

ResolvableType.forInstance(new Bar()).hasUnresolvableGenerics();
```

PR #30092, opened 2023-03-09
Remove extra copy of headers/cookies in WebClient

yuzawa-san:
When using the reactive webclient with default headers (and cookies) AND headers (and cookies) specified at the request level, there is an additional intermediate map created prior to copying to the final ClientRequest. I believe this can be eliminated in the changes proposed here. 

Here is part of the icicle graph of memory allocations:
image

The stuff done in the middle section should go away and be done on the right. Additionally, the fixes in https://github.com/spring-projects/spring-framework/pull/29972 will improve the forEach and ReadOnlyHttpHeaders performance.

The final copy into the ClientRequest used `addAll` but I used `putAll` on the defaults and request-provided values in order to preserve the order of operations. Since it appears that nothing else is setting the headers/cookies on that ClientRequest.Builder using add vs put should not matter.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/reactive/server/DefaultWebTestClient.java
+++ b/spring-test/src/main/java/org/springframework/test/web/reactive/server/DefaultWebTestClient.java
@@ -370,8 +370,8 @@ class DefaultWebTestClient implements WebTestClient {
 
 		private ClientRequest.Builder initRequestBuilder() {
 			ClientRequest.Builder builder = ClientRequest.create(this.httpMethod, initUri())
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Thanks for the suggestion. I have added an extra small refactoring, eliminating the intermediate `HttpRequest` passed into `DefaultResponseSpec` and passing the HttpMethod and URI instead since that's all that was needed.

yuzawa-san:
@rstoyanchev thanks you, could you please consider the sibling PR to this one https://github.com/spring-projects/spring-framework/pull/29972? that one has more performance gains.

PR #30093, opened 2023-03-09
Optimize AggregatesSpliterator to avoid repetitive work #25508 issue …

harshal-gav:
Depends on #25508

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/MergedAnnotationsCollection.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/MergedAnnotationsCollection.java
@@ -249,7 +249,7 @@ final class MergedAnnotationsCollection implements MergedAnnotations {
 						annotationResult, this.mappingCursors[annotationResult]);
 				this.mappingCursors[annotationResult]++;
 				if (mergedAnnotation == null) {
-					return tryAdvance(action);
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @harshal-gav,

Are you proposing this as an extended optimization if #25508 is merged?

Can you please provide further insight on the rationale for the proposed changes?

harshal-gav:
The proposed change of adding additional parameters to the tryAdvance method in the AggregatesSpliterator class can help avoid repetitive work by reducing the number of times the tryAdvance method needs to be called.

In the current implementation, when mergedAnnotation == null, the tryAdvance method is called with only the aggregate and action parameters. This means that if the tryAdvance method is not able to advance the spliterator, it will need to be called again with the same parameters. This can result in repetitive work, especially if the tryAdvance method is expensive or if the spliterator is advanced only one element at a time.

By adding additional parameters to the tryAdvance method, such as lowestDistance and annotationResult, the method can make use of previously computed values and avoid repetitive work. For example, if the lowestDistance and annotationResult values are computed during a previous call to the tryAdvance method, they can be reused in subsequent calls instead of recomputing them from scratch.

Overall, this change can help improve the performance of the AggregatesSpliterator class by reducing the number of times the tryAdvance method needs to be called, and by making better use of previously computed values.

Issue #30097, opened 2023-03-09
Allow SnakeYaml 2.0 runtime compatibility


strehle:
**Affects:** 5.3.x

---
Please upgrade snakeyaml to 2.0 in branch 5.3.x, e.g. or take https://github.com/spring-projects/spring-framework/commit/9712bb62fdc772144c1876e5210eeec77c31e026

This would help many projects to get free of CVE blaming regarding snakeyaml

bclozel:
SnakeYaml is an optional dependency for Spring Framework. The commit you're pointing out merely raises the minimum version and changes an implementation to avoid deprecation warnings.

Is there a runtime issue with SnakeYaml 2.0 and Spring Framework 5.3.x? Can you report the stacktrace here and a minimal application reproducing the problem? Note that Spring Boot 2.7.x is already targeting SnakeYaml 2.0 compatibility (see spring-projects/spring-boot#34405) and it is based on Spring Framework 5.3.x.

strehle:
Stack
java.lang.NoSuchMethodError: org.yaml.snakeyaml.representer.Representer: method 'void <init>()' not found
        at org.springframework.beans.factory.config.YamlProcessor.createYaml(YamlProcessor.java:187)
        at org.springframework.beans.factory.config.YamlProcessor.process(YamlProcessor.java:164)
        at org.springframework.beans.factory.config.YamlMapFactoryBean.createMap(YamlMapFactoryBean.java:124)
        at org.springframework.beans.factory.config.YamlMapFactoryBean.getObject(YamlMapFactoryBean.java:104)
        at org.cloudfoundry.identity.uaa.impl.config.YamlServletProfileInitializer.initialize(YamlServletProfileInitializer.java:123)


See github action run for https://github.com/cloudfoundry/uaa/pull/2219/checks from https://github.com/cloudfoundry/uaa/pull/2219

strehle:
as far as I have understood snakeyaml the change from
https://github.com/spring-projects/spring-framework/blob/main/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java#L184-L189 should work also with 1.33 but is then also compatible with 2.0

bclozel:
Closed with d00fd4c.

PR #30101, opened 2023-03-10
Fix: use 2.0 compatible signature

strehle:
https://github.com/spring-projects/spring-framework/issues/30097


bclozel:
Thanks but this has been taken care of in d00fd4c.

PR #30107, opened 2023-03-12
Use IllegalStateException in ReactiveTestTransactionManager.doCommit()

izeye:
This PR changes to use `IllegalStateException` in `ReactiveTestTransactionManager.doCommit()` as no arguments are involved there.


snicoll:
thanks @izeye, the code on `main` has changed in the meantime so I've applied that to `6.0.x`.

PR #30114, opened 2023-03-13
Merge values from `Forwarded` and `X-Forwarded-*` headers

jmesny:
As specified in [RFC 7239](https://www.rfc-editor.org/rfc/rfc7239) (and documented on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)), parameters from the `Forwarded` header are optional.
Having a `Forwarded` header does not imply having a forwarded host value in it. For this reason, even if a `Forwarded` header is present, it is still needed to check for additional values provided by the `X-Forwarded-*` header family, when one is missing from the `Forwarded` header.

The problem was encountered in a real case scenario where a request passing through a proxy reached our server with the following headers:
```
Forwarded: for=1.2.3.4; by=Akamai-Nevada
X-Forwarded-Proto: https
X-Forwarded-Host: example.com
X-Forwarded-Port: 443
X-Forwarded-For: 1.2.3.4
```
Our  use of the `server.forward-headers-strategy=framework` Spring Boot property that activates the `ForwardedHeaderFilter` usage makes the `HttpServletRequest` to provide the following values:
```
httpServletRequest.getScheme() ==> http
httpServletRequest.getServerName() ==> localhost
httpServletRequest.getServerPort() ==> 64511 (random value)
```
Our expectation would be to have instead:
```
httpServletRequest.getScheme() ==> https
httpServletRequest.getServerName() ==> example.com
httpServletRequest.getServerPort() ==> 443
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.OptionalInt;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


rstoyanchev:
@jmesny, did the proxy create both standard and non-standard headers, or were they added by different proxies? It's a bit unclear how this came about, and I think it's important to try and address that further upstream rather than take them all.

We have a [security warning](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#filters-forwarded-headers) in our docs that we are not in a good position to judge whether such headers came from a trusted proxy or from external clients, and recommend that such concerns be dealt with by a proxy. 

Note also that the spec recommends in [section 7.4](https://www.rfc-editor.org/rfc/rfc7239#section-7.4) that proxies should be converting such headers.

rstoyanchev:

Issue #17920, opened 2016-04-29
Reactive types support for @Cacheable methods [SPR-14235]


spring-projects-issues:
**[Pablo Díaz-López](https://jira.spring.io/secure/ViewProfile.jspa?name=padilo)** opened **[SPR-14235](https://jira.spring.io/browse/SPR-14235?redirect=false)** and commented

Currently when using cache annotations on beans it caches the Observables like other types, So it will not cache its value.

I tried to use the following pattern to handle it:

```java
@Cacheable("items")
public Observable<Item> getItem(Long id) {
    return Observable.just(id)
        .map(myrepo::getById)
        .cache();
}
```

In the happy path, as we cache the `Observable` values works pretty well, but if `getById` throws an exception the observable is cached with the exception which isn't how it should work.

It would be also very nice to have support to `Single`.

If you give me some advice I can try to do a PR to solve this.

---

**Affects:** 4.2.5

**Sub-tasks:**
- #22152 a

**Referenced from:** pull request https://github.com/spring-projects/spring-framework/pull/1066

1 votes, 8 watchers

cbornet:
Hi, what is the status of this ? If I understand well, there was discussion about a `ReactiveCache` since JSR-107 is a blocking API. Has something evolved since the opening of this issue ?

snicoll:
Nothing has been evolving in that area I am afraid and given the lack of proprietary implementations, I  think cache vendors themselves aren't keen to explore that route either. This leaves us with the blocking model which isn't a great fit with whatever `@Cacheable` has to offer. 

There is a cache operator in reactor though and I remember so initial work trying to harmonize this here. Perhaps @simonbasle can refresh my memory?

simonbasle:
I made an attempt at an opinionated API towards caching in `reactor-extra` (see [reactor-addons repo](https://github.com/reactor/reactor-addons/tree/master/reactor-extra/src/main/java/reactor/cache)), but that has not seen much focus nor feedback since then, so I wouldn't claim it is perfect.

tomasulo:
Are there any updates on this?

snicoll:
No, the latest update on is here. Most cache libraries (including JCache) are still on a blocking model so [my comment above](https://github.com/spring-projects/spring-framework/issues/17920#issuecomment-509610787) still stands.

k631583871:
Are there any updates on this?

snicoll:
@k631583871 I already replied to that question just above your comment.

[...]

Issue #17559, opened 2015-04-30
Support caching annotations for `CompletableFuture` return values [SPR-12967]


spring-projects-issues:
**[Lukas Krecan](https://jira.spring.io/secure/ViewProfile.jspa?name=dart0)** opened **[SPR-12967](https://jira.spring.io/browse/SPR-12967?redirect=false)** and commented

It would be nice to support caching of ListenableFuture (CompletableFuture) results. Imagine that your method returns ListenableFuture and you want to cache the result (not the future itself). If I am not mistaken it is not currently supported.

I may send you a pull request, but I'd like to hear if it makes sense. It may be a bit confusing since the cache would be filled/evicted not upon the method exit but upon the future completion.

---

**Affects:** 4.1.6

spring-issuemaster:
Bulk closing outdated, unresolved issues. Please, reopen if still relevant.

jhoeller:
As explained in #17920 (specifically https://github.com/spring-projects/spring-framework/issues/17920#issuecomment-1646004504), we have a `CompletableFuture` based arrangement with the Caffeine AsyncCache as a primary provider now. That effort comes with equivalent reactive support but can be used standalone with just `CompletableFuture` as well, effectively addressing this ticket here too.

PR #30122, opened 2023-03-15
CaffeineCacheManager supports async loading value

jizhuozhi:


bclozel:
Sorry, but we don't accept PRs without any rationale nor explanation, especially for new features and new API.

If I understand correctly, this change would turn cache entries into `CompletableFuture<V>`. As outlined in #17559, this is not supported (and it won't be supported by the JRS, see jsr107/jsr107spec#312). If you wish to use this feature, you can use Caffeine directly. We might reconsider this in the future with #17920, but we're missing critical bits right now to work on that.

Also, please run the build locally before submitting a PR - this change is failing the build.
Thanks!

jizhuozhi:
Hello, @bclozel . It's my mistake. I just ran the test but forgot to check the style.

My intention is not to get CompletableFuture to make asynchronous calls, but to be able to use asynchronous API when refreshing the cache (this can avoid executor blocking). 

The problem I have encountered is that when using Spring Cloud LoadBalancer (it uses CaffeineCacheManager to cache service instances), I don't want to get the instance from the registry when it expires, but pull the latest content from the registry regularly And replace the content in the cache (Caffeine provides the ability to refresh regularly). Most registry SDKs provide asynchronous APIs to avoid blocking the calling thread, but now CaffeineCacheManager only supports synchronous loading of values, so I have to convert asynchronous to synchronous and block here, which will affect the update of other cache values.

In fact, because `com.github.benmanes.caffeine.cache.AsyncLoadingCache#synchronous` is used to create a cached view, for the caller of the cache, it is still synchronously obtaining data, and only the IO that occurs during refresh is affected operation only

jhoeller:
While this has been declined initially (and cannot be reopened anymore as a pull request), I'm effectively addressing this along with #17920 now. `CaffeineCacheManager` has a general `setAsyncCacheMode(boolean)` method now, as well as a `setAsyncCacheLoader` method which explicitly sets that flag as well. We not only build internally async caches, we also leverage the Caffeine `AsyncCache` interface for our new `CompletableFuture`-based `retrieve` operations now.

PR #30136, opened 2023-03-18
Optimize some iterations in BodyExtractor and BodyInserter

yuzawa-san:
Using iterators for trivial findFirst and toList cases uses less CPU and memory (in these cases, not while costing readability)
Here is an icicle graph example of the CPU usage before 
image
and here it is after:
image
There was about a 10-15% overhead with the streams usage, which we can save on by doing this.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
@@ -375,7 +375,8 @@ public abstract class BodyInserters {
 		MediaType mediaType = outputMessage.getHeaders().getContentType();
 		for (HttpMessageWriter<?> messageWriter : context.messageWriters()) {
 			if (messageWriter.canWrite(bodyType, mediaType)) {
-				return write(publisher, bodyType, mediaType, outputMessage, context, cast(messageWriter));
+				HttpMessageWriter<Object> typedMessageWriter = cast(messageWriter);
+				return write(publisher, bodyType, mediaType, outputMessage, context, typedMessageWriter);
 			}
 		}
 		return Mono.error(unsupportedError(bodyType, context, mediaType));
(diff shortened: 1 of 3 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/function/BodyInserters.java
```diff
@@ -373,12 +373,12 @@ else if (adapter != null) {
...
 		MediaType mediaType = outputMessage.getHeaders().getContentType();
-		return context.messageWriters().stream()
-				.filter(messageWriter -> messageWriter.canWrite(bodyType, mediaType))
-				.findFirst()
-				.map(BodyInserters::cast)
-				.map(writer -> write(publisher, bodyType, mediaType, outputMessage, context, writer))
-				.orElseGet(() -> Mono.error(unsupportedError(bodyType, context, mediaType)));
+		for (HttpMessageWriter<?> messageWriter : context.messageWriters()) {
+			if (messageWriter.canWrite(bodyType, mediaType)) {
+				return write(publisher, bodyType, mediaType, outputMessage, context, cast(messageWriter));
```
simonbasle, line 378: compared to the `BodyExtractor` case, I find that `cast` call is a little too buried among all these parameters, making it less visible. would you mind extracting it to an intermediate local variable? this should be efficiently optimized away by the JIT compiler, so that is purely from a readability perspective.

simonbasle (changes requested):
indeed these kind of stream usage can get an order of magnitude slower especially for small collections... this looks good to me overall, thanks for the PR @yuzawa-san 👍 
one minor comment to address from a readability standpoint and I think we can merge this in time for 6.0.8

yuzawa-san:
@simonbasle updated with intermediate variable.

https://github.com/spring-projects/spring-framework/pull/29972 also gets around another (I would say more expensive) streams usage in ReadOnlyHttpHeaders.
https://github.com/spring-projects/spring-framework/blob/4e896c8125c957ffa63238d10444f76d9913b549/spring-web/src/main/java/org/springframework/http/ReadOnlyHttpHeaders.java#L150-L156
I left this one there because that PR avoids that entrySet and also I have been working on a branch where I got ReadOnlyHttpHeaders to use `CollectionUtils.unmodifiableMultiValueMap` internally.

simonbasle (approved):
thanks for the change @yuzawa-san 👍

PR #30137, opened 2023-03-18
Lenient handling of unencoded URIs in ForwardedHeaderTransformer

joakimmohn:
Currently `ForwardedHeaderTransformer` assumes the url is url encoded which is not necessarily the case.

I experienced an issue where the a query parameter as part of the OpenID Connect flow contains a "=" character which fails the `verifyUriComponent` check in `HierarchicalUriComponents`.


rstoyanchev (commented):
The URI should be encoded, or otherwise it cannot be parsed reliably. I don't think we should have to make such checks, and this wouldn't be the only place that would require it.

I don't where the root cause lies. It could be that OpenID Connect sends it that way, and it should be fixed there. We also made some recent improvements to how the URI is initialized for Reactor Netty in 6.0.6 and 5.3.26. It's worth trying with those versions if you haven't.

rstoyanchev:
I've solved this in a slightly different way, essentially still avoiding encoding but also not causing a strict check to be performed.

PR #30138, opened 2023-03-18
Use String.equals() in LiteralPathElement

yuzawa-san:
[String.equals()](https://github.com/openjdk/jdk/blob/jdk-17%2B0/src/java.base/share/classes/java/lang/String.java#L1027) eventually calls [StringLatin1.equals()](https://github.com/openjdk/jdk/blob/jdk-17%2B0/src/java.base/share/classes/java/lang/StringLatin1.java#L94) which uses processor-specific intrinsics. They are available for major architectures and typically use some sort of processor-optimized vector operations, which should be faster than the current for-loop.

Before:
image

After:
image

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/pattern/LiteralPathElement.java
+++ b/spring-web/src/main/java/org/springframework/web/util/pattern/LiteralPathElement.java
@@ -31,6 +31,8 @@ class LiteralPathElement extends PathElement {
 
 	private final char[] text;
 
+	private final String textString;
+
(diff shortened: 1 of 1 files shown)


poutsma:
I changed the PR to remove the `char[]` field in favor of the `String`, as both contain the same information.

I then ran our JMH `PathMatchingBenchmark` against this PR, and it does look like it increases the operations/second, while not significantly increasing memory usage. So I am going to merge this PR for 6.0.8.

PR #30139, opened 2023-03-18
Cache ServerHttpRequest::getMethod in AbstractServerHttpRequest

yuzawa-san:
ServerHttpRequest.method() is called twice on each HttpMethodPredicate and within CorsUtils.isPreFlightRequest(). Each call has to do a roundtrip all the way into netty and it does a bunch of String hashing to resolve the proper Spring HttpMethod. I moved this work to be done once upon request construction since the field read is cheaper.


poutsma:
Thanks for leaving a PR.

It seems only fair that we introduce the same enhancement for the other `ServerHttpRequest` implementations, so I moved your changes to `AbstractServerHttpRequest`.

PR #30156, opened 2023-03-21
Fixed circular reference exception using @Async annotation

wangzhengsi:
### Problem description

Using the `@Async` annotation when `beanA` and `beanB` are circularly referenced will throw an error.

### Reasons

The bean with `@Async` annotation is created as a proxy instance during initialization, and the bean that is retrieved from the cache of the earlier singleton is still the original instance.

### Solution

When the bean is fetched from the cache of the earlier singleton, it will be created as a proxy instance. The initialization phase will be checked and processing will be skipped if the proxy instance has already been created.

### Source Code

- [BeanA](https://github.com/wangzhengsi/0321Code/blob/master/src/main/java/com/tempura/test/service/BeanA.java)

- [BeanB](https://github.com/wangzhengsi/0321Code/blob/master/src/main/java/com/tempura/test/service/BeanB.java)

### Error message

```
org.springframework.beans.factory.BeanCurrentlyInCreationException: Error creating bean with name 'beanA': Bean with name 'beanA' has been injected into other beans [beanB] in its raw version as part of a circular reference, but has eventually been wrapped. This means that said other beans do not use the final version of the bean. This is often the result of over-eager type matching - consider using 'getBeanNamesOfType' with the 'allowEagerInit' flag turned off, for example.
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:622) ~[spring-beans-5.1.9.RELEASE.jar:5.1.9.RELEASE]

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/AbstractAdvisingBeanPostProcessor.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/AbstractAdvisingBeanPostProcessor.java
@@ -16,7 +16,9 @@
 
 package org.springframework.aop.framework;
 
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.springframework.aop.Advisor;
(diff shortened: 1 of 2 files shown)


sbrannen (changes requested):
Please introduce a test that fails before the proposed changes and passes afterwards.

wangzhengsi:
Start the project to see the error message
[https://github.com/wangzhengsi/0321Code](url)

PR #30161, opened 2023-03-22
Handle all exceptions for stored proc output param retrieval in `SharedEntityManagerCreator`

### LINKED ISSUE #30160: `SharedEntityManagerCreator` does not handle all exceptions for stored proc output param retrieval
**Affects:** Spring Boot 6.0.4

In the `SharedEntityManagerCreator` class, the following code block invokes `getOutputParameterValue` which can throw a `HibernateException`.

https://github.com/spring-projects/spring-framework/blob/800b13492b627d02da5d3591ee9ae9c04833da81/spring-orm/src/main/java/org/springframework/orm/jpa/SharedEntityManagerCreator.java#L424-L435

The problem is that the catch block only handles `IllegalArgumentException`.

This causes a connection leak in the HikariCP pool because the subsequent `EntityManagerFactoryUtils.closeEntityManager(this.entityManager);` line will never be executed, and the entity manager will never be closed.

A possible solution is to change the catch block to handle the more general `Exception` or `RuntimeException` instead of `IllegalArgumentException`.

giuseppemilicia:
- Fixes #30160

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/SharedEntityManagerCreator.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/SharedEntityManagerCreator.java
@@ -430,7 +430,7 @@ public abstract class SharedEntityManagerCreator {
 									entry.setValue(storedProc.getOutputParameterValue(key.toString()));
 								}
 							}
-							catch (IllegalArgumentException ex) {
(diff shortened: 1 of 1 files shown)


sbrannen:
This has been merged into `main` with an appropriate commit message and backported to `5.3.x`.

Thanks

PR #30166, opened 2023-03-22
Replace `Collections.unmodifiableList(new ArrayList(..))` with `List.copyOf()`

ssang1105:
This PR replaces `Collections.unmodifiableList()` with `List.copyOf()`. 
It creates a new immutable list by copying the elements of the given list.
Using `List.copyOf()`  results in more readable and prevents unexpected bugs, and makes the code easier to maintain.


rstoyanchev (commented):
There is one more such case in `DefaultMetadataExtractor`.

ssang1105:
@rstoyanchev  I found and fix it

PR #30168, opened 2023-03-22
Minor refactoring in PayloadMethodArgumentResolver

divcon:
remove duplicate logic

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/support/PayloadMethodArgumentResolver.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/support/PayloadMethodArgumentResolver.java
@@ -136,11 +136,7 @@ public class PayloadMethodArgumentResolver implements HandlerMethodArgumentResol
 
 		Class<?> targetClass = resolveTargetClass(parameter, message);
 		Class<?> payloadClass = payload.getClass();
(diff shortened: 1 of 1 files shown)


spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/support/PayloadMethodArgumentResolver.java
```diff
@@ -151,9 +147,9 @@ public Object resolveArgument(MethodParameter parameter, Message<?> message) thr
 				throw new MessageConversionException(message, "Cannot convert from [" +
 						payloadClass.getName() + "] to [" + targetClass.getName() + "] for " + message);
 			}
-			validate(message, parameter, payload);
-			return (isOptionalTargetClass ? Optional.of(payload) : payload);
 		}
+		validate(message, parameter, payload);
+		return (isOptionalTargetClass ? Optional.of(payload) : payload);
```
s50600822, line 152: I wonder if it'd be more natural to early return when ClassUtils.isAssignable, then the rest don't have to be nested (not sure why it originally nested).

rstoyanchev: But then you would be back to having duplicate logic since validation and `Optional` wrapping is required regardless. Also, conversion is likely the more common case anyway, so it's okay for that to be performed first.

PR #30183, opened 2023-03-24
Fix example in Javadoc for `@EnableWebSocket`

ghostg00:
The echoWebSocketHandler method is not in the WebSocketConfigurer interface. It should be the @Bean annotation


sbrannen:
Hi @ghostg00,

Good catch, and congratulations on submitting your first PR to the Spring Framework! 👍 

This has been merged into `main` and backported to `5.3.x`.

Thanks

d799907825:
good job!

PR #30189, opened 2023-03-25
Optimize array creation in SpEL `ConstructorReference`

TAKETODAY:
- Create primitive arrays directly instead of using `Array#newInstance`
- Replace if-else blocks with enhanced switch statement


TAKETODAY:
Thank you

PR #30192, opened 2023-03-25
Improve performance of canRead() in HttpMessageReader's

yuzawa-san:
* A lot of the decoders use MimeType.ALL, so use MimeType.WILDCARD_TYPE for faster String.equals() using reference comparisons, specifically in MimeType.isWildcardType().
* Move cheaper checks to the front of the canRead implementations. The media/mime checks appear to be cheaper than Class.isAssignableFrom which is cheaper than the ResolvableType checks, so putting them in the correct order (while preserving the and/or logic) gives an improvement. 

before icicle graph for CPU from my [stress test](https://gist.github.com/yuzawa-san/5102be26fb9e4472419c23b39c4524b2). this is for a JSON payload, thus the other readers should be skipped and take as little cpu as possible:
image

after (see how MultipartHttpMessageReader and FormHttpMessageReader have shrunk):
image

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/MimeType.java
+++ b/spring-core/src/main/java/org/springframework/util/MimeType.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -268,7 +268,8 @@ public class MimeType implements Comparable<MimeType>, Serializable {
 	 * @return whether the subtype is a wildcard
 	 */
 	public boolean isWildcardSubtype() {
-		return WILDCARD_TYPE.equals(getSubtype()) || getSubtype().startsWith("*+");
+		String subtype = getSubtype();
+		return WILDCARD_TYPE.equals(subtype) || subtype.startsWith("*+");
 	}
 
 	/**
--- a/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
@@ -181,7 +181,7 @@ public abstract class MimeTypeUtils {
 
 	static {
 		// Not using "parseMimeType" to avoid static init cost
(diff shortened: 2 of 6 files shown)


spring-core/src/main/java/org/springframework/util/MimeType.java
```diff
@@ -268,7 +268,8 @@ public boolean isWildcardType() {
 	 * @return whether the subtype is a wildcard
 	 */
 	public boolean isWildcardSubtype() {
-		return WILDCARD_TYPE.equals(getSubtype()) || getSubtype().startsWith("*+");
+		String subtype = getSubtype();
+		return WILDCARD_TYPE.equals(subtype) || subtype.startsWith("*+");
```
rstoyanchev, line 272: Aside from extracting a variable, how does this improve performance?

spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
```diff
@@ -181,7 +181,7 @@ public abstract class MimeTypeUtils {
 
 	static {
 		// Not using "parseMimeType" to avoid static init cost
-		ALL = new MimeType("*", "*");
+		ALL = new MimeType(MimeType.WILDCARD_TYPE, MimeType.WILDCARD_TYPE);
```
rstoyanchev, line 184: This is just using constants instead of String literals, and it should be invoked only once as a static initializer. I don't see how this improves performance.

yuzawa-san: It allows the reference check to work https://github.com/openjdk/jdk/blob/jdk-17%2B35/src/java.base/share/classes/java/lang/String.java#L1822

spring-core/src/main/java/org/springframework/util/MimeType.java
yuzawa-san: see response in MimeTypeUtils

rstoyanchev: Yes I see it there, and thanks for clarifying, but the change seems unrelated. Whether the subtype is accessed through a local variable or not, it is the same String. Am I missing anything?

PR #30200, opened 2023-03-27
Add an error handler and an interceptor for RestTemplate

ZouHongxue:
### Why Have This PR?
In daily development, functions are quite common. It's a default implementation also an example.

1. Do not throw the exception, check the HTTP status outside. [Reference](https://stackoverflow.com/questions/38093388/spring-resttemplate-exception-handling)
2. Log the response especially when an exception happens. [Reference](https://stackoverflow.com/questions/7952154/spring-resttemplate-how-to-enable-full-debugging-logging-of-requests-responses)


ZouHongxue:
Noted with thanks.

PR #30207, opened 2023-03-27
Support Named Parameters in Stored Procedures

### LINKED ISSUE #9084: Support named parameters in CallableStatementCreator implementation [SPR-4406]
**[Mike Calmus](https://jira.spring.io/secure/ViewProfile.jspa?name=mcalmus)** opened **[SPR-4406](https://jira.spring.io/browse/SPR-4406?redirect=false)** and commented

Named parameters have been added for some operations, but the execution of a CallableStatement does not support named parameters.

The following code block from CallableStatementFactory.CallableStatementCreatorImpl would be affected or need to be extended:

```
int sqlColIndx = 1;
for (int i = 0; i < declaredParameters.size(); i++) {
	SqlParameter declaredParameter = (SqlParameter) declaredParameters.get(i);
	if (!declaredParameter.isResultsParameter()) {
		// So, it's a call parameter - part of the call string.
		// Get the value - it may still be null.
		Object inValue = this.inParameters.get(declaredParameter.getName());
		if (declaredParameter instanceof ResultSetSupportingSqlParameter) {
			// It's an output parameter: SqlReturnResultSet parameters already excluded.
			// It need not (but may be) supplied by the caller.
			if (declaredParameter instanceof SqlOutParameter) {
				if (declaredParameter.getTypeName() != null) {
					cs.registerOutParameter(sqlColIndx, declaredParameter.getSqlType(), declaredParameter.getTypeName());
				}
				else {
					if (declaredParameter.getScale() != null) {
						cs.registerOutParameter(sqlColIndx, declaredParameter.getSqlType(), declaredParameter.getScale().intValue());
					}
					else {
						cs.registerOutParameter(sqlColIndx, declaredParameter.getSqlType());
					}
				}
				if ((declaredParameter).isInputValueProvided() || inValue != null) {
					StatementCreatorUtils.setParameterValue(csToUse, sqlColIndx, declaredParameter, inValue);
				}
			}
		}
		else {
			// It's an input parameter- must be supplied by the caller.
			if (!this.inParameters.containsKey(declaredParameter.getName())) {
				throw new InvalidDataAccessApiUsageException(
						"Required input parameter '" + declaredParameter.getName() + "' is missing");
			}
			StatementCreatorUtils.setParameterValue(csToUse, sqlColIndx, declaredParameter, inValue);
		}
		sqlColIndx++;
	}
}
```

---

**Issue Links:**
- #17398 Support for named parameters in SimpleJdbcCall
- #8193 StoredProcedure enhancement to allow parameters to be optional

10 votes, 8 watchers

marschall:
StoredProcedure binds by name or index depending on the method called.

The behavior after this PR is:
- if `#execute` is called with an array then indexed bind variables are used
- if `#execute` is called with a `Map` then named bind variables are used

This would be my expectation from reading the API.

The code has a couple of limitations:
- there is unfortunately quite a bit of copy and paste
- `SqlValue` and `SqlTypeValue` are not supported for named binds, they would need a `#setValue` or `#setTypeValue` method respectively that takes a parameter name rather than a parameter index

Existing users will get different behavior if they are calling the `Map` based method. This may break if:
- the supplied parameter names do not match the parameter names of the procedure
- if the driver does not support named parameters
- they use `SqlValue` or `SqlTypeValue`

Closes gh-9084

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/CallableStatementCreatorFactory.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/CallableStatementCreatorFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -117,6 +117,14 @@ public class CallableStatementCreatorFactory {
 	}
 
 	/**
+	 * Return a new CallableStatementCreator instance given these parameters.
+	 * @param params list of parameters (may be {@code null})
+	 */
+	public CallableStatementCreator newNamedCallableStatementCreator(@Nullable Map<String, ?> params) {
+		return new NamedCallableStatementCreatorImpl(params != null ? params : new HashMap<>());
+	}
+
+	/**
 	 * Return a new CallableStatementCreator instance given this parameter mapper.
 	 * @param inParamMapper the ParameterMapper implementation that will return a Map of parameters
 	 */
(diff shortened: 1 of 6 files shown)


jhoeller:
Since our entire JDBC infrastructure is based on indexed parameters, I do not see us supporting named bindings for stored procedures at the JDBC level, exactly for the reason that it leads to CallableStatement-specific duplication of quite a bit of code that can othewise be reused with PreparedStatement-based code paths.

The `execute(Map)` support in `StoredProcedure` is meant to match local parameter names declared via `declareParameter`, possibly matching database-level parameter names but not necessarily so. Supporting database parameter names directly would allow for arbitrary ordering at the Java declaration level, I suppose, but this is arguably not a significant enough benefit to justify the processing complexity within Spring's JDBC support. Since the SQL types are important as well, a well-ordered sequence of `declareParameter` calls with name plus SQL type does seem quite appropriate in any case, aligned with the database-level declaration but explicitly declared with the same ordering at the Java level.

From that perspective, I'd rather close this PR. Thanks for your efforts, in any case!

marschall:
Fair enough should we then also close #9084 ?

PR #30218, opened 2023-03-28
Remove `String.formatter()` from DefaultServerRequestObservationConvention

yuzawa-san:
The formatter is overkill since it has to parse the format string each time and we are only doing concatenation here. I see String.format is use elsewhere, typically in exception and debug log calls which is ok, but this was one notable place found in [my stress test](https://gist.github.com/yuzawa-san/5102be26fb9e4472419c23b39c4524b2) with actuator enabled.

before cpu:
image

after cpu (most of the time is spend in toLowerCase since the concat is cheaper than the String.formatter()):
image


bclozel:
Thanks a lot @yuzawa-san !

Issue #26072, opened 2020-11-11
`SimpleJdbcCall` with `namedBinding=true` for MS SQL Server


iignateva:
spring version: 5.0.8.RELEASE

I try to use `setNamedBinding(true)` on `SimpleJdbcCall` and get the generated statement in oracle syntax (`call @P0 = proc_name Param1 => @P1, Param2=> @P2 ...`).

I found this comment: https://github.com/spring-projects/spring-framework/issues/17398#issuecomment-453423378 

Was it implemented for the MS SQL Server?

jhoeller:
MS SQL Server seems to use the same `@<parameter> = <value>` syntax as Sybase, so we should be able to deal with this along with #30231.

PR #30231, opened 2023-03-29
Fix `SimpleJdbcCall` named parameter binding for Sybase ASE

giuseppemilicia:
This commit fixes an issue with Sybase named parameter binding in the code. The `namedParamBindingToUse` method has been updated to properly handle named parameters in Sybase stored procedures. Prior to this fix, the method may not have returned the correct named parameter for a given parameter name, causing issues with stored procedure execution.

This fix is necessary to ensure proper execution of Sybase stored procedures that use named parameters.

When working with Sybase ASE stored procedures, parameters should be supplied in the order of their create procedure statement, unless they are supplied in the form `@<parameter> = <value>`, in which case they can be supplied in any order. For more information, see the Sybase ASE documentation: https://help.sap.com/docs/SAP_ASE/b65d6a040c4a4709afd93068071b2a76/aa8728babc2b101491a4f578b9082925.html?locale=en-US


snicoll:
I wonder if that PR could also help with #26072

PR #30238, opened 2023-03-30
Rename MockMVC matcher methods to prevent regression in user tests

### LINKED ISSUE #30220: Wrong MockRestRequestMatchers.header() method in spring-test being invoked (JDK issue?)
**Affects:** 2.7.10

---

Hello everyone.

I'm uncertain if this is an issue with spring-test or with Java compiler itself but, from Spring Boot 2.7.10 (which incorporates spring-test 5.3.26), we have been having an odd issue with tests calling the wrong `MockRestRequestMatchers.header()` method.

In spring-test 5.3.25 there were a couple of `header()` methods, but the one that interest us is the one with this signature:

```public static RequestMatcher header(String name, Matcher<? super String>... matchers)```

this was used in testing like this:

```.andExpect(header("X-MY-HEADER", equalTo("myValue")))```

where `equalTo()` is defined in Hamcrest as

```public static <T> Matcher<T> equalTo(T operand)```

this worked normally.

From Spring Boot 2.7.10 (spring-test 5.3.26) a new `header()` method has been added with this signature:

```public static RequestMatcher header(String name, Matcher<? super List<String>> matcher)```

and this is where things are breaking: the same call above is now entering this new method, as opposed to the previous. I find this surprising because the signature doesn't match, `equalTo("myValue")` returns `Matcher<String>`, which doesn't align to `Matcher<? super List<String>>`, and yet it's coming in. If I change the call to `header()` to explicitly call out the type like this:

```.andExpect(header("X-MY-HEADER", CoreMatchers.<String>equalTo("myValue")))```

then things go back to normal and the call is again directed to the old `header()` method.

I'm guessing that, via type erasure, the signatures of the two `header()` methods are being simplified to just `header(String, Matcher)` and, there being two possible candidates that can match the call, Java is just picking the first one. I'm not sure what could be done from spring-framework's perspective, though.

Thoughts?

simonbasle:
This commit changes the name of two recently introduced methods in the
`MockRestRequestMatchers` class for header and queryParam. These have
been found to cause false negatives in user tests, due to the new
overload taking precedence in some cases.

Namely, using a `Matcher` factory method which can apply to both `List`
and `String` will cause the compiler to select the newest list overload,
by instantiating a `Matcher<Object>`.

This can cause false negatives in user tests, failing tests that used
to pass because the Matcher previously applied to the first String in
the header or queryParam value list. For instance, `equalsTo("a")`.

The new overloads are recent enough and this has enough potential to
cause an arbitrary number of user tests to fail that we break the API
to eliminate the ambiguity, by renaming the methods with a `*List`
suffix.

Closes gh-30220
Closes gh-30238
See gh-29953
See gh-28660

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/MockRestRequestMatchers.java
@@ -125,11 +125,11 @@ public abstract class MockRestRequestMatchers {
 	 * @param name the name of the query parameter whose value(s) will be asserted
 	 * @param matcher the Hamcrest matcher to apply to the entire list of values
 	 * for the given query parameter
-	 * @since 5.3.26
+	 * @since 5.3.27
 	 * @see #queryParam(String, Matcher...)
 	 * @see #queryParam(String, String...)
 	 */
-	public static RequestMatcher queryParam(String name, Matcher<? super List<String>> matcher) {
+	public static RequestMatcher queryParamList(String name, Matcher<? super List<String>> matcher) {
 		return request -> {
 			MultiValueMap<String, String> params = getQueryParams(request);
 			List<String> paramValues = params.get(name);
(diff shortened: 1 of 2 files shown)


sbrannen (approved):

simonbasle:
Backport issue: #30235

Issue #29861, opened 2023-01-20
ContentDisposition.toString() should include both regular and extended filename parameter


arjohn-telecats:
The Content-Disposition header generated by ContentDisposition.toString() includes either the regular `filename` parameter or the extended `filename*` parameter, depending on the value of the `charset` attribute. For backwards compatibilities with older browser it would be preferable to _always_ include the regular parameter, in addition to the extended parameter. This also seems to be what RFC 6266 suggests:

RFC 6266, section 4.3:
_Many user agent implementations predating this specification do not understand the "filename*" parameter.  Therefore, when both "filename" and "filename*" are present in a single header field value, recipients SHOULD pick "filename*" and ignore "filename". This way, senders can avoid special-casing specific user agents by sending both the more expressive "filename*" parameter, and the "filename" parameter as fallback for legacy recipients (see [Section 5] for an example)._

RFC 6266, section 5 (Examples):
_This example is the same as the one above, but adding the "filename" parameter for compatibility with user agents not implementing [RFC 5987]:_
```
     Content-Disposition: attachment;
                          filename="EURO rates";
                          filename*=utf-8''%e2%82%ac%20rates
```
_Note: Those user agents that do not support the [RFC 5987] encoding ignore "filename*" when it occurs after "filename"._

poutsma:
While having both parameters sounds like a good idea, I am having difficulty seeing how we can transform any given String into an US-ASCII string for the `filename` parameter. For instance, given the filename `€ rates`, I don't see how we can transliterate '€' into `EUR`, as seen in the example.

arjohn-telecats:
Transliterating this is way out of scope for the ContentDisposition class imho. Options that might work for the regular attribute:
1) Simple url-encode any non-ascii characters (using utf8?) and hope that the browser decodes it in the same way.
2) Replace any non-ascii characters with a replacement character, an underscore for example.
3) Extended the API and allow developers to set both attributes separately.

The first option works for me with the latest Chrome and Firefox releases.

poutsma:
I decided to resolve this by encoding any non-ASCII characters as quoted-printable (RFC 2047), as that seems to be the most common way to do so. Moreover, we already had decoding support for it.

arjohn-telecats:
[...]

PR #30252, opened 2023-03-31
Quote '?' in Quoted Printable filename

luozhenyu:
# Bad case
```java
String filename = "filename with ?问号.txt";
ContentDisposition cd = ContentDisposition.attachment()
    .filename(filename, StandardCharsets.UTF_8)
    .build();
String[] parts = cd.toString().split("; ");

// attachment; filename="=?UTF-8?Q?filename=20with=20?=E9=97=AE=E5=8F=B7.txt?="
String quotedPrintableFilename = parts[0] + "; " + parts[1];
// expected: filename with ?问号.txt
// actual: filename with
System.out.println(ContentDisposition.parse(quotedPrintableFilename).getFilename());
```

# Cause
In previous implementation, the question mark (?) is not quoted. If a quoted character (=E9=97=AE) follows, the combination "?=" will be interpreted as end mark of "encoded-word".

> 8-bit values which correspond to printable ASCII characters other than "=", "?", and "_" (underscore), MAY be represented as those characters.
> https://www.rfc-editor.org/rfc/rfc2047#section-4.2

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
+++ b/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
@@ -599,7 +599,7 @@ public final class ContentDisposition {
 	}
 
 	private static boolean isPrintable(byte c) {
-		return (c >= '!' && c <= '<') || (c >= '>' && c <= '~');
+		return (c >= '!' && c <= '<') || (c >= '@' && c <= '~') || c == '>';
 	}
 
 	private static String encodeQuotedPairs(String filename) {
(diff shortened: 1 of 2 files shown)


rstoyanchev (commented):
This seems related to the recent change in #29861 and the added regular `"filename"` parameter. Prior to that for `filename(..., UTF_8)` we created only a `"filename*"` parameter via `encodeRfc5987Filename`, and that part of the test works.

poutsma:
Thank you for submitting a PR. I have made a couple of changes also related to RFC 2047, so that spaces are now encoded as underscores, and that underscores themselves are handled as non-printable.

Issue #30256, opened 2023-03-31
PostgreSQL stored procedure call fails due escaping call with curly braces


venkateshguduru:
Hi Team, I am facing an issue while invoking the PostgreSQL stored procedure call, it fails due escaping call with curly braces.

**Problem Statement**

When invoking the PostgreSQL stored procedure, it returns error:

```
bad SQL grammar [{call sample_sp(?, ?)}]; nested exception is org.postgresql.util.PSQLException
```

**Root cause**

PostgreSQL doesn't support/need escaping call with curly braces, need to make it conditional.

**Issue class**

https://github.com/spring-projects/spring-framework/blob/main/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/CallMetaDataContext.java#L645-L668

**Fix**

Add an `if` condition to add curly braces only for non PostgreSQL calls.

```java
if(this.metaDataProvider instanceof PostgresCallMetaDataProvider) {
	callString = new StringBuilder("call ");
}else {
	callString = new StringBuilder("{call ");
}
```

sbrannen:
Closing as superseded by #30257.

Please refrain from creating an issue **and** a PR for the same topic. The PR is sufficient.

Thanks

snicoll:
Reopening as the PR has been declined

jhoeller:
This seems to be the case outlined here: https://stackoverflow.com/questions/65116924/x-is-a-procedure-use-call-when-i-am-already-using-call
In order to make PostgreSQL compliant with standard JDBC syntax for stored procedures and not just functions, you may specify `escapeSyntaxCallMode=callIfNoReturn` as a JDBC driver property. Not sure why that is not the driver's default but it is the recommended solution to this problem.

From the Spring side, we are going to close this issue and stick with standard JDBC syntax.

PR #30257, opened 2023-03-31
Support PostgreSQL stored procedure `call` syntax in `CallMetaDataContext`

venkateshguduru:
Fix for PostgreSQL stored procedure call fails due escaping call with curly braces,

- Addresses #30256


sbrannen:
Please refrain from creating PRs with a title like `Fix for #30256`.

Instead, use a concise and descriptive title, like the one I've introduced for this PR.

See also: [contributing guidelines](https://github.com/spring-projects/spring-framework/blob/main/CONTRIBUTING.md)

snicoll:
@venkateshguduru the instanceof looks a bit unusual to me. Shouldn't we try to abstract that using a method on `CallMetaDataProvider` that the postgres impl will implement. I am not sure what it should be and whether this is warranted though. The lack of tests is also problematic.

snicoll:
OK I am going to close this as I don't think we should do an instanceof there.

PR #30260, opened 2023-03-31
Fix SameSite attribute in MockMvcHttpConnector

### LINKED ISSUE #30249: MockMvcHttpConnector not populating sameSite attribute in cookie
I'm trying to assert that the `SameSite` attribute is being set correctly using the `MockMvcWebTestClient` but I noticed that it wasn't being populated correctly.

https://github.com/spring-projects/spring-framework/blob/9b38e43c1754563465c275c253fb4b6fd1cf9345/spring-test/src/main/java/org/springframework/test/web/servlet/client/MockMvcHttpConnector.java#L192-L202

The actual cookie in question is of type `MockCookie`

https://github.com/spring-projects/spring-framework/blob/19cf5035343922111e7c03b60933bf178170f9d4/spring-test/src/main/java/org/springframework/mock/web/MockCookie.java#L80-L89

I think that the `MockMvcHttpConnector` should be updated to use the new Servlet 6 `Cookie.getAttribute` to get the `SameSite` attribute.

At the same time I notice that the `MockCookie` doesn't have the `SameSite` attribute in the attribute map. I think the Cookie implementation on Servlet 6 generally delegate the setters to `setAttributeInternal`. So I think that `MockCookie` also needs a corresponding update to set the `SameSite` as an attribute.

justin-tay:
Fixes #30249

* Modified `MockCookie` to populate attributes using `setAttribute`.
* Modified `MockMvcHttpConnector` to set `sameSite` using `Cookie.getAttribute`
* Added tests

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockCookie.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockCookie.java
@@ -39,15 +39,13 @@ import org.springframework.util.StringUtils;
 @SuppressWarnings("removal")
 public class MockCookie extends Cookie {
 
-	private static final long serialVersionUID = 4312531139502726325L;
+	private static final long serialVersionUID = 1198809317225300389L;
 
+	private static final String SAME_SITE = "SameSite";
+	private static final String EXPIRES = "Expires";
 
 	@Nullable
 	private ZonedDateTime expires;
-
-	@Nullable
-	private String sameSite;
-
 
 	/**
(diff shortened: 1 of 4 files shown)


sbrannen:
Hi @justin-tay,

Could you please split this PR into two separate PRs to address these two topics?

1. Update `MockCookie` to make use of Servlet 6.0 APIs and semantics for "attributes".
2. Support `SameSite` attribute in `MockMvcHttpConnector`.

sbrannen:
Superseded by:

- #30263
- #30264

PR #30271, opened 2023-04-03
Specify initial capacity when creating `ArrayList` in `SpringFactoriesLoader`

wizard4wu:
When the ArrayList is created firstly, provides the capacity for the list. 

It can reduce unnecessary space.


sbrannen:

Thanks

PR #30294, opened 2023-04-06
Fix comparison of `title` in `equals()` and `hashCode()` of `ProblemDetail`

quaff:
Lazy computed title property should be taken into account

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ProblemDetail.java
+++ b/spring-web/src/main/java/org/springframework/http/ProblemDetail.java
@@ -42,6 +42,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Rossen Stoyanchev
  * @author Juergen Hoeller
+ * @author Yanming Zhou
  * @since 6.0
  * @see <a href="https://datatracker.ietf.org/doc/html/rfc7807">RFC 7807</a>
  * @see org.springframework.web.ErrorResponse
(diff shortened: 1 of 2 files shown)


simonbasle (changes requested):
thanks for the PR @quaff but what's the rationale behind this?

using the `getTitle()` over `this.title` isn't that much of an improvement because if the field is not set, the getter derives the result from the `this.status` field (namely, the reason phrase of the enum which is constant). since `this.status` already participates in both `equals()` and `hashcode()`, this change doesn't add new information to these methods.

it even arguably can make them less performant, because now instead of a simple null equality + int comparison, it adds resolution of an enum from the status code + getting the enum's reason phrase to the mix.

so overall I don't see the benefit of this change.

quaff:
> thanks for the PR @quaff but what's the rationale behind this?
> 
> using the `getTitle()` over `this.title` isn't that much of an improvement because if the field is not set, the getter derives the result from the `this.status` field (namely, the reason phrase of the enum which is constant). since `this.status` already participates in both `equals()` and `hashcode()`, this change doesn't add new information to these methods.
> 
> it even arguably can make them less performant, because now instead of a simple null equality + int comparison, it adds resolution of an enum from the status code + getting the enum's reason phrase to the mix.
> 
> so overall I don't see the benefit of this change.

You can review the test I added, currently the `ProblemDetail` is not equal after serialize then deserialize.
The actual use case is:
```java
		ResponseEntity<ProblemDetail> resp = this.testRestTemplate.exchange(
				RequestEntity.method(HttpMethod.GET, path).header(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON_VALUE).build(),
				ProblemDetail.class);
		ProblemDetail expected = ProblemDetail.forStatusAndDetail(HttpStatus.CONFLICT, "Conflicts arise");
		expected.setTitle(HttpStatus.CONFLICT.getReasonPhrase());
		expected.setInstance(URI.create(path));
		assertThat(resp.getBody()).isEqualTo(expected);
```
Title must be explicit set before comparison, but it is implicit at server side, that's not idiomatic.

rstoyanchev (commented):
Yes that makes sense to use `getTitle()` to get a consistent value.

PR #30295, opened 2023-04-06
Fix DisposableBeanAdapter throws NPE

edfeff:
This bug was previously found in the #10879 , and Jhoeller has already fixed it once.
However, it is still reproduced  in spring-framework v6.0.7，I am trying to fix this bug。

ref: https://github.com/spring-projects/spring-framework/issues/10879


snicoll:
@edfeff unfortunately, this is ignoring the problem rather than fixing the cause. I am afraid we can't accept this as a solution. If you're experiencing such a NPE with `6.0.7`, please open an issue with a small sample that we can use to reproduce it. Thanks!

PR #30300, opened 2023-04-07
Check ResponseStatusException reason as MessageSource code for ProblemDetail

### LINKED ISSUE #30222: Consider using reason as detailMessageCode for ResponseStatusException
Please see https://github.com/spring-projects/spring-boot/issues/34791

quaff:
detail in ProblemDetail will be localized
Closes GH-30222

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/ErrorResponse.java
+++ b/spring-web/src/main/java/org/springframework/web/ErrorResponse.java
@@ -40,6 +40,7 @@ import org.springframework.lang.Nullable;
  * {@code @RestController} or {@code RestControllerAdvice} class.
  *
  * @author Rossen Stoyanchev
+ * @author Yanming Zhou
  * @since 6.0
  * @see ErrorResponseException
  */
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/web/server/ResponseStatusException.java
```diff
@@ -132,4 +134,11 @@ public String getMessage() {
 		return getStatusCode() + (this.reason != null ? " \"" + this.reason + "\"" : "");
 	}
 
+	@Override
+	public String getDetailMessageCode() {
+		if (StringUtils.hasLength(this.reason) && this.getClass() == ResponseStatusException.class) {
```
quaff, line 139: It's arguable to apply changes to subclasses or not.

rstoyanchev (commented):
The problem is that `reason` may or may not be a message code, and this change suppresses the default message code determination. Is there a reason you can't use the default message code, i.e. `problemDetail.org.springframework.web.ErrorResponseException`, or otherwise creating a dedicated sub-class also allows you to set the message code and arguments?

quaff:
> GH-30222

I'm trying to fix the inconsistency in boot https://github.com/spring-projects/spring-boot/issues/34791

rstoyanchev:
The referenced Boot issue was closed because you can't fix any of this on the Boot side, so I don't know what you mean. Here I'm saying that the PR in its present form can't go forward because it assumes the reason is a message code, without any option to change that, also suppressing the ability to use the same message code as for all other exceptions. 

Whatever the solution, it would need to be much more explicit about passing a message code, but again I'm wondering whether switching to the default message code would work for you.

quaff:
@rstoyanchev I updated the commit that will break nothing just enhancement, please review it.

quaff:
@rstoyanchev Could you take a look?

rstoyanchev:
Apologies for the delay. Yes, if used as a fallback it should work.

PR #26761, opened 2021-04-05
Allow `PropertySourcesPlaceholderConfigurer` subclass to customize `PropertyResolver`

1zg12:
## Background

Prior to Spring 5.2, when we need to customize the PropertiesLoader, we can subclass `PropertyPlaceholderConfigurer`, and override the `resolvePlaceholder` method.

Now with `PropertySourcesPlaceholderConfigurer` deprecated `PropertyPlaceholderConfigurer`, the `resolvePlaceholder` method has been moved out of the `PlaceHolderConfigurer` and into a separate `PropertyResolver`.

Somehow the `PropertyResolver` is hardcoded [*]: https://github.com/spring-projects/spring-framework/blob/01bea34569cd3dd00e8691a08ab1a2fc53d6c13e/spring-context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java#L157

Due to the hardcode, the only way left for developers to override the `resolvePlaceholder` now becomes to duplicate the whole `postProcessBeanFactory` method and then provide custom `PropertyResolver`.

## Changes

Instead of hardcoding a `PropertyResolver` as existing and make it very difficult to customize, this PR will point to a factory method to provide the `PropertyResolver` as needed.

I believe there is no side effect on this. It only adds the flexibility for customization which has been lost.

#### Note [*]

Even though `PropertyPlaceholderConfigurer` was also using a hardcoded class, https://github.com/spring-projects/spring-framework/blob/01bea34569cd3dd00e8691a08ab1a2fc53d6c13e/spring-beans/src/main/java/org/springframework/beans/factory/config/PropertyPlaceholderConfigurer.java#L210

It's a private inner class, which even doesn't allow override. However, after several redirects, it eventually points to the `resolvePlaceholder` of the current class implementation
https://github.com/spring-projects/spring-framework/blob/01bea34569cd3dd00e8691a08ab1a2fc53d6c13e/spring-beans/src/main/java/org/springframework/beans/factory/config/PropertyPlaceholderConfigurer.java#L250

PR #30304, opened 2023-04-07
Allow `PropertySourcesPlaceholderConfigurer` to customize `PropertyResolver`

terminux:
## Overview

Currently, `PropertySourcesPlaceholderConfigurer` only uses `PropertySourcesPropertyResolver` by default, and cannot customize the resolver.

For example, at the Spring Boot level, it may be necessary to replace `PropertySourcesPropertyResolver` with [ConfigurationPropertySourcesPropertyResolver](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/ConfigurationPropertySourcesPropertyResolver.java) to support short-circuiting.

## Related Issues

- spring-projects/spring-boot#28687
- spring-projects/spring-boot#34900


snicoll:
Thanks for the PR @terminux. We've fixed this in the meantime, see https://github.com/spring-projects/spring-framework/pull/26761

Issue #30285, opened 2023-04-04
Introduce assertions against `Cookie` attributes in `CookieResultMatchers` for `MockMvc`


sbrannen:
## Overview

Servlet API 6.0 introduced `jakarta.servlet.http.Cookie.getAttribute(String)` for accessing attributes other than the standard attributes such as `Max-Age`, `Secure`, etc.

However, there is currently no way to assert the values of such attributes in MockMvc. For example, you cannot assert the value of the `SameSite` attribute.

In contrast, `org.springframework.test.web.reactive.server.CookieAssertions.sameSite(String, String)` provides first-class support for asserting against the `SameSite` attribute when using `WebTestClient`.

## Related Issues

- #30264

## Deliverables

- [x] Introduce assertions against `Cookie` attributes in `CookieResultMatchers` for `MockMvc`.
- [x] Consider introducing a dedicated assertion for the `SameSite` cookie attribute.
- [x] Add DSL support in Kotlin in `CookieResultMatchersDsl.kt`

PR #30308, opened 2023-04-10
Add since tags to sameSite() and attribute() in CookieResultMatchersDsl

izeye:
This PR adds `@since` tags to the new `sameSite()` and `attribute()` functions in the `CookieResultMatchersDsl`.

See gh-30285


snicoll:
Thanks again Johnny!

PR #30318, opened 2023-04-11
Update documentation to clarify Elvis operator semantics

mmanashirov:
Updated ternary expression that could be misleading for those unfamiliar with Groovy's Elvis operator semantics.
Also added a caution snippet to distinguish the varying interpretations and reinforce the expected behavior in SpEL.

### DIFF
--- a/framework-docs/src/docs/asciidoc/core/core-expressions.adoc
+++ b/framework-docs/src/docs/asciidoc/core/core-expressions.adoc
@@ -1702,13 +1702,13 @@ ternary operator.
 
 The Elvis operator is a shortening of the ternary operator syntax and is used in the
 https://www.groovy-lang.org/operators.html#_elvis_operator[Groovy] language.
-With the ternary operator syntax, you usually have to repeat a variable twice, as the
+With the ternary operator syntax, you usually have to repeat a variable multiple times, as the
 following example shows:
 
 [source,groovy,indent=0,subs="verbatim,quotes"]
 ----
 	String name = "Elvis Presley";
-	String displayName = (name != null ? name : "Unknown");
+	String displayName = (name != null && name.length() > 0) ? name : "Unknown";
 ----
 
 Instead, you can use the Elvis operator (named for the resemblance to Elvis' hair style).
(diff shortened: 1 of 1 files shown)


framework-docs/src/docs/asciidoc/core/core-expressions.adoc
```diff
@@ -1775,6 +1775,11 @@ example shows how to use the Elvis operator in a `@Value` expression:
 This will inject a system property `pop3.port` if it is defined or 25 if not.
 =====
 
+CAUTION: Beware that the behavior of the Elvis operator follows boolean semantics as opposed
+to object reference semantics. Thus, rather than returning the first operand if it is non-null,
+it returns the first operand if it evaluates to true based on the rules that comprise
+https://groovy-lang.org/semantics.html#the-groovy-truth[The Groovy Truth].
+
```
sbrannen, line 1782: That's not entirely true.

SpEL does not honor all of Groovy's "truth" rules.

The SpEL Elvis operator checks for values that are non-null **and** non-empty (for Strings).

simonbasle:
Thanks @mmanashirov for pointing that subtlety. As @sbrannen has pointed out, the wording in the change is not entirely accurate but the concern is valid. I'll treat it more as a ticket than a PR and will come up with a satisfactory wording in a separate commit, closing this as superseded. No more action required on your part 👍

mmanashirov:
Thanks for pointing out that this behavior is specific to Strings. After experiencing a slight burn as a result of this behavior on Strings, I made the false assumption that it consistently behaves according to Groovy's rules for other types as well. You just saved me from another burn in the future! Thank you.

simonbasle:
superseded by #30352

sbrannen:
> Thanks for pointing out that this behavior is specific to Strings. 

You're welcome.

> After experiencing a slight burn as a result of this behavior on Strings, 

I can totally see how you got burned by that: it was an undocumented "feature".

> I made the false assumption that it consistently behaves according to Groovy's rules for other types as well.

Well, you know what they say about assumptions... `<fill in the blank>`

I still recall what my 10th grade history teacher told my class about assumptions. Though I don't think it's appropriate to quote that here. 😇

mmanashirov:
We must have had the same 10th grade history teacher. 😂

PR #30340, opened 2023-04-14
Fix link to releases feed

jstansel:


jstansel:

snicoll:
Good catch, thanks @jstansel.

PR #30343, opened 2023-04-16
Use putIfAbsent in Netty5HeadersAdapter

divcon:
For readability, use the putIfAbsent() method.


yuzawa-san:
https://github.com/spring-projects/spring-framework/pull/29972 is a similar optimization

PR #30361, opened 2023-04-21
Optimize DefaultLifecycleProcessor::stopBeans

aooohan:
Optimize `DefaultLifecycleProcessor` by reducing sorting and refactor `stopBeans` method.

Do this change due to `LifecycleGroup` has been sorted and member in which has same phase, so resort `LifecycleGroupMember`  list to be meaningless.

If any improvements can do, feel free to tell me.


simonbasle (approved):
👍 This is similar to changes made to `startBeans` in https://github.com/spring-projects/spring-framework/pull/25506 and since the `LifecycleGroupMember` `phases` map is created right before iteration, there shouldn't be any change in phase within the group's members

simonbasle:
LGTM. I've added some polish to the javadoc since it was making references to `Comparable` nature.

PR #30365, opened 2023-04-23
Fix link in Javadoc of `PassThroughFilterChain`

kzander91:
The Javadoc in `PassThroughFilterChain` incorrectly says that it implements `FilterConfig`, when in fact it implements `FilterChain`.


sbrannen:

Thanks

And... thanks for updating the "test fixture" copy as well. 👍

Issue #30334, opened 2023-04-13
Refactor ReloadableResourceBundleMessageSource slightly to extend it to read from other file types (json, db) 


Sahilshetye:
I was hoping to refactor and move this line:
`this.cachedProperties.put(filename, propHolder); `

from [line 461](https://github.com/spring-projects/spring-framework/blob/d32027df92479c0e2fbf4502afe71b4e80a2ac41/spring-context/src/main/java/org/springframework/context/support/ReloadableResourceBundleMessageSource.java#L461)  to [line 395](https://github.com/spring-projects/spring-framework/blob/d32027df92479c0e2fbf4502afe71b4e80a2ac41/spring-context/src/main/java/org/springframework/context/support/ReloadableResourceBundleMessageSource.java#L395)

**Rationale:**
This allows me to extend ReloadableResourceBundleMessageSource and override just `refreshProperties()` and `loadProperties()` in the child class to extend reading and loading resources from a different source than .xml or .properties. My use case requires me to use JSON or db query

```java 
@Override
protected PropertiesHolder refreshProperties(String filename, @Nullable PropertiesHolder propHolder){
    PropertiesHolder prop = super.refreshProperties(filename, propHolder);
    if (propHolder.getProperties().isEmpty()){
        //try refreshing as json. since XML and  prop are not present)
    }
}
```
And This Class has a lot of useful reload logic which is pointless to reimplement if the functionality is already present.

snicoll:
Thanks for sharing your use case @Sahilshetye.

simonbasle:
@Sahilshetye do you absolutely need to override `refreshProperties`? Moving that line makes the logic a bit harder to follow and to prove that the correct `propHolder` has been added to the cache, so I'm not a fan of that proposal.

Maybe instead we could make the set of extensions tried within `refreshProperties` overridable. You would then just override this new `getResourceExtensions` method as well as `loadProperties`, and that's it.

I need to discuss this with the team, but does that sound like a plausible solution to your particular use case?

snicoll:
Closing in favor of PR #30369

Sahilshetye:
> der to follow and to prove that the correct `propHolder` has been added to the cache, so I'm not a fan of that proposal.

Works for me.

PR #30369, opened 2023-04-24
Add ReloadableResourceBundleMessageSource Resource selection hook

simonbasle:
This commit enables sub-classes to better customize resource selection
and resource loading in `ReloadableResourceBundleMessageSource`, without
a need to duplicate the caching logic of `refreshProperties`.

See gh-30334
Closes gh-30369


snicoll (approved):
LGTM!

PR #30388, opened 2023-04-27
Publish release candidates and milestones to maven central

navaneeth-mysore-sonatype:
I work for Sonatype and we'd love to have your milestone and release candidates in maven central.


bclozel:
Thanks for the proposal, but Spring has a long standing policy of publishing milestones and snapshots on [our own artifactory repo](https://repo.spring.io). Unless there are specific motivations for this we're unaware of, we are not going to change policy at this time.

PR #30391, opened 2023-04-28
Throwable and Error should not be caught

yang-zhang99:
- Throwable and Error should not be caught


bclozel:
Sorry but we don't accept contributions without any background or rationale about the change. If this is about code styling and personal preferences, we also tend to decline those unless it's explicitly listed in [our code style rules](https://github.com/spring-projects/spring-framework/wiki/Code-Style).

PR #30396, opened 2023-04-28
Support multiple style of parsing/printing Durations

simonbasle:
This commit introduces a notion of different styles for the formatting
of `Duration`.
The `DurationFormat` annotation is added to ease selection of a style,
which are represented as `DurationFormat.Style` enum, as well as a
supported time unit represented as `DurationFormat.Unit` enum.

`DurationFormatter` has been retroffited to take such a `Style`,
optionally, at construction. The default is still the JDK style aka
ISO-8601.

This introduces the new `SIMPLE` style which uses a single number + a
short human-readable suffix. For instance `-3ms` or `2h`.

This has the same semantics as the `DurationStyle` in Spring Boot and
is intended as a replacement for that feature, providing access to the
feature to projects that only depend on Spring Framework.

Finally, the `@Scheduled` annotation is improved by adding detection
of the style and parsing for the String versions of initial delay,  fixed
delay and fixed rate.

See gh-22013
See gh-22474

### DIFF
--- a/spring-context/src/main/java/org/springframework/format/datetime/standard/DurationFormatter.java
+++ b/spring-context/src/main/java/org/springframework/format/datetime/standard/DurationFormatter.java
@@ -18,7 +18,6 @@ package org.springframework.format.datetime.standard;
 
 import java.text.ParseException;
 import java.time.Duration;
-import java.time.temporal.ChronoUnit;
 import java.util.Locale;
 
 import org.springframework.format.Formatter;
@@ -39,7 +38,7 @@ class DurationFormatter implements Formatter<Duration> { //TODO why is this one 
 
 	private final DurationFormat.Style style;
 	@Nullable
-	private final ChronoUnit defaultUnit;
+	private final DurationFormat.Unit defaultUnit;
 
 	/**
 	 * Create a {@code DurationFormatter} following JSR-310's parsing rules for a Duration
@@ -52,7 +51,7 @@ class DurationFormatter implements Formatter<Duration> { //TODO why is this one 
 	/**
 	 * Create a {@code DurationFormatter} in a specific {@link DurationFormat.Style}.
 	 * <p>When a unit is needed but cannot be determined (e.g. printing a Duration in the
-	 * {@code SIMPLE} style), {@code ChronoUnit#MILLIS} is used.
+	 * {@code SIMPLE} style), {@code DurationFormat.Unit#MILLIS} is used.
 	 */
 	public DurationFormatter(DurationFormat.Style style) {
 		this(style, null);
@@ -60,7 +59,7 @@ class DurationFormatter implements Formatter<Duration> { //TODO why is this one 
 
 	/**
 	 * Create a {@code DurationFormatter} in a specific {@link DurationFormat.Style} with an
-	 * optional {@code ChronoUnit}.
+	 * optional {@code DurationFormat.Unit}.
 	 * <p>If a {@code defaultUnit} is specified, it may be used in parsing cases when no
 	 * unit is present in the string (provided the style allows for such a case). It will
 	 * also be used as the representation's resolution when printing in the
@@ -68,9 +67,9 @@ class DurationFormatter implements Formatter<Duration> { //TODO why is this one 
 	 * unit.
 	 *
 	 * @param style the {@code DurationStyle} to use
-	 * @param defaultUnit the {@code ChronoUnit} to fall back to when parsing and printing
+	 * @param defaultUnit the {@code DurationFormat.Unit} to fall back to when parsing and printing
 	 */
-	public DurationFormatter(DurationFormat.Style style, @Nullable ChronoUnit defaultUnit) {
+	public DurationFormatter(DurationFormat.Style style, @Nullable DurationFormat.Unit defaultUnit) {
 		this.style = style;
 		this.defaultUnit = defaultUnit;
 	}
--- a/spring-context/src/main/java/org/springframework/format/datetime/standard/DurationFormatter.java
+++ b/spring-context/src/main/java/org/springframework/format/datetime/standard/DurationFormatter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


simonbasle:
Note: the first commit was closer to the arrangement found in Boot, but there was too much logic in the `annotation` package for my taste, and it was exposing `Unit` as a wrapper over `ChronoUnit`. This difference in package structure was also increasing the risk of package tangle, so I extracted the parsing and printing logic into `DurationFormatterUtils`.

spring-context/src/main/java/org/springframework/format/datetime/standard/DurationFormatter.java
```diff
@@ -18,28 +18,79 @@
...
+ * supporting additional {@code DurationFormat.Style} styles.
  *
  * @author Juergen Hoeller
  * @since 4.2.4
- * @see Duration#parse
+ * @see DurationFormatterUtils
+ * @see DurationFormat.Style
  */
-class DurationFormatter implements Formatter<Duration> {
+class DurationFormatter implements Formatter<Duration> { //TODO why is this one package-private ? make public and change since taglet ?
```
simonbasle, line 38: @jhoeller I was wondering if there was a particular reason this `Formatter` has been kept package-private so far?

simonbasle:
third iteration (missing a bit of documentation) is reintroducing `DurationFormat.Unit`, bearing part of the logic in dealing with suffixes and giving a better sense of exactly which units are actually supported (unlike with `ChronoUnit` where eg. `MILLENIA` would compile but fail at runtime)

philwebb:
I might be a bit biased, but I personally quite liked the logic being in the `DurationStyle` enum rather than another `*Utils` class. Especially as `DurationFormatterUtils` now depends on an enum that's inside the `org.springframework.format.annotation.DurationFormat` annotation.

I also think it's a bit easier to use. E.g. `DurationFormatterUtils.parse("10s", DurationFormat.Style.SIMPLE)` vs `DurationStyle.SIMPLE.parse("10s")`.

Perhaps the enum can be extracted from the annotation and moved to the `org.springframework.format.datetime.standard` package?

[These tests in Spring Boot](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/convert/DurationStyleTests.java) show how that API might look when being used.

snicoll:
Spring Framework has `Utils` classes in various places so I can understand why Simon went down that route. And having all that code as part of the annotation package felt wrong. 

I like the idea of moving the unit class outside of the annotation package. It looks like it would improve the usability of the code and address the concern of having too much code in the annotation package.

simonbasle:
I went that route for consistency with Framework current package arrangement, with some extra effort in avoiding too much logic in the annotation. `Unit`-related methods felt simple enough that they could be implemented in the enum, as a middle ground.

The trouble with moving the `Unit` (or `Style` for that matter) outside of the annotation package is that the `@DurationFormat` annotation depends on the Style and Unit as part of its API. We can have package `org.springframework.format.datetime.standard` depend on `org.springframework.format.annotation` (there's precedent here), but not the other way around.

philwebb:
I missed the existing `org.springframework.format.datetime.standard` -> `org.springframework.format.annotation` dependency. That does indeed complicate things!

I wonder what `DurationFormatterUtils` would look like if it didn't use enums from `org.springframework.format.annotation`? Perhaps it could offer public `parseIso8601`, `parseSimple` and `parseDetected` methods and leave something else to work out which of those methods to call if the annotation is present. I guess the `DurationFormat.Unit` somewhat messes with that, but `ChronoUnit` or `TimeUnit` could be an option (at the expense of runtime failures if a user is picks `ChronoUnit.MILLENIA`.

I think as it stands `DurationFormatterUtils` is quite coupled to `@DurationFormat`. That shows up in `ScheduledAnnotationBeanPostProcessor` which is pulling in `DurationFormat.Unit` so that it can call `fromChronoUnit` and then `DurationFormatterUtils.detectAndParse`. In fact, I think `ScheduledAnnotationBeanPostProcessor` is the only reason that `DurationFormatterUtils` is public and I think `detectAndParse` is the only method it needs.

AFAICT we don't have any other public utils classes under `org.springframework.format`. I wonder if using `DurationFormatter` directly in `ScheduledAnnotationBeanPostProcessor` might help us keep the utils package-private?

E.g., in `ScheduledAnnotationBeanPostProcessor`:

```java
private static Duration toDuration(String value, TimeUnit timeUnit) {
	new DurationFormatter(timeUnit).parse(value);
}
```

bclozel:
I've just reviewed this PR and rebased against the main branch. There were some changes on the `@Scheduled` annotation which create a conflict. I've resolved them locally with the following (see the 3rd item in the bullet point list):

```java
/**
 * Execute the annotated method with a fixed period between the end of the
 * last invocation and the start of the next.
 * <p>The duration String can be in several formats:
 * <ul>
 *     <li>a plain integer &mdash; which is interpreted to represent a duration in
 *     milliseconds by default unless overridden via {@link #timeUnit()} (prefer
 *     using {@link #fixedDelay()} in that case)</li>
 *     <li>any of the known {@link org.springframework.format.annotation.DurationFormat.Style
 *     DurationFormat.Style}: the {@link org.springframework.format.annotation.DurationFormat.Style#ISO8601 ISO8601}
 *     style or the {@link org.springframework.format.annotation.DurationFormat.Style#SIMPLE SIMPLE} style
 *     &mdash; using the {@link #timeUnit()} as fallback if the string doesn't contain an explicit unit</li>
 *     <li>one of the above, with Spring-style "${...}" placeholders as well as SpEL expressions</li>
 * </ul>
 * @return the delay as a String value &mdash; for example a placeholder,
 * or a {@link org.springframework.format.annotation.DurationFormat.Style#ISO8601 java.time.Duration} compliant value
 * or a {@link org.springframework.format.annotation.DurationFormat.Style#SIMPLE simple format} compliant value
 * @since 3.2.2
 * @see #fixedDelay()
 */
```

> AFAICT we don't have any other public utils classes under org.springframework.format. I wonder if using DurationFormatter directly in ScheduledAnnotationBeanPostProcessor might help us keep the utils package-private?

I've tried that approach, and this involves moving the detection code to the `DurationFormat.Style` enum and the usage pattern doesn't seem ideal:

```java
private static Duration toDuration(String value, TimeUnit timeUnit) {
	DurationFormat.Unit unit = DurationFormat.Unit.fromChronoUnit(timeUnit.toChronoUnit());
	DurationFormat.Style style = DurationFormat.Style.detect(value);
	try {
		return new DurationFormatter(style, unit).parse(value, null);
	} catch (ParseException exc) {
		throw new IllegalArgumentException(exc);
	}
}
```


As Simon said:

> ensuring good compatibility/migrability with what Boot has is definitely a priority, but this felt like the best arrangement to me so far.

I think we don't have enough information at this point to decide whether this arrangement is good enough for Spring Boot. I'm rescheduling this to 6.1.x as we need to find a proper time to experiment with both Spring Framework and Spring Boot SNAPSHOTs.

philwebb:
There's a comment on https://github.com/spring-projects/spring-boot/issues/37562#issuecomment-2036949439 which might be relevant to this feature. It would be nice to support Kotlin `1.5s` style values.

PR #30402, opened 2023-04-30
Use suspending style filter function in Kotlin code sample

jbotuck:
This is really a  question in the form of a PR. Is this code sample equivalent of the original (assuming `renewToken` is a suspend function)?

### DIFF
--- a/framework-docs/src/docs/asciidoc/web/webflux-webclient.adoc
+++ b/framework-docs/src/docs/asciidoc/web/webflux-webclient.adoc
@@ -1029,7 +1029,7 @@ any response content, whether expected or not, is released:
 ----
 	fun renewTokenFilter(): ExchangeFilterFunction {
 		return ExchangeFilterFunction { request: ClientRequest, next: ExchangeFunction ->
-			mono {
+			mono (Dispatchers.Unconfined){
 				val response = next.exchange(request).awaitSingle()
 				if (response.statusCode().value() == HttpStatus.UNAUTHORIZED.value()) {
 					response.releaseBody().awaitSingleOrNull()
--- a/framework-docs/src/docs/asciidoc/web/webflux-webclient.adoc
+++ b/framework-docs/src/docs/asciidoc/web/webflux-webclient.adoc
@@ -1027,19 +1027,15 @@ any response content, whether expected or not, is released:
 [source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
 .Kotlin
 ----
-	fun renewTokenFilter(): ExchangeFilterFunction? {
-		return ExchangeFilterFunction { request: ClientRequest?, next: ExchangeFunction ->
-			next.exchange(request!!).flatMap { response: ClientResponse ->
+	fun renewTokenFilter(): ExchangeFilterFunction {
+		return ExchangeFilterFunction { request: ClientRequest, next: ExchangeFunction ->
+			mono {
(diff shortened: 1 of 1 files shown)


jbotuck:
@sdeleuze

framework-docs/src/docs/asciidoc/web/webflux-webclient.adoc
```diff
@@ -1027,19 +1027,16 @@ any response content, whether expected or not, is released:
 [source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
 .Kotlin
 ----
-	fun renewTokenFilter(): ExchangeFilterFunction? {
-		return ExchangeFilterFunction { request: ClientRequest?, next: ExchangeFunction ->
-			next.exchange(request!!).flatMap { response: ClientResponse ->
+	fun renewTokenFilter(): ExchangeFilterFunction {
+		return ExchangeFilterFunction { request: ClientRequest, next: ExchangeFunction ->
+			mono {
```
sdeleuze, line 1032: I think `mono(Dispatchers.Unconfined)` should be used here instead of `mono` which may default to `Dispatchers.Default` which is baked by a shared thread pool. That would also be consistent by `CoWebFilter` introduced by @poutsma.

sdeleuze (changes requested):

framework-docs/src/docs/asciidoc/web/webflux-webclient.adoc
jbotuck: Also, why can't I find the suspend function in CoWebFilter documented anywhere? https://docs.spring.io/spring-framework/docs/current/kdoc-api/spring-web/org.springframework.web.server/-co-web-filter/index.html

sdeleuze: > Are there any plans for a CoExchangeFilterFunction?
See me answer in the PR level comment.

> Also, why can't I find the suspend function in CoWebFilter documented anywhere?
`CoWebFilter` is documented in the API documentation you linked, and is "just" a `WebFilter` implementation. I don't think the reference documentation includes a `WebFilter` implementation code sample where we should translate the Kotlin version with `CoWebFilter`.

sdeleuze:
> Are there any plans for a CoExchangeFilterFunction?
After a related discussion with @poutsma, we think that the proper way to move forward is to introduce indeed a `CoExchangeFilterFunction` as part of Spring Framework 6.1, then update the documentation to leverage it. Please create a related issue and we will likely implement it.

As a consequence, I close this PR unmerged.

framework-docs/src/docs/asciidoc/web/webflux-webclient.adoc
jbotuck: I'd expect that a protected method intended to be overridden like https://github.com/spring-projects/spring-framework/blob/f7367ceae53f6f812b8634fc324933aaf63785c9/spring-web/src/main/kotlin/org/springframework/web/server/CoWebFilter.kt#L48 would show up in the kdoc, but I might just be wrong about that

sdeleuze: Good catch, maybe a Dokka issue for that specific use case where we have 2 `filter` functions pretty close in term of signature.

Can you please create a related issue in https://github.com/Kotlin/dokka linking this source code and the generated kdoc?

PR #30403, opened 2023-04-30
HTTP interface client should not have blockingTimeout set and leave it to the underlying client by default

### LINKED ISSUE #30248: HTTP interface client should not set a default blockingTimeout value and instead let the underlying HTTP client determine the behavior
**Affects:**  Spring Boot 3.0.5 / Spring Framework 6.0.7

---

As mentioned in the title, it seems the webclient timeout value does not work when using the HTTP interface.
But the same timeout value works when using the webclient directly as below.

```
    @EventListener
    public void onStartup(ApplicationStartedEvent event) {
        // Non-blocking, working fine
        event.getApplicationContext().getBean(WebClient.class)
                .get()
                .uri("http://localhost:8080/test")
                .retrieve().bodyToMono(String.class)
                .subscribe(System.out::println);

        // Blocking, working fine
        String result = event.getApplicationContext().getBean(WebClient.class)
                .get()
                .uri("http://localhost:8080/test")
                .retrieve().bodyToMono(String.class)
                .block();
        System.out.println(result);

        // Blocking but with HttpInterface, is failing
        SampleApi sampleApi = event.getApplicationContext().getBean(SampleApi.class);
        System.out.println(sampleApi.get().getStatusCode());
    }

```

I have added a minimal sample application that demonstrates the above issue.
https://github.com/pmverma/demo-http-interface-webclient-timeout

All codes are in a single file [DemoHttpInterfaceWebclientTimeoutApplication.java](https://github.com/pmverma/demo-http-interface-webclient-timeout/blob/master/src/main/java/com/example/demohttpinterfacewebclienttimeout/DemoHttpInterfaceWebclientTimeoutApplication.java)

Regards,
Mohan

OlgaMaciaszek:
Fixes gh-30248.

### DIFF
--- a/framework-docs/src/docs/asciidoc/integration/rest-clients.adoc
+++ b/framework-docs/src/docs/asciidoc/integration/rest-clients.adoc
@@ -497,6 +497,11 @@ Annotated, HTTP exchange methods support the following return values:
 TIP: You can also use any other async or reactive types registered in the
 `ReactiveAdapterRegistry`.
 
+TIP: For non-reactive types, blocking from a reactive publisher is performed
+under the hood by the framework. By default, it is done without a timeout.
+You can set a timeout for blocking by calling `blockTimeout(Duration blockTimeout)`
+on `HttpServiceProxyFactory.Builder`.
+
 
 [[rest-http-interface-exceptions]]
 === Exception Handling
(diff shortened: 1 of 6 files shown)


rstoyanchev (approved):
Overall, looks good. 

We just need to expand on the documentation a bit. It's not that we block without a timeout, but rather we defer to the underlying client, which is in a better position to provide more control over timeout, e.g. connecting vs request handling, etc. So the `blockingTimeout` at our level is a higher level convenience, but without any settings out of the box.

PR #30419, opened 2023-05-04
fix object equals

shenjianeng:


snicoll:
Thanks for the PR but the code is fine as it is.

Issue #30360, opened 2023-04-20
ApplicationEventPublisher - PayloadApplicationEvent not received by EventListener if event fired from child context


robtha:
If PayloadApplicationEvent is published via ApplicationEventPublisher from child context, it will not be received by EventListeners in the parent context. This worked in Spring 5.x

Tested with spring versions: 6.0.7, 6.0.8

I have attached a sample application to reproduce the problem.

PayloadApplicationEventS are only received by the EventListener in the child context but not in parent context.
Subclasses of ApplicationEvent are not affected and still received by the listener in the parent context.

kind regards
Robert

[sample.zip](https://github.com/spring-projects/spring-framework/files/11285709/sample.zip)

simonbasle:
This is indeed a regression linked to https://github.com/spring-projects/spring-framework/pull/24599

simonbasle:
Superseded by PR gh-30420.

jhoeller:
We eventually decided to introduce a focused revision for addressing this regression, consistently pre-resolving event types vs payload types now.

PR #30420, opened 2023-05-04
Fix publishing of PayloadApplicationEvents in parent context

simonbasle:
This commit introduces a more defensive propagation of an event
and the optionally associated eventType when calling `publishEvent`,
especially when a parent context is present.

Notably, it avoids a situation where the parent `publishEvent` call
would generate a new `PayloadApplicationEvent` with the wrong eventType
(`PayloadApplicationEvent<PayloadApplicationEvent<T>>`).

This commit also adds test coverage (which includes cases where an event
is fired before the multicaster is set) and a defensive check against
inconsistent publishing of a PayloadApplicationEvent (even though that
is something that is expected to be done mostly internally by the
framework).

See gh-30360
Closes gh-30420


simonbasle:
superseded by https://github.com/spring-projects/spring-framework/commit/e228f4ba6479694975c93e2f2a908ed91481bc65

PR #30433, opened 2023-05-05
Add non-null precondition checks in `DefaultEntityResponseBuilder`

srivatsa-cfp:


poutsma:
Merged, thank you.

I did remove the latter two null checks, because they were unnecessary. Any `equals` method should be able to deal with `null` parameters.

PR #30435, opened 2023-05-06
Fix migration to asciidoctor tabs

rwinch:
The automation to asciidoctor tabs did not migrate properly for all cases this commit fixes the migration.

For example, [this](https://docs.spring.io/spring-framework/reference/core/aop/ataspectj/advice.html#aop-ataspectj-advice-proceeding-with-the-call) does not properly render as tabs due to the migration not working properly.


rstoyanchev:
I think this applies to 6.0.x as well. I'll rebase it on 6.0.x before merging.

rwinch:
@rstoyanchev You are right. I messed up which branch to target. I've updated it to point at 6.0.x now.

PR #30457, opened 2023-05-09
Replace magic numbers with constants

sigee:


snicoll:
Thanks for the suggestion, but I don't think we need to change this in tests.

PR #30458, opened 2023-05-09
Fix code style issues (From C-style to Java-style array declaration)

sigee:


snicoll:
This is our copy of cglib and while we're maintaining it now, we prefer to only change what's really necessary.

PR #30459, opened 2023-05-09
Fix control flow issues (Extract common part from if)

sigee:


snicoll:
@sigee thanks for the PR but we prefer the expressiveness of complete if/else blocks. I wouldn't call the current code having a control flow issue.

PR #30462, opened 2023-05-10
Support `@Bean` 'lite' mode when no `@Bean` methods are declared locally

### LINKED ISSUE #30449: `@Bean` 'lite' mode not supported if `@Bean` methods are not declared locally
I'm not sure it's a bug or undocumented limitation. 

If no `@Bean` defined in `Config` then `@Bean` defined in `BaseConfig` will be ignored.
```java
@ContextConfiguration(classes = Config.class)
public class InjectionTests {

	@Autowired
	private String bean1;

	@Autowired
	private String bean2;

	@Test
	void test() {
		assertThat(bean1).isEqualTo("bean1");
		assertThat(bean2).isEqualTo("bean2");
	}

}

class Config extends BaseConfig {

}

class BaseConfig {
	@Bean
	String bean1() {
		return "bean1";
	}
	@Bean
	String bean2() {
		return "bean2";
	}
}
```
there are many workarounds
1. move any one of `@Bean` to `Config`
```java
class Config extends BaseConfig {
	@Bean
	String bean2() {
		return "bean2";
	}
}

class BaseConfig {
	@Bean
	String bean1() {
		return "bean1";
	}
}
```
2. mark `@Configuration` or `@Component` on `Config`
```java
@Configuration
// @Component
class Config extends BaseConfig {

}

class BaseConfig {
	@Bean
	String bean1() {
		return "bean1";
	}
	@Bean
	String bean2() {
		return "bean2";
	}
}
```
3. use `@Import` instead of `ContextConfiguration`
```java
@Import(Config.class)
public class InjectionTests {

	@Autowired
	private String bean1;

	@Autowired
	private String bean2;

	@Test
	void test() {
		assertThat(bean1).isEqualTo("bean1");
		assertThat(bean2).isEqualTo("bean2");
	}

}
```

quaff:
ConfigurationClassUtils should check bean methods in super classes also
Fix GH-30449

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
@@ -349,7 +349,7 @@ class ConfigurationClassParser {
 		if (!memberClasses.isEmpty()) {
 			List<SourceClass> candidates = new ArrayList<>(memberClasses.size());
 			for (SourceClass memberClass : memberClasses) {
-				if (ConfigurationClassUtils.isConfigurationCandidate(memberClass.getMetadata()) &&
+				if (ConfigurationClassUtils.isConfigurationCandidate(memberClass.getMetadata(), this.metadataReaderFactory) &&
 						!memberClass.getMetadata().getClassName().equals(configClass.getMetadata().getClassName())) {
 					candidates.add(memberClass);
 				}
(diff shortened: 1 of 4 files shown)


quaff:
Test case added, could you review it? @jhoeller @sbrannen

jhoeller:
This is not within the intentions of our configuration class check which may literally apply to every bean in an application context. Driven by the fact that we're typically introspecting class files in a classpath scan there, not fully resolved classes, we intentionally do not rely on the superclass/interface hierarchy of such beans. We only search for `@Bean` methods on superclasses and interfaces if there is an indication on the given class to begin with, either a stereotype or local `@Bean` methods. With the latter local method check, we're just trying to be lenient; in the original version of that code, we insisted on a stereotype for any lite configuration class. I'm afraid we cannot be even more lenient than that when that would incur significant extra parsing cost.

quaff:
> I'm afraid we cannot be even more lenient than that when that would incur significant extra parsing cost.
I don't think it would incur significant extra parsing cost since it's a rare case, at least document should be improved if you insist.

PR #30466, opened 2023-05-10
Use cached playbook in build dir

rwinch:
This is an update to the per branch build to allow placing the playbook in the build dir. It needs to happen with gh-30465


rstoyanchev:
This is merged.

PR #30478, opened 2023-05-11
HttpClient based ClientHttpRequestFactory

mdeinum:
As JDK17 is now the baseline it is possible to use the HttpClient provided by Java.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/JdkClientClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/JdkClientClientHttpRequest.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2023-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


poutsma:
Thank for your submitting this PR, @mdeinum. I made several changes to the PR.

- Removed duplicate `Client` in types names, to mirror the reactive versions in `org.springframework.http.client.reactive`.
- Removed request buffering in favor of an `OutputStream` to `Flow.Publisher<ByteBuffer>` bridge, see [OutputStreamPublisher](https://github.com/spring-projects/spring-framework/blob/dd57ec9fa08babc5c1e4469a5d3c293506d8b5b5/spring-web/src/main/java/org/springframework/http/client/OutputStreamPublisher.java). Thank you, @OlegDokuka!
- Made request and types package private, similar to other request and response types in the package.
- Various other small improvements.

alexcrownus:
JdkClientHttpRequestFactory is not compatible with BufferingClientHttpRequestFactory #31027

Issue #26021, opened 2020-11-03
Add support for Class-based mapping in DatabaseClient


mp911de:
Spring Data R2DBC's `DatabaseClient` provided a rich fluent API including functionality that allowed a convenient mapping of results into a particular type through `as(Class)`. During the migration of `DatabaseClient` we decided to drop that functionality to provide functionality that we knew how to provide from a Spring Framework perspective. Target object mapping in Spring Data used core Spring Data functionality that isn't available on the Spring Framework level.

We see a lot of demand for a method that is convenient to use from the caller side and that applies convention-based mapping of primitives  (Java primitives and R2DBC primitives) and data classes.

Right now, mapping requires a mapping function to be used through `map(BiFunction<Row, RowMetadata, T>)`.

jhoeller:
I've added this to our backlog for the time being but I'd be happy to bring it into a 5.3.x release soon.

To get started, ideally we'd have at least a sketch of what such a method's implementation should look like. Are we trying to make this fully pluggable or are we rather just baking in some common cases out of the box, leaving the rest up to custom mapping functions?

mp911de:
Generally speaking, it would be convenient to use Spring Data (or any other mapping framework) if it is on the classpath when mapping query results, regardless of the SQL connector technology (JDBC or R2DBC). A pluggable variant could provide streamlined instantiation strategies (for example reflection-less) or apply column-name mappings but it comes with increased complexity and the need for an SPI.

A non-pluggable arrangement could serve for the most common use-cases. Such an API needs to differentiate whether the target type is expected to map onto a single column (`as(Long.class)` (Java-primitive), `as(LocalDateTime.class)` (R2DBC primitive), `as(Blob.class)` (R2DBC primitive), `as(Json.class)` (R2DBC Postgres primitive)) or whether the result should be mapped onto a data class/record/POJO.

Since R2DBC driver primitives vary across drivers, it doesn't make sense to update our framework code for each new driver that we discover. I wonder whether it would make sense to provide an API on R2DBC level that allows consumers to identify R2DBC and driver primitives (r2dbc/r2dbc-spi#192).

schauder:
Could we port `BeanPropertyRowMapper` to R2DBC?

cparaskeva:
Any updates on that ?!

pawelryznar:
@mp911de by this 
> Right now, mapping requires a mapping function to be used through `map(BiFunction<Row, RowMetadata, T>)`.

[...]

PR #30530, opened 2023-05-23
Port BeanPropertyRowMapper and DataClassRowMapper for r2dbc

simonbasle:


simonbasle:
see discussion in gh-26021. cc @mp911de and @jhoeller

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/BeanPropertyRowMapper.java
```diff
@@ -0,0 +1,436 @@
...
+ *
+ * @author Thomas Risberg
+ * @author Juergen Hoeller
+ * @author Sam Brannen
+ * @author Simon Baslé
+ * @since 6.1
+ * @param <T> the result type
+ * @see DataClassRowMapper
+ */
+public class BeanPropertyRowMapper<T> implements BiFunction<Row, RowMetadata, T> {
```
mp911de, line 92: It might even be sufficient to let this be `Function<Readable>` for mapping stored procedure outputs in addition to `Row`s.

mp911de:
As `Readable` is not a sealed interface, you're right; there could be potential additional types. In practice, we haven't seen any being around. Also, there was no indication that someone would come up with an additional subtype that isn't a Row or OutParameters.

PR #30534, opened 2023-05-24
Always fall back to original method if annotation pointcut used

### LINKED ISSUE #27119: Not annotation pointcut matches dynamic proxy methods incorrectly
"target" and "not @annotation" combination not works as expected, but "target" and "@annotation" combination works fine.
```java
	// works as expected
	@Around("execution(public * *(..)) and target(repository) and @annotation(org.springframework.transaction.annotation.Transactional)")
	public Object interceptPositive(ProceedingJoinPoint pjp, Repository repository) throws Throwable {
		Method method = ((MethodSignature) pjp.getSignature()).getMethod();
		if (!method.isAnnotationPresent(Transactional.class))
			throw new RuntimeException("annotation should present here");
		return pjp.proceed();
	}

	// not works as expected
	@Around("execution(public * *(..)) and target(repository) and not @annotation(org.springframework.transaction.annotation.Transactional)")
	public Object interceptNegative(ProceedingJoinPoint pjp, Repository repository) throws Throwable {
		Method method = ((MethodSignature) pjp.getSignature()).getMethod();
		if (method.isAnnotationPresent(Transactional.class))
			throw new RuntimeException("annotation should not present here");
		return pjp.proceed();
	}
```
[test-pointcut.zip](https://github.com/spring-projects/spring-framework/files/6751811/test-pointcut.zip) is a test project

quaff:
Prior to this commit, `AspectJExpressionPointcut` didn't fall back to the original method if `!@annotation()` is used, which can result in a false positive.

- fixes #27119

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/aspectj/AspectJExpressionPointcut.java
+++ b/spring-aop/src/main/java/org/springframework/aop/aspectj/AspectJExpressionPointcut.java
@@ -78,6 +78,7 @@ import org.springframework.util.StringUtils;
  * @author Juergen Hoeller
  * @author Ramnivas Laddad
  * @author Dave Syer
+ * @author Yanming Zhou
  * @since 2.0
  */
 @SuppressWarnings("serial")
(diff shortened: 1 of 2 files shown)


quaff:
Would you take a look? @sbrannen @jhoeller

sbrannen:
Yes, we'll take a look.

Thanks for the PR.

PR #30537, opened 2023-05-24
Remove non-empty filename check in `ContentDisposition` builder

Foolkin:
The ContentDisposition Builder must be consistent with ContentDisposition and accept a nullable filename.


sbrannen:

Thanks

PR #30546, opened 2023-05-25
Support parsing `long` dates in `InstantFormatter`

### LINKED ISSUE #30312: Add support for converting a String to an Instant
What I'm trying to achieve:

```java
@ConfigurationProperties
record ApplicationProps(Instant instant) {}
```

with the following `application.properties`:
```
instant = 1680612880417
```
where the value is millis since the epoch, see [`Instant.ofEpochMilli()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/Instant.html#ofEpochMilli(long)).

Edit:
Currently this fails with a
```
...
Caused by: java.lang.IllegalArgumentException: Parse attempt failed for value [123456789]
	at org.springframework.format.support.FormattingConversionService$ParserConverter.convert(FormattingConversionService.java:224) ~[spring-context-6.0.6.jar:6.0.6]
	at org.springframework.core.convert.support.ConversionUtils.invokeConverter(ConversionUtils.java:41) ~[spring-core-6.0.6.jar:6.0.6]
	... 117 common frames omitted
Caused by: java.time.format.DateTimeParseException: Text '123456789' could not be parsed at index 0
```

Perhaps this is a good candidate for a default binding.

Considerations:
- Why are we choosing `Instant.ofEpochMilli()` and not `Instant.ofEpochSecond()`? It just felt more natural. Is this uncertainty reason enough to refuse this request? That said, there already is a registered-by-default `Converter<Long, Instant>` which uses `ofEpochMilli`. Therefore, an additional Formatter would be consistent with the already-existing Converter.
- Do we also need to add more similar formatters for types like `java.util.Date` or `java.sql.Timestamp`? I do no think think so as neither of them has a simple `ofEpochMilli()`-like creator method or clear semantics. Other `java.time` also are not good targets for this unless OffsetDateTime/ZonedDateTime is considered okay given a default timezone.

RemusRD:
- addresses #30312

### DIFF
--- a/spring-context/src/test/java/org/springframework/format/datetime/standard/InstantFormatterTests.java
+++ b/spring-context/src/test/java/org/springframework/format/datetime/standard/InstantFormatterTests.java
@@ -140,7 +140,7 @@ class InstantFormatterTests {
 
 		@Override
 		public Stream<Arguments> provideArguments(ExtensionContext context) {
-			return random.longs(DATA_SET_SIZE, Instant.ofEpochMilli(Long.MIN_VALUE).toEpochMilli(), Instant.ofEpochMilli(Long.MAX_VALUE).toEpochMilli())
+			return random.longs(DATA_SET_SIZE, Long.MIN_VALUE, Long.MAX_VALUE)
 					.mapToObj(Instant::ofEpochMilli)
 					.map(instant -> instant.truncatedTo(ChronoUnit.MILLIS))
 					.map(Arguments::of);
--- a/spring-context/src/test/java/org/springframework/format/datetime/standard/InstantFormatterTests.java
+++ b/spring-context/src/test/java/org/springframework/format/datetime/standard/InstantFormatterTests.java
@@ -19,6 +19,7 @@ package org.springframework.format.datetime.standard;
 import java.text.ParseException;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
 import java.util.Random;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


RemusRD:
See https://github.com/spring-projects/spring-framework/issues/30312

spring-context/src/test/java/org/springframework/format/datetime/standard/InstantFormatterTests.java
```diff
@@ -121,5 +132,19 @@ Stream<?> provideArguments() {
...
 	}
+	private static final class RandomEpochMillisProvider implements ArgumentsProvider {
+
+		private static final long DATA_SET_SIZE = 10;
+
+		private static final Random random = new Random();
+
+		@Override
+		public Stream<Arguments> provideArguments(ExtensionContext context) {
+			return random.longs(DATA_SET_SIZE, Instant.ofEpochMilli(Long.MIN_VALUE).toEpochMilli(), Instant.ofEpochMilli(Long.MAX_VALUE).toEpochMilli())
```
simonbasle, line 143: this can be simplified:
```suggestion
			return random.longs(DATA_SET_SIZE, Long.MIN_VALUE, Long.MAX_VALUE)
```

simonbasle:
> Wondering if the formatting shouldn't have been done by my Intellij IDEA? Maybe I have something not configured properly? Or is there any gradle task to format the code?

Thanks for fixing the formatting. There isn't really a Gradle task, but there is some guidance in the [`CONTRIBUTING.md`](https://github.com/spring-projects/spring-framework/blob/main/CONTRIBUTING.md) pointing to wiki pages. There's also a `.editorconfig` at the root of the project and some exported settings in `src/idea`.

PR #30570, opened 2023-05-31
Fix invalid link in transaction resources

kihyuk-sung:
Hello. I found that `Distributed transactions in Spring, with and without XA` url in Spring Framework - Data Access - Transaction Management - Further Resources is not linked to the valid url.
https://docs.spring.io/spring-framework/reference/data-access/transaction/resources.html

https://www.infoworld.com/article/2077963/distributed-transactions-in-spring--with-and-without-xa.html

This url contains double dash (--). In that case, asciidoctor use some specific methods described in followed links.

https://docs.asciidoctor.org/asciidoc/latest/macros/complex-urls/
https://github.com/asciidoctor/asciidoctor/issues/3965

I fix this problem using link prefix and double plus line macro(++) because it preserve the original url.

There are another solutions to solve this problem. If you don't like this way, please edit this document page.

### DIFF
--- a/framework-docs/modules/ROOT/pages/data-access/transaction/resources.adoc
+++ b/framework-docs/modules/ROOT/pages/data-access/transaction/resources.adoc
@@ -4,7 +4,7 @@
 
 For more information about the Spring Framework's transaction support, see:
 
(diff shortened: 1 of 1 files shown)


snicoll:
Good catch, thanks @kihyuk-sung!

PR #30593, opened 2023-06-05
Fix example in Javadoc for `MultipartBodyBuilder`

heoYH:
The example of MultipartBodyBuilder was wrong, so it was corrected.


sbrannen:
Hi @heoYH,

Good catch, and congratulations on submitting your first PR for the Spring Framework! 👍 

Although `.body(BodyInserters.fromMultipartData(multipartBody))` would work, I assume the intention was to use `.bodyValue(multipartBody)` instead.

So I'll merge this PR to give you credit and then change the example to use the more succinct `bodyValue()` method.

sbrannen:

Thanks

PR #30599, opened 2023-06-06
Adding Reminder Information to websocket doc

heartape:
I didn't use SockJS when I first started using Stomp, but the entire document used this option, which caused me to be unable to connect to the browser. Later, when I discovered this option, I was able to restore it to normal. I think it is necessary to provide a reminder to prevent spending time investigating such worthless issues.


sdeleuze:
I understand how the documentation was potentially confusing related to the default usage of SockJS, but this has been revised via #30857 so I think the reminder is not needed anymore.

PR #30636, opened 2023-06-11
Resolve default value when argument converts to null

### LINKED ISSUE #29550: Space in integer request parameter neither triggers the default value nor generates a 400
**Affects:** spring-web 5.3.22

### Expectation

Given an endpoint with an `Integer` parameter named `limit` with a default value of `10`, when a user makes a request with `limit=%20` then the system should either return a `400 - Bad Request` or have the default value applied.

### Actual

Given an endpoint with an `Integer` parameter named `limit` with a default value of `10`,  when a user makes a request with `limit=%20` then the system accepts the request and the limit variable has a `null` value.

This happens because when checking for the default value, Spring will not trim the string, so `" "` will be a value and will not trigger the default value.

But when binding the value to the variable, [this line](https://github.com/spring-projects/spring-framework/blob/main/spring-beans/src/main/java/org/springframework/beans/propertyeditors/CustomNumberEditor.java#L105) will resolve to `true` (because `allowEmpty` is set to `true` and `!StringUtils.hasText(" ")` resolves to `true`), so `null` will be set (instead of generating an exception).

I have no idea how to fix it, because this does not seem to be a bug in the code, but a mismatch of expectations.

### Steps to reproduce

1. Clone the reproducer https://github.com/oscarfh/spring-reproducer
2. Start the application
3. Call in your terminal: `curl "http://localhost:8080/endpoint?limit=`
   - "50" will be returned, this is the default value set for the limit endpoint, meaning that the default value was applied.
4. Call `curl "http://localhost:8080/endpoint?limit=a"`
   - Note that a 400 is returned, meaning that limit does not support string values.
5. Call `curl "http://localhost:8080/endpoint?limit=%20"`
   - Notice how "null" is returned, meaning that the `limit` variable was assigned the `null` value.

This is unexpected because, due to the fact that this variable is an integer with a default value. You either expect it to be the integer supplied by the user or the default value. You do not expect it to be null. This null will then very likely cause a null pointer exception in your code.

Nheyll:
Resolves  #29550
I just added a condition that checks if the argument is converted to null after being resolved a first time. If it does and it has a default value, the default value is resolved.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/annotation/AbstractNamedValueMethodArgumentResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/method/annotation/AbstractNamedValueMethodArgumentResolver.java
@@ -123,6 +123,11 @@ public abstract class AbstractNamedValueMethodArgumentResolver implements Handle
 			WebDataBinder binder = binderFactory.createBinder(webRequest, null, namedValueInfo.name);
 			try {
 				arg = binder.convertIfNecessary(arg, parameter.getParameterType(), parameter);
+				if (arg == null && namedValueInfo.defaultValue != null) {
+					arg = binder.convertIfNecessary(
+							resolveEmbeddedValuesAndExpressions(namedValueInfo.defaultValue),
+							parameter.getParameterType(), parameter);
+				}
 			}
 			catch (ConversionNotSupportedException ex) {
 				throw new MethodArgumentConversionNotSupportedException(arg, ex.getRequiredType(),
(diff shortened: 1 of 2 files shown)


quaff:
You should add a test case.

Nheyll:
> You should add a test case.

Just pushed my test case

izeye:
https://github.com/spring-projects/spring-framework/issues/29550 that this PR seems to try to resolve has been resolved in b98c1ec36a2a00ff9cc3a7fa5c9fd0e776a97225 already.

PR #30651, opened 2023-06-12
Introduce `ObservationThreadLocalTestExecutionListener` in the TestContext framework

marcingrzejszczak:
Without this change we have no way to pass the `ObservationRegistry` that is registered in the given test's `ApplicationContext` as the one that should be used by the `ObservationThreadLocalAccessor`.

With this change we introduce a `TestExecutionListener` that will set an `ObservationRegistry` registered in the test's `ApplicationContext` for each test class and after each test class will restore the previously stored `ObservationRegistry` instance.


marcingrzejszczak:
cc @rstoyanchev after upgrading the context propagation library I needed to update one class

jonatan-ivanov (approved):

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ReactiveTypeHandler.java
```diff
@@ -515,11 +516,11 @@ private static class ContextSnapshotHelper {
 
 		public static Object writeReactorContext(Object returnValue) {
 			if (Mono.class.isAssignableFrom(returnValue.getClass())) {
-				ContextSnapshot snapshot = ContextSnapshot.captureAll();
+				ContextSnapshot snapshot = ContextSnapshotFactory.builder().build().captureAll();
```
rstoyanchev, line 519: I think there is no real reason to require 1.0.3 here. Couldn't we continue using the deprecated method a while longer and suppress the deprecation?

rstoyanchev (changes requested):
I'm only commenting on the change in spring-webmvc. I have not reviewed the spring-test changes.

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ReactiveTypeHandler.java
marcingrzejszczak: The build fails if I don't do this :shrug:

spring-test/spring-test.gradle
```diff
@@ -50,6 +50,8 @@ dependencies {
 	optional("io.projectreactor:reactor-test")
 	optional("org.jetbrains.kotlinx:kotlinx-coroutines-core")
 	optional("org.jetbrains.kotlinx:kotlinx-coroutines-reactor")
+	optional('io.micrometer:micrometer-observation')
+	optional('io.micrometer:context-propagation')
```
sbrannen, line 54: Please sort dependencies alphabetically in the build script.

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+import io.micrometer.observation.ObservationRegistry;
+import io.micrometer.observation.contextpropagation.ObservationThreadLocalAccessor;
+
+import org.springframework.context.ApplicationContext;
+import org.springframework.core.Conventions;
+import org.springframework.test.context.TestContext;
+import org.springframework.test.context.TestExecutionListener;
+
+/**
+ * {@code ObservationThreadLocalTestListener} is an extension of the {@link TestExecutionListener}
```
sbrannen, line 28:
```suggestion
 * {@code ObservationThreadLocalTestExecutionListener} is an implementation of the {@link TestExecutionListener}
```

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+ * SPI that updates the {@link ObservationThreadLocalAccessor} with the {@link ObservationRegistry}
+ * taken from the {@link ApplicationContext} present in the {@link TestContext}.
+ *
+ * This implementation is not thread-safe.
```
sbrannen, line 32:
```suggestion
 * <p>This implementation is not thread-safe.
```

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @author Marcin Grzejszczak
+ * @since 6.0
```
sbrannen, line 35:
```suggestion
 * @since 6.1
```

I imagine this would be added in 6.1.

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class ObservationThreadLocalTestListener implements TestExecutionListener {
+
+	/**
+	 * Attribute name for a {@link TestContext} attribute which contains the previously
+	 * set {@link ObservationRegistry} on the {@link ObservationThreadLocalAccessor}.
+	 * After all tests from the given test class get ran, the previously stored {@link ObservationRegistry}
```
sbrannen, line 42:
```suggestion
	 * <p>After all tests from the current test class have completed, the previously stored {@link ObservationRegistry}
```

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+	 * will be restored. If tests are ran concurrently this might cause issues
```
sbrannen, line 43:
```suggestion
	 * will be restored. If tests are executed concurrently this might cause issues
```

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+ * {@code ObservationThreadLocalTestListener} is an extension of the {@link TestExecutionListener}
+ * SPI that updates the {@link ObservationThreadLocalAccessor} with the {@link ObservationRegistry}
+ * taken from the {@link ApplicationContext} present in the {@link TestContext}.
+ *
+ * This implementation is not thread-safe.
+ *
+ * @author Marcin Grzejszczak
+ * @since 6.0
+ */
+public class ObservationThreadLocalTestListener implements TestExecutionListener {
```
sbrannen, line 37:
```suggestion
public class ObservationThreadLocalTestExecutionListener extends AbstractTestExecutionListener {
```

How do you envision this listener being registered?

Will users have to register it manually via `@TestExecutionListeners`? (I'd typically recommend against that)

Or should this listener be automatically registered as a default (like every other listener in `spring-test`)?

If the latter, this listener needs to be declared in `/spring-test/src/main/resources/META-INF/spring.factories`.

In any case, this listener should extend `AbstractTestExecutionListener` and provide a reasonable "order" value. See other concrete implementations in `spring-test` to determine which order to use (or ask for assistance if needed).

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+	 * set {@link ObservationRegistry} on the {@link ObservationThreadLocalAccessor}.
+	 * After all tests from the given test class get ran, the previously stored {@link ObservationRegistry}
+	 * will be restored. If tests are ran concurrently this might cause issues
+	 * unless the {@link ObservationRegistry} is always the same (which should be the case most frequently).
+	 */
+	public static final String PREVIOUS_OBSERVATION_REGISTRY = Conventions.getQualifiedAttributeName(
+			ObservationThreadLocalTestListener.class, "previousObservationRegistry");
+
+	@Override
+	public void beforeTestClass(TestContext testContext) throws Exception {
```
sbrannen, line 50: Please add method-level Javadoc.

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+		testContext.setAttribute(PREVIOUS_OBSERVATION_REGISTRY, ObservationThreadLocalAccessor.getInstance().getObservationRegistry());
+		testContext.getApplicationContext().getBeanProvider(ObservationRegistry.class)
+				.ifAvailable(observationRegistry -> ObservationThreadLocalAccessor.getInstance().setObservationRegistry(observationRegistry));
+	}
+
+	@Override
+	public void afterTestClass(TestContext testContext) throws Exception {
```
sbrannen, line 57: Please add method-level Javadoc.

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	public static final String PREVIOUS_OBSERVATION_REGISTRY = Conventions.getQualifiedAttributeName(
+			ObservationThreadLocalTestListener.class, "previousObservationRegistry");
+
+	@Override
+	public void beforeTestClass(TestContext testContext) throws Exception {
+		testContext.setAttribute(PREVIOUS_OBSERVATION_REGISTRY, ObservationThreadLocalAccessor.getInstance().getObservationRegistry());
+		testContext.getApplicationContext().getBeanProvider(ObservationRegistry.class)
+				.ifAvailable(observationRegistry -> ObservationThreadLocalAccessor.getInstance().setObservationRegistry(observationRegistry));
+	}
```
sbrannen, line 54: Please do your best to wrap "production" code around 100 characters.

spring-test/src/main/java/org/springframework/test/context/observation/package-info.java
```diff
@@ -0,0 +1,9 @@
+/**
+ * Micrometer Observation support classes for the <em>Spring TestContext Framework</em>.
```
sbrannen, line 2:
```suggestion
 * Observation support classes for the <em>Spring TestContext Framework</em>.
```

If this package is intended to provide support for observation use cases in general, I'm wondering if the name of the `TestExecutionListener` should include `Micrometer...`?

spring-test/src/test/java/org/springframework/test/context/observation/ObservationThreadLocalTestListenerTests.java
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.test.context.TestContext;
+
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.willAnswer;
+import static org.mockito.Mockito.mock;
+
+class ObservationThreadLocalTestListenerTests {
```
sbrannen, line 37:
```suggestion
class ObservationThreadLocalTestExecutionListenerTests {
```

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
```diff
@@ -0,0 +1,63 @@
...
+public class ObservationThreadLocalTestListener implements TestExecutionListener {
+
+	/**
+	 * Attribute name for a {@link TestContext} attribute which contains the previously
+	 * set {@link ObservationRegistry} on the {@link ObservationThreadLocalAccessor}.
+	 * After all tests from the given test class get ran, the previously stored {@link ObservationRegistry}
+	 * will be restored. If tests are ran concurrently this might cause issues
+	 * unless the {@link ObservationRegistry} is always the same (which should be the case most frequently).
+	 */
+	public static final String PREVIOUS_OBSERVATION_REGISTRY = Conventions.getQualifiedAttributeName(
```
sbrannen, line 46: Do you foresee that another `TestExecutionListener` would need to interact with the attribute with this name?

If not, please change the visibility to `private`.

sbrannen (changes requested):
Thanks for the PR.

I've made an initial review and requested several changes.

I've also raised a few points that need to be decided on.

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
marcingrzejszczak: We've made the whole train of changes in Micrometer Core (1.10.8), Context Propagation (1.0.3), Micrometer Tracing (1.0.7) and Reactor all in patch versions to fix bugs related to context propagation with reactor. We need this to be working in a patch release IMO. Without this change context propagation will not work properly. I would say that adding this component is a bug fix not an enhancement.

spring-test/src/main/java/org/springframework/test/context/observation/ObservationThreadLocalTestListener.java
marcingrzejszczak: I don't. I'll change this to private

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ReactiveTypeHandler.java
marcingrzejszczak: Yes, exactly. The current used method is marked as deprecated

spring-test/spring-test.gradle
marcingrzejszczak: I've changed the order

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ReactiveTypeHandler.java
marcingrzejszczak: Ah ok, let me fix that then

PR #30659, opened 2023-06-13
Add Http headers support in getForEntity

kooixh:
# Motivation

Create support for Http headers in `getForEntity` to allow simpler code. Currently if one needs to add headers in a `GET` call one would have to use `restTemplate.exchange()` which is verbose and could different from coding convention of using `restTemplate.postForEntity()`, etc. 

## Use

```jave
HttpHeaders headers = new HttpHeaders();
headers.setBearerAuth("abcd);
ResponseEntity<Object> res = restTemplate.getForEntity(url , headers, Object.class);
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestOperations.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestOperations.java
@@ -104,6 +104,20 @@ public interface RestOperations {
 			throws RestClientException;
 
 	/**
+	 * Retrieve an entity by doing a GET on the specified URL with the specified headers
+	 * The response is converted and stored in a {@link ResponseEntity}.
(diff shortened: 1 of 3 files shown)


rstoyanchev:
Thanks for the suggestion but we don't want to add even more methods to `RestTemplate`. We would also have to add the same for other HTTP methods, and that's not sustainable. This is why in 6.1 we're adding a new `RestClient` on the same infrastructure but with a fluent API. In the mean time, you can use the `exchange` method with a `RequestEntity`:

```java
RequestEntity<Void> requestEntity = RequestEntity
		.get(url)
		.headers(headers -> headers.setBearerAuth("abcd"))
		.build();

ResponseEntity<String> response = 
		new RestTemplate().exchange(requestEntity, String.class);
```

PR #30671, opened 2023-06-14
Add possibility to escape property placeholders

### LINKED ISSUE #9628: Support escaping prefix and separator in property placeholders [SPR-4953]
**[Chris Lee](https://jira.spring.io/secure/ViewProfile.jspa?name=chrislee)** opened **[SPR-4953](https://jira.spring.io/browse/SPR-4953?redirect=false)** and commented

There is presently no way to inject a placeholder expression (`${...}`) that won't be picked up by `PropertyPlaceholderConfigurer`. Ideally we should be able to inject a string that contains `${...}` for later use in its target bean without involvement from `PropertyPlaceholderConfigurer`.

Perhaps a simple escape mechanism using the conventional backslash, such as `"The current value is \\${...}"` would suffice.

---

**Affects:** 2.5.4, 2.5.5

11 votes, 5 watchers

bdshadow:
This commit allows to escape property placeholders to later processed by user. PropertyPlaceholderHelper allows to set custom escape character. "\" is used as the default one.

Closes gh-9628

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/PropertyPlaceholderHelper.java
+++ b/spring-core/src/main/java/org/springframework/util/PropertyPlaceholderHelper.java
@@ -63,6 +63,7 @@ public class PropertyPlaceholderHelper {
 
 	private final boolean ignoreUnresolvablePlaceholders;
 
+	private final char escapeCharacter;
 
 	/**
 	 * Creates a new {@code PropertyPlaceholderHelper} that uses the supplied prefix and suffix.
@@ -71,7 +72,23 @@ public class PropertyPlaceholderHelper {
 	 * @param placeholderSuffix the suffix that denotes the end of a placeholder
 	 */
 	public PropertyPlaceholderHelper(String placeholderPrefix, String placeholderSuffix) {
-		this(placeholderPrefix, placeholderSuffix, null, true);
+		this(placeholderPrefix, placeholderSuffix, null, true, '\\');
+	}
+
+	/**
+	 * Creates a new {@code PropertyPlaceholderHelper} that uses the supplied prefix and suffix.
+	 * Default escape character '\' is used.
+	 * @param placeholderPrefix the prefix that denotes the start of a placeholder
+	 * @param placeholderSuffix the suffix that denotes the end of a placeholder
+	 * @param valueSeparator the separating character between the placeholder variable
+	 * and the associated default value, if any
+	 * @param ignoreUnresolvablePlaceholders indicates whether unresolvable placeholders should
+	 * be ignored ({@code true}) or cause an exception ({@code false})
+	 */
+	public PropertyPlaceholderHelper(String placeholderPrefix, String placeholderSuffix,
+			@Nullable String valueSeparator, boolean ignoreUnresolvablePlaceholders) {
+
+		this(placeholderPrefix, placeholderSuffix, valueSeparator, ignoreUnresolvablePlaceholders, '\\');
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


snicoll:
Thanks for the PR. I've played with the code a bit and already polished quite a bit so I'll probably cherry-pick if you have time to update it. That doesn't work for default value having a property placeholder in it, something like:

```
${my.prop:\\${my.fallback}}
```

if `my.prop` it's trying to resolve `my.fallback` rather than using `${my.fallback}` as default value.

bdshadow:
@snicoll yes, i will fix it
thank you for the review

snicoll:
One thing I had in mind would be to have a data structure that represents a bit of text and some extra information. You could then handle those cases:

* It's just text and you output as is
* It's a placeholder and it needs to be resolved

Perhaps with a method that needs to "emit" the value and does the replacement by calling a function. I haven't really looked at the problem in details but it looks like we may remove the escape character "too soon" and we end up in another branch that is not aware that it should be escaped. Having a data structure may also simplify a bit of the code.

Food for thoughts, and thanks for looking at it!

bdshadow:
Tried for some time to solve it. The code really becomes ugly. The problem with the existing mr is indeed that the escape character is removed "too soon". While trying to fix it, I've come up with some more tests, which I'm not sure how they should work:
```
p1=abc
p2=\${undefined:${p1}}

p3=\${p1}
p4=\${undefined:${p3}}

p5=abc\${
```
What should `p2`, `p4` and `p5` be resolved to? 
Should `p2` become `${undefined:abc}`? Do we want to resolve nested placeholders inside an escaped placeholder? If yes, then `p4` must become `${undefined:${p1}}`?
A separate case is `p5`. Should it become "abc${"? Or should it be treated as not a placeholder and stay the same"abc\${"?
Probably these are questions to ask in the initial ticket

snicoll:
It's fine we can continue here as we're trying to figure out the best way to move forward with the code. Thanks for the follow-up and I am not surprised things get messy with corner cases.

To answer your questions:

* `p2=\${undefined:${p1}}`  should be `${undefined:${p1}}`as whatever that's inside the prefix/suffix that's escaped should be rendered as is.
* `\${undefined:${p3}}` same `${undefined:${p3}}`
* `abc\${` is a good question. It should be rendered as `abc\${` I guess since it's not a valid expression.

I don't know if the above helps or makes it even worse. Let me know :)

bdshadow:
Finally, i've updated it :)
Actually, i added 2 commits. The first one just fixes the problem we discussed above. 
In the second one, i tried to introduce a data structure, which you mentioned. I'm not sure if it really helps, but at least it lowers the complexity of the code, which sonar was complaining about. I tried to add 2 classes `Property` and `Placeholder`, but in my opinion, it led to more complexity as `Placeholder` can become a `Property` and there were unnecessary conversions back and forward all the time. 
I also thought about some static initializers like `Property.of(...), maybe it can simplify reading the code too.
Looking forward to your opinion and am ready to polish it further.

snicoll:
@bdshadow I want to acknowledge that I've seen this. Getting the espacing to work is quite involved as we both feared. If nobody beats me to it, I'll need some quality time to review and provide more feedback if needed. That won't happen before we start 6.2 in anger in a month or so.

bdshadow:
Thank you @snicoll. Totally understand. Looking forward to it

snicoll:
It turns out quite tricky and, while I don't have a definitive implementation for this, I did rewrite the parser from scratch as I felt this was the only way to address this issue and other issues reported against it. Thanks for the PR and your efforts.

PR #30696, opened 2023-06-20
Remove redundant assertion in `ReactorServerHttpResponse` constructor

### LINKED ISSUE #30686: A dead code may be found in 'org.springframework.http.server.reactive.ReactorServerHttpResponse' at line 59
Is the not null assertion necessary here? It seems like a dead code.
In  `org.springframework.http.server.reactive.ReactorServerHttpResponse`, line 59, if the response is null, JVM will throw NPE at line 58 and never go to line 59.  The assertion will not throw IllegalArgumentException with the specific message at any time.

```java
	public ReactorServerHttpResponse(HttpServerResponse response, DataBufferFactory bufferFactory) {
		super(bufferFactory, new HttpHeaders(new NettyHeadersAdapter(response.responseHeaders())));
		Assert.notNull(response, "HttpServerResponse must not be null");
		this.response = response;
	}
```

https://github.com/spring-projects/spring-framework/blob/842569c9e519ec7d11fc86691577cee067c7ef47/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpResponse.java#L59

kooixh:
Closes gh-30686

Remove redundant Assert


spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpResponse.java
```diff
@@ -56,7 +56,6 @@ class ReactorServerHttpResponse extends AbstractServerHttpResponse implements Ze
 
 	public ReactorServerHttpResponse(HttpServerResponse response, DataBufferFactory bufferFactory) {
 		super(bufferFactory, new HttpHeaders(new NettyHeadersAdapter(response.responseHeaders())));
```
sbrannen, line 58:
```suggestion
		super(bufferFactory, new HttpHeaders(new NettyHeadersAdapter(Objects.requireNonNull(response, "HttpServerResponse must not be null").responseHeaders())));
```

Although it will throw a `NullPointerException` instead of an `IllegalArgumentException`, if we use `Objects.requireNonNull` we can still have a custom error message.

So let's go with that.

sbrannen (changes requested):

spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpResponse.java
kooixh: Made the change

sbrannen (approved):

spring-webflux/src/main/java/org/springframework/web/reactive/result/method/InvocableHandlerMethod.java
sbrannen, line 1: Why did you make `spring-webflux/src/main/java/org/springframework/web/reactive/result/method/InvocableHandlerMethod.java` an "executable" file?

In any case, I will not merge that change.

sbrannen:
This has been merged into `main` in 374522464610a2e29354d9e1c9a4ec5d46afb8c8 and revised in 3d33d2baa9de61b07908695d3c4f1a9bcc8b0490.

Thanks

PR #30703, opened 2023-06-20
Add MultipartFile support for MultipartBuilder

OlgaMaciaszek:


rstoyanchev (commented):
`MultipartFile` also contains `name`, `orginalFileName`, and `contentType`, which would be presented in part headers. 

Thinking about this further though, `MultipartBodyBuilder` cannot support `MultipartFile` directly because it is in a lower level package. We'll need to move this support up to the level of HTTP interface client where we can resolve a `MultipartFile` and populate the `MultipartBodyBuilder`.

PR #30710, opened 2023-06-21
Use `Long.parseLong(CharSequence,int,int,int)` to avoid intermediate `String` creation

kilink:
Where possible, switch to the `Long.parseLong` variant that accepts a start and end index for the supplied `CharSequence`, thus avoiding making an unnecessary copy of the input.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/unit/DataSize.java
+++ b/spring-core/src/main/java/org/springframework/util/unit/DataSize.java
@@ -174,10 +174,11 @@ public final class DataSize implements Comparable<DataSize>, Serializable {
 	public static DataSize parse(CharSequence text, @Nullable DataUnit defaultUnit) {
 		Assert.notNull(text, "Text must not be null");
 		try {
-			Matcher matcher = DataSizeUtils.PATTERN.matcher(StringUtils.trimAllWhitespace(text));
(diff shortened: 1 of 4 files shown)


sbrannen:
@kilink, thanks for the PR.

Have you experienced any performance issues which inspired you to make the proposed changes?

kilink:
> @kilink, thanks for the PR.
> 
> Have you experienced any performance issues which inspired you to make the proposed changes?

I did not observe any issue in particular, but I had noticed that the `Integer.parseInteger` variant that takes start / end indices is being used throughout the codebase, and it seemed like the same optimization hadn't been applied for instances of `Long.parseLong` where it is possible to avoid creating a substring.

sbrannen:
> I did not observe any issue in particular, but I had noticed that the `Integer.parseInteger` variant that takes start / end indices is being used throughout the codebase, and it seemed like the same optimization hadn't been applied for instances of `Long.parseLong` where it is possible to avoid creating a substring.

Thanks for the feedback.

Indeed, we made that change for `Integer.parseInt` in conjunction with #27680.

In light of that, I'll schedule this improvement for 6.1 M4.

sbrannen:

Thanks

PR #30654, opened 2023-06-13
Add support for `private` init/destroy methods in AOT mode

essobedo:
## Motivation

In AOT mode, if the init/destroy methods are private and auto-discovered thanks to an annotation, the generated code sets the init and destroy method names with the fully qualified name of the declaring class as prefix to prevent naming conflicts which causes an error of the following type when starting the application (native image or with `-Dspring.aot.enabled=true`):

``` 
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'demoBean': Could not find an init method named 'com.example.demo.DemoBean.init' on bean with name 'demoBean'
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1770) ~[spring-beans-6.0.10-SNAPSHOT.jar!/:6.0.10-SNAPSHOT]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:598) ~[spring-beans-6.0.10-SNAPSHOT.jar!/:6.0.10-SNAPSHOT]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:520) ~[spring-beans-6.0.10-SNAPSHOT.jar!/:6.0.10-SNAPSHOT]
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:326) ~[spring-beans-6.0.10-SNAPSHOT.jar!/:6.0.10-SNAPSHOT]
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234) ~[spring-beans-6.0.10-SNAPSHOT.jar!/:6.0.10-SNAPSHOT]
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:324) ~[spring-beans-6.0.10-SNAPSHOT.jar!/:6.0.10-SNAPSHOT]
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:200) ~[spring-beans-6.0.10-SNAPSHOT.jar!/:6.0.10-SNAPSHOT]
[...]

Issue #30718, opened 2023-06-21
Package-private init/destroy methods are not always invoked


sbrannen:
## Overview

While working on #30692, I noticed that we currently do not invoke _shadowed_ package-private init/destroy methods declared in a different package.

Instead, we only invoke the last such init/destroy method found in the type hierarchy.

We should therefore track "non-visible" package-private init/destroy methods by their fully-qualified method name, analogous to how we track `private` init/destroy methods.

## Related Issues

- #10614

PR #30724, opened 2023-06-22
Support [package-]private init/destroy methods in AOT mode

sbrannen:
## Overview

Prior to this commit, private (and non-visible package-private) init/destroy methods were not supported in AOT mode. The reason is that such methods are tracked using their fully-qualified method names, and the AOT support for init/destroy methods previously did not take fully-qualified method names into account. In addition, the invocation order of init/destroy methods differed between standard JVM mode and AOT mode.

This commit addresses these issues in the following ways.

- `AbstractAutowireCapableBeanFactory.invokeCustomInitMethod()`, `DisposableBeanAdapter.determineDestroyMethod()`, and `BeanDefinitionPropertiesCodeGenerator.addInitDestroyHint()` now parse fully-qualified method names to locate the correct init/destroy methods.

- `AbstractAutowireCapableBeanFactory` and `DisposableBeanAdapter` delegate to a new `MethodDescriptor` record which encapsulates the parsing of fully qualified method names; however, `BeanDefinitionPropertiesCodeGenerator` duplicates this logic since it resides in a different package, and we do not currently want to make `MethodDescriptor` public.

- Init/destroy methods detected via annotations (such as `@PostConstruct` and `@PreDestroy`) are now invoked _prior_ to init/destroy methods that are explicitly configured by name or convention. This aligns with the invocation order in standard JVM mode; however, `InitializingBean#afterPropertiesSet()` and `DisposableBean#destroy()` are still invoked before annotated init/destroy methods in AOT mode which differs from standard JVM mode.

- Unit and integration tests have been updated to test the revised behavior.

## Related Issues

- #10614
- #28083
- #30654 
- #30755

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InitDestroyAnnotationBeanPostProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InitDestroyAnnotationBeanPostProcessor.java
@@ -182,7 +182,7 @@ public class InitDestroyAnnotationBeanPostProcessor implements DestructionAwareB
 	private static String[] safeMerge(@Nullable String[] existingNames, Collection<LifecycleMethod> detectedMethods) {
 		Stream<String> detectedNames = detectedMethods.stream().map(LifecycleMethod::getIdentifier);
 		Stream<String> mergedNames = (existingNames != null ?
-				Stream.concat(Stream.of(existingNames), detectedNames) : detectedNames);
+				Stream.concat(detectedNames, Stream.of(existingNames)) : detectedNames);
 		return mergedNames.distinct().toArray(String[]::new);
 	}
 
--- a/spring-beans/src/main/java/org/springframework/beans/factory/aot/BeanDefinitionPropertiesCodeGenerator.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/aot/BeanDefinitionPropertiesCodeGenerator.java
@@ -72,6 +72,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Phillip Webb
  * @author Stephane Nicoll
+ * @author Sam Brannen
  * @since 6.0
  */
 class BeanDefinitionPropertiesCodeGenerator {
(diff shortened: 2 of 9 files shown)


sbrannen:
@jhoeller and @snicoll, if either of you have time, I would appreciate a quick review of this PR.

In any case, I think we should decide if we want to address the following (with a separate GitHub issue).

> however, InitializingBean#afterPropertiesSet() and DisposableBean#destroy() are still invoked before annotated init/destroy methods in AOT mode which differs from standard JVM mode.

This difference in behavior stems from the fact that we merge init/destroy method names during AOT processing, and that causes `@PostConstruct`/`@PreDestroy` methods to be invoked _after_ `InitializingBean#afterPropertiesSet()` and `DisposableBean#destroy()` when running in AOT mode.

sbrannen:
Fixed in 3181dca5efe48b74b017d5461d7572af000ab52c

sbrannen:
Reopening to sure this has been tested in a native image, regarding reflection hints.

sbrannen:
> Reopening to ensure this has been tested in a native image, regarding reflection hints, etc.

I have confirmed that the fix works within a GraalVM native image by creating the sample application as described in #30654 and modifying the `build.gradle` script as follows.

```groovy
ext['spring-framework.version'] = '6.0.11-SNAPSHOT'

repositories {
	mavenCentral()
	maven {
		url "https://repo.spring.io/snapshot"
	}
}
```

To be thorough, I also verified the recently introduced support for shadowed package-private init/destroy methods (see #30718).

Using the following classes...

```java
// resides in a different package than the other beans
public class SubpackageSuperBean {

	@PostConstruct
	void init() {
		System.out.println("Init SubpackageSuperBean");
	}

	@PreDestroy
	void destroy() {
		System.out.println("Destroy SubpackageSuperBean");
	}

}
```

```java
public class SuperBean extends SubpackageSuperBean {

	@PostConstruct
	private void init() {
		System.out.println("Init SuperBean");
	}

	@PreDestroy
	private void destroy() {
		System.out.println("Destroy SuperBean");
	}

}
```

```java
public class DemoBean extends SuperBean {

	@PostConstruct
	private void init() {
		System.out.println("Init DemoBean");
	}

	@PreDestroy
	private void destroy() {
		System.out.println("Destroy DemoBean");
	}

}
```

... the output from the native executable is:

```
Init SubpackageSuperBean
Init SuperBean
Init DemoBean

Destroy DemoBean
Destroy SuperBean
Destroy SubpackageSuperBean
```

In light of that, I am closing this issue.

PR #30746, opened 2023-06-25
Fix remove the side effect of ConstructorArgumentValues#addIndexedArgumentValue using deep-copy.

remeio:
Fix remove the side effect of `ConstructorArgumentValues#addIndexedArgumentValue` using **deep-copy**.

Like `org.springframework.beans.factory.config.ConstructorArgumentValues#addArgumentValues`


jhoeller:
The current behavior of that `addIndexedArgumentValue` variant is intentional: It allows for registering existing `ValueHolder` instances in custom arrangements, just like `addGenericArgumentValue(ValueHolder)` does. If that is not desirable, either use one of the overloaded methods or call `.copy()` on the existing `ValueHolder` instance before passing it in.

In contrast, `addArgumentValues(ConstructorArgumentValues)` takes a full copy of a separate `ConstructorArgumentValues` instance with all of its `ValueHolder` instances. In such a scenario, the `ValueHolder` instances have not been individually passed in by the caller, hence the implicit deep copy there..

Issue #17162, opened 2014-12-19
Exception thrown by TransactionSynchronization#afterCompletion is eaten and there is no way to catch and perform any operation [SPR-12560]


spring-projects-issues:
**[Mohammad](https://jira.spring.io/secure/ViewProfile.jspa?name=adnan)** opened **[SPR-12560](https://jira.spring.io/browse/SPR-12560?redirect=false)** and commented

Exception thrown in TransactionSynchronization's afterComplete method doesn't be propagated back to calling method. Reason is org.springframework.transaction.support.TransactionSynchronizationUtils class invokeAfterCompletion has below implementation -

> public static void invokeAfterCompletion(List\<TransactionSynchronization> synchronizations, int completionStatus) {
> if (synchronizations != null) {
> for (TransactionSynchronization synchronization : synchronizations) {
> try {
> synchronization.afterCompletion(completionStatus);
> }
> catch (Throwable tsex) {
> logger.error("TransactionSynchronization.afterCompletion threw exception", tsex);
> }
> }
> }
> }

Since it is calling multiple synchronizer in a row hence they are eating exception and just logging it(which is clearly written as java doc). I think there should be a way to throw exception so that if required client code can perform some operation.

One solution would be to run afterComplete method in different threads and propagate exceptions with that thread. For programmatic transaction management, we would get know exception (Considering declarative transaction interceptors would not help as control won't come to client code)

Reference question in stackoverflow is - http://stackoverflow.com/questions/27563561/transactionsynchronization-sallow-runtimeexception-that-is-thrown-while-aftercom?noredirect=1#comment43558721_27563561

---

**Affects:** 4.0.3

spring-issuemaster:
Bulk closing outdated, unresolved issues. Please, reopen if still relevant.

slinstaedt:
`org.springframework:spring-tx:jar:5.3.18` which should be part of spring boot 2.5.12.

KSH-code:
it was fixed by https://github.com/spring-projects/spring-framework/pull/30776
so it is included in https://github.com/spring-projects/spring-framework/releases/tag/v6.0.11 and https://github.com/spring-projects/spring-framework/releases/tag/v5.3.29

bclozel:
Thanks @KSH-code !

While the original issue was more about making that exception available through a callback of some sort, the reason for reopening it is indeed about the logging level.

Closing this issue as a superseded by #30776

stefanrybacki:
[...]

PR #30776, opened 2023-06-30
Fix log level on error with @TransactionalEventListener

KSH-code:
Synchronous methods with `@TransactionalEventListener` logs in debug level.
However, most production logs are filtered by error level. Thus most developers cannot detect `@TransactionalEventListener` method throws error whether or not.

It was changed from https://github.com/spring-projects/spring-framework/commit/95110d825715a70dbc7fadc0e3ebd42f44e6bdfb. I think this changed an error because it is not related to  commit message.

ref: https://github.com/spring-projects/spring-framework/issues/17162


jhoeller:
Good point, it is not clear why that log level got downgraded to debug, this might have been an accident indeed. While regular transaction synchronizations are not meant to throw an exception there to begin with, this may certainly happen for transactional event processing and should be visible there.

PR #30786, opened 2023-06-30
fix(Beans): support kotlin context receiver in constructor

alexismanin:
Includes a minor change to improve support for kotlin context receiver.
I mark this PR as a draft for now because context receiver is still experimental so:

1. There might be changes in the future in context receivers, making this PR obsolete (or at least requires to modify it)
2. I cannot add unit tests without activating the experimental feature, which I'd rather not.

The PR only fix constructor call when a class is annotated with a context receiver.
Concretely, it means that this PR allows:

```kt
context(AnotherBeanType)
@Component class MyBean() {
    // ... class content ...
}
```

Note that, the above construct works on a project **without** Kotlin support. In such case, no Kotlin reflection is involved, and the "java" constructor is called instead.

For those interested in testing, I've setup [a small demo project](https://github.com/alexismanin/spring-kt-ctx-receiver-demo) that:

 - Use context receiver on @Bean annotated method (already works with Spring-Boot 3.1.1)
 - Contains commented constructor based component (because a change is required for it to work)
 - Contains a branch `feat/by-constructor` that shows constructor binding works when kotlin spring plugin is removed from build (but it forces to open all classes manually).

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -873,13 +873,17 @@ public abstract class BeanUtils {
 			}
 
 			List<KParameter> parameters = kotlinConstructor.getParameters();
-
-			Assert.isTrue(args.length <= parameters.size(),
-					"Number of provided arguments must be less than or equal to the number of constructor parameters");
-			if (parameters.isEmpty()) {
+			final int nbExpectedParameters = parameters.size();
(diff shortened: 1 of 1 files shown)


sdeleuze:
Given the status of the discussion on Kotlin side where everything is still being discussed on design side, I have the feeling we are still pretty far from something we can support in Spring. A consequence, I close this PR unmerged. Let's reconsider later when there is something more concrete.

PR #30788, opened 2023-06-30
Allows JdkClientHttpRequest.DISALLOWED_HEADERS to be customized.

### LINKED ISSUE #30787: Allow JdkClientHttpRequest.DISALLOWED_HEADERS to be customized
Since https://bugs.openjdk.org/browse/JDK-8213696 the JDK `HttpClient`'s list of disallowed headers can be customized by setting `jdk.httpclient.allowRestrictedHeaders=host`. This is a feature many users desire in a proxy, including Spring Cloud Gateway. Maybe `JdkClientHttpRequest` can read the same property since it silently skips the header where, I think, `HttpClient` fails with an exception.

/cc @poutsma

spencergibb:
It does this by reading the System property `jdk.httpclient.allowRestrictedHeaders` and removing those set from the default list.

Fixes gh-30787

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
@@ -25,7 +25,9 @@ import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.ByteBuffer;
 import java.time.Duration;
-import java.util.List;
+import java.util.Collections;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Flow;
 
@@ -33,6 +35,7 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.lang.Nullable;
 import org.springframework.util.StreamUtils;
+import org.springframework.util.StringUtils;
 
 /**
  * {@link ClientHttpRequest} implementation based the Java {@link HttpClient}.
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
```diff
@@ -48,8 +51,19 @@ class JdkClientHttpRequest extends AbstractStreamingClientHttpRequest {
...
 	 */
-	private static final List<String> DISALLOWED_HEADERS =
-			List.of("connection", "content-length", "expect", "host", "upgrade");
+	protected static final Set<String> DISALLOWED_HEADERS = getDisallowedHeaders();
+
+	private static Set<String> getDisallowedHeaders() {
+		TreeSet<String> headers = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+		headers.addAll(Set.of("connection", "content-length", "expect", "host", "upgrade"));
+
+		String headersToAllow = System.getProperty("jdk.httpclient.allowRestrictedHeaders");
```
spencergibb, line 60: Maybe there is a setter somewhere and we set the property rather than trying to read it?

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
```diff
@@ -48,8 +51,19 @@ class JdkClientHttpRequest extends AbstractStreamingClientHttpRequest {
 	 * The JDK HttpRequest doesn't allow all headers to be set. The named headers are taken from the default
 	 * implementation for HttpRequest.
 	 */
-	private static final List<String> DISALLOWED_HEADERS =
-			List.of("connection", "content-length", "expect", "host", "upgrade");
+	protected static final Set<String> DISALLOWED_HEADERS = getDisallowedHeaders();
```
spencergibb, line 54: `protected` was changed to test without much setup.

spencergibb: `AbstractMockWebServerTests` dispatch for `/echo` checks that `Host` is `localhost:port`, so unsure how to test otherwise.

PR #30793, opened 2023-07-02
Update inline comment in `ProxyTransactionManagementConfiguration`

shin-mallang:
The original comment explained that protected methods allow @Transactional.
However, in practice, default method is also allowed.
So this should be clarified to avoid confusion.


sbrannen:
Hi @shin-mallang,

> The original comment explained that protected methods allow @transactional. However, in practice, default method is also allowed. So this should be clarified to avoid confusion.

Interface `default` methods are always supported by `AnnotationTransactionAttributeSource`, since interface `default` methods are `public`.

The comment in question was added to explain why the `AnnotationTransactionAttributeSource` constructor is now (since 6.0) invoked with `false` -- which sets the `publicMethodsOnly` flag to `false` (which enables support for `protected` `@Transactional` methods).

shin-mallang:
@sbrannen 

I wasn't referring to the interface's default methods, I meant the methods whose access modifiers are package private.
I think this needs to be added in a separate comment, doesn't it?

PR #30803, opened 2023-07-04
Use `Collections.emptyEnumeration` where appropriate

kilink:
Use the empty `Enumeration` singleton returned by `Collections.emptyEnumeration`; there were a couple places that explicitly passed an empty collection to `Collections.enumeration`, which may as well just use `Collections.emptyEnumeration`.


sbrannen:

Thanks

PR #30825, opened 2023-07-06
Update inline comment in ProxyTransactionManagementConfiguration

shin-mallang:
I apologize for the lack of clarity in my previous PR(#30793). 

The original comment explained that protected methods allow @transactional.
However, in practice, `package-private` method is also allowed.
So this should be clarified to avoid confusion.


sbrannen:
Hi @shin-mallang,

Thanks for clarifying that you meant "package-private" when you said "default" in the original PR.

> So this should be clarified to avoid confusion.

The comment in question is an internal comment for maintainers of the framework to point out the reason for invoking the constructor with `false`, and it sufficiently conveys that.

Since the comment is not user-facing (such as Javadoc or the reference manual), I don't think it warrants spending so much time on this internal comment. Plus we would need to update `ProxyCachingConfiguration` as well if we wanted to be consistent.

However, if you find places in Javadoc or in the reference manual that incorrectly state that `@Transactional`, `@Cacheable`, `@CachePut`, etc. can only be used on `public` methods, feel free to open a PR to address that.

Issue #29249, opened 2022-10-04
kotlin script template tests are broken on Java 19


bclozel:
When running `org.springframework.web.reactive.result.view.script.KotlinScriptTemplateTests` on Java 19, we're getting the following failure:

```
Failed to evaluate script [org/springframework/web/reactive/result/view/script/kotlin/render.kts]
java.lang.IllegalStateException: Failed to evaluate script [org/springframework/web/reactive/result/view/script/kotlin/render.kts]
	at org.springframework.web.reactive.result.view.script.ScriptTemplateView.loadScripts(ScriptTemplateView.java:302)
	at org.springframework.web.reactive.result.view.script.ScriptTemplateView.createEngineFromName(ScriptTemplateView.java:276)
	at org.springframework.web.reactive.result.view.script.ScriptTemplateView.setApplicationContext(ScriptTemplateView.java:241)
	at org.springframework.web.reactive.result.view.script.KotlinScriptTemplateTests.createViewWithUrl(KotlinScriptTemplateTests.java:91)
	at org.springframework.web.reactive.result.view.script.KotlinScriptTemplateTests.render(KotlinScriptTemplateTests.java:78)
	at org.springframework.web.reactive.result.view.script.KotlinScriptTemplateTests.renderTemplateWithFrenchLocale(KotlinScriptTemplateTests.java:48)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at 
...
org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:71)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
Caused by: javax.script.ScriptException: ERROR Cannot inline bytecode built with JVM target 17 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option (ScriptingHost574059d5_Line_0.kts:11:58)
[...]

PR #30896, opened 2023-07-15
Enable `KotlinScriptTemplateTests` with Java 20

izeye:
This PR changes to enable the `KotlinScriptTemplateTests` with Java 20 as 66a1be2d86f9e56e41086a07245e31d099616303 says that Java 21 is not supported and `./gradlew :spring-webflux:test -PtestToolchain=20 --tests KotlinScriptTemplateTests` seems to work with this change.

See gh-29249


sbrannen:
Good catch! 👍 

Though you missed `KotlinScriptTemplateTests` in `spring-webmvc`.

No worries, I updated that in 0611192dacc995891fd99904d2c4273fe9212d28.

Issue #26489, opened 2021-02-01
Spring WebClient multipart/form-data does not include a content length


JoeBalduz:
**Affects:** Spring Boot 2.4.2

---

Hi, I've noticed that multipart/form-data does not include a content length when the request is created with multipart/form-data
. This has caused an issue with trying to access a route that requires a content-length to be set or else I get a 411. I've created a small demo to recreate the problem which can found here https://github.com/JoeBalduz/spring-webclient-multipart-no-content-length.

I also saw that #26046 appears to be the same issue as mine but it is closed. Is this possibly something that can be fixed in a future version so that the content-length is set automatically for multipart/form-data? I've tried setting the content-length myself in multiple ways but the closest I got was being off by about 10 bytes.

rstoyanchev:
It is not easy to compute a content-length. It requires serializing and buffering the entire content in memory, which is hardly ideal for a multipart request in particular. Granted we may know the length of a `Resource` (e.g. a file), but we'd still need to serialize headers, multipart boundaries, and all other parts with their content to know the content-length of the entire request.

Can you elaborate on "trying to access a route that requires a content-length"? For a multipart request that doesn't seem like a reasonable requirement.

Keep in mind also that we do set the content-length on individual parts where the length is know. For example for a file `Resource` I would expect the part headers to have a content-length. That is a more feasible expectation I think.

JoeBalduz:
There's an API route I need to access where if the Content-Length header isn't set on the http request, I get a 411 error and I'm not able to upload my file, along with a string, to that route. I don't control that API, so I've had to switch to using RestTemplate in order to send that request however I'd prefer to not use RestTemplate since it is blocking.

rstoyanchev:
The `FormHttpMessageConverter` also doesn't set a content-length for a multipart request. Can you elaborate?

JoeBalduz:
I'm not exactly sure where the content-length header is being set when I use RestTemplate but I can see in wireshark that it has a content-length header and I'm no longer getting the 411 error when sending a request to the API I'm accessing and the request is working. Here is how I'm using RestTemplate:
`
val headers = HttpHeaders()
headers.contentType = MediaType.MULTIPART_FORM_DATA
val httpEntity = HttpEntity(multipartBodyBuilder.build(), headers)
val response = restTemplate.postForEntity("url/im/accessing", httpEntity, String::class.java)
`

JoeBalduz:
[...]

PR #30899, opened 2023-07-16
Add example of buffering with WebClient to the reference documentation

### LINKED ISSUE #28758: Add example of buffering with WebClient to the reference documentation
The code samples in #26489 are very useful and I keep coming back to them. Recently I forgot the exact title and spent a while to remember how to find it. Perhaps it is a good candidate to add to the documentation, perhaps in [this section][section]? I would be willling to create that PR but I figured it probably makes sense to ask first.

[section]: https://docs.spring.io/spring-framework/docs/5.2.22.RELEASE/spring-framework-reference/web-reactive.html#webflux-client-filter

Nheyll:
Resolves #28758

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webflux-webclient/client-filter.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webflux-webclient/client-filter.adoc
@@ -148,5 +148,73 @@ Kotlin::
 ----
 ======
 
+The example below demonstrates how to use the `ExchangeFilterFunction` interface to create
+a custom filter class that helps with computing a `Content-Length` header for `PUT` and `POST`
+`multipart/form-data` requests using buffering.
 
+[tabs]
+======
+Java::
++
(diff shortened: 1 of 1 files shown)


sdeleuze:
Thanks for this contribution, merged via 411b355f2c47c5835329225032645f4b2ee40f9d and refined via 3ccbe3d1cba54840a8f7beb22f911ac3eb94e683.

Pro456458:
Hi I come across to this issue  [26489](https://github.com/spring-projects/spring-framework/issues/26489) "**Spring WebClient multipart/form-data does not include a content length**" . Provided solution working fine and inside ExchangeFilterFunction, content-length is well calculated in case of uploading file in a single chunk. But in case of uploading file in multiple chunk (for large size) it started failing and giving exception 
{The length of the file input stream does not match the provided length. Expected: X, Actual (approximate): 'Y' "} 
Could @sdeleuze @spring-projects-issues  @Nheyll @nertzy any one help me to provide the solution how to calculate the content for each chunk

PR #30903, opened 2023-07-17
Move blocking view resolution to boundedElastic

Arooba-git:
Hi! :) 

We used BlockHound to identify blocking operations in reactive spring modules; it turns out resolveViewName operation is blocking the reactive pipeline in webflux:

Screen Shot 2023-07-17 at 7 53 44 PM
This PR fixes the blocking call as well as updates a test to support the updated, reactive version.

The changes have been validated with build and test commands as mention in guidelines. 👍


poutsma:
The problem with the changes made by this PR is that *all* view resolution is done one the bounded elastic scheduler; not just blocking ones like FreeMarker. And because context switches can be expensive, it is not desirable for this to happen.

I will address the concern underlying this PR (i.e. blocking template resolution when using FreeMarker) by localizing the usage of the bounded elastic scheduler to just the blocking operations in `FreeMarkerView`, and not all view resolutions.

PR #30916, opened 2023-07-20
Correct inaccuracy in spring docs

luchob:
Spring uses not only `BeanPostProcessor`-s to process callback interfaces. For example `AbstractAutowireCapableBeanFactory#initializeBean` is calling `afterPropertiesSet` 
on the `InitializingBean`-s.

Verified for at least `spring-beans` `v6.0.10`


snicoll:
@luchob the intention of the paragraph is to explain that the framework uses `BeanPostProcessor` internally to implement the features exposed there, and that users can implement custom callback with a similar mechanism. While you are correct, I don't see a value of providing additional details that serve no purpose in the context of that section.

PR #30928, opened 2023-07-23
Move blocking writeResource operation to boundedElastic

Arooba-git:
Hi! 🙂 

We detected a blocking call in http-codec module with the help of BlockHound:
sf2-blocking

This PR fixes the blocking code so the pipeline remains reactive end to end. :)  We re-ran the tests and also compared the performance (CPU usage and latency) before and after the fix:

**Before**
sf-latency-before
sf2-cpu-bef

**After**
sf2-cpu-aft
sf2-latency-after


poutsma:
As with #30903, this PR uses the bounded elastic scheduler for all writing operations, including zero-copy operations that do not burden the CPU. I will address this by localizing the usage of the bounded elastic scheduler to just the blocking operations.

PR #30934, opened 2023-07-24
Skip array sort when the length of array not greater than 1

remeio:
Performance optimization(`org.springframework.beans.factory.support.DefaultListableBeanFactory#resolveMultipleBeans`):

For the sorting condition of `Collection`  is  `result instanceof List<?> list && list.size() > 1`, when the size of collection isn't greater than 1, it will skip sorting.

Similarly, skip array sort when the length of array not greater than 1.


sbrannen:

Thanks

PR #30936, opened 2023-07-24
Support `@RSocketExchange` for responding

### LINKED ISSUE #29240: Support `@RSocketExchange` for annotated responders
It's better to addopt unifed annotation contract for both server and client side. And I think the [RSocket Interface annotation ](https://docs.spring.io/spring-framework/docs/6.0.0-M6/reference/html/web-reactive.html#rsocket-interface)  is more powerful and meaningful

OlgaMaciaszek:
Fixes gh-29240.

### DIFF
--- a/spring-messaging/src/test/java/org/springframework/messaging/rsocket/service/RSocketServiceIntegrationTests.java
+++ b/spring-messaging/src/test/java/org/springframework/messaging/rsocket/service/RSocketServiceIntegrationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.messaging.rsocket.service;
 
 import java.time.Duration;
+import java.util.stream.Stream;
 
 import io.rsocket.SocketAcceptor;
 import io.rsocket.core.RSocketServer;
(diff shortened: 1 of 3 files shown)


spring-messaging/src/test/java/org/springframework/messaging/rsocket/service/RSocketServiceIntegrationTests.java
```diff
@@ -124,6 +145,23 @@ interface Service {
...
+		public Mono<String> echoAsync(String payload) {
+			return Mono.delay(Duration.ofMillis(10)).map(aLong -> payload + " async");
+		}
+
+		@Override
+		public Flux<String> echoStream(String payload) {
+			return Flux.interval(Duration.ofMillis(10)).map(aLong -> payload + " " + aLong);
+		}
+
+	}
```
rstoyanchev, line 162: I don't think we need both `ServerController` and `ExchangeServerController`. `@MessageMapping` methods already have test coverage, and it is not a goal here to test that. We're instead testing requesting with `@RSocketExchange` methods. Now we can also test responding. 

I suggest keeping the test class as it was before, but simply replacing `@MessageMapping` with `@RSocketExchange`.

rstoyanchev (changes requested):
For the documentation, while it is logical to mention `@RSocketExchange` alongside `@MessageMapping`, I'm not sure I like the result. It gets repetitive and becomes a distraction.

I think it would be better to create a dedicated section under "Annotated Responders" called `@RSocketExchange`, probably right after [@MessageMapping](https://docs.spring.io/spring-framework/reference/rsocket.html#rsocket-annot-messagemapping) where we can properly explain all that needs explaining. For a start what it is in contrast to `@MessageMapping` (with a reference to the RSocket Service section), that it is is supported for responding as an alternative to `@MessageMapping` so you can essentially use it for requesting and responding, that it supports very similar but slightly different method signature as it needs to remain neutral to requester vs responder, and so on. The RSocket Service proxy section could also mention that the annotation is supported for responding to make sure that it's discoverable. Other than that I don't think we need to mention it everywhere alongside `@MessageMapping`. This will become even more obvious with the same changes for `@HttpExchange` since we have much more extensive documentation that mentions `@RequestMapping` many times.

spring-messaging/src/test/java/org/springframework/messaging/rsocket/service/RSocketServiceIntegrationTests.java
OlgaMaciaszek: I have removed the new controller and made the existing one implement the `Service` without annotating it directly; also have slightly modified the service to additionally verify class-level `@RSocketExchange`. Let me know if that looks good to you.

PR #30942, opened 2023-07-25
Use `readNBytes` in `StringHttpMessageConverter` when `contentLength` is available

kilink:
When the content length is known, use readNBytes on the InputStream in StringHttpMessageConverter, which avoids some extra copying and allocations.

This is essentially the same optimization as #30010.


sbrannen:
This has been merged into `6.0.x` and `main` in 7636eecb481d9d5ec9a145ec800dec09b90743da and revised in c0506422908ef0d985c0133f7e1c1059e6cb34e1.

Thanks

PR #30456, opened 2023-05-09
Fix String comparison issue in `SpringExtension`

sigee:

PR #30963, opened 2023-07-28
`IllegalStateException` cannot be thrown in `SpringExtension`

### LINKED ISSUE #30962: IllegalStateException Cannot be thrown
**Issue Overview:**

In the current implementation of the validateAutowiredConfig method, there is an issue with the string comparison when checking for violations. This issue can lead to unexpected behavior and potential exceptions during runtime.

**Issue Description:**
The `validateAutowiredConfig` method is responsible for validating the presence of certain annotations on test methods and test lifecycle methods within a test class. It ensures that no methods are annotated with `@Autowired` and instead suggests using annotations like `@Qualifier` or `@Value` on individual method parameters.

In the original code, the validation of violations relies on comparing the errorMessage string with NO_VIOLATIONS_DETECTED. However, this comparison is done using the != operator, which might not work as intended for string comparison. This is because the != operator checks for reference equality (whether both variables point to the same memory location), not the content of the strings. Consequently, if the errorMessage and NO_VIOLATIONS_DETECTED strings have different memory locations (which is very likely in this case), the condition will not be satisfied even if the errorMessage string has the same content as NO_VIOLATIONS_DETECTED.

Similarly, in the `validateRecordApplicationEventsConfig` method, there is a similar issue with the string comparison of the `errorMessage` variable. The original code used the != operator, and I have corrected it by using the equals method.

These fixes ensure that the methods correctly handle the error message comparison and throw an IllegalStateException only when necessary.

```
if (errorMessage != NO_VIOLATIONS_DETECTED) {
    throw new IllegalStateException(errorMessage);
}

```

**Affects:** 6.0.x

**Fix:**
The proposed fix involves changing the string comparison from != to .equals() to compare the content of the two strings properly.

```
if (!errorMessage.equals(NO_VIOLATIONS_DETECTED)) {
    throw new IllegalStateException(errorMessage);
}

```

alidandach:
Fix incorrect string comparison:

The issue occurred when checking for the presence of specific error messages that determine whether to throw an IllegalStateException.

closes #30962

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/junit/jupiter/SpringExtension.java
+++ b/spring-test/src/main/java/org/springframework/test/context/junit/jupiter/SpringExtension.java
@@ -172,7 +172,7 @@ public class SpringExtension implements BeforeAllCallback, AfterAllCallback, Tes
 								testClass.getName(), Arrays.toString(methodsWithErrors)));
 			}, String.class);
 
-		if (errorMessage != NO_VIOLATIONS_DETECTED) {
+		if (!errorMessage.equals(NO_VIOLATIONS_DETECTED)) {
 			throw new IllegalStateException(errorMessage);
 		}
 	}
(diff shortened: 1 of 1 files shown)


sbrannen:
- duplicate of #30456

Issue #30913, opened 2023-07-19
Support `@HttpExchange` for server side handling


rstoyanchev:
While `@HttpExchange` was initially created for client use, the idea was always that an `@HttpExchange` interface is an HTTP service contract, neutral to client or server. 

In addition to simplifying client code, there are also cases where a shared interface between client and server makes provides an easy way for clients to access server APIs and keep up with changes. This leads to increased coupling that won't be a good fit for a public API but may be exactly the goal for internal APIs. This is commonly used with Spring Cloud and OpenFeign as described in https://github.com/spring-projects/spring-framework/issues/30733#issuecomment-1613576238.

Similar support is also planned for the RSocket interface client in #29240.

OlgaMaciaszek:
I'd like to work on that.

rstoyanchev:
Superseded by #30913.

Issue #30733, opened 2023-06-23
The `HttpServiceProxyFactory` allows for the customization of request metadata generation


DanielLiu1123:
The current `HttpServiceProxyFactory` only supports handling annotations based on `@HttpExchange`. However, in certain scenarios, users may need to generate HTTP request based on custom annotations. For example, some `Spring Cloud OpenFeign` users prefer or have already been accustomed to using annotations based on `@RequestMapping`, or some may want to generate HTTP request based on `jakarta.ws.rs-api` annotations.

If the framework allows users to determine the logic for generating request metadata instead of being "hardcoded" within the framework, it would greatly enhance its extensibility.

rstoyanchev:
In general nothing against flexibility, but specifically for `@RequestMapping` annotations, those were designed for server side use and their semantics are not neutral to client vs server use. Mapping with URL patterns vs declaring a concrete URL is just one example, but there are many others.

We created `@HttpExchange` annotations specifically for this reason. The method signature is largely compatible with `@RequestMapping` in terms of supported annotations and method arguments, but far from the same. The former is a subset of the latter, as it is focused on declaring a concrete request rather than on mapping requests to a server handler. 

Theoretically, given that `@HttpExchange` is neutral to client vs server, and so it could be supported it on the server side as an alternative to `@RequestMapping`. You would just need to accept working with a more limited range of options for the method signature. The other way around, however, supporting `@RequestMapping` for client side use, is not something we are going to do. It's what Spring Cloud Feign did and caused many issues.

Keep in mind also that the annotation and its attributes is just one side. The other is support for specific types of arguments and annotations. In that sense, you can't really only customize the annotation. You also need to support its semantics, including argument types and annotations. In other words it wouldn't be that simple to support `jakarta.ws.rs-api` annotations.

Is it just what you are accustomed or something, or is there some other reason for this?

DanielLiu1123:
Thank you for your response.

Firstly, I understand that `@RequestMapping` is specifically designed for server-side implementation, and it has numerous attributes that are not suitable for the client-side. However, my intention is not to support `@RequestMapping` as a client-side implementation, but rather to provide the capability to support other annotations.

[...]

PR #30980, opened 2023-08-02
Support `@HttpExchange` for server-side handling 

OlgaMaciaszek:


quaff:
I would like using `RequestMapping` instead of `HttpExchange` like feign does, not contrary.

OlgaMaciaszek:
@quaff This is not something we're planning to support. Please see more details in @rstoyanchev 's [comment](https://github.com/spring-projects/spring-framework/issues/30733#issuecomment-1609276552).

OlgaMaciaszek:
Fixes gh-30913.

PR #30995, opened 2023-08-04
Switching thread context classloader during mvc method execution

graceyu-own:
My project consists of a spring boot project and multiple extension packages (which can be understood as plugin).

Each extension package has its own spring application context and can use MVC related functions (such as defining controllers, interceptors, etc.), the extension package loader will load the Jar package through a custom URLCLassLoader and read the handlerMethods and interceptors in RequestMappingHandlerMapping and register them in the main springboot RequestMappingHandlerMapping.

When the main springboot program receives a request call and is scheduled through the DispatcherServlet, it will call the handler method registered with the extension package. At this time, the thread context classloader does not switch to the classLoader where the extension package is located.

At this point, the code implemented within the method may experience a ClassNotFoundException.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java
+++ b/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java
@@ -241,7 +241,9 @@ public class InvocableHandlerMethod extends HandlerMethod {
 	@Nullable
 	protected Object doInvoke(Object... args) throws Exception {
 		Method method = getBridgedMethod();
+		ClassLoader beforeClassLoader = Thread.currentThread().getContextClassLoader();
 		try {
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks for the contribution, but I don't think we can change the current class loader for all Spring applications. Most applications will not use this plugin mechanism and we can't inflict such a change on all.

Maybe try to [hook your plugin using a `HandlerInterceptor`](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-servlet/handlermapping-interceptor.html)?

graceyu-own:
Thank you for your suggestion, but the application context of my extension packages does not have a servlet context.
I attempted to rewrite the createInvocableHandlerMethod of RequestMappingHandlerAdapter, it can solve my problem.
-----------------------------------
image

PR #31012, opened 2023-08-08
Fixed an error connecting to another database ignoring the "spring.datasource.url" setting in "application.properties"

pyg410:
Fixed a bug where "catalogName" in "TableMetaData" and "DatabaseMetaData" were connected to different databases by directly putting data into the "tableMeta" variable without verifying whether the database names were the same.

1. problem situation
I had 2 databases (fast_sns, jdbchw).
The database "jdbchw" was connected even though "fast_sns" was specified in "spring.datasource.url" of "application.properties".
As a result of debugging, I found that it was a problem with the "next()" method of the While statement, and added a code to verify it.

스크린샷 2023-08-08 오후 8 57 52

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/GenericTableMetaDataProvider.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/GenericTableMetaDataProvider.java
@@ -20,11 +20,7 @@ import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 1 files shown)


snicoll:
@pyg410 thanks for the PR but we can't really consider it without test cases that cover the problem you've described.

spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/GenericTableMetaDataProvider.java
```diff
@@ -321,6 +317,9 @@ private void locateTableAndProcessMetaData(DatabaseMetaData databaseMetaData,
 				tmd.setCatalogName(tables.getString("TABLE_CAT"));
 				tmd.setSchemaName(tables.getString("TABLE_SCHEM"));
 				tmd.setTableName(tables.getString("TABLE_NAME"));
+				if(!Objects.equals(tmd.catalogName, tables.getString("TABLE_CAT"))){
+					continue;
```
qiangmao, line 321: `public class Main {
    public static void main(Stirng[] args) {
    }
}`

PR #31015, opened 2023-08-09
Change Kotlin `Any` to be a nullable type in AOP refdoc examples

pstrsr:
When using "Any" as return type, as suggested by the [documentation example](https://docs.spring.io/spring-framework/reference/core/aop/ataspectj/example.html), in a `@Around`-Aspect calls to void methods with `pjp.proceed()` will fail, with a cryptic NullPointerException error:

`java.lang.NullPointerException: pjp.proceed() must not be null`

I have made an example project to showcase this behavior:
[spring-aop-docs-bug
](https://github.com/AlmostFamiliar/spring-aop-docs-bug)


sdeleuze:
Merged with additional similar changes, thanks!

PR #31016, opened 2023-08-09
refactor: rename variable name with 'camelCase' style.

remeio:
from `_sbr` to `sbr`.


snicoll:
@remeio thanks for the PR but there are a lot of places where we'd have to replace this for consistency. Renaming the variable in one file is a stylistic change that we prefer to apply ourselves ayway.

PR #30745, opened 2023-06-25
Improve documentation on inferred destruction callbacks

Nheyll:
I gave the #29546 a try.

- I divided [_Destruction Callbacks_](https://docs.spring.io/spring-framework/docs/5.3.x/reference/html/core.html#beans-factory-lifecycle-disposablebean) section into two subsections, one for the inference of destroy methods and the other for the existing documentation on destruction callbacks. 
- I took the liberty to call the latter _Explicit destruction methods_ in opposition to _Inferred destruction methods_.
- _Inferred destruction methods_ is written with a Java-first approach, with the the help of the [_Receiving Lifecycle Callbacks_ subsection](https://docs.spring.io/spring-framework/docs/5.3.x/reference/html/core.html#beans-java-lifecycle-callbacks) which now references _Inferred destruction methods_.
- I updated some references due to the addition of a new subsection.

PR #31025, opened 2023-08-10
Improve documentation on inferred destruction callbacks #30745

harshal-gav:
In the "Java" code snippet, the class `ExampleBean` is missing the `implements DisposableBean` declaration. Correction: Ensure the class implements the `DisposableBean` interface.

1. Added : `DisposableBean` after the class name to indicate that the class is implementing the `DisposableBean` interface. 
2. Added the `override` keyword before the destroy method to indicate that it's being overridden from the `DisposableBean` interface. 
3. Retained the `cleanup` function as-is, in case you have additional cleanup tasks you want to perform that are specific to this class.

### DIFF
--- a/framework-docs/modules/ROOT/pages/core/beans/factory-nature.adoc
+++ b/framework-docs/modules/ROOT/pages/core/beans/factory-nature.adoc
@@ -189,24 +189,28 @@ Java::
 +
 [source,java,indent=0,subs="verbatim,quotes",role="primary"]
 ----
-	public class ExampleBean {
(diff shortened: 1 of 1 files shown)


sbrannen:
As stated in the documentation, the `ExampleBean` class intentionally does not implement `DisposableBean` since its `cleanup` method is configured as the `destroy-method` in the corresponding XML configuration.

Whereas, the `AnotherExampleBean` class in the subsequent example does implement `DisposableBean`.

PR #31076, opened 2023-08-19
Wrap ternary operator within parentheses

70825:
Wrapped ternary operator within parentheses as outlined in [Code Style](https://github.com/spring-projects/spring-framework/wiki/Code-Style#ternary-operator).


snicoll:
Thanks @70825. For the record, we prefer to do things like this ourselves as it often requires to look a bit deeper for consistency. I did another pass and found several other occurrences, see https://github.com/spring-projects/spring-framework/commit/2b76c4d847b066759b02a78a66a148b407e8bda3.

Issue #30648, opened 2023-06-12
Determine why certain JPA tests fail on Java 18+


sbrannen:
This is a follow-up to #30185.

Affected tests can be seen in commit 0ca02ce67720b7ff2b755fd19421b50e89df003d.

bdshadow:
I tried to remove `@DisabledForJreRange` for three tests (`ApplicationManagedEntityManagerIntegrationTests`,  `ContainerManagedEntityManagerIntegrationTests`, `EclipseLinkEntityManagerFactoryIntegrationTests`) in the spring-orm module and successfully ran the tests with:
```
./gradlew :spring-orm:test --tests "org.springframework.orm.jpa.eclipselink.EclipseLinkEntityManagerFactoryIntegrationTests" -PtestToolchain=19
```
(of course, changing `--tests` for every test).
```
dmitrii.bocharov A0111:~/dev/spring-framework$ java --version
openjdk 19.0.2 2023-01-17
OpenJDK Runtime Environment (build 19.0.2+7-44)
OpenJDK 64-Bit Server VM (build 19.0.2+7-44, mixed mode, sharing)
```
To be sure java 19 is used, i put a breakpoint on some test and `Runtime.version()` showed me the 19th version.
Maybe i'm doing smth wrong, but these 3 orm tests seem fine.

The mentioned kotlin tests can't be fixed by now, because the disabledReason = "Kotlin doesn't support Java 20+ yet" is still valid.

snicoll:
Eclipselink tests have been restored via #31078.

snicoll:
That actually fixed it.

PR #31078, opened 2023-08-20
Fix spring-orm tests with eclipselink and Java 18+

bdshadow:


bdshadow:
@sdeleuze  it contains JPA update, because the issue reported by @sbrannen, which I fixed is about JPA: https://github.com/spring-projects/spring-framework/issues/30648

bdshadow:
actually, that's why i made 2 commits. I can remove the one for kotlin fixes

snicoll:
Let's have a look to the other commit indeed.

snicoll:
Unfortunately, we can't change the dependency of Eclipselink like that as we need to build against 3.0 to match our JPA 3.0  / EE 9 baseline. We should probably see if we can change the dependency when the tests run with Java 21.

snicoll:
However, @jhoeller pointed out that Eclipselink `3.0.4-RC1` was released this week and could be used instead.

Issue #30688, opened 2023-06-19
Condition evaluation on beans with similar factory methods are mixed


wang-xiaowu:
**Affects:** SpringBoot 3.1.0

---

i am using `@Configuration(enforceUniqueMethods = false)` to allow the configuration has the save name of beans, but if i set `sync: false`，it show the error `Parameter 0 of method xxx required a bean of type xxx that could not be found.`.
it only execute the code which is on the top, for example i put the `SyncFlatMessageHandlerImpl` bean on the top of `AsyncFlatMessageHandlerImpl` and set `sync: true`，it also show the error `Parameter 0 of method xxx required a bean of type xxx that could not be found.`. because it did not execute the code in `AsyncFlatMessageHandlerImpl`
```java
@Configuration(enforceUniqueMethods = false)
public class RabbitMqClientAutoConfiguration {

    @Bean
    @ConditionalOnProperty(value = "sync", havingValue = "true", matchIfMissing = true)
    public MessageHandler messageHandler() {
        return new AsyncFlatMessageHandlerImpl();
    }

    @Bean
    @ConditionalOnProperty(value = "sync", havingValue = "false")
    public MessageHandler messageHandler() {
        return new SyncFlatMessageHandlerImpl();
    }

    @Bean(initMethod = "start", destroyMethod = "stop")
    public RabbitMqCanalClient rabbitMqCanalClient(MessageHandler messageHandler) {
        .......
        return rabbitMqCanalClient;
    }
}
```

mdeinum:
This isn't allowed in Java regardless of what the `enforceUniqueMethods` states. You cannot have 2 methods with the exact same signature in a Java class. 

The `enforceUniqueMethods` is there when you have methods with the same name but different signatures (like including a different dependent dependency), to eliminate errors with this it can be enforced to differentiate the names (which is the default).

bclozel:
As highlighted by @mdeinum, this is not a valid code snippet. @wang-xiaowu Please provide a minimal sample application that reproduces this issue.

snicoll:
I am not sure how it's related to `enforceUniqueMethods`. It looks like the identity of the thing that matches the condition is not separate between the first, and the second method. Looking at the auto-configuration report, we get this:

```
UserConfig#user:
      Did not match:
         - @ConditionalOnProperty (sync=true) found different value in property 'sync' (OnPropertyCondition)
      Matched:
         - @ConditionalOnProperty (sync=false) matched (OnPropertyCondition)
[...]

Issue #31073, opened 2023-08-18
Deprecate use of several bean factory methods for the same bean in a configuration class


snicoll:
Consider the following use case:

```java
@Configuration
public class Config {
	
	@Bean(name = "myBean")
	MyBean someBean() { ... }
	
	@Bean(name = "myBean")
	MyBean someBean(MyDependency) { ... }
}
```

The container accepts an override for `myBean` as long as it is in the same configuration class to mimic the resolution of the greediest satisfiable constructors for bean factory methods. 

Such an arrangement is much better served using `Optional` or `ObjectProvider` with a single bean factory method as it makes it explicit that some argument may or may not be available.

jhoeller:
Bean method overloading is effectively turned off by default as of 6.0, see #22609. From that perspective, formal deprecation does not seem urgent and might be better off in 7.0.

jhoeller:
The `enforceUniqueMethods` flag is formally deprecated now, with the default `true` being the only non-deprecated behavior now. This prevents regular use of overloaded factory methods as well as `@Bean` method overrides with the same bean name but different method names. For the latter scenario, `@Configuration(enforceUniqueMethods=false)` makes it work as before in 7.0 as well (previously that kind of arrangement was not covered by the `enforceUniqueMethods` flag at all).

PR #31080, opened 2023-08-20
Conditions on overloaded beans not honored

astrologic7:
As pointed out in the issue #30688 the conditions on the overloaded beans aren't honored in all cases.
This is due to the fact that the `ConfigurationClassBeanDefinitionReader` class bases its logic about which bean method should be skipped on the method name only. The solution provides a brand-new method `String getMethodSignature();` in the `MethodMetadata` class, based on which will be decided if a bean should be skipped.


quaff:
I confirm this PR fixes https://github.com/spring-projects/spring-framework/issues/30688

quaff:
Related to https://github.com/spring-projects/spring-framework/issues/31073

snicoll:
Thanks for the PR but we've declined this use case, see https://github.com/spring-projects/spring-framework/issues/30688#issuecomment-1687931033

Issue #26091, opened 2020-11-13
@DynamicPropertySource is not supported via @NestedTestConfiguration semantics


fbiville:
Reproduced with Spring Framework 5.3.1 (via Spring Boot 2.4.0).

As initially described [here](https://jira.spring.io/browse/DATACMNS-1829) and [here](https://github.com/spring-projects/spring-boot/issues/23929#issuecomment-726701029), nested tests do not work with `@DynamicPropertySource`.

You can reproduce the issue with this sample Spring Data JPA (2.4.1) repository: https://github.com/fbiville/spring-data-jpa-nested (see README for reproduction one-liners).

As suggested by @wilkinsona [here](https://github.com/spring-projects/spring-boot/issues/23929#issuecomment-726711390),  replacing the annotation with `ApplicationContextInitializer` works.

sbrannen:
Thanks so much for raising the issue!

When implementing #19930, my goal was to provide support for inheriting enclosing class configuration for all class-level annotations; however, I simply overlooked `@DynamicPropertySource` in the process. 

`@DynamicPropertySource` is technically a `static` method-level annotation, but it in fact contributes to the context cache key in the _Spring TestContext Framework_.

So I apologize for missing that.

I've already reproduced the issue in an integration test using only the Spring Framework, and the fix requires some changes in `DynamicPropertiesContextCustomizerFactory` to honor `@NestedTestConfiguration` semantics. So I'll make sure to provide proper support for inheriting `@DynamicPropertySource` configuration in `@Nested` test classes in Spring Framework 5.3.2.

sbrannen:
I've added support for _inheriting_ `@DynamicPropertySource` methods from enclosing classes.

This will be released with Spring Framework 5.3.2, but feel free to try it out in an upcoming snapshot.

As a workaround for 5.3.0 and 5.3.1, you can create a test base class (typically `abstract`) or test interface that only includes the `@DynamicPropertySource` method and then extend or implement that class or interface in each `@Nested` test class that needs it. Examples can be seen in the `DynamicPropertySourceFromSuperclassTests` and `DynamicPropertySourceFromInterfaceTests` nested test classes in [DynamicPropertySourceNestedTests](https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/test/java/org/springframework/test/context/junit/jupiter/nested/DynamicPropertySourceNestedTests.java).

PR #31083, opened 2023-08-21
`@DynamicPropertySource` in `@Nested` test class cannot override dynamic properties from enclosing class

### LINKED ISSUE #37040: @Nested classes with nested @DynamicPropertySource get called in order, but don't get applied
It seems that `@Nested`  and `@DynamicPropertySource` do not play together when both outer and inner classes have a source defined. Neither `@DirtiesContext` nor inheriting / overwritting seems to do the trick, (overwritting expects a full setup, while inheriting has the methods called in proper order but they don't seem to get applied).

I've added a manual test to the existing [test class](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/test/java/org/springframework/test/context/junit/jupiter/nested/DynamicPropertySourceNestedTests.java) that should fail as I'd expect `"1"` to be the host but it is not. 

```java
SpringJUnitConfig
class DynamicPropertySourceNestedTests {

    private static final String TEST_CONTAINER_IP = "DynamicPropertySourceNestedTests.test.container.ip";

    private static final String TEST_CONTAINER_PORT = "DynamicPropertySourceNestedTests.test.container.port";

    static DemoContainer container = new DemoContainer();

    @DynamicPropertySource
    static void containerProperties(DynamicPropertyRegistry registry) {
        registry.add(TEST_CONTAINER_IP, container::getIpAddress);
        registry.add(TEST_CONTAINER_PORT, container::getPort);
    }

    private static void assertServiceHasInjectedValues(Service service) {
        assertThat(service.getIp()).isEqualTo("127.0.0.1");
        assertThat(service.getPort()).isEqualTo(4242);
    }

    @Test
    @DisplayName("@Service has values injected from @DynamicPropertySource")
    void serviceHasInjectedValues(@Autowired Service service) {
        assertServiceHasInjectedValues(service);
    }

    @Nested
    class Testing {
        @DynamicPropertySource
        static void containerProperties(DynamicPropertyRegistry registry) {
            registry.add(TEST_CONTAINER_IP, () -> "1"); // even though it is called after outer dynamic properties it does not get applied
            registry.add(TEST_CONTAINER_PORT, container::getPort);
        }

        @Test
        void serviceHasInjectedValues(@Autowired Service service) {
            assertServiceHasInjectedValues(service);
        }
    }

    @Configuration
    @Import(Service.class)
    static class Config {
    }

    static class Service {

        private final String ip;

        private final int port;


        Service(@Value("${" + TEST_CONTAINER_IP + ":10.0.0.1}") String ip, @Value("${" + TEST_CONTAINER_PORT + ":-999}") int port) {
            this.ip = ip;
            this.port = port;
        }

        String getIp() {
            return this.ip;
        }

        int getPort() {
            return this.port;
        }
    }

    static class DemoContainer {

        String getIpAddress() {
            return "127.0.0.1";
        }

        int getPort() {
            return 4242;
        }
    }
}
```

quaff:
Fix https://github.com/spring-projects/spring-boot/issues/37040

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizerFactory.java
+++ b/spring-test/src/main/java/org/springframework/test/context/support/DynamicPropertiesContextCustomizerFactory.java
@@ -35,6 +35,7 @@ import org.springframework.test.context.TestContextAnnotationUtils;
  *
  * @author Phillip Webb
  * @author Sam Brannen
+ * @author Yanming Zhou
  * @since 5.2.5
  * @see DynamicPropertiesContextCustomizer
  */
@@ -54,10 +55,10 @@ class DynamicPropertiesContextCustomizerFactory implements ContextCustomizerFact
 	}
 
 	private void findMethods(Class<?> testClass, Set<Method> methods) {
-		methods.addAll(MethodIntrospector.selectMethods(testClass, this::isAnnotated));
 		if (TestContextAnnotationUtils.searchEnclosingClass(testClass)) {
 			findMethods(testClass.getEnclosingClass(), methods);
 		}
+		methods.addAll(MethodIntrospector.selectMethods(testClass, this::isAnnotated));
 	}
 
 	private boolean isAnnotated(Method method) {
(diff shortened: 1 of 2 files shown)


sbrannen:
## Related Issues

- #26091

sbrannen:
This has been merged into `6.0.x` and `main` in 368036cab433e6f4b1dc1045725a2677e4400d70 and revised in 620a87bcbc966be0c71a3c66c78c5633936f66c8.

This has also been backported to `5.3.x` (see #31085).

Thanks for noticing [the issue](https://github.com/spring-projects/spring-boot/issues/37040) in the Spring Boot issue tracker, fixing it, and submitting the PR! 👍 

----

Side note: after attempting to backport this to `5.3.x`, I realized why I had never tested it in the first place. Static members in inner classes were not supported in Java until Java 16 with the introduction of records (see https://bugs.openjdk.org/browse/JDK-8254321), and we didn't start authoring tests in Java 17 until 6.0. Thus, I had to remove the `DynamicPropertySourceOverridesEnclosingClassTests` class on the `5.3.x` branch.

PR #31091, opened 2023-08-22
Fix invalid type name in RSocket section of the reference documentation

aahlenst:
There seems to be a typo in the `@RSocketExchange` documentation. The code sample declares the type of the generated proxy as `RepositoryService` but `RadarService` does not extend any interface. I assume it should be `RadarService` instead.


snicoll:
Good catch @aahlenst

Issue #28703, opened 2022-06-24
`GenericApplicationContext` does not honor `ProtocolResolver` when a resource loader is set via `setResourceLoader()`


arteymix:
I'm trying to intercept certain URL patterns to resolve into the correct `UrlResource`, but the application context does not honour custom protocol resolvers.

In pseudocode:

```java
GenericApplicationContext applicationContext = new GenericApplicationContext();
applicationContext.setResourceLoader(someResourceLoader);
applicationContext.addProtocolResolver(new CustomUrlResolver());
applicationContext.getResource("http://some.custom.url/"); // never resolved via CustomUrlResolver
```

The `CustomUrlResolver` resolver is never invoked because `someResourceLoader` takes precedence in `getResource`. Unfortunately, there's no way to work around this issue by adding a delegating resource loader, since there's no accessor for the resource loader.

I believe that the `setResourceLoader(...)` call is done somewhere in Spring Boot.

I had to do `setResourceLoader(null)`. This is likely going to break things.

arteymix:
In that case, you can put the for-loop within the `this.resourceLoader != null` condition.

sbrannen:
> In that case, you can put the for-loop within the `this.resourceLoader != null` condition.

Yes. You are of course correct, and that's exactly how I implemented it. 👍 

When I said "at a glance", I meant that literally. It was the last thing I _glanced_ at before signing off for the day, and I hadn't even put any thought into how to optimize the code. 😇 

This is in place now for 5.3.22 (available soon in snapshots).

PR #31101, opened 2023-08-23
Introduce configuration to diable @Scheduled tasks

### LINKED ISSUE #28073: Introduce NoOpTaskScheduler for disabling @Scheduled tasks in test setups
If an application defines a `@Scheduled(fixedRateString = "PT6H")` tasks, then this task will be running on any `@SpringBootTest` `junit` test that does not explicit mock the bean containing the scheduled task.

```
@Service
public class SchedulerService {
    @Scheduled(fixedRateString = "PT6H")
    public void runOnStartup() {

   }
}
```

This gets tedious if there are many more `@Scheduled` tasks in the application that launch on startup by usage of `fixedRate` or `fixedRateString`.

It would be great if there could be a possibility to completely deactivate the execution of `@Scheduled` tasks in junit.

Just like it is possible to deactivate any `@Cacheable` annotation by simply setting `spring.cache.type=none` dring junit tests.

Something like: `spring.enable.scheduling=false` (default: true)

Leewei222:
Prior to this commit, only cron @Scheduled tasks can be disabled with the special cron expression value that indicates a disabled trigger: "-".

This commit enables this configuration to fixedDelay and fixedRate @Scheduled tasks.

Fixes gh-28073

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/annotation/ScheduledAnnotationBeanPostProcessor.java
@@ -462,14 +462,15 @@ public class ScheduledAnnotationBeanPostProcessor
 				if (StringUtils.hasLength(fixedDelayString)) {
 					Assert.isTrue(!processedSchedule, errorMessage);
 					processedSchedule = true;
-					try {
-						fixedDelay = toDuration(fixedDelayString, scheduled.timeUnit());
+					if (!Scheduled.CRON_DISABLED.equals(fixedDelayString)) {
+						try {
+							fixedDelay = toDuration(fixedDelayString, scheduled.timeUnit());
+						} catch (RuntimeException ex) {
+							throw new IllegalArgumentException(
+									"Invalid fixedDelayString value \"" + fixedDelayString + "\" - cannot parse into long");
+						}
+						tasks.add(this.registrar.scheduleFixedDelayTask(new FixedDelayTask(runnable, fixedDelay, initialDelay)));
 					}
-					catch (RuntimeException ex) {
-						throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


snicoll:
@Leewei222 can you share how this is fixing #28073? The issue there is rather a disable switch in testing scenario. Your proposal would require to change the underlying annotation. Also `-` for fixedDelay and fixedRate looks quite strange to me.

Leewei222:
Hi @snicoll, thanks for the feedback.

The scheduled jobs are usually configured with property place holders. In some environments, we want to disable some of the schedulers. For cron scheduled job, this can easily be done by the following code 
```
@Scheduled(cron = "${cron}")
 public void scheduleTaskWithCron() {
     // do something
}

// application.properties
cron=-
```
However, this cannot work for fixedDelay and fixedRate scheduled tasks. I think there should be an easy way for us to disable them too! My proposal can disable them with the following code
```
@Scheduled(fixedDelayString= "${fixedDelay}")
 public void scheduleTaskWithFixedDelay() {
     // do something
}

@Scheduled(fixedRateString= "${fixedRate}")
public void scheduleTaskWithFixedRate() {
     // do something
}

// application.properties
fixedDelay=-
fixedRate=-
```

If `-` for fixedDelay and fixedRate doesn't make sense, what about using negative number?

snicoll:
OK that explains the use case. I don't think this actually fixes #28703, or at least not directly so let's discuss the merit of this one regardless.

> If - for fixedDelay and fixedRate doesn't make sense, what about using negative number?

Both `fixedDelay` and `fixedRate` are actually numbers already with a convenience String-based attribute counter part. The default value for the long-based ones is already `-1` to indicate that the value should not be taken into account. We can't unfortunately reuse this to signal something else.

If you want to disable some of the schedulers, I'd rather use a `@Profile` to prevent the scheduled bean to be exposed in the context at all. This should require some code refactoring but I believe this will for the better and more explicit than what you were suggesting.

PR #31113, opened 2023-08-25
Fix return value of HandlerMethod is ignored if reason of @ResponseStatus present

quaff:
See https://github.com/spring-projects/spring-boot/issues/37098#issuecomment-1692958851


quaff:
It breaks tests introduced by d52fc3bd2fb96de0f92e51782523ecd2b06646b0, I'm not sure it's a bug should be fixed, maybe we should update the javadoc of `ResponseStatus` to clarify current behavior instead of fixing this.

quaff:
Superseded by https://github.com/spring-projects/spring-framework/pull/31121

PR #31115, opened 2023-08-26
Use notEmpty() to check whether the collection/array is empty or not.

remeio:
When `collection` is `null`:
- `Assert.isTrue(collection.size() > 0, message)` will throw `NullPointerException`.
- `Assert.notEmpty(collection, message)` will throw `IllegalArgumentException`.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/PropertySourceRegistry.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/PropertySourceRegistry.java
@@ -64,7 +64,7 @@ class PropertySourceRegistry {
 			encoding = null;
 		}
 		String[] locations = propertySource.getStringArray("value");
-		Assert.isTrue(locations.length > 0, "At least one @PropertySource(value) location is required");
+		Assert.notEmpty(locations, "At least one @PropertySource(value) location is required");
 		boolean ignoreResourceNotFound = propertySource.getBoolean("ignoreResourceNotFound");
 
 		Class<? extends PropertySourceFactory> factoryClass = propertySource.getClass("factory");
--- a/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
@@ -622,7 +622,7 @@ public abstract class DataBufferUtils {
 	 * @since 5.2
 	 */
 	public static Matcher matcher(byte[]... delimiters) {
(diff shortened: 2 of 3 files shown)


spring-context/src/main/java/org/springframework/context/annotation/PropertySourceRegistry.java
```diff
@@ -64,7 +64,7 @@ void processPropertySource(AnnotationAttributes propertySource) throws IOExcepti
 			encoding = null;
 		}
 		String[] locations = propertySource.getStringArray("value");
-		Assert.isTrue(locations.length > 0, "At least one @PropertySource(value) location is required");
+		Assert.notEmpty(locations, "At least one @PropertySource(value) location is required");
```
sbrannen, line 67: This is unnecessary since `org.springframework.core.annotation.AnnotationAttributes.getStringArray(String)` will never return `null`.

spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
```diff
@@ -622,7 +622,7 @@ public static Matcher matcher(byte[] delimiter) {
 	 * @since 5.2
 	 */
 	public static Matcher matcher(byte[]... delimiters) {
-		Assert.isTrue(delimiters.length > 0, "Delimiters must not be empty");
+		Assert.notEmpty(delimiters, "Delimiters must not be empty");
```
sbrannen, line 625: We rely on nullability constraints in such cases, and `delimiters` should never be `null`, since `@Nullable` is not declared for that method parameter.

If a user supplies a `null` value for the var-args array, the resulting `NullPointerException` should suffice to inform them of their error.

spring-core/src/main/java/org/springframework/core/io/support/PropertySourceProcessor.java
```diff
@@ -80,7 +80,7 @@ public void processPropertySource(PropertySourceDescriptor descriptor) throws IO
 		String name = descriptor.name();
 		String encoding = descriptor.encoding();
 		List<String> locations = descriptor.locations();
-		Assert.isTrue(locations.size() > 0, "At least one @PropertySource(value) location is required");
+		Assert.notEmpty(locations, "At least one @PropertySource(value) location is required");
```
sbrannen, line 83: The `locations()` property of the `PropertySourceDescriptor` record is not declared as `@Nullable`.

See previous comment for further details on nullability constraints.

sbrannen (commented):
In light of the comments I've added to the review, we are not inclined to make the proposed changes.

PR #31121, opened 2023-08-28
Clarify and log that return value of HandlerMethod is ignored if reason of @ResponseStatus present

quaff:
Supersedes GH-31113
See https://github.com/spring-projects/spring-boot/issues/37098#issuecomment-1692958851


bclozel:
I don't think we should add log statements for this. We could improve the javadoc to let the users know that the reason will be used to render the response entirely (and keep referring to the `sendError` method). I don't think it makes things clearer to refer to annotated methods in the description either.

Issue #30750, opened 2023-06-26
@Order on nested @Configuration bug


mironbalcerzak:
**Affects:** 5.2.24.RELEASE, 5.3.28, 6.0.10

**Summary:**

`@Order` on nested `@Configuration` classes is not respected in certain scenarios.

Once `static` is removed, bean loading occurs as expected (and that's against what is documented for `@Configuration` annotation).

Here are diagrams and related code:

https://github.com/mironbalcerzak/poc-spring-config-ordering

Here's SO: https://stackoverflow.com/questions/76552375/order-ing-of-configuration-for-bean-overriding

Have a great day =)

mironbalcerzak:
@bclozel  any feedback

jhoeller:
Revisiting the ordering affect for configuration classes along with documenting the current behavior in #30177.

Generally speaking, ordering among configuration classes is not meant to be a hard declaration - in particular not for bean overrides by name in those configuration classes -, just a soft ordering between otherwise arbitrary or name-based registration order.

jhoeller:
The effect that you are seeing comes from the classpath scan finding the nested classes directly rather than through their containing class. As a consequence, it processes them in the order that it found them in the classpath. Through declaring those nested classes as non-static, classpath scanning does not consider them as independent anymore, so they will actually be processed through their containing class then - with the order for nested classes respected there. You can achieve the same effect by removing the `@Configuration` stereotype from the nested classes (e.g. A1, A2) so that classpath scanning does not identify them anymore; this works with static classes as well since they will only be found through their containing class then.

As a recommendation, if you rely on ordering for any such structure, do not use classpath scanning at all for that part. And even more generally, try to not rely on ordering at all. Bean definition overriding can be surprising even in straightforward scenarios, and such nesting with ordering for bean definition override purposes is bound to cause maintenance headaches.

PR #31123, opened 2023-08-28
When context uses component scan, parent class conditions are ignored in favor of nested class conditions.

### LINKED ISSUE #29372: Parent Conditionals are ignored on component scanning
Imagine you have the following configuration and the associated properties:

<details>
<summary>MyConfiguration.java</summary>

```java
@Configuration
@EnableConfigurationProperties(MyProperties.class)
public class MyConfiguration {

  @Configuration
  @ConditionalOnProperty(value = "my.enabled", havingValue = "true")
  static class MyEnabledConfiguration {

    @Configuration
    @ConditionalOnProperty(value = "my.version", havingValue = "v1")
    static class MyEnabledFirstConfiguration {
      @Bean
      public MyClient firstClient() {
        return new FirstClient();
      }
    }

    @Configuration
    @ConditionalOnProperty(value = "my.version", havingValue = "v2", matchIfMissing = true)
    static class MyEnabledSecondConfiguration {
      @Bean
      public MyClient secondClient() {
        return new SecondClient();
      }

    }
  }

  @Configuration
  @ConditionalOnProperty(value = "my.enabled", havingValue = "false", matchIfMissing = true)
  static class MyDisabledConfiguration {

    @Bean
    public MyClient thirdClient() {
      return new ThirdClient();
    }
  }
```
</details>

<details>
<summary>MyProperties.java</summary>

```java
@Getter
@Setter
@Validated
@ConfigurationProperties(prefix = "my")
public class MyProperties {

  private boolean enabled;

  private Version version = Version.V1;

  public enum Version {
    V1,
    V2
  }
}
```

</details>

In following case I expect the only single bean of type `ThirdClient`

```yaml
my:
  enabled: false
  version: v1
```

And it actually works fine, if you write your test as following: 

```java 
@SpringBootTest(classes = MyConfiguration.class)
class MySpringBootTest {

    @Autowired
    private ApplicationContext context;

    @Test
    void contextLoads() {
        assertThat(context.getBeansOfType(MyClient.class).size()).isEqualTo(1);
        assertThat(context.getBean(MyClient.class)).isInstanceOf(ThirdClient.class);
    }
}
```

However if you are using component scan, the parent conditional is ignored.

```diff
+ @SpringBootTest
- @SpringBootTest(classes = MyConfiguration.class)
class MySpringBootTest {
    ...
}
```

In this case you get the following two beans instead of expected single one as in example above. 
* FirstClient
* ThirdClient

astrologic7:
The solution is really simple and non-invasive: in the context scan flow, when `scan(basePackages)` is invoked, we deal with resources instead of classes. Before any further evaluations, these resources need to be sorted based on how many "**$**" are present in the filename. When a valid candidate is excluded because of a non-matching condition, the corresponding filename is added to a list. Before registering a candidate, this list is consulted. The candidate is registered if the list doesn't contain the candidate's parent filename.

Closes #29372

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ClassPathScanningCandidateComponentProvider.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ClassPathScanningCandidateComponentProvider.java
@@ -19,11 +19,7 @@ package org.springframework.context.annotation;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -53,13 +49,10 @@ import org.springframework.core.type.filter.AnnotationTypeFilter;
 import org.springframework.core.type.filter.AssignableTypeFilter;
 import org.springframework.core.type.filter.TypeFilter;
 import org.springframework.lang.Nullable;
-import org.springframework.stereotype.Component;
-import org.springframework.stereotype.Controller;
-import org.springframework.stereotype.Indexed;
-import org.springframework.stereotype.Repository;
-import org.springframework.stereotype.Service;
+import org.springframework.stereotype.*;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
+import org.springframework.util.StringUtils;
 
 /**
  * A component provider that scans for candidate components starting from a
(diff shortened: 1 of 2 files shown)


EdoardoP92:
@bclozel do you think that can be merged?

EdoardoP92:
@snicoll is it possible to get feedback about this contribution?

snicoll:
@EdoardoP92 paging individual team members isn't going to help this PR to be reviewed faster. We're doing our best with our current priorities and we'll get to it as soon as possible. Thank you for your patience.

jhoeller:
Closing this PR along with #29372 and #30750 since classpath scanning is meant to find such nested classes independently when they carry their own stereotype annotation.

PR #31125, opened 2023-08-28
Add more tests to DefaultClientRequestObservationConvention

annabuyevich:
Adding tests to increase coverage and test expected behavior for `DefaultClientRequestObservationConvention`
class. I would love to contribute more if possible.

Added the following test:

1. Response with 400 status
2. Response with unknown status code
3. Null Request
4. Null Response
5. Error in Context


bclozel:
Thanks @annabuyevich !

PR #31129, opened 2023-08-29
Add @JsonValue To HttpMethod

ghost:
As the HttpMethod was changed from enum to class, it was changed differently from the previous one during serialization.
I added @JsonValue and changed it to look the same as before.


bclozel:
Thanks but we don't accept Jackson specific annotations on such types as they are not meant for serialization and this case is not covered by our backward compatibility rule.

Please use a custom serializer and register it with Jackson, at your own risk.

PR #31132, opened 2023-08-29
Avoid @Async create Future object when return value is not instance of Future type

fangfei1080:
submit method return a value, but it is useless, and memory is added. 
execute method can avoid this scene.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/interceptor/AsyncExecutionAspectSupport.java
+++ b/spring-aop/src/main/java/org/springframework/aop/interceptor/AsyncExecutionAspectSupport.java
@@ -291,7 +291,7 @@ public abstract class AsyncExecutionAspectSupport implements BeanFactoryAware {
 			return executor.submit(task);
 		}
 		else if (void.class == returnType) {
-			executor.submit(task);
+			executor.execute(task);
 			return null;
 		}
 		else {


spring-aop/src/main/java/org/springframework/aop/interceptor/AsyncExecutionAspectSupport.java
```diff
@@ -291,7 +291,7 @@ else if (Future.class.isAssignableFrom(returnType)) {
 			return executor.submit(task);
 		}
 		else if (void.class == returnType) {
-			executor.submit(task);
+			executor.execute(task);
```
mdeinum, line 294: Have you tried compiling this? I doubt it would even compile as `execute` takes a `Runnable` whilst `submit` takes a `Callable`. The type of `task` is `Callable` and not `Runnable` and as a `Callable` isn't a `Runnable` I doubt this would even compile.

flyfang-jpg: You are right. I just compiled source code local. It occured a problem with incompatible param type. I'm sorry to think Executor's API only, but ignored the param. Thanks for to corret my first commit (a little depressed). I will try to do beeter on this project. Thanks again!

PR #31150, opened 2023-09-01
WebFlux blocking configuration documentation fix

mdeinum:
The `BlockingExecutionConfigurer.setExecutor` takes an `AsyncTaskExecutor` not an `Executor` as the documentation states. With this commit this has been corrected.


sdeleuze:
Merged, thanks!

PR #31164, opened 2023-09-03
Consolidate HTTP Interface client support for MultipartFile and `@RequestPart`

### LINKED ISSUE #31139: HTTP Interface client should not fail if `@RequestPart` used with `MultipartFile`
In #30342 we added support for `MultipartFile` in a separate arguments resolver. The resolver is documented as not requiring `@RequestPart` since it's recoginized by type and can be made optional with `java.util.Optional`. However, it also shouldn't fail if that combination is used, and right now that's not the case as mentioned in as per https://github.com/spring-projects/spring-framework/issues/30342#issuecomment-1691949955, since the resolver for `@RequestPart` is used first.

I think we should reconsider a tighter integration between the two resolvers, and maybe even merge them into the `RequestPartArgumentResolver` so it supports `@RequestPart` with `MultipartFile` but also just `MultipartFile`.

/cc @OlgaMaciaszek

bdshadow:
As described in the ticket, I've merged `MultipartFileArgumentResolver` into the `RequestPartArgumentResolver`. I also think this is a better option since the javadoc for`@RequestPart` explicitly mentions work with the `MultipartFile.`

However, I'm unsure if `MultipartFileArgumentResolver` must be removed now. Probably it can, since the release 6.1 isn't out yet, so it's present only in the pre-releases.


rstoyanchev (changes requested):
Thanks for the pull request. Indeed, as we merge the two into one, we would no longer need `MultipartFileArgumentResolver` and since it is not yet released, it can be removed, and likewise the test classes should also be merged into one. 

Would you like to apply those changes?

bdshadow:
@rstoyanchev yes, i will do it

PR #31175, opened 2023-09-05
Polish resolveArgument method in RequestResponseBodyMethodProcessor

shin-mallang:
```java
String name = Conventions.getVariableNameForParameter(parameter);
```
The above code is only used in conditional statements.
So I moved it into the conditional statements.
With this, we can expect a slight performance improvement in some case.


shin-mallang:
And I have a question.

WebUtils's `isIncludeRequest()` method  looks like this.
```java
public static boolean isIncludeRequest(ServletRequest request) {
    return (request.getAttribute(INCLUDE_REQUEST_URI_ATTRIBUTE) != null);
}
```
It has bracket that is unnecessary and inconsistent with the other methods.
Can I change this to the following?
```java
public static boolean isIncludeRequest(ServletRequest request) {
    return request.getAttribute(INCLUDE_REQUEST_URI_ATTRIBUTE) != null;
}
```

shin-mallang:
I was wondering if there are any situations where binderFactory is null.
If you don't mind, could you let me know?

sdeleuze:
I merged this refinement. For your question, I am not sure how common it is, maybe @poutsma or @rstoyanchev know more.

shin-mallang:
@sdeleuze 
thank you :)

PR #31186, opened 2023-09-07
Improve specificity of the `UriComponentsBuilder` `path()` method

HollowedKnight:
The documentation of the function specifically mentions that the path is appended but the naming of the method
did not reflect said behaviour.


rstoyanchev:
Thanks for the suggestion, but it is too late for such naming changes. If we change `path()`, similar arguments can be made to add a verb to any other builder methods. We would also have to deprecate existing methods to avoid further confusion from having parallel methods.

Issue #31127, opened 2023-08-28
Replace Reactive `awaitSingle()` usages by the `Mono` variant


sdeleuze:
Could improve Reactor/Coroutine context interroperability.

PR #31188, opened 2023-09-07
Introduce awaitReceive and awaitReceiveOrNull APIs

### LINKED ISSUE #30926: web.reactive.function.server.ServerRequest.awaitBody() exception handling
**Affects:** Spring v6.0.6

## What is the recommended approach on handling ServerRequest.awaitBody()

Right now, the API is using under the hood `awaitSingle()` from `kotlinx.coroutines.reactive`, which is intentional from
the spring team. Now, the API is throwing either `NoSuchElementException` or `IllegalArgmuentException` depending on if
the publisher emits zero or more than one values.

**Question:** In general code, are we expected to catch them both? Although, the operator is being used on a `Mono`,
which technically can never throw `IllegalArgmuentException`?

On the other hand, since we always extract the body (thus we are not directly affected by how a client writes in body)
in either a `Flux` or `Mono` if the client emits more than one values, but we use `ServerRequest.awaitBody()` the body
is converted **properly**, and we are retrieving a collection. Though, this process breaks in serialization 
(as expected).

### Usage in general code

The most "blur" part is about throwing `IllegalArgmuentException`. That is because, when using `kotlinx.serialization`
for general "catch-all" around deserialization it is recommended to catch `IllegalArgmuentException`. In that case, we
have to know beforehand if we need to also check if  `IllegalArgmuentException` is type of `SerializationException` 
(so we can distinguish between them).

When I am writing code to handle `ServerRequest.awaitBody()`, I find it a one-way path to use 
`ServerRequest.awaitBodyOrNull()` API to avoid all the above points.

My exception handling looks like this:

```kotlin
suspend inline fun <reified T : Any> ServerRequest.awaitAndRequireBody(): T {
    val body = try {
        awaitBodyOrNull<T>()
    } catch (e: IllegalArgumentException) { // serialization error
        throw IllegalArgumentException(bodyTypeErrorMessage<T>())
    }
    requireNotNull(body) { bodyTypeErrorMessage<T>() }
    return body
}
```

Which makes `ServerRequest.awaitBody()` kind of "obsolete" API (at least imho).

## Proposal

I recommend to at least document about the fact that `ServerRequest.awaitBody()` throws exceptions which are expected to be caught from the user.

Also, happy to provide a pr with the changes.

GeorgePap-719:
## Summary

`ServerRequest.awaitBody*` APIs have some problems:

1. `ServerRequest.awaitBody` forces you to think about the coroutines-bridge exception handling, since it uses
   `kotlinx.coroutines.reactive.awaitSingle`. Also, it is not documented that it might throw `NoSuchElementException`
   and `IllegalArgumentException` for **coroutines-related** errors.

2. `ServerRequest.awaitBody` might throw `IllegalArgumentException` for two different reasons, for serialization-related
   errors and for coroutines-related errors.

3. `ServerRequest.awaitBodyOrNull` should never throw as the signature indicates. The convention established by the
   stdlib mandate that an operation with the name `xxxOrNull` returns `null` instead of throwing in case there is an
   error.

## Proposal

Introduce APIs that address all the above.

Add `ServerRequest.awaitReceiveNullable`.

- This API "abstracts" away the coroutine-bridging exception handling from the user
- Throws only in case of serialization error
- Returns `null` in case user is expecting the body to be "missing"

Add `ServerRequest.awaitReceive`.

- This API "abstracts" away the coroutine-bridging exception handling from the user
- Throws only in case of serialization error

Both APIs introduce a "mindset" that only serialization-wise can something go wrong, in other words, user input.

**Extensions:**

Based on these functions, the user can create their own function for a general catch-all and return `null` pattern

```
runCatching { awaitReceiveOrNull }.getOrNull()
```

## Deprecations/Migrations

Deprecate `ServerRequest.awaitBodyOrNull` which can be replaced with `ServerRequest.awaitReceiveNullable`, as
**behavior** wise they are pretty close.

Deprecate `ServerRequest.awaitBody` with not direct replacement, but promote the usage of `ServerRequest.awaitReceive`.

Closes gh-30926

### DIFF
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/ServerRequestExtensions.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/ServerRequestExtensions.kt
@@ -31,12 +31,14 @@ import reactor.core.publisher.Mono
 import java.net.InetSocketAddress
 import java.security.Principal
 import kotlin.reflect.KClass
+import kotlin.reflect.KType
+import kotlin.reflect.full.starProjectedType
 
 /**
  * Extension for [ServerRequest.bodyToMono] providing a `bodyToMono<Foo>()` variant
  * leveraging Kotlin reified type parameters. This extension is not subject to type
  * erasure and retains actual generic type arguments.
- * 
+ *
  * @author Sebastien Deleuze
  * @since 5.0
  */
@@ -76,9 +78,25 @@ fun <T : Any> ServerRequest.bodyToFlow(clazz: KClass<T>): Flow<T> =
 /**
  * Non-nullable Coroutines variant of [ServerRequest.bodyToMono].
  *
+ * ### Deprecation
+ *
+ * This method is deprecated in favor of [awaitReceive].
+ *
+ * * It is difficult to reason about its exception handling,
+ *   since it forces you to handle both types of errors: serialization and coroutines-related.
+ * * In the case of [IllegalArgumentException] it is indistinguishable if the exception came from coroutines
+ *   or serialization unless checked explicit for `SerializationException` type.
+ * * It is unclear whether [IllegalArgumentException] can be thrown at all from coroutines-perspective
+ *   as the receiver is a [Mono].
+ *
  * @author Sebastien Deleuze
  * @since 5.2
  */
+@Deprecated(
+	message = "Deprecated in favor of awaitReceive.",
+	level = DeprecationLevel.WARNING,
+	replaceWith = ReplaceWith("this.awaitReceive")
+)
 suspend inline fun <reified T : Any> ServerRequest.awaitBody(): T =
 		bodyToMono<T>().awaitSingle()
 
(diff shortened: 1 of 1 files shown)


sdeleuze:
Thanks for providing this PR, but I don't think we should merge the proposed changes.

> ServerRequest.awaitBody forces you to think about the coroutines-bridge exception handling, since it uses
kotlinx.coroutines.reactive.awaitSingle. Also, it is not documented that it might throw NoSuchElementException
and IllegalArgumentException for coroutines-related errors.

> ServerRequest.awaitBody might throw IllegalArgumentException for two different reasons, for serialization-related
errors and for coroutines-related errors.

Please refer for [this comment](https://github.com/spring-projects/spring-framework/issues/30926#issuecomment-1711368209) for related feedback about current APIs and related documentation and see https://github.com/spring-projects/spring-framework/issues/31189 follow-up issue.

> ServerRequest.awaitBodyOrNull should never throw as the signature indicates. The convention established by the
stdlib mandate that an operation with the name xxxOrNull returns null instead of throwing in case there is an
error.

I don't think this is correct, `Mono<T>.awaitSingleOrNull` can (and should) throw/emit exception at codec level for example, the `null` value is used when for example there is no body, which is something different.

For those reasons, I decline this PR, but thanks for the discussion.

GeorgePap-719:
Hey, thanks for the feedback!

> I don't think this is correct, Mono<T>.awaitSingleOrNull can (and should) throw/emit exception at codec level for example, the null value is used when for example there is no body, which is something different.

I agree with this statement.

I tried to raise the issue about the naming in the function itself  `awaitBodyOrNull`. As i see it, i found the problem kinda similar to the old `Reactive<T>.awaitSingleOrNull` , where it is deprecated for similar reasons, the function was named xxxOrNull but was throwing an exception for a specific error.

So in the same manner, imo, i expect a function named `awaitBodyOrNull` to return `null` in case of an error, nevertheless if it is at codec level or coroutines-related error.

sdeleuze:
As far as I know, `Reactive<T>.awaitSingleOrNull` was deprecated because there is no garantee to have a single element, and is replaced by `Mono<T>.awaitSingleOrNull` which keeps the same behavior is term of exception management. See #31127 related issue.

GeorgePap-719:
I was checking the deprecation section here: [Deprecation note](https://github.com/Kotlin/kotlinx.coroutines/blob/7e2d23740c71b8ec5a7e2440f02b36e0519496a6/reactive/kotlinx-coroutines-reactive/src/Await.kt#L122C7-L122C7).

> because there is no guarantee to have a single element

But that might be also the case. I cannot argue here since I'm not expert in coroutines.

sdeleuze:
Interesting, the deprecation may be partially misleading because as far as I can tell `Publisher<T>.awaitFirstOrNull()` has a documentation saying `if this publisher has produced an error, throws the corresponding exception` and that's the case for  `Mono<T>.awaitSingleOrNull` as well. I will discuss that with the Kotlin team.

GeorgePap-719:
I'm quite interested as well in what the kotlin team has to say.

Nevertheless, I will drop my two cents just for the sake of the discussion:

From my understanding most coroutines operators propagate (and throw) all exceptions
in the chain and, also might throw a `CancelationException` as well (but this is documented always I
think). They are not trying to convey these kinds of exceptions in the signatures but exceptions the
functions/operators themselves throw for their own semantic value.

Examples:

`Publisher<T>.awaitSingle()`: Propagates all exceptions and also throws `CancellationException`.
The KDoc states that the function itself throws two additional exceptions for its own semantic
value: `NoSuchElementException` if the publisher does not emit any value, and
`IllegalArgumentException` if the publisher emits more than one value.
Which can be translated that the function itself does some validations to ensure some semantics
(emits only one value or throws the corresponding exception).

`Publisher<T>.awaitSingleOrNull()`: Propagates all exceptions and also throws
`CancellationException`. The KDoc states that the function itself for its own semantics throws an
`IllegalArgumentException` if the publisher emits more than one value or returns `null` in case
there were non emitted. The problem here was that all `xxxOrNull` functions should return `null`
in case of an error for their own semantic value, however here it is not the case.

`Mono<T>.awaitSingleOrNull()`: Propagates all exceptions and also throws `CancellationException`.
The KDoc states that the function itself for its own semantic value returns `null` in case of an
error, specifically: "If the Mono completed without a value, `null` is returned".

`Publisher<T>.awaitFirstOrNull()`: To avoid being repetitive, this function follows the same pattern
as the other ones and in case of an error for its own semantic value returns `null`.

sdeleuze:
We got confirmation from the Kotlin team the wording from the deprecation notes is misleading (they may or not refine, up to them).

GeorgePap-719:
I see, nevertheless, thanks for the discussion.

PR #31195, opened 2023-09-10
Pass correct `SqlParameterSource` to `NamedParameterJdbcTemplate` in `DefaultJdbcClient`

making:
When using `RowCallbackHandler` or `ResultSetExtractor` in `JdbcClient` and passing a parameter with `paramSource`, `No value supplied for the SQL parameter 'xxxxxx'` occurs because the SqlParameterSource used internally is different.


sbrannen:
Good catch, @making! 👍 

I'll get this merged into `main` soon.

PR #31197, opened 2023-09-10
Use `Method#isDefault` to find default methods in interfaces

refeccd:
Use `Method#isDefault` to replace the logic in `ReflectionUtils#findConcreteMethodsOnInterfaces`.

- see #31198


sbrannen:

Thanks

PR #31202, opened 2023-09-11
Add status handler to recognize unknown status codes outside of 4xx/5…

duesenklipper:
…xx and throw an UnknownHttpStatusCodeException

org.springframework.web.reactive.function.client.UnknownHttpStatusCodeException claims in its Javadoc that it will be thrown to signal that "an unknown (or custom) HTTP status code is received". Up until Spring 5/Spring Boot 2 this was indeed true. We recently upgraded to Spring Boot 3/Spring 6, and our code that expected to receive this exception no longer works.

It turns out that the DEFAULT_STATUS_HANDLER in DefaultWebClient tests HttpStatusCode.isError as a predicate before calling DefaultClientResponse#createException. createException is the only place UnknownHttpStatusException is instantiated.

This is a behavior change from Spring 5, since now only 4xx/5xx custom status codes can lead to a UnknownHttpStatusException, in spite of what the exception's javadoc says.

This PR splits DEFAULT_STATUS_HANDLER into one for errors, and one for unknown codes. Both call the same function, with different predicates - one calls isError(), the other calls a new function isWellKnownStatusCode(). I kept the logic for both WebClientResponseException and UnknownHttpStatusCodeException in createException(), since there still can be a custom 4xx/5xx code. I also provide a test that validates the behavior for custom codes outside of that range, in addition to the existing test which only verified for a custom 5xx code.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpStatusCode.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpStatusCode.java
@@ -89,6 +89,15 @@ public sealed interface HttpStatusCode extends Serializable permits DefaultHttpS
 	}
 
 	/**
+	 * Checks whether the given status code is a well-known HTTP status code or not
+	 * @param statusCode the HTTP status code (potentially non-standard)
+	 * @return {@code true} if the status code corresponds to a standard HTTP status code
+	 */
+	static boolean isWellKnownStatusCode(HttpStatusCode statusCode) {
+		return HttpStatus.resolve(statusCode.value()) != null;
+	}
+
+	/**
 	 * Return an {@code HttpStatusCode} object for the given integer value.
 	 * @param code the status code as integer
 	 * @return the corresponding {@code HttpStatusCode}
--- a/spring-web/src/main/java/org/springframework/http/HttpStatusCode.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpStatusCode.java
@@ -89,12 +89,11 @@ public sealed interface HttpStatusCode extends Serializable permits DefaultHttpS
 	}
 
 	/**
-	 * Checks whether the given status code is a well-known HTTP status code or not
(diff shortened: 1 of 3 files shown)


spring-web/src/main/java/org/springframework/http/HttpStatusCode.java
```diff
@@ -88,6 +88,15 @@ default boolean isSameCodeAs(HttpStatusCode other) {
...
 	}
 
+	/**
+	 * Checks whether the given status code is a well-known HTTP status code or not
+	 * @param statusCode the HTTP status code (potentially non-standard)
+	 * @return {@code true} if the status code corresponds to a standard HTTP status code
+	 */
+	static boolean isWellKnownStatusCode(HttpStatusCode statusCode) {
+		return HttpStatus.resolve(statusCode.value()) != null;
+	}
```
quaff, line 98: How about this? 
```java
	boolean isWellKnown() {
		return HttpStatus.resolve(this.value()) != null;
	}
```

duesenklipper: Very good point, thank you. I changed it to a default method on the interface.

rstoyanchev (changes requested):
It looks this may have happened inadvertently as part of [changes to introduce](https://github.com/spring-projects/spring-framework/commit/28ac0d3883e0c683fb995dffc0af7c8a82f3c3d0#diff-0e1ae29a66f61a7a221793c9ac3897ceef75105c83823ed1eff0e473e0eaf7bcR494) the `HttpStatusCode` interface. I think we should simply revert to how the `Predicate` was before, i.e. checking for codes greater than or equal to 400. I don't see any benefit from the additional default handler.

@poutsma what do you think?

rstoyanchev:
Tentatively scheduled for 6.1 but still pending team review.

Issue #31212, opened 2023-09-13
Make it easier to execute a method on a particular type of thread


wilkinsona:
**Affects:** 6.1

With the introduction of virtual thread support, I anticipate that there will be an increased need for executing a method on a particular type of thread. Specifically, when a Servlet web server is using virtual threads for request handling, I anticipate a need to switch execution to a platform thread to avoid pinning of virtual threads caused by unavoidable synchronization in third-party code. [This Tomcat mailing list discussion](https://marc.info/?l=tomcat-user&m=169408819217183&w=2) is one example of this problem.

This can be achieved today using an `@Async` annotated method that references a `TaskExecutor` that is using platform threads but it comes with a reasonably large amount of boiler plate as you have to move code into a separate method and work with the `Future` API.

To simplify things, I would like Framework to deal with the `Future` on my behalf so that I can write a "normal" method and have it automatically run on a thread from a particular task executor. There's prior art for this sort of functionality in Micronaut with its [`@ExecuteOn` annotation](https://docs.micronaut.io/latest/api/io/micronaut/scheduling/annotation/ExecuteOn.html). In a Spring app, I envisage that a similar annotation could be used on any method that's an entry point into the application such as a method on a `@Controller`.

rstoyanchev:
After an internal discussion, it would not be straight forward to implement this for Spring MVC if at all. In the very least it would require starting an async request, which is not expected, and might not be feasible to implement at all while supporting all that controller methods can do. 

It would be much easier to do this on WebFlux where switching threads can done seamlessly. On WebFlux, the benefit would be to get off the event loop and onto a thread pool for blocking. Note that WebFlux does [provide a mechanism for that](https://docs.spring.io/spring-framework/reference/web/webflux/config.html#webflux-config-blocking-execution). An annotation could make it easier, but we also don't want to introduce an annotation like that if it can't apply more broadly.

For now we are going to close this since [JEP 491](https://openjdk.org/jeps/491) is going to significantly address the original concern. If there is anything comes up in the future, we can look to address it with a more targeted solution, and also taking into account specific constraints with blocking handling in Spring MVC.

PR #31227, opened 2023-09-14
Eliminate synchronized block to avoid thread pinning in SingletonSupplier

quaff:


jhoeller:
Have you seen a concrete effect within a virtual thread here? As long as the target supplier does not do any I/O or other blocking operations, this should not make a difference. That said, `SingletonSupplier` comes as a reusable component, so I can see why it should be able to deal with blocking scenarios as well.

quaff:
It's very common that target supplier invoke I/O, for example querying data from database. This is inspired by [This Tomcat mailing list discussion](https://marc.info/?l=tomcat-user&m=169409979623061&w=2) from https://github.com/spring-projects/spring-framework/issues/31212.

jhoeller:
Internally we only use `SingletonSupplier` for lazy data structures and lazy bean references. So one-time late initialization only, quite different from per-operation synchronization in a database driver. Nevertheless, this looks easy enough to revise, even if just as a defensive measure.

PR #31235, opened 2023-09-14
Improve wording in Javadoc for `@Autowired`

Victory-Canopus:


spring-beans/src/main/java/org/springframework/beans/factory/annotation/Autowired.java
```diff
@@ -30,7 +30,7 @@
  * <h3>Autowired Constructors</h3>
  * <p>Only one constructor of any given bean class may declare this annotation with the
  * {@link #required} attribute set to {@code true}, indicating <i>the</i> constructor
- * to autowire when used as a Spring bean. Furthermore, if the {@code required}
+ * to be autowired when used as a Spring bean. Furthermore, if the {@code required}
```
sbrannen, line 33: It's actually grammatically correct as-is; however, the proposed change may be easier to understand for most readers. So we'll merge the PR to improve the wording, and I've updated the title accordingly.

sbrannen (approved):

sbrannen:

Thanks, and congratulations on your first contribution to the Spring Framework! 👍

spring-beans/src/main/java/org/springframework/beans/factory/annotation/Autowired.java
StellarStrategist: Thanks! It's my first-ever GitHub contribution.

PR #31248, opened 2023-09-16
Introduce `initialize()` method in `AbstractRoutingDataSource` and `AbstractRoutingConnectionFactory`

violetbeach:
Situation is that you want to add DataSource to the DataSources of AbstractRoutingDataSource during runtime.

When i put a Data Source, I need to synchronize target Data Sources to resolved Data Sources.

At that time, I call `afterPropertiesSet()` as below.

```java
@Component
public class MultiDatabaseManager {

    private final Map<Object, Object> dataSourceMap = new ConcurrentHashMap<>();

    private AbstractRoutingDataSource multiDataSource;

    public DataSource createMultiDataSource() {
        MultiDataSource multiDataSource = new MultiDataSource();
        multiDataSource.setTargetDataSources(dataSourceMap);
        multiDataSource.setDefaultTargetDataSource(defaultDataSource());
        return multiDataSource;
    }

    public void addDataSource(String hostIp, ...) throws SQLException {
        DataSource dataSource = DataSourceBuilder
                .create()
                .username(username)
                .password(password)
                .url(url)
                .driverClassName(driverClassName)
                .build();

        try (Connection c = dataSource.getConnection()) {
            dataSourceMap.put(hostIp, dataSource);
            multiDataSource.afterPropertiesSet(); // unclear intentions
        }
    }

}
```

However, `afterPropertiesSet()` is unclear.

So I suggest extract `refresh()` method. Then, intent can be exposed.

Please feel free to share your opinions or advice.

thanks you.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/lookup/AbstractRoutingDataSource.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/lookup/AbstractRoutingDataSource.java
@@ -116,7 +116,7 @@ public abstract class AbstractRoutingDataSource extends AbstractDataSource imple
 
 	@Override
 	public void afterPropertiesSet() {
-		refresh();
+		initialize();
 	}
 
 	/**
@@ -124,7 +124,7 @@ public abstract class AbstractRoutingDataSource extends AbstractDataSource imple
 	 * and defaultTargetDataSource to resolvedDefaultDataSource.
 	 * @throws IllegalArgumentException in case of targetDataSources is null
 	 */
-	public void refresh() {
+	public void initialize() {
 		if (this.targetDataSources == null) {
 			throw new IllegalArgumentException("Property 'targetDataSources' is required");
 		}
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/lookup/AbstractRoutingDataSource.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/lookup/AbstractRoutingDataSource.java
@@ -116,6 +116,15 @@ public abstract class AbstractRoutingDataSource extends AbstractDataSource imple
 
 	@Override
 	public void afterPropertiesSet() {
+		refresh();
+	}
+
+	/**
+	 * Synchronizes targetDataSources to resolvedDataSources
+	 * and defaultTargetDataSource to resolvedDefaultDataSource.
+	 * @throws IllegalArgumentException in case of targetDataSources is null
+	 */
+	public void refresh() {
 		if (this.targetDataSources == null) {
 			throw new IllegalArgumentException("Property 'targetDataSources' is required");
 		}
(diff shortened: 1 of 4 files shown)


sbrannen (changes requested):
Thanks for the proposal and PR.

Can you please also introduce a similar `refresh()` method in `AbstractRoutingConnectionFactory` along with a corresponding test?

violetbeach:
> Can you please also introduce a similar `refresh()` method in `AbstractRoutingConnectionFactory` along with a corresponding test?

Sure. I've done.

Please let me know if you need anything else. Thank you for your review!

spring-jdbc/src/main/java/org/springframework/jdbc/datasource/lookup/AbstractRoutingDataSource.java
```diff
@@ -116,6 +116,15 @@ public void setDataSourceLookup(@Nullable DataSourceLookup dataSourceLookup) {
...
 	public void afterPropertiesSet() {
+		refresh();
+	}
+
+	/**
+	 * Synchronizes targetDataSources to resolvedDataSources
+	 * and defaultTargetDataSource to resolvedDefaultDataSource.
+	 * @throws IllegalArgumentException in case of targetDataSources is null
+	 */
+	public void refresh() {
```
sbrannen, line 127: Please rename to `initialize()`.

spring-r2dbc/src/main/java/org/springframework/r2dbc/connection/lookup/AbstractRoutingConnectionFactory.java
```diff
@@ -127,6 +127,14 @@ public void setConnectionFactoryLookup(ConnectionFactoryLookup connectionFactory
...
 	@Override
 	public void afterPropertiesSet() {
+		refresh();
+	}
+
+	/**
+	 * Synchronizes targetConnectionFactories to resolvedConnectionFactories
+	 * and defaultTargetConnectionFactory to resolvedDefaultConnectionFactory.
+	 */
+	public void refresh() {
```
sbrannen, line 137: Please rename to `initialize()`.

sbrannen (changes requested):
Thanks for making the changes to `AbstractRoutingConnectionFactory` as well.

After further consideration, the team has decided that we would like these new methods to be named `initialize()` instead of `refresh()`, since the latter implies that there is first class support for dynamic refreshes at runtime -- which is not the case.

violetbeach:
> After further consideration, the team has decided that we would like these new methods to be named `initialize()` instead of `refresh()`, since the latter implies that there is first class support for dynamic refreshes at runtime -- which is not the case.

I get it.

Thank you for considering it and the kind comment!

sbrannen (approved):

sbrannen:
This has been merged into `main` in 6d2d8a36c2b608e0c127efa20ef9ea8f8238ff84 and revised in d50ec68ad7bbbc195353e1ba59e2a613bc80013a.

Thanks for the contribution, and congratulations on getting your first PR for the Spring Framework merged! 👍

PR #31253, opened 2023-09-18
Optimize SimpleAliasRegistry#canonicalName

kang-hl:


snicoll:
Thanks for the PR but "optimize" with no context is not really helpful. I think the code is fine as it is and I don't see a visible optimization. Going forward, please provide a description so that we can make an informed call of why you're contributing.

Issue #31261, opened 2023-09-18
It is possible to fetch an uninitialized Bean when cycling dependencies


nineya:
Following up on the latest comments in #25667，Due to the lack of thread-safe judgment for earlySingletonObjects in the getSingleton(String beanName, boolean allowEarlyReference) method, the method may return an uninitialized Bean. And without any processing thereafter, spring returns the uninitialized bean directly to the client that calls the BeanFactory.getBean method.

nineya:
Screenshot of the cause of the problem.
![image](https://github.com/spring-projects/spring-framework/assets/58282918/01011c3f-2219-43f3-9405-ae85cf28e875)

snicoll:
There's no need to create both an issue and a PR as one superseded the other.

PR #26376, opened 2021-01-12
Avoid getting objects that have not completed initialization

lijinxiong:
in org.springframework.beans.factory.support.DefaultSingletonBeanRegistry

```java
@Nullable
protected Object getSingleton(String beanName, boolean allowEarlyReference) {
// Quick check for existing instance without full singleton lock
Object singletonObject = this.singletonObjects.get(beanName);
if (singletonObject == null && isSingletonCurrentlyInCreation(beanName)) {
	singletonObject = this.earlySingletonObjects.get(beanName);
	if (singletonObject == null && allowEarlyReference) {
		synchronized (this.singletonObjects) {
			// Consistent creation of early reference within full singleton lock
			singletonObject = this.singletonObjects.get(beanName);
			if (singletonObject == null) {
				singletonObject = this.earlySingletonObjects.get(beanName);
				if (singletonObject == null) {
					ObjectFactory<?> singletonFactory = this.singletonFactories.get(beanName);
					if (singletonFactory != null) {
						singletonObject = singletonFactory.getObject();
						this.earlySingletonObjects.put(beanName, singletonObject);
						this.singletonFactories.remove(beanName);
					}
				}
			}
		}
	}
}
return singletonObject;
}
```

Unit test

```java
public class Father {

	private Son son;

	public Son getSon() {
		return son;
	}

	public void setSon(Son son) {
		this.son = son;
	}
}
```

```java
public class Son {

	private Father father;

	private Mother mother;

	public Father getFather() {
		return father;
	}

	public void setFather(Father father) {
		this.father = father;
	}

	public Mother getMother() {
		return mother;
	}

	public void setMother(Mother mother) {
		this.mother = mother;
	}
}
```

```java
public class Mother {

	public Mother() {

		// Simulation takes a long time、just for testing
		try {
			Thread.sleep(4 * 1000);
		} catch (InterruptedException e) {
			e.printStackTrace();
		}

	}
}

```

```java
	@Test
	public void testGetLazySingleton() throws InterruptedException {
		testGetSingleton("father");
	}

	@Test
	public void testGetNonLazySingleton() throws InterruptedException {
		testGetSingleton("father-non-lazy");
	}

	private void testGetSingleton(String fatherBeanName) throws InterruptedException {

		CountDownLatch countDownLatch = new CountDownLatch(2);
		AtomicInteger occurExceptionCount = new AtomicInteger(0);
		DefaultListableBeanFactory beanFactory = new DefaultListableBeanFactory();
[...]

PR #31262, opened 2023-09-18
Fix an issue with uninitialized beans being fetched when circular dependencies are used

nineya:
The PR changed the org. Springframework. Beans. Factory. Support. DefaultListableBeanFactory resolveNamedBean method, To resolve the issue of getting an uninitialized bean that can occur when concurrently executing getBean to get a bean that contains a loop dependency.
Ref: #31261


snicoll:
See also #26376

jhoeller:
For circular reference resolution, any access to a Spring bean needs to be able to receive an early-exposed singleton whenever needed, so we can't artificially block in `resolveNamedBean` there. If you'd like to avoid the circular reference problem completely, set your factory to `setAllowCircularReferences(false)` which will avoid any exposure to `earlySingletonObjects` to begin with, therefore never returning an early singleton object from `resolveNamedBean` either.

Note that we are considering to disallow circular references by default in a future Spring Framework generation. However, as long as we need to be able to resolve circular references, we need to keep those bean retrieval methods as lenient as they are.

PR #31279, opened 2023-09-20
Refactoring getSingleton method to improve readability

yinabameguru:
Recently I was learning source code and wanted to make some contributions to the project, so I tried refactoring the getSingleton method

Inverting the logic of the if statements and putting return inside of these if statements.  The benefit is the code will tend to be readability and less deeply indented.


snicoll:
@yinabameguru thanks but we generally don't accept PR that are stylistic change as it is mostly subjective. If we did accept this change, there are many places in the framework that would also need to change for consistency.

PR #31291, opened 2023-09-22
Stop sharing `GenericConversionService` between test methods

quaff:
Some tests called mutating methods like `addConverterFactory()`, it will potentially affect other tests.

### DIFF
--- a/spring-core/src/test/java/org/springframework/core/convert/support/GenericConversionServiceTests.java
+++ b/spring-core/src/test/java/org/springframework/core/convert/support/GenericConversionServiceTests.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


spring-core/src/test/java/org/springframework/core/convert/support/GenericConversionServiceTests.java
```diff
@@ -61,11 +62,16 @@
...
  */
 class GenericConversionServiceTests {
 
-	private final GenericConversionService conversionService = new GenericConversionService();
+	private GenericConversionService conversionService;
 
+	@BeforeEach
+	void init() {
+		conversionService = new GenericConversionService();
+	}
```
sbrannen, line 74: This test class uses JUnit Jupiter with the default `@TestInstance` lifecycle mode which is `PER_METHOD`, meaning that the test instance will be instantiated for each test method.

Consequently, the `GenericConversionService` will be recreated for every instance/test.

sbrannen:
Closing as "works as designed".

See https://github.com/spring-projects/spring-framework/pull/31291#pullrequestreview-1640117970.

PR #31299, opened 2023-09-23
Upgrade to snakeyaml 2.2

Sineaggi:
Snakeyaml 2.2 [has a fix](https://bitbucket.org/snakeyaml/snakeyaml/wiki/Changes) where the package `org.yaml.snakeyaml.inspector` is not exported from its module, requiring a manual `--add-exports` when running spring on the module path.

### DIFF
--- a/framework-platform/framework-platform.gradle
+++ b/framework-platform/framework-platform.gradle
@@ -148,6 +148,6 @@ dependencies {
 		api("org.webjars:webjars-locator-core:0.53")
 		api("org.xmlunit:xmlunit-assertj:2.9.1")
 		api("org.xmlunit:xmlunit-matchers:2.9.1")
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks but Spring Framework only has an optional dependency on this library and only compiles against it. We would rather deal with dependency upgrades and baseline upgrades ourselves.

You should instead upgrade this dependency directly in your application as Spring Framework will not enforce any version for this library and does not list it in its POMs.

Sineaggi:
I have a branch that's slowly attempting to migrate spring-framework to the module system, and it's the "compiles against it" part that's failing (specifically due to the `org.yaml.snakeyaml.inspector` module not being exported). Hence this mr.

Issue #21758, opened 2018-08-29
Support `BigInteger` and `BigDecimal` literals in SpEL expressions


spring-projects-issues:
**[Rimal](https://jira.spring.io/secure/ViewProfile.jspa?name=rimal)** opened **[SPR-17225](https://jira.spring.io/browse/SPR-17225?redirect=false)** and commented

Can we please add a suffix support for BigInteger as well in SpEL. 

Currently, we can add a long value as 12345456L (with the L suffix) but for a BigInteger, we need to mention it as new `java.math.BigInteger('9248040402')`.

It would be really helpful if we could mention a BigInteger as 9248040402BI or something on these lines.

---
No further details from [SPR-17225](https://jira.spring.io/browse/SPR-17225?redirect=false)

ovmarkushyn:
@spring-projects-issues 
Could you please consider [PR#31301](https://github.com/spring-projects/spring-framework/pull/31301) as a possible solution for the request #21758.

sbrannen:
**Team Decision**:

We consider the proposal to be a bit of a niche feature which does not warrant the complexity it would introduce within the Spring Expression Language (see PR #31301).

In general, at this point in time we do not intend to introduce new language-level features in the Spring Expression Language.

In light of that, we are closing this issue.

In any case, thank you for the proposal!

sbrannen:
Please note that it's always possible to register your own custom functions for such purposes, as long as you are in control of the setup of the `EvaluationContext`.

For example, the following test demonstrates how to register a custom `bigInt` function and use it within expressions instead of `new java.math.BigInteger('9248040402')`.

```java
@Test
void bigIntegerFunction() throws Exception {
	SpelExpressionParser parser = new SpelExpressionParser();
	StandardEvaluationContext context = new StandardEvaluationContext();
	Method method = BigInteger.class.getMethod("valueOf", long.class);
	context.registerFunction("bigInt", method);

	Expression expression = parser.parseExpression("3 + #bigInt(5)");
	BigInteger result = expression.getValue(context, BigInteger.class);
	assertThat(result).isEqualTo(BigInteger.valueOf(8));
}
```

And you could register a similar function (for example, `bitDecimal`) for `BigDecimal` support.

Hope you find this useful!

PR #31301, opened 2023-09-23
Support `BigInteger` and `BigDecimal` literals in SpEL expressions

ovmarkushyn:
### Context

The change satisfies the request made in #21758 which involves adding support for a suffix to `BigInteger`, similar to 'L' for `long`.

### Usage

The proposed solution makes it possible to use both `BigInteger` and `BigDecimal` for both HEX and non-HEX number literals.

The suffixes `BI` and `BD` are used for `BigInteger` and `BigDecimal`, respectively. Both suffixes can be specified in lower/upper case or mixed.

### Solution

Consider please diagram attached.

It shows all the elements that have been changed as part of the solution.

**Yellow** - represent existing elements.

**Green** - "big number" (`BigInteger` and `BigDecimal`) extension of `Tokenizer`.

In general, the solution is implemented similarly to the `L` support, aiming for as much consistency as possible with the current design.

Whole "big number" related code is isolated in 2 inner classes, one for **HEX** literal (Big**Hex**IntegerConcernExtension) and another one for **non-HEX** one (Big**NonHex**NumberConcernExtension) both having common parent `BigNumberConcernExtension`.

`BigNumberConcernExtension` keeps all the logic related to suffix detection, extraction, position shifting.

`BigHexIntegerConcernExtension` and `BigNonHexNumberConcernExtension` are responsible for handling the specifics of token addition.

[](https://github.com/spring-projects/spring-framework/assets/1871503/6ec2f923-933e-472d-a6b0-01d6477d9556)


spring-expression/src/main/java/org/springframework/expression/spel/BigNumberConcern.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Marker of "big number" ({@link java.math.BigInteger} and {@link java.math.BigDecimal}) concern extension for SpEL engine.
+ * Solution extends related elements similar to how L/l suffix support is done.
+ * It covers: HEX, non-HEX(including exponent) literals.
+ * Requirement: <a href="https://github.com/spring-projects/spring-framework/issues/21758">Suffix support for BigInteger in SPEL [SPR-17225] #21758</a>
+ *
+ * @author Oleksandr Markushyn
+ */
+public @interface BigNumberConcern {
```
sbrannen, line 27: Please remove the `@BigNumberConcern` from this PR.

We do not use marker annotations in Spring Framework to track features, and its presence in this PR makes it harder to review the actual changes in the PR.

sbrannen (changes requested):

spring-expression/src/main/java/org/springframework/expression/spel/BigNumberConcern.java
ovmarkushyn: Removed.

ovmarkushyn:
@sbrannen 
BigNumberConcern is also removed from PR description and diagram.

sbrannen:
Hi @ovmarkushyn,

In light of https://github.com/spring-projects/spring-framework/issues/21758#issuecomment-1937763112, we are closing this PR as well.

Having said that, we do appreciate the hard work you put into this PR! 👍 

Thanks,

Sam

PR #31305, opened 2023-09-24
Remove unused imports

alidandach:
### Summary:

- Switched from direct reference (!=) to Objects.equals() for string comparisons to ensure consistent content comparison, null safety, and clearer code intention.
- Remove unused imports

### DIFF
--- a/buildSrc/src/main/java/org/springframework/build/JavaConventions.java
+++ b/buildSrc/src/main/java/org/springframework/build/JavaConventions.java
@@ -27,7 +27,6 @@ import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginExtension;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.jvm.toolchain.JavaLanguageVersion;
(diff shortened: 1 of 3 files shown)


sbrannen:
The proposed modification to `SpringExtension` has already been rejected twice: in #30962 and #30456.

There was in fact nothing wrong with the implementation, and it worked as designed.

> Switched from direct reference (!=) to Objects.equals() for string comparisons to ensure consistent content comparison, null safety, and clearer code intention.

There was no issue with null safety, and there was no need to compare the content of the strings.

`NO_VIOLATIONS_DETECTED` is only intended to serve as a _marker_ `String` (or a `null` object in the sense of the "Null Object Pattern") which in `SpringExtension` serves the purpose of a boolean `false` value without the need to compare the content of two strings for every single test method invocation in a test suite.

However, since this is the 3rd PR to suggest a change to this code, I concede that the intention could be clearer.

In light of that, I'm repurposing this PR to address the removal of the unused imports, and I'll push a different revision to `SpringExtension` to make the intention clearer.

sbrannen:
See commit 7aae97bcd469105ac2f2aa76ce93f38df455e3b9.

PR #26028, opened 2020-11-04
Use `computeIfAbsent` in `SpelCompiler` and `AdvisedSupport`

hzmpay:
Use computeIfAbsent to simplify some code, in addition, the content of some Javadoc has been adjusted.

PR #31306, opened 2023-09-25
Optimize AdvisedSupport#getInterceptorsAndDynamicInterceptionAdvice

kang-hl:


snicoll:
@kang-hl can you share the optimization here?

snicoll:
That's somehow reverting #26028

PR #31309, opened 2023-09-25
Remove `MethodCacheKey`

kang-hl:
When calling the computeIfAbsent method, the hashCode method will only be called once, and the equals method will probably not be called. Using MethodCacheKey is inefficient.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
@@ -63,6 +63,7 @@ import org.springframework.util.ObjectUtils;
  * @author Rod Johnson
  * @author Juergen Hoeller
  * @author Sam Brannen
+ * @author Hailin Kang
  * @see org.springframework.aop.framework.AopProxy
  */
 public class AdvisedSupport extends ProxyConfig implements Advised {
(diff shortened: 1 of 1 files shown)


jhoeller:
The use of `MethodCacheKey` in an isolated `ConcurrentHashMap` benchmark is consistently faster than using `Method` itself as a key there, possibly due to the pre-computed hash codes of the existing entries in the cache. In a regular scenario, it is about 30% faster; with overloaded methods in the mix (which leads to Method hash code collisions and `equals` calls), it is even 40% faster.

As a consequence, it seems sensible to preserve our existing caching arrangement there.

kang-hl:
Are you benchmarking with the latest code? If not, you should use the latest code.




---Original---
From: "Juergen ***@***.***&gt;
Date: Fri, Nov 24, 2023 00:19 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [spring-projects/spring-framework] Remove `MethodCacheKey` (PR#31309)




 
The use of MethodCacheKey in an isolated ConcurrentHashMap benchmark is consistently faster than using Method itself as a key there, possibly due to the pre-computed hash codes of the existing entries in the cache. In a regular scenario, it is about 30% faster; with overloaded methods in the mix (which leads to Method hash code collisions and equals calls), it is even 40% faster.
 
As a consequence, it seems sensible to preserve our existing caching arrangement there.

PR #31312, opened 2023-09-25
Add FastjsonHttpMessageConverter

mek1986:
add the com.alibaba.fastjson2 support for HttpMessage Converter.

### DIFF
--- a/framework-platform/framework-platform.gradle
+++ b/framework-platform/framework-platform.gradle
@@ -23,6 +23,7 @@ dependencies {
 	api(platform("org.mockito:mockito-bom:5.5.0"))
 
 	constraints {
+		api("com.alibaba.fastjson2:fastjson2:2.0.40")
 		api("com.fasterxml:aalto-xml:1.3.2")
 		api("com.fasterxml.woodstox:woodstox-core:6.5.1")
 		api("com.github.ben-manes.caffeine:caffeine:3.1.8")
--- a/spring-messaging/spring-messaging.gradle
+++ b/spring-messaging/spring-messaging.gradle
@@ -8,6 +8,7 @@ dependencies {
 	api(project(":spring-core"))
 	optional(project(":spring-context"))
 	optional(project(":spring-oxm"))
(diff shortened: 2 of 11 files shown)


bclozel:
@mek1986 thanks for the PR, but discussing it beforehand would have been a good first step. I'm not sure the Spring wants to add another variant for this. Spring Boot starters are well designed for this and I think we might reduce the auto-detection in Framework in the future anyway. We will discuss that as a team and get back to you.

mek1986:
> @mek1986 thanks for the PR, but discussing it beforehand would have been a good first step. I'm not sure the Spring wants to add another variant for this. Spring Boot starters are well designed for this and I think we might reduce the auto-detection in Framework in the future anyway. We will discuss that as a team and get back to you.

ok,thank you.

wenshao:
[fastjson](https://github.com/alibaba/fastjson) is very widely used in China, and [fastjson2](https://github.com/alibaba/fastjson2) is an upgraded version of fastjson. Spring's built-in support for fastjson will make it more convenient for Chinese users.

bclozel:
It looks like [fastjson2 already offers excellent support for Spring Framework](https://github.com/alibaba/fastjson2/blob/main/docs/spring_support_cn.md) 5 and 6. Maybe the only missing part would be a Spring Boot starter to auto-configure this?

I'm not sure merging this would be an improvement. Instead of improving support it would be harder for the fastjson team to evolve the converter along the core support. Spring Framework has backwards compatibility constraints and cannot adopt the latest changes in libraries without raising the baseline.

Maybe we can link to fastjson support in the reference docs directly? This would make fastjson more discoverable to users worldwide (not just china) and would avoid the downsides of merging the support here.

bclozel:
We've discussed this and we think that a link from the [Spring Boot community starters page](https://github.com/spring-projects/spring-boot/tree/main/spring-boot-project/spring-boot-starters#community-contributions) would be a better fit, if in the future the fastjson team decides to release such an artifact with their project.

PR #31313, opened 2023-09-25
Use lambda expressions for message handlers in `StandardWebSocketHandlerAdapter`

alidandach:
This PR removes anonymous classes and uses lambda expressions instead which reduce the verbosity of code and enhance the readability of 'onOpen` method.


sbrannen:
@alidandach, we cannot accept this PR since it breaks the code.

In the future, please make sure you read the comments in the code you are modifying.

```java
// The following inner classes need to remain since lambdas would not retain their
// declared generic types (which need to be seen by the underlying WebSocket engine)
```

PR #31316, opened 2023-09-26
Allow capacity = 0 in `ConcurrentLruCache` again

kcichonczyk:
Fix for ConcurrentLruCache initialised with capacity = 0, previous behaviour allowed capacity = 0 (current public constructor allows it according to its javadoc)


kcichonczyk:

PR #31320, opened 2023-09-27
Polish Conditional Statements

chaewon121:
I found that redundant validations were affecting readability. 
Therefore, I refactored the code by removing the redundant validations and improved readability.


snicoll:
Thank you for making your first contribution to Spring Framework. FTR, I've aligned the reactive and mvc versions and added missing tests, see [7a05d23](https://github.com/spring-projects/spring-framework/commit/7a05d23597e765762148aa32ecbec89c7497810e)

Issue #31114, opened 2023-08-25
Instrument RestClient for Observability


bclozel:
`RestClient` has been recently introduced and we should instrument it for observability.
This client variant uses the same infrastructure as the `RestTemplate`, so we should share the observation context and conventions. The client style is more functional, so the instrumentation itself will be similar to `WebClient`.

PR #31325, opened 2023-09-27
Allow custom observation convention for `RestClient`

hpoettker:
The PR adds the possibility to use a custom `ObservationConvention` in the `RestClient`, and to set it through its builder.

Observability has been added to `RestClient` in #31114 based on the same infrastructure used by `RestTemplate`. Setting a custom `ObservationConvention` is supported with both `RestTemplate` and `WebClient`, and I was surprised that `RestClient` currently does not support it.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
@@ -107,6 +107,9 @@ final class DefaultRestClient implements RestClient {
 
 	private final ObservationRegistry observationRegistry;
 
+	@Nullable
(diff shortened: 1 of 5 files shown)


bclozel:
Custom observation conventions can be configured on the registry itself directly. While this change aligns more with the other clients there are other instrumentations in Framework that don't offer a configuration option on the instrumented directly.

I'll discuss that with the observability team to get their opinion.

bclozel:
Thanks @hpoettker , we've merged your contribution. We're mostly doing this for consistency with other HTTP clients but we don't intend to provide this option for all instrumentations as they can be configured with `GlobalObservationConvention` as well.

PR #31328, opened 2023-09-27
Fix regression in JmsAccessor#setSessionAcknowledgeMode

vpavic:
This commit partially reverts 3b8dd0a5, which introduced a strict validation in `JmsAccessor#setSessionAcknowledgeMode` that prevents use of vendor-specific acknowledge modes.

---

See also:
- https://github.com/spring-projects/spring-boot/pull/37576


snicoll:
Thanks again @vpavic

PR #31330, opened 2023-09-28
Add missing `conversionService` field in doc example

jihuayu:


snicoll:
@jihuayu good catch and thank you for making your first contribution to Spring Framework.

PR #31346, opened 2023-10-02
Make `CorsConfiguration.OriginPattern` public

Tr1ple-F:
The nested `OriginPattern` class in `CorsConfiguration` is amazingly useful and should be available for re-use by applications.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
+++ b/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
@@ -725,7 +725,7 @@ public class CorsConfiguration {
 	 * Contains both the user-declared pattern (e.g. "https://*.domain.com") and
 	 * the regex {@link Pattern} derived from it.
 	 */
-	private static class OriginPattern {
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @Tr1ple-F,

What concrete need do you have for using `OriginPattern` in your own application?

Tr1ple-F:
I am extending OAuth2RedirectStrategy and only want to allow a given subset
of redirect patterns (similar to the CORS Origins). For this it seems
better to me to reuse the existing OriginPattern class.

Also I don't see a reason why it should be specifically private.

Sam Brannen ***@***.***> schrieb am Mo., 2. Okt. 2023, 15:14:

rstoyanchev (commented):
`OriginPattern` is mainly for internal use. When providing public API we would need to put more thought into what is exposed, and ensure it is broadly applicable since once it is public we would need to be prepared to enhance and evolve what it does, e.g. nested or top-level, builder, support for additional syntax, etc.

Considering that it is a fairly small and straight forward class, it would be less trouble if you incorporate what it does into your own code.

PR #31352, opened 2023-10-03
Optimize uses of `onErrorResume()`

kzander91:
Replace uses of `onErrorResume()` with
- `onErrorMap()` in places where it is just used to map to a different exception.
- `onErrorComplete()` where it just maps to `Mono.empty()`.
- `onErrorReturn()` where it just maps to `Mono.just()`.

The first change achieves the same using less code, while the latter two changes improve efficiency as of Reactor 3.4.23 (see reactor/reactor-core#3159) by avoiding the unnecessary creation of `Mono`/`Flux` instances.


bclozel:
Thanks @kzander91 !

PR #31364, opened 2023-10-05
EMFU#doGetTransactionalEntityManager to check for actual transaction

ettavolt:
… rather than just a synchronization. Otherwise, at least SharedEntityManagerCreator will create a 'lingering' EM inside NEVER/NOT_SUPPORTED propagation, which is closed only after exiting from inside that boundary.

I did look at some other usage of isSynchronizationActive, and I'm not sure if it is more proper to use isActualTransactionActive there. I suspect this might be true for a couple of places.

isActualTransactionActive and doGetTransactionalEntityManager seem to be [around](https://github.com/spring-projects/spring-framework/blame/a3d7dc09ef7af361705704ca9f457b4901815663/spring-orm/src/main/java/org/springframework/orm/jpa/EntityManagerFactoryUtils.java) [here](https://github.com/spring-projects/spring-framework/blame/main/spring-tx/src/main/java/org/springframework/transaction/support/TransactionSynchronizationManager.java) since before the recorded history, so we can't say if isActualTransactionActive was added after or before doGetTransactionalEntityManager.

Edit of #29430, which I messed up by pushing to main in my repository.

### DIFF
--- a/integration-tests/src/test/java/org/springframework/orm/jpa/SharedEntityManagerCreatorTransactionSyncTests.java
+++ b/integration-tests/src/test/java/org/springframework/orm/jpa/SharedEntityManagerCreatorTransactionSyncTests.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2022-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.orm.jpa;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.EntityManagerFactory;
+import jakarta.persistence.Query;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.support.AbstractPlatformTransactionManager;
+import org.springframework.transaction.support.TransactionTemplate;
+
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+/**
+ * @author Arseniy Skvortsov
+ */
+@ExtendWith(MockitoExtension.class)
+public class SharedEntityManagerCreatorTransactionSyncTests {
+
+	@Test
+	public void noSharingInNeverPropagation() {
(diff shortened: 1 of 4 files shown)


spring-orm/src/test/java/org/springframework/orm/jpa/JpaTransactionManagerTests.java
```diff
@@ -417,9 +418,9 @@ public void testPropagationSupportsAndRequiresNewAndEarlyAccess() {
 		assertThat(TransactionSynchronizationManager.isSynchronizationActive()).isFalse();
 
 		Object result = tt.execute(status -> {
-			EntityManagerFactoryUtils.getTransactionalEntityManager(factory);
+			assertThat(EntityManagerFactoryUtils.getTransactionalEntityManager(factory)).isNull();
```
ettavolt, line 421: PROPAGATION_SUPPORTS says, that if there is no existing transaction, the code will be executed non-transactionally, and, thus it makes sense for get Transactional EntityManager to return null.

That also means no EM resource is bound to the thread via the factory.

jhoeller:
Sorry for noticing your original issue before: It's not just about actual transactions, that code is about reusing an `EntityManager` for a given boundary intentionally. As long as we have a managed boundary there, we'll expose the same `EntityManager` instance within that scope, reusing its first-level cache. This has been the case ever since the inception of Spring's JPA support. So it works by design, accepting a longer-open `EntityManager` handle for reuse within the scope.

While it is true that such resource handles may 'linger' when a transaction gets suspended, such a scenario is usually not good practice to begin with. The common suspension scenarios have `REQUIRED` (potentially with `readOnly=true`) in combination with `REQUIRES_NEW`, whereas `SUPPORTS` is mostly used on its own for read-only scenarios where no actual transaction is desired. `NOT_SUPPORTED` and `NEVER` are mostly there for completeness (aligned with EJB), we rarely see those in practice.

An open `EntityManager` does not necessarily mean a bound JDBC `Connection`. In case of longer-lived `EntityManager` instances, the connection release mode in the persistence provider can be configured accordingly, aggressively releasing each obtained `Connection` back to the pool after every operation. With such a setup, the `EntityManager` is quite literally just a first-level cache for a particular processing step in the application, occupying some memory but not holding any resources.

Last but not least, synchronization can be turned off on the transaction manager. `JpaTransactionManager` will then only manage an actual transactional `EntityManager` for each required boundary (pre-exposed by `JpaTransactionManager` for each transaction rather than lazily obtained through `doGetTransactionalEntityManager`), not accepting any other scoped resources.

integration-tests/src/test/java/org/springframework/orm/jpa/SharedEntityManagerCreatorTransactionSyncTests.java
```diff
@@ -0,0 +1,62 @@
...
+		given(em.createQuery("x")).willReturn(query);
+		given(em.isOpen()).willReturn(true);
+		tt.setPropagationBehavior(TransactionDefinition.PROPAGATION_NEVER);
+		tt.executeWithoutResult(status -> {
+			SharedEntityManagerCreator
+					.createSharedEntityManager(emf)
+					.createQuery("x")
+					.executeUpdate();
+			verify(query).executeUpdate();
+			verify(em).close();
```
ettavolt, line 59: @jhoeller, what do you think about this expectation? Namely, that transaction is closed after one-off `executeUpdate` inside NEVER propagation?

PR #31381, opened 2023-10-08
Remove redundant `static` declaration on `Search` class

OnlyWick:
I found a very small mistake. : )

[9.3. Field (Constant) Declarations.](https://docs.oracle.com/javase/specs/jls/se17/html/jls-9.html#jls-9.3)

> Every field declaration in the body of an interface delaration is implicitly public, static, and final. It is permitted to redundantly specify any or all of these modifiers for such fields.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/MergedAnnotations.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/MergedAnnotations.java
@@ -507,7 +507,7 @@ public interface MergedAnnotations extends Iterable<MergedAnnotation<Annotation>
 	 *
 	 * @since 6.0
 	 */
-	static final class Search {
+	class Search {
 
 		static final Predicate<Class<?>> always = clazz -> true;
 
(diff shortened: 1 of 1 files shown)


OnlyWick:
@sbrannen

OnlyWick:
As described in the Java Language Specification, I feel like simply removing "static final" would suffice. I can't think of any other side effects at the moment. If there are any other side effects, please let me know.

OnlyWick:
I just woke up... I haven't slept all night and I forgot to check the other properties of the interface.

OnlyWick:
"final" does not need to be removed, but you have also made the constructor private.

OnlyWick:
So, I just need to remove "static". @sbrannen

sbrannen:
Hi @OnlyWick,

For future reference, please do not @-mention individual maintainers of the framework. Somebody from the team will get to your PR when they have time.

Also, please try to refrain from submitting PRs with a single small change like that. We prefer to manage such changes in batches. So if you find something like that, please attempt to find all such occurrences in the code base and submit a single PR to address the issue on a broader scale.

Thanks

sbrannen:
> [9.3. Field (Constant) Declarations.](https://docs.oracle.com/javase/specs/jls/se17/html/jls-9.html#jls-9.3)
> 
> > Every field declaration in the body of an interface delaration is implicitly public, static, and final. It is permitted to redundantly specify any or all of these modifiers for such fields.

I assume you meant to reference "9.5. Member Class and Interface Declarations".

PR #31384, opened 2023-10-09
Remove jdbc execute "closeResources" parameter

thdwoqor:
I don't know why the `boolean closeResources` parameter is needed
Except for the `queryForStream` method, virtually all of them are `true` and there is no reason for `queryForStream` to be false either

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
@@ -377,8 +377,9 @@ public class JdbcTemplate extends JdbcAccessor implements JdbcOperations {
 	// Methods dealing with static SQL (java.sql.Statement)
 	//-------------------------------------------------------------------------
 
+	@Override
 	@Nullable
-	private <T> T execute(StatementCallback<T> action, boolean closeResources) throws DataAccessException {
+	public  <T> T execute(StatementCallback<T> action) throws DataAccessException {
 		Assert.notNull(action, "Callback object must not be null");
 
 		Connection con = DataSourceUtils.getConnection(obtainDataSource());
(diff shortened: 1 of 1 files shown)


snicoll:
> there is no reason for queryForStream to be false either

What makes you think that?

sbrannen:
Hi @thdwoqor,

Unfortunately we cannot accept this, since it breaks the contract of `queryForStream()`.

> I don't know why the `boolean closeResources` parameter is needed Except for the `queryForStream` method, virtually all of them are `true` and there is no reason for `queryForStream` to be false either

`queryForStream()` returns a `Stream` that the user is required to close, as per the documentation in the Javadoc for that method.

The reason that `queryForStream()` invokes `execute(new StreamStatementCallback(), false)` is that the `StreamStatementCallback` has an `onClose()` method which closes the resources.

In light of that, I am closing this issue as invalid.

thdwoqor:
@snicoll 
@sbrannen

```java
@Override
	public <T> Stream<T> queryForStream(String sql, RowMapper<T> rowMapper) throws DataAccessException {
		class StreamStatementCallback implements StatementCallback<Stream<T>>, SqlProvider {
			@Override
			public Stream<T> doInStatement(Statement stmt) throws SQLException {
				ResultSet rs = stmt.executeQuery(sql);
				Connection con = stmt.getConnection();
				return new ResultSetSpliterator<>(rs, rowMapper).stream().onClose(() -> {
					JdbcUtils.closeResultSet(rs);
					JdbcUtils.closeStatement(stmt); // this
					DataSourceUtils.releaseConnection(con, getDataSource()); // this
				});
			}
			@Override
			public String getSql() {
				return sql;
			}
		}

		return result(execute(new StreamStatementCallback(), false));
	}
```

The queryForStream method uses the 
closeStatement and releaseConnection in doInStatement, and isn't that effectively closeResources true?

sbrannen:
> The queryForStream method uses the
> closeStatement and releaseConnection in doInStatement, and isn't that effectively closeResources true?

No.

It actually does not invoke that in `doInStatement`.

Rather, as I mentioned in https://github.com/spring-projects/spring-framework/pull/31384#issuecomment-1752919505, that code is a lambda expression supplied to `onClose()` which is only invoked if the user closes the `Stream`.

sbrannen:
@thdwoqor, if you have further questions, please keep our policy in mind:

Thanks for getting in touch, but it feels like this is a question that would be better suited to [Stack Overflow](http://stackoverflow.com/). As mentioned in [the guidelines for contributing](https://github.com/spring-projects/spring-framework/blob/master/CONTRIBUTING.md#discuss), we prefer to use the issue tracker only for bugs and enhancements. Feel free to update this issue with a link to the re-posted question (so that other people can find it) or add some more details if you feel this is a genuine bug.

PR #31415, opened 2023-10-11
Replace deprecated method `getBuildDir()`

GVictorG7:
According to the [documentation ](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getBuildDir--), the method `getBuildDir()` is deprecated and should be replaced with `getLayout().getBuildDirectory()`.


GVictorG7:
The implementation of the deprecated method `getBuildDir()` is `org.gradle.api.internal.project.DefaultProject#getBuildDir`, and to be more explicit:
`    @Deprecated
    public File getBuildDir() {
        return (File)this.getLayout().getBuildDirectory().getAsFile().get();
    }`
Therefore, in order to remove the usage of deprecated method and the warning, a solution is simply to apply the method implementation directly in the code.

PR #31432, opened 2023-10-14
Update to io.freefair.aspectj 8.4

larsgrefer:
Starting with version 8.4 the custom configuration to change the source directory from `src/*/aspectj` to `src/*/java` is not necessary anymore.

see https://github.com/freefair/gradle-plugins/issues/881


bclozel:
Thanks @larsgrefer for maintaining this plugin, it really makes our life easier!

PR #31434, opened 2023-10-15
Polish lambda expressions usage

GVictorG7:
Polish lambda expressions usage from `buildSrc` module and remove associated compiler and sonar warnings.


bclozel:
Sorry but we're fine the way things are. We don't use Sonar in our build and our style checks are happy with this part of the codebase.

PR #31460, opened 2023-10-19
Add support for Gradle plugin org.owasp.dependencycheck

wfouche:
https://owasp.org/www-project-dependency-check/ :  "Dependency-Check is a Software Composition Analysis (SCA) tool that attempts to detect publicly disclosed vulnerabilities contained within a project’s dependencies. It does this by determining if there is a Common Platform Enumeration (CPE) identifier for a given dependency. If found, it will generate a report linking to the associated CVE entries."

Start the OWASP dependency-check scan with command:

$ ./gradlew dependencyCheckAggregate

The dependency-check security vulnerability report is located at: **./build/reports/dependency-check-report.html**


wfouche:
Of course if works on my machine, and produces great results and shows vulnerabilities in htmlunit 2.70 and other JARs.   :-)

[dependency-check-report.zip](https://github.com/spring-projects/spring-framework/files/13047910/dependency-check-report.zip)

PR #31463, opened 2023-10-20
JdbcTemplate should store generated keys only if database driver supports it 

maciejmiklas:
Not every database supports `PreparedStatement#getGeneratedKeys()`, for example, SQLLitle does not, and calling this method throws an exception, it has been discussed here: https://github.com/xerial/sqlite-jdbc/issues/996 

`JdbcTemplate`  calls `PreparedStatement#getGeneratedKeys()` whether it's supported or not, and in the case of SQLLitle, it causes an exception.

This PR changes the behavior of `JdbcTemplate` that it will call `PreparedStatement#getGeneratedKeys()` only if the underlying driver supports it by querying:  `DatabaseMetaData#supportsGetGeneratedKeys()`.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
@@ -1568,6 +1568,10 @@ public class JdbcTemplate extends JdbcAccessor implements JdbcOperations {
 	}
 
 	private void storeGeneratedKeys(KeyHolder generatedKeyHolder, PreparedStatement ps, int rowsExpected) throws SQLException {
+		if(!JdbcUtils.supportsGeneratedKeys(ps.getConnection())){
+			logger.trace("Generated keys is not supported");
+			return;
+		}
 		List<Map<String, Object>> generatedKeys = generatedKeyHolder.getKeyList();
 		ResultSet keys = ps.getGeneratedKeys();
 		if (keys != null) {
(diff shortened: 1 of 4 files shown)


gotson:
It might be worth reading at similar questions asked on jooq, as apparently `supportsGetGeneratedKeys` may not be reliable in some drivers.

https://github.com/jOOQ/jOOQ/issues/8996#issuecomment-520735606

There's also https://github.com/jOOQ/jOOQ/issues/8993#issuecomment-515900680 where apparently returning null from `getGeneratedKeys`, instead an empty resultset should be returned.

maciejmiklas:
This is a good argument - thank you for checking it.

jhoeller:
So your code is calling a `JdbcTemplate` operation with a `KeyHolder` argument there but would be happy for that key holder not to be populated? Note that there are non-KeyHolder overloads of those methods which never trigger a `getGeneratedKeys` call to begin with, in case you do not need those keys anyway.

If we consider key retrieval as sort-of-optional even for our `KeyHolder`-based operations, we could simply catch `SQLFeatureNotSupportedException` on actual access, I suppose, just like we check for `null` there. That would avoid any potential inconsistency with `DatabaseMetaData` support checks.

However, I'm still concerned by the mismatch with the javadoc for our `KeyHolder`-based operations which explicitly state that "generated keys will be put into the given KeyHolder". If the JDBC driver does not support key retrieval, arguably those `JdbcTemplate` operations are meant to fail...

maciejmiklas:
@jhoeller you are correct - we can use execute instead of update.

snicoll:
Thanks for the follow-up. I've created #31486 to review the Javadoc.

prrvchr:
Hi all,
Here is a JDBC 4.1 compatible SQLite version with `getGeneratedKeys()` support.
[sqlite-jdbc-3.45.1.5-SNAPSHOT.jar](https://github.com/prrvchr/sqlite-jdbc/releases/download/3.45.1.5-SNAPSHOT/sqlite-jdbc-3.45.1.5-SNAPSHOT.jar)

PR #31471, opened 2023-10-22
Fix typo in RootBeanDefinition Api Doc

sen-woo:
The 'GenericBeanDefinition' link and words do not match, so I think it would be a good idea to modify it.


bclozel:
This is by design and makes sense in the context of the sentence. The [link label](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#link) is plural in order to be grammatically correct here.

Thanks!

PR #31483, opened 2023-10-24
Resolve to empty `MultiValueMap` when no matrix variables are provided

bernie-schelberg-invicara:
When no matrix variables are provided to an endpoint with a `MultiValueMap` parameter, the argument should resolve to a `MultiValueMap`, not to `Collections.emptyMap()`.


snicoll:
@bernie-schelberg-invicara good catch and thank you for making your first contribution to Spring Framework. I've harmonized the method so that it has to return a `MultiValueMap` for consistency. I've also adapted the reactive variant. See [9aa707e](https://github.com/spring-projects/spring-framework/commit/9aa707ec4b04fe5a27d93143a42bec8cd30c7666)

PR #31506, opened 2023-10-26
Support direct shard database operation routing in Spring JDBC

meedbek:
This PR is a follow up to https://github.com/spring-projects/spring-framework/pull/30988.
Goal is to add support for direct database operation routing in sharded databases in Spring JDBC. 

This functionality is achieved by acquiring a shard connection through the JDBC API 
DataSource.createConnectionBuilder().shardingKey(key).build().

In this PR we will focus on one of the two approaches that were introduced in the previous PR.

We introduce a DataSource adapter class that applies sharding keys to every standard getConnection() call. The sharding keys are specified through a ShardingKeyProvider object.

Here's an example of how to use it:

```java
ShardingKeyDataSourceAdapter dataSourceAdapter = new ShardingKeyDataSourceAdapter(dataSource);

dataSourceAdapter.setShardingKeyProvider(new ShardingKeyProvider() {
    public ShardingKey getShardingKey() throws SQLException {
        // Example: The sharding key is derived from the user name.
        String name = SecurityContextHolder.getContext().getAuthentication().getName();
        return dataSource.createShardingKeyBuilder().subKey(name, JDBCType.VARCHAR).build();
    }
    
    public ShardingKey getSuperShardingKey() throws SQLException {
        return null;
    }
});

JdbcTemplate shardingJdbcTemplate = new JdbcTemplate(dataSourceAdapter);

// The SQL query will be executed directly in the shard corresponding 
// to the sharding key returned from the ShardingKeyProvider.getShardingKey() method.
shardingJdbcTemplate.execute(SQL);
```

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/ShardingKeyProvider.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/ShardingKeyProvider.java
@@ -8,19 +8,10 @@ import org.springframework.lang.Nullable;
 /**
  * Interface defines methods for retrieving sharding keys, which are used to establish
  * direct shard connections (in the context of sharded databases). This is used as a
- * way of providing the sharding key, besides the thread-bound sharding keys in
+ * way of providing the sharding key in
  * {@link org.springframework.jdbc.datasource.ShardingKeyDataSourceAdapter}.
  *
- * <p>It is particularly useful in scenarios where using {@link DirectShardCallbackTemplate} is not
- * practical, like when the sharding key is independent of the specific query being executed.
- * For instance, the sharding key may depend on an HTTP session (authenticated user's data).</p>
- *
- * <p>In cases where used in parallel with {@link DirectShardCallbackTemplate}, the sharding key
- * provided in the {@link DirectShardCallbackTemplate} method is the one that will be used, and
- * not the one provided by {@link ShardingKeyProvider}.</p>
- *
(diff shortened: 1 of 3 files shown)


snicoll:
@meedbek I think this PR would be actionable if we decide to merge https://github.com/spring-projects/spring-framework/pull/30988, which we haven't. Am I missing something?

meedbek:
> @meedbek thanks for the PR, but I think it would only be actionable if we decide to merge #30988, which we haven't. Am I missing something?

meedbek:
> @meedbek thanks for the PR, but I think it would only be actionable if we decide to merge #30988, which we haven't. Am I missing something?

meedbek:
> @meedbek thanks for the PR, but I think it would only be actionable if we decide to merge #30988, which we haven't. Am I missing something?

I am taking over [@Anir-ln](https://github.com/Anir-Ln) work on the feature that's why I opened a new PR.
I am focusing on one side of the feature for now, I will look into the rest later.

snicoll:
What do you mean by "taking over"? You mean replacing? If so, then there should be some discussion to close the other PR then.

jeandelavarene:
Hallo @jhoeller, we would like to have a meeting with you to review the comments you made in the older PR https://github.com/spring-projects/spring-framework/pull/30988 regarding transactions. Would that be possible? If so, how to proceed? Thank you.

PR #31513, opened 2023-10-27
Support "--" end of options in SimpleCommandLineArgsParser

### LINKED ISSUE #21949: Support for bare double dash in argument parsing with SimpleCommandLineArgsParser [SPR-17416]
**[Phil Webb](https://jira.spring.io/secure/ViewProfile.jspa?name=pwebb)** opened **[SPR-17416](https://jira.spring.io/browse/SPR-17416?redirect=false)** and commented

Originally raise [as a Spring Boot issue](https://github.com/spring-projects/spring-boot/issues/14909):

I would like Spring Boot application to support guideline 10 of POSIX.1-2017 - Utility argument convention section:

| The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

The desired behavior is to recognize the `--` and leave all the arguments to the right of it alone.

Just for context: in `2.0.x` passing `--` to a SpringBoot application results in an error (sample trace below). The double dash is treated as an invalid option (whose name is zero-length) and there seems to be no way to configure this behavior.

```
java.lang.IllegalArgumentException: Invalid argument syntax: --
	at org.springframework.core.env.SimpleCommandLineArgsParser.parse(SimpleCommandLineArgsParser.java:75)
	at org.springframework.core.env.SimpleCommandLinePropertySource.<init>(SimpleCommandLinePropertySource.java:90)
	at org.springframework.boot.DefaultApplicationArguments$Source.<init>(DefaultApplicationArguments.java:76)
	at org.springframework.boot.DefaultApplicationArguments.<init>(DefaultApplicationArguments.java:42)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:321)
	at com.netflix.genie.GenieAgentApplication.run(GenieAgentApplication.java:63)
```

---

**Reference URL:** https://github.com/spring-projects/spring-boot/issues/14909

bclozel:
Prior to this commit, the `SimpleCommandLineArgsParser` would reject
"--" arguments as invalid. As reported by the community, the POSIX
utility conventions (Guideline 10) state that

> The first -- argument that is not an option-argument should be
> accepted as a delimiter indicating the end of options.
> Any following arguments should be treated as operands, even if they
> begin with the '-' character.

This commit updates `SimpleCommandLineArgsParser` to not reject "--"
arguments and instead to consider remaining arguments as non-optional.

Fixes gh-21949

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/env/SimpleCommandLineArgsParser.java
+++ b/spring-core/src/main/java/org/springframework/core/env/SimpleCommandLineArgsParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,13 @@ package org.springframework.core.env;
  * <em>without spaces</em> by an equals sign ("="). The value may optionally be
  * an empty string.
  *
+ * <p>This parser supports the POSIX "end of options" delimiter, meaning that
+ * any {@code "--"} (empty option name) in the command signals that all remaining
+ * arguments will non-optional. For example, here {@code "--opt=ignored"} is considered
+ * as a non-optional argument.
+ * <pre class="code">
+ * --foo=bar -- --opt=ignored</pre>
+ *
  * <h4>Valid examples of option arguments</h4>
  * <pre class="code">
  * --foo
@@ -53,6 +60,7 @@ package org.springframework.core.env;
  *
  * @author Chris Beams
  * @author Sam Brannen
+ * @author Brian Clozel
  * @since 3.1
  */
 class SimpleCommandLineArgsParser {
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/core/env/SimpleCommandLineArgsParser.java
```diff
@@ -74,13 +81,18 @@ public CommandLineArgs parse(String... args) {
...
 					optionName = optionText.substring(0, indexOfEqualsSign);
 					optionValue = optionText.substring(indexOfEqualsSign + 1);
+					if (optionName.isEmpty()) {
+						throw new IllegalArgumentException("Invalid argument syntax: " + arg);
+					}
 				}
 				else {
 					optionName = optionText;
+					if (optionName.isEmpty()) {
+						// End of options delimiter, all remaining args must be ignored
```
reevik, line 91: It must not, but the implementation can ignore them.

bclozel:
Thanks @reevik , I've updated my PR.

snicoll (approved):

PR #31523, opened 2023-10-30
Add Kotlin extension for `PropertyResolver#getProperty(String, Class<T>, T)`

wakingrufus:
Allows Kotlin caller to get a non-nullable property, using a default value when the property is not set. A method already exists on PropertyResolver which does this, but takes a type parameter.
The extension method can eliminate the parameter via reified type, similar to the other extension methods which exist already.

Remove `@Disabled` from PropertyResolverExtensionsKotlinTests to allow tests to run

### DIFF
--- a/spring-core/src/test/kotlin/org/springframework/core/env/PropertyResolverExtensionsKotlinTests.kt
+++ b/spring-core/src/test/kotlin/org/springframework/core/env/PropertyResolverExtensionsKotlinTests.kt
@@ -19,7 +19,7 @@ package org.springframework.core.env
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.verify
-import org.junit.jupiter.api.Disabled
+import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 
 /**
@@ -27,7 +27,6 @@ import org.junit.jupiter.api.Test
  *
  * @author Sebastien Deleuze
  */
-@Disabled
 class PropertyResolverExtensionsKotlinTests {
 
 	val propertyResolver = mockk<PropertyResolver>()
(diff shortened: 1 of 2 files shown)


wakingrufus:
Hi, @sdeleuze since you were most recent committer to these files, would you mind reviewing this? 
Thanks!

sbrannen:
> Hi, @sdeleuze since you were most recent committer to these files, would you mind reviewing this?

@wakingrufus, please refrain from @-mentioning individual team members. Somebody from the team will process the issue when they get a chance.

Furthermore, I've added the `theme:kotlin` label which Sebastien monitors.

wakingrufus:
> @wakingrufus, please refrain from @-mentioning individual team members. Somebody from the team will process the issue when they get a chance.
> 
> Furthermore, I've added the `theme:kotlin` label which Sebastien monitors.

Ok, Sorry about that. I have had my PRs sit unreviewed until someone is looped in when contributing to other projects in the past. Sorry to bring in that assumption here. Thank you for taking a look at my PR so quickly.

spring-core/src/test/kotlin/org/springframework/core/env/PropertyResolverExtensionsKotlinTests.kt
```diff
@@ -19,15 +19,14 @@ package org.springframework.core.env
...
-import org.junit.jupiter.api.Disabled
+import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 
 /**
  * Mock object based tests for PropertyResolver Kotlin extensions.
  *
  * @author Sebastien Deleuze
  */
-@Disabled
```
vatsasiddhartha, line 30: we can remove the @Disabled annotation..

Issue #30645, opened 2023-06-12
Built-in Web Support for Method Bean Validation


rstoyanchev:
Currently bean validation is applied to `@ModelAttribute` and `@RequestBody` arguments if they are also annotated with `jakarta.validation.Valid` or Spring's `@Validated` with validation groups. You can put constraint annotations directly on other arguments, but that requires `@Validated` on the controller class with `MethodValidationPostProcessor` applying method validation via AOP. 

This is not intuitive and creates challenges for handling validation errors, e.g. to handle either `ConstraintViolationException` or `MethodArgumentNotValidException`. Also need to introspect method parameters to know if the failure is for a request header, path variable, or other, see #26219. It can lead to double validation, see #24913. It does not support reactive types, see #20781. 

We need built-in method validation in web handling for a more integrated experience, default web exception handling of constraint violations to RFC 7807 responses. 

This is an umbrella issue for the improvement:

- [x] #29825
- [x] #30644
- [x] #30813
- [x] #30643
- [x] #30653

unwx:
Is it possible to disable this built-in validation?

I looked into the code and noticed that the [HandlerMethodValidator](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/method/annotation/HandlerMethodValidator.java) is automatically used when `jakarta.validation.Validator` is present and when [@Validated](https://github.com/spring-projects/spring-framework/blob/main/spring-context/src/main/java/org/springframework/validation/annotation/Validated.java) is not used.

In [ControllerMethodResolver](https://github.com/spring-projects/spring-framework/blob/main/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/ControllerMethodResolver.java) and in [RequestMappingHandlerAdapter](https://github.com/spring-projects/spring-framework/blob/main/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerAdapter.java):
```java
private final static boolean BEAN_VALIDATION_PRESENT = ClassUtils.isPresent("jakarta.validation.Validator", HandlerMethod.class.getClassLoader());
```
...
```java
if (BEAN_VALIDATION_PRESENT) {
    List<HandlerMethodArgumentResolver> resolvers = this.argumentResolvers.getResolvers();
    this.methodValidator = HandlerMethodValidator.from(
        this.webBindingInitializer, 
        this.parameterNameDiscoverer,
[...]

Issue #29825, opened 2023-01-16
Spring MVC and WebFlux method validation support that also adapts `ConstraintViolation`s


rstoyanchev:
Introduce support for method validation in Spring MVC and Spring WebFlux. This requires changes first at a lower level to enhance `MethodValidationInterceptor` infrastructure to adapt `ConstraintViolation`s to `MessageSourceResolvable` errors, further aggregated into a `BindingResult` for cascaded violations on Object method parameters.

neroncap:
Hi! I'm interesting in make some custom validations for the entire @RequestHeader, i was testing this issue and is not possible. It is planned to be included in further versions?

rstoyanchev:
Yes it is possible as indicated in the summary above, but the goal is to make it better integrated with Spring MVC and WebFlux, and easier to handle any exceptions. 

If you need help to make it work as it is currently, this is not the right place.

rstoyanchev:
This issue has been re-purposed to match the initial support in 6.1 M1, and the changes for that are now available in `6.1.0-SNAPSHOT`. Below is a brief summary of the changes, ahead of a more detailed documentation update in #30643. 

**spring-context changes**

`MethodValidationInterceptor` now delegates to `MethodValidationAdapter`, comparable to `SpringValidatorAdapter` but for method validation. It produces a `MethodValidationResult` with `ConstraintViolation`s adapted to `ParameterValidationResult` each containing a list of `MessageSourceResolvable` errors, also grouped by method parameter. A `ParameterValidationResult` may also be `ParameterErrors`, an `Errors` instance with cascaded violations adapted to field errors.

`MethodValidationResult` provides a method to throw `MethodValidationException`. `MethodValidationInterceptor` makes use of. This exception is a `ConstraintValidationException` so same as before, but with more information.

**Spring MVC and WebFlux changes**

Spring MVC and WebFlux have been updated to use `MethodValidationAdater` to apply method validation, and raise `MethodValidationException` in case of constraint violations. In order to be used, remove `@Validated` on the class level (i.e. opt out of AOP method validation), ensure `jakarta.validation.Validator` such as `LocalValidatorFactoryBean` is configured for use in Spring MVC or WebFlux config, and then it should apply to controller methods `@Constraint` annotations.

[...]

PR #31530, opened 2023-11-01
Improve method validation support for errors on elements within a container

k-seth:
Expands on the validation behaviour added as part of the work for https://github.com/spring-projects/spring-framework/issues/30645, motivated by an issue with Sets containing an object with nested Lists, introduced in https://github.com/spring-projects/spring-framework/issues/29825. The adapter should now better handle nested objects for types beyond those of List or Map.


rstoyanchev:
For additional context, see https://github.com/spring-projects/spring-framework/issues/30645#issuecomment-1783969150. In short, currently we support `List` and `Map` containers. The goal is to also support arrays and sets, and other container types for which there is a registered Jakarta `ValueExtractor`.

rstoyanchev:
Thanks for the report and changes. Element container handling is not only improved but also simpler.

PR #31532, opened 2023-11-01
Introduce CompilationException::getDiagnosticCode

quaff:
Fix TestCompilerTests with locales other than english

See GH-31407


snicoll:
Thanks for the PR but I don't think this is is the right fix for this. The status code can be also implementation dependent and you can have more than one error per compilation round. Before changing the API again, I want to see how we can provide more information in the exception, if any.

PR #31537, opened 2023-11-02
Improved ObjectInputStream of DefaultDeserializer automatically return resources by applying try-with-resources syntax.

ch4570:
In the DefaultDeserializer class, there was a problem with not properly releasing resources used by the ObjectInputStream. 

I've improved it by using try-with-resources to ensure that resources are properly released, 
thereby preventing memory leaks.


snicoll:
@ch4570 thanks for the PR but the Javadoc of `Deserializer#deserialize` states:

> Note: Implementations should not close the given InputStream (or any decorators of that InputStream) but rather leave this up to the caller.

PR #31531, opened 2023-11-01
Use switch expression where feasible

quaff:

PR #31538, opened 2023-11-02
Use switch expression instead of If-Else

only-juun:
**[Use switch expression instead of If-Else]**

I refactored the code using the new switch expression introduced since Java 12, which I used to make the code more concise and readable. 
This code uses Java's new switch expression to check each ResourceResolver subtype and set the corresponding boolean flag. The case L -> syntax allows us to concisely write the executable code for each case in a single line. Note that ignored is the variable used in the pattern matching, which is not actually used in this case.
The variable name ignored is used to indicate that this variable is not actually used, meaning that we are only checking to see if it is of type VersionResourceResolver, and not interested in any particular instance of that type.

Refactoring the code in this way improves readability and allows us to omit the break statement that would have been required in a traditional switch statement.

**[Changes to avoid unnecessary null checks]**

"== null" check is a common and intuitive way to ensure that an object is not initialized in Java. However, I considered a different approach to improve the readability and maintainability of the code. 
In this code, the pathPrefixes field is initialized to a LinkedHashMap upon declaration, eliminating the need for a null check.

**[Use switch expression where feasible]**

In Java 12 and later versions, I use the new switch expression to make my code more concise.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/config/PathMatchConfigurer.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/config/PathMatchConfigurer.java
@@ -35,12 +35,10 @@ public class PathMatchConfigurer {
 	@Nullable
 	private Boolean trailingSlashMatch;
 
-
 	@Nullable
 	private Boolean caseSensitiveMatch;
 
-	@Nullable
-	private Map<String, Predicate<Class<?>>> pathPrefixes;
+	private Map<String, Predicate<Class<?>>> pathPrefixes = new LinkedHashMap<>();;
 
 
 	/**
@@ -79,9 +77,6 @@ public class PathMatchConfigurer {
 	 * @since 5.1
 	 */
 	public PathMatchConfigurer addPathPrefix(String prefix, Predicate<Class<?>> predicate) {
-		if (this.pathPrefixes == null) {
-			this.pathPrefixes = new LinkedHashMap<>();
-		}
 		this.pathPrefixes.put(prefix, predicate);
 		return this;
 	}
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/config/ResourceChainRegistration.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/config/ResourceChainRegistration.java
@@ -81,15 +81,13 @@ public class ResourceChainRegistration {
 	public ResourceChainRegistration addResolver(ResourceResolver resolver) {
 		Assert.notNull(resolver, "The provided ResourceResolver should not be null");
 		this.resolvers.add(resolver);
-		if (resolver instanceof VersionResourceResolver) {
-			this.hasVersionResolver = true;
-		}
-		else if (resolver instanceof PathResourceResolver) {
-			this.hasPathResolver = true;
-		}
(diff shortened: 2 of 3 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/config/PathMatchConfigurer.java
```diff
@@ -35,12 +35,10 @@ public class PathMatchConfigurer {
 	@Nullable
 	private Boolean trailingSlashMatch;
 
-
 	@Nullable
 	private Boolean caseSensitiveMatch;
 
-	@Nullable
-	private Map<String, Predicate<Class<?>>> pathPrefixes;
+	private Map<String, Predicate<Class<?>>> pathPrefixes = new LinkedHashMap<>();;
```
sbrannen, line 41: Please revert all changes related to `pathPrefixes`.

The code was intentionally written without assignment, since path prefixes are optional.

The `LinkedHashMap` is lazily created on demand. Consequently, the map is only created if it's actually used.

spring-webflux/src/main/java/org/springframework/web/reactive/config/ResourceChainRegistration.java
```diff
@@ -81,15 +81,13 @@ public ResourceChainRegistration(boolean cacheResources, @Nullable Cache cache)
...
-		else if (resolver instanceof WebJarsResourceResolver) {
-			this.hasWebjarsResolver = true;
-		}
+
+        switch (resolver) {
+            case VersionResourceResolver ignored -> this.hasVersionResolver = true;
+            case PathResourceResolver ignored -> this.hasPathResolver = true;
+            case WebJarsResourceResolver ignored -> this.hasWebjarsResolver = true;
+        }
+
```
sbrannen, line 90: These changes break the build.

Please revert these changes.

[Pattern Matching for switch](https://openjdk.org/jeps/441) did not become an "enabled" feature until Java 21, and we have a Java 17 baseline.

More importantly, please ensure that you actually run the build before submitting a PR.

spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientResponseException.java
```diff
@@ -302,44 +302,30 @@ public static WebClientResponseException create(
...
-				case SERVICE_UNAVAILABLE:
-					return new WebClientResponseException.ServiceUnavailable(statusText, headers, body, charset, request);
-				case GATEWAY_TIMEOUT:
-					return new WebClientResponseException.GatewayTimeout(statusText, headers, body, charset, request);
-			}
-		}
-		return new WebClientResponseException(statusCode, statusText, headers, body, charset, request);
-	}
+		return switch (statusCode) {
+			case HttpStatus httpStatus -> {
```
sbrannen, line 306: As stated above:

[Pattern Matching for switch](https://openjdk.org/jeps/441) did not become an "enabled" feature until Java 21, and we have a Java 17 baseline.


Thus, this code does not compile.

In addition, the switch expression refactoring for this class is already covered by #31531.

sbrannen:
Hi @only-juun,

Unfortunately, we cannot accept the proposed changes (see my review comments for details), but we're looking forward to your next contribution.

Cheers,

Sam

PR #31546, opened 2023-11-03
Change an existing switch statement to the newer syntax

ch4570:
This PR updates the switch statements from Java 11 and earlier, applying modern syntax to eliminate redundant break statements and enhance code readability by making it block-oriented.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/serializer/DefaultDeserializer.java
+++ b/spring-core/src/main/java/org/springframework/core/serializer/DefaultDeserializer.java
@@ -66,11 +66,9 @@ public class DefaultDeserializer implements Deserializer<Object> {
 	 */
 	@Override
 	public Object deserialize(InputStream inputStream) throws IOException {
-		ObjectInputStream objectInputStream = new ConfigurableObjectInputStream(inputStream, this.classLoader);
-		try {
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/core/serializer/DefaultDeserializer.java
```diff
@@ -66,11 +66,9 @@ public DefaultDeserializer(@Nullable ClassLoader classLoader) {
 	 */
 	@Override
 	public Object deserialize(InputStream inputStream) throws IOException {
-		ObjectInputStream objectInputStream = new ConfigurableObjectInputStream(inputStream, this.classLoader);
-		try {
+		try (ObjectInputStream objectInputStream = new ConfigurableObjectInputStream(inputStream, this.classLoader)) {
```
snicoll, line 69: This is an unrelated  change from a previous PR. Please revert.

snicoll (changes requested):

bclozel:
This code belongs to our ASM fork, I'm not sure we should apply cosmetic changes there as it's likely to add conflicts when we upgrade.

snicoll:
Ah I had missed that. Yeah, in that case, I am in favor of closing this.

PR #31554, opened 2023-11-06
Add duplicate key exception error code for SAP HANA database

baratrax:


jhoeller:
Since this matches the Hana error code in `sql-error-codes.xml`, let's bring this in sync for 6.0.14 - and also update the duplicated check in the R2DBC `ConnectionFactoryUtils` class accordingly.

PR #31576, opened 2023-11-08
Update BeanInfoFactory.java

NWOM:
Method Name: Changed the method name from getBeanInfo to createBeanInfo to make its purpose more self-explanatory.

Generics: Introduced a generic type parameter <T> to specify the type of BeanInfo returned by the factory. This makes the interface more flexible and type-safe.

Exception Handling: Introduced a custom exception BeanInfoCreationException for better error handling. This exception provides constructors to allow for more detailed error messages and the ability to attach a cause to the exception.

Improved JavaDoc Comments: Enhanced the JavaDoc comments for the interface and its methods to provide more information about their purpose and usage.

Annotations: Retained the @Nullable annotation to indicate that the returned BeanInfo can be null.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanInfoFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanInfoFactory.java
@@ -22,39 +22,39 @@ import java.beans.IntrospectionException;
 import org.springframework.lang.Nullable;
 
 /**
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the suggestion, but we can't modify a public interface like that as it would be a breaking change. The introduction of the generics is also problematic.

Issue #28055, opened 2022-02-15
Introduce landing page for JavaDoc and KDoc


sbrannen:
## Overview

Spring Framework's Kotlin API documentation is not discoverable via https://spring.io/projects/spring-framework#learn.

The `API Doc.` link points to the [JavaDoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/), and there is no way to navigate to the KDoc unless you follow one of the links buried in the reference manual.

## Deliverables

- [ ] Introduce a new HTML "landing page" for the API docs that provides links to the [JavaDoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/) and [KDocs](https://docs.spring.io/spring-framework/docs/current/kdoc-api/).
   - For example, this could be a new `api/index.html` file so that `/api` would be the link for the landing page.
- [ ] Update spring.io so that the `API Doc.` links point to the new landing page.
- [ ] Ensure that the landing page is included in the ZIP distribution that gets unzipped on the spring.io server.

simonbasle:
Superseded by gh-31587

PR #31587, opened 2023-11-09
Link to KDoc API documentation from Javadoc overview

simonbasle:
This commit also fixes the javadoc task to take the overview.html asset into account, since this is where the link to KDoc is set up.

Note that the output directory for kdoc is changed to be consistent with what's in the docsZip, so that the link works in local builds too.

See gh-28055


sbrannen (approved):
I haven't verified that actual result in the build, so I'm trusting that all works.

Otherwise, I think this looks good.

Thanks! 👍

bclozel (approved):
Thanks for taking care of this!

PR #31598, opened 2023-11-13
Avoid duplicate JAR resources in `PathMatchingResourcePatternResolver` on MS Windows

huyachigege:
To avoid duplicate JAR resources on MS Windows.

If you use `mybatis-plus` and your `mapper-locations` starts with `classpath*:`, Spring will find duplicate `UrlResource`s for XML files which results in a crash on application startup.

Since their `getCleanedUrl` values are different mybatis will throw an exception, even though they represent the same resource.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java
+++ b/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java
@@ -430,7 +430,8 @@ public class PathMatchingResourcePatternResolver implements ResourcePatternResol
 					int prefixIndex = filePath.indexOf(':');
 					if (prefixIndex == 1) {
 						// Possibly "c:" drive prefix on Windows, to be upper-cased for proper duplicate detection
-						filePath = StringUtils.capitalize(filePath);
+						// to resolve find duplicate jar resource on windows
+						filePath = "/" + StringUtils.capitalize(filePath);
 					}
 					// # can appear in directories/filenames, java.net.URL should not treat it as a fragment
 					filePath = StringUtils.replace(filePath, "#", "%23");


huyachigege:
jar:file:F:/****

huyachigege:
ClassLoader.getResources("") will generate path like 'jar:file:/F:***' and addClassPathManifestEntries() will generate path like ''jar:file:F:***'. They are the same actually ,but UrlResource.getCleanedUrl() is different ,so you will add duplicate UrlResource into result of method PathMatchingResourcePatternResolver.getResources(location).
My English is not good, I hope you can understand.

sbrannen:
Hi @huyachigege,

Congratulations on creating your first PR ever! 👍 

We'll look into it.

> My English is not good, I hope you can understand.

Yes, your explanation is good enough to follow.

I edited the text to make it slightly clearer, and hopefully I didn't change the meaning you wished to convey.

huyachigege:
> Hi @huyachigege,
> 
> Congratulations on creating your first PR ever! 👍
> 
> We'll look into it.
> 
> > My English is not good, I hope you can understand.
> 
> Yes, your explanation is good enough to follow.
> 
> I edited the text to make it slightly clearer, and hopefully I didn't change the meaning you wished to convey.

Thanks, that is what I mean.

jhoeller:
`addClassPathManifestEntries` delegates to `hasDuplicate` before adding a new entry, and that method is meant to detect exactly such a difference in leading slashes. Could you try to find out why this is not correctly identifying the duplicate in your scenario?

huyachigege:
> `addClassPathManifestEntries` delegates to `hasDuplicate` before adding a new entry, and that method is meant to detect exactly such a difference in leading slashes. Could you try to find out why this is not correctly identifying the duplicate in your scenario?

line 497 will get root resources, it will containes 'jar:file:/F:xxx/app.jar!/boot-info/classes' and 'jar:file:F:xxx/app.jar!' ,they are diffrent now. But child of them will be 'jar:file:/F:xxx/app.jar!/.../xxx.xml' and 'jar:file:F:xxx/app.jar!/.../xxx.xml' ，they are duplicate on windows. If on linux, they are both 'jar:file:/...' so it works on linux.
You can try on windows and use properies l just told.

huyachigege:
> Hi @huyachigege,
> 
> > line 497 will get root resources,
> 
> That doesn't map to the current code on the `main` branch.
> 
> With which version of Spring Framework are you experiencing the duplicates?

5.x

> Hi @huyachigege,
> 
> > line 497 will get root resources,
> 
> That doesn't map to the current code on the `main` branch.
> 
> With which version of Spring Framework are you experiencing the duplicates?

5.3.x,sorry for that

sbrannen:
#### Related Issues

- #20539

sbrannen:
> If you use `mybatis-plus` and your `mapper-locations` starts with `classpath*:`

Can you please provide us the exact location pattern you are using?

Specifically, what do you have for `<???>` in `classpath*:<???>`?

huyachigege:
> > If you use `mybatis-plus` and your `mapper-locations` starts with `classpath*:`
> 
> Can you please provide us the exact location pattern you are using?
> 
> Specifically, what do you have for `<???>` in `classpath*:<???>`?

classpath*:/**/*_MYSQL.xml

sbrannen:
> classpath*:/**/*_MYSQL.xml

Thanks. That helps.

Can you please tell us which version of Java and Spring Boot you're using as well?

huyachigege:
> > classpath*:/**/*_MYSQL.xml
> 
> Thanks. That helps.
> 
> Can you please tell us which version of Java and Spring Boot you're using as well?

OpenJDK-11,SpringBoot 2.6.12.

sbrannen:
> line 497 will get root resources, it will contain `jar:file:/F:xxx/app.jar!/boot-info/classes` and `jar:file:F:xxx/app.jar!`. They are different now. But child of them will be `jar:file:/F:xxx/app.jar!/.../xxx.xml` and `jar:file:F:xxx/app.jar!/.../xxx.xml`. They are duplicate on windows. If on Linux, they are both `jar:file:/...`, so it works on Linux.

Based on that feedback, it appears that we are getting two different roots (`rootDirUrl` values) for the same underlying JAR:

- `jar:file:F:xxx/app.jar!` -- for the actual Spring Boot application JAR
- `jar:file:/F:xxx/app.jar!/boot-info/classes` -- from the MANIFEST of the Spring Boot application JAR

Then, in `findPathMatchingResources()` we add the results of `doFindPathMatchingJarResources(rootDirResource, rootDirUrl, subPattern)` without checking for duplicates.

The `hasDuplicate()` method Juergen mentioned is used at a lower level and therefore does not detect duplicates at this higher level.

sbrannen:
This has been merged into `main` in d5874ab99ef62b3da8a9269aa526b43328e968a9, revised in e71117dcdf8c713b0a364d896ea2221183eac64b, and backported to `5.3.x`.

Thanks

Issue #31607, opened 2023-11-14
When using Oracle, `JdbcClient.update(KeyHolder)` does not work without explicit key column names


zakaria-shahen:
Hello
When use Oracle DB The `JdbcClient.update(KeyHolder)` method not work with named parameters 

Example:
```java
        jdbcClient.sql(insert_sql)
                .paramSource(objectModel)
                .update(keyHolder);
        var id = keyHolder.getKey();
```
and throw
```
org.springframework.dao.DataRetrievalFailureException: The generated key type is not supported. Unable to cast [oracle.sql.ROWID] to [java.lang.Number].
	at org.springframework.jdbc.support.GeneratedKeyHolder.getKeyAs(GeneratedKeyHolder.java:86) ~[spring-jdbc-6.1.0-RC2.jar:6.1.0-RC2]
	at org.springframework.jdbc.support.GeneratedKeyHolder.getKey(GeneratedKeyHolder.java:65) ~[spring-jdbc-6.1.0-RC2.jar:6.1.0-RC2]
```

and when use `NamedParameterJdbcOperations` as workaround, it works 
```java
        namedParameterJdbcOperations.update(
                insert_sql, new SimplePropertySqlParameterSource(objectModel), keyHolder, new String[] {"id"}
        );
```

jhoeller:
I ended up introducing an `update(KeyHolder generatedKeyHolder, String... keyColumnNames)` variant which supports both indexed and named parameters (even if it is more likely to be used with named parameters).

PR #31608, opened 2023-11-14
Updated `JdbcClient.update` to enable returning generated keys when using named parameters and Oracle DB. 

zakaria-shahen:
Updated `JdbcClient.update` to enable returning generated keys when using named parameters and Oracle DB. 
issue: https://github.com/spring-projects/spring-framework/issues/31607


jhoeller:
I ended up introducing an `update(KeyHolder generatedKeyHolder, String... keyColumnNames)` variant which supports both indexed and named parameters (even if it is more likely to be used with named parameters) which I'm about to commit against #31607, just in time for the 6.1 GA release tomorrow.

PR #31636, opened 2023-11-21
Add BearerAuthenticationInterceptor

sadv1r:
Added ClientHttpRequestInterceptor implementation for RestClient, WebClient or RestTemplate requests with externally provided Bearer token like with BasicAuthenticationInterceptor:

```java
@Configuration
public class RestClientConfiguration {

    @Value("${example.token}")
    private String token;

    @Bean
    public RestClient restClient() {
        return RestClient.builder()
                .baseUrl("https://api.example.com")
                .requestInterceptor(new BearerAuthenticationInterceptor(() -> token))
                .build();
    }

}
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/support/BearerAuthenticationInterceptor.java
+++ b/spring-web/src/main/java/org/springframework/http/client/support/BearerAuthenticationInterceptor.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2002-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


spring-web/src/main/java/org/springframework/http/client/support/BearerAuthenticationInterceptor.java
```diff
@@ -0,0 +1,61 @@
...
+ * token, unless a custom {@code Authorization} header has already been set.
+ *
+ * @author Dmitry Ivanov
+ * @see HttpHeaders#setBearerAuth
+ * @see HttpHeaders#AUTHORIZATION
+ * @since 6.1
+ */
+public class BearerAuthenticationInterceptor implements ClientHttpRequestInterceptor {
+
+	private final String token;
```
quaff, line 38: It's better to use `Supplier<String>` to support dynamic token.

sadv1r: Great idea, thanks!

bclozel:
Thanks for the proposal but we think that bearer token support is better suited in Spring Security with the OAuth2 support. If you're using another authentication mechanism that uses bearer token, creating your own interceptor is simple enough not to rely on a Framework provided one.

I'm closing this issue as a result.

See https://docs.spring.io/spring-security/reference/servlet/oauth2/index.html#oauth2-client

PR #31638, opened 2023-11-21
Refresh cache in `ReloadableResourceBundleMessageSource`

vershnik:
This change allows subclasses to call getMergedProperties(locale) when
cacheSeconds > 0 and get refreshed contents. Useful for SPA webapps to
deliver all translations to the browser.

Since all individual files can be refreshed concurrently, the thread that
refreshes mergedHolder can get stale data for any file being refreshed
so simply comparing maximal fileTimestamp is not enough to detect
change - used hashCode for this purpose.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/ReloadableResourceBundleMessageSource.java
+++ b/spring-context/src/main/java/org/springframework/context/support/ReloadableResourceBundleMessageSource.java
@@ -30,6 +30,7 @@ import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 
 import org.springframework.context.ResourceLoaderAware;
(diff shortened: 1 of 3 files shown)


snicoll:
@vershnik this PR has a a commit from @andreibastun. The author should be the one submitting the PR as they're the one signing the CLA.

andreibastun:
@snicoll  Sorry, I didn't notice that my second account was singed in the git when commiting. Now I accepthed CLA from both accounts.

snicoll:
Thanks for the PR but we're not keen to make `ReloadableResourceBundleMessageSource` more involved than it already is. We suggest that you implement the refresh in a custom sub-class of yours and if that turns out to be too convoluted, we can discuss how to add a hookpoint to make it easier.

andreibastun:
@snicoll Ok, that makes sense, thanks.  I have created another PR  [32118](https://github.com/spring-projects/spring-framework/pull/32118)  that does not add functionality, but just extracs some methods from getMergedProperties to be more easily reused in subclasses.

PR #31639, opened 2023-11-21
Optimization: convert internal classes to records

freesme:


snicoll:
@freesme thanks for the PR but we prefer to apply stylistic changes like that ourselves.

freesme:
> @freesme thanks for the PR but we prefer to apply stylistic changes like that ourselves.

Got it, thanks for the feedback.

PR #31647, opened 2023-11-22
Treat kotlin.Unit as void

### LINKED ISSUE #31648: Spring MVC: Unexpected bytes added to the HTTP response for kotlin.Unit return type in controller methods
Affects: Spring Boot 3.2.0-RC2, Spring Web 3.2.0-RC2, Kotlin 1.9.20

### Issue
Using a Controller, with return type `kotlin.Unit`, which uses directly `HttpServletResponse` to write the response, adds unexpected bytes to the response. The following is a controller example where a file is read and written to the output directly:

```
    @GetMapping(path = ["public/file.txt"])
    fun getFile(
        httpServletResponse: HttpServletResponse
    ) {
        httpServletResponse.status = 200

        httpServletResponse.outputStream.write(Files.readAllBytes(Path.of("src/main/resources/file.txt")))
    }
```
The following shows the bytes expected vs current:

```
Current: [49, 50, 51, 52, 53, 54, 55, 56, 57, 123, 125]
Expected: [49, 50, 51, 52, 53, 54, 55, 56, 57]
```

123 and 125 bytes are `{` and `}`

### How to reproduce the issue

This is a [Spring Boot project](https://github.com/spring-projects/spring-boot/files/13429547/issuebytes.zip) with a test that reproduces the issue.

### Debugging

My guess would be: there is a new converter which is taking `kotlin.Unit` and tries to serialize it, outputting `{` and `}`.

quaff:
Fix https://github.com/spring-projects/spring-framework/issues/31648

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/KotlinDetector.java
+++ b/spring-core/src/main/java/org/springframework/core/KotlinDetector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


sdeleuze:
Thanks for your PR, but I fixed the related issue via 441e2105338835c1d43dc176d6527c791febc8d9.

PR #31650, opened 2023-11-22
Optimize loops in BeanDefinitionPropertiesCodeGenerator

freesme:
-  There are two iterations of `propertyValues` in the code, combining the logic into one loop can reduce overhead.
- `propertyValue.getName()` already cached but not used.


bclozel:
Thanks for your contribution @freesme !

PR #31659, opened 2023-11-23
RestClient documentation improvements

Johannes-Rost:
Confusion between request/response and builder/built response in the documentation auf RestClient corrected.


snicoll:
@Johannes-Rost thanks again!

PR #31663, opened 2023-11-24
Polish GenericTypeResolver Javadoc

quaff:


snicoll:
Good stuff, thanks again @quaff

PR #31679, opened 2023-11-24
Clarify that DI is a type of IoC

Nijco:
The documentation currently states that Inversion of Control (IoC) and Dependency Injection (DI) are the same thing. Although the two terms are related, they are not synonymous, but DI is a type of IoC.

I believe this change is important because using the terms interchangeably really muddies the water about the meaning of these two fundamental concepts of software frameworks.


sbrannen:
Hi @Nijco,

This has been merged into `main` in a56bf874768cac4dd6643a3525083c63f0035a60 and revised in 23dc5696d9acd89922c66a2c00f857f6b1e2fa41.

Thanks

PR #31686, opened 2023-11-26
add InitDestroyAnnotationBeanPostProcessor to BeanFactoryInitializationAotProcessor

### LINKED ISSUE #30755: Init/destroy methods are invoked in different order in AOT mode
## Overview

This issue serves as a placeholder until I find more time to provide details.

In the interim, the difference in invocation order between standard JVM mode and AOT mode can be viewed in the following tests.

## Tests

JVM mode:

https://github.com/spring-projects/spring-framework/blob/2ccf4cab8b4b437915c297be979d2e569fc273ef/spring-context/src/test/java/org/springframework/context/annotation/InitDestroyMethodLifecycleTests.java#L123-L143

AOT mode:

https://github.com/spring-projects/spring-framework/blob/2ccf4cab8b4b437915c297be979d2e569fc273ef/spring-context/src/test/java/org/springframework/context/annotation/InitDestroyMethodLifecycleTests.java#L170-L203

## Known Issues

- `InitDestroyAnnotationBeanPostProcessor.processAheadOfTime(RegisteredBean)` gets invoked twice in the same AOT processing phase: once via `CommonAnnotationBeanPostProcessor` and once via `InitDestroyAnnotationBeanPostProcessor`.
- `InitializingBean` and `DisposableBean` are treated specially even in AOT mode.

bdshadow:
closes gh-30755

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InitDestroyAnnotationBeanPostProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InitDestroyAnnotationBeanPostProcessor.java
@@ -31,25 +31,34 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Stream;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.aot.generate.GeneratedMethod;
+import org.springframework.aot.generate.GenerationContext;
+import org.springframework.aot.hint.ExecutableMode;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationAotContribution;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationAotProcessor;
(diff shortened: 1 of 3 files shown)


bdshadow:
I understand that `InitDestroyAnnotationBeanPostProcessorTests` must be improved further (and i'm ready to do it), but anyway made a mr to discuss if the whole idea of the fix is correct

As Sam wrote here https://github.com/spring-projects/spring-framework/pull/30724#issuecomment-1602534343, the problem was indeed in merging  init/destroy method names during AOT processing and then there're treated equally in runtime, however it's not true in non-aot mode. Because `@PostConstruct` annotations are considered externallyManaged and are processed differently. So i think the correct solution is to register `CommonAnnotationBeanPostProcessor` (or other  `InitDestroyAnnotationBeanPostProcessor`-s) when they're used.

snicoll:
Thanks for the PR. Unfortunately, I don't think we should register the post-processor. It's a bit backwards as the AOT processing is meant to replace the post-processor by some optimized code.

Removing the method names from the generated code is also not something we'd like to pursue. Ideally the full list of methods to invoke would be in generated code, and the distinction with externally managed would not exist. I haven't looked at the problem enough to know if that's possible without changing something in the core container. 

With that said, do you want to continue working on this or should we close this?

snicoll:
Alright I am going to close this now as this approach is not what we'd consider to fix the related issue. Thanks for the PR, in any case!

PR #31692, opened 2023-11-27
Refactored code 

Nashyn:
I refactored code in 6 files with 6 different methods at some places to make the code more readable and reduce complexity and removed some smells identified.


bclozel:
Sorry but I'm closing this PR as there are way too many issues with it, for changes that don't fix actual problems but are mainly cosmetic.

Nashyn:
Can you tell me where did you identify the issues so i can try to fix it? If possible it would be helpful. Thanks

PR #31696, opened 2023-11-27
Refactoring done

Nashyn:
I have refactored the code again with same method and solved the issues previously arising


snicoll:
> and solved the issues previously arising

Sorry, what do you mean?

Nashyn:
I had raised a pull request earlier but it was closed because it was causing issues so i solved it and now i have created a new pull request

snicoll:
I am afraid this is in the same state as previously discussed in #31692. If you want to contribute, then please avoid making stylistic change and focus on one contribution at a time.

PR #31723, opened 2023-11-30
Link to `spring.factories` used in the TestContext framework in the reference manual

perlun:
This type of change would be useful (since it makes it easier for the reader to find this file), but... will it be hard to support with different versions of the docs? Given that it should technically refer to the "same version" of the file (replacing `main` in the URL with the tag in question), to be really great.

### DIFF
--- a/framework-docs/modules/ROOT/pages/testing/testcontext-framework/tel-config.adoc
+++ b/framework-docs/modules/ROOT/pages/testing/testcontext-framework/tel-config.adoc
@@ -82,8 +82,9 @@ issue is addressed through support for automatic discovery of default
 
 Specifically, the `spring-test` module declares all core default `TestExecutionListener`
 implementations under the `org.springframework.test.context.TestExecutionListener` key in
-its `META-INF/spring.factories` properties file. Third-party frameworks and developers
-can contribute their own `TestExecutionListener` implementations to the list of default
+{spring-framework-code}/spring-test/src/main/resources/META-INF/spring.factories[its
+`META-INF/spring.factories` properties file]. Third-party frameworks and developers can
+contribute their own `TestExecutionListener` implementations to the list of default
 listeners in the same manner through their own `META-INF/spring.factories` properties
 file.
 
@@ -207,4 +208,3 @@ Kotlin::
 	}
 ----
 ======
-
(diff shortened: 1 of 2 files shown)


framework-docs/modules/ROOT/pages/testing/testcontext-framework/tel-config.adoc
```diff
@@ -82,10 +82,10 @@ issue is addressed through support for automatic discovery of default
 
 Specifically, the `spring-test` module declares all core default `TestExecutionListener`
 implementations under the `org.springframework.test.context.TestExecutionListener` key in
-its `META-INF/spring.factories` properties file. Third-party frameworks and developers
-can contribute their own `TestExecutionListener` implementations to the list of default
-listeners in the same manner through their own `META-INF/spring.factories` properties
-file.
+[its `META-INF/spring.factories` properties file](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/main/resources/META-INF/spring.factories).
```
snicoll, line 85: This has a `spring-framework-code` attribute that should be used rather than the hardcoded link. See `antora.yml`.

snicoll (changes requested):
Thanks for the PR. I've added a comment for your consideration.

sbrannen (changes requested):
Hi @perlun,

In addition to the `spring-framework-code` change that @snicoll requested, would you be willing to make a similar change to the section on [`ContextCustomizerFactory`](https://docs.spring.io/spring-framework/reference/testing/testcontext-framework/ctx-management/context-customizers.html#testcontext-context-customizers-automatic-discovery) configuration?

If not, we can handle that ourselves.

In any case, please let us know so that we can proceed with this PR.

framework-docs/modules/ROOT/pages/testing/testcontext-framework/tel-config.adoc
perlun: Thanks for your feedback @snicoll, makes sense. :+1: Nice to learn about some of the Asciidoc features, seems like a nice usage of it in this case.

I adjusted both this and incorporated the suggestion from @sbrannen, please re-check. (Sam - it did perhaps not make as much sense in the https://docs.spring.io/spring-framework/reference/testing/testcontext-framework/ctx-management/context-customizers.html#testcontext-context-customizers-automatic-discovery case, since it referred to both "Spring Framework and Spring Boot", but the link only refers to the Spring Framework one.. but, possibly still a step in the right direction. :thinking:)

sbrannen: Yes, indeed. That's a step in the right direction. I'll rework the wording locally after merging the PR.

sbrannen (approved):

PR #31737, opened 2023-12-01
Introduce `toString(Charset)` in `FastByteArrayOutputStream`

kilink:
Add a toString overload that accepts a Charset to mirror the method that has been on ByteArrayOutputStream since JDK10; additionally, add a special case for when a single buffer is in use internally to avoid the need to resize.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
+++ b/spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
@@ -19,6 +19,7 @@ package org.springframework.util;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.charset.Charset;
 import java.security.MessageDigest;
 import java.util.ArrayDeque;
 import java.util.Deque;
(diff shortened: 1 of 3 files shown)


kilink:
Note this is somewhat of a follow up to #31731 and #30709; `ContentCachingRequestWrapper` was updated to use `FastByteArrayOutputStream`, which didn't have the `toString(Charset)` overload that we were previously calling when it was just `ByteArrayOutputStream`; adding a similar method to `FastByteArrayOutputStream` allows for a nice optimization when there's only a single buffer, as we can pass the byte array + offset + len to the String constructor and avoid any copying.

spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
```diff
@@ -162,9 +163,26 @@ public void close() {
...
 
+	/**
+	 * Converts the buffer's contents into a string by decoding the bytes using
+	 * the specified {@link java.nio.charset.Charset charset}.
+	 *
+	 * @param      charset  the {@linkplain java.nio.charset.Charset charset}
+	 *             to be used to decode the {@code bytes}
+	 * @return     a String decoded from the buffer's contents
+	 */
+	public String toString(Charset charset) {
```
sbrannen, line 177: For future reference, please keep in mind that we do not indent Javadoc tag content and that we need to add a `@since` tag for new elements in the public API.

However, I'll address those issues when merging, so there's no need to update this PR.

sbrannen: In addition, references to instance fields must be prefixed with `this.`.

See Spring Framework's [Code Style](https://github.com/spring-projects/spring-framework/wiki/Code-Style) documentation for details.

PR #31738, opened 2023-12-02
Polish Javadoc

iaruru:
- remove final keyword in final class.

  I noticed that `WriteResultSubscription` is declared as a final class, so in my opinion, I removed the final keyword from methods, as they are already within a final class

- fix minor typos in java comments(javadoc)

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/WriteResultPublisher.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/WriteResultPublisher.java
@@ -117,7 +117,7 @@ class WriteResultPublisher implements Publisher<Void> {
 		}
 
 		@Override
-		public final void request(long n) {
+		public void request(long n) {
 			if (rsWriteResultLogger.isTraceEnabled()) {
 				rsWriteResultLogger.trace(this.publisher.logPrefix +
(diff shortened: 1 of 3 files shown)


spring-web/src/main/java/org/springframework/http/server/reactive/WriteResultPublisher.java
```diff
@@ -117,7 +117,7 @@ public WriteResultSubscription(WriteResultPublisher publisher) {
 		}
 
 		@Override
-		public final void request(long n) {
+		public void request(long n) {
```
sbrannen, line 120: Although `final` is technically unnecessary in the current arrangement, we prefer to make stylistic changes like this on our own when we deem it necessary.

So I'll revert those changes when merging.

sbrannen (approved):
Hi @dogglezz,

As mentioned in the comments, I'm refocusing this PR on the Javadoc fixes and have renamed the PR accordingly.

sbrannen:
This has been merged into `main` in decb22a93de643bbe07f245c6374f7f8f9ae9ee8 and revised in 47cdc7c5f0bf686c92fbea874834ea700a64953a.

Thanks

PR #31754, opened 2023-12-05
Remove redundant methods which are identical to its super method

quaff:


snicoll:
Thanks for the suggestion but we don't feel those to be redundant.

PR #31757, opened 2023-12-05
Introduce overloaded `MockPart` constructor that accepts the `Content-Type`

HyeongMokJeong:
This PR introduces a constructor to MockPart, enabling the configuration of Content-Type. 

When testing controllers that expect a specific format, such as JSON, as @RequestPart, this resolves the issue of encountering a 415 status code problem.

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockPart.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockPart.java
@@ -70,6 +70,19 @@ public class MockPart implements Part {
 		this.headers.setContentDispositionFormData(name, filename);
 	}
 
+	/**
+	 * Constructor for a part with a filename, byte[] content and MediaType mediaType.
+	 * @see #getHeaders()
+	 */
+	public MockPart(String name, @Nullable String filename, @Nullable byte[] content, @Nullable MediaType) {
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @HyeongMokJeong,

Out of curiosity, why do you think the content type warrants a dedicated constructor for `MockPart`?

Have you considered setting the content type after instantiating the `MockPart` as follows?

```java
MockPart mockPart = new MockPart("myPart", """
		{"foo": "bar"}
		""".getBytes());

mockPart.getHeaders().setContentType(MediaType.APPLICATION_JSON);
```

HyeongMokJeong:
Hi @sbrannen 

Thank you for revising the title and providing feedback.

I may not have advanced skills, but I think it would be convenient if `HttpHeaders` could be configured more flexibly.
In that context, Can I hear your opinion on initializing `HttpHeaders` through the constructor?

Like this
```
    public MockPart(String name, @Nullable String filename, @Nullable byte[] content, @Nullable HttpHeaders headers) {
        Assert.hasLength(name, "'name' must not be empty");
        this.name = name;
        this.filename = filename;
        this.content = (content != null ? content : new byte[0]);
        this.headers = headers;
    }
```

Issue #31254, opened 2023-09-18
HeaderContentNegotiationStrategy.resolveMediaTypes throws unexpected IllegalArgumentException 


mhankus:
**Affects:** 6.0.11

Parsing of http accept header throws IllegalArguemntException which is unexpected, and causes unexpected application behavior.
Example stack trace

```
Too many elements
java.lang.IllegalArgumentException: Too many elements
at org.springframework.util.Assert.isTrue(Assert.java:122)
at org.springframework.util.MimeTypeUtils.sortBySpecificity(MimeTypeUtils.java:365)
at org.springframework.web.accept.HeaderContentNegotiationStrategy.resolveMediaTypes(HeaderContentNegotiationStrategy.java:55)
at org.springframework.web.accept.ContentNegotiationManager.resolveMediaTypes(ContentNegotiationManager.java:128)
at org.springframework.web.servlet.mvc.condition.ProducesRequestCondition.getAcceptedMediaTypes(ProducesRequestCondition.java:290)
at org.springframework.web.servlet.mvc.condition.ProducesRequestCondition.getMatchingCondition(ProducesRequestCondition.java:208)
at org.springframework.web.servlet.mvc.method.RequestMappingInfo.getMatchingCondition(RequestMappingInfo.java:401)
at org.springframework.web.servlet.mvc.method.RequestMappingInfoHandlerMapping.getMatchingMapping(RequestMappingInfoHandlerMapping.java:110)
at org.springframework.web.servlet.mvc.method.RequestMappingInfoHandlerMapping.getMatchingMapping(RequestMappingInfoHandlerMapping.java:68)
```

Class HeaderContentNegotiationStrategy currently catches only InvalidMediaTypeException, but IllegalArgumentException can also be thrown here.
```
public List<MediaType> resolveMediaTypes(NativeWebRequest request)
			throws HttpMediaTypeNotAcceptableException {

		String[] headerValueArray = request.getHeaderValues(HttpHeaders.ACCEPT);
		if (headerValueArray == null) {
			return MEDIA_TYPE_ALL_LIST;
		}

		List<String> headerValues = Arrays.asList(headerValueArray);
		try {
			List<MediaType> mediaTypes = MediaType.parseMediaTypes(headerValues);
			MimeTypeUtils.sortBySpecificity(mediaTypes);
			return !CollectionUtils.isEmpty(mediaTypes) ? mediaTypes : MEDIA_TYPE_ALL_LIST;
		}
[...]

PR #31769, opened 2023-12-06
Fix condition for "Too many elements" in `MimeTypeUtils.sortBySpecificity()`

izeye:
This PR fixes condition for "Too many elements" in `MimeTypeUtils.sortBySpecificity()` that seems to have been changed accidentally in 05c3ffb2fbdf358c6a23309a3118b0a64ecb4b40 to align with its Javadoc again.

See gh-31254

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
@@ -362,7 +362,7 @@ public abstract class MimeTypeUtils {
 	 */
 	public static <T extends MimeType> void sortBySpecificity(List<T> mimeTypes) {
 		Assert.notNull(mimeTypes, "'mimeTypes' must not be null");
-		if (mimeTypes.size() >= 50) {
+		if (mimeTypes.size() > 50) {
 			throw new InvalidMimeTypeException(mimeTypes.toString(), "Too many elements");
 		}
 


jandroalvarez:
@sbrannen why haven't you rolled back to the assertion implementation instead of sticking to the InvalidMimeTypeException and removing the equals in the mime types size condition?

sbrannen:
@jandroalvarez, please note that this PR was closed over 3 months ago in 6.1.2.

If you believe you have discovered a further regression or bug, please open a new issue.

Thanks

sbrannen:
To answer your question, it appears there may have been an oversight.

> Why haven't you changed the `catch` clause as originally requested in #31254? The same runtime crash is still happening.

I suppose you're proposing to change what we currently have from:

```java
catch (InvalidMediaTypeException ex) {
	throw new HttpMediaTypeNotAcceptableException(
			"Could not parse 'Accept' header " + headerValues + ": " + ex.getMessage());
}
```

To:

```java
catch (InvalidMediaTypeException | InvalidMimeTypeException ex) {
	throw new HttpMediaTypeNotAcceptableException(
			"Could not parse 'Accept' header " + headerValues + ": " + ex.getMessage());
}
```

Or:

```java
catch (IllegalArgumentException ex) {
	throw new HttpMediaTypeNotAcceptableException(
			"Could not parse 'Accept' header " + headerValues + ": " + ex.getMessage());
}
```

I'm not convinced that we should catch `IllegalArgumentException` there.

But if you're effectively suggesting the former (`InvalidMediaTypeException | InvalidMimeTypeException`), then please [create a new issue](https://github.com/spring-projects/spring-framework/issues/new).

Cheers!

sbrannen:
@jandroalvarez, to ensure the issue you've reported does not get lost, I am going to go ahead and create a new issue based on your input now.

So, there's no need for you to create a new issue.

Once I've created the new issue, I'll post a link back here.

Issue #31669, opened 2023-11-24
DataBinder fails with "No primary or single unique constructor found for interface MultipartFile"


mschneid:
The following example works with Spring Framework 6.0, but not with 6.1:

```java
class MyForm {
	private String name;
	private MultipartFile file;

	MyForm(String name, @Nullable MultipartFile file) {
		this.name = name;
		this.file = file;
	}

	// ...
}

@Controller
public class FileUploadController {
	@PostMapping("/form")
	public String handleFormUpload(MyForm form, BindingResult errors) {
		if (!form.getFile().isEmpty()) {
			byte[] bytes = form.getFile().getBytes();
			// store the bytes somewhere
			return "redirect:uploadSuccess";
		}
		return "redirect:uploadFailure";
	}
}
```

The above code throws the exception `java.lang.IllegalStateException: No primary or single unique constructor found for interface org.springframework.web.multipart.MultipartFile` if the file is not specified in the request.

The method `DataBinder#shouldCreateObject(MethodParameter)` was introduced with #31488. I think this method should also check whether the parameter is optional or not by calling `MethodParameter#isOptional()`.

snicoll:
> I think this method should also check whether the parameter is optional or not by calling MethodParameter#isOptional().

I am not sure about that. Can you check that you compile your code with `-parameters`, see the second paragraph [of this section of the upgrade notes](https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-6.x#core-container). And you should have seen a warning in the logs stating as much with `6.0.x`. If that doesn't help, please move that code in text into a sample we can actually run. You can attach a zip to this issue or push the code to a separate GitHub repository.

jhoeller:
@rstoyanchev is the new check in DataBinder possibly mis-handling `MultipartFile` as a plain bind object there?

mschneid:
@snicoll But Spring Boot sets parameters automatically:
```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-compiler-plugin</artifactId>
  <configuration>
    <parameters>true</parameters>
  </configuration>
</plugin>
```

However, I have created a minimal, fully verifiable example [here](https://github.com/mschneid/spring6.1-databind).

snicoll:
@mschneid apologizes, we've got many report around this and I thought it could have been one more occurrence.

rstoyanchev:
[...]

PR #31778, opened 2023-12-07
Exclude `Part` from nested constructor binding in WebFlux

HeartPattern:
Related to: https://github.com/spring-projects/spring-framework/issues/31669 and https://github.com/spring-projects/spring-framework/commit/9ade52dbe25bdcf5751590c98bacfe7053f5dddf

WebFlux has same issue with nullable multipart file handling. MultipartFile and Part should be excluded from nested constructor binding in WebExchangeDataBinder.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/bind/support/WebExchangeDataBinder.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/support/WebExchangeDataBinder.java
@@ -20,6 +20,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
+import org.springframework.core.MethodParameter;
+import org.springframework.web.multipart.MultipartFile;
 import reactor.core.publisher.Mono;
 
 import org.springframework.beans.MutablePropertyValues;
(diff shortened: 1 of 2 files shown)


sbrannen (changes requested):
Can you please introduce a test that fails before the change and passes after the change?

sbrannen:
> However, it seems like Part and FilePart is not the only problem. Any other nullable types is not bind properly through constructor after [ea398d7](https://github.com/spring-projects/spring-framework/commit/ea398d7b7e72d1e1a9f7a002a291f4e0b8660e70). For example, in kotlin

@HeartPattern, can you please create a separate issue to discuss that?

HeartPattern:
Open another PR addressing that problem.
https://github.com/spring-projects/spring-framework/pull/31800

sbrannen (approved):

spring-web/src/main/java/org/springframework/web/bind/support/WebExchangeDataBinder.java
```diff
@@ -87,6 +89,13 @@ public Mono<Void> construct(ServerWebExchange exchange) {
 				.then();
 	}
 
+	@Override
+	protected boolean shouldConstructArgument(MethodParameter param) {
+		Class<?> type = param.nestedIfOptional().getNestedParameterType();
+		return (super.shouldConstructArgument(param) &&
+				!MultipartFile.class.isAssignableFrom(type) && !Part.class.isAssignableFrom(type));
+	}
```
rstoyanchev, line 97: This is a necessary change indeed for `Part` but `MultipartFile` is not used in WebFlux, so we can leave that out.

rstoyanchev:
I'm processing this locally as we are close to the 6.1.2 release. No need send further updates.

Issue #31742, opened 2023-12-03
Provide more control over charset parameter when sending form data


ffray:
The default charset-parameter added to the `Content-Type` header when using `FormHttpMessageWriter` or `FormHttpMessageConverter` with [SPR-16613](https://github.com/spring-projects/spring-framework/issues/21154) is causing some trouble if the server does not support specifying a charset.

As the spec for `application/x-www-urlencoded` did not exist for a long time and the one provided by [whatwg - application/x-www-form-urlencoded](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) is still somewhat vague,
both clients and servers have implemented various approaches for dealing with this MIME-type.

Earlier HTML specs mention the non-intuitive behavior:
> Parameters on the [application/x-www-form-urlencoded](https://www.w3.org/TR/2014/REC-html5-20141028/iana.html#application/x-www-form-urlencoded) MIME type are ignored. In particular, this MIME type does not support the charset parameter.

At the moment I'm struggling to integrate a piece of software against such server, while trying to use Spring Securities `SpringReactiveOpaqueTokenIntrospector`, as there is currently no way to override the default behavior.

Though I'll use an entire custom implementation, I'd like to propose supporting that adding the charset can be turned off.
I've already created a patch and tests for the `FormHttpMessageWriter` and `BodyInserters` which uses `Hints` to control omitting the charset and would like to contribute if this proposal is accepted.

ffray:
I like the custom codec solution way better than my code.
Didn't notice this simple approach, which comes quite handy!

rstoyanchev:
This was also reported in #22588 and it led to the creation of the `getMediaType` method. Considering the lack of clear guidance in the spec, and that some servers don't handle it well, we should make it easier to submit without the charset parameter.

Issue #22588, opened 2019-03-13
Spring Webflux Webclient APPLICATION_FORM_URLENCODED issue


Plankton93:
I am trying to create a post request via `WebClient` to get Microsoft powerbi token.

```java
MultiValueMap<String, String> credentials = new LinkedMultiValueMap<>();
credentials.add("xxx","xxx");
credentials.add("yyy","yyy");

Mono<String> response = webClientService.getWebclient()
                    .post()
                    .uri(oauthUrl)
                    .contentType(MediaType.APPLICATION_FORM_URLENCODED)
                    .accept(MediaType.APPLICATION_JSON)
                    .body(BodyInserters.fromFormData(credentials))
                    .retrieve()
                    .bodyToMono(String.class);

            log.debug("RESPONSE STRING = "+response.block());
```

But i am always getting 400 bad request exception -> Suppressed: java.lang.Exception: #block terminated with an error.

**I had similar code on Spring Boot 2.0.0.RELEASE where it worked fine. But updating to Spring Boot version 2.1.2.RELEASE makes this crashing. Any suggestions?**

FYI: 

```java
public WebClient getWebclient() {
        TcpClient client = TcpClient.create()
                .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 10000)
                .doOnConnected(connection -> connection.addHandlerLast(new ReadTimeoutHandler(15)).addHandlerLast(new WriteTimeoutHandler(15)));

        return WebClient.builder()
                .clientConnector(new ReactorClientHttpConnector(HttpClient.from(client).followRedirect(true)))
                .filter(logRequest())
                .filter(logResponse())
                .build();
    }
```

I also tried to add:

```java
ExchangeStrategies strategies = ExchangeStrategies.builder()
                .codecs(configurer -> {
                    configurer.registerDefaults(true);
                    configurer.customCodecs().reader(new FormHttpMessageReader());
                })
                .build();
```

but it did not help.

When i curl it i get successful response:

```
     curl -X POST -d 'grant_type=password&client_id=abc&resource=url&scope=openid&username=username&password=password=' -vD --  https://microsofrtUrl
```
```
POST /common/oauth2/token HTTP/1.1
> Host: xxx
> User-Agent: curl/7.61.1
> Accept: */*
> Content-Length: 216
> Content-Type: application/x-www-form-urlencoded

```

rstoyanchev:
[...]

PR #31781, opened 2023-12-07
Provide more control over charset parameter when writing form data

rstoyanchev:
Some servers don't support `application/x-www-form-urlencoded` with a charset parameter, and there isn't clear spec guidance. In light of this, we can avoid adding a charset parameter unless it deviates from the default `UTF-8`.

See #31742 for more details.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageWriter.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/FormHttpMessageWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -60,13 +60,8 @@ import org.springframework.util.MultiValueMap;
 public class FormHttpMessageWriter extends LoggingCodecSupport
 		implements HttpMessageWriter<MultiValueMap<String, String>> {
 
-	/**
-	 * The default charset used by the writer.
-	 */
+	/** The default charset used by the writer. */
 	public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
-
-	private static final MediaType DEFAULT_FORM_DATA_MEDIA_TYPE =
-			new MediaType(MediaType.APPLICATION_FORM_URLENCODED, DEFAULT_CHARSET);
 
 	private static final List<MediaType> MEDIA_TYPES =
 			Collections.singletonList(MediaType.APPLICATION_FORM_URLENCODED);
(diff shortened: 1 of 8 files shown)


hannahBoat:
I'm also experiencing this issue the API we're interacting with doesn't permit application/x-www-form-urlencoded;charset=UTF-8 only application/x-www-form-urlencoded. What's the eta for a release of this fix?

bclozel:
This will be released with 6.2.0, to be released next November.

hannahBoat:
Ok that's quite some time away, is there any chance of being added to a patch release?

rstoyanchev:
@hannahBoat the `getMediaType` method is protected since #22588, so you can override it as a workaround as shown in https://github.com/spring-projects/spring-framework/issues/31742#issuecomment-1839237948.

hannahBoat:
This workaround doesn't work for us unfortunately. We are using Spring Security Ouath2. Which calls into `DefaultClientCredentialsTokenResponseClient` 

The `FormHttpMessageConverter` is set within the constructor of `DefaultClientCredentialsTokenResponseClient`
```
    public DefaultClientCredentialsTokenResponseClient() {
        RestTemplate restTemplate = new RestTemplate(Arrays.asList(new FormHttpMessageConverter(), new OAuth2AccessTokenResponseHttpMessageConverter()));
        restTemplate.setErrorHandler(new OAuth2ErrorResponseErrorHandler());
        this.restOperations = restTemplate;
    }
```

It's the `FormHttpMessageConverter` that's causing the issue. In particular the `writeForm` method. This append's the charset to the content-type header. There's a line that then throws an error if the charset is absent.

```
	private void writeForm(MultiValueMap<String, Object> formData, @Nullable MediaType contentType,
			HttpOutputMessage outputMessage) throws IOException {

		contentType = getFormContentType(contentType);
		outputMessage.getHeaders().setContentType(contentType);

		Charset charset = contentType.getCharset();
		Assert.notNull(charset, "No charset"); // should never occur

		byte[] bytes = serializeForm(formData, charset).getBytes(charset);
		outputMessage.getHeaders().setContentLength(bytes.length);

		if (outputMessage instanceof StreamingHttpOutputMessage streamingOutputMessage) {
			streamingOutputMessage.setBody(new StreamingHttpOutputMessage.Body() {
				@Override
				public void writeTo(OutputStream outputStream) throws IOException {
					StreamUtils.copy(bytes, outputStream);
				}

				@Override
				public boolean repeatable() {
					return true;
				}
			});
		}
		else {
			StreamUtils.copy(bytes, outputMessage.getBody());
		}
	}

```

We would need to extend FormHttpMessageConverter, re write a good proportion of the code to omit the 'charset'. Then feed this into a new 'RestTemplate'

```
RestTemplate restTemplate = new RestTemplate(Arrays.asList(new CustomFormHttpMessageConverter(), new OAuth2AccessTokenResponseHttpMessageConverter()));
restTemplate.setErrorHandler(new OAuth2ErrorResponseErrorHandler());
```

To subsequently get handle on the `DefaultClientCredentialsTokenResponseClient` from the Spring OAuath setup and run setRestOperations with the new restTemplate. 

This feels very convoluted. Particularly when we need to security patch and update spring-web. We may need to update the 'custom' code each time. Is there a more streamlined solution to this issue?

PR #31798, opened 2023-12-09
Ensure that Observation is stopped and Scope is closed in doReceiveAndExecute()

izeye:
This PR changes to try to ensure that `Observation` is stopped and `Scope` is closed in `AbstractPollingMessageListenerContainer.doReceiveAndExecute()` as it seems to be possible for the `finally` clause not to be invoked if an exception is thrown out of the try-finally clause.


bclozel:
Thanks @izeye !

PR #31800, opened 2023-12-09
Ensure `DataBinder` once again binds null values for nullable/optional constructor arguments

HeartPattern:
After https://github.com/spring-projects/spring-framework/commit/ea398d7b7e72d1e1a9f7a002a291f4e0b8660e70, DataBinder failed to bind nullable value through constructor.

We should check if param is nullable and do not trying to resolve nested constructor.


rstoyanchev (commented):
We cannot always exclude optional arguments only because they are optional, or we would never be able to inject a successfully constructed nested argument.

The question I have is how was this working before in the success case? Such a nested class would not have been instantiated. Could you provide more details?

rstoyanchev:
See also #31821, which seems to be the same case. If you could, please clarify either way.

rstoyanchev:
Thanks for the report and a PR. I am preparing a fix for this under #31821.

PR #31802, opened 2023-12-09
Use `String.repeat()` instead of explicit cycle for repeating strings

Drezir:


snicoll:
@Drezir thanks for the contribution. Going forward, please run the build prior to submit the change as it was failing with several checkstyle violations, see https://github.com/spring-projects/spring-framework/commit/d0574197ea5bd2682c6cc478d7685ae7c9351f6e. I also preferred to left our CGLIB and ASM forks unchanged.

Drezir:
> @Drezir thanks for the contribution. Going forward, please run the build prior to submit the change as it was failing with several checkstyle violations, see [d057419](https://github.com/spring-projects/spring-framework/commit/d0574197ea5bd2682c6cc478d7685ae7c9351f6e). I also preferred to left our CGLIB and ASM forks unchanged.

Sorry for that, will do better next time. Thank you

PR #31808, opened 2023-12-11
Fix Comparators.nullsLow and Comporators.nullsHigh behavior

mathieu-amblard:
`nullsLow` should s…ort null values lower than non-null values, `nullsHigh` should sort null values higher than non-null values
The reverse behavior has been introduced in Spring Framework 6.1.X, it was correct before.
See https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/test/java/org/springframework/util/comparator/NullSafeComparatorTests.java which tests the same comparators used before.


sbrannen:

This has been merged into `main` in a01384068ab0c3124e8ca9a48b0fe5de873d5097 and revised in 7432a96b48bff92cad8ea298411b32396210cd19.

Thanks

PR #31845, opened 2023-12-14
Replace deprecated use of `project.buildDir` in `framework-api` Gradle script

bohub12:
Replaces methods that will be deprecated by gradle

_Reference_
- [gradle 8.5 official documentation - Upgrading version 8.x to latest](https://docs.gradle.org/current/userguide/upgrading_version_8.html#project_builddir)

### DIFF
--- a/framework-api/framework-api.gradle
+++ b/framework-api/framework-api.gradle
@@ -28,7 +28,7 @@ javadoc {
 		header = rootProject.description
 		use = true
 		overview = project.relativePath("$rootProject.rootDir/framework-docs/src/docs/api/overview.html")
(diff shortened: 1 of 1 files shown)


sbrannen (changes requested):
As stated in the Gradle documentation:

> The switch from a File to a DirectoryProperty requires adaptations in build logic. The main impact is that you cannot use the property inside a String to expand it. Instead, you should leverage the dir and file methods to compute the location you want.

Thus, the change you have suggested is invalid.

Please update your PR with appropriate build logic.

bohub12:
I have fixed DirectoryProperty to not be handled inside a String as written in the documentation.

snicoll:
Thanks for following up @bohub12.

PR #31847, opened 2023-12-14
Combine if-statements in `ScopedProxyBeanDefinitionDecorator`

achhibi:
This pull request simplifies the code and makes it more readable by combining two conditions into a single statement using the logical AND operator (`&&`). Now, the enclosed statement executes only when both conditions (`a` and ` b`) are `true`.

Your review and consideration for merging are appreciated to improve overall code quality.


snicoll:
@achhibi as I've mentioned previously, we prefer to do such stylistic code change ourselves in general. I am going to merge this one but please the above in consideration going forward.

achhibi:
@snicoll Thank you for taking the time to review; I will take your feedback into consideration.

PR #31869, opened 2023-12-20
Complete frame callback when opcode is not PONG

leonchen83:
onWebSocketFrame method should complete callback. more details see issue [JETTY 11088](https://github.com/jetty/jetty.project/issues/11088)


sbordet (approved):
Good catch!

sdeleuze:
@sbordet From your perspective, should we apply this fix to the Jetty 11 support we have in Spring Framework 6.0.x as well?

sbordet:
@sdeleuze Jetty 11 does not need this fix because it does not use `Callback`.

The corresponding signature in Jetty 11 would be:

```java
@OnWebSocketFrame
public void onWebSocketFrame(Frame frame)
```

so no `Callback`.
In Jetty 11 there is the requirement that the frame payload must be consumed synchronously in the method invocation, or copied to be consumed asynchronously, because when returning from the `onWebSocketFrame()` method the frame payload buffer is recycled.

PR #31879, opened 2023-12-21
Polish of getOrder

Yaklede:
I polished code of the getOrder of OrderComparator,
I made it easier to read the codes


snicoll:
@Yaklede thanks we prefer to make stylistic code change like that ourselves as it's largely subjective. I believe the code is fine as it is.

PR #31886, opened 2023-12-22
Specify use of regexp Path Pattern in javadoc

FBibonne:
As far I saw how the framework process web requests Mapping and as far I read the code of CaptureVariablePathElement, when using  a path pattern with a regexp (like `{spring:[a-z]+}`) in a @RequestMapping, the regexp can match only one path segment of the path of the request. So this PR proposes to improve the javadoc of PathPattern to specify that the regexp match can only happen in one path segment.


bclozel:
Thanks @FBibonne for your contribution!

PR #31899, opened 2023-12-23
Fix formatting in scheduling.adoc

kzander91:
This fixes two minor formatting issues in the task scheduling documentation.


jhoeller:
Good catch, thanks for the PR!

PR #31917, opened 2023-12-28
Fix usage of WebClientAdapter in reference documentation

0x1306e6d:
`WebClientAdapter#forClient(WebClient)` is deprecated. Use `WebClientAdapter#create(WebClient)` instead and fix typo.


snicoll:
Good catch @ghkim3221 and thank you for making your first contribution to Spring Framework. There was an extra usage in the same file so I've fixed that as well, see [b1c0b65](https://github.com/spring-projects/spring-framework/commit/b1c0b6566630a8605366c18e015d8d0bf2df5390)

ghkim3221:
Thank you !!!

PR #31924, opened 2023-12-29
Use UTF-8 by default for JSON multipart content in `ContentRequestMatchers`

### LINKED ISSUE #31923: ContentRequestMatchers.MultipartHelper default charset
**Affects:** 6.1.2

---

While using the [@RestClientTest](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/autoconfigure/web/client/RestClientTest.html) I had an issue where I sent a [Multipart](https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#_multipart) request and wanted to confirm in a test that it was correctly formed.

But during testing I noticed that the request was handled by MultipartHelper and it didn't parse correctly my `application/json` part in `UTF-8`, resulting in some symbols being malformed. Although, it worked correctly while using the deprecated `application/json; charset=utf-8`. https://github.com/spring-projects/spring-framework/blob/f5b4f7d9e89ba21e2fd842f71d8a9af1dcfa0535/spring-test/src/main/java/org/springframework/test/web/client/match/ContentRequestMatchers.java#L361

After further investigation I noticed that `DiskFileItemFactory` is used with default values, with the charset set to `this.defaultCharset = "ISO-8859-1"`.

Would it make sense for `MultipartHelper` to default to UTF-8?

```java
FileUpload fileUpload = new FileUpload();
DiskFileItemFactory factory = new DiskFileItemFactory();
factory.setDefaultCharset(StandardCharsets.UTF_8.name());
fileUpload.setFileItemFactory(factory);
```

azdanov:
This should resolve https://github.com/spring-projects/spring-framework/issues/31923

I've added default UTF-8 charset to be used in `DiskFileItemFactory`.

This should allow ContentRequestMatchers to correctly handle multipartData which contains unicode.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/client/match/ContentRequestMatchers.java
+++ b/spring-test/src/main/java/org/springframework/test/web/client/match/ContentRequestMatchers.java
@@ -60,6 +60,8 @@ import static org.springframework.test.util.AssertionErrors.assertTrue;
  */
 public class ContentRequestMatchers {
 
+	private static final String DEFAULT_ENCODING = "UTF-8";
+
 	private final XmlExpectationsHelper xmlHelper;
 
 	private final JsonExpectationsHelper jsonHelper;
@@ -363,7 +365,9 @@ public class ContentRequestMatchers {
 		public static MultiValueMap<String, ?> parse(MockClientHttpRequest request) {
 			try {
 				FileUpload fileUpload = new FileUpload();
-				fileUpload.setFileItemFactory(new DiskFileItemFactory());
+				DiskFileItemFactory factory = new DiskFileItemFactory();
+				factory.setDefaultCharset(DEFAULT_ENCODING);
+				fileUpload.setFileItemFactory(factory);
 
 				List<FileItem> fileItems = fileUpload.parseRequest(new UploadContext() {
 					private final byte[] body = request.getBodyAsBytes();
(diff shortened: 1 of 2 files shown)


spring-test/src/main/java/org/springframework/test/web/client/match/ContentRequestMatchers.java
```diff
@@ -363,7 +365,9 @@ private static class MultipartHelper {
 		public static MultiValueMap<String, ?> parse(MockClientHttpRequest request) {
 			try {
 				FileUpload fileUpload = new FileUpload();
-				fileUpload.setFileItemFactory(new DiskFileItemFactory());
+				DiskFileItemFactory factory = new DiskFileItemFactory();
+				factory.setDefaultCharset(DEFAULT_ENCODING);
+				fileUpload.setFileItemFactory(factory);
```
rstoyanchev, line 370: Switching the general default could break existing tests, and that should wait for a major or minor release. In the meantime, we could narrow the scope and default to UTF-8 specifically for JSON. That would be safer and addresses the original issue.

rstoyanchev (changes requested):
Let me know if you agree with the proposed change below and whether you'd like to do it?

azdanov:
I agree, but I don't have time now. So feel free to add.

PR #31932, opened 2024-01-02
Fix Kotlin example for simpler SELECT variant using IN

janjouketjalsma:
Removed unused code, fixed bind parameter name.


snicoll:
@janjouketjalsma thank you for making your first contribution to Spring Framework.

PR #31939, opened 2024-01-03
Update `autowired-qualifiers.adoc` to refer the `-parameters` Java compiler flag

dodie:
As described in [Parameter Name Retention section for the Spring Upgrade Guide for 6.1](https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-6.x#parameter-name-retention), Spring only matches the injection point name (e.g. parameter name of a constructor) against the target bean names if the `-parameters` Java compiler flag is present.


jhoeller:
Indeed, that's worth pointing out right there! Thanks for the pull request.

PR #31942, opened 2024-01-03
Add error handling function in R2DBC `DatabaseClient`

### LINKED ISSUE #31256: Add error handling strategy to R2DBC DatabaseClient
We currently have an issue where the database might throw unexpected exceptions. To handle that we would like to catch those exceptions and close the connection to the database. I.e we need to add this code:

```
    @Override
    public <T> Mono<T> inConnection(Function<Connection, Mono<T>> action) throws DataAccessException {
        return super.inConnection(con -> action.apply(con)
                .onErrorResume(e -> {
                    handleError(e, con);
                    return Mono.error(e);
                }));
    }

    @Override
    public <T> Flux<T> inConnectionMany(Function<Connection, Flux<T>> action) throws DataAccessException {
        return super.inConnectionMany(con -> action.apply(con)
                .onErrorResume(e -> {
                    handleError(e, con);
                    return Mono.error(e);
                }));
    }
```

However since DefaultDatabaseClient is package private it is not possible to extend it. 

The easiest solution would be to add a delegating client which just delegates all calls to the underlying client and adds that error handling on top. However the internal call structure of DefaultDatabaseClient makes that impossible. Calls to the "sql" method returns an instance of the internal class DefaultGenericExecuteSpec. Since that class is an internal class it has a reference to the DefaultDatabaseClient, that reference gets exposed externally and when a client calls the "map" method on the ExecuteSpec. That reference is used to create an FetchSpec and then the FetchSpec will use that reference to fetch the data, not the delegating client completely bypassing the custom error handling.

I propose to expose DefaultGenericExecuteSpec externally and let the caller provide a client with which to do the requests. Then a delegating client could create a new instance with itself as client.

injae-kim:
Closes gh-31256

### Motivation
```java
// users want to handle error like this on various cases!
    @Override
    public <T> Mono<T> inConnection(Function<Connection, Mono<T>> action) throws DataAccessException {
        return super.inConnection(con -> action.apply(con)
                .onErrorResume(e -> {
                    handleError(e, con); // here
                    return Mono.error(e);
                }));
    }

    @Override
    public <T> Flux<T> inConnectionMany(Function<Connection, Flux<T>> action) throws DataAccessException {
        return super.inConnectionMany(con -> action.apply(con)
                .onErrorResume(e -> {
                    handleError(e, con); // here
                    return Mono.error(e);
                }));
    }
```
- We found that users want to handle error by user's custom error handling logic on `DatabaseClient`, but currently impossible

### Modification
```java
DatabaseClient.builder()
    .handleInConnectionError((t, conn) -> { /* error handling logic */})
    .build()
```
- Add error handling function on R2DBC DatabaseClient and it's builder

### Result
- Now user can use their custom error handling logic on R2DBC DatabaseClient
- Close https://github.com/spring-projects/spring-framework/issues/31256

### DIFF
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,18 +17,10 @@
 package org.springframework.r2dbc.core;
 
 import java.util.Map;
-import java.util.function.BiFunction;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.function.Supplier;
+import java.util.function.*;
 
-import io.r2dbc.spi.ConnectionFactory;
+import io.r2dbc.spi.*;
 import io.r2dbc.spi.Readable;
-import io.r2dbc.spi.Result;
-import io.r2dbc.spi.Row;
-import io.r2dbc.spi.RowMetadata;
-import io.r2dbc.spi.Statement;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -56,6 +48,7 @@ import org.springframework.util.Assert;
  *
  * @author Mark Paluch
  * @author Juergen Hoeller
+ * @author Injae Kim
  * @since 5.3
  */
 public interface DatabaseClient extends ConnectionAccessor {
@@ -145,6 +138,12 @@ public interface DatabaseClient extends ConnectionAccessor {
 		 * @see NamedParameterExpander
 		 */
 		Builder namedParameters(boolean enabled);
+
+		/**
+		 * Handle the given error on {@link ConnectionAccessor#inConnection}
+		 * or {@link ConnectionAccessor#inConnectionMany}.
+		 */
+		Builder handleInConnectionError(BiConsumer<? super Throwable, ? super Connection> handleInConnectionError);
 
 		/**
 		 * Apply a {@link Consumer} to configure this builder.
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
@@ -150,10 +150,12 @@ public interface DatabaseClient extends ConnectionAccessor {
 		Builder namedParameters(boolean enabled);
 
 		/**
-		 * Handle the given error on {@link ConnectionAccessor#inConnection}
+		 * Handle an exception thrown from {@link ConnectionAccessor#inConnection}
 		 * or {@link ConnectionAccessor#inConnectionMany}.
+		 * @param errorHandler a {@code BiConsumer} that handles the connection error
+		 * @since 6.2
 		 */
-		Builder onConnectionError(BiConsumer<? super Throwable, ? super Connection> onConnectionError);
+		Builder onConnectionError(BiConsumer<? super Throwable, ? super Connection> errorHandler);
 
 		/**
 		 * Apply a {@link Consumer} to configure this builder.
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
@@ -17,10 +17,20 @@
 package org.springframework.r2dbc.core;
 
 import java.util.Map;
-import java.util.function.*;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 
-import io.r2dbc.spi.*;
+import io.r2dbc.spi.Connection;
+import io.r2dbc.spi.ConnectionFactory;
 import io.r2dbc.spi.Readable;
+import io.r2dbc.spi.Result;
+import io.r2dbc.spi.Row;
+import io.r2dbc.spi.RowMetadata;
+import io.r2dbc.spi.Statement;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
@@ -153,7 +153,7 @@ public interface DatabaseClient extends ConnectionAccessor {
 		 * Handle the given error on {@link ConnectionAccessor#inConnection}
 		 * or {@link ConnectionAccessor#inConnectionMany}.
 		 */
-		Builder handleInConnectionError(BiConsumer<? super Throwable, ? super Connection> handleInConnectionError);
+		Builder onConnectionError(BiConsumer<? super Throwable, ? super Connection> onConnectionError);
 
 		/**
 		 * Apply a {@link Consumer} to configure this builder.
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
@@ -83,7 +83,7 @@ final class DefaultDatabaseClient implements DatabaseClient {
 
 	private final ExecuteFunction executeFunction;
 
-	private final BiConsumer<? super Throwable, ? super Connection> handleInConnectionError;
+	private final BiConsumer<? super Throwable, ? super Connection> onConnectionError;
 
 	@Nullable
 	private final NamedParameterExpander namedParameterExpander;
@@ -91,13 +91,13 @@ final class DefaultDatabaseClient implements DatabaseClient {
 
 	DefaultDatabaseClient(BindMarkersFactory bindMarkersFactory, ConnectionFactory connectionFactory,
 						  ExecuteFunction executeFunction,
-						  BiConsumer<? super Throwable, ? super Connection> handleInConnectionError,
+						  BiConsumer<? super Throwable, ? super Connection> onConnectionError,
 						  boolean namedParameters) {
 
 		this.bindMarkersFactory = bindMarkersFactory;
 		this.connectionFactory = connectionFactory;
 		this.executeFunction = executeFunction;
-		this.handleInConnectionError = handleInConnectionError;
+		this.onConnectionError = onConnectionError;
 		this.namedParameterExpander = (namedParameters ? new NamedParameterExpander() : null);
 	}
 
@@ -130,10 +130,10 @@ final class DefaultDatabaseClient implements DatabaseClient {
 			Connection connectionToUse = createConnectionProxy(connectionCloseHolder.connection);
 					try {
 						return action.apply(connectionToUse)
-								.doOnError(t -> handleInConnectionError.accept(t, connectionCloseHolder.connection));
+								.doOnError(t -> onConnectionError.accept(t, connectionCloseHolder.connection));
 					}
 					catch (Throwable ex) {
-						handleInConnectionError.accept(ex, connectionCloseHolder.connection);
+						onConnectionError.accept(ex, connectionCloseHolder.connection);
 						if (ex instanceof R2dbcException) {
 							String sql = getSql(action);
 							return Mono.error(ConnectionFactoryUtils.convertR2dbcException(
@@ -158,10 +158,10 @@ final class DefaultDatabaseClient implements DatabaseClient {
 			Connection connectionToUse = createConnectionProxy(connectionCloseHolder.connection);
 					try {
 						return action.apply(connectionToUse)
-								.doOnError(t -> handleInConnectionError.accept(t, connectionCloseHolder.connection));
+								.doOnError(t -> onConnectionError.accept(t, connectionCloseHolder.connection));
 					}
 					catch (Throwable ex) {
-						handleInConnectionError.accept(ex, connectionCloseHolder.connection);
+						onConnectionError.accept(ex, connectionCloseHolder.connection);
 						if (ex instanceof R2dbcException) {
 							String sql = getSql(action);
 							return Flux.error(ConnectionFactoryUtils.convertR2dbcException(
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -68,6 +69,7 @@ import org.springframework.util.StringUtils;
  * @author Bogdan Ilchyshyn
  * @author Simon Baslé
  * @author Juergen Hoeller
+ * @author Injae Kim
  * @since 5.3
  * @see DatabaseClient#create(ConnectionFactory)
  */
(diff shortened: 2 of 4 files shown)


jhoeller:
Thanks for the pull request! From a quick first glance, this looks pretty good to me :-)

A code-style note: Please do not use star imports such as `io.r2dbc.spi.*`; we always list all imported classes explicitly. You may run the `check` build target instead of `test`, including our Checkstyle task as well which checks for such common code style issues.

A naming note: `handleInConnectionError` is descriptive but sounds like a protected template method rather than a callback. Maybe `onConnectionError` with an event-style `on` prefix instead? The specific semantics of being called for exceptions that come up within the `inConnection` method could be put into a javadoc, still making it descriptive enough.

injae-kim:
oh thanks a lot for detailed review @jhoeller !

- Fix `import *` to import all classes explicitly
- Fix naming `handleInConnectionError` to `onConnectionError`
- Run `checkStyle` and `build` and check success

✅ I updated PR like above based on your suggestion. thanks!

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
```diff
@@ -146,6 +149,12 @@ interface Builder {
 		 */
 		Builder namedParameters(boolean enabled);
 
+		/**
+		 * Handle the given error on {@link ConnectionAccessor#inConnection}
+		 * or {@link ConnectionAccessor#inConnectionMany}.
+		 */
+		Builder onConnectionError(BiConsumer<? super Throwable, ? super Connection> onConnectionError);
```
sbrannen, line 156:
```suggestion
		/**
		 * Handle an exception thrown from {@link ConnectionAccessor#inConnection}
		 * or {@link ConnectionAccessor#inConnectionMany}.
		 * @param errorHandler a {@code BiConsumer} that handles the connection error
		 * @since 6.2
		 */
		Builder onConnectionError(BiConsumer<? super Throwable, ? super Connection> errorHandler);
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
```diff
@@ -81,16 +83,22 @@ final class DefaultDatabaseClient implements DatabaseClient {
 
 	private final ExecuteFunction executeFunction;
 
+	private final BiConsumer<? super Throwable, ? super Connection> onConnectionError;
```
sbrannen, line 86:
```suggestion
	private final BiConsumer<? super Throwable, ? super Connection> errorHandler;
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
```diff
@@ -81,16 +83,22 @@ final class DefaultDatabaseClient implements DatabaseClient {
...
 	@Nullable
 	private final NamedParameterExpander namedParameterExpander;
 
 
-	DefaultDatabaseClient(BindMarkersFactory bindMarkersFactory, ConnectionFactory connectionFactory,
-			ExecuteFunction executeFunction, boolean namedParameters) {
+	DefaultDatabaseClient(BindMarkersFactory bindMarkersFactory,
+			ConnectionFactory connectionFactory,
+			ExecuteFunction executeFunction,
+			BiConsumer<? super Throwable, ? super Connection> onConnectionError,
```
sbrannen, line 95:
```suggestion
			BiConsumer<? super Throwable, ? super Connection> errorHandler,
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
```diff
@@ -81,16 +83,22 @@ final class DefaultDatabaseClient implements DatabaseClient {
...
+			boolean namedParameters) {
 
 		this.bindMarkersFactory = bindMarkersFactory;
 		this.connectionFactory = connectionFactory;
 		this.executeFunction = executeFunction;
+		this.onConnectionError = onConnectionError;
```
sbrannen, line 101:
```suggestion
		this.errorHandler = errorHandler;
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
```diff
@@ -122,11 +130,17 @@ public <T> Mono<T> inConnection(Function<Connection, Mono<T>> action) {
 			// Create close-suppressing Connection proxy
 			Connection connectionToUse = createConnectionProxy(connectionCloseHolder.connection);
 					try {
-						return action.apply(connectionToUse);
+						return action.apply(connectionToUse)
+								.doOnError(t -> this.onConnectionError.accept(t, connectionCloseHolder.connection));
```
sbrannen, line 134:
```suggestion
								.doOnError(t -> this.errorHandler.accept(t, connectionCloseHolder.connection));
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
```diff
@@ -122,11 +130,17 @@ public <T> Mono<T> inConnection(Function<Connection, Mono<T>> action) {
...
 					}
-					catch (R2dbcException ex) {
-						String sql = getSql(action);
-						return Mono.error(ConnectionFactoryUtils.convertR2dbcException("doInConnection", sql, ex));
+					catch (Throwable ex) {
+						this.onConnectionError.accept(ex, connectionCloseHolder.connection);
```
sbrannen, line 137:
```suggestion
						this.errorHandler.accept(ex, connectionCloseHolder.connection);
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
```diff
@@ -144,11 +158,17 @@ public <T> Flux<T> inConnectionMany(Function<Connection, Flux<T>> action) {
 			// Create close-suppressing Connection proxy, also preparing returned Statements.
 			Connection connectionToUse = createConnectionProxy(connectionCloseHolder.connection);
 					try {
-						return action.apply(connectionToUse);
+						return action.apply(connectionToUse)
+								.doOnError(t -> this.onConnectionError.accept(t, connectionCloseHolder.connection));
```
sbrannen, line 162:
```suggestion
								.doOnError(t -> this.errorHandler.accept(t, connectionCloseHolder.connection));
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClient.java
```diff
@@ -144,11 +158,17 @@ public <T> Flux<T> inConnectionMany(Function<Connection, Flux<T>> action) {
...
 					}
-					catch (R2dbcException ex) {
-						String sql = getSql(action);
-						return Flux.error(ConnectionFactoryUtils.convertR2dbcException("doInConnectionMany", sql, ex));
+					catch (Throwable ex) {
+						this.onConnectionError.accept(ex, connectionCloseHolder.connection);
```
sbrannen, line 165:
```suggestion
						this.errorHandler.accept(ex, connectionCloseHolder.connection);
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClientBuilder.java
```diff
@@ -42,6 +45,8 @@ class DefaultDatabaseClientBuilder implements DatabaseClient.Builder {
 
 	private ExecuteFunction executeFunction = Statement::execute;
 
+	private BiConsumer<? super Throwable, ? super Connection> onConnectionError = (t, conn) -> {};
```
sbrannen, line 48:
```suggestion
	private BiConsumer<? super Throwable, ? super Connection> errorHandler = (t, conn) -> {};
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClientBuilder.java
```diff
@@ -76,6 +81,13 @@ public DatabaseClient.Builder namedParameters(boolean enabled) {
 		return this;
 	}
 
+	@Override
+	public DatabaseClient.Builder onConnectionError(
+			BiConsumer<? super Throwable, ? super Connection> onConnectionError) {
+		this.onConnectionError = onConnectionError;
+		return this;
+	}
```
sbrannen, line 89:
```suggestion
	public DatabaseClient.Builder onConnectionError(
			BiConsumer<? super Throwable, ? super Connection> errorHandler) {
		this.errorHandler = errorHandler;
		return this;
	}
```

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DefaultDatabaseClientBuilder.java
```diff
@@ -91,7 +103,8 @@ public DatabaseClient build() {
 		}
 
 		return new DefaultDatabaseClient(
-				bindMarkers, this.connectionFactory, this.executeFunction, this.namedParameters);
+				bindMarkers, this.connectionFactory,
+				this.executeFunction, this.onConnectionError, this.namedParameters);
```
sbrannen, line 107:
```suggestion
				this.executeFunction, this.errorHandler, this.namedParameters);
```

spring-r2dbc/src/test/java/org/springframework/r2dbc/core/DefaultDatabaseClientUnitTests.java
```diff
@@ -130,6 +132,98 @@ public void onComplete() {
 		verify(connection, times(1)).close();
 	}
 
+	@Test
+	void onConnectionError() {
+		AtomicInteger onErrorCounter = new AtomicInteger();
+		BiConsumer<Throwable, Connection> onConnectionError = (t, conn) -> {
```
sbrannen, line 138:
```suggestion
		BiConsumer<Throwable, Connection> errorHandler = (t, conn) -> {
```

spring-r2dbc/src/test/java/org/springframework/r2dbc/core/DefaultDatabaseClientUnitTests.java
```diff
@@ -130,6 +132,98 @@ public void onComplete() {
...
+			onErrorCounter.incrementAndGet();
+			conn.getMetadata();
+		};
+
+		DefaultDatabaseClient databaseClient = (DefaultDatabaseClient) databaseClientBuilder
+				.onConnectionError(onConnectionError)
```
sbrannen, line 144:
```suggestion
				.onConnectionError(errorHandler)
```

spring-r2dbc/src/test/java/org/springframework/r2dbc/core/DefaultDatabaseClientUnitTests.java
```diff
@@ -130,6 +132,98 @@ public void onComplete() {
...
+					assertThat(onErrorCounter.get()).isOne();
+					verify(connection, times(1)).getMetadata();
+				});
+		verify(connection, times(1)).close();
+	}
+
+	@Test
+	void onConnectionErrorActionThrow() {
+		AtomicInteger onErrorCounter = new AtomicInteger();
+		BiConsumer<Throwable, Connection> onConnectionError = (t, conn) -> {
```
sbrannen, line 160:
```suggestion
		BiConsumer<Throwable, Connection> errorHandler = (t, conn) -> {
```

spring-r2dbc/src/test/java/org/springframework/r2dbc/core/DefaultDatabaseClientUnitTests.java
```diff
@@ -130,6 +132,98 @@ public void onComplete() {
...
+			onErrorCounter.incrementAndGet();
+			conn.getMetadata();
+		};
+
+		DefaultDatabaseClient databaseClient = (DefaultDatabaseClient) databaseClientBuilder
+				.onConnectionError(onConnectionError)
```
sbrannen, line 166:
```suggestion
				.onConnectionError(errorHandler)
```

spring-r2dbc/src/test/java/org/springframework/r2dbc/core/DefaultDatabaseClientUnitTests.java
```diff
@@ -130,6 +132,98 @@ public void onComplete() {
...
+					assertThat(onErrorCounter.get()).isOne();
+					verify(connection, times(1)).getMetadata();
+				});
+		verify(connection, times(1)).close();
+	}
+
+	@Test
+	void onConnectionErrorInConnectionMany() {
+		AtomicInteger onErrorCounter = new AtomicInteger();
+		BiConsumer<Throwable, Connection> onConnectionError = (t, conn) -> {
```
sbrannen, line 184:
```suggestion
		BiConsumer<Throwable, Connection> errorHandler = (t, conn) -> {
```

spring-r2dbc/src/test/java/org/springframework/r2dbc/core/DefaultDatabaseClientUnitTests.java
```diff
@@ -130,6 +132,98 @@ public void onComplete() {
...
+			onErrorCounter.incrementAndGet();
+			conn.getMetadata();
+		};
+
+		DefaultDatabaseClient databaseClient = (DefaultDatabaseClient) databaseClientBuilder
+				.onConnectionError(onConnectionError)
```
sbrannen, line 190:
```suggestion
				.onConnectionError(errorHandler)
```

spring-r2dbc/src/test/java/org/springframework/r2dbc/core/DefaultDatabaseClientUnitTests.java
```diff
@@ -130,6 +132,98 @@ public void onComplete() {
...
+	@Test
+	void onConnectionErrorInConnectionManyActionThrow() {
+		AtomicInteger onErrorCounter = new AtomicInteger();
+		BiConsumer<Throwable, Connection> onConnectionError = (t, conn) -> {
+			onErrorCounter.incrementAndGet();
+			conn.getMetadata();
+		};
+
+		DefaultDatabaseClient databaseClient = (DefaultDatabaseClient) databaseClientBuilder
+				.onConnectionError(onConnectionError)
```
sbrannen, line 212:
```suggestion
				.onConnectionError(errorHandler)
```

sbrannen (changes requested):
Please rename instances of the `BiConsumer` to `errorHandler` (or perhaps `connectionErrorHandler`).

Thanks

jhoeller:
@sbrannen @injae-kim indeed, what we register via `onConnectionError` is an error handler function, terminology-wise, and that should be reflected in the argument name.

@mp911de does the overall arrangement look sensible to you?

mp911de:
Taking a step back, I wonder what we actually want to achieve. Generally, connections are released (closed) upon an exception and the error signal is dispatched to the subscriber. 

`DatabaseClient` has a concept of functions (`ExecuteFunction`, `StatementFilterFunction`) so I'm wondering whether it would rather make sense to have `ConnectionFilterFunction` that acts as hook for all `inConnection[Many]` invocations. `inConnection` was built with the intent to serve as customization hook. Introducing a filter function that can serve this purpose would take the error handler into consideration and provide customization beyond just error handling.

injae-kim:
>Taking a step back, I wonder what we actually want to achieve.

I just want to make user can handleError with connection like `handleError(t, conn)` on `inConnection[Many]`!
Currently it's impossible cause connection is only used inside of `inConnection[Many]`.

> I'm wondering whether it would rather make sense to have ConnectionFilterFunction that acts as hook for all inConnection[Many] invocations.

@mp911de , thanks a lot for your suggestion!
Can you share some example code about `ConnectionFilterFunction` usage? 👍

injae-kim:
>  (@mp911de) Introducing a `ConnectionFilterFunction` that can serve this purpose would take the error handler into consideration and provide customization beyond just error handling.

Hi @@sbrannen, @jhoeller what do you think about above suggestion?

If you agree with this, please share to me~! I'll update this PR with this way!
(introducing `ConnectionFilterFunction` instead of `BiConsumer<> errorHandler`)

spring-r2dbc/src/main/java/org/springframework/r2dbc/core/DatabaseClient.java
injae-kim: oh thanks a lot for detailed review!
after checking https://github.com/spring-projects/spring-framework/pull/31942#issuecomment-1877220128, I'll update PR.

injae-kim: ✅ updated! rename all `BiConsumer onConnectionError` to `errorHandler`.

sbrannen (approved):

injae-kim:
Rebase to upstream/main to resolve conflict on test~!

snicoll:
Looking at the latest state of this PR, I don't really see how the consumer is related to "connection error".  The function that uses the connection may also lead to an error, and if closing the connection fails, it is not invoked. Looking at the javadoc of `onConnectionError` I don't believe it matches what the algorithm does.

injae-kim:
https://github.com/spring-projects/spring-framework/blob/fdf187ec465da1d8cc04a8f1b4b892a32f346e33/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/ConnectionAccessor.java#L43-L47

```java
/**
 * Handle an exception thrown from {@link ConnectionAccessor#inConnection}
 * or {@link ConnectionAccessor#inConnectionMany} 👉 (new) during executing a callback {@link Function} within a {@link Connection} scope.
 * @param errorHandler a {@code BiConsumer} that handles the connection error
 * @since 6.2
 */
Builder onConnectionError(BiConsumer<? super Throwable, ? super Connection> errorHandler)
```

`onConnectionError` actually consumes error thrown during `action.apply(connection)`.
And `action` is a callback function within a connection scope, based on `ConnectionAccessor.inConnection*` comment.

So it's good to enhance `onConnectionError` javadoc like above? matching with what it really does.

snicoll:
> So it's good to enhance onConnectionError javadoc like above? matching with what it really does.

I don't think so and I am not sure that is where we want to go anyway. Brainstorming with @mp911de, we'd like to take a step back and resume the conversation on the issue. Specifically by sharing a representative sample of what that would achieve. It's unclear if you're affected by that issue as well or if you're trying to help based on the original report. Either way, let's continue chatting on the issue. Thanks for the PR, in any case!

injae-kim:
> Either way, let's continue chatting on the issue.

aha I got it! let's continue to discuss on issue~ thanks a lot for review 🙇

PR #31945, opened 2024-01-04
Upgrade to AssertJ 3.25.1, use AssertJ BOM

scordio:


sbrannen:

Thanks, @scordio! 👍

spring-beans/src/test/java/org/springframework/beans/factory/config/YamlProcessorTests.java
```diff
@@ -146,9 +146,7 @@ void flattenedMapIsSameAsPropertiesButOrdered() {
 	void standardTypesSupportedByDefault() {
 		setYaml("value: !!set\n  ? first\n  ? second");
 		this.processor.process((properties, map) -> {
-			// Assert on Properties as a Map due to bug in AssertJ 3.25.0
-			Map<Object, Object> propsAsMap = new LinkedHashMap<>(properties);
-			assertThat(propsAsMap).containsExactly(entry("value[0]", "first"), entry("value[1]", "second"));
+			assertThat(properties).containsExactly(entry("value[0]", "first"), entry("value[1]", "second"));
 			assertThat(map.get("value")).asInstanceOf(InstanceOfAssertFactories.type(Set.class))
 					.satisfies(set -> assertThat(set).containsExactly("first", "second"));
```
scordio, line 151: I'm thinking about how to simplify these two lines and get rid of the `@SuppressWarnings("unchecked")`.

Unfortunately, the following:

```java
assertThat(map.get("value")).asInstanceOf(InstanceOfAssertFactories.COLLECTION)
		.containsExactly("first", "second");
```

is not good enough because it doesn't ensure that `actual` is a `Set` instance.

By design, we didn't introduce any `InstanceOfAssertFactory` for types that don't have more specific assertions, i.e., there is no `AbstractSetAssert` today because it wouldn't have any added value compared to [`AbstractCollectionAssert`](https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractCollectionAssert.html).

A way to avoid the `@SuppressWarnings` and get the same behavior could be:

```java
assertThat(map.get("value")).asInstanceOf(new InstanceOfAssertFactory<>(Set.class, Assertions::<String> assertThat))
		.containsExactly("first", "second");
```

which is clearly not nicer...

sbrannen: Nope, not nicer, but thanks for trying. 😉 

We often run into issues with generics and unchecked warnings when using AssertJ, but... we don't spend too much time on it. Rather, we just do our best to get it working and move on.

scordio: I'm not giving up! https://github.com/assertj/assertj/issues/3322

I'll also look for other cases in the codebase that could benefit from improvements and feel free to tag me when you encounter such cases... I love puzzles 😄

sbrannen: Grazie mille, Stefano! 🙇

scordio: assertj/assertj#3325 will bring an improvement 🙂

PR #31947, opened 2024-01-04
Check for AOP proxies in AbstractHandlerMethodMapping

MeneDev:
# Motivation

When enabling Spring Security MethodSecurity, I noticed that the Controller under test was not recoginized as Controller anymore in MockMvc Tests.

This provides a fix, but needs more investigation. The way I solved it is probably not ideal, and if the way is fine, it still needs refactoring. Maybe the problem is also to be found in the test setup and is considered expected behavior. I will add further information soon, but I cannot share the code causing the behavior directly. I did add a test wich seems to replicate the behavior.

I'll try to contribute over the weekend, but may only be available again on monday.

- [ ] Checkstyle
- [ ] All tests pass locally


MeneDev:
I realized this is expected behavior as documented under https://docs.spring.io/spring-framework/reference/6.1-SNAPSHOT/web/webmvc/mvc-controller/ann.html#mvc-ann-requestmapping-proxying

PR #31951, opened 2024-01-05
Fix references to "application/*+xml" in Javadoc

pri88yank:
Small typo in comments of Media Type supported


snicoll:
@pri88yank thank you for making your first contribution to Spring Framework. I've found two more references to that typo and I've fixed that in https://github.com/spring-projects/spring-framework/commit/549f6c1e8007191ec627302524e9c4586efca587

PR #31965, opened 2024-01-07
Delete some useless code

The-Gamer-01:
This pr is my first pr to spring-framework.
This pr uses IDEA code analysis to delete some useless code.

### DIFF
--- a/spring-core/src/main/java/org/springframework/asm/FieldWriter.java
+++ b/spring-core/src/main/java/org/springframework/asm/FieldWriter.java
@@ -278,7 +278,7 @@ final class FieldWriter extends FieldVisitor {
    *
    * @param attributePrototypes a set of attribute prototypes.
    */
-  final void collectAttributePrototypes(final Attribute.Set attributePrototypes) {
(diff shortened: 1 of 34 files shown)


snicoll:
@The-Gamer-01 thanks for the PR, but we prefer to do such stylistic change ourselves as they are largely subjective, especially the ones that are applied automatically by a tool.

If you're looking to contribute, my recommendation would be to do that based on the area of the framework that you're using. You might find a typo in the doc/javadoc or something that could be phrased better. Or you might find a limitation in the code based on your use cases. Thanks again for the PR, in any case!

The-Gamer-01:
> @The-Gamer-01 thanks for the PR, but we prefer to do such stylistic change ourselves as they are largely subjective, especially the ones that are applied automatically by a tool.
> 
> If you're looking to contribute, my recommendation would be to do that based on the area of the framework that you're using. You might find a typo in the doc/javadoc or something that could be phrased better. Or you might find a limitation in the code based on your use cases. Thanks again for the PR, in any case!

OK，thanks

PR #31977, opened 2024-01-08
Add missing preposition in reference guide

SuhasBk:
Adding a missing preposition - "to" in WebSocket Token Authentication doc.


snicoll:
Good catch @SuhasBk and thank you for making your first contribution to Spring Framework.

PR #31981, opened 2024-01-08
Update to Commons IO 2.15.1

wfouche:


bclozel:
Thanks but we usually maintenance upgrades in bulk, especially if they don't need code changes. We compile against those versions and unlike Spring Boot, we don't provide any dependency management for it.

PR #31986, opened 2024-01-09
Add `spring.context.expression.maxLength` property

### LINKED ISSUE #31952: Make maximum SpEL expression length configurable for `StandardBeanExpressionResolver`
**Affects:** Spring Framework 5.2.24+

### Background

Since #30325 (implemented in b73f5fcac22555f844cf27a7eeb876cb9d7f7f7e) the length of SpEL expressions is limited by default to 10000. As I understand it this is a mitigation against potential ReDoS exploits. However, in some cases this limitation is too low and prevents upgrading to recent Spring Framework versions.

While #30380 (implemented in aefcb9d2d6ee385c14d83852bee50b0307b42ce4) adds support for a custom `maximumExpressionLength` the feature is only accessible if one instantiates the `SpelParserConfiguration` class themselves.

In my case I would like to configure the `SpelParserConfiguration` created in the class `StandardBeanExpressionResolver` to accept my very long property by raising the `maximumExpressionLength` to a higher value than its default (10000).

https://github.com/spring-projects/spring-framework/blob/07097976efe954d33d4d7104709e06a614e4fc3f/spring-context/src/main/java/org/springframework/context/expression/StandardBeanExpressionResolver.java#L91-L105

### Use case

I've got a huge map in my config:

```properties
myproperty={\
  a: {\
    x: { host: '10.1.1.1', port: 1234 },\
    y: { host: '10.1.1.1', port: 1234 },\
    z: { host: '10.1.1.1', port: 1234 }\
  },\
  b: {\
    x: { host: '10.1.1.1', port: 1234 },\
    y: { host: '10.1.1.1', port: 1234 },\
    z: { host: '10.1.1.1', port: 1234 }\
  },\
  c: {\
    x: { host: '10.1.1.1', port: 1234 },\
    y: { host: '10.1.1.1', port: 1234 },\
    z: { host: '10.1.1.1', port: 1234 }\
  },\
  # and so on, altogether 15000 characters
}
```

It is used by a property:

```
@Value("#{${myproperty}}")
private Map<String, Map<String,Map<String,String>>> myproperty;
```

If I try to start my application I get the following exception:

```
org.springframework.expression.spel.SpelEvaluationException: EL1079E: SpEL expression is too long, exceeding the threshold of '10,000' characters"}}
```

### Proposal

Make the parameter `maximumExpressionLength` of `SpelParserConfiguration` configurable when it is instantiated in `StandardBeanExpressionResolver.java` (see the snippet above). Example (not sure what a conformant property name would be):

```properties
spring.standardBeanExpressionResolver.maximumExpressionLength=20000
```

bencehornak:
Fixes #31952

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/expression/StandardBeanExpressionResolver.java
+++ b/spring-context/src/main/java/org/springframework/context/expression/StandardBeanExpressionResolver.java
@@ -23,11 +23,13 @@ import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanExpressionException;
 import org.springframework.beans.factory.config.BeanExpressionContext;
 import org.springframework.beans.factory.config.BeanExpressionResolver;
+import org.springframework.core.SpringProperties;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.expression.Expression;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.ParserContext;
+import org.springframework.expression.spel.SpelCompilerMode;
 import org.springframework.expression.spel.SpelParserConfiguration;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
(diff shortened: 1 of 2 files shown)


sbrannen:
Thanks for the PR, @bencehornak.

Unfortunately, I've already done the work.

In the future, you might want to ask before spending your time on a PR.

sbrannen:
In case you're interested, feel free to check out what I committed in https://github.com/spring-projects/spring-framework/commit/785598629abda944343a02307ad82a79bb31b589 to compare with your work.

There are certainly some similarities. 😉

PR #31991, opened 2024-01-09
Document exception handling for HTTP Interface client with `RestClient` and `RestTemplate`

### LINKED ISSUE #31782: HTTP Interface Exception Handling section only mentions WebClient
**Affects:**  6.1.1
---

In the [exception handling](https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-http-interface-exceptions) section, I had expected to see an example using the newly added `RestClient`:

```java
RestClient restClient = RestClient.builder()
    .defaultStatusHandler(HttpStatusCode::isError, resp -> ...)
    .build();
```

Or maybe even a small callout that the same setup/arrangement exists for `RestClient`.

The introduction section before it gives example for each client as well:

image

OlgaMaciaszek:
Fixes gh-31782.

### DIFF
--- a/framework-docs/modules/ROOT/pages/integration/rest-clients.adoc
+++ b/framework-docs/modules/ROOT/pages/integration/rest-clients.adoc
@@ -1079,10 +1079,30 @@ underlying HTTP client, which operates at a lower level and provides more contro
 
 [[rest-http-interface-exceptions]]
 === Exception Handling
+In order to provide a custom way of handling errors, you can register response
+status handlers on the underlying HTTP clients.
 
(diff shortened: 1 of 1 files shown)


rstoyanchev (changes requested):
Thanks for the changes.

Could we structure this more like a bullet list as it is in the beginning of the HTTP Interface client section, and also follow the same order? In other words:

For `RestClient`:

_name exception raised by default for 4xx and 5xx..._

... source code snippet...

_reference to Javadoc for more details_

For `WebClient`:

... same ...

For `RestTemplate`:

... same ...

I think that would make it more readable and easier to focus on the relevant part.

Issue #29148, opened 2022-09-13
DefaultWebClient logging sensitive information in URI


itsnagaraj:
**Affects:** v5.3.22 and prior. Module is spring-webflux

The `insertCheckpoint` method in `DefaultWebClient` seems to log the URI. In few scenarios this URI can have sensitive information like email or phone number or name etc.,. The framework doesn't provide any controls to switch it off and the only way I can think of right now is to create a duplicate of this class in our classpath and override this method. Any other suggestions are appreciated.? If future versions of spring-webflux can provide a way to override the behaviour that will provide us a strategic alternative.

```
private <T> Mono<T> insertCheckpoint(Mono<T> result, HttpStatusCode statusCode, HttpRequest request) {
	HttpMethod httpMethod = request.getMethod();
	URI uri = request.getURI();
	String description = statusCode + " from " + httpMethod + " " + uri + " [DefaultWebClient]";
       return result.checkpoint(description);
}

PR #31992, opened 2024-01-09
Exclude URI query from remaining WebClient checkpoints

aaronrosser:
**Affects:** latest (v6.1.2)
**Module:** spring-webflux

Continuing from this issue https://github.com/spring-projects/spring-framework/issues/29148 and its fix https://github.com/spring-projects/spring-framework/commit/f9d8367379371f52887a210f1bc8ff8682d596e2 regarding full request URIs including sensitive query parameters being included in mono checkpoints.

This PR sanitises URIs included in 
- the mono checkpoint in `DefaultWebClient.DefaultRequestBodyUriSpec::exchange`
- the message of `WebClientResponseException`s created via `WebClientResponseException::create`

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientResponseException.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientResponseException.java
@@ -16,6 +16,8 @@
 
 package org.springframework.web.reactive.function.client;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientResponseException.java
```diff
@@ -99,7 +102,19 @@ public WebClientResponseException(
 
 	private static String initMessage(HttpStatusCode status, String reasonPhrase, @Nullable HttpRequest request) {
 		return status.value() + " " + reasonPhrase +
-				(request != null ? " from " + request.getMethod() + " " + request.getURI() : "");
+				(request != null ? " from " + request.getMethod() + " " + getUriToLog(request.getURI()) : "");
+	}
+
+	private static URI getUriToLog(URI uri) {
```
aaronrosser, line 108: Maybe this method should go in `WebClientUtils`? 🤔

rstoyanchev:
Thanks for the pull request. I added a method to `WebClientUtils` and used it in all places of logging, including one additional one in `DefaultClientResponse`.

aaronrosser:
> Thanks for the pull request. I added a method to WebClientUtils and used it in all places of logging, including one additional one in DefaultClientResponse.

That was quick. Thank you 😀

PR #32008, opened 2024-01-10
Update docs on server side use of `@HttpExchange` to make it more clear when it is a good fit

OlgaMaciaszek:


rstoyanchev (changes requested):
Let's have the interface examples replace the controller ones altogether, and add an explanation of the use case along the lines of https://github.com/spring-projects/spring-framework/issues/30913#issue-1811415968.

PR #32009, opened 2024-01-10
Fix `DefaultDataBuffer#getNativeBuffer()` to set correct limit

### LINKED ISSUE #30967: The method getNativeBuffer() in DefaultDataBuffer returns misconfigured ByteBuffer
**Affects:** 6.0.11

---

The class `DefaultDataBuffer` has method `getNativeBuffer` with follow implementation:

```java
	public ByteBuffer getNativeBuffer() {
		this.byteBuffer.position(this.readPosition);
		this.byteBuffer.limit(readableByteCount());
		return this.byteBuffer;
	}

        @Override
	public int readableByteCount() {
		return this.writePosition - this.readPosition;
	}
```

The limit of the byte buffer to return should be set to the `this.writePosition`, because if the condition `(this.writePosition - this.readPosition) <= this.readPosition` is satisfied, the limit of the returned byte buffer will be equal to its current position, so no single byte could be read from this buffer, even if the condition `this.writePosition > this.readPosition` is satisfied. So how it should be:

```java
	public ByteBuffer getNativeBuffer() {
		return this.byteBuffer.limit(this.writePosition).position(this.readPosition);
	}
```

injae-kim:
Closes gh-30967

### Motivation
```java
// DefaultDataBuffer
	public ByteBuffer getNativeBuffer() {
		this.byteBuffer.position(this.readPosition);
		this.byteBuffer.limit(readableByteCount()); // 👈 limit should be writePosition, not readableByteCount
		return this.byteBuffer;
	}
```
- We found that `DefaultDataBuffer#getNativeBuffer` set `byteBuffer#limit` as `readableByteCount()` incorrectly so fix to `writePosition`

### Modification
- Fix `DefaultDataBuffer#getNativeBuffer()` to set correct limit as `writePosition` instead of `readableByteCount()`

### Result
- Now `DefaultDataBuffer#getNativeBuffer()` return native bytebuffer with correct limit
- Close #30967

### DIFF
--- a/spring-core/src/test/java/org/springframework/core/io/buffer/DefaultDataBufferTests.java
+++ b/spring-core/src/test/java/org/springframework/core/io/buffer/DefaultDataBufferTests.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.io.buffer;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


spring-core/src/test/java/org/springframework/core/io/buffer/DefaultDataBufferTests.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author Injae Kim
+ * @since 6.2
+ */
+class DefaultDataBufferTests {
+
+	private final DefaultDataBufferFactory bufferFactory = new DefaultDataBufferFactory();
+
+	@Test // gh-30967
+	void getNativeBuffer() {
```
injae-kim, line 38: image

FYI) with current logic, this test failed~!

injae-kim:
Oh thank you for your additional polishing :)

I'm so happy that this is my first contribution on spring-framework!
I'll continue to contribute by another PR 🚀

PR #32015, opened 2024-01-11
Upgrade to JMH 1.37

wfouche:
Gradle plugin me.champeau.jmh 0.7.2 uses JMH 1.37


wfouche:
Dependencies report generated with command

`./gradlew dependencyUpdates`

to report file

`build/dependencyUpdates/report.txt`

which indicated that `jmh-generator-bytecode` needed to be at 1.37 as well (therefore, it was added).

Issue #28491, opened 2022-05-20
`@HttpExchange` for default methods support


wonwoo:
```java
@HttpExchange("http://localhost:8080")
public interface GreetingClient {

    @GetExchange("/greeting")
    Flux<Greeting> greetings();

    @PostExchange("/greeting")
    Mono<Greeting> greetings(@RequestBody Greeting greeting);

    default Flux<Greeting> defaultMethod() {
        return greetings(new Greeting("default method test"))
                .flatMapMany(it -> greetings());
    }
}
```

defaultMethod method I hope it works!

PR #32018, opened 2024-01-12
Add invokeDefaultMethod utils for JDK dynamic proxy created by Spring AOP

### LINKED ISSUE #28495: Invoke default methods in dynamic proxies automatically
## Overview

See #28491 for background information.

In JDK dynamic proxies created by Spring AOP, we could potentially introduce support for automatically invoking default methods (or make it easier via a utility).

## Available Java APIs

- `java.lang.reflect.Method.isDefault()`
- `java.lang.reflect.InvocationHandler.invokeDefault(Object, Method, Object...)`

injae-kim:
Closes gh-28495

### Motivation
```java
@HttpExchange("http://localhost:8080")
public interface GreetingClient {

    @GetExchange("/greeting")
    Flux<Greeting> greetings();

    default Flux<Greeting> defaultMethod() {
        return greetings(new Greeting("default method test"))
                .flatMapMany(it -> greetings());
    }
}
```
- Like https://github.com/spring-projects/spring-framework/issues/28491, we may need to support default method on more proxies

> In JDK dynamic proxies created by Spring AOP, .. we cound **make it(support default method) easier via a utility.**

- Based on https://github.com/spring-projects/spring-framework/issues/28495#issue-1243003876, I want to introduce `ReflectiveMethodInvocationUtils#invokeDefaultMethod` first to support default method easily via a utility and reduce duplicated code!

### Modification
- Add `ReflectiveMethodInvocationUtils#invokeDefaultMethod`
- Replace duplicated default method invocation code on `HttpServiceProxyFactory` and `RSocketServiceProxyFactory`

### Result
- Now we can easily support default method via a utiliy in JDK dynamic proxies created by Spring AOP
- Close gh-28495

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/ReflectiveMethodInvocationUtils.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/ReflectiveMethodInvocationUtils.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


jhoeller:
On review, I don't really see a need for extracting common logic there since it is a pretty straight invocation anyway, just a condition and a cast but then being able to return straight away.

From where we stand right now, #28495 is rather about where else to invoke default methods the same way. If we identify such further places, I don't mind copying the same condition logic there.

injae-kim:
> rather about where else to invoke default methods the same way. If we identify such further places, I don't mind copying the same condition logic there.

Aha~ I understood. I think #28495 means extracting common logic 😅

Thanks you!

PR #32031, opened 2024-01-15
Implementation of a spring HttpHandler directly from a Jetty core Handler

gregw:
Implement a Jetty core Handler that uses a spring HttpHandler, without the overhead of servlets and using reactive async abstraction.


gregw:
Ooops. Sorry this is too early for this PR.   I will close and make a new one when this is vaguely workable....

gregw:
If there is any interest.... look at https://github.com/webtide/spring-framework/pull/1 for this being developed.

PR #32034, opened 2024-01-15
Fix `RestClient` to use defaultRequest on `RestClient.Builder`

### LINKED ISSUE #32028: `DefaultRestClientBuilder.defaultRequest(...)` doesn't do anything
`DefaultRestClientBuilder.defaultClient(Consumer<RestClient.RequestHeadersSpec<?>>)` can be used to set a property, but that property is never actually used when building the `RestClient`.

Is this related to #31625?

injae-kim:
Closes gh-32028

### Motivation
- On #32028 , we found that `DefaultRestClientBuilder.defaultRequest(...)` doesn't do anything
  - Cause `RestClientBuilder.build()` doesn't pass `defaultRequest` to `RestClient`

### Modification
- Fix `RestClient` to use defaultRequest on `RestClient.Builder`

### Result
- Now `RestClient` uses `defaultRequest` that passed from `RestClientBuilder.build()` well

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
```diff
@@ -452,6 +458,9 @@ private <T> T exchangeInternal(ExchangeFunction<T> exchangeFunction, boolean clo
 			URI uri = null;
 			try {
 				uri = initUri();
+				if (defaultRequest != null) {
+					defaultRequest.accept(this);
+				}
```
injae-kim, line 463: https://github.com/spring-projects/spring-framework/blob/faba044735ad173d96b4b1adab62752b48b2c758/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L481-L484

FYI) almost same usage of `defaultRequest` on `WebClient`~!

PR #32064, opened 2024-01-19
Allow for AbstractUrlHandlerMapping to add/remove handlers at runtime

### LINKED ISSUE #26798: Allow for AbstractUrlHandlerMapping to add/remove handlers at runtime
The requirements is to be able to add and remove mappings for urls at runtime.

It is possible right now with the `AbstractHandlerMethodMapping` and its `registerMapping(T mapping, Object handler, Method method)` and `unregisterMapping(T mapping)` API.

It would be great if `AbstractUrlHandlerMapping` would make its `protected void registerHandler(String urlPath, Object handler) throws BeansException, IllegalStateException {` as `public` and also would introduce an `unregisterHalder(String urlPath)`.

Related Spring Integration issue: https://github.com/spring-projects/spring-integration/issues/3533.

Right I'm enforced to copy/paste a lot from Spring MVC to have a backward compatibility and implement a requested feature.

Thanks

awmeinema:
Closes  #26798

Made `void registerHandler(String urlPath, Object handler) throws BeansException, IllegalStateException` public and introduced a public `unregisterHalder(String urlPath)`.

Makes possible to add and remove handlers at runtime.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractUrlHandlerMapping.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractUrlHandlerMapping.java
@@ -403,13 +403,14 @@ public abstract class AbstractUrlHandlerMapping extends AbstractHandlerMapping i
 
 	/**
 	 * Register the specified handler for the given URL path.
+	 * <p>This method may be invoked at runtime after initialization has completed.
 	 * @param urlPath the URL the bean should be mapped to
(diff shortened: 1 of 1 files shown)


snicoll:
@awmeinema thanks for making your first contribution to Spring Framework. I've polished things up a bit and added tests in [ad0c488](https://github.com/spring-projects/spring-framework/commit/ad0c4887670ee130f631af91343df43eff8b388b)

PR #32070, opened 2024-01-20
Replace with Java Style array declaration

mnhock:


snicoll:
This is a particular part of the codebase here (our fork of cglib). We generally avoid doing polish in that area of the code as we modify only what we need from that external project. Thanks anyway for the PR.

PR #32071, opened 2024-01-20
Remove unnecessary conversion to 'String'

mnhock:


snicoll:
@mnhock thank you for making your first contribution to Spring Framework.

snicoll:
@mnhock Juergen made me aware that this isn't unnecessary actually as it account for writing `null` if the value is `null`, which would throw an exception now. I did miss that as the method parameter is not `@Nullable`. Turns out it's old code that's called in many places. Thanks for the suggestion, in any case.

PR #32073, opened 2024-01-20
Single character string argument in 'String.indexOf()' call

mnhock:


snicoll:
Thanks for the suggestion but I think the code is fine a it is.

PR #32078, opened 2024-01-22
Cache validationGroups for subsequent requests

### LINKED ISSUE #32068: Avoid reflecting to obtain `validationGroups`  every time the call is made
https://github.com/spring-projects/spring-framework/blob/1a52c56bd42c1a9ba8cc8f8849d5f5cb742a97e5/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java#L241-L245

can call ``getValidationGroups()`` in the `InvocableHandlerMethod` construction method, and take `validationGroups` returned by ``getValidationGroups()`` method as a member variable. Avoid reflecting to obtain `validationGroups`  every time the call is made

quaff:
Close GH-32068

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java
+++ b/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


brucelwl:
@quaff Would it be better to initialize `validationGroups` in method `setMethodValidator(@Nullable MethodValidator methodValidator)` ?

example code:
```
public void setMethodValidator(@Nullable MethodValidator methodValidator) {
   this.methodValidator = methodValidator;
   this.validationGroups = getValidationGroups();
}
```

jhoeller:
@quaff I eventually went with a non-lazy-init approach as suggested by @brucewl, dropping the `getValidationGroups()` method completely. Also, applying the same change to the reactive `InvocableHandlerMethod` variant as well.

Issue #32067, opened 2024-01-19
Reference Doc URL is Not Found


BykaWF:
[Screencast from 01-19-2024 07:04:12 PM.webm](https://github.com/spring-projects/spring-framework/assets/119706327/3b0e0e35-64f5-4cf5-a23f-dc9bb61c2aa0)

But this link works fine https://docs.spring.io/spring-framework/reference/index.html

NilavroSeal2000:
In the web page Reference doc is redirecting to a different link [https://docs.spring.io/spring-framework/reference/6.1/index.html](url) 
The actual link for the document is [https://docs.spring.io/spring-framework/reference/index.html](url)

snicoll:
@sdeleuze https://docs.spring.io/spring-framework/reference/6.1 does not work anymore. Could it be related to a recent change?

jcw1031:
[https://docs.spring.io/spring-framework/reference/](https://docs.spring.io/spring-framework/reference/)

sdeleuze:
Should be fixed by https://github.com/spring-projects/spring-framework/pull/32083.

PR #32083, opened 2024-01-22
Fixes atlas-extension order

rwinch:
The atlas-extension must be registered before the latest-version-extension so that the latest version logic is applied to versions imported from the atlas-extension.

Closes gh-32067


rwinch:
Note that this was only recently exposed because of https://github.com/spring-projects/spring-framework/pull/32066 which ensures that .htaccess was published on single branch builds (not just a full build)

PR #32088, opened 2024-01-23
Use count in ParamsRequestCondition#getValueMatchCount

Ryan-Dia:
The `getValueMatchCount` method was refactored to use a `Stream` instead of a `for-each loop`, eliminating the need for the unnecessary local variable `count`. This change was made to reduce the risk of accidental modifications during future maintenance, as the implementation is possible without using a local variable. Additionally, the refactoring was done for enhanced stability and maintainability, especially since there was no need to interrupt the loop under specific conditions.

- [x]  The test code was executed and passed successfully.


snicoll:
@Ryan-Dia congratulations for making your first contribution to Spring Framework.

snicoll:
@Ryan-Dia the team has discussed this change today and we've decided to revert it. I had missed that the method you've changed is in a hot code path and we explicitly avoid using the Stream API in such places.

Ryan-Dia:
Thank you for reviewing the changes I proposed. I understand the decision to revert them, recognizing that the modified method is in a hot code path. I appreciate the feedback and will ensure my future contributions are more closely aligned with these best practices

PR #32089, opened 2024-01-23
Fix getMostSpecificMethod to return original method if target class is not in hierarchy

### LINKED ISSUE #32087: Parameter based cache keys on repository methods can't resolve after 6.1.2->6.1.3
**Affects:** 6.1.3

A change was made in 6.1.3 to the AopUtils.getMostSpecificMethod and BridgeMethodResolver.getMostSpecificMethod that impacts how cache keys get resolved on Spring Data repositories.

In my setup, I have a repository which redefines the save(<T>) method, so that I can put a @CachePut annotation on top. The key for this particular cache is defined using the method's parameter.

```
@Repository
public interface ProfileRepository extends JpaRepository<Profile, Integer> {
	
	@CachePut(value="ProfileRepoByCode", key="#profile.getCode()")
	Profile save(Profile profile);
}
```
In 6.1.2, when CacheOperationMetadata.targetMethod was set, the method reference would be to the custom repository's method. In 6.1.3, this now points to the SimpleJpaRepository.save method. SimpleJpaRepository.save and I use different parameter names for the object being saved, which results in MethodBasedEvaluationContext.lazyLoadArguments no longer being able to find a match when evaluating the SpEL expression.

The attached code works when Spring Boot 3.2.1 (Framework 6.1.2) is used, but fails when upgraded to Spring Boot 3.2.2 (Framework 6.1.3).

[cacheBridgeMethod-1.zip](https://github.com/spring-projects/spring-framework/files/14017775/cacheBridgeMethod-1.zip)
[Framework_6_1_2.txt](https://github.com/spring-projects/spring-framework/files/14017850/Framework_6_1_2.txt)
[Framework_6_1_3.txt](https://github.com/spring-projects/spring-framework/files/14017851/Framework_6_1_3.txt)

quaff:
Closes GH-32087

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/BridgeMethodResolver.java
+++ b/spring-core/src/main/java/org/springframework/core/BridgeMethodResolver.java
@@ -46,6 +46,7 @@ import org.springframework.util.ReflectionUtils.MethodFilter;
  * @author Rob Harrop
  * @author Juergen Hoeller
  * @author Phillip Webb
+ * @author Yanming Zhou
  * @since 2.0
  */
 public final class BridgeMethodResolver {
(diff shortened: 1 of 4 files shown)


jhoeller:
@quaff thanks for the PR which I'm mostly using for my fix for #32087. However, I'm avoiding the extra constraints (and extra overhead) in `ClassUtils.getMostSpecificMethod` since that is being called in rather expensive code paths. The check in `BridgeMethodResolver` should be sufficient for our purposes here.

quaff:
> @quaff thanks for the PR which I'm mostly using for my fix for #32087. However, I'm avoiding the extra constraints (and extra overhead) in `ClassUtils.getMostSpecificMethod` since that is being called in rather expensive code paths. The check in `BridgeMethodResolver` should be sufficient for our purposes here.

It's better to amend the commit to revert `ClassUtils.getMostSpecificMethod` part, contributors like be the author.

PR #32094, opened 2024-01-23
Remove unnecessary else statements in AbstractMediaTypeExpression

Ryan-Dia:
In this PR, I've refactored the `compareTo` method in the `AbstractMediaTypeExpression` class. The main change involves simplifying the conditional structure by removing the unnecessary `else` statements. This enhances readability and align the code style with the `equals` method

**Changes Made:**
- Removed unnecessary `else` statements in the `compareTo` method.
- Retained the original logic and functionality.


snicoll:
Thanks for the PR but things like this are largely subjective and we don't really accept PRs like this. In any case, I think that the current code reads better anyway.

PR #32095, opened 2024-01-23
Remove unnecessary else statements in RequestMethodsRequestCondition

Ryan-Dia:
The main change is the simplification of the conditional structure by removing unnecessary else statements.

**Changes Made:**
- Removed redundant validations
- Removed unnecessary `else` statements 
- Retained the original logic and functionality.

- [x] The test code was executed and passed successfully.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/condition/RequestMethodsRequestCondition.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/condition/RequestMethodsRequestCondition.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


snicoll:
I don't think the else here is unnecessary.  It is now because there's a `return` but the algorithm may change in the future so the code as it is is more robust IMO. Thanks for the PR, in any case.

Ryan-Dia:
@snicoll 
>It is now because there's a return but the algorithm may change in the future so the code as it is is more robust IMO.

I agree with this perspective.  However, I'm curious about your thoughts on modifying the nested logic within the if statement.

📌 [Reference](https://github.com/spring-projects/spring-framework/blob/d7778c0212d0e23b86fcd4585d749307c89c5bef/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/condition/ParamsRequestCondition.java#L93)

before
```java
if (isEmpty() && other.isEmpty()) {
	  return this;
  }
  else if (other.isEmpty()) {
	  return this;
  }
  else if (isEmpty()) {
	  return other;
  }
```		

after
```js
if (other.isEmpty()) {
	  return this;
  }
  else if (isEmpty()) {
	  return other;
  }
```

PR #32126, opened 2024-01-26
Skip overriding primary bean definition with non-primary one

### LINKED ISSUE #32125: Skip overriding primary bean definition with non-primary one
Here is my case, I defined a default bean in xml configuration, sometimes I want override that bean with another implementation via `@Component` and  `@Primary`, it's fine if injection is autowired by type but not name, the ideal solution is pinning the primary bean definition make it cannot be overridden by non-primary one.

Here is test case:
```java
package test;

import org.springframework.test.context.ContextConfiguration;

@SpringJUnitConfig
@ContextConfiguration(locations = "ctx.xml")
public class OverrideBeanTests {

	@Autowired
	private Service service;

	@Test
	void test() {
		assertThat(service).isInstanceOf(CustomService.class);
	}

}
```
```java
package test;

public class Service {

}
```
```java
package test;

import org.springframework.context.annotation.Primary;
import org.springframework.stereotype.Component;

@Primary
@Component("service")
public class CustomService extends Service {

}

```
```xml
<beans default-autowire="byName"
	xmlns="http://www.springframework.org/schema/beans"
	xmlns:context="http://www.springframework.org/schema/context"
	https://www.springframework.org/schema/beans/spring-beans.xsd
	http://www.springframework.org/schema/context
	https://www.springframework.org/schema/context/spring-context.xsd">
	<context:component-scan base-package="test"/>
	<bean id="service" class="test.Service"/>
</beans>
```

quaff:
Fix GH-32125

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


snicoll:
Going forward, please do not create both an issue and a PR as one supersedes the other. As the original issue has been declined, I am closing this as well, see https://github.com/spring-projects/spring-framework/issues/32125#issuecomment-1911787730

PR #32134, opened 2024-01-26
Support multiple `@RequestMapping` annotations

youssef3wi:
Closes #32043


sbrannen:
Hi @youssef3wi,

Unfortunately, this PR introduces breaking changes by renaming protected methods which are effectively part of the public API.

In addition, I have added the "pending-design-work" label to #32043 to make it clearer that the team wishes to "investigate the feasibility of supporting multiple `@RequestMapping` annotations on a single handler method" before. One of the reasons we chose that wording in the issue's description is that we are aware of the fact that existing methods are based on the assumption that there is a single annotation mapping per method.

If you have ideas, feel free to share them on the issue to begin a discussion.

Thanks!

sbrannen:
As a side note, any PR that wishes to address this topic would need to address it not only in WebFlux but also in Web MVC.

PR #32145, opened 2024-01-28
Update basics.adoc

1515biubiubiu:
According to the context content, whether the jump here should point to the resources.adoc file.


sdeleuze:

PR #32150, opened 2024-01-29
Remove identical method declaration from sub interface

quaff:


snicoll:
`TaskExecutor` predates `Executor` so what you see is what we'd like to keep.

PR #32164, opened 2024-01-31
Polish `StatementCreatorUtilsTests`

quaff:
1. mark field `preparedStatement` as final
2. avoid passing `null` to `@NotNull` parameter


snicoll:
Thanks @quaff, I've reverted 1. I agree with the reasoning but we're not consistent at the moment so going in one direction without a clear understanding of what the expectation is is not a good idea.

PR #32172, opened 2024-01-31
Clarify UnknownContentTypeException

michaldo:
UnknownContentTypeException message should contain information important for analyze error. 

Before the PR message contains only response content type. After the PR message includes also status code.

Maybe response content should be also included. It would be useful for analyzing error, but on the other hand content may be very large.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/UnknownContentTypeException.java
+++ b/spring-web/src/main/java/org/springframework/web/client/UnknownContentTypeException.java
@@ -79,7 +79,7 @@ public class UnknownContentTypeException extends RestClientException {
 			HttpStatusCode statusCode, String statusText, HttpHeaders responseHeaders, byte[] responseBody) {
 
 		super("Could not extract response: no suitable HttpMessageConverter found " +
-				"for response type [" + targetType + "] and content type [" + contentType + "]");
+				"for type [" + targetType + "] and response: code [" + statusCode.value() + "], content type [" + contentType + "]");
 
 		this.targetType = targetType;
 		this.contentType = contentType;


snicoll:
> UnknownContentTypeException message should contain information important for analyze error.

It does already IMO and the exception contains far more than the status code. 

> Maybe response content should be also included. It would be useful for analyzing error, but on the other hand content may be very large.

I think you got why things are the way they are. The exception contains the information for you to look at, but relying on logging message only to understand what was wrong in the setup of the application is not a good idea.

michaldo:
> relying on logging message only to understand what was wrong in the setup of the application is not a good idea.

When I investigate rest client error, I want one: I want see response. For performance and cost all responses are not logged - that is my experience in many projects/companies.

When I investigate rest client error, I DO rely on logged message. That is often the only evidence - all others may gone.
Usually it works, because error message contains response body and I can reproduce what failed in isolation. See `org.springframework.web.client.DefaultResponseErrorHandler#getErrorMessage`

UnknownContentTypeException means that response content type is different than I expected. They are two reasons:
1. wrong or missing request header `Accept`
2. request header `Accept` correct but response is not compatible.

Reason 1 I can easy check: construct request locally and inspect header.
To analyze reason 2 I need more information about response in exception message

bclozel:
If 1) the request body can be very large and 2) we don't know what do to do with this content-type - then it means we have no reliable way to parse or print the body in logs (it could be binary for all we know). I understand that it makes things more difficult to investigate but I can't find a way to reliably improve the exception.

michaldo:
> body(..)  could be binary for all we know

Here you don't care binary: https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/client/DefaultResponseErrorHandler.java#L157

I understand your concern about adding body to exception message. But in the PR I just add response status code

PR #32173, opened 2024-02-01
Change FlushGuardedOutputStream to override write() methods properly

testn:
FlushGuardedOutputStream does not override write(byte[]) and write(byte[], int, int) method. Therefore, the previous code calls the delegate's write(int) which performs a lot of checks on buffer length while creating a lot of garbage. By overriding write(byte[]) and write(byte[], int, int) help improving performance by calling into the respective methods directly


snicoll:
Thanks for the PR @testn but this code is deprecated and removed in the current line so we're not going to make further changes to it.

PR #32189, opened 2024-02-02
Optimize Map methods in ServletAttributesMap

kilink:
ServletAttributesMap inherited default implementations of the size and isEmpty methods from AbstractMap which delegates to the Set returned by entrySet. ServletAttributesMap's entrySet method made this fairly expensive, since it would copy the attributes to a List, then use a Stream to build the Set. To avoid the cost, add implementations of isEmpty / size that don't need to call entrySet at all. Additionally, change entrySet to return a Set view that simply lazily delegates to the underlying servlet request for iteration.


bclozel:
Thanks @kilink for your contribution, this has been merged in main and will be backported to 6.0.x as well in #32197

PR #32215, opened 2024-02-07
Initialize (another) Map with correct size in RequestPredicates

kilink:
Fix another instance where a LinkedHashMap was initialized with an initial capacity that would always cause a resize / rehash to occur. Switch to CollectionUtils.newLinkedHashMap to size the map appropiately for the expected number of items.

See d5cb1d9ad, this is just another case I missed.


bclozel:
Thanks @kilink ! I missed that one too while reviewing the previous PR.

PR #32233, opened 2024-02-09
Fix issue when the allowedOrigin list contains strings with spaces

SecurityRisk:
### Issue
When the `List<String> allowedOrigins` contains Strings that have spaces, it will cause the CORS comparison checks with the Origin header to not be equal

### Resolution
By trimming the values when they get inserted into `allowedOrigins`, the comparison checks will pass. Since we trim the spaces out of the comma delimited String, we might as well do it for the `List<>` too. There would be no reason that a URL would contain a space before or after the string

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
+++ b/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
@@ -183,7 +183,7 @@ public class CorsConfiguration {
 			setAllowedOrigins(DEFAULT_PERMIT_ALL);
 		}
 		parseCommaDelimitedOrigin(origin, value -> {
-			value = trimTrailingSlash(value);
+			value = trimTrailingSlash(value).trim();
 			this.allowedOrigins.add(value);
 		});
 	}


sdeleuze:
It makes sense to me to trim comma separated values to handle correctly `origin1, origin2`, but I am unsure about the use case that require triming individual list elements. Could you please provide more details on your use case?

SecurityRisk:
So, I know this is improper from yaml to have spaces in a property, but imagine you create a property in the application.properties and use that property in an `@Value("#{'allowedOrigins'.split(',')})` with the property being:
`allowedOrigins=http://localhost:8080, http://localhost:8000`

While yes, you shouldn't have spaces in a property in yaml, this sort of setup will result in only the first origin being allowed. In order to resolve this, the property needs to be changed or the split regex needs to change to `'\\s*,\\s*'`. However, from a user perspective, no obvious errors happen with this configuration. The user will keep getting CORS errors for the subsequent Origins and not really understand why. So to me, I would think always trimming the List to be accurate before any equality checking

sdeleuze:
Thanks for providing more details. I am not convinced Spring Framework should workaround such misuse, so I prefer decline this change proposal.

PR #32236, opened 2024-02-10
Fixes syntax error in JdbcClient examples

spencergibb:
Looks like a copy-paste error from the JdbcTemplate examples where the SQL is the first parameter, rather than the fluent style of the new JdbcClient.


snicoll:
Thanks! I've found a few more that I fixed in [b1f6401](https://github.com/spring-projects/spring-framework/commit/b1f6401e4f02ead72958ea1dcc242451ce12fd2d)

Issue #32234, opened 2024-02-09
UriComponentsBuilder double encodes query parameters


bahrigencsoy:
**Affects:** Tested on 4,5,6.

`URIComponentsBuilder` double encodes the request parameter when `replaceQueryParam` is called. This pattern is used in [spring-cloud-gateway](https://github.com/spring-cloud/spring-cloud-gateway/blob/v4.1.0/spring-cloud-gateway-server/src/main/java/org/springframework/cloud/gateway/filter/factory/RewriteRequestParameterGatewayFilterFactory.java#L63)

Here is an accompanying unit test for `UriComponentsBuilderTests.java` class:

```java
@Test
void doubleEncodingOfQueryParamReplacement() throws Exception {
  String url = "http://localhost:8080/app/api/device/queues?devices=100%2C200%2C300%2C400&tenant=root";
  URI uri = new URI(url);

  UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromUri(uri);
  uriComponentsBuilder.replaceQueryParam("tenant", "default");

  URI replaced = uriComponentsBuilder.build().toUri();
  assertThat(replaced.toString()).isEqualTo("http://localhost:8080/app/api/device/queues?devices=100%2C200%2C300%2C400&tenant=default");
  // ^Fails with "http://localhost:8080/app/api/device/queues?devices=100%252C200%252C300%252C400&tenant=default"
}
```

VenkatramanMuthukrishnan:
https://github.com/spring-projects/spring-framework/pull/32237

rstoyanchev:
From the Javadoc of `replaceQueryParam`:

> please, review the Javadoc of [queryParam(String, Object...)](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriBuilder.html#queryParam(java.lang.String,java.lang.Object...)) for further notes on the treatment and encoding of individual query parameters.

That in turn says:

[...]

PR #32237, opened 2024-02-10
Fix double encoding of query parameters in UriComponentsBuilder (#32234)

VenkatramanMuthukrishnan:


spring-web/src/main/java/org/springframework/web/util/HierarchicalUriComponents.java
```diff
@@ -638,7 +638,7 @@ public boolean isAllowed(int c) {
 		QUERY_PARAM {
 			@Override
 			public boolean isAllowed(int c) {
-				if ('=' == c || '&' == c) {
+				if ('=' == c || '&' == c || ',' == c ) {
```
VenkatramanMuthukrishnan, line 641: Added comma as a exclusion as it was skipping in query-param encode

rstoyanchev:
Closing along with the issue. Thanks for the pull request in any case.

PR #32248, opened 2024-02-12
speed up the finding the applicable AOP advisors by parallelism

caitsithx:
the Determining the sublist of a list of candidate advisors that are applicable to the given class can be simply parallelized using the parallelStream to speed up the spring-context start-up process.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/autoproxy/AbstractAdvisorAutoProxyCreator.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/autoproxy/AbstractAdvisorAutoProxyCreator.java
@@ -118,18 +118,10 @@ public abstract class AbstractAdvisorAutoProxyCreator extends AbstractAutoProxyC
 	 * @param beanClass the target's bean class
 	 * @param beanName the target's bean name
 	 * @return the List of applicable Advisors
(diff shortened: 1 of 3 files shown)


snicoll:
@caitsithx thanks for the suggestion but we don't intend to use parallel streams internally anywhere by design. If we parallelize, then at a higher level where we manage the granularity ourselves.

> to speed up the spring-context start-up process.

Did you measure it? Given how much micro-switching this would involve at that level, we're very much doubtful this would actually make a difference. Going forward, please consider measuring and sharing a performance benchmarks as part of a a change like this.

caitsithx:
> @caitsithx thanks for the suggestion but we don't intend to use parallel streams internally anywhere by design. If we parallelize, then at a higher level where we manage the granularity ourselves.
> 
> > to speed up the spring-context start-up process.
> 
> Did you measure it? Given how much micro-switching this would involve at that level, we're very much doubtful this would actually make a difference. Going forward, please consider measuring and sharing a performance benchmarks as part of a a change like this.

Thank you @snicoll for the reviewing.
1) Got it. a fine-grained granularity could be more efficient, controllable and more tune-able. I will make some changes.
2) In fact I am working with a spring application with several thousands of beans, it could remove 1 mins during the startup by parallelizing the Determining of the candidate advisors. as you suggested, I will attach a performance test result.

PR #32250, opened 2024-02-12
Include default request options early on.

### LINKED ISSUE #32053: WebClient and RestClient's defaultRequest(..) is not invoked early enough
### Background
```java
/**
 * Global option to specify a header to be added to every request,
 * if the request does not already contain such a header.
 */
Builder defaultHeader(String header, String... values);
```
`WebClient` and `RestClient` have `default[Header|Cookie|Request|..]` that useful to set `default value` on request.

https://github.com/spring-projects/spring-framework/blob/899de4f3bfc373bdf1c0e67c7160331bd056825b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L502-L508

As you can see above code, `defaultHeader` value is set only when it's not contained in headers.

### Question
```java
/**
 * Provide a consumer to customize every request being built.
 * @param defaultRequest the consumer to use for modifying requests
 */
Builder defaultRequest(Consumer<RequestHeadersSpec<?>> defaultRequest);
```
https://github.com/spring-projects/spring-framework/blob/899de4f3bfc373bdf1c0e67c7160331bd056825b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L481-L484

But on `defaultRequest(..)`, it can override request's values unlike `defaultHeader` cause it's called on every request being built.

### Reproduce
```java
// DefaultWebClientTests.java
@Test
void defaultRequest_override() {
	ThreadLocal<String> context = new NamedThreadLocal<>("foo");

	WebClient client = this.builder
			.defaultRequest(spec -> spec.accept(MediaType.APPLICATION_JSON)) // default
			.build();

	client.get().uri("/path")
			.accept(MediaType.IMAGE_PNG) // set
			.retrieve().bodyToMono(Void.class).block(Duration.ofSeconds(10));

	ClientRequest request = verifyAndGetRequest();
	assertThat(request.headers().getAccept()).isEqualTo(MediaType.IMAGE_PNG); // ❌Expected: <image/png> but actual: <application/json>
}
```
image

I think `defaultRequest(spec -> spec.accept(MEDIA_TYPE))` will override all media types that set by `accept(..)` on this request.

Q. is above behavior of `defaultRequest(..)` is intended? thanks!

youssef3wi:
Fixes #32053.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
@@ -220,6 +220,11 @@ final class DefaultWebClient implements WebClient {
 
 		DefaultRequestBodyUriSpec(HttpMethod httpMethod) {
 			this.httpMethod = httpMethod;
+
+			// Add defaultRequest option to WebClient.Builder
+			if (defaultRequest != null) {
+				defaultRequest.accept(this);
+			}
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


snicoll:
@youssef3wi thanks but the related issue is already assigned to @poutsma.

PR #32259, opened 2024-02-13
Ignore trailing semicolons when parsing Accept-Language header

schnapster:
Something I'm seeing in the wild is a trailing semicolon for `Accept-Language` headers. Partial example stack trace:

```
java.lang.IllegalArgumentException: range=en;
    at java.util.Locale$LanguageRange.<init>(Locale.java:3099)
    at sun.util.locale.LocaleMatcher.parse(LocaleMatcher.java:525)
    at java.util.Locale$LanguageRange.parse(Locale.java:3214)
    at org.springframework.http.HttpHeaders.getAcceptLanguage(HttpHeaders.java:505)
    at org.springframework.http.HttpHeaders.getAcceptLanguageAsLocales(HttpHeaders.java:526)
```

I don't think it would hurt to be more lenient here and drop the trailing semicolon before attempting to parse it.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -502,6 +502,9 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 	 */
 	public List<Locale.LanguageRange> getAcceptLanguage() {
 		String value = getFirst(ACCEPT_LANGUAGE);
+		if (value != null) {
+			value = StringUtils.trimTrailingCharacter(value, ';');
+		}
 		return (StringUtils.hasText(value) ? Locale.LanguageRange.parse(value) : Collections.emptyList());
 	}
 
--- a/spring-web/src/test/java/org/springframework/http/HttpHeadersTests.java
+++ b/spring-web/src/test/java/org/springframework/http/HttpHeadersTests.java
@@ -489,6 +489,19 @@ class HttpHeadersTests {
 		assertThat(headers.getAcceptLanguageAsLocales()).element(0).isEqualTo(Locale.FRANCE);
 	}
 
+	@Test
+	void acceptLanguageTrailingSemicolon() {
+		String headerValue = "en-us,en;";
+		headers.set(HttpHeaders.ACCEPT_LANGUAGE, headerValue);
+		assertThat(headers.getFirst(HttpHeaders.ACCEPT_LANGUAGE)).isEqualTo(headerValue);
+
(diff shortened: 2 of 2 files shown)


rstoyanchev (commented):
@schnapster, the header is parsed via `Locale.LanguageRange.parse`, which supports a weighted list of ranges where each may have a `q` parameter. However, the parsing logic doesn't allow a lot of flexibility and looks specifically for `";q="`. Note also it first splits by `,` to create a list of range value. So the problem may still occur if there is a trailing `;` in one of the other ranges.

We don't want to replicate the parsing logic from `Locale`, but we may be able to do a split by `,` and then check for invalid ";" in each. However, this is better done defensively, i.e. allowing parsing to fail first and then reacting to that in fallback mode.

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -502,6 +502,9 @@ public void setAcceptLanguage(List<Locale.LanguageRange> languages) {
 	 */
 	public List<Locale.LanguageRange> getAcceptLanguage() {
 		String value = getFirst(ACCEPT_LANGUAGE);
+		if (value != null) {
+			value = StringUtils.trimTrailingCharacter(value, ';');
```
jdicke, line 506: ';' should be a variable

jdicke (approved):

spring-web/src/main/java/org/springframework/http/HttpHeaders.java
```diff
@@ -502,6 +502,9 @@ public void setAcceptLanguage(List<Locale.LanguageRange> languages) {
 	 */
 	public List<Locale.LanguageRange> getAcceptLanguage() {
 		String value = getFirst(ACCEPT_LANGUAGE);
+		if (value != null) {
```
kmesiab, line 505: While I think a null check is fine, why not also validate whether this string in non empty and at least [n] chars?

schnapster:
@rstoyanchev apologies for the late follow-up, I'm a bit out of the loop on this by now. Could you please provide a concrete header to demonstrate your example? Or ideally a (pseudo-cody) test case similar to the one I added? I'd be happy to take another stab at this if I understand better what the additional faults are that you want Spring to be able to handle. We can also handle that in another issue / PR, imho this one already provides incremental value by itself.

poutsma:
@schnapster Thank you for submitting a PR. I have combined your code and the suggestions of @rstoyanchev into 099d0168577ab498529f18a48439a8235c14839a.

PR #32272, opened 2024-02-14
Correct documentation for `userDestinationPrefix` in `MessageBrokerRegistry`

Xednar:
Changes documentation for setUserDestinationPrefix to match getTargetDestination in DefaultUserDestinationResolver

Current documentation is misleading, because the default implementation appends "-user" + sessionId to the actualDestination


rstoyanchev (commented):
The `targetDestination` is for matching source destination that are used to target a user. This is translated to a target destination that is a unique queue name (and does not contain "user" in it.

jdicke (approved):

PR #32281, opened 2024-02-15
Use Spliterator of underlying collection in MutablePropertyValues / MutablePropertySources

kilink:
Delegate to the spliterator method of the underyling collection in MutablePropertyValues and MutablePropertySources. In both cases, those collection types have specialized Spliterator implementations. Delegating to these Spliterators also means the characteristics of the Spliterator are properly set.

Note: this spliterator was already exposed via the stream() method, since that would delegate to the underlying collection. This mostly just makes things consistent.


jdicke (approved):
Nice

PR #32291, opened 2024-02-18
Avoid resizing of fixed-size `HashSet`/`LinkedHashSet` variants

kilink:
Add helpers to CollectionUtils for building HashSets and LinkedHashSets that can hold an expected number of elements without needing to resize / rehash.

I also went ahead and updated places throughout the code base that appeared to be potentially sizing HashSet / LinkedHashSet incorrectly. This is by no means exhaustive.

See ff11467a0c6, where this was added for HashMap / LinkedHashMap.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/CollectionUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/CollectionUtils.java
@@ -22,8 +22,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -85,7 +87,7 @@ public abstract class CollectionUtils {
 	 * @see #newLinkedHashMap(int)
 	 */
 	public static <K, V> HashMap<K, V> newHashMap(int expectedSize) {
-		return new HashMap<>(computeMapInitialCapacity(expectedSize), DEFAULT_LOAD_FACTOR);
+		return new HashMap<>(computeInitialCapacity(expectedSize), DEFAULT_LOAD_FACTOR);
 	}
 
 	/**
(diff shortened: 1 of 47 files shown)


spring-core/src/main/java/org/springframework/util/CollectionUtils.java
```diff
@@ -105,6 +107,30 @@ public static <K, V> LinkedHashMap<K, V> newLinkedHashMap(int expectedSize) {
...
+	 * any immediate resize/rehash operations to be expected.
+	 * @param expectedSize the expected number of elements (with a corresponding
+	 * capacity to be derived so that no resize/rehash operations are needed)
+	 * @see #newHashSet(int)
+	 */
+	public static <E> LinkedHashSet<E> newLinkedHashSet(int expectedSize) {
+		return new LinkedHashSet<>(computeMapInitialCapacity(expectedSize), DEFAULT_LOAD_FACTOR);
+	}
+
 	private static int computeMapInitialCapacity(int expectedSize) {
```
sbrannen, line 134:
```suggestion
	private static int computeInitialCapacity(int expectedSize) {
```

Probably better to rename this method since it's now used for maps and sets.

kilink: I thought about that, but noticed that [Guava](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Sets.java#L262) and the [JDK](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/HashSet.java#L401) both refer to map capacity helpers in their versions of this helper. It sort of makes sense I guess since it ends up being backed by a Map. Don't really have a strong opinion so I'll make your suggested change.

kilink:
I also wanted to note that part of the reason for this PR is that I'd like to be able to rely on this helper in Spring projects that can't use the JDK19 helpers, and where we don't necessarily want to rely on Guava.

spring-core/src/main/java/org/springframework/util/CollectionUtils.java
sbrannen: > It sort of makes sense I guess since it ends up being backed by a Map.

True. I just thought the new name would require less cognitive overhead for anyone reading the code. Though, to be honest, it's an internal helper method. So it doesn't really matter. 

In any case, thanks for making the change.

On a side note, I just learned that there's a new `HashSet.newHashSet()` factory method in Java 19. So thanks for that. 👍 

Though we still have a Java 17 baseline.

spring-core/src/main/java/org/springframework/util/CollectionUtils.java
```diff
@@ -102,10 +104,34 @@ public static <K, V> HashMap<K, V> newHashMap(int expectedSize) {
...
 
-	private static int computeMapInitialCapacity(int expectedSize) {
+	/**
+	 * Instantiate a new {@link HashSet} with an initial capacity
+	 * that can accommodate the specified number of elements without
+	 * any immediate resize/rehash operations to be expected.
+	 * @param expectedSize the expected number of elements (with a corresponding
+	 * capacity to be derived so that no resize/rehash operations are needed)
+	 * @see #newLinkedHashSet(int)
+	 */
```
sbrannen, line 117: In the future, please remember to add `@since` tags to newly introduced non-private` methods and to update the copyright headers of all affected files.

Though no need to do that for this PR: I'll take care of that when merging.

sbrannen:
This has been merged into `main` in e1a32d4ba9cb88931a34172927a396fae2043133, revised in b9c304b8905e6a0112dd888418e5fa56a0fe0c6e, and _extended_ in 644887e09447e464d580e8bef3c8fa3cff20c5bf.

Thanks

PR #32294, opened 2024-02-19
Use `ConcurrentHashMap.newKeySet` where feasible

kilink:
In places where a ConcurrentHashMap was used as a set by wrapping it with Collections.newSetFromMap, switch to just using the set returned by ConcurrentHashMap.newKeySet directly.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultSingletonBeanRegistry.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultSingletonBeanRegistry.java
@@ -16,7 +16,6 @@
 
 package org.springframework.beans.factory.support;
 
-import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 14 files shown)


spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultSingletonBeanRegistry.java
```diff
@@ -16,7 +16,6 @@
 
 package org.springframework.beans.factory.support;
 
-import java.util.Collections;
```
sbrannen, line 19: Surprise: I just noticed that 902e5707a88522a158985138e4ec8b1061184d1c introduced additional usage of `java.util.Collections`, so the code no longer compiles when merged into `main`.

Though fret not: I'll add the import back locally. 😉

sbrannen:

Thanks

PR #32301, opened 2024-02-21
Clarify fallback beans are included for autowiring Arrays, Collections, and Maps

quaff:
Test is improved to verify that.


jhoeller:
The same is true for primary beans, so I've extended the test to both primary and fallback beans and added a cross-reference in the javadoc notes.

PR #32318, opened 2024-02-23
Optimize HttpHeaders.getAcceptLanguageAsLocales

kilink:
The HttpHeaders.getAcceptLanguageAsLocales was incurring overhead from using a Stream, as well as calling the fairly expensive Locale.getDisplayName method.

Switch to using an ArrayList, and skipping over wildcard ranges to avoid needing to check the display name.

Here is a CPU flamegraph from one of our apps where this method call stands out.

image

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -527,10 +527,14 @@ public class HttpHeaders implements MultiValueMap<String, String>, Serializable 
 		if (ranges.isEmpty()) {
 			return Collections.emptyList();
 		}
-		return ranges.stream()
-				.map(range -> Locale.forLanguageTag(range.getRange()))
-				.filter(locale -> StringUtils.hasText(locale.getDisplayName()))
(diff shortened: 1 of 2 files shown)


jhoeller:
@poutsma the present check there seems worth optimizing in the 6.1.x line. `getDisplayName` being a specific culprit in that method, but also `.stream()` usage worth reviewing in other `HttpHeaders` methods.

I'm not sure about the intent of the original `getDisplayName` check there and whether the range check replacement is 100% equivalent for our purposes, so I'll leave it up to you to refine this if necessary.

kilink:
I'm fairly certain checking of display name was to handle wildcards as described in rfc4647. Based on the validation in `Locale.LanguageRange.parse`, it appeared to me to be the only possible way to end up with a Locale instance without a display name, as all other invalid language tags are rejected.

```java
Locale.forLanguageTag("*").getDisplayName(); // empty
Locale.forLanguageTag("*-CA").getDisplayName(); // empty
```

poutsma:
Merged. Thanks for submitting a PR, @kilink !

PR #32367, opened 2024-03-04
AbstractErrors: use isEmpty() instead of length()

alidandach:
This pull request proposes an enhancement to the doSetNestedPath method where I replaced the check for an empty string using .length() > 0 with the more concise !nestedPath.isEmpty() in the conditional statement. This change improves readability and aligns with common Java best practices for string emptiness checks.


jhoeller:
Thanks, good catch!

Issue #32360, opened 2024-03-04
`StringUtils.uriDecode` decodes strings with non-ASCII characters incorrectly


Romanow88:
**Affects:** 6.1.3

`StringUtils.uriDecode` will take two different paths depending if "%" is found. One of these does not handle non-ASCII characters.

Fast path:

```java
StringUtils.uriDecode("ü", StandardCharsets.UTF_8) // returns "ü"
```

Replace path:

```java
StringUtils.uriDecode("%20ü", StandardCharsets.UTF_8) // returns " �"
```

dumbbelloper:
Hello, I've submitted a Pull Request that aims to address this issue: [PR #32373](https://github.com/spring-projects/spring-framework/pull/32373)

poutsma:
StringUtils::uriDecode expects the input to be encoded, which means that the `ü` character should have been provided as `%C3%BC` instead. I will add a documentation note that makes this behavior clear.

Note that `StringUtils::uriDecode` is a low-level, internal utility method exposed via higher-level components, such as `UriComponentsBuilder`, that are better at dealing with [en|de]coded URIs.

PR #32373, opened 2024-03-05
Fix non-ASCII character handling in StringUtils.uriDecode

dumbbelloper:
StringUtils.uriDecode now correctly handles non-ASCII characters regardless of the presence of "%" encoding.

Previously, the method took two different paths depending on whether "%" was found, leading to incorrect handling of non-ASCII characters in the absence of "%" encoding.

This fix ensures that all characters, including non-ASCII ones, are properly decoded using the provided Charset, improving the method's reliability and consistency across all inputs.

This change addresses issues with decoding multibyte characters and ensures compatibility with a wider range of character encodings, enhancing the utility's overall functionality.

#32360


poutsma:
Thank you for submitting a PR. Unfortunately, we have decided to take a different route in 646bd7f8933756802ae9f4e6450f0b2a23191304.

PR #32377, opened 2024-03-05
Use `String#isEmpty` where feasible

alidandach:
This PR replaces checks for empty strings (`"".equals(...)`) with the `String#isEmpty` method.

This change not only improves readability but also aligns with best practices in Java for string emptiness checking.


sbrannen:
Thanks for the PR, @alidandach.

We don't normally modify classes in our CGLIB fork; however, we'll make an exception to the rule in this case since the proposed changes align with idiomatic Java and potentially result in a performance improvement (albeit negligible).

PR #32380, opened 2024-03-06
Improve Javadoc for `substituteNamedParameters()` in `NamedParameterUtils` regarding empty lists

quaff:
Document that an empty list is not guaranteed to be supported by the database – for example, MySQL and PostgreSQL don't support `foo in ()`.


sbrannen:
This has been merged into `main` in https://github.com/spring-projects/spring-framework/commit/be136d79eec43125f21cfae650d4874829bb4a35 and revised in https://github.com/spring-projects/spring-framework/commit/6461eec582dc259dba2b6ee9b56ffbdcc50ce2f5.

Thanks

PR #32386, opened 2024-03-07
Support binding indexed values for record class

### LINKED ISSUE #32379: Nested record containing a query param of type List not being resolved correctly
**Affects:** \every version up to 3.3.0-SNAPSHOT (tested with both 3.2.3 and 3.3.0-SNAPSHOT)

Hello!

I came across an issue with nested records used as query parameters. Here is what my Controller looks like:
```
@RestController
@RequestMapping("/api")
public class DemoController {

    @GetMapping
    public void test(RecordParamHolder paramHolder) {
        System.out.println("Record: " + paramHolder.recordParamName()); // not working when passing indexed param list
        System.out.println("POJO: " + paramHolder.pojoParamName()); // always working
    }
}
```

`RecordParamHolder` looks like this (as the names state 1st one is a POJO and the 2nd one is a record):
```
public record RecordParamHolder(
        PojoListParam pojoParamName,
        RecordListParam recordParamName
) {}
```

Both `PojoListParam` and `RecordListParam` contain one field of type `List<String>`. And so when I try to pass a list of params like this: `GET localhost:8080/api?recordParamName.in%5B0%5D=abc` (square brackets escaped) I see that the list is not being resolved and is `NULL`. However, similar request works fine for nested POJO class – `GET localhost:8080/api?pojoParamName.in%5B0%5D=abc`.

I know that I can pass a list like this: `paramName=value1,value2,etc.`, but this approach doesn't work when there is a comma in any of the values (see this issue for example #29411). I know I can also pass it like this `paramName=value1&paramName=value2&etc`, but I was told it's inconvenient to build such request from JavaScript because paramName is being overwritten and only the last one is passed.

Anyway, I believe it should work properly for nested records just as it works for nested POJOs. Attaching a [Minimal Reproducible Example](https://github.com/spring-projects/spring-framework/files/14500838/demo.zip).

Thanks in advance!

quaff:
Fix GH-32379

### DIFF
--- a/spring-context/src/main/java/org/springframework/validation/DataBinder.java
+++ b/spring-context/src/main/java/org/springframework/validation/DataBinder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,6 +49,7 @@ import org.springframework.beans.PropertyValues;
 import org.springframework.beans.SimpleTypeConverter;
 import org.springframework.beans.TypeConverter;
 import org.springframework.beans.TypeMismatchException;
+import org.springframework.core.CollectionFactory;
 import org.springframework.core.KotlinDetector;
 import org.springframework.core.MethodParameter;
 import org.springframework.core.ResolvableType;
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Thanks for taking a look and the suggestion @quaff. We are going to address this more comprehensively with #32426 by adding support for constructor binding for collections and maps.

PR #32387, opened 2024-03-07
add synchronized code when add new data source

huanccwang:
Currently we use AbstractRoutingDataSource to dynamically switch data source. But sometimes switch would have some exception, then we found when switch concurrently, exception may happen. So add synchronized code when add new data source(initialize() and determineTargetDataSource()).


jhoeller:
Please note that `AbstractRoutingDataSource` is generally not designed for concurrent reconfiguration. It expects to be populated in a configuration step and then to act against that prepared configuration at operation time.

How do you dynamically switch the DataSource, actually? Calling `setTargetDataSources` and then `initialize` again? There is also a race condition between the setter and the init method invocation then. I'm afraid that simple synchronization/locking in the existing methods won't be sufficient, this rather requires some specific design for dynamic reconfigurability.

PR #32398, opened 2024-03-08
Fix DefaultServerObservation httpUrl value

ruansheng8:
See `DefaultServerRequestObservationConvention.java` :
```java
	@Override
	public KeyValues getHighCardinalityKeyValues(ServerRequestObservationContext context) {
		// Make sure that KeyValues entries are already sorted by name for better performance
		return KeyValues.of(httpUrl(context));
	}

```

httpUrl is HighCardinalityKey, It should use `context.getCarrier().getURI().toString()` in DefaultServerRequestObservationConvention#httpUrl(ServerRequestObservationContext context) (line:174)

image


ruansheng8:
I think the http.url should be contains http request query parameters

bclozel:
I don't think we should. If anything, we should align in the future with OTel conventions and those are separating path and query in different tags. I'm not sure when we should do that as this would not be a backwards compatible change and I don't know if HTTP OTel conventions are stable at this point.

Do you have another spec in mind pointing to the behavior you're describing here?

ruansheng8:
Thanks for the answer. I looked at the definition of `opentelemetry` and you are right.

PR #32402, opened 2024-03-08
Add Spring Boot Converter to "Configuring a Global Date and Time Format" documentation

m-dutaut:
Update the "Configuring a Global Date and Time Format" so that its application does not break support of Duration properties [readable](https://docs.spring.io/spring-boot/docs/2.1.8.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-conversion-duration) format (1s, 2h,3d...)

Without `ApplicationConversionService.addApplicationConverters(conversionService)` loading Duration annotated with `@Value` would fail because the `DurationConverter` registered within the `DateTimeFormatterRegistrar` does not support "readable" Duration format.


snicoll:
@m-dutaut `ApplicationConversionService` is a Spring Boot concept, and so is the feature you are referring to. We can't document that here.

PR #32405, opened 2024-03-10
Avoid cloning empty `Annotation` array in `TypeDescriptor`

kilink:
Rework `AnnotatedElementAdapter` to avoid cloning the underlying `Annotation` array if it is empty when `getAnnotations()` is called.

Additionally, make the class static and add a factory method that returns a singleton instance for `null` or an empty `Annotation` arrays.


jhoeller:
Backported to 6.1.6, 6.0.19, 5.3.34 as well now since it is a straightforward internal optimization in a hot code path.

sbrannen:
- see also #32476

PR #32425, opened 2024-03-12
Use lambda expressions 

alidandach:
used lambda expressions and method references where possible, reducing the need for extraneous code blocks.


snicoll:
@alidandach thank you but these are largely subjective and we prefer to do such changes ourselves.

Issue #32415, opened 2024-03-11
Mismatch between documented and actual message code handling for ErrorResponse


ZeroCyan:
There is a discrepancy between the documentation and the actual implementation regarding the message codes for ErrorResponses.

[https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-ann-rest-exceptions.html#mvc-ann-rest-exceptions-i18n](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-ann-rest-exceptions.html#mvc-ann-rest-exceptions-i18n) describes the following pattern:

> The default strategy for message codes follows the pattern: problemDetail.\[type|title|detail].\[fully qualified exception class name]

As you can see in the following references, `detail` is not used when searching for the message code. This does happen for `problemDetail.title` and `problemDetail.type`

https://github.com/spring-projects/spring-framework/blob/0435e9c49e299a4db345cc120f29e844ccca96f1/spring-web/src/main/java/org/springframework/web/ErrorResponse.java#L162

https://github.com/spring-projects/spring-framework/blob/0435e9c49e299a4db345cc120f29e844ccca96f1/spring-web/src/main/java/org/springframework/web/ErrorResponse.java#L172

https://github.com/spring-projects/spring-framework/blob/0435e9c49e299a4db345cc120f29e844ccca96f1/spring-web/src/main/java/org/springframework/web/ErrorResponse.java#L184

Expected message code according to documentation: 
- problemDetail.detail.org.springframework.web.bind.MethodArgumentNotValidException

Actual message code that gets resolved
- problemDetail.org.springframework.web.bind.MethodArgumentNotValidException

snicoll:
Closing in favor of PR #32415

PR #32446, opened 2024-03-14
Default strategy for ProblemDetail error codes wrongly document how "detail" is supported

LinorDolev:
Doc fix for #32415
Removed 'details' attribute from documentation as it is not being used in implementation at all.


snicoll:
@LinorDolev thank you for your first contribution. For the record, while "detail" is not supported the way it was documented, there is a strategy for it. I've polished the change to make that more explicit.

PR #32448, opened 2024-03-15
Improve containsValue Method for CharSequence Comparison

alidandach:
**Problem**

The current implementation of the `containsValue` method checks if a provided value is an instance of `String` and then compares this value to the entries within headers using `String.equals()`. This approach limits the comparison strictly to String instances and does not account for other CharSequence implementations like `StringBuilder` or `StringBuffer`, which might also represent valid sequences of characters to be compared.

**Solution**

This MR introduces an improvement to the `containsValue` method, enabling it to handle comparisons with any objects implementing the `CharSequence` interface, not just String. This is achieved by converting both the value and the entry value to String before comparison, ensuring that the comparison logic remains effective while becoming more flexible.

**Changes Made**

-  Check if value is an instance of CharSequence instead of strictly String.
-  Convert value to String (if it's a CharSequence) before the comparison.
-  Update the comparison logic to ensure it correctly handles CharSequence implementations.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/support/Netty5HeadersAdapter.java
+++ b/spring-web/src/main/java/org/springframework/http/support/Netty5HeadersAdapter.java
@@ -115,9 +115,12 @@ public final class Netty5HeadersAdapter implements MultiValueMap<String, String>
 
 	@Override
 	public boolean containsValue(Object value) {
-		return (value instanceof String &&
-				StreamSupport.stream(this.headers.spliterator(), false)
-						.anyMatch(entry -> value.equals(entry.getValue())));
+		if (value instanceof CharSequence) {
+			String valueStr = value.toString();
+			return StreamSupport.stream(this.headers.spliterator(), false)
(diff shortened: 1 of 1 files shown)


bclozel:
I think this change would break the implementation here. While netty headers can indeed contain `CharSequence` keys and values, the adapter here is only exposing a `MultiValueMap<String, String>`.

This change would make the following behavior possible, which I think would be inconsistent:

```
HttpHeaders nettyHeaders = HttpHeaders.newHeaders();
nettyHeaders.add(new StringBuilder().append("test"), new StringBuilder().append("spring"));
Netty5HeadersAdapter adapter = new Netty5HeadersAdapter(nettyHeaders);

// we got a List<String> here
assertThat(adapter.getFirst("test")).isInstanceOf(String.class).isEqualTo("spring");

// this is true, you can get a String "spring"
assertThat(adapter.containsValue("spring")).isTrue();
// this should be false, you will never get a `StringBuilder` value from the map
assertThat(adapter.containsValue(new StringBuilder().append("spring"))).isFalse();
```

PR #32462, opened 2024-03-16
Resolve property-dependent parameter names for exception messages

### LINKED ISSUE #32323: Property dependent `@RequestParam` parameters are not resolved on bad request
**Affects:** \3.2.2

---
I came across this issue a while ago and since I can't find it anywhere nor create an easy workaround, I create this issue.

A bit of context first : 
I created my own fully customizable authentication service built over spring, and on the controller, I have the following method :
```java
@PostMapping("${security.authentication.log-in.mapping}")
public ResponseEntity<Void> handleLogIn(@RequestParam("${security.authentication.identifier.parameter.name}") String identifier,
                                        @RequestParam("${security.authentication.raw-password.parameter.name}") String rawPassword) {
    //My authentication logic
}
```

With the following properties :
```properties
security.authentication.log-in.mapping=logIn
security.authentication.identifier.parameter.name=username
security.authentication.raw-password.parameter.name=password

```

The references to the properties are resolved as expected during normal situations, thus allowing me to fully customize the authentication endpoint mapping and parameter names.

However, if one of the property is misspelled or missing, The default error message of the BAD_REQUEST response does not resolve the property see the image below (mispelling **usernam** on purpose) :
After sending the following POST request `http://localhost:443/authentication/logIn?usernam=user&password=pwd`
image

The expected message would be `"Required parameter 'username' is not present."`

andrea-mauro:
When named value method arguments are defined through placeholders or SpeL and an exception is thrown because the required parameter is missing or can't be converted, the exception message references the placeholder instead of the resolved named value. Example:
```
Required header '${systemProperty}' is not present.
```

This PR enhances the error message by passing the resolved parameter name to the error handlers, so that the exception thrown is more relevant.

Closes #32323

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/reactive/AbstractNamedValueMethodArgumentResolver.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/reactive/AbstractNamedValueMethodArgumentResolver.java
@@ -99,7 +99,7 @@ public abstract class AbstractNamedValueMethodArgumentResolver implements SyncHa
 			else if (namedValueInfo.required && !nestedParameter.isOptional()) {
 				handleMissingValue(resolvedName.toString(), nestedParameter, message);
 			}
-			arg = handleNullValue(namedValueInfo.name, arg, nestedParameter.getNestedParameterType());
+			arg = handleNullValue(resolvedName.toString(), arg, nestedParameter.getNestedParameterType());
 		}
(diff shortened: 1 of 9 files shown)


simonbasle (changes requested):
Looks good to me, but IIRC there's a similar pattern in `spring-webflux` as well (in `AbstractNamedValueArgumentResolver`). can you cover that one as well?

andrea-mauro:
@simonbasle The occurrence in `spring-webflux` is now covered as well. Please take a look.

simonbasle (approved):

PR #32467, opened 2024-03-17
Make SimpleAliasRegistry.aliasNames final

izeye:
This PR changes `SimpleAliasRegistry.aliasNames` to `final`.


jhoeller:
Good catch, thanks!

Issue #23844, opened 2019-10-22
Spring AOP should not apply UndeclaredThrowableException to checked exceptions from Kotlin methods


ryandanielspmc:
Currently, when using Spring AOP with Kotlin, `UndeclaredThrowableException` is thrown when a method with advice throws a checked exception. There is a **_workaround_** in Kotlin in that you can add `@Throws` annotations to methods, but this is not really feasible as it would require annotating every method of every project, and methods in other libraries might not be under your control.

What would solve this problem would be to have an option in Spring to disable the throwing of `UndeclaredThrowableException` and simply throw the original exception.

ryandanielspmc:
Looks like the relevant code is here:
https://github.com/spring-projects/spring-framework/blob/9bd3a535cd524cf771423e67feeea523ac985f87/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java#L752-L759

caio-bos-ifood:
This issue is also impacting me when using resilience4j circuit breaker. The AOP is encapsulating the real exception so the circuit breaker is never switched to OPEN.

jhoeller:
The reason for that custom `UndeclaredThrowableException` throwing logic in `CglibAopProxy` is that interface-based proxies unavoidably get such a `UndeclaredThrowableException` out of the JDK implementation, so for consistency with `JdkDynamicAopProxy`, we manually enforce it in the CGLIB variant.

So whatever we do about it, we can only do it for CGLIB proxies. @sdeleuze what's your take on this? Is CGLIB+Kotlin a common enough case to be handled specifically - or on the inverse, are JDK proxies with Kotlin uncommon enough? Are checked exceptions without explicit declaration common enough in Kotlin?

Generally speaking for the scenarios above, why the need for checked exceptions coming out of an interceptor to begin with? From my perspective, this is bad practice with Java-based AOP, even more so with Kotlin-based AOP, and it's also a general Java interoperability issue when used in Kotlin-based arrangements.

ryandanielspmc:
@jhoeller 

> why the need for checked exceptions coming out of an interceptor to begin with?

In our case, we wanted to add logging around `@Service` methods with AOP. And in this scenario, the logging code should not change the behavior of the methods being logged.

Additionally, this method of logging would allow us to:

- Avoid manual modification of dozens of methods with identical boilerplate code to add logging
- Avoid modification of each of these again later if we want to change how the logging works
[...]

Issue #26854, opened 2021-04-23
@Transactional should treat UndeclaredThrowableException as a checked exception


Sam-Kruglov:
I read [this blogpost](https://blog.pchudzik.com/201911/transactional-errors/) describing this problem:
```java
@SneakyThrows
@Transactional
public void lombokSurprise() {
    jdbcTemplate.execute("insert into test_table values('lombok!')");
    throw new Exception("Simple exception");
}
```
This code is supposed to commit the transaction but it does not. The problem is that Spring wraps checked exceptions that aren't declared in the signature within an [UndeclaredThrowableException](https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/UndeclaredThrowableException.html) over [here](https://github.com/spring-projects/spring-framework/blob/01e50fb60ad0d82f96fe4df6054ed6029e2f6a12/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java#L767) and from that point treats it as unchecked. 

I think we should treat this one as a checked exception to preserve the promised behavior. I believe it's controlled here:

https://github.com/spring-projects/spring-framework/blob/01e50fb60ad0d82f96fe4df6054ed6029e2f6a12/spring-tx/src/main/java/org/springframework/transaction/interceptor/DefaultTransactionAttribute.java#L171-L188

So, we could append `|| ex instanceof UndeclaredThrowableException` to solve it.

It will be a breaking change, but I think most people expect it to work that way anyway.

sbrannen:
> So, we could append `|| ex instanceof UndeclaredThrowableException` to solve it.

If you are proposing that be added in the `rollbackOn()` method in `DefaultTransactionAttribute`, that would not have any effect since `UndeclaredThrowableException` extends `RuntimeException`.

Are you perhaps proposing that the `cause` of the `UndeclaredThrowableException` be taken into consideration?

sbrannen:
@Sam-Kruglov, are you aware that you can configure `@Transactional` as follows?

```java
@Transactional(noRollbackFor = UndeclaredThrowableException.class)
@SneakyThrows
public void lombokSurprise() {
    jdbcTemplate.execute("insert into test_table values('lombok!')");
    throw new Exception("Simple exception");
}
```

That should help you achieve your goal.

Sam-Kruglov:
Editing ‘@Transactional’ seems like a workaround, I think it’s more intuitive if it’s the default behavior. 

[...]

PR #32469, opened 2024-03-18
Consistent handling of undeclared checked exceptions in CGLIB proxies

LeMikaelF:
This PR fixes a regression in 5.2.RC2.

Before this version, if an object was proxied with CGLIB and threw an undeclared checked exception, for example using lombok's `@SneakyThrows`, the proxy interceptors would see the target's exception as is, and the CGLIB generated class would then wrap the exception in an `UndeclaredThrowableException`, to match the behaviour of `java.lang.reflect.Proxy`, so that callers would see the target's exception wrapped in an `UndeclaredThrowableException`. But [starting with 5.2.RC2](https://github.com/spring-projects/spring-framework/commit/9bd3a535cd524cf771423e67feeea523ac985f87), the `UndeclaredThrowableException` is created in the process of invoking the target, inside the proxy, and so the interceptors see the target's exception only after it has been wrapped in an `UndeclaredThrowableException`, instead of seeing the raw exception from the target. An exception was since then added to prevent this in Kotlin applications, where all exceptions are unchecked ([relevant issue](https://github.com/spring-projects/spring-framework/issues/23844)), but not for Java.

Concretely, this means for example that the rollback behaviour of `@Transactional` methods changes in 5.2.RC2 for Java applications. One situation where this can occur is that if an application uses CGLIB proxies and undeclared checked exceptions, transactions will stop rolling back on those exceptions with the 5.2 upgrade. Another case is that if a post-5.2 application uses a CGLIB proxy with `@Transactional`, extracting an interface from a bean can cause the transaction to stop rolling back on undeclared checked exceptions. Below is the relevant section from a [reproducer](https://github.com/LeMikaelF/transactional-reproducer/blob/main/src/test/java/com/mikaelfrancoeur/transactionalreproducer/TransactionalIssueTest.java).

```java
@Test
void givenCglibProxyAndNoRollbackForRuntimeException_whenUndeclaredCheckedException_thenRollsBack() {
    assertThat(Enhancer.isEnhanced(myClass.getClass())).isTrue();

    assertThrowsUndeclaredThrowableException(() -> myClass.throwException(new CheckedException()));

    verify(platformTransactionManager).rollback(any()); // fails in 5.2.RC1, passes in 5.2.RC2
}

@Test
void givenJdkProxyAndNoRollbackForRuntimeException_whenUndeclaredCheckedException_thenRollsBack() {
    assertThat(Proxy.isProxyClass(myInterfaceImpl.getClass())).isTrue();

    assertThrowsUndeclaredThrowableException(() -> myInterfaceImpl.throwException(new CheckedException()));

    verify(platformTransactionManager, never()).rollback(any());
}
```
This regression also seems to affect Resilience4j's circuit breaker, according to [one report](https://github.com/spring-projects/spring-framework/issues/23844#issuecomment-747061071).

The intent behind the new error-handling introduced in 5.2.RC2 was to avoid a bug in CGLIB, but it also changed the behaviour of AOP interceptors. I tried fixing this by reusing a similar error-handling pattern, but the issue is that CGLIB proxies don't always use interceptors, in some cases they use dispatchers to directly invoke the target, so the user-facing error behaviour became really messy.

The best solution to this issue is therefore to do the undeclared-exception handling in the generated code. It turns out, there's a fork of CGLIB (also Apache 2) where the issue that prompted the custom error-handling in the first place was fixed ([issue](https://github.com/cglib/cglib/issues/189) and [unmerged](https://github.com/cglib/cglib/pull/190) PR in CGLIB). I reused the implementation from that PR committed it with the original author's name (@hengyunabc), and did some light refactor to it.

This MR therefore consists of 3 parts:
* Fix CGLIB
* Revert the error-handling to use CGLIB's `UndeclaredThrowableTransformer`
* Add a test suite that can catch regressions like the one in 5.2.RC2


LeMikaelF:
I pushed a commit that fixes the checkStyle issues. 

For reference, I found this other related issue: https://github.com/spring-projects/spring-framework/issues/26854

spring-core/src/main/java/org/springframework/cglib/transform/impl/UndeclaredThrowableTransformer.java
```diff
@@ -55,10 +55,20 @@ public CodeEmitter begin_method(int access, final Signature sig, final Type[] ex
             return e;
         }
         return new CodeEmitter(e) {
-            private Block handler;
-            /* init */ {
-                handler = begin_block();
-            }
+	        private final boolean isConstructor = Constants.CONSTRUCTOR_NAME.equals(sig.getName());
+            private Block handler = begin_block();
+	        private boolean calToSuperWasSeen;
```
andrei-ivanov, line 60: typo perhaps in `cal`?

jhoeller: I'm preparing a revision of this to separate `ClassLoaderAwareGeneratorStrategy` from an `UndeclaredThrowableStrategy` delegate, to be committed soon. Noticed this only already, renaming it to `callToSuperSeen`.

Issue #31217, opened 2023-09-13
WebFlux temporary file not always deleted with parallel uploads.


davefarrelly:
I am working on a project using spring-webflux:6.0.8 and I have noticed an issue where temporary files created during a FilePart upload are not always getting deleted when uploads are done in parallel. 

However, if I do slow sequential uploads, the temporary files are getting deleted consistently as expected.

I have a reproducer with a single endpoint that receives a `FilePart`. Sending ~30 requests in parallel with with 28MB file seems to reproduce the issue consistently.

[spring-tempfile-reproducer.zip](https://github.com/spring-projects/spring-framework/files/12599520/spring-tempfile-reproducer.zip)

rstoyanchev:
I'm not able to reproduce the issue. There may be something more specific here depending on how the client uses creates and uses connections. Can you share the client code as well?

This is what I used:

```java
public class ClientApp {

	public static void main(String[] args) {

		FileSystemResource resource = new FileSystemResource("..."); // 30 MB file

		WebClient client = WebClient.create();

		Flux.range(1, 50)
				.flatMap(aLong -> client.post()
						.uri("http://localhost:9890/upload")
						.body(BodyInserters.fromMultipartData("uploadedFile", resource))
						.retrieve()
						.toBodilessEntity())
				.blockLast();
	}

}
```

davefarrelly:
@rstoyanchev thanks for the response! 

I was actually using JMeter with the "bzm - Parallel Controller" plugin to make the requests. 

I will test against our real application with a client similar to what you provided and also try 6.0.10 and get back to you with the results.

davefarrelly:
@rstoyanchev quick update, I tried updating to 6.0.10 and surprisingly none of the temporary files are getting deleted on this version. (Note. this is still using JMeter). Is there any config changes or anything needed with this version?

rstoyanchev:
Thanks for the additional info. I should have mentioned, might as well use the latest 6.0.13 (instead of 6.0.10), just in case. Also if you get a chance to run my sample code above, just to confirm whether it reproduces it for you or not.

davefarrelly:
Yeah I tested 6.0.13 and seems to be behaving the same as 6.0.10, none of the files are getting deleted. 

I ran the sample you gave above and results were the same on my end 🤔

davefarrelly:
[...]

PR #32482, opened 2024-03-18
Fix WebFlux temporary file not always deleted with parallel uploads

qeeqez:
# Fixes gh-31217. Delete temp files after multipart exchange

### Now temporary files removed as expected and we do not have gigabytes of data stored in temp folder after multipart exchange.

Actual fix is only in DefaultParts.java
All other changes - cleaning temp data after test runs

p.s. @poutsma pull request is for 6.1.x as per assigned milestone


poutsma:
I thank you for submitting a PR, and I appreciate the help with this issue.

However, by removing the `blockingOperation` calls, expensive blocking file operations will be scheduled on a  non-blocking thread, which would introduce serious performance problems. So despite the fact that this PR resolves #31217, I am afraid that we need to reject it, as merging it would introduce new issues.

Issue #32491, opened 2024-03-19
@Transactional of the target class is not considered for methods defined in a subclass


markusheiden:
If a (public) method is declared at a superclass of an `@Transactional` target class, the method won't be wrapped in a transaction.

The problem lies in `AbstractFallbackTransactionAttributeSource`:
```java
	protected TransactionAttribute computeTransactionAttribute(Method method, @Nullable Class<?> targetClass) {
		// Don't allow non-public methods, as configured.
		if (allowPublicMethodsOnly() && !Modifier.isPublic(method.getModifiers())) {
			return null;
		}

		// The method may be on an interface, but we need attributes from the target class.
		// If the target class is null, the method will be unchanged.
                // -> This will resolve to the method at the superclass. This method has no @Transactional.
		Method specificMethod = AopUtils.getMostSpecificMethod(method, targetClass);

		// First try is the method in the target class.
                // -> This will resolve to the method at the superclass too because the method has not been overwritten.
		TransactionAttribute txAttr = findTransactionAttribute(specificMethod);
		if (txAttr != null) {
			return txAttr;
		}

		// Second try is the transaction attribute on the target class.
                // -> This will resolve to the superclass. The superclass has no @Transactional.
		txAttr = findTransactionAttribute(specificMethod.getDeclaringClass());
		if (txAttr != null && ClassUtils.isUserLevelMethod(method)) {
			return txAttr;
		}

                // Bug: This third case is missing:
                // -> The target class has @Transactional.
		txAttr = findTransactionAttribute(targetClass);
		if (txAttr != null && ClassUtils.isUserLevelMethod(method)) {
			return txAttr;
		}

		if (specificMethod != method) {
			// Fallback is to look at the original method.
			txAttr = findTransactionAttribute(method);
			if (txAttr != null) {
				return txAttr;
			}
			// Last fallback is the class of the original method.
			txAttr = findTransactionAttribute(method.getDeclaringClass());
			if (txAttr != null && ClassUtils.isUserLevelMethod(method)) {
				return txAttr;
			}
		}

		return null;
	}
```

jhoeller:
[...]

PR #32492, opened 2024-03-19
@Transactional of the target class is not considered for methods defined in a superclass

markusheiden:
Fix for #32491


jhoeller:
See https://github.com/spring-projects/spring-framework/issues/32491#issuecomment-2051884041

PR #32507, opened 2024-03-21
Add Reactive support to Spring AOP

kse-music:
original [issue ](https://github.com/spring-projects/spring-security/issues/14731)


snicoll:
Thanks for the PR but let's keep the discussion in Spring Security for now. If we'd have to address something, it wouldn't be done as you suggested anyway.

PR #32511, opened 2024-03-22
Skip comparing variableResolver if underlying type is Class

quaff:
Before this commit, given:
```java
ResolvableType type1 = ResolvableType.forClassWithGenerics(IProvider.class, String.class);
ResolvableType type2 = ResolvableType.forClassWithGenerics(IProvider.class,ResolvableType.forClass(StringProvider.class).as(IProvider.class).getGenerics());
assertThat(type1).isEqualTo(type2);
```
will cause:
```
Expected :org.springframework.core.ResolvableTypeTests$IProvider<java.lang.String>
Actual   :org.springframework.core.ResolvableTypeTests$IProvider<java.lang.String>
```


snicoll:
Thanks for the PR but you should be using `equalsType` for that type of comparison. The regular `equals` implementation is rather strict since it is primarily used for cache lookups, not for type comparisons.

PR #32521, opened 2024-03-22
Apply instanceof pattern matching in ASM fork

FelixDes:
This pull request applies the instanceof pattern matching on spring-core module. This enhancement will improve the code readability.


sbrannen:
Hi @FelixDes,

Thanks for the PR.

Unfortunately, these are ASM classes, and the following is our policy on such classes.

Issue #32129, opened 2024-01-26
Harmonize ChannelRegistration to use Executor rather than TaskExecutor


snicoll:
With the fix made in #32081, we noticed that `ChannelRegistration` requires a `TaskExecutor` but what it really needs is an `Executor` really. We'll use this issue to harmonize the types.

PR #32524, opened 2024-03-23
Fix Javadoc `@since` tag for `ChannelRegistration.executor()`

izeye:
The signature for the `ChannelRegistration.executor()` has been changed in f526b23fd728ea3eb6cf7245142f4563df34dbbd, but its Javadoc `@since` tag update seems to have been missed.

See gh-32129


sbrannen:
> The signature for the `ChannelRegistration.executor()` has been changed in [f526b23](https://github.com/spring-projects/spring-framework/commit/f526b23fd728ea3eb6cf7245142f4563df34dbbd), but its Javadoc `@since` tag update seems to have been missed.

Although that's true, it was actually introduced in 6.1.4 with the wrong parameter type.

So, perhaps #32129 should have been backported to `6.1.x` as a bug fix?

Let's see what the rest of the team thinks.

snicoll:
That's not accurate. It was introduced this way on purpose to facilitate the migration from 6.1.x to 6.2.x, that is with the executor term but taking a `TaskExecutor` in 6.1.x.

PR #32532, opened 2024-03-26
Prevent package summaries from being truncated due to incorrect sentence detection

deejay1:
Javadoc doesn't seem to like having `e.g.` as it thinks the sentence ends there which is usually incorrect and results in broken descriptions. Replacing the breakable space with a nonbreakable one works around it.


snicoll:
@deejay1 thanks for the report but I am not a fan of the change. Rather, we could update the phrasing to avoid the use of ,".e.g." or make the first sentence shorter (that's definitely the case for one of them). Do you want to amend your proposal?

deejay1:
@snicoll yeah, I'm not that much of a fan either, I'll see if I can come up with something better

deejay1:
@snicoll @sbrannen I reworded the doc strings a little bit, should be better now.

sbrannen:

Thanks

Issue #32113, opened 2024-01-25
DataBufferLimitException in WebFlux controllers should result in 413 instead of 500


singhbaljit:
When using Spring WebFlux, there is a maximum limit on the payload size, defined by `spring.codec.max-in-memory-size` (default value is 256KB). If a request is made with a body exceeding this size, `DataBufferLimitException` is thrown. The default error handler then returns a `500 Internal Server Error` response. This is actually not a server side error, but instead a server side validation. This should result in `413 Content Too Large` instead.

Spring Boot version: 3.2.1

singhbaljit:
Few more thoughts: the error handler probably shouldn't map `DataBufferLimitException` to `413` directly, as there can be other reasons for the exception. For example, `WebClient` can also throw this exception, which likely is a server side error. I suppose we should map (i.e. `onErrorMap`) to another exception that always result in a `413`. But I'm not that familiar with the spring-framework code, so just guessing. 

Also, I do not know if what if this happens in WebMVC all well; that is a 500 instead of a 413, but it should be straightforward to validate.

PR #32558, opened 2024-03-31
Handle DataBufferLimitException as HTTP 413 responses

mysend12:
origin issue: https://github.com/spring-projects/spring-framework/issues/32113

hello. In WebFlux, when a DataBufferLimitException occurs because the payload is larger than the setting value (spring.codec.max-in-memory-size), a 413 Too Large Body status code should be returned, but a 500 status code was returned, so that part has been corrected.

1. A ServerWebInputTooLargeException exception was declared, which inherits ResponseStatusException.
2. When determining whether there is an error in the request value in AbstractMessageReaderArgumentResolver, if a DataBufferLimitException occurs, it is converted to ServerWebInputTooLargeException.
3. Added tooLargeBody test code.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/ServerWebInputTooLargeException.java
+++ b/spring-web/src/main/java/org/springframework/web/server/ServerWebInputTooLargeException.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


mamadaliev (approved):
LGTM

spring-web/src/main/java/org/springframework/web/server/ServerWebInputTooLargeException.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.springframework.http.HttpStatus;
+import org.springframework.lang.Nullable;
+
+/**
+ * Exception for errors that fit response status 413 (payload too large) for use in
+ * Spring Web applications.
+ *
+ * @author Kim Bosung
+ * @since 5.0
```
injae-kim, line 28:
```suggestion
 * @since 6.2
```
since 6.2 is correct?

injae-kim:
origin issue: #32113

- Can you update PR description and commit message to include "Fixes #32113" ?
- Can you update "copywright 2024" on top of each files?

snicoll:
@injae-kim thanks for trying to help but please refrain from providing guidelines to contributors. The first one is actually wrong, we do not need a reference to the original issue in the commit message.

injae-kim:
oh sorry, I got it! actually PR author @mysend12 is my friend and I'm helping him to contribute :)

bclozel:
Thanks for your contribution @mysend12 , this is now merged in main and will be shipped with the next 6.2 milestone.

PR #32560, opened 2024-04-01
Remove link to unrelated method in javadoc of AnnotatedBeanDefinitionReader

quaff:
It should link `doRegisterBean(Class<T> beanClass, String name, Class<? extends Annotation>[] qualifiers, Supplier<T> supplier, BeanDefinitionCustomizer[] customizers)` instead of `registerBean(Class, String, Class...)` but it is private, to make things more simple I'd like to remove it instead of linking all variants of `registerBean()`.


snicoll:
Yeah, I'd rather remove the link altogether. Thanks again!

PR #32568, opened 2024-04-03
Add create ClientHttpRequestFactory documentation

rentius2407:
This is a proposal to add an explanation on how to create a ClientHttpRequestFactory with settings (connect and read timeout) and building a RestClient using the configured request factory.


bclozel:
`ClientHttpRequestFactories` belongs to Spring Boot, we can't really document that here. It's [documented here](https://docs.spring.io/spring-boot/docs/current/reference/html/io.html#io.rest-client.restclient.ssl). Thanks for the proposal, but I'm declining this PR.

PR #32570, opened 2024-04-03
Support ordering of AOP advice annotations at the method level

cbxbj:
AOP advice annotations sorting is more flexible


snicoll:
Thanks for the suggestion but please consider raising an issue to discuss the change before actually working on it. `@Order` is meant to be used to order "types" and/or components, not methods. It's available at the method level in conjunction with `@Bean`.

Issue #29491, opened 2022-11-15
Improve 404 error handling in Spring MVC


rstoyanchev:
There are two places in Spring MVC that can call `sendError(SC_NOT_FOUND)`. One is the `DispatcherServlet` and the other `ResourceHttpRequestHandler`. This prevents applications from handling 404 errors, and also prevents the possibility for default error handling with an RFC 7807 problem details. By contrast, WebFlux raises `ResponseStatusException(NOT_FOUND)` for both cases.

The `DispatcherServlet` can be configured to raise `NoHandlerFoundException` through its `throwExceptionIfNoHandlerFound` property, but by default it's false. For `ResourceHttpRequestHandler` there is no way to customize the behavior.

The goal for this issue is to phase out hard-coded use of `sendError` and allow consistent 404 handling:

1. Set the `throwExceptionIfNoHandlerFound` property to `true` by default, and deprecate it for eventual removal. This will result in raising `NoHandlerFoundException` by default.
2. Update `ResourceHttpRequestHandler` to raise a similar `NoResourceFoundException`, and provide default handling in `ResponseEntityExceptionHandler` as well as in `DefaultHandlerExceptionResolver`.

This should help to address https://github.com/spring-projects/spring-boot/issues/7653. Boot will also need to update its own `throwExceptionIfNoHandlerFound` property that enforces a `false` value by default on `DispatcherServlet`.

rstoyanchev:
An additional fix in #30930 ensures built-in handling of `NoResourceFoundException`.

PR #32574, opened 2024-04-03
Update javadoc for NoHandlerFoundException after 6.1 changes

SanderHestvik:
Reflect change of default throwExceptionIfNoHandlerFound after 6.1


snicoll:
See #29491

snicoll:
Thanks for the PR @SanderHestvik but I don't think we should be documenting a deprecated method the way you suggest. Rather, the new default should be configured, and the deprecated behavior should be left out. I've done that in https://github.com/spring-projects/spring-framework/commit/d955549037e632b76c7b716855bffcd673ffda04

PR #32628, opened 2024-04-12
Fix typos in comment and test

laterlaugh:


spring-core/src/main/java/org/springframework/cglib/proxy/BridgeMethodResolver.java
```diff
@@ -110,7 +110,7 @@ public void visitMethodInsn(
                                 && currentMethod != null) {
                             Signature target = new Signature(name, desc);
                             // If the target signature is the same as the current,
-                            // we shouldn't change our bridge becaues invokespecial
+                            // we shouldn't change our bridge because invoke special
```
sbrannen, line 113: Please note that `invokespecial` is an instruction in the JVM.

Thus, the comment is correct as it is.

So I'll remove that change when merging the PR.

sbrannen:

Thanks

PR #32634, opened 2024-04-14
Use `hasText()` instead of `hasLength()` in `AbstractAutoProxyCreator`

dukbong:
In the context of the code, using the hasText method instead of hasLength seems more appropriate.
This is because "beanName" should consist of non-empty, non-whitespace characters.


sbrannen:
> This is because "beanName" should consist of non-empty, non-whitespace characters.

That's correct: the bean name should not be "empty".

However, checks for that should be applied earlier in the bean definition creation/processing phases.

Consequently, we do not need to apply such checks in components like `AbstractAutoProxyCreator`.

Thanks anyway!

PR #32635, opened 2024-04-14
Modify error message when timeout is less than TIMEOUT_DEFAULT

remeio:
When `timeout` is zero(0), this method `setTimeout` runs normally, it will not throw the `IlleglArgumentException`. In other words, zero is valid timeout for this method. The valid arguments:
- TIMEOUT_DEFAULT(-1)
- Zero(0)
- Positive integer(>=1)

So, I think `non-negative` is better in this method.

(Actually when timeout is zero, it will throw `TransactionTimedOutException` on transaction starting)


simonbasle:
A zero timeout is indeed a valid input, and it is up to the underlying transaction implementation to interpret this.
Of particular note is how this gets interpreted if the 0 is passed down to a Jakarta `UserTransaction`:
>  If the value is zero, the transaction service restores the default value. If the value is negative a SystemException is thrown.

simonbasle (approved):

PR #32637, opened 2024-04-15
Optimized @Cacheable in multiple layer caches

liaolunhui:
Let's say you have two level caches.
layer 1: local in-memory cache, faster speed but shorter time to live.
layer 2: Redis cache, slower speed bug longer time to live.

You can using multiple `@Cacheable` annotation like below:

` @Caching(cacheable = {@Cacheable(cacheManager = "local", cacheNames = "cdnLog", key = "#clientIp"),
            @Cacheable(cacheManager = "redis", cacheNames = "cdnLog", key = "#clientIp")})
int countByClientIpAndDateBetween(String clientIp);`

We are expecting if local in-memory cache is null and Redis cache exists, use Redis cache to fill local in-memory cache.
But it is not currently, it only put caches when all layer caches are null.
This pull request optimized as our expectation.

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
@@ -548,8 +548,19 @@ public abstract class CacheAspectSupport extends AbstractCacheInvoker
 
 		// Collect puts from any @Cacheable miss, if no cached value is found
 		List<CachePutRequest> cachePutRequests = new ArrayList<>(1);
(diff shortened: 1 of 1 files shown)


quaff:
You should always include tests to cover your changes.
And it maybe off topic, how do you evict local cache if it's updated in another instance.

snicoll:
> This pull request optimized as our expectation.

@liaolunhui thanks for the PR but you can't change the behavior to your expectation without a validation that the previous behavior is wrong. The current behavior is the expected behavior (do not invoke the method if the key computes to an entry) and it's not the job of the abstraction to copy data over. If you want to implement such complex cache arrangement, you should research a cache library that allows to sync them.

PR #32646, opened 2024-04-16
Prevent potential NullPointerException.

dukbong:
Contribution summary: Added null check to prevent potential NullPointerException in visitIndexedArgumentValues method.


sbrannen:
The map supplied to `visitIndexedArgumentValues()` cannot be `null`.

Thus the change in this PR is invalid.

Please refrain from introducing `@Nullable` where it is not applicable.

PR #32647, opened 2024-04-16
Omit empty resource description in `DeprecatedBeanWarner`'s log message

dukbong:
In this context, the hasText() method seems more appropriate. This is because considering cases where the resourceDescription string contains only whitespace ensures clearer logging, as it accounts for scenarios where the string is not empty but consists solely of whitespace characters.


sbrannen:

Thanks

dukbong:
> This has been merged into `main`.
> 
> Thanks

Thanks! @sbrannen

PR #32648, opened 2024-04-16
Refactor ContentDisposition class

onjik:
### 1st Commit (8293b4b1b4313f81d20c646525b6f7661c31950d)
- In the old code, the code returns the same object every time, but it was always creating the object.
- In the new code, I improved this by making the constant return.
- Update License Date Range(https://github.com/spring-projects/spring-framework/wiki/Code-Style#license)
- Add Test code for empty
- Add `boolean isEmpty()` for check is empty Content Disposition(In the old code, you should have checked everything or called a heavy toString() method.)

### 2nd Commit (876e41cbc073917360fd90c8ff1b8d2b6a07d79d)
- I added null check on static parse method, like the rest of the static methods.
- The old code had the possibility of NPE occurring.

### 3rd Commit (7c9d1154784a9b6bcae05421eebe986b28170dd8)
- In the old code, the type is used as a simple string, so there was a possibility of human error.
```java
// like
ContentDisposition.builder("iline")
```
- So I added a String-based constant value, and a simple annotation so that beginners don't panic.
- And I also Add `@see` java doc tag on static builder method
```java
// you can use like this
ContentDisposition.builder(Type.FORM_DATA)
```
---
Overall, there is no change in the code interface, but I improve the way it works internally. Everything works the same way as before.

Thank you !!

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
+++ b/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
@@ -347,6 +347,7 @@ public final class ContentDisposition {
 	 * @see #toString()
 	 */
 	public static ContentDisposition parse(String contentDisposition) {
+		Assert.notNull(contentDisposition, "'contentDisposition' must not be null");
 		List<String> parts = tokenize(contentDisposition);
 		String type = parts.get(0);
 		String name = null;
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal, but I don't think this improves the codebase nor make the implementation more safe.

PR #32652, opened 2024-04-16
Make BodyBuilder.bodyValueAndAwait() a reified type of function

GeorgePap-719:
## Summary

`ServerResponse.BodyBuilder.bodyValueAndAwait` API is broken when it is used with an application it
uses for serialization the `kotlinx.serialization` library and the provided body uses generics.
The library is unable to serialize the body since it is a type of `Any` and spring falls back
to default one. This could lead to unexpected responses.

For example:

```kotlin
@Serializable
data class SomeBody(@SerialName("user_id") val userId: Int, val name: String)

suspend fun returnBody(request: ServerRequest): ServerResponse {
    return ServerResponse
        .ok()
        // Note: `fun ServerResponse.BodyBuilder.bodyValueAndAwait(body: Any)`
        // is not a reified function and is subject to type erasure.
        // In cases where response is a collection, the object is serialized as `Collection<Any>`,
        // and the response does not have the proper format, which is not expected.
        // Actual response: [{"userId":1,"name":"name"}]
        // Expected response : [{"user_id":1,"name":"name"}]
        .bodyValueAndAwait(listOf(SomeBody(1, "name")))
}
```

A minimum reproducible example can be found [here](https://github.com/GeorgePap-719/interview-template/blob/reproduce-spring-issue/src/main/kotlin/ServerResponseRouter.kt).

## Proposal

Make BodyBuilder.bodyValueAndAwait() a reified type of function. We can also migrate the API without
breaking existing functionality.

## Alternatives

Only document the behavior and accept it as the intended behavior.

### DIFF
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/ServerResponseExtensions.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/ServerResponseExtensions.kt
@@ -20,6 +20,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.reactor.awaitSingle
 import org.reactivestreams.Publisher
 import org.springframework.core.ParameterizedTypeReference
+import org.springframework.core.ReactiveAdapterRegistry
 import org.springframework.http.MediaType
 import reactor.core.publisher.Mono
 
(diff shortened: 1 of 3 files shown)


spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/ServerResponseExtensions.kt
```diff
@@ -54,14 +55,69 @@ inline fun <reified T : Any> ServerResponse.BodyBuilder.body(producer: Any): Mon
...
  * @throws IllegalArgumentException if `body` is a [Publisher] or an
- * instance of a type supported by [org.springframework.core.ReactiveAdapterRegistry.getSharedInstance],
+ * instance of a type supported by [org.springframework.core.ReactiveAdapterRegistry.getSharedInstance].
  */
+@Suppress("DeprecatedCallableAddReplaceWith")
+@Deprecated(
+	message = "Use `bodyValueAndAwait<T>()` instead, which is not subject to type erasure.",
+	level = DeprecationLevel.WARNING
+)
+@JvmName("bodyValueAndAwait0") // to avoid platform declaration clash
```
sdeleuze, line 95: Can you provide more details on the issue you see if you remove this `@JvmName("bodyValueAndAwait0")` annotation?

GeorgePap-719: If you remove this annotation then the compilation will fail with error:
```text
Platform declaration clash: The following declarations have the same JVM signature (bodyValueAndAwait(Lorg/springframework/web/reactive/function/server/ServerResponse$BodyBuilder;Ljava/lang/Object;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;):
    suspend inline fun <reified T : Any> ServerResponse.BodyBuilder.bodyValueAndAwait(body: T, `$completion`: Continuation<ServerResponse>): Any? defined in org.springframework.web.reactive.function.server
    suspend fun ServerResponse.BodyBuilder.bodyValueAndAwait(body: Any, `$completion`: Continuation<ServerResponse>): Any? defined in org.springframework.web.reactive.function.server
```

When the bytecode is generated the generics are erased so the compiler just sees two functions with the same type, because in the end `T` is replaced with `Any`.

GeorgePap-719:
Sorry for force-push while in-review. Minor changes in KDocs and error message.

PR #32655, opened 2024-04-17
Consolidate filter handling: applyFilters.

dukbong:
This PR consolidates the functionality of the previous methods, ***enableFilters*** and ***disableFilters***, which were performing similar tasks, thus eliminating redundancy.

The new ***applyFilters*** method integrates these two operations into one, enhancing code reusability and reducing duplication. This change makes the code more concise and easier to maintain. Additionally, the new method accepts an ***enable parameter*** to determine whether to enable or disable the filters, allowing both actions to be handled within a single method.


bclozel:
I don't think this improves the code.
Please avoid submitting PRs for stylistic changes in the future.

dukbong:
> I don't think this improves the code. Please avoid submitting PRs for stylistic changes in the future.

I apologize. I won't submit PRs for stylistic changes or personal issues in the future.

PR #32656, opened 2024-04-17
Synchronize access to volatile variables in AsyncRequestInterceptor

dukbong:
This PR introduces synchronized blocks to synchronize access to volatile variables in the AsyncRequestInterceptor class. Previously, modifications to the volatile variables timeoutInProgress and errorInProgress were not synchronized, potentially leading to race conditions and inconsistent behavior in a multi-threaded environment.

By enclosing access to these variables within synchronized blocks, this PR ensures proper synchronization and consistency. This enhancement improves the reliability and stability of the AsyncRequestInterceptor class, particularly benefiting scenarios involving asynchronous web requests.


bclozel:
Thanks for the proposal, but I believe synchronization is performed externally in this case.
Please avoid sending PRs for hypothetical issues that you have encountered or replicated yourself.

PR #32662, opened 2024-04-18
Add implementation of the isCandidateClass method to do full introspection

vjh0107:
Hello,

This PR implements the method `isCandidateClass`, which introspects annotations at the type, method, and field levels to determine if the provided annotation is applied. Although it was mentioned in the comment, it was not implemented, so I have done it!

I ensured that adhered to the existing Spring team's code style and made minor changes to ReflectionUtils to align with the reflection code style.

Since this is my second PR, there might be issues. If there are any problems, I would appreciate the opportunity to fix them.

Thank you!

Screenshot 2024-04-18 at 10 51 09

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationUtils.java
@@ -19,22 +19,26 @@ package org.springframework.core.annotation;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Array;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Proxy;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import org.springframework.core.BridgeMethodResolver;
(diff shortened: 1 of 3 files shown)


jhoeller:
I'm afraid this is not the actual intention behind `isCandidateClass`: That method is rather meant to make quick decisions to rule out classes which will never carry the given annotation to begin with. If `isCandidateClass` returns `true`, the caller will then proceed with a full annotation search depending on the semantics of the given annotation. On the other hand, if `isCandidateClass` returns `false`, the caller will entirely skip that search. It's effectively a simple performance optimization, skipping the expensive reflection interaction for a given class if it will never reveal any such annotation at all.

vjh0107:
Thank you for your reply.
According to the comments, it is commented that the analysis should be performed at type, method, or field level. but there is no logic introspecting at type, method, or field level. Looking at where the isCandidateClass method is used, it seems that after checking through the annotation, the logic I performed was executed again, which does not seem to cause performance issues. The most important thing is that it does not perform the action corresponding to the method name. Could you please review it again when you have time?
@jhoeller

sbrannen:
> According to the comments, it is commented that the analysis should be performed at type, method, or field level. 

The Javadoc does not state that.

The Javadoc states that the method determines the the supplied class is a "candidate" for carrying the specified annotation. The Javadoc does not state that the method will actually check at the method or field level.

In summary, this method makes a best effort attempt to guess if certain annotations are likely to be present in a given class.


> it seems that after checking through the annotation, the logic I performed was executed again, which does not seem to cause performance issues. 

Doing the same thing twice is always a performance issue, simply because there is no need to do the same thing twice.

Please keep in mind that this is an internal utility method that is only intended to be used within the Framework itself, and the maintainers are aware of what this method is intended to do and how it should be used.

vjh0107:
Thank you for your kind explanation. @sbrannen

PR #32673, opened 2024-04-19
Remove redundant conditions and enhance safety with List.of() method

dukbong:
- Revised the portion where null check was performed after `Assert.notNull()`, ***eliminating redundant null checks***
- Enhanced code safety by replacing `Arrays.asList()` with `List.of()` to ***create immutable lists***


sbrannen:
@dukbong, please refrain from opening PRs and then closing them and replacing them with additional PRs, since this introduces unnecessary noise in the issue tracker.

PR #32674, opened 2024-04-19
Remove redundant conditions

dukbong:
Revised the portion where null check was performed after Assert.notNull(), eliminating redundant null checks.

### DIFF
--- a/spring-core/src/main/java/org/springframework/aot/hint/ExecutableHint.java
+++ b/spring-core/src/main/java/org/springframework/aot/hint/ExecutableHint.java
@@ -127,9 +127,10 @@ public final class ExecutableHint extends MemberHint implements Comparable<Execu
 		 * @param mode the required mode
 		 * @return {@code this}, to facilitate method chaining
 		 */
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/aot/hint/ExecutableHint.java
```diff
@@ -127,9 +127,10 @@ public static class Builder {
 		 * @param mode the required mode
 		 * @return {@code this}, to facilitate method chaining
 		 */
+		@SuppressWarnings("NullAway")
 		public Builder withMode(ExecutableMode mode) {
 			Assert.notNull(mode, "'mode' must not be null");
-			if ((this.mode == null) || !this.mode.includes(mode)) {
+			if (!this.mode.includes(mode)) {
```
sbrannen, line 133: `mode` and `this.mode` refer to two different objects.

Thus, the proposed change is invalid.

sbrannen:
Closing since the proposed change is invalid.

PR #32675, opened 2024-04-19
Enhance safety with List.of() method.

dukbong:
Enhanced code safety by replacing Arrays.asList() with List.of() to create immutable lists


snicoll:
> Enhanced code safety

The list is not exposed anywhere so I don't understand the argument. Brian already asked you to not submit such PR, which adds maintenance overhead on our side. I'd expect many other case where this change would need to be applied if we were to accept this so this would also add inconsistency to the codebase.

PR #32704, opened 2024-04-25
Support varargs invocations in SpEL for varargs array subtype

LeMikaelF:
There was a bug in SpEL where an array passed into a varargs method could be wrapped one too many times if the component type of the argument didn't match the component type of the parameter. For example:

```java
@Test
void givenArrayFromVariable_whenInvokeVarargsMethod_thenCorrectArguments() {
	ExpressionParser parser = new SpelExpressionParser();
	StandardEvaluationContext context = new StandardEvaluationContext();
	Recorder recorder = new Recorder();
	context.setVariable("recorder", recorder);
	context.setVariable("myArray", new String[] { "a", "b" });

	parser.parseExpression("#recorder.record(#myArray)").getValue(context);

	assertThat(recorder.getArgs()).asInstanceOf(ARRAY).containsExactly("a", "b");
}

private static class Recorder {
	@Nullable private Object[] args;

        // test would pass with (String... args)
	public void record(Object... args) {
		this.args = args;
	}
	
	@Nullable
	public Object[] getArgs() {
		return args;
	}
}
```

This fails with the following error message:

```
Expecting actual:
  [["a", "b"]]
to contain exactly (and in same order):
  ["a", "b"]
```

I've added more exhaustive test cases for the method where the bug was, as well as a test in `SpelCompilationCoverageTests`. I've added this last test just next to another test that was commented out with a TODO, and had to comment out part of my test because of the same bug, but fixing this exposes another underlying bug that I've fixed in another branch, and I was able to uncomment the tests that were commented out. The fix is ready, I'm just waiting for this one to be merged.

I may also have accidentally uncovered another bug in SpEL. At line 4651 in `SpelCompilationCoverageTests`, the method `seventeen()` isn't actually invoked with `()`, yet it's still executed and its result value is available. I'm not sure this is desirable behaviour.

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
@@ -36,6 +36,8 @@ import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.MethodInvoker;
 
+import static org.springframework.util.ObjectUtils.isArray;
+
 /**
  * Utility methods used by the reflection resolver code to discover the appropriate
  * methods, constructors, and fields that should be used in expressions.
@@ -457,14 +459,18 @@ public abstract class ReflectionHelper {
 	 * @return a repackaged array of arguments where any varargs setup has been done
 	 */
 	public static Object[] setupArgumentsForVarargsInvocation(Class<?>[] requiredParameterTypes, Object... args) {
-		// Check if array already built for final argument
 		int parameterCount = requiredParameterTypes.length;
+		Assert.notEmpty(requiredParameterTypes, "Required parameter types must not be empty");
+
+		Class<?> lastRequiredParameterType = requiredParameterTypes[parameterCount - 1];
+		Assert.isTrue(lastRequiredParameterType.isArray(), "Method must be varargs");
+
 		int argumentCount = args.length;
+		Object lastArgument = argumentCount > 0 ? args[argumentCount - 1] : null;
 
 		// Check if repackaging is needed...
 		if (parameterCount != args.length ||
-				requiredParameterTypes[parameterCount - 1] !=
-						(args[argumentCount - 1] != null ? args[argumentCount - 1].getClass() : null)) {
+			(!isArray(lastArgument) && differentTypes(lastRequiredParameterType, lastArgument))) {
 
 			// Create an array for the leading arguments plus the varargs array argument.
 			Object[] newArgs = new Object[parameterCount];
@@ -477,7 +483,7 @@ public abstract class ReflectionHelper {
 			if (argumentCount >= parameterCount) {
 				varargsArraySize = argumentCount - (parameterCount - 1);
 			}
-			Class<?> componentType = requiredParameterTypes[parameterCount - 1].componentType();
+			Class<?> componentType = lastRequiredParameterType.componentType();
 			Object varargsArray = Array.newInstance(componentType, varargsArraySize);
 			for (int i = 0; i < varargsArraySize; i++) {
 				Array.set(varargsArray, i, args[parameterCount - 1 + i]);
@@ -489,6 +495,9 @@ public abstract class ReflectionHelper {
 		return args;
 	}
 
+	private static boolean differentTypes(Class<?> lastRequiredParameterType, @Nullable Object lastArgument) {
+		return lastArgument == null || lastRequiredParameterType != lastArgument.getClass();
+	}
 
 	/**
 	 * Arguments match kinds.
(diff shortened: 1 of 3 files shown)


sbrannen:
Hi @LeMikaelF,

> This fails with the following error message:

I just copied and pasted the example from this issue's description into my IDE, and I cannot reproduce the error you've shown.

That test passes for me on `main`.

What versions of Spring Framework and Java are you using?

LeMikaelF:
My bad, I had two similar tests and copy-pasted the wrong one. The tests above passes because the bug only affects arrays, not lists. This is the failing test:

```java
@Test
void givenArrayFromVariable_whenInvokeVarargsMethod_thenCorrectArguments() {
	ExpressionParser parser = new SpelExpressionParser();
	StandardEvaluationContext context = new StandardEvaluationContext();
	Recorder recorder = new Recorder();
	context.setVariable("recorder", recorder);
	context.setVariable("myArray", new String[] { "a", "b" });

	parser.parseExpression("#recorder.record(#myArray)").getValue(context);

	assertThat(recorder.getArgs()).asInstanceOf(ARRAY).containsExactly("a", "b");
}
```

I will update the PR description.

dcollins123:
Could you add '*' before the array var so the elements become individual args? 
parser.parseExpression("#recorder.record(*#myArray)").getValue(context);

sbrannen:
> I may also have accidentally uncovered another bug in SpEL. At line 4651 in `SpelCompilationCoverageTests`, the method `seventeen()` isn't actually invoked with `()`, yet it's still executed and its result value is available. I'm not sure this is desirable behaviour.

That's to be expected: when `seventeen` is encountered in that expression without `()`, it is treated as a property reference that resolves to the `seventeen()` method in the root context object.

sbrannen:
> I've added this last test just next to another test that was commented out with a TODO, and had to comment out part of my test because of the same bug, but fixing this exposes another underlying bug that I've fixed in another branch, and I was able to uncomment the tests that were commented out. The fix is ready, I'm just waiting for this one to be merged.

Do you mean that you have a second fix for which you plan to submit a PR that will address the following TODO regarding compilation of varargs method invocations?

```java
// TODO Determine why the String[] is passed as the first element of the Object... varargs array instead of the entire varargs array.
```

sbrannen:
> Could you add `*` before the array var so the elements become individual args?

@dcollins123, `*` is not proper syntax for a SpEL expression.

spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
```diff
@@ -457,14 +459,18 @@ private static boolean isFirstEntryInArray(Object value, @Nullable Object possib
...
+		Assert.isTrue(lastRequiredParameterType.isArray(), "Method must be varargs");
+
 		int argumentCount = args.length;
+		Object lastArgument = argumentCount > 0 ? args[argumentCount - 1] : null;
 
 		// Check if repackaging is needed...
 		if (parameterCount != args.length ||
-				requiredParameterTypes[parameterCount - 1] !=
-						(args[argumentCount - 1] != null ? args[argumentCount - 1].getClass() : null)) {
+			(!isArray(lastArgument) && differentTypes(lastRequiredParameterType, lastArgument))) {
```
sbrannen, line 473: This can be simplified by using `java.lang.Class.isInstance(Object)`.

However, there is no need to update this PR: I'll do that when merging.

spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
```diff
@@ -36,6 +36,8 @@
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.MethodInvoker;
 
+import static org.springframework.util.ObjectUtils.isArray;
```
sbrannen, line 39: We do not use static imports for utility methods in the core Spring Framework.

sbrannen (approved):
Thanks for the raising the issue, the PR itself, and for the additional tests for var-args invocations.

I left a few comments which I will address locally before merging.

spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectionHelper.java
LeMikaelF: Thank you, I didn't think of that!

LeMikaelF:
> Do you mean that you have a second fix for which you plan to submit a PR that will address the following TODO regarding compilation of varargs method invocations?

Yes, exactly.

sbrannen:
> > Do you mean that you have a second fix for which you plan to submit a PR that will address the following TODO regarding compilation of varargs method invocations?
> 
> Yes, exactly.

@LeMikaelF, that would be great!

PR #32717, opened 2024-04-27
feat: contribution workflow added

opensourcevk:
testing evidence:
image


snicoll:
Thanks for the suggestion but we don't need this

PR #32733, opened 2024-04-30
Fix incorrect class reference syntax in Kotlin code sample

Seungpang:
Hi Team,

I noticed an incorrect class reference in the documentation, and I've attempted to correct it.


sbrannen:
This has been merged into `6.1.x` and `main`.

Thanks

sdeleuze:
Refined to use the related Kotlin extension.

PR #32745, opened 2024-05-01
Sort AspectJ annotated advice methods using `@Order`

Puppy4C:
This PR provides the ability to use `@Order` sorting for Advice Annotation.

Example:

```java
@Aspect
public class MyAspect {

    @Order(1)
    @Before("execution(* org.example.AopObj.*(..))")
    public void method1() {
        System.out.println("Before method-1");
    }

    @Order(2)
    @Before("execution(* org.example.AopObj.*(..))")
    public void method2() {
        System.out.println("Before method-2");
    }
    
}
```

`method1()` will be executed before `method2()` because the sequence number of `method1()` is smaller.


snicoll:
Thanks for the PR. This was declined in #29732 and #32570

PR #32755, opened 2024-05-03
Added read-committed transaction isolation and consistency support to TransactionAwareCacheDecorator

werner77:
The decorator now maintains consistency within the transaction and ensures all changes are only
committed to the target cache on successful completion of the transaction.

- Added complete isolation and consistency support for concurrent transactions.
- Added accompanying Javadoc documentation.
- Added unit test testing the isolation properties introduced.
- Maintained backwards compatibility (existing unit test is still there and succeeds)

See also issue: https://github.com/spring-projects/spring-framework/issues/17353

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/transaction/TransactionAwareCacheDecorator.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/transaction/TransactionAwareCacheDecorator.java
@@ -23,7 +23,7 @@ import org.springframework.transaction.support.TransactionSynchronization;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


spring-context-support/src/main/java/org/springframework/cache/transaction/TransactionAwareCacheDecorator.java
```diff
@@ -79,18 +111,50 @@ public Object getNativeCache() {
...
 	@Override
 	@Nullable
 	public <T> T get(Object key, @Nullable Class<T> type) {
-		return this.targetCache.get(key, type);
+		final var transactionState = getTransactionState();
+		final var wrapper = transactionState == null ? null : transactionState.get(key);
+		if (wrapper != null) {
+			// Unwrap
+			final var value = convertValue(wrapper);
+			return cast(value, type);
```
GovindarajanL, line 131: Potential NPE, consider handling the case where convertValue(wrapper) returns null before calling cast(value, type).

PR #32764, opened 2024-05-05
Add `@Router` and `@Handler`

earlgrey02:
In general, annotation-based controller allows you to refer to controllers with unique annotations such as `@Controller`, while functional endpoint refers to routers and handlers with the traditional `@Configuration` and `@Component` annotations. I felt **there was a need for annotations to uniquely refer to routers and handlers more explicitly(
Whether used simply for writing explicit code or for AOP)**, so posted this PR.


snicoll:
Thanks for the PR but we don’t intend to create first class annotations like suggested. 

We prefer to provide a consistent way and let integrators create any specific stereotypes if needed.

Issue #28552, opened 2022-06-01
Deprecate trailing slash match and change default value from true to false


vpavic:
> Whether to match to URLs irrespective of the presence of a trailing slash. If enabled a method mapped to `"/users"` also matches to `"/users/"`.
> The default value is `true`.

Even though this behavior has been long present in Spring, it introduces ambiguity that can (combined with some other choices) easily have consequences in shape of security vulnerabilities. Consider this example:

```java
@SpringBootApplication
@RestController
public class SampleApplication {

    public static void main(String[] args) {
        SpringApplication.run(SampleApplication.class, args);
    }

    @GetMapping("/resources")
    String resources() {
        return "Hello from /resources";
    }

    @GetMapping("/resources/{id}")
    String resourceById(@PathVariable Long id) {
        return "Hello from /resources/" + id;
    }

    @Bean
    SecurityFilterChain filterChain(HttpSecurity httpSecurity) throws Exception {
        return httpSecurity
                .authorizeHttpRequests(requests -> {
                    requests.antMatchers("/resources").hasRole("admin");
                    requests.antMatchers("/resources/**").hasRole("user");
                    requests.anyRequest().denyAll();
                })
                .httpBasic(Customizer.withDefaults())
                .build();
    }

}
```

```properties
spring.security.user.password=password
spring.security.user.roles=user
```

Default user (with role `user`) will get `403` attempting to `GET /resources` but can avoid protection by issuing `GET /resources/`, which wouldn't be possible with trailing slash matching disabled.

Let me note that I'm aware that using `mvcMatchers` instead of `antMatchers` would have prevented this issue but that doesn't change the fact that there are many configurations out there relying on `antMatchers` and that sometimes `antMatchers` are simply more suitable for other reasons.

Also, I personally see little real benefit of having trailing slash matching enabled because:
- if application renders server-side generated views navigation is either way established using hyperlinks
- if application exposes APIs then even more so it is expected that requests are aligned with the API docs

For places where it's really needed for application to respond to both requests, I'd argue that it's either way better solution to configure redirects instead of application wide ambiguous request mappings.

Please consider making this change for `6.0`.

vpavic:
> From that perspective, the default is not unreasonable, and more than just HATEOAS style navigation or API guidance, even just manually typing a URL somewhere.

[...]

PR #32796, opened 2024-05-11
Align StandaloneMockMvcBuilder with trailing slash defaults

evan-mc:
Spring no longer supports trailing slashes by default, as of https://github.com/spring-projects/spring-framework/issues/28552. However, unit tests that are setup using the `StandaloneMockMvcBuilder` still have default support for trailing slashes. This causes an inconsistency that does not reflect what spring does by default. By changing the default to false, unit tests will accurately reflect the base implementation, which will cause less confusion when unit tests are working and production code is not.


sbrannen:
@rstoyanchev, what do you think about switching the default here?

rstoyanchev:
Yes, it makes sense to align the default, but maybe in 6.2 only.

PR #32804, opened 2024-05-13
Support varargs invocations in SpEL for varargs array subtype in compiled expressions

LeMikaelF:
This PR adds support for varargs invocations where the varargs argument is an array with a component type that is a subtype of the parameter varargs component type. For example, the method accepts an `Object... args` parameter, and it is passed a `String[]` argument. https://github.com/spring-projects/spring-framework/pull/32704 already added support for this in interpreted expressions. There were a few tests that were commented out because compiled expressions didn't support this yet, so I uncommented them and added more tests for functions, methods, and constructors.

I will be happy to address any comments on this PR.

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/ast/SpelNodeImpl.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/ast/SpelNodeImpl.java
@@ -16,9 +16,8 @@
 
 package org.springframework.expression.spel.ast;
 
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Member;
-import java.lang.reflect.Method;
+import java.lang.reflect.Executable;
+import java.util.Objects;
 import java.util.function.Supplier;
 
 import org.springframework.asm.MethodVisitor;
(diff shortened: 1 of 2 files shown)


spring-expression/src/main/java/org/springframework/expression/spel/ast/SpelNodeImpl.java
```diff
@@ -214,62 +213,67 @@ protected ValueRef getValueRef(ExpressionState state) throws EvaluationException
 	 * and if it is then the argument values will be appropriately packaged into an array.
 	 * @param mv the method visitor where code should be generated
 	 * @param cf the current codeflow
-	 * @param member the method or constructor for which arguments are being set up
+	 * @param executable the method or constructor for which arguments are being set up
 	 * @param arguments the expression nodes for the expression supplied argument values
 	 */
-	protected static void generateCodeForArguments(MethodVisitor mv, CodeFlow cf, Member member, SpelNodeImpl[] arguments) {
```
sbrannen, line 220: Since this is a `protected` method in a `public` type, we cannot simply change the method signature.

Instead, we would need to introduce an overloaded variant that accepts `Executable` and deprecate the existing variant that accepts `Member`.

spring-expression/src/main/java/org/springframework/expression/spel/ast/SpelNodeImpl.java
```diff
@@ -214,62 +213,67 @@ protected ValueRef getValueRef(ExpressionState state) throws EvaluationException
...
-				arrayType = arrayType.substring(1); // trim the leading '[', may leave other '['
-				// build array big enough to hold remaining arguments
-				CodeFlow.insertNewArrayCode(mv, childCount - p, arrayType);
-				// Package up the remaining arguments into the array
-				int arrayindex = 0;
-				while (p < childCount) {
-					SpelNodeImpl child = arguments[p];
+			int argCount = arguments.length;
+			String varargsType = paramDescriptors[paramCount - 1];
+			String varargsComponentType = varargsType.substring(1); // trim the leading '[', may leave other '['
```
sbrannen, line 234: Although I agree that these name changes make the code easier to follow, we should avoid including refactorings like this in a PR, since these changes make it much more challenging to follow what the PR is trying to achieve.

spring-expression/src/main/java/org/springframework/expression/spel/ast/SpelNodeImpl.java
```diff
@@ -214,62 +213,67 @@ protected ValueRef getValueRef(ExpressionState state) throws EvaluationException
...
 				}
 			}
+			else if (varargsType.equals(arguments[argCount - 1].getExitDescriptor())) {
+				// varargs type matches type of last argument
+				cf.generateCodeForArgument(mv, arguments[argCount - 1], paramDescriptors[paramCount - 1]);
+			}
+			else {
+				// last argument is an array and varargs component type is a supertype of argument component type
+				cf.generateCodeForArgument(mv, arguments[argCount - 1], varargsComponentType);
+			}
```
sbrannen, line 254: It's not clear to me how we can be certain that "varargs component type is a supertype of argument component type" is a valid assumption vs. wishful thinking.

I had thought we would rely on actual type information to make this decision, and I will experiment with that.

spring-expression/src/test/java/org/springframework/expression/spel/SpelCompilationCoverageTests.java
```diff
@@ -2277,31 +2285,33 @@ public static String join(String...strings) {
...
+		// single string argument
 		expression = parser.parseExpression("#doFormat('hey %s', 'there')");
 		context = new StandardEvaluationContext();
 		context.registerFunction("doFormat",
 				DelegatingStringFormat.class.getDeclaredMethod("format", String.class, Object[].class));
 		((SpelExpression) expression).setEvaluationContext(context);
 
 		assertThat(expression.getValue(String.class)).isEqualTo("hey there");
-		assertThat(((SpelNodeImpl) ((SpelExpression) expression).getAST()).isCompilable()).isTrue();
+		assertThat(((SpelExpression) expression).getAST().isCompilable()).isTrue();
```
sbrannen, line 2298: I see that you have made this change numerous times within this PR.

As I mentioned in a previous comment, we should avoid including refactorings like this in a PR, since these changes make it much more challenging to follow what the PR is trying to achieve.

sbrannen (approved):
Thanks for the proposed fix as well as all of the additional tests! 👍 

I've left some comments, but there's no need to address those in this PR at this time, since I plan to experiment and potentially make changes locally before merging this work.

LeMikaelF:
> Thanks for the proposed fix as well as all of the additional tests! 👍
> 
> I've left some comments, but there's no need to address those in this PR at this time, since I plan to experiment and potentially make changes locally before merging this work.

Thank you for the comments! I'll keep an eye out for the final fix.

sbrannen:
Current work on this PR can be viewed in the following feature branch which is based on this PR.

https://github.com/spring-projects/spring-framework/compare/main...sbrannen:spring-framework:issues/gh-32804-spel-varargs-array-subtype-compilation

sbrannen:
This has been merged into `main` in 29bb7b907c354ebc494afdfe38a0ed6f72f6272c and 12727a2c4f9ba6b7db3a5afa09a43419a5f2c6e7 and revised in 8fe4493a7d5af7df63165ceb8f6d984d7f285a3d.

To see the combined result, view the merged commit 061c13d367740ec144ca7eb89f54f0477db6aac7.

Thanks again for finding the source of the error and taking the initiative to submit a proposal with comprehensive test coverage! 👍

PR #32814, opened 2024-05-14
Add new method to set default headers in RestClient

doljae:
## Motivation

The `RestClient.Builder` has [a method](https://github.com/spring-projects/spring-framework/blob/376d7839e0fb6acf902d33eb578b1df143bcaf0b/spring-web/src/main/java/org/springframework/web/client/RestClient.java#L281) that takes a parameter of type `Consumer<HttpHeaders>` to set the default headers.

```kotlin
val defaultHeaders = HttpHeaders().apply {
    contentType = MediaType.APPLICATION_JSON
    setBasicAuth("encodedCredentials")
}

val restClient = RestClient.builder()
    .baseUrl("https://spring.io")
    .defaultHeaders {
        defaultHeaders.forEach { headerName, headerValue -> it[headerName] = headerValue }
    }
    .build()
```

In addition to this, it would be nice to have an intuitive method to set `HttpHeaders` directly, just like the methods in the common builder pattern.

## Modifications

- Add new method to set default headers in `RestClient`
```java
Builder defaultHeaders(HttpHeaders httpHeaders);
```

## Result

```kotlin
val defaultHeaders = HttpHeaders().apply {
    contentType = MediaType.APPLICATION_JSON
    setBasicAuth("encodedCredentials")
}

val restClient = RestClient.builder()
    .baseUrl("https://spring.io")
    .defaultHeaders(defaultHeaders)
    .build()
```
Now we can now intuitively set our custom `HttpHeaders` as the default header.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultRestClientBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultRestClientBuilder.java
@@ -277,6 +277,15 @@ final class DefaultRestClientBuilder implements RestClient.Builder {
 	}
 
 	@Override
+	public RestClient.Builder defaultHeaders(HttpHeaders httpHeaders) {
+		this.initHeaders();
+		if (this.defaultHeaders != null && httpHeaders != null) {
+			this.defaultHeaders.putAll(httpHeaders);
+		}
+		return this;
+	}
+
+	@Override
 	public RestClient.Builder defaultRequest(Consumer<RestClient.RequestHeadersSpec<?>> defaultRequest) {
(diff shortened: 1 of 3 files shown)


skystarxtogether:
There is one problem.

snicoll:
Thanks for the suggestion but we prefer to be in control over the `HttpHeaders` instance and offering a consumer or a way to set one header is a common pattern with `WebClient` and `RestClient`.

You can put all values using `.addAll`.

doljae:
Like you said, I didn't notice that we can use `addAll()` instead of iterating copy key-value to default header in `RestClient.builer()`.

Thanks your detailed explanation @snicoll 🙂

PR #32852, opened 2024-05-20
try to solve issue # 32043

lchwork:


mdeinum:
> We should also keep in mind that we may potentially decide to support multiple composed `@RequestMapping` annotations (e.g., `@PostMapping` and `@PutMapping`) on a single method _without_ making `@RequestMapping` itself a `@Repeatable `annotation.

According to the issue this should work *without making `@RequestMapping` an `@Repeatble`*. Your PR makes it an `@Repeatable`.

snicoll:
Thanks for trying @icha024 but the issue is flagged "pending-design-work" which means it needs our attention before crafting a solution. As such, this PR isn't actionable. Please consider the label next time and don't hesitate to ask on an issue before spending time on it.

PR #32864, opened 2024-05-22
Modernize Antora Build

rwinch:
- Use same playbook as docs-build
- Use Env Variables to cause partial build (same as docs-build)
- Use package.json so that dependencies can be updated with dependabot


rwinch:
You will want to make the same changes for all Antora based branches

sdeleuze:
Merged in `6.0.x`, `6.1.x` and `main` branches. Thanks @rwinch.

PR #32874, opened 2024-05-22
Use Threadlocal.remove() instead of .set(null) in documentation

Seungpang:
Hi Team,

This PR updates the Spring Framework documentation to recommend using `ThreadLocal.remove()` instead of `ThreadLocal.set(null)` when unsetting a `ThreadLocal` value.

The current documentation suggests using `ThreadLocal.set(null)` to unset a `ThreadLocal` value. However, this approach can lead to memory leaks and other unexpected behaviors, particularly in multi-threaded and multi-classloader environments. Using `ThreadLocal.set(null)` does not completely remove the `ThreadLocal` instance, which means the associated memory may not be fully released. This can cause data inconsistencies and unexpected behaviors.

In the Spring Framework itself, `ThreadLocal.remove()` is used to ensure that the `ThreadLocal` values are properly cleaned up, preventing potential memory leaks and ensuring consistent behavior across different environments.


snicoll:
Good catch, thanks @Seungpang

PR #32880, opened 2024-05-23
Add update-antora-ui-spring.yml workflow

marcusdacoregio:
This workflow will take care of opening PRs to update the `antora-ui-spring` artifact in the `antora-playbook.yml` file from supported branches and from `docs-build`.

More details about the workflow at https://github.com/spring-io/spring-doc-actions?tab=readme-ov-file#update-antora-ui-spring

The action is using a commit hash as its version for now since we still don't have a tag with the workflow, once the tag is created I'll open a PR to use it.


.github/workflows/update-antora-ui-spring.yml
```diff
@@ -0,0 +1,35 @@
...
+  issues: write
+  contents: write
+
+jobs:
+  update-antora-ui-spring:
+    runs-on: ubuntu-latest
+    name: Update on Supported Branches
+    strategy:
+      matrix:
+        branch: [ '5.3.x', '6.0.x', '6.1.x', 'main' ]
```
sdeleuze, line 19: @marcusdacoregio Thanks for this PR. We don't use Antora on `5.3.x`, can I refine this PR to remove this branch?

marcusdacoregio: Sure, feel free to make any changes

sdeleuze:
@marcusdacoregio With @bclozel, we are discussing having potential duplicated workflow when we will create the `6.2.x` branch. What about merging this PR to the `docs-build` branch instead of `main` to make sure we keep a single one?

marcusdacoregio:
@sdeleuze I made something wrong while changing this PR. I've opened https://github.com/spring-projects/spring-framework/pull/32893 targeting `docs-build` and also removed the `5.3.x` branch from the list.

PR #32896, opened 2024-05-26
Additional Assert methods

jthugg:
These changes might not be important.

Sometimes user need(or want) to check the expressions or objects and throw specific exception.
These changes would be helpful in such situations.


snicoll:
@jthugg thanks for the suggestion but, as stated in the Javadoc, this class for mostly for our own use:

> Mainly for internal use within the framework; for a more comprehensive suite of assertion utilities consider org. apache. commons. lang3.Validate from Apache Commons Lang , Google Guava's Preconditions , or similar third-party libraries.

As we don't have the need for those methods, we won't add them.

PR #32900, opened 2024-05-26
Add test cases for hasHttpHandlerDecorator method

gdmrw:
Improve coverage


snicoll:
Thanks for the PR but `hasHttpHandlerDecorator` is trivial and the tests here consist of setting something on the builder and asserting it has been set. While it does indeed increase coverage, those tests do not bring a lot of value on their own.

PR #32901, opened 2024-05-26
add more test case to improve coverage 

june-rains:
Add more test case to improve coverage  for SqlQuery class in spring-jdbc and EscapedErrors class in spring-web


snicoll:
Thanks for the PR. Is this part of an assignment or something? While we actively seek contributions, this one seems a bit random to me.

PR #32931, opened 2024-06-01
Add coroutine variant of `WebExceptionHandler`

earlgrey02:
As in the case of `CoWebFilter`, I added `CoWebExceptionHandler`, **a coroutine variant** of `WebExceptionHandler` used for global exception handling in Spring WebFlux.

### DIFF
--- a/spring-web/src/main/kotlin/org/springframework/web/server/CoWebExceptionHandler.kt
+++ b/spring-web/src/main/kotlin/org/springframework/web/server/CoWebExceptionHandler.kt
@@ -3,11 +3,13 @@ package org.springframework.web.server
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.reactor.mono
 import reactor.core.publisher.Mono
+import kotlin.coroutines.CoroutineContext
 
 abstract class CoWebExceptionHandler : WebExceptionHandler {
-	final override fun handle(exchange: ServerWebExchange, ex: Throwable): Mono<Void> =
-		mono(Dispatchers.Unconfined) { coHandle(exchange, ex) }.then()
-
+	final override fun handle(exchange: ServerWebExchange, ex: Throwable): Mono<Void> {
+		val context = exchange.attributes[CoWebFilter.COROUTINE_CONTEXT_ATTRIBUTE] as CoroutineContext?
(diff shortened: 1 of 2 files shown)


sdeleuze:
We need to be careful about not going too far in terms of Coroutines duplications of Reactive contrats, but I think I am in favor of that one for 2 reaons:
 - That seems indeed pretty close to `CoWebFilter` use cases which is pretty popular among Kotlin developers I think
 - This is not yet reflected in the PR, but `CoWebExceptionHandler` could introduced the same kind of Coroutines context management [as in `CoWebFilter`](https://github.com/spring-projects/spring-framework/blob/416eff1b0488076a5f7ff2f552e46714736df526/spring-web/src/main/kotlin/org/springframework/web/server/CoWebFilter.kt#L38-L45).

@rstoyanchev Would you be ok for that?

@earlgrey02 What about refining your PR to provide Coroutines context management like `CoWebFilter` does?

rstoyanchev:
If `CoWebFilter` exists then yes a `CoWebExceptionHandler` next to it makes sense.

earlgrey02:
> We need to be careful about not going too far in terms of Coroutines duplications of Reactive contrats, but I think I am in favor of that one for 2 reaons:
> 
> * That seems indeed pretty close to `CoWebFilter` use cases which is pretty popular among Kotlin developers I think
> * This is not yet reflected in the PR, but `CoWebExceptionHandler` could introduced the same kind of Coroutines context management [as in `CoWebFilter`](https://github.com/spring-projects/spring-framework/blob/416eff1b0488076a5f7ff2f552e46714736df526/spring-web/src/main/kotlin/org/springframework/web/server/CoWebFilter.kt#L38-L45).
> 
> @rstoyanchev Would you be ok for that?
> 
> @earlgrey02 What about refining your PR to provide Coroutines context management like `CoWebFilter` does?

Instead of creating a separate `COROUTINE_CONTEXT_ATTRIBUTE` for `CoWebExceptionHandler`, I reused the `COROUTINE_CONTEXT_ATTRIBUTE` used in the existing `CoWebFilter` and handler(`CoroutineContextAware HandlerFunction`).

sdeleuze:
Merged and polished, thanks for your contribution.

PR #32941, opened 2024-06-03
Add a way to set the timeout to reactive jdk client

m4tt30c91:
Add a setTimeout method to the http reactive jdk native connector

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpConnector.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpConnector.java
@@ -21,6 +21,7 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.ByteBuffer;
+import java.time.Duration;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


jhoeller:
Thanks for the suggestion! We should name that method `setReadTimeout`, aligned with ´JdkClientHttpRequestFactory`, and possibly pass the `Duration` straight into an extended `JdkClientHttpRequest` constructor (since the request class is not public anyway, we can flexibly change the constructor signature to accept a `Duration` parameter that we can directly apply there).

Feel free to revise the PR accordingly, otherwise we can also implement it directly.

matteocristoforo-fabrick:
Sure, gonna review the pull request accordingly; is it ok if I add a commit on top, or would you prefer a single one to keep it cleaner?

m4tt30c91:
Sure, gonna review the pull request accordingly; is it ok if I add a commit on top, or would you prefer a single one to keep it cleaner?

snicoll:
Both are fine but if you can squash that’s better.

m4tt30c91:
I did review the pull request, let me know if that is ok.

snicoll:
Thanks for making your first contribution to Spring Framework, @m4tt30c91.

PR #32954, opened 2024-06-04
Extract Method PR

MaxNstk:
Enhancing the readability of the code using Extract Method


snicoll:
Thanks but such refactoring is subjective and we prefer to do these ourselves. As it stands the existing code looks fine.

PR #32955, opened 2024-06-04
Adding hash constant

MaxNstk:
Adding constant for hash, removing the explicitly call of the 31 number.


snicoll:
Thanks for the PR, but as indicated in the other PR you've raised, these changes are subjective and the code is fine as it is.

PR #32956, opened 2024-06-04
Inverting conditions and improving readability of the code

MaxNstk:
Removing ifs and improving code quality


snicoll:
Same as the other two. Thanks for the PR, in any case!

PR #32957, opened 2024-06-05
Entity name in MappingSqlQuery example of reference guide is wrong

soglad:
In the document of "Modeling JDBC Operations as Java Objects", the getCustomer(Long id) method should be getActor(Long id).  This commit corrected this issue.


snicoll:
@soglad good catch, thanks very much for making your first contribution to Spring Framework.

PR #32966, opened 2024-06-05
Use HttpStatusCode consistently in reference guide

ypyf:
Updated the first argument to `onStatus(...)` to use `HttpStatusCode` instead of `HttpStatus`.


snicoll:
Both are valid but I can see how we want the doc to be consistent there.

ypyf:
> Thanks for the PR. Both are valid but I can see how we want the doc to be consistent there.

if I understand correctly, `Predicate<HttpStatus>` could not safely assign  to `Predicate<HttpStatusCode>` although `HttpStatus` is a subtype of `HttpStatusCode`.

snicoll:
You're right, I missed the link with the predicate.

PR #32991, opened 2024-06-09
Add Conditional Property Loading Based on Environment

kevchen266:
Purpose:
This feature allows Spring Boot applications to conditionally load properties based on the active environment, without requiring complex dynamic configurations. It simplifies configuration management by enabling properties to be loaded only if certain conditions are met, thus reducing the need for manual configuration changes and restarts. This enhancement addresses the current issue where the static configuration approach requires restarting the application to apply any changes, leading to potential downtime and manual interventions.

Implementation Details:

ConditionalOnProperty Annotation: A custom annotation that specifies conditions under which a bean should be created based on property values.
OnPropertyCondition Class: Implements the logic to check if a specified property exists in the environment and matches the expected value.
MyConfiguration Class: Uses the @ConditionalOnProperty annotation to conditionally create a bean.
MyFeatureBean Class: Represents a simple bean with properties and methods to demonstrate functionality.
ConditionalOnPropertyTest Class: Verified that the bean is created when the property is set.
Verified that the bean is not created when the property is not set.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConditionalOnProperty.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConditionalOnProperty.java
@@ -1,0 +1,14 @@
+package org.springframework.context.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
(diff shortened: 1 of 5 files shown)


snicoll:
@kevchen266 thanks for the PR. I don't know if you've searched the issue tracker before working on this, but this was requested before and declined (see #18372 and the linked issues). Sorry for the wasted time but the efforts of moving the code here isn't something that would change our mind.

Issue #30509, opened 2023-05-20
Support to propagate a savepoint operations via TransactionSynchronization


kazuki43zoo:
I propose to support to propagate a savepoint operation via `TransactionSynchronization`.

## Background

I'm the MyBatis's developer. Now, If application developer use the savepoint(=`PROPAGATION_NESTED`) feature together with MyBatis, If perform select statement after rolling back a savepoint, an application may get the rolled back data from MyBatis's local cache.
I want to the opportunity for clearing local cache that hold on MyBatis module at savepoint operations(at roll backed mainly) for resolving potential problem.

## Example Testing Code

```java
package com.example;

import org.apache.ibatis.annotations.Insert;
import org.apache.ibatis.annotations.Mapper;
import org.apache.ibatis.annotations.Select;
import org.apache.ibatis.annotations.Update;
import org.assertj.core.api.Assertions;
import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.transaction.PlatformTransactionManager;
import org.springframework.transaction.TransactionDefinition;
import org.springframework.transaction.support.DefaultTransactionDefinition;
import org.springframework.transaction.support.TransactionOperations;
import org.springframework.transaction.support.TransactionTemplate;

@SpringBootTest
class MybatisSpringGh785ApplicationTests {

  @Autowired
  MessageMapper mapper;

  @Autowired
  PlatformTransactionManager transactionManager;

  @Test
  void contextLoads() {

    // TransactionOperations for standard JDBC Transaction
    DefaultTransactionDefinition txDef = new DefaultTransactionDefinition(TransactionDefinition.PROPAGATION_REQUIRED);
    TransactionOperations txOperations = new TransactionTemplate(transactionManager, txDef);

    // TransactionOperations for JDBC savepoint Transaction
    DefaultTransactionDefinition savepointTxDef = new DefaultTransactionDefinition(
        TransactionDefinition.PROPAGATION_NESTED);
    TransactionOperations savepointTxOperations = new TransactionTemplate(transactionManager, savepointTxDef);

    // Insert test data
    txOperations.executeWithoutResult(s -> {
      mapper.insert(new Message(1, "Hello!"));
    });

    // Do Testing
    txOperations.executeWithoutResult(txStatus -> {

      // Select latest message and load to local cache
      Message messageBefore = mapper.select(1);

[...]

PR #32992, opened 2024-06-09
Do not swallow exceptions on failing to release a savepoint

ccorbacho:
By swallowing the exception on failing to release a savepoint, we are hiding client code from database errors that they may need to be aware of or handle.

e.g. if the connection is now dead when trying to release the savepoint, by swallowing the exception here, client code has no way of knowing this now until the next operation on the database, which will yield a confusing message about the problem and obscures where the failure occurred as it will be in a different part of their code that throws any exception.

Swallowing the exception here is also inconsistent with all the other database calls in the Transaction Object, where we always raise an exception up and let client code decide how to handle the error.


jhoeller:
`releaseSavepoint` has a bit of a double role in serving as a sort-of `commit` as well as a resource cleanup step. On review, I agree that propagation the exceptions make sense, focusing on the commit-like role rather than the cleanup role. This may have side effects in certain scenarios so it seems like a fine change worth trying in 6.2, along with #30509. Thanks for raising this!

PR #32999, opened 2024-06-11
Update Kotlin examples in `autowired.adoc`

lsc4719:
Hello, I recently started to read Spring documentations and I really enjoy it.
Thank you for your great contributions.
I think I found something to fix in the [doc](https://docs.spring.io/spring-framework/reference/core/beans/annotation-config/autowired.html) that explains annotation `@Autowired`.

In the doc, java code shows setter-based injection

```java
	public class MovieRecommender {
		private Set<MovieCatalog> movieCatalogs;
		@Autowired
		public void setMovieCatalogs(Set<MovieCatalog> movieCatalogs) {
			this.movieCatalogs = movieCatalogs;
		}
		// ...
	}
```

and Kotlin code is property-based injection example.
```kotlin
class MovieRecommender {
		@Autowired
		lateinit var movieCatalogs: Set<MovieCatalog>
		// ...
	}
```
but both must be equivalent as much as possible cause they are in the same box.

### DIFF
--- a/framework-docs/modules/ROOT/pages/core/beans/annotation-config/autowired.adoc
+++ b/framework-docs/modules/ROOT/pages/core/beans/annotation-config/autowired.adoc
@@ -255,7 +255,7 @@ Kotlin::
 ----
 	class MovieRecommender {
 
-		@Autowired
+		@set:Autowired
 		lateinit var movieCatalogs: Set<MovieCatalog>
 
 		// ...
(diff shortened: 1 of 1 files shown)


lsc4719:
I added some more `set:`s due to the same reason.

lsc4719:
I closed this PR. Cause for Kotlin, I guess, setter-based/property-bases injections are not distinguishable.

PR #33017, opened 2024-06-12
Fix package description for org.springframework.validation.annotation

izeye:
It seems to have been missed in https://github.com/spring-projects/spring-framework/pull/32532.


jhoeller:
Thanks, good catch!

PR #33023, opened 2024-06-13
spring-form JSP tags should escape HTML value based on response character encoding

stonio:
I notice that spring-form JSP tags always escapes attribute `value` with character encoding `ISO-8859-1` defined in class `WebUtils`. It is preferable to align with response character encoding (likely `UTF-8`).

Spring Framework taglib should only use "HTML escape function" based on response character encoding.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/tags/form/AbstractFormTag.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/tags/form/AbstractFormTag.java
@@ -92,7 +92,7 @@ public abstract class AbstractFormTag extends HtmlEscapingAwareTag {
 	 * as required. This version is <strong>not</strong> {@link PropertyEditor}-aware.
 	 */
 	protected String getDisplayString(@Nullable Object value) {
-		return ValueFormatter.getDisplayString(value, isHtmlEscape());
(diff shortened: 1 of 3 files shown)


poutsma:
Thank you for submitting a PR. I decided to take a simpler approach to solve the issue in [e622555](https://github.com/spring-projects/spring-framework/commit/e62255512a6657e468ae5fc019b4ba5f3d7cc223), so that we  did not have to change `ValueFormatter`.

stonio:
Please re-open #33023.
Few others Spring form tags (such as  `<form:select>` and `<form:options>`) are still ignoring response character encoding.

Examples:
```jsp
<%-- Select items as List --%>
<form:select path="p" items='${["café", "café crème"]}' />

<%-- Select items as Map --%>
<form:select path="p" items='<%= java.util.Map.of("café", "café au lait") %>'  />

<%-- Options items as Map --%>
<form:select path="p">
    <form:options items='<%= java.util.Map.of("café", "café au lait") %>' />
</form:select>
```

poutsma:
Hi @stonio,

I am no longer employed by Broadcom and no longer have commit rights on the Spring Framework, so I am not involved in its maintenance. I would suggest to open a new issue.

PR #33042, opened 2024-06-16
Update Contract.java   -> Changed @since 6.2 -> @since 6.2.0

immanuellukwago:
Changed @since 6.2 -> @since 6.2.0


snicoll:
@immanuellukwago I am not sure what led you to change this. If you look in the codebase, you'll see that we use `X.Y` format for majors. Also, we have a checkstyle rule to validate this and your change makes the build fail.

PR #33073, opened 2024-06-20
Use error handler for reactive cache aspect

zizare:
Use cacheErrorHandler on reactive caching aspect.

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
@@ -506,7 +506,10 @@ public abstract class CacheAspectSupport extends AbstractCacheInvoker
 
 		for (Cache cache : context.getCaches()) {
 			if (CompletableFuture.class.isAssignableFrom(context.getMethod().getReturnType())) {
-				CompletableFuture<?> result = cache.retrieve(key);
+				CompletableFuture<?> result = Optional.ofNullable(cache.retrieve(key)).map(c -> c.exceptionally(ex -> {
+					getErrorHandler().handleCacheGetError((RuntimeException) ex, cache, key);
+					return null;
+				})).orElse(null);
 				if (result != null) {
 					return result.thenCompose(value -> (CompletableFuture<?>) evaluate(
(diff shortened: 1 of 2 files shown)


spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
```diff
@@ -506,9 +506,12 @@ private Object findInCaches(CacheOperationContext context, Object key,
 
 		for (Cache cache : context.getCaches()) {
 			if (CompletableFuture.class.isAssignableFrom(context.getMethod().getReturnType())) {
-				CompletableFuture<?> result = cache.retrieve(key);
+					CompletableFuture<?> result = cache.retrieve(key);
```
simonbasle, line 509: nit: formatting left-over from the 1st commit

spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
```diff
@@ -1136,12 +1139,30 @@ public Object findInCaches(CacheOperationContext context, Cache cache, Object ke
 				if (adapter.isMultiValue()) {
 					return adapter.fromPublisher(Flux.from(Mono.fromFuture(cachedFuture))
 							.switchIfEmpty(Flux.defer(() -> (Flux) evaluate(null, invoker, method, contexts)))
-							.flatMap(v -> evaluate(valueToFlux(v, contexts), invoker, method, contexts)));
+							.flatMap(v -> evaluate(valueToFlux(v, contexts), invoker, method, contexts))
+							.onErrorResume(RuntimeException.class, ex -> {
+								try {
+									getErrorHandler().handleCacheGetError((RuntimeException) ex, cache, key);
+									return Flux.defer(() -> (Flux) evaluate(null, invoker, method, contexts));
```
simonbasle, line 1146: I'm not entirely convinced that the `defer` is necessary here (similarly to how there's no defer in the `flatMap` above). was there a particular reason for it?

zizare: You right it's not necessary here.
Removed it

simonbasle:
Thanks for the contribution @zizare 👍 this will be released in 6.1.11 and 6.2.0-M5, both coming up on July 11

Issue #26264, opened 2020-12-14
AntPathPatternMatcher hardcoded "/" separator in variables glob


jonenst:
If you don't use the default separator, "/" should not be special:

```
jshell> new AntPathMatcher("\0").extractUriTemplateVariables("tot{;}", "totoa");
$11 ==> {;=oa}
```

```
jshell> new AntPathMatcher("\0").extractUriTemplateVariables("tot{/}", "totoa");
|  Exception java.lang.IllegalStateException: Pattern "tot{/}" is not a match for "totoa"
|        at AntPathMatcher.extractUriTemplateVariables (AntPathMatcher.java:516)
|        at (#12:1)
```
Expected {/=oa}

note: the previous example is treated as an exact match:
```
jshell> new AntPathMatcher("\0").extractUriTemplateVariables("tot{/}", "tot{/}");
$19 ==> {}
```

See https://github.com/spring-projects/spring-framework/blob/17e6cf1cc1b57bd65bfae8da8cddc8314a89408d/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java#L645 there are slashes hardcoded in there

tafjwr:
I worked on this issue in #33085

bclozel:
Closing in favor of #33085

PR #33085, opened 2024-06-22
AntPathPatternMatcher hardcoded "/" separator in variables matching and comparator

tafjwr:
Related: #26264 

When AntPathMatcher is constructed with a path separator other than default character "/", the functionality of extraction of URI template variables doesn’t work properly. If the matcher is specified with a non-default character, and a part of URI variable template includes default separator as a variable name, the variable name key mapping leads to exceptions or unexpected results. This is because the regex for URI template variables is hard-coded with "/", which is not considered as a part of variable key, although it can be placed if the matcher has a non-default path separator. To fix this, I updated it to pass a path separator to that regex.

Currently, for example, the matcher with non-default path separator seems to be leveraged on WebSocketConfiguration, in default, to map a "."-separated resource, which is a part of URI segment. Thus, for instance, if the path separator is specified as ".", extracting “price.stock.aaa“ with the template "price.stock.{ticker/symbol}" leads to IllegalStateException.

NOTE: A hard-coded variable pattern is still remaining. It seemed to be indirectly used in PathMatcher#compare implementation, but I’m not sure if this erroneous code causes some bug behaviors or not and is tracked by the team. And for isolation reasons, I would like to exclude it from the scope of code fix in this PR.

https://github.com/spring-projects/spring-framework/blob/b134f253b5a72200821501dbbf921d7d75ff964a/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java#L81

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java
+++ b/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java
@@ -462,7 +462,7 @@ public class AntPathMatcher implements PathMatcher {
 			matcher = this.stringMatcherCache.get(pattern);
 		}
 		if (matcher == null) {
-			matcher = new AntPathStringMatcher(pattern, this.caseSensitive);
+			matcher = new AntPathStringMatcher(pattern, this.pathSeparator, this.caseSensitive);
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for your contribution @tafjwr , this is now in main and should be available in SNAPSHOTs soon.

PR #33111, opened 2024-06-27
 Fix ClassCastException from return value validation with proxy when adaptConstraintViolations=true

sbluemel:
Fixed method arguments of `applyReturnValueValidation` to be called with the return value, no the method arguments.
Added some tests with "adaptConstraintViolations=true" which also failed before the fix.

But I'm not into reactive programming, so I couln't figure out a way to do return value validation on Monos.


sbluemel:
@rstoyanchev I had some time to look into it as well. But I'm unsure about the reactive tests.
Please have a look.

sbluemel:
There are now similiar tests in the original pull request by @imvtsl.

PR #33137, opened 2024-07-02
Add header-only constructor to create MessageHeaderAccessor

Nephery:
Add a constructor that lets users create a `MessageHeaderAccessor` object given only a headers map.

This is a small optimization in high volume scenarios to avoid creating unnecessary `Message<?>` objects when you only have a headers map.

The specific use case I'm looking at, is being able to create a `MessageHeaderAccessor` object to process headers from a "consolidated headers" object (such as from the [Rabbit Binder](https://docs.spring.io/spring-cloud-stream/reference/rabbit/rabbit_overview/receiving-batch.html#consumer-side-batching) or [Solace Binder](https://github.com/SolaceProducts/solace-spring-cloud/tree/master/solace-spring-cloud-starters/solace-spring-cloud-stream-starter#batch-consumers)) created for a `Message<List<?>>` consumed from [Spring Cloud Stream's batched consumers](https://docs.spring.io/spring-cloud-stream/reference/spring-cloud-stream/producing-and-consuming-messages.html#batch-consumers).

e.g.: Being able to do something like this:

```java
@Bean
Consumer<Message<List<Payload>>> input() {
	return batchMsg -> {
		List<Payload> batchedPayloads = batchMsg.getPayload();
		List<Map<String, Object>> batchedHeaders = (List<Map<String, Object>>) batchMsg.getHeaders().get(SolaceBinderHeaders.BATCHED_HEADERS);

		for (int i = 0; i < batchedPayloads.size(); i++) {
			Map<String, Object> headersMap = batchedHeaders.get(i);
			MessageHeaderAccessor accessor = null;
			if (headersMap instanceof MessageHeaders headers) {
				accessor = MessageHeaderAccessor.getAccessor(headers, null);

			}
			if (accessor == null || || !headers.isMutable()) {
				accessor = new MessageHeaderAccessor(headersMap);
			}

			// Do some logic with the accessor
		}
	};
}
```

With the current implementation, instead of just being able to do:
```java
MessageHeaderAccessor accessor = new MessageHeaderAccessor(headersMap);
```

I would need to do this, which is pointless if the accessor's constructor will just extract the headers from the `Message<?>`:
```java
MessageHeaderAccessor accessor = new MessageHeaderAccessor(new GenericMessage<?>(payload, headersMap));
```

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/support/MessageHeaderAccessor.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/support/MessageHeaderAccessor.java
@@ -128,7 +128,7 @@ public class MessageHeaderAccessor {
 	 * A constructor to create new headers.
 	 */
 	public MessageHeaderAccessor() {
-		this(null);
+		this((Map<String, Object>) null);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


snicoll:
@Nephery thanks for the suggestion but until you have signed the CLA, we can't really continue this conversation here. One problem with your change is that there are now two constructors with a single argument that are nullable so this will break backward compatibility.

Nephery:
In that case, I'll close the PR and raise it as an issue.

About your concern, I was wondering if it was a concern as well. I have a potential suggestion to work around it, but lets continue that discussion in the issue once I raise it.

Issue #31746, opened 2023-12-04
NotReadablePropertyException due to mismatch between `ConstraintViolation` property path and `BindingResult` target in `MethodValidationAdapter`


k-seth:
The title is pretty vague, sorry - couldn't think of something better. There is _plenty_ of detail here to make up for that though.

Spring 6.1.x brought method bean validation (https://github.com/spring-projects/spring-framework/issues/30645) - as part of the feedback to this enhancement, I provided a PR https://github.com/spring-projects/spring-framework/pull/31530 to attempt to refine this further.

In hindsight, the solution I provided also has its problems. Essentially, all the issues stem from the use of the leaf bean - whose behaviour is, unfortunately, not as I had understood at the time. As I have come to release, the change I submitted had some unintended side effects that only became clear as I fully started migrating to Spring Boot 3.2.

**1. Cascaded validation can fail**
In cases where a validation error occurs on a nested bean, `SpringValidatorAdopt#getRejectedValue` may attempt to get the raw value for the property. As a result, one will see an error like `JSR-303 validated property 'professor.name' does not have a corresponding accessor for Spring data binding`. This is because the code attempts to walk the full property path, starting from the provided bean. Unfortunately, the leaf bean points to the nested bean with the cascaded validation, not the initial bean where validation started. Thus, the property path is invalid from this context (Go figure, despite giving the exact bean in question, it is no good). The JSR error is not thrown in all cases, as paths which include collections with non-indexed/keyed access are fine because the `getRejectedValue` method explicitly skips walking paths with `[]` in them - instead, it uses the value on the violation directly - because there is no way to access the element in question. 

As an example, consider the following code:

```java
public void registerCourse(@Valid Course course) {
}

private record Course(@NotBlank String title, @Valid Set<Person> students, @Valid Person professor) {
}

private record Person(@Size(min = 1, max = 10) String name, List<@NotBlank String> hobbies) {
}
```
The failure can be caused by having a validation error on the professor parameter - say the name, as my error above uses.

[...]

PR #33150, opened 2024-07-04
Fix adaptation of violations on Set method parameter

snussbaumer:
Since commit 8552e149b5faae31cab6c7d9d3f0f189904e4a8e, validation of Set values (for example `Set<@NotBlank String>`) crashes with error : No way to unwrap Iterable without index

With this fix we don't know which value exactly is the problem, but at least we have a working validation

Maybe there is a way to know exactly which value(s) are the culprit, but without an index I don't know how to find it. 

Or maybe there is a better way to fix the problem, I trust you'll do a thorough review and find it, I'm not that familiar with the codebase 😄 .

Thanks

This is the error without the fix : 
```
java.lang.IllegalStateException: No way to unwrap Iterable without index
	at org.springframework.util.Assert.state(Assert.java:78)
	at org.springframework.validation.beanvalidation.MethodValidationAdapter.adaptViolations(MethodValidationAdapter.java:358)
	at org.springframework.validation.beanvalidation.MethodValidationAdapter.validateArguments(MethodValidationAdapter.java:247)
	at org.springframework.validation.beanvalidation.MethodValidationAdapterTests.testArgs(MethodValidationAdapterTests.java:237)
	at org.springframework.validation.beanvalidation.MethodValidationAdapterTests.validateValueSetArgument(MethodValidationAdapterTests.java:223)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

### DIFF
--- a/spring-context/src/main/java/org/springframework/validation/beanvalidation/MethodValidationAdapter.java
+++ b/spring-context/src/main/java/org/springframework/validation/beanvalidation/MethodValidationAdapter.java
@@ -346,6 +346,10 @@ public class MethodValidationAdapter implements MethodValidator {
 					value = map.get(key);
 					container = map;
 				}
+				else if (arg instanceof Set<?>) {
+					value = arg;
+					container = null;
+				}
 				else if (arg instanceof Optional<?> optional) {
 					value = optional.orElse(null);
 					container = optional;
--- a/spring-context/src/test/java/org/springframework/validation/beanvalidation/MethodValidationAdapterTests.java
+++ b/spring-context/src/test/java/org/springframework/validation/beanvalidation/MethodValidationAdapterTests.java
@@ -19,6 +19,7 @@ package org.springframework.validation.beanvalidation;
 import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Locale;
+import java.util.Set;
 import java.util.function.Consumer;
 
 import jakarta.validation.Valid;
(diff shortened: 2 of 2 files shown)


snussbaumer:
Maybe there is a way to know exactly which value(s) are the culprit, but without an index I don't know how to find it.

snussbaumer:
I the fix could then be backported to 6.1.x that would be great

rstoyanchev (commented):
There is an underlying limitation in that there is no way to access the value in the `Set` that caused the violation (see https://github.com/jakartaee/validation/issues/194).

Initially, we used to create a `ParameterErrors` instance for arrays and collections, with the collection/array as the backing object, and the container + container index allowing you to access the value. It sort of worked for `Set` but there was no index and no way to access the actual value. There was an attempt to improve that in #31530 for Set and other (custom) container types, but it was based on an incomplete understanding of the available metadata, leading to #31746 and the subsequent change d7ce13c7633e902fd928043caf30a3a0f9380508 is what made it impossible to use a collection that is not indexed.

The subsequent change 8552e149b5faae31cab6c7d9d3f0f189904e4a8e that you referenced further refined that so that collections/arrays are more correctly represented with `ParameterValidationResult` as directly validated method parameters (rather than with `ParameterErrors` for cascaded violations on an Object), so there is no longer a backing object to set, just the argument value. That did not change anything for sets though. 

I think we could create `ParameterValidationResult` with the `Set` as the value that was validated, rather than the actual element, which we cannot access, but there would be no container and no container index. That would allow the error to be reported as before, but if there are multiple errors on the same `Set` there would be no way of telling which one is which. This wouldn't be anything new, but would simply restore what was possible in 6.1.2.

PR #33155, opened 2024-07-07
feat: Add CollectionFormat feature

### LINKED ISSUE #33154: Support various ways to encode collections in URL parameters for declarative HTTP interfaces
I've been making good use of the [declarative HTTP interface feature](https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-http-interface) that was added in the latest spring release.
I would like HTTP interface to be a bit more versatile and support different encoding methods for `@RequestParam`.

Currently, the encoding method for `@RequestParam` of type collection seems to be repeating the same key with &, which seems to be related to the way `RequestParamArgumentResolver` is implemented.

```java
interface TestRepository {
	@GetExchange("/test")
	Repository test(@RequestParam List<String> param);
}

@Service
@RequiredArgsConstructor
public class TestService {
        private TestRepository repository;

        public void test() {
                List<String> param = List.of("1", "2", "3");
                repository.test(param); // Generated URI path -> GET /test?param=1&param=2&param=3
        }
}

```

Of course Modern web frameworks expect `@RequestParam` to be deserialized to collection type automatically, either by repeating the same key value as &, or by parsing strings joined by certain delimiters.
However, if this is not possible due to limitations of older web frameworks or API specs, we have to join the strings with a specific delimiter before the request and then pass the strings as `@RequestParam`, which is not cool.

It would be nice to have the ability to encode `@RequestParam` via [@CollectionFormat](https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/CollectionFormat.java), which is supported by Spring Cloud OpenFeign. I think users who use declarative HTTP interface will have experience with Spring Cloud OpenFeign and will expect a smooth transition.

```java
interface TestRepository {
	@GetExchange("/test1")
	Repository test1(@RequestParam @CollectionFormat(CSV) List<String> param);

	@GetExchange("/test2")
	Repository test2(@RequestParam @CollectionFormat(SSV) List<String> param);
}

@Service
@RequiredArgsConstructor
public class TestService {
        private TestRepository repository;

        public void test() {
                List<String> param = List.of("1", "2", "3");
                repository.test1(param); // Generated URI path -> GET /test1?param=1,2,3
                repository.test2(param); // Generated URI path -> GET /test2?param=1 2 3
        }
}

```

I'm curious to know what the maintainers think about this 🙂

doljae:
## Motivation:

Resolve #33154

## Modifications:

- Add new annotation `@CollectionFormat` with simple test.

## Additional info:

- If we have a way to format it according to the checkstyle rules, please let me know 🙏

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/bind/annotation/CollectionFormat.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/annotation/CollectionFormat.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.bind.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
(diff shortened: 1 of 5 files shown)


rstoyanchev:
Closing for now, as per discussion under #33154, we are going to take a different direction.

PR #33156, opened 2024-07-07
DefaultErrorResponseBuilder does not implement headers(Consumer)

TAKETODAY:
https://github.com/spring-projects/spring-framework/blob/f1658079a5503eb4df191a27a2782b2e7a02bb4f/spring-web/src/main/java/org/springframework/web/DefaultErrorResponseBuilder.java#L65-L80


snicoll:
@TAKETODAY thanks very much. I've polished your contribution and added a test.

PR #33165, opened 2024-07-08
Update tomcat version to 10.1.25 in framework-platform.gradle

fredbalves:
Fixes CVE-2024-34750


bclozel:
Thanks, but we like to take care of this ourselves. We're merely compiling against Tomcat's API, so we're not vulnerable to this CVE.

PR #33166, opened 2024-07-08
Harmonize phrasing in URI Encoding section

hunhee98:
This is a documentation-only change and does not impact any functionality of the framework.


snicoll:
@hunhee98 thanks. I am not a native speaker but I think "historic reasons" is valid. However, "historical reasons" being used in the vast majority of the code base, it's still worth harmonizing.

Issue #29514, opened 2022-11-18
Update documentation to mention Java 17+ baseline


Andrew-Russell-fingo:
I believe that the latest version of spring-context-6.0.0.jar
And uploaded to Maven Central
Has been compiled with the wrong jvm version, so cannot work inside a jdk8 vm

Last Version: 5.3.24, compiled for JDK8
From https://repo1.maven.org/maven2/org/springframework/spring-context/5.3.24/spring-context-5.3.24.module

```
{
  "formatVersion": "1.1",
  "component": {
    "group": "org.springframework",
    "module": "spring-context",
    "version": "5.3.24",
    "attributes": {
      "org.gradle.status": "release"
    }
  },
  "createdBy": {
    "gradle": {
      "version": "7.5.1"
    }
  },
  "variants": [
    {
      "name": "apiElements",
      "attributes": {
        "org.gradle.category": "library",
        "org.gradle.dependency.bundling": "external",
        "org.gradle.jvm.version": 8,
        "org.gradle.libraryelements": "jar",
        "org.gradle.usage": "java-api",
        "org.jetbrains.kotlin.localToProject": "public",
        "org.jetbrains.kotlin.platform.type": "jvm"
      },
...
```
Now from version 6.0.0 , compiled for JDK17
https://repo1.maven.org/maven2/org/springframework/spring-context/6.0.0/spring-context-6.0.0.module

```
{
  "formatVersion": "1.1",
  "component": {
    "group": "org.springframework",
    "module": "spring-context",
    "version": "6.0.0",
    "attributes": {
      "org.gradle.status": "release"
    }
  },
  "createdBy": {
    "gradle": {
      "version": "7.5.1"
    }
  },
  "variants": [
    {
      "name": "apiElements",
      "attributes": {
        "org.gradle.category": "library",
        "org.gradle.dependency.bundling": "external",
        "org.gradle.jvm.environment": "standard-jvm",
        "org.gradle.jvm.version": 17,
        "org.gradle.libraryelements": "jar",
        "org.gradle.usage": "java-api",
        "org.jetbrains.kotlin.platform.type": "jvm"
      },
...
```

ttddyy:
@Andrew-Russell-fingo Spring Framework 6 requires Java 17+ as a baseline. So, it doesn't work with Java8.

[...]

Issue #29476, opened 2022-11-13
Double set of annotations needed to support REST service and client with same interface


ilgrosso:
Let's suppose we want to implement the typical use case for JAX-RS developers: for a given REST service, define:

1. interface
2. server-side implementation
3. proxy client

With introduction of [Declarative HTTP Interface](https://docs.spring.io/spring-framework/docs/6.0.x/reference/html/integration.html#rest-http-interface), this can now be accomplished with:

### interface

```java
@HttpExchange("/rest/users")
@RequestMapping("/rest/users")
public interface UserService {
    @GetExchange(value = "/")
    @GetMapping(value = "/")
    List<User> search();
}
```

### controller

```java
@RestController
public class UserServiceImpl implements UserService {
  public List<User> search() {
    return List.of();
  }
}
```

### client

```java
WebClient client = WebClient.builder().baseUrl(baseUrl).build();
HttpServiceProxyFactory factory = HttpServiceProxyFactory.builder(WebClientAdapter.forClient(client)).build();

UserService userService = factory.createClient(UserService.class);
userService.seatch();
```

## The trouble

So far, so good: all works.

The point is: is it really necessary to double each annotation on the interface, one time for the controller, one time for the client? Both annotations are providing the same information.

quaff:
It confuse me too.

bclozel:
The sample you've shown indeed begs the question - why not share the server annotations for the interface client generation? `@RequestMapping` method signatures can be very flexible and in many cases, they can depend on server types or worse, are not precise enough to generate an actual HTTP request.

Spring Cloud OpenFeign has tested this approach and is moving away from it on purpose. See spring-cloud/spring-cloud-openfeign#678. This also has been discussed in more details in this talk: https://youtu.be/5LNOnVJKW_4?t=1156

ilgrosso:
Thank you for explanation and further details @bclozel

quaff:
> 

Could spring provider built-in annotations that combine server and client annotations. like this
```java
@GetMapping
@GetExchange
public @interface Get {

	@AliasFor(annotation = {GetMapping.class, GetExchange.class})
	String[] value() default {};

}
```

bclozel:
@quaff unfortunately, this wouldn't solve the problems listed in the issues/talk I've pointed to.

quaff:
> @quaff unfortunately, this wouldn't solve the problems listed in the issues/talk I've pointed to.

[...]

PR #33213, opened 2024-07-14
Upgrade to Gradle 8.9

diguage:
[Gradle 8.8 crashed on MacOS 11.x (8.7 still works as expected) · Issue #29476 · gradle/gradle](https://github.com/gradle/gradle/issues/29476), 

and Gradle 8.9 fix the bug, [Disable file-system watching on macOS 11 and before by lptr · Pull Request #29514 · gradle/gradle](https://github.com/gradle/gradle/pull/29514).

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.8-bin.zip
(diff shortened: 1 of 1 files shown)


snicoll:
@diguage thanks but we prefer to do those upgrades ourselves.

diguage:
> @diguage thanks but we prefer to do those upgrades ourselves.

OK, as long as the problem is fixed, no matter who commits the code.

snicoll:
> OK, as long as the problem is fixed, no matter who commits the code.

Hopefully, looking at the commit of https://github.com/spring-projects/spring-framework/issues/33214 and your proposal, you'll understand this isn't about who commits the code but to make sure to do things correctly.

PR #33220, opened 2024-07-15
Allow formatting of Collection values for `@RequestParam` with HTTP interface client

### LINKED ISSUE #33154: Support various ways to encode collections in URL parameters for declarative HTTP interfaces

doljae:
## Motivation:

Make it possible to format `Collection`, `@RequestParam` arguments to a single String value. On the server side, the `ConversionService` parses a String value to the target method parameter type. On the client side, the reverse could be done to format the `MethodParameter` to a String. 

See original request and discussion under #33154.

## Modifications:

Add customization for handling collection types to `RequestParamArgumentResolver`

## Additional info:

- I plan to polish the code after checking with the maintainer to make sure I'm working in the direction I intended.
- If we have a way to format it according to the checkstyle rules, please let me know 🙏

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/service/invoker/RequestParamArgumentResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/service/invoker/RequestParamArgumentResolver.java
@@ -18,8 +18,12 @@ package org.springframework.web.service.invoker;
 
 import org.springframework.core.MethodParameter;
 import org.springframework.core.convert.ConversionService;
+import org.springframework.http.MediaType;
 import org.springframework.lang.Nullable;
 import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.service.annotation.*;
+
+import java.util.List;
 
 /**
  * {@link HttpServiceArgumentResolver} for {@link RequestParam @RequestParam}
@@ -54,9 +58,16 @@ import org.springframework.web.bind.annotation.RequestParam;
  */
 public class RequestParamArgumentResolver extends AbstractNamedValueArgumentResolver {
 
+	private boolean formatAsSingleValue;
 
 	public RequestParamArgumentResolver(ConversionService conversionService) {
 		super(conversionService);
+		this.formatAsSingleValue = true;
+	}
+
+	public RequestParamArgumentResolver(ConversionService conversionService, boolean formatAsSingleValue) {
+		super(conversionService);
+		this.formatAsSingleValue = formatAsSingleValue;
 	}
 
 
@@ -64,8 +75,45 @@ public class RequestParamArgumentResolver extends AbstractNamedValueArgumentReso
 	@Nullable
 	protected NamedValueInfo createNamedValueInfo(MethodParameter parameter) {
 		RequestParam annot = parameter.getParameterAnnotation(RequestParam.class);
+		boolean formatAsSingleValue = this.formatAsSingleValue;
+		if (annot == null) {
+			return null;
+		}
+
+		List<String> invalidContentType = List.of(
+				MediaType.APPLICATION_FORM_URLENCODED_VALUE.toLowerCase(),
+				MediaType.MULTIPART_FORM_DATA_VALUE.toLowerCase(),
+				MediaType.MULTIPART_FORM_DATA_VALUE.toLowerCase(),
+				MediaType.MULTIPART_MIXED_VALUE.toLowerCase(),
+				MediaType.MULTIPART_RELATED_VALUE.toLowerCase());
+
+		PostExchange annot1 = parameter.getMethodAnnotation(PostExchange.class);
+		if (annot1 != null && invalidContentType.contains(annot1.contentType().toLowerCase())) {
+			formatAsSingleValue = true;
+		}
+
+		PutExchange annot2 = parameter.getMethodAnnotation(PutExchange.class);
+		if (annot2 != null && invalidContentType.contains(annot2.contentType().toLowerCase())) {
+			formatAsSingleValue = true;
+		}
+
+		DeleteExchange annot3 = parameter.getMethodAnnotation(DeleteExchange.class);
+		if (annot3 != null && invalidContentType.contains(annot3.contentType().toLowerCase())) {
+			formatAsSingleValue = true;
+		}
+
+		PatchExchange annot4 = parameter.getMethodAnnotation(PatchExchange.class);
+		if (annot4 != null && invalidContentType.contains(annot4.contentType().toLowerCase())) {
+			formatAsSingleValue = true;
+		}
+
+		HttpExchange annot5 = parameter.getDeclaringClass().getAnnotation(HttpExchange.class);
+		if (annot5 != null && invalidContentType.contains(annot5.contentType().toLowerCase())) {
+			formatAsSingleValue = true;
+		}
+
 		return (annot == null ? null :
-				new NamedValueInfo(annot.name(), annot.required(), annot.defaultValue(), "request parameter", true));
+				new NamedValueInfo(annot.name(), annot.required(), annot.defaultValue(), "request parameter", formatAsSingleValue));
 	}
 
 	@Override
--- a/spring-web/src/main/java/org/springframework/web/service/invoker/RequestParamArgumentResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/service/invoker/RequestParamArgumentResolver.java
@@ -55,11 +55,11 @@ import org.springframework.web.bind.annotation.RequestParam;
  */
 public class RequestParamArgumentResolver extends AbstractNamedValueArgumentResolver {
 
-	private boolean formatAsSingleValue;
+	private boolean formatAsSingleValue = true;
+
 
 	public RequestParamArgumentResolver(ConversionService conversionService) {
 		super(conversionService);
-		this.formatAsSingleValue = true;
 	}
 
 	public RequestParamArgumentResolver(ConversionService conversionService, boolean formatAsSingleValue) {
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/web/service/invoker/RequestParamArgumentResolver.java
```diff
@@ -54,18 +58,62 @@
...
+		if (annot == null) {
+			return null;
+		}
+
+		List<String> invalidContentType = List.of(
+				MediaType.APPLICATION_FORM_URLENCODED_VALUE.toLowerCase(),
+				MediaType.MULTIPART_FORM_DATA_VALUE.toLowerCase(),
+				MediaType.MULTIPART_FORM_DATA_VALUE.toLowerCase(),
+				MediaType.MULTIPART_MIXED_VALUE.toLowerCase(),
+				MediaType.MULTIPART_RELATED_VALUE.toLowerCase());
```
rstoyanchev, line 88: I would replace this with a protected method that takes the same input and decides whether to format as single value or not. The method can return `true` if `formatAsSingleValue` is set except when the content type is form data or multipart data. For the multipart check, you could use `contentType.getType().equals("multipart")`. 

The protected method would allow a subclass to further refine the decision. It would be a good idea to expose the formatAsSingleValue property with a getter for subclasses to be able to check. I think the property could also have a setter rather than being a constructor argument.

spring-web/src/test/java/org/springframework/web/service/invoker/RequestParamArgumentResolverTests.java
```diff
@@ -57,12 +59,155 @@ void requestParam() {
...
+		void getForm(@RequestParam String param1, @RequestParam List<String> param2);
+
+		@PutExchange(contentType = "application/x-www-form-urlencoded")
+		void putForm(@RequestParam String param1, @RequestParam List<String> param2);
+
+		@PutExchange(contentType = "application/x-www-form-urlencoded")
+		void deleteForm(@RequestParam String param1, @RequestParam List<String> param2);
+
+		@PatchExchange(contentType = "application/x-www-form-urlencoded")
+		void patchForm(@RequestParam String param1, @RequestParam List<String> param2);
```
rstoyanchev, line 197: I don't think we need the same test for every HTTP method. It would be sufficient to have one test that makes sure formatting can be suppressed.

spring-web/src/test/java/org/springframework/web/service/invoker/RequestParamArgumentResolverTests.java
```diff
@@ -57,12 +59,155 @@ void requestParam() {
...
 	}
 
+	@HttpExchange(contentType = "application/x-www-form-urlencoded")
+	private interface Service2 {
+
+		@GetExchange
+		void getForm(@RequestParam String param1, @RequestParam List<String> param2);
+	}
```
rstoyanchev, line 205: I don't think HTTP GET with form data is not a valid scenario. HTTP GET is not expected to have a body.

spring-web/src/test/java/org/springframework/web/service/invoker/RequestParamArgumentResolverTests.java
```diff
@@ -57,12 +59,155 @@ void requestParam() {
...
+
+	@HttpExchange
+	private interface Service3 {
+
+		@GetExchange
+		void getForm(@RequestParam String param1, @RequestParam List<String> param2);
+	}
```
rstoyanchev, line 212: Again I don't think adding more test cases really adds much value.

rstoyanchev (changes requested):
Looking at the changes so far, I've decided to make the following update dcabddddc21825de1ffa452803729926563f0fe0 to make it easier access to `HttpExchange` metadata in argument resolvers. Now `RequestParamArgumentResolver` can override the overloaded `createNamedValueInfo` method and get easy access to the HTTP method and content type.

See further comments below.

spring-web/src/test/java/org/springframework/web/service/invoker/RequestParamArgumentResolverTests.java
doljae: On d4f666b commit, I clean up code & leave only the one valid test you mentioned.

spring-web/src/test/java/org/springframework/web/service/invoker/RequestParamArgumentResolverTests.java
doljae: On d4f666b commit, I clean up code & leave only the one valid test you mentioned.

spring-web/src/test/java/org/springframework/web/service/invoker/RequestParamArgumentResolverTests.java
doljae: Same as https://github.com/spring-projects/spring-framework/pull/33220#discussion_r1685286621

spring-web/src/test/java/org/springframework/web/service/invoker/RequestParamArgumentResolverTests.java
doljae: Same as https://github.com/spring-projects/spring-framework/pull/33220#discussion_r1685286621

spring-web/src/main/java/org/springframework/web/service/invoker/RequestParamArgumentResolver.java
doljae: On https://github.com/spring-projects/spring-framework/commit/d4f666bd318749a05f06ca700daad640b57ea685 commit, I add new protected method `isMultiValueFormContentType(MediaType contentType)` and use it on new override method you made 🙂

doljae:
On [95a76ee](https://github.com/spring-projects/spring-framework/pull/33220/commits/95a76ee10b218d5a2bb2ebe923719f505857684b) commit I tried to applying coding style which is used in the project. Please let me know if I need to make any additional changes.

One more thing, I was wondering what your thoughts are on supporting a variety of delimiters besides the comma that `@Collectionformat` supports.

rstoyanchev (approved):
Thanks for the updates, I'll take it from here. For a different delimiter, you can customize the `ConversionService`.

yvasyliev:
Hello @rstoyanchev, I need some clarification here.

I've got a case where I need to support different collection encoding methods within the same HTTP client:

```java
@HttpExchange("/api/v1")
public interface MyClient {
    @GetExchange("/list")
    String sendList(List<String> params); // must be encoded as `params=value1&params=value2`

    @GetExchange("/csv")
    String sendCsv(List<String> params); // must be encoded as `params=value1,value2`
}
```

Could you please suggest how can I tackle it after f967f6f?

With Feign client I could do the following:

```java
@FeignClient("myClient")
public interface MyClient {
    @GetMapping("/api/v1/list")
    String sendList(List<String> params); // will be encoded as `params=value1&params=value2`

    @CollectionFormat(feign.CollectionFormat.CSV)
    @GetMapping("/api/v1/csv")
    String sendCsv(List<String> params); // will be encoded as `params=value1,value2`
}
```

I would expect to see a `@CollectionFormat` alternative for HTTP Interface to be applied to each request parameter individually. It will provide users with high customization capabilities:

```java
@HttpExchange("/api/v1")
public interface MyAnotherClient {
    @GetExchange("/list")
    String sendDifferentParams(
        List<String> params,
        @CollectionFormat(CSV) List<String> csvParams
    ); // will be encoded as `params=value1&params=value2&csvParams=value1,value2`
}
```

doljae:
Hello @yvasyliev .
I'm not the one who you mentioned but I received an email from your comment.
You can use it with [milestone version](https://mvnrepository.com/artifact/org.springframework/spring-webmvc/6.2.0-M7) with [this way](https://github.com/spring-projects/spring-framework/issues/33154#issuecomment-2226988544).

As I know, you're right. Currently we can decide only one way of handling collection values with `@RequestParam` annotation in single declarative HTTP interface.

I'm not the maintainer but I think Spring core already have an `ArgumentResolver` and it's derived class, we can decide the way to handle collection values with registering customized `ArgumentResolver` class. `@CollectionFormat`'s implementation is different from the one in spring core and I'm guessing that their intent is to be cautious about adding new annotations. Again, this is just my opinion 🙂

PR #33222, opened 2024-07-17
Add support for making MapAccessor read-only

quaff:
Align to `ReflectivePropertyAccessor`

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/expression/MapAccessor.java
+++ b/spring-context/src/main/java/org/springframework/context/expression/MapAccessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


snicoll:
Before we can consider this PR, we'd need to understand the use case. You are talking about alignment but I don't see that as necessary (or inconsistent at the moment). Also, please refrain from doing unrelated changes. All those formatting in the tests should go away.

quaff:
> Before we can consider this PR, we'd need to understand the use case.

I need `EvaluationContext` like `SimpleEvaluationContext.forReadOnlyDataBinding()` plus `MapAccessor`.

> Also, please refrain from doing unrelated changes. All those formatting in the tests should go away.

OK, I will revert those changes.

sbrannen:
Hi @quaff,

> I need `EvaluationContext` like `SimpleEvaluationContext.forReadOnlyDataBinding()` plus `MapAccessor`.

In order to achieve your goal, you'll actually need to make use of `withAssignmentDisabled()` as well -- for example, something like the following.

```java
SimpleEvaluationContext context = SimpleEvaluationContext
		.forPropertyAccessors(new MapAccessor(false), DataBindingPropertyAccessor.forReadOnlyAccess())
		.withAssignmentDisabled()
		.build();
```

- see #33319

quaff:
Thanks for your caution.

PR #33233, opened 2024-07-18
Support JAXBElement in Jaxb2RootElementHttpMessageConverter

deblockt:
On the current Spring version, usage of `Jaxb2RootElementHttpMessageConverter` in `RestClient` `messageConverters` don't allow to serialize `JAXBElement `.

This PR add this feature.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/xml/Jaxb2RootElementHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/xml/Jaxb2RootElementHttpMessageConverter.java
@@ -121,7 +121,7 @@ public class Jaxb2RootElementHttpMessageConverter extends AbstractJaxb2HttpMessa
 
 	@Override
 	public boolean canWrite(Class<?> clazz, @Nullable MediaType mediaType) {
-		return (AnnotationUtils.findAnnotation(clazz, XmlRootElement.class) != null && canWrite(mediaType));
+		return ((JAXBElement.class.isAssignableFrom(clazz) || AnnotationUtils.findAnnotation(clazz, XmlRootElement.class) != null) && canWrite(mediaType));
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


snicoll:
@deblockt thanks for the PR but it looks a little strange to me. Shouldn't the conversion happen via the content-type rather than via supporting such a type? Where would you be using this exactly?

deblockt:
Hi @snicoll , This PR is the equivalent of this commit on `WebClient` https://github.com/mdeinum/spring-framework/commit/9de58d78da431a4a94d08cab964221b62a3c527b.

In my project, I try to serialize XML object using `RestClient` with this code.
```
restClient = RestClient.builder()
                .messageConverters((httpMessageConverters -> {
                    httpMessageConverters.add(new Jaxb2RootElementHttpMessageConverter());
                }))
                .build();
restClient
                    .post().uri("/")
                    .contentType(MediaType.APPLICATION_XML)
                    .body(new ObjectFactory().createPojo(pojo))
                    .retrieve()
                    .body(MyPojo.class);
```
Where `new ObjectFactory().createPojo(pojo)` return a `JAXBElement<MyPojo>`.

The `ObjectFactory` and `MyPojo` classes are generated using `jaxb2-maven-plugin` and this plugin don't annotate classes with `@XmlRootElement` so we should use the `JAXBElement` to generate a valid XML.

snicoll:
Thanks for the additional details.

> The ObjectFactory and MyPojo classes are generated using jaxb2-maven-plugin and this plugin don't annotate classes with @XmlRootElement so we should use the JAXBElement to generate a valid XML.

Yeah I didn't mean that `@XmlRootElement` was mandatory but rather that the content type should drive the conversion, not the type to serialize but I can see how that's short sighted. And the fact we've done this for codecs means we should do it for MVC as well.

snicoll:
Thanks again @deblockt!

PR #33241, opened 2024-07-19
fix timezone parsing

yibijianghu:


snicoll:
@yibijianghu thanks for the PR but you haven't described what this code change is supposed to fix. And we can't really consider it until you do with unit tests.

yibijianghu:
if timeZoneString is "GMTabc"， this method will return GMT. but if timeZoneString is "abc", it will throw  IllegalArgumentException. this is not right. GMTabc or abc  should  throw exception.

snicoll:
>  this is not right. GMTabc or abc should throw exception.

@yibijianghu the Javadoc of `parseTimeZoneString` states:

> timeZoneString – the time zone String, following TimeZone. getTimeZone(String) [...]

So it's working as documented. `TimeZone.getTimeZone("GMTabc")` returns GMT indeed and that's a behavior of the JDK.

PR #33247, opened 2024-07-20
Add missing Javadoc since in AbstractUrlHandlerMapping

izeye:
This PR adds a Javadoc `@since` tag for the `AbstractUrlHandlerMapping.unregisterHandler()`.

See gh-32064


snicoll:
Thanks again @izeye - there were two `add` methods added in this commit so I've polished the proposal to add missing `@since` tag for them too.

PR #33248, opened 2024-07-20
Stop referring to Oracle STRUCT and ARRAY as they are deprecated

ngocnhan-tran1996:
I see `StructDescriptor` and `ArrayDescriptor` have been deprecated.

Replace for `StructDescriptor`
>Use factory method Connection.createStruct to create an instance of java.sql.Struct directly.

Replace for `ArrayDescriptor`
>Use factory method OracleConnection.createOracleArray to create an instance of java.sql.Array directly.

[StructDescriptor javadoc](https://docs.oracle.com/en/database/oracle/oracle-database/23/jajdb/oracle/sql/StructDescriptor.html)
[ArrayDescriptor javadoc](https://docs.oracle.com/en/database/oracle/oracle-database/23/jajdb/oracle/sql/ArrayDescriptor.html)

That why I make this PR. Is it necessary to update document? If it is necessary, please update kotlin section.


snicoll:
Good catch @ngocnhan-tran1996. I've moved the code snippet to an actual class in case this happens again in the future.

ngocnhan-tran1996:
@snicoll 

Thanks for your review.

PR #33250, opened 2024-07-21
Include query parameters in MockHttpServletRequest

### LINKED ISSUE #33249: query parameters ignored in post request by HtmlUnitRequestBuilder
Bug in both spring boot 3.4.0-M1 and 3.3.2 (spring-test 6.2.0-M6 and 6.1.11).
When submitting a form in a real browser the query string is included in the parameters.
This allow me to write `@PostMapping(params = "form=1")` to detect which form was submitted.

It works in production, but not a unittest.
When using HtmlUnit in a test annotated with `@SpringBootTest` this fails.
`HtmlUnitRequestBuilder.params` reads `this.webRequest.getParameters()`, but `WebRequest.getParameters` ignores query string parameters for post.

I've tested this in spring boot 3.4.0-M1 as I hoped the switch to a newer htmlunit version would have fixed it.
The bug also exists in 3.3.2 and probably all previous versions.

I've put together a minimal testcase here: https://github.com/Crydust/testcasehtmlunit

Crydust:
WebRequest.getParameters() contains parameters from the request body.
We need to insert the query string parameters ourselves.
Parsing logic for the query mirrors WebRequest GET request approach.
Unfortunately the WebRequest.normalize method isn't public
and NameValuePair.normalized is marked as subject to change.
That means we'll need to deal with potential null values ourselves.
We fall back to an empty string to emulate the normalized behavior.

Closes gh-33249

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/HtmlUnitRequestBuilder.java
@@ -37,6 +37,7 @@ import jakarta.servlet.http.HttpSession;
 import org.htmlunit.FormEncodingType;
 import org.htmlunit.WebClient;
 import org.htmlunit.WebRequest;
+import org.htmlunit.http.HttpUtils;
 import org.htmlunit.util.KeyDataPair;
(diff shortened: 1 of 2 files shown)


Crydust:
I should probably review more edge cases before submitting this. Closing for now.

PR #32977, opened 2024-06-07
Support `canEncode()` for `JAXBElement` in `Jaxb2XmlEncoder`

deblockt:
## Overview

This PR provides the ability to serialize a `JAXBElement` using the `Jaxb2XmlEncoder` by updating `canEncode()` accordingly.

## Related Issues

- #30552
- #32972

PR #33258, opened 2024-07-22
Jaxb2XmlEncoder does not support JAXBElement subtypes

olbinski:
# Overview
This PR provides the ability to serialize a JAXBElement using the Jaxb2XmlEncoder by updating canEncode() accordingly.

It is fix this [`Pull Request`](https://github.com/spring-projects/spring-framework/pull/32977)

[Here](https://github.com/spring-projects/spring-framework/blob/a4c2f291d986ccf988f37af5466265a7f534c16a/spring-oxm/src/main/java/org/springframework/oxm/jaxb/Jaxb2Marshaller.java#L632) you can find similar example of usage


bclozel:
Thanks for your contribution @olbinski - sorry we missed this case earlier.

PR #33260, opened 2024-07-23
Update FlashMap.java

lamtrinhdev:
Adding more comments for FlashMap.java


bclozel:
Thanks for your first contribution but I believe this comment is misleading. This is not setting an expiration delay (as in number of seconds until expiration), but actually an Epoch time (as in `System.currentTimeMillis()`).

As a result I don't think this change is valid.

lamtrinhdev:
Dear @bclozel ,

Thank you for your feedback, @bclozel . You are right, let me close this PR.

Thanks,
Lam

PR #33266, opened 2024-07-23
Fix error on Jaxb2RootElementHttpMessageConverter

deblockt:
Can encode can now support child class of JAXBElement

fix issue on https://github.com/spring-projects/spring-framework/pull/32977

See comment  https://github.com/spring-projects/spring-framework/pull/32977#issuecomment-2243482641


snicoll:
@deblockt the code as it is is ok and that change would actually introduced the problem you've described.

PR #33277, opened 2024-07-26
Code sample has wrong reference to configurePathMatching

sheip9:
In interface WebFluxConfigurer, the method to path matching is "configurePathMatching", NOT "configurePathMatch". Detail is in https://github.com/spring-projects/spring-framework/blob/main/spring-webflux/src/main/java/org/springframework/web/reactive/config/WebFluxConfigurer.java#L129


snicoll:
@sheip9 thanks very much for making your first contribution to Spring Framework. To avoid that from happening again, I've moved the code sample to actual code.

Issue #33212, opened 2024-07-13
`ConversionService` cannot convert primitive array to `Object[]`


sbrannen:
## Overview

The `DefaultConversionService` currently cannot convert a primitive array to `Object[]`. This is due to the use of `GenericConversionService.canBypassConvert(...)` in `ArrayToArrayConverter` which assumes a conversion from `int[]` (or any other primitive array) to `Object[]` is unnecessary because "an `int` (or other primitive type) can be converted to an `Object` by returning the source object unchanged," although the latter is not true for primitives.

Thus an attempt to convert from `int[]` to `Object[]` returns the source `int[]` array unchanged and incompatible with `Object[]`.

This strikes me as a bit strange since the following conversions (and other similar conversions) are all supported. See the various `convert*()` test methods in `DefaultConversionServiceTests`.

- `int[]` --> `Integer[]`
- `int[]` --> `String[]`
- `int[]` --> `float[]`
- `Object[]` --> `Integer[]`
- `Object[]` --> `int[]`

The latter makes it clear that you can convert from `Object[]` to `int[]` but not from `int[]` to `Object[]`.

Note that this also affects our varargs support in SpEL expressions. See the various `@Disabled` tests in https://github.com/spring-projects/spring-framework/commit/ae5dd54115bd9e21917a66255a60f03d50d935f0 for examples.

We should decide if we want to support conversions from primitive arrays to `Object[]` and otherwise consider documenting this as a known limitation.

## Related Issues

- #14200

snicoll:
@sbrannen FTR, there was a PR reported against this that I've closed as you were already assigned. In case you haven't started anything and the PR addresses the issue, we can reconsider.

PR #33297, opened 2024-07-31
Feature: Support primitive array to Object[] conversion.

mcvayc:
This commit fixes issue #33212.  

## Context: 
Previously, attempting to convert a primitive array to an Object array caused a java.lang.ClassCastException.  

This happened because the check to bypass array conversion only checked if the types of the array elements were convertible.  This is not sufficient to determine if two arrays are convertible, because—while an int can be cast to an Integer—an int[] cannot be cast to an Integer[].  

## Solution
This commit fixes the issue by only bypassing conversion for arrays if one of the following conditions is true:
- the source and target array elements have the same primitive type
- the source and target array elements have non-primitive types

## Comments
This is my first contribution so please feel free to make suggestions.  I moved the boolean logic into a separate function to improve code clarity, but I'm happy to inline the logic to improve performance.


snicoll:
@mcvayc thanks for the PR but the issue is already assigned so we can't accept this contribution. Going forward, please ask before starting the implementation.

sbrannen:
Hi @mcvayc,

In case you're interested in the approach I took to address this issue, see commit https://github.com/spring-projects/spring-framework/commit/cb6a5baac508921486fd13a7a91cab9d7625868b.

Regards,

Sam

mcvayc:
Thanks, that is simpler.

PR #33306, opened 2024-08-02
`JdbcClient` should be exposed `JdbcTemplate`

ngocnhan-tran1996:
I think `JdbcClient` should be exposed `JdbcTemplate` like `NamedParameterJdbcTemplate` or `AbstractJdbcCall`.


snicoll:
Thanks for the suggestion but the analogy is not the same. `NamedParameterJdbcTemplate` provides additional features on top of `RestTemplate` while `JdbcClient` is a new API that uses `JdbcTemplate` internally.

Issue #27467, opened 2021-09-25
Non-escaped closing curly brace in RegEx results in initialization error on Android


manuel2295:
**Affects:** \<5.2.4.RELEASE>

---

I am trying to use `UriComponentsBuilder.newInstance() `in my android application. But I keep getting this exception upon initialization. This happens starting from version 5.2.4.RELEASE.

```
java.lang.ExceptionInInitializerError
        at org.springframework.web.util.UriComponentsBuilder.newInstance(UriComponentsBuilder.java:172)
 Caused by: java.util.regex.PatternSyntaxException: Syntax error in regexp pattern near index 96
    ^(([^:/?#]+):)?(//(([^@\[/?#]*)@)?(\[[\p{XDigit}:.]*[%\p{Alnum}]*]|[^\[/?#:]*)(:(\d*(?:\{[^/]+?})?))?)?([^?#]*)(\?([^#]*))?(#(.*))?
```

sbrannen:
This appears to be a regression introduced in #24470 that only affects Android applications.

https://github.com/spring-projects/spring-framework/commit/273812f9c5122014260a854d81c27fb2e0a017ac#diff-b7ee321c4c2864dd3493c171d139fc498e11de4ef18cfc57dee3c6f5a84e3dc7L83

sbrannen:
Fixed via f0aa4f4857fbe197c162c9639d24d5bb460eb8f7

sbrannen:
@manuel2295, 

Congratulations on opening your first GitHub issue _ever_. 👍 

The fix for this regression was applied to Spring Framework `5.2.x` in commit 5a5f118efcd399c1473556145a03dd579790179c and will be available in upcoming snapshots for 5.2.18.

We would be grateful if you could verify that the fix works for you in `5.2.18.BUILD-SNAPSHOT` and post back here with a comment to let us know. Details on how to consume snapshots can be found [here](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Artifacts#snapshots).

Thanks

sbrannen:
Our [CI server](https://ci.spring.io/teams/spring-framework/pipelines/) appears to be down at the moment, so I'm not sure when the snapshot will be published.

sbrannen:
[Snapshots for 5.2.18](https://repo.spring.io/ui/native/snapshot/org/springframework/spring-web/5.2.18.BUILD-SNAPSHOT) have been published now. So feel free to try it out and report back.

PR #33316, opened 2024-08-04
Polishing Code

zinzoddari:
HI 🙋🏻‍♀️

1. Remove redundant Regex escapes
2. Polishing Javadoc and comment

Thank you 🙇🏻‍♀️


sbrannen:
Regarding unnecessary escaping in regular expressions, that's actually intentional due to an issue with Android.

- See #27467

PR #33339, opened 2024-08-07
Improve optional parameter checks for HTTP Interface argument

### LINKED ISSUE #31467: Null Handling of HTTP Interface
https://docs.spring.io/spring-framework/docs/6.0.0/reference/html/integration.html#rest-http-interface

I would like to have the default behavior (even without using of [@NonNullApi](https://docs.spring.io/spring-framework/docs/6.1.0-RC1/javadoc-api/org/springframework/lang/NonNullApi.html) or with `@NonNullApi` if that is impossible ) for parameters and return values is, respectively, neither to accept nor to produce null values and use `@Nullable` on a parameter or return value that can be null.

OlgaMaciaszek:
Fixes gh-31467.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/service/invoker/RequestBodyArgumentResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/service/invoker/RequestBodyArgumentResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


spring-web/src/main/java/org/springframework/web/service/invoker/RequestBodyArgumentResolver.java
```diff
@@ -68,9 +71,16 @@ public boolean resolve(
 			return false;
 		}
 
+		parameter = parameter.nestedIfOptional();
```
rstoyanchev, line 74: I don't think we need this here since the only place we check the parameter type is in the reactive section, but we don't expect an `Optional` wrapper around a reactive/async type.

rstoyanchev (commented):
Thanks for the changes.

Overall looks good. One minor suggestion for all the argument resolvers to check for `null` first and exit early. I think that would make it easier to read. Something like:

```java
if (argument == null) {
	Assert.isTrue(parameter.isOptional(), "HttpMethod is required");
	return true;
}

// argument is not null...

```

rstoyanchev (approved):

RAWMEN888 (commented):
[](https://th.xhamster.com/creator/content-protection)

Issue #33342, opened 2024-08-08
Support content disposition setter in  ResponseEntity builder


rfigueroa:
**Affects:** 6.x

Allow Response entity to set content disposition header e.g.

```
ResponseEntity.ok().contentDisposition(contentDisposition).build()
```

---

PR #33343, opened 2024-08-08
Add content disposition setter to ResponseEntity

rfigueroa:
see https://github.com/spring-projects/spring-framework/issues/33342

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ResponseEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/ResponseEntity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/ResponseEntity.java
```diff
@@ -567,6 +577,12 @@ public BodyBuilder contentType(MediaType contentType) {
 			return this;
 		}
 
+		@Override
+		public BodyBuilder contentDisposition(ContentDisposition contentDisposition) {
+			this.headers.setContentDisposition(contentDisposition);
```
GovindarajanL, line 582: This looks well written, Just a small comment 
Consider adding a null check for the contentDisposition parameter. If null values are not allowed, you could throw an IllegalArgumentException

sdeleuze:
I understand the ask and thanks for providing a PR, but it is hard to argue why we should add this one and many others, + there is also `setContentDispositionFormData()`, so I think I would recommend to just use [`headers(Consumer<HttpHeaders> headersConsumer)`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/ResponseEntity.HeadersBuilder.html#headers(java.util.function.Consumer)) for that need. As a consequence, I am going to decline this PR.

PR #33350, opened 2024-08-08
Fix syntax error in `RestClient` documentation

crusherd:
When copy & pasting the error handling example with the lambda expression, Eclipse complains that a ";" is missing in the throw clause. This update adds the missing ";"


sbrannen:
Thanks for the PR, @crusherd.

This has been merged into `6.1.x` and `main`.

For your next PR, please make sure you rebase on `main` before submitting the PR.

For example, this PR should have contained a single commit (https://github.com/spring-projects/spring-framework/pull/33350/commits/e45b4054ca3566e03286295c97244a6552f57137) on top of `main` instead of 3 commits.

crusherd:
Thank you, I will consider this.
The starting point was the "Edit this page" functionality of the webpage itself. From there I created a fork and the PR. On PR creation I got the information, that there is a merge conflict which I resolved in the web browser.

sbrannen:
> The starting point was the "Edit this page" functionality of the webpage itself. From there I created a fork and the PR. On PR creation I got the information, that there is a merge conflict which I resolved in the web browser.

Ah, OK.

I haven't tried that workflow personally, so I'm unfamiliar with it.

Perhaps there's a way to rebase on main after the fact via the web browser. If not, you can rebase on main with a git client and then force push. And otherwise we'll sort it out when merging. For this particular case, I effectively cherry-picked that single commit onto `6.1.x` which sorted out the issues.

Issue #32756, opened 2024-05-03
`MediaTypeAssert` risk false negative with `isNotEqual` assertion


simonbasle:
In `MediaTypeAssert`, we have both `isEqualTo(Object)` and `isEqualTo(String)`, but only `isNotEqualTo(Object)`.
The later is inherited from `AbstractAssert`.

As a `String` is never equal to a `MediaType`, the following test _appears_ to pass but it does for the wrong reasons (false negative):

```java
@Test
void isNotEqualWhenDifferentShouldPass() {
	assertThat(mediaType("application/json")).isNotEqualTo(MediaType.TEXT_HTML_VALUE);
}
```

To make the issue more evident, one can try the following test instead, which doesn't pass with the current implementation:

```java
@Test
void isNotEqualWhenSameShouldFail() {
	assertThatExceptionOfType(AssertionError.class)
			.isThrownBy(() -> assertThat(mediaType("application/json")).isNotEqualTo(MediaType.APPLICATION_JSON_VALUE))
			.withMessageContaining("Media type");
}
```

(note that `isNotEqual` isn't currently tested in `MediaTypeAssertTests`, such tests should be added as part of fixing this issue)

simonbasle:
We should consider if the issue stems from the naming of `isEqualTo(String)`. This method's intention is to parse the provided `String` and compare it to the `actual MediaType`. As such it could be named slightly differently in order to avoid confusion with common `AbstractAssert` assertions.

For instance: `isEqualToValueOf(String)`.

In that case, we can probably defer the introduction of an opposite `isNotEqualToValueOf` method until we get feedback that this would actually be useful.

PR #33351, opened 2024-08-08
Add Javadoc since for MediaTypeAssert.isNotEqualTo()

izeye:
This PR add a Javadoc `@since` tag for the `MediaTypeAssert.isNotEqualTo(String)` method.

See gh-32756


sbrannen:
Please note that `MediaTypeAssert` will first be released with 6.2 which is documented at the class level.

Thus, there is no need to add `@since 6.2` on methods introduced when the class is introduced.

PR #33354, opened 2024-08-08
Fix typo in the bean validation section of the reference manual

GoncaloPT:
Documentation typo:
the error code seems to be on the Person class and therefore Size.person, instead of Size.student.


snicoll:
Thanks @GoncaloPT. There were actual further such wrong use a few lines down that I've polished in https://github.com/spring-projects/spring-framework/commit/df7ac9b308b13f1d340bbb1c2b80b5241d0c9172

PR #33364, opened 2024-08-10
Replace `String.join` with `String.repeat` for generating placeholders in `TableMetaDataContext`

luckygc:
Replaced the use of String.join combined with Collections.nCopies to generate SQL placeholders with the more concise and efficient String.repeat method.

- This change simplifies the code and improves readability.
- Performance might be slightly improved as String.repeat is more direct.

No functionality change; only a refactoring to improve code clarity and efficiency.

Benchmark          Mode  Cnt      Score      Error   Units
JoinTest.nCopies  thrpt    5   5938.720 ±   84.293  ops/ms
JoinTest.repeat   thrpt    5  37268.560 ± 1977.043  ops/ms

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/TableMetaDataContext.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/TableMetaDataContext.java
@@ -350,7 +350,7 @@ public class TableMetaDataContext {
 				throw new InvalidDataAccessApiUsageException(message);
 			}
 		}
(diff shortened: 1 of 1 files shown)


sbrannen:
@luckygc, please note that there is no need to close a PR and open a new one (##33365) identical to the previous one (this PR).

Instead, you can push additional commits to the existing PR.

luckygc:
@sbrannen Thank you for the guidance. I understand now and will make sure to push additional commits to the existing PR next time instead of creating a new one.

PR #33365, opened 2024-08-10
Replace String.join with String.repeat for generating placeholders

luckygc:
Replaced the use of String.join combined with Collections.nCopies to generate SQL placeholders with the more concise and efficient String.repeat method.

- This change simplifies the code and improves readability.
- Performance might be slightly improved as String.repeat is more direct.

No functionality change; only a refactoring to improve code clarity and efficiency.

benchmark

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/TableMetaDataContext.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/TableMetaDataContext.java
@@ -17,7 +17,6 @@
 package org.springframework.jdbc.core.metadata;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


snicoll (changes requested):
@luckygc thanks for the PR, but have you tried to run the build? Your change breaks if `columnCount` is `0`. There's even a test for it:

```
java.lang.IllegalArgumentException: count is negative: -1

	at java.base/java.lang.String.repeat(String.java:4417)
	at org.springframework.jdbc.core.metadata.TableMetaDataContext.createInsertString(TableMetaDataContext.java:354)
	at org.springframework.jdbc.core.simple.TableMetaDataContextTests.testTableWithSingleColumnGeneratedKey(TableMetaDataContextTests.java:150)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

luckygc:
@snicoll Sorry, I didn't consider the case where columnCount is 0. I have revised the code and resubmitted it. The changes have been tested.

snicoll:
Thanks for the update but we'd like to keep the code as it is. It's not in a hot path and the updated code is much less readable.

PR #33366, opened 2024-08-12
Make CacheControl immutable

kashike:
The current `CacheControl` class has (in my opinion) a few drawbacks:
1) it is mutable - there's been times this has caused me (and others, I would assume) headaches when stored in a `static final` field for re-usability
2) there's no way to remove a directive from an already-created `CacheControl`

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/CacheControl.java
+++ b/spring-web/src/main/java/org/springframework/http/CacheControl.java
@@ -51,42 +51,55 @@ import org.springframework.util.StringUtils;
  */
 public class CacheControl {
 
-	@Nullable
-	private Duration maxAge;
-
-	private boolean noCache = false;
-
-	private boolean noStore = false;
-
-	private boolean mustRevalidate = false;
-
-	private boolean noTransform = false;
-
-	private boolean cachePublic = false;
-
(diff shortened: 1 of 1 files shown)


bclozel:
I think I would rather try to make CacheControl immutable than introducing a builder for this. Additionally this builder is not guiding developers to typical setups and might also allow invalid combinations.

Would you like to give the immutability update a try? At first glance this should not require API changes.

kashike:
@bclozel I've pushed an initial set of changes that make `CacheControl` immutable - is this along the lines of what you were looking for? Is there any sort of validation you want to introduce here (ex: disallow public & private) to disallow invalid combinations?

snicoll:
This looks good @kashike, thanks.

snicoll:
@kashike sorry but looking at how it's used downstream we've decided to revert this. It really comes as a surprise that the instance can no longer be reused as it was before. I don't know if having an immutable version of `CacheControl` is worth the change but if you wan to purse this, please open a new issue so that we can discuss if there's another way.

kashike:
Hey @snicoll! I am still interested in this, but before I make another attempt it would be useful to see how `CacheControl` is being used downstream - are you able to share what kind of issues were encountered?

PR #33368, opened 2024-08-12
Fix predicate setup in test

tafjwr:
`Predicate#and()` composes the predicate as its return value, so now, the and clauses are ignored at the assertion.


snicoll:
Thanks @tafjwr - I've reworked the tests so that we don't use any predicate at all.

PR #33370, opened 2024-08-12
Separate out ETag validation and parsing logic

kashike:
This started off as fixing a [bug where weak-prefixed values were not properly validated](https://github.com/spring-projects/spring-framework/compare/main...kashike:feature/etag?expand=1#diff-e2d6218e6585f8b7e32682f6f8f7aed22dbd76d862ec92e86e0902243889556aL1076), and turned into a larger refactor to extract all of the repeated validation and parsing logic into a common `ETag` class.

If you'd like me to go a different route, just let me know. 😃

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ETag.java
+++ b/spring-web/src/main/java/org/springframework/http/ETag.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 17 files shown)


kashike:
Hey @rstoyanchev! The validation logic checks for a prefix of `W/` when it should be checking for `W/"`. As per the spec (see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) and [here](https://datatracker.ietf.org/doc/html/rfc7232#section-2.3)) this allows invalid values to pass through.

rstoyanchev:
Yes, I see now. Could you submit a separate PR with just the bug fix please? There is still chance to have that applied to the upcoming releases this week. As for the broader refactoring, we can discuss that separately.

kashike:
@rstoyanchev #33374

rstoyanchev:
Thanks. I've processed #33374 and backported it to 6.0.x and 5.3.x. 

As for the wider refactoring, coincidentally we had plans also to make efficiency improvements under #33372, and that will include encapsulation of the logic for better re-use. 

I'm closing this for now, but please do follow #33372, and we'll see if there is anything left after it.

Issue #32776, opened 2024-05-07
Document ControllerAdviceBean as internal usage


sinanoezdemir:
### Enhancement/ Misleading description in javadoc. 

In the Javadoc of `ControllerAdviceBean` you can read 

> Encapsulates information about an @ControllerAdvice Spring-managed bean without necessarily requiring it to be instantiated.
The findAnnotatedBeans(ApplicationContext) method can be used to discover such beans. **However, a ControllerAdviceBean may be created from any object, including ones without an @ControllerAdvice annotation.**

[ControllerAdviceBean](https://github.com/spring-projects/spring-framework/blob/168276aaab14456cfcc1b0bd440a26e0682f8f90/spring-web/src/main/java/org/springframework/web/method/ControllerAdviceBean.java#L47)

Reading this I assume, that it should be possible to create ControllerAdviceBeans and register them programmatically (without using the `@ControllerAdvice` annotation).

As far as I understood the [ExceptionHandlerExceptionResolver](https://github.com/spring-projects/spring-framework/blob/168276aaab14456cfcc1b0bd440a26e0682f8f90/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ExceptionHandlerExceptionResolver.java#L78) is the class keeping the ControllerAdviceBeans to apply matching ExceptionHandlers when an exception occurs. 

However, when [initialising the exceptionHandlerAdviceCache](https://github.com/spring-projects/spring-framework/blob/168276aaab14456cfcc1b0bd440a26e0682f8f90/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ExceptionHandlerExceptionResolver.java#L302) it uses ControllerAdviceBean.findAnnotatedBeans which scans the application context for beans annotated with `@ControllerAdvice`. 

The exceptionHandlerAdviceCache is otherwise not modifiable. So there is actually no other way to add ControllerAdviseBeans besides using the annotation.

I think that's why the Javadoc is misleading. 

However, it would be great to have the ability to define ControllerAdviseBeans without using the `@ControllerAdvice` annotation.  

- _Context:_ I am developing in a micro service environment and we have multiple micro services using the same ControllerAdvise which we extracted into a Library. I do not want libraries to contain spring component annotations. I rather would like to instantiate a bean in the configurations of each micro service. With the current implementation it's not possible to "just" remove the annotation and instantiate that bean in a configuration. 

I assumed something similar to this should work.

```java
[...]

PR #33401, opened 2024-08-18
Add tests for `ControllerAdviceBean#resolveBean()` and revise existing tests

tafjwr:
## Overview

I think `ControllerAdviceBean#resolveBean()` should be tested, because it is a public API and depends on the cache system, so I'd like to add tests for that.

## Related Issues

- #32776

### DIFF
--- a/spring-web/src/test/java/org/springframework/web/method/ControllerAdviceBeanTests.java
+++ b/spring-web/src/test/java/org/springframework/web/method/ControllerAdviceBeanTests.java
@@ -89,9 +89,9 @@ class ControllerAdviceBeanTests {
 
 	@Test
 	void equalsHashCodeAndToStringForBeanName() {
-		String beanName = SimpleControllerAdvice.class.getSimpleName();
-		ControllerAdviceBean bean1 = createControllerAdviceBean(SimpleControllerAdvice.class);
-		ControllerAdviceBean bean2 = createControllerAdviceBean(SimpleControllerAdvice.class);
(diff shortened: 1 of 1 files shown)


tafjwr:
I found my branch is stale and the tests are broken. I'm fixing them.

tafjwr:
Fixed. Plus, I noticed that the some test method pairs perform precisely same the tests (`xxxForBeanName` and `xxxForBeanInstance`). There is no longer `ControllerAdviceBean` constructor with Bean type, so `xxxForBeanInstance` test methods could be omitted.

https://github.com/spring-projects/spring-framework/blob/328b72133f39a1027f8796007336c9d2aca73abe/spring-web/src/test/java/org/springframework/web/method/ControllerAdviceBeanTests.java#L98-L106

sbrannen:
> Plus, I noticed that the some test method pairs perform precisely the same tests (`xxxForBeanName` and `xxxForBeanInstance`). There is no longer `ControllerAdviceBean` constructor with Bean type, so `xxxForBeanInstance` test methods could be omitted.

Good catch.

Please remove the obsolete "bean instance" tests.

tafjwr:
Thank for your advice! I removed the obsolete tests on 8d045ad

sbrannen:
This has been merged into `main` in 4ffeddb099f8fcc9eb6b89a98c57d60246ab6ca2 and revised in 346b6f740a98f81e0f55aa641a9bb39fec51580a.

Thanks

Issue #33397, opened 2024-08-16
RestClient doesn't open a scope for the processing of the request


KayWu:
RestTemplate opens the scope when handling requests, whereas RestClient does not.
This will cause the custom ClientHttpRequestInterceptor for RestClient to be unable to access the currently active scope within the intercept method.

https://github.com/spring-projects/spring-framework/blob/2eff5cb46318eab8e1ed9a7abe27428bc6fb1477/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java#L896

https://github.com/spring-projects/spring-framework/blob/2eff5cb46318eab8e1ed9a7abe27428bc6fb1477/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java#L541

KayWu:
Hi @prkksh , DefaultRestClient starts the observation, but not activate the scope in the current thread, while RestTemplate does both things.
We can see that both codes call observation.start(), but only RestTemplate calls observation.openScope.

https://github.com/spring-projects/spring-framework/blob/2eff5cb46318eab8e1ed9a7abe27428bc6fb1477/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java#L892-L904

https://github.com/spring-projects/spring-framework/blob/2eff5cb46318eab8e1ed9a7abe27428bc6fb1477/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java#L533-L544

When observation.openScope() is called, it activates the observation in the current thread, making any code executed later in the same thread aware of this observation. If a new observation is created, the existing observation will be treated as the parent observation.

However, if an observation is started but not activated with openScope(), subsequent code execution will not be aware of this observation.

I have implemented a ClientHttpRequestInterceptor, which performs some operations that create new observations. Since these operations are executed sequentially, the observations should also be linked together in sequence.

This works fine with RestTemplate, but it doesn't work with RestClient because RestClient does not activate the scope, causing these observations to be independent rather than linked.

prkksh:
[PR](https://github.com/spring-projects/spring-framework/pull/33404) for this

PR #33404, opened 2024-08-18
Open observation scope in RestClient

prkksh:
This is to address issue - RestClient doesn't open a scope for the processing of the request #33397. The scope is now opened and properly closed, which ensures that any observations captured during the execution are appropriately handled, even in the presence of exceptions.

Changes
- Refactored exchangeInternal method to manage Observation.Scope more effectively.
- Added a test to verify the proper creation and closure of the observation scope.


bclozel:
Thanks for the proposal @prkksh but the issue was already assigned and work was in progress. You can see that I went with different changes in #33397.

PR #33425, opened 2024-08-25
Add SameSite enum support to ResponseCookie

zinzoddari:
I have added the `SameSite` enum to `ResponseCookie` while ensuring that the existing string-based handling is still supported.

The reason for adding the enum is to enhance type safety. Additionally, if a value that is not provided is entered, it will default to `Lax` as per the specification (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#name-the-samesite-attribute).

I would appreciate it if you could review the PR!

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ResponseCookie.java
+++ b/spring-web/src/main/java/org/springframework/http/ResponseCookie.java
@@ -50,14 +50,14 @@ public final class ResponseCookie extends HttpCookie {
 	private final boolean partitioned;
 
 	@Nullable
-	private final String sameSite;
+	private final SameSite sameSite;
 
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal, but we're going to decline this PR.
Promoting this to an enum doesn't bring much, as `SameSite` can accept any value (and the actual resulting behavior depends on the browser configuration). Exposing this in our public API would require deprecating the `String` variant and would create some edge cases like `builder.sameSite(null)`.

We might revisit this in the future if the SameSite spec evolves and requires more elaborate parsing or features on the server side.

Thanks for your contribution!

PR #33449, opened 2024-08-30
Recommend use of self injection instead of `AopContext.currentProxy()`

quaff:
Self-Injection is better than horrendous `AopContext.currentProxy()`

### DIFF
--- a/framework-docs/modules/ROOT/pages/core/aop/proxying.adoc
+++ b/framework-docs/modules/ROOT/pages/core/aop/proxying.adoc
@@ -193,9 +193,8 @@ in the advice associated with a method invocation getting a chance to run.
 Okay, so what is to be done about this? The best approach (the term "best" is used
 loosely here) is to refactor your code such that the self-invocation does not happen.
 This does entail some work on your part, but it is the best, least-invasive approach.
-The next approach is absolutely horrendous, and we hesitate to point it out, precisely
-because it is so horrendous. You can (painful as it is to us) totally tie the logic
-within your class to Spring AOP, as the following example shows:
+The next approach is a little counter-intuitive, you can inject self by field or method
+(constructor injection is not supported), as the following example shows:
 
 [tabs]
 ======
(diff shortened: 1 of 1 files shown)


framework-docs/modules/ROOT/pages/core/aop/proxying.adoc
```diff
@@ -205,9 +204,12 @@ Java::
 ----
 	public class SimplePojo implements Pojo {
 
+	    @Autowired
+        private SimplePojo self;
```
sbrannen, line 208:
```suggestion
        private Pojo self;
```

Unless you are certain CGLIB proxies are being created, this would need to be `Pojo`.

Though, there's no need to update this PR. I'll address that after merging.

sbrannen:
Hi @quaff,

> Self-Injection is better than horrendous `AopContext.currentProxy()`

Yes, I totally agree that with! 👍 

However, after further analysis, I realized that we cannot use `@Autowired` in the example you modified in this PR.

In addition, I noticed there was room for improvement in other areas.

In light of that, I am closing this PR as superseded by the following.

- #33454
- #33455 

Thanks anyway for the PR!

sbrannen:
@quaff, we may not actually have a "self injection" example anywhere in the reference manual. So, if you'd like to submit a new PR with such an example, feel free to do so.

It should probably go in the new _Self Injection_ sidebar here: https://raw.githubusercontent.com/spring-projects/spring-framework/main/framework-docs/modules/ROOT/pages/core/beans/annotation-config/autowired.adoc#beans-autowired-annotation-self-injection

quaff:
> @quaff, we may not actually have a "self injection" example anywhere in the reference manual. So, if you'd like to submit a new PR with such an example, feel free to do so.
> 
> It should probably go in the new _Self Injection_ sidebar here: https://raw.githubusercontent.com/spring-projects/spring-framework/main/framework-docs/modules/ROOT/pages/core/beans/annotation-config/autowired.adoc#beans-autowired-annotation-self-injection

@sbrannen I created https://github.com/spring-projects/spring-framework/pull/33462.

PR #33452, opened 2024-08-30
Expand acronyms `FQN` and `FQCN`

quaff:
We should insist on using either one to keep consistency, `FQCN` is more common in the Java ecosystem.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/ConstructorResolver.java
@@ -1242,7 +1242,7 @@ class ConstructorResolver {
 	 * Return a {@link Predicate} for a parameter type that checks if its target
 	 * value is a {@link Class} and the value type is a {@link String}. This is
 	 * a regular use cases where a {@link Class} is defined in the bean
-	 * definition as an FQN.
+	 * definition as fully-qualified class name.
 	 * @param valueType the type of the value
 	 * @return a predicate to indicate a fallback match for a String to Class
 	 * parameter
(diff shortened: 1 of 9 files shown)


sbrannen (changes requested):
I'd argue that we should follow best practices for technical writing by avoiding the use of undefined acronyms.

In other words, we should never use either FQN or FQCN without first defining what is meant by the acronym used, and if we are not going to use the term or acronym multiple times in the same paragraph/section/Javadoc, we should just omit the use of the acronym altogether.

Would you like to revise this PR accordingly?

quaff:
> we should follow best practices for technical writing by avoiding the use of undefined acronyms.

Agreed totally.

> Would you like to revise this PR accordingly?

I'd like to, do you mean replacing all `FQCN` with `fully-qualified class name` ?

sbrannen:
> I'd like to, 

👍 

> do you mean replacing all `FQCN` with `fully-qualified class name` ?

Yes, when it's used in text/Javadoc/comments.

When it's used in a code/syntax example, I usually go with something like `<fully-qualified class name>.<method-name>` instead of `FQN.methodName`.

quaff:
> > I'd like to,
> 
> 👍
> 
> > do you mean replacing all `FQCN` with `fully-qualified class name` ?
> 
> Yes, when it's used in text/Javadoc/comments.
> 
> When it's used in a code/syntax example, I usually go with something like `<fully-qualified class name>.<method-name>` instead of `FQN.methodName`.

@sbrannen done.

sbrannen:
This has been merged into `main` in 019c0b1d4e59c6f158c16b7fc6f7cecd6e96e89d and polished in 717b972f88c13ad774e533d1571a5668971d779f.

Thanks

PR #33459, opened 2024-09-01
Update Content-Length when body changed by Interceptor

imzhoukunqiang:
Hi maintainers, while working with the framework, I discovered a bug that I believe warrants attention.

In the `org.springframework.http.client.InterceptingClientHttpRequest` , the request body may be changed by `interceptors`, but `Content-Length` in the request headers didn't. This may cause the server receive the incorrect request body. 
I created a simple demo to show this bug . [DEMO](https://github.com/imzhoukunqiang/spring-http-request-bug-demo)

The fix for this is to reset the `Content-Length` before sending the request at `org.springframework.http.client.InterceptingClientHttpRequest.InterceptingRequestExecution#execute`

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/InterceptingClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/InterceptingClientHttpRequest.java
@@ -93,7 +93,10 @@ class InterceptingClientHttpRequest extends AbstractBufferingClientHttpRequest {
 				 * After intercepted by the interceptors, the length of the body may change.
 				 * Set the latest body.length to "Content-Length"
 				 */
-				request.getHeaders().setContentLength(body.length);
+				long contentLength = request.getHeaders().getContentLength();
+				if (contentLength >-1 & contentLength != body.length) {
+					request.getHeaders().setContentLength(body.length);
+				}
 
 				HttpMethod method = request.getMethod();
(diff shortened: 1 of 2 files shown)


bclozel:
I am not sure we should do this. Do we really set the content length header in all cases and for all converters? If an interceptor is changing the request body I would argue it needs to change the content type as well as the content length of it was set already.
I think some http clients can send bodies with chunked transfer encoding.

imzhoukunqiang:
> I am not sure we should do this. Do we really set the content length header in all cases and for all converters? If an interceptor is changing the request body I would argue it needs to change the content type as well as the content length of it was set already. I think some http clients can send bodies with chunked transfer encoding.

I agree with your point of view. I'm not sure whether setting `Content-Length` is the responsibility of the user or the framework. However, typically when we use `RestTemplate`, we do not set `Content-Length`, and it is handled by the framework.

And if after executing the interceptor chain, the `Content-Length` exists in the request header and does not equal `body.length`, should it be corrected to the correct value?

bclozel:
Your description makes sense. However, isn't your PR setting it in all cases?

imzhoukunqiang:
> Your description makes sense. However, isn't your PR setting it in all cases?

Updated.

bclozel:
Thanks for your contribution @imzhoukunqiang , this is now merged.

PR #33462, opened 2024-09-02
Add "self injection" code example

quaff:


bclozel:
Thanks for the proposal, but I don't think this code snippets really helps here, especially for a use case we're advising against.

Issue #31907, opened 2023-12-25
Allow `Propagation.NOT_SUPPORTED` with `@TransactionalEventListener`


xak2000:
After #30679 and #31414 this doesn't work anymore:

```java
@TransactionalEventListener(fallbackExecution = true)
@Transactional(propagation = Propagation.NOT_SUPPORTED)
public void onUserDeleted(UserDeletedEvent userDeletedEvent) {
	someOtherService.anyTransactionalOrNonTransactionalMethod();
}

```

I think, `NOT_SUPPORTED` should be allowed here, as any other `Propagation` that implies suspending of the current transaction (that is in an undefined state, as @odrotbohm mentioned).

The whole purpose of restriction it to support only `REQUIRES_NEW` was to catch invalid usages of `@TransactionalEventListener`, right?

But `NOT_SUPPORTED` is also totally valid usage, as the method will never be executed in a current transaction (that was active at the moment of calling the listener). Current transaction will be suspended in exactly the same way as it is supsended with `REQUIRES_NEW`.

Probably, `Propagation.NEVER` is also valid here, as it will never execute the method if a transaction exists. But this has much less practical use: why do you need `Propagation.NEVER` on a `@TransactionalEventListener` in the first place? The transaction will always be active at the moment of calling such listener, except when `fallbackExecution = true` and listener was actually called on an event, published from non-transactional method. So, while I doubt it will be practial case, it is still valid in my opinion and should not prevent Application context to start.

Anyway, most importantly `Propagation.NOT_SUPPORTED` is totally valid and perfectly practical usage of `@TransactionalEventListener`. I would even argue that this is the best combination as `Propagation.NOT_SUPPORTED` serves as a failsafe way to prevent execution of the listener in an old transaction (that is in a "bad" state at the moment of execution) and the decision to start a new transaction or not for the code inside of the listener ifself is usually out of the scope of the listner and more in the scope of the business-method, that the listener calls.

`@TransactionalEventListener` + `@Transactional(propagation = Propagation.NOT_SUPPORTED)` means that `someOtherService.anyTransactionalOrNonTransactionalMethod()` will be executed correctly if it is annotated with `@Transactional` (with any `propagation `) or if it is not annotated at all.

[...]

Issue #31414, opened 2023-10-11
Enforce REQUIRES_NEW for correct transaction configuration on transactional event listeners


jhoeller:
Following up on https://github.com/spring-projects/spring-framework/issues/30679#issuecomment-1757848560, we decided to make the check more restrictive: only accepting `@Transactional` when declared with REQUIRES_NEW on an `@TransactionalEventListener` method.

Issue #30679, opened 2023-06-15
Detect invalid transaction configuration for transactional event listeners


odrotbohm:
## Problem

It's common practice to use a simple (no custom attributes) `@Transactional` on methods to execute them wrapped in a transaction. Unfortunately, for methods declared as `@TransactionalEventListener`, this doesn't work as expected:

```java
@Component
class SomeListener {

  @Transactional
  @TransactionalEventListener
  void on(SomeEvent event) { /* … */ }
}
```

Those methods are invoked in the transaction cleanup phase of an original, other transaction. This means that this particular, other transaction is still running, although in an undefined state. With a simple `@Transactional` declared on those methods, the code would be executed in an undefined transactional state. An obvious solution to this problem is to declare a `Propagation` of `REQUIRES_NEW` on the listener method, so that it will run in a new transaction:

```java
@Component
class SomeListener {

  @Transactional(propagation = Propagation.REQUIRES_NEW)
  @TransactionalEventListener
  void on(SomeEvent event) { /* … */ }
}
```

## Proposed solution

It would be nice if the erroneous arrangement shown above would be rejected, ideally at bootstrap (potentially even AOT?) time. We have to consider that a plain `@Transactional` is still valid in the case of the method declared to be invoked asynchronously (using `@Async`). In other words, a declaration like this is still valid, as the asynchronous invocation triggers the execution on a separate thread and thus a clean slate regarding transactions.

```java
@Component
class SomeListener {

  @Async
  @Transactional
  @TransactionalEventListener
  void on(SomeEvent event) { /* … */ }
}
```

quaff:
After https://github.com/spring-projects/spring-framework/commit/842569c9e519ec7d11fc86691577cee067c7ef47
```java
@Component
class SomeListener {

  @Transactional(propagation = Propagation.REQUIRES_NEW)
  @TransactionalEventListener
  void on(SomeEvent event) { /* … */ }
}
```
will not be valid, is it intentional? @jhoeller

jhoeller:
Good point, the check seems too restrictive now. I'll revise it to be lenient towards REQUIRES_NEW.

jhoeller:
[...]

PR #33470, opened 2024-09-03
Throw runtime error for `@TransactionalEventListener` when `@Transactional` propagation level is not `REQUIRES_NEW` or `NOT_SUPPORTED`

Jaeyoung22:
After #30679, #31414, and #31907, if a method is a `@TransactionalEventListener` with a propagation level other than `@Transactional(propagation = Propagation.REQUIRES_NEW)` or `@Transactional(propagation = Propagation.NOT_SUPPORTED)`, a runtime error will be thrown.

However, there is no runtime error yet for `@Transactional` with all propagation level on the top of the class.

For classes that use `@Service` with `@Transactional` at the top and don't attach this annotation to the method where the write operation takes place, there were cases where it didn't work as intended because no runtime error was thrown.

So I'd like to contribute to this.


Jaeyoung22:
@jhoeller 
hello.
I hope this isn't rude, but I haven't gotten any feedback in quite some time.

Do you mind if I ask for a review?
Please let me know if there are any issues with the PR.

PR #33472, opened 2024-09-04
Add exception message when logging resolved exceptions

LucaBarden:
For more insight into which exception has been resolved, the thrown exception will be included in the log message accordingly

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractHandlerExceptionResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractHandlerExceptionResolver.java
@@ -244,7 +244,7 @@ public abstract class AbstractHandlerExceptionResolver implements HandlerExcepti
 	 */
 	protected void logException(Exception ex, HttpServletRequest request) {
 		if (this.warnLogger != null && this.warnLogger.isWarnEnabled()) {
-			this.warnLogger.warn(buildLogMessage(ex, request));
+			this.warnLogger.warn(buildLogMessage(ex, request), ex);
 		}
 	}
 
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractHandlerExceptionResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractHandlerExceptionResolver.java
@@ -255,7 +255,7 @@ public abstract class AbstractHandlerExceptionResolver implements HandlerExcepti
 	 * @return the log message to use
 	 */
 	protected String buildLogMessage(Exception ex, HttpServletRequest request) {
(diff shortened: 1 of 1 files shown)


Duckulus:
Huge W

rstoyanchev (commented):
The warnLogger is intentionally minimal, and may be used in production for example. At most we could add the Exception message in addition to the Exception name. Logging the full exception is not in scope I'm afraid. You can, however, override the `buildLogMessage` method.

LucaBarden:
> The warnLogger is intentionally minimal, and may be used in production for example. At most we could add the Exception message in addition to the Exception name. Logging the full exception is not in scope I'm afraid. You can, however, override the `buildLogMessage` method.

We are using it in production and have searched for issues too long most of the time and would have appreciated the Stacktrace included.

bclozel:
I agree with Rossen's assessment. A large part of the Spring community will be concerned by logs becoming too noisy. I'm sure lots of teams do not use Kibana or similar and will be negatively impacted by this change.

Please update your PR to log the exception message in the default implementation of `buildLogMessage`.

LucaBarden:
Let me know if you had a different style in mind

spring-webmvc/src/main/java/org/springframework/web/servlet/handler/AbstractHandlerExceptionResolver.java
```diff
@@ -255,7 +255,7 @@ protected void logException(Exception ex, HttpServletRequest request) {
 	 * @return the log message to use
 	 */
 	protected String buildLogMessage(Exception ex, HttpServletRequest request) {
-		return "Resolved [" + LogFormatUtils.formatValue(ex, -1, true) + "]";
+		return "Resolved [" + LogFormatUtils.formatValue(ex, -1, true) + "] :" + ex.getMessage();
```
rstoyanchev, line 258: This should be in the formatValue call:
```suggestion
		return "Resolved [" + LogFormatUtils.formatValue(ex + "] :" + ex.getMessage(), -1, true);
```

rstoyanchev:
Thanks for the updates. Apologies for any confusion, but on closer look, `Throwable#toString` [already does this](https://github.com/openjdk/jdk/blob/9ebc2ecbf613da3bcee1dd5e8920a26d5f6d6df7/src/java.base/share/classes/java/lang/Throwable.java#L523-L527), so that's about as far as we can take it. If you want full exception logging, you can override `buildLogMessage` . 

Note that the Spring MVC Auto-Configuration in Spring Boot makes it easy to replace the `ExceptionHandlerExceptionResolver` instance through a `WebMvcRegistrations` bean. See [the docs on that](https://docs.spring.io/spring-boot/reference/web/servlet.html#web.servlet.spring-mvc.auto-configuration).

PR #33473, opened 2024-09-04
Avoid empty array allocations in ResolvableType

kilink:
Avoid allocating empty arrays in a few places in ResolvableType.


sdeleuze:
Merged, thanks.

On a very simple webapp, this change reduces the number of `ResolvableType[]` instances from 439 to 395. On bigger apps, it will likely lead to bigger gains.

kilink:
> Merged, thanks.
> 
> On a very simple webapp, this change reduces the number of `ResolvableType[]` instances from 439 to 395. On bigger apps, it will likely lead to bigger gains.

I should have included some data, in the app I was looking at a heap dump for, I saw ~700 duplicate empty arrays coming from ResolvableType, or about 11kb of waste.

Issue #27619, opened 2021-10-28
Efficient webjars version resolution natively in Spring


dsyer:
Spring (MVC and Webflux) has a `ResourceResolver` abstraction that can be used to resolve the versions in webjars, avoiding the need to maintain the version explicitly in 2 or more places (build file and HTML source). E.g. (from Petclinic):

```html
 <script src="/webjars/jquery/jquery.min.js"></script>
```

Resolves to `classpath:/META-INF/resources/webjars/jquery/<version>/jquery.min.js` at runtime.

Spring Boot carries the responsibility of configuring the resource resolver, and currently it uses the `webjars-locator-core` (https://github.com/webjars/webjars-locator-core) library to do that, so version resolution only works if that library is on the classpath. The `WebJarsAssetLocator` from that library has a very broad and powerful API for locating files inside webjars, but there are some issues, namely:

1. It is fairly inefficient, since it scans the whole `/META-INF/resources/webjars` classpath on startup (in a constructor!).
2. It has 2 awkward dependencies (github classpath scanner and jackson)
3. It doesn't work in a native image (https://github.com/spring-projects-experimental/spring-native/issues/157) because of the classpath scanning

But we don't need `webjars-locator-core` to just do version resolution, which is all Spring Boot offers, because webjars have a very well-defined structure. They all have a `pom.properties` with the version in it, and they only use a handful of well-known group ids, so they are easy to locate. It might be a good idea to implement it in Framework, since it is so straightforward and only depends on reading resources from the classpath.

All of the issues above could be addressed just by providing a simpler version resolver natively (and configuring the resource config in a native image with a hint).

vpavic:
I gave this solution a spin in one of my projects and so far the impressions are good.

Is it OK if I refine it a bit and provide a PR to replace the existing (WebJars Locator based) `WebJarsResourceResolver` implementations?

bclozel:
@vpavic I don't think we should implement the resolution mechanism in Spring Framework directly, as this is webjars-locator's purpose in the first place. As far as I know there is no concrete specification here to follow, so we might derive from current or future behavior of the official library.

GraalVM native becoming an important topic in the Java community, I think we should instead think about proper native support in webjars. Frameworks could instead drive webjars-locator at build time; the library could resolve all available resources and dump them in an index and possibly GraalVM resource configuration, since those resources will be needed at runtime. At runtime, Frameworks could then query the locator library for resources and it could use its own index (no scanning involved!) to resolve resources.

[...]

PR #33495, opened 2024-09-06
Simplify WebJars support

vpavic:
https://github.com/spring-projects/spring-framework/issues/27619#issuecomment-2333340105


bclozel:
It looks like this proposal is making `WebJarsResourceResolver` depend on two different types: `WebJarVersionLocator` (from `org.webjars:webjars-locator-lite`) and `WebJarAssetLocator` (from `org.webjars:webjars-locator-core`). Because those two types are used by public constructors, I think this class now needs both dependencies to be on the classpath to be loaded. Doesn't this defeats the purpose? Isn't it going to make things harder for maintenance and deprecation in the future?

Please advise.

vpavic:
> I think this class now needs both dependencies to be on the classpath to be loaded.

Can you clarify why do you believe this is the case?

Before submitting this PR, I built Framework snapshot with these changes locally and used it in a sample app to test both scenarios - with only `org.webjars:webjars-locator-lite` available on classpath, and with only `org.webjars:webjars-locator-core` available.

bclozel:
I have just tried to create a sample application with the following dependencies:

```
dependencies {
	implementation 'org.springframework.boot:spring-boot-starter-web'
	implementation 'org.webjars:webjars-locator-lite'
	testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
}
```

Trying to instantiate manually the resolver like so:

```
@SpringBootApplication
public class WebjarsApplication {

		WebJarsResourceResolver webJarsResourceResolver = new WebJarsResourceResolver(new WebJarVersionLocator());
		SpringApplication.run(WebjarsApplication.class, args);
	}

}
```

Yields the following:

```
$ ./gradlew assemble

> Task :compileJava FAILED
/Users/bclozel/workspace/tmp/webjars/src/main/java/com/example/webjars/WebjarsApplication.java:13: error: cannot access WebJarAssetLocator
                WebJarsResourceResolver webJarsResourceResolver = new WebJarsResourceResolver(new WebJarVersionLocator());
                                                                  ^
  class file for org.webjars.WebJarAssetLocator not found
1 error

```

Am I missing something?

vpavic:
Ah, I see - the problem is when instantiating resolver in the application code, which doesn't compile if both implementations are not present on the classpath at that time. Admittedly in my tests I was focused on scenarios where it's the Framework instantiating resolver based on what's available on the classpath.

But shouldn't it be possible to overcome this issue if constructor accepted simply `Object` for the midterm while both options are supported? Such constructor could be immediately deprecated, in favor of a factory method - potentially even adding one that allows `WebJarsResourceResolver` to be created using `BiFunction<String, String, String>` which would increase its ability to be reused with different locator implementations.

vpavic:
This works for me:

```java
public WebJarsResourceResolver(Object webJarLocator) {
	this.webJarLocator = switch (webJarLocator.getClass().getName()) {
		case "org.webjars.WebJarVersionLocator" -> ((org.webjars.WebJarVersionLocator) webJarLocator)::fullPath;
		case "org.webjars.WebJarAssetLocator" -> ((org.webjars.WebJarAssetLocator) webJarLocator)::getFullPathExact;
		default -> throw new IllegalArgumentException("Unsupported WebJar locator implementation");
	};
}
```

While such constructor isn't pretty, it's only for the midterm and doesn't break anyone while providing flexibility for the future while avoiding the following (IMO) downsides of the current solution:

- keeping the same inflexibility that was present until now (what happens if there's another WebJar locator implementation few years down the line?)
- code duplication
- you end up with awkwardly named `LiteWebJarVersionLocator` as your only implementation
- breaking the continuity (version control history wise) of `WebJarsResourceResolver`

bclozel:
Thanks for the proposal, but I think we're fine with the current tradeoffs.

PR #33496, opened 2024-09-06
Avoid storing duplicate empty array in MethodParameter field

kilink:
Avoid storing duplicate empty arrays in the parameterAnnotations field of MethodParameter.

In a heap dump of a sample app, there were 1,293 empty `Annotation[]` duplicates, or about ~20kb of waste.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/MethodParameter.java
+++ b/spring-core/src/main/java/org/springframework/core/MethodParameter.java
@@ -647,7 +647,7 @@ public class MethodParameter {
 				// for inner classes, so access it with the actual parameter index lowered by 1
 				index = this.parameterIndex - 1;
 			}
-			paramAnns = (index >= 0 && index < annotationArray.length ?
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/core/MethodParameter.java
```diff
@@ -647,7 +647,7 @@ public Annotation[] getParameterAnnotations() {
 				// for inner classes, so access it with the actual parameter index lowered by 1
 				index = this.parameterIndex - 1;
 			}
-			paramAnns = (index >= 0 && index < annotationArray.length ?
+			paramAnns = (index >= 0 && index < annotationArray.length && annotationArray[index].length > 0 ?
```
sdeleuze, line 650: This optimization introduces a small change of behavior, but it looks acceptable to me as adaptation of an empty array is IMO unlikely to produce something else than an empty array.

PR #33507, opened 2024-09-09
Avoid empty array allocations in AnnotationTypeMapping

kilink:
Avoid empty array allocations in a few places in AnnotationTypeMapping.

In a heap dump for a production app, there were a couple hundred empty int arrays referenced from AnnotationTypeMapping instances (~3kb of waste); there were also about ~60 empty MirrorSet arrays (~1kb of waste).

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMapping.java
+++ b/spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMapping.java
@@ -69,6 +69,8 @@ final class AnnotationTypeMapping {
 
 	private static final MirrorSet[] EMPTY_MIRROR_SETS = new MirrorSet[0];
 
+	private static final int[] EMPTY_INT_ARRAY = new int[0];
+
 
 	@Nullable
 	private final AnnotationTypeMapping source;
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/core/annotation/AnnotationTypeMapping.java
```diff
@@ -750,7 +758,7 @@ class MirrorSet {
 
 			private int size;
 
-			private final int[] indexes = new int[attributes.size()];
+			private final int[] indexes = attributes.size() > 0 ? new int[attributes.size()] : EMPTY_INT_ARRAY;
```
sdeleuze, line 761: @kilink Could you please double check if you have a use case where `attributes.size() == 0` at this line? I was not able to find one in Petclinic JDBC. If you can't, please revert that specific change (and keep other ones that looks legit).

sdeleuze:
@kilink Any chance you could provide a feedback? If not, I plan to merge your PR without the commented change.

kilink:
> @kilink Any chance you could provide a feedback on my question above? If not, I plan to merge your PR without the commented change.

Sorry, have been somewhat busy and hadn't got a chance to look at it. What you said it probably correct, I was likely just over-aggressive in trying to eliminate possible empty array allocations here. I can push up a change that removes that line.

kilink:
I double-checked, I only see the problem in `AnnotationTypeMapping` (and indirectly `TypeMappedAnnotation`):

image

I went ahead and reverted that line.

PR #33515, opened 2024-09-10
Replace `e.g.` with `for example`

quaff:
Latin abbreviation is not readable for non-native English speakers.


sbrannen (changes requested):
I agree that we should stop using `e.g.` in our documentation; however, we should do that consistently across Javadoc and the reference manual.

Searching for `e.g.` (ignoring case) reveals 1168 occurrences in `*.java` files and 76 occurrences in `*.adoc` files on `main`. We also use `e.g.` in `*.dtd` and `*.xsd` files.

So, if we're going to do this, we should do it for all occurrences.

In addition, for statements like `E.g. the following class implements a singleton`, we should write `For example, the following class implements a singleton` (note the comma).

For places where ` e.g. ` comes in the middle of a sentence, it should typically be replaced with `, for example, ` and sometimes with ` &mdash; for example, `, depending on the context.

Please let us know if you'd like to update this PR to address all such occurrences.

quaff:
I'll try.

quaff:
@sbrannen I've updated.

PR #33529, opened 2024-09-12
Ensure that SseEmitter outputs UTF-8 encoded text

YMNNs:
The character encoding is not specified in the `extendResponse` method of `SseEmitter`, so the character encoding actually received by the browser is `ISO_8859_1`. 

This PR modifies the response header in the `extendResponse` method to encode it in `UTF-8`.

Should we add a constructor that accepts other character encodings?

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java
@@ -46,6 +46,8 @@ public class SseEmitter extends ResponseBodyEmitter {
 
 	private static final MediaType TEXT_PLAIN = new MediaType("text", "plain", StandardCharsets.UTF_8);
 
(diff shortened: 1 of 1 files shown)


bclozel:
I don't believe this is necessary, ISO_8859_1 is the expected encoding for servlet applications.

YMNNs:
> I don't believe this is necessary, ISO_8859_1 is the expected encoding for servlet applications.

We should consider that this is an international project, and users may send information in other languages ​​through SSE, which will cause the browser to receive garbled characters.

Another reason is that line 47 (`private static final MediaType TEXT_PLAIN = new MediaType("text", "plain", StandardCharsets.UTF_8);`) has specified the encoding as UTF-8, and using two encodings in the same class will destroy consistency. I believe this situation is caused by omissions in coding.

YMNNs:
@bclozel What do you think then?

bclozel:
The `TEXT_PLAIN` media type you're pointing is not sent with the response, but merely used to let the message converters know how they should convert the data using UTF-8.

I think server sent events should always be decoded as UTF-8 content, so there is no need to add a charset parameter to the media type. If the data is garbled, it means that you didn't serialize data properly in the first place.

PR #33538, opened 2024-09-13
RequestParam Capitalization

Zemke:
WIP. Just wondering if something like this has a chance of making it. The idea is to have essentially case-insensitive `@RequestParam`s. Especially useful when using Enums.


bclozel:
Thanks for the proposal, but we're not going to pursue this. We would rather rely on conversion services for this and not expose this information at the controller level.

Issue #31604, opened 2023-11-14
Make WebSocketMessageBrokerStats optionally export stats to micrometer


nyilmaz:
Currently we access stats via logs which `WebSocketMessageBrokerStats` produce. Would you consider exporting those metrics to any stats collector? If interested, we can create a pr for this.

https://stackoverflow.com/questions/57481608/how-to-access-and-export-websocketmessagebrokerstats-internal-counts

bclozel:
Sorry for the delayed response.

Spring Framework only depends on the `Observation` API and this use case doesn't fit as `Observation` is all about timer metrics and traces. Here, we could expose stats information and let developers bind `Gauges` to their `MeterRegistry` using the "micrometer-core" dependency.

The log line contains several parts:
```
WebSocketSession[...], stompSubProtocol[...], stompBrokerRelay[...], inboundChannel[...], outboundChannel[...], sockJsScheduler[...]
```

The "inboundChannel", "outboundChannel" and "sockJsScheduler" are all Stringly typed as this information is extracted directly from `ThreadPoolExecutor#toString`. I don't think we can expose more detailed stats. At best, we could expose the `Executor` directly and let developers extract information from there.

As for the "WebSocketSession", "stompSubProtocol", "stompBrokerRelay", they all have dedicated `Stats` interface that are already public. We could add methods like the following to `WebSocketMessageBrokerStats` and deprecate the String versions to avoid redundancy:

```java
@Nullable
public StompSubProtocolHandler.Stats getStompSubProtocolStats() {
	return (this.stompSubProtocolHandler != null ? this.stompSubProtocolHandler.getStats() : null);
}
```

This would allow applications to register gauges with the metric names and tags of their choices:

```java
MeterRegistry meterRegistry = ...;

Gauge.builder("spring.stomp.frames", stats.getStompSubProtocolStats(), StompSubProtocolHandler.Stats::getTotalConnect)
			.tag("type", "CONNECT")
			.description("number of CONNECT frames processed")
			.register(meterRegistry);

Gauge.builder("spring.stomp.frames", stats.getStompSubProtocolStats(), StompSubProtocolHandler.Stats::getTotalConnected)
			.tag("type", "CONNECTED")
			.description("number of CONNECTED frames processed")
			.register(meterRegistry);
```

The only downside here is that such metrics would probably fit more with `Counter`, but counters need to be incremented manually by the instrumentation rather than polling like Gauges do.

@nyilmaz What do you think about this proposal?

[...]

PR #33543, opened 2024-09-16
Replace deprecated references in WebSocketMessageBrokerStats.toString()

izeye:
This PR replaces deprecated references in the `WebSocketMessageBrokerStats.toString()`.

- See gh-31604


rstoyanchev (commented):
The only potential issue is if those methods are overridden by a sub-class, or if the corresponding info methods on the component providing the stats is overridden. Granted that seems unlikely, but do you see any harm in leaving toString as is for the time being?

izeye:
@rstoyanchev No, I just wanted to clean up the deprecated references.

rstoyanchev:
Okay, thanks for the suggestion. We can defer that until the deprecations are removed since it is internal only.

Issue #33545, opened 2024-09-17
`ZoneIdEditor` throws wrong exception type for `TypeConverterSupport`


harmonic-ben:
Spring Framework Version: Latest (6.1.13) and below

`spring-beans/src/main/java/org/springframework/beans/propertyeditors/ZoneIdEditor.java` calls `ZoneId.of` which throws a `java.time.DateTimeException`.

But `org.springframework.beans.TypeConverterSupport.convertIfNecessary(String, Object, Class<T>, TypeDescriptor)` is expecting an `IllegalArgumentException`.

This manifests in a variety of ways. Best example is in `web-mvc` assuming a controller with:

```java
   @GetMapping("/")
    public void get(
            @RequestParam(name = "integer", required = false)
            int integer,
            @RequestParam(name = "zoneId")
            ZoneId zoneId) {

    }
 ```

This means a `get` request of `/?integer=helloworld` results in a `MethodArgumentTypeMismatchException` which can easily be bound to 400 response.

However a `get` request of `/?zoneId=helloworld` results in a `DateTimeException` which binds to 500 response.

The integer parameter binding works because it throws `NumberFormatException` which extends from `IllegalArgumentException`.

PR #33550, opened 2024-09-17
ZoneIdEditor Add catch exception

ngocnhan-tran1996:
Issue https://github.com/spring-projects/spring-framework/issues/33545

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/propertyeditors/ZoneIdEditor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/propertyeditors/ZoneIdEditor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-beans/src/main/java/org/springframework/beans/propertyeditors/ZoneIdEditor.java
```diff
@@ -38,7 +39,13 @@ public void setAsText(String text) throws IllegalArgumentException {
 		if (StringUtils.hasText(text)) {
 			text = text.trim();
 		}
-		setValue(ZoneId.of(text));
+		try {
+
+			setValue(ZoneId.of(text));
+		}
+		catch (Exception ex) {
+			throw new IllegalArgumentException("Failed to convert ZoneId for " + text, ex);
```
mdeinum, line 47: Other editors that throw exceptions wrap the `text` inside `'` chars, so `"Failed to convert ZoneId for '" + text + "'"` as the message. That way it is easy to identify trailing and/or leading spaces.

jhoeller:
Note that I've prepared a fix for #33545 locally already, just not committed it to the 6.1.x branch yet where it will get forward-merged to the main branch. This will get pushed to the repo later today.

@ngocnhan-tran1996 please pay attention to assignments on issues, otherwise an unrequested PR is effectively duplicated work. Thanks for your efforts, in any case!

In terms of exception message, I'm actually using the original exception message as-is without extra words or quoting of the value since the given value is contained there in the `DateTimeException` anyway. Also, we're exclusively catching `DateTimeException` there, not general `Exception`.

PR #33556, opened 2024-09-17
Add read timeout setter on HttpComponentsClientHttpRequestFactory

cmagnuson:
This brings the HttpComponentsClientHttpRequestFactory into alignment with JDK, Jetty, OkHttp3 and SimpleClient ClientHttpRequestFactory's.

The read timeout is a pretty common value to set, so I think it is helpful to allow configuring it directly from the factory, vs. supplying a custom `RequestConfig` and `HttpClient` to the factory.


simonbasle:
thanks for the contribution @cmagnuson. Looks good to me, I will polish the javadoc/formatting and add some tests before merging, fixing the checkstyle error along the way

simonbasle (approved):

PR #33566, opened 2024-09-19
Polish

izeye:
This PR polishes a bit.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
@@ -59,7 +59,7 @@ class JdkClientHttpRequest extends AbstractClientHttpRequest {
 	private final HttpRequest.Builder builder;
 
 
-	public JdkClientHttpRequest(HttpMethod httpMethod, URI uri, DataBufferFactory bufferFactory,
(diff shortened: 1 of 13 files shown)


nastae (approved):

spring-web/src/main/java/org/springframework/http/client/reactive/JdkClientHttpRequest.java
```diff
@@ -59,7 +59,7 @@ class JdkClientHttpRequest extends AbstractClientHttpRequest {
 	private final HttpRequest.Builder builder;
 
 
-	public JdkClientHttpRequest(HttpMethod httpMethod, URI uri, DataBufferFactory bufferFactory,
+	JdkClientHttpRequest(HttpMethod httpMethod, URI uri, DataBufferFactory bufferFactory,
```
sbrannen, line 62: Although I fully understand the desire to remove `public` here, historically the code base often makes constructors `public` even if the enclosing type is not `public`.

Look at the other `JdkClientHttpRequest` implementation as well as neighboring classes for examples.

In light of that, I removed this change when merging the PR.

sbrannen:
This has been merged into `main` in 8931b75b958144b1496da1b163553acd03e11ebd and partially revised in a related commit (8931b75b958144b1496da1b163553acd03e11ebd).

Thanks

Issue #33434, opened 2024-08-27
Updates to resource handling for functional endpoints


rstoyanchev:
The built-in handling of resources in Spring MVC and WebFlux gets updated occasionally, but the functional programming model hasn't stayed up-to-date. Those should be functionally equivalent where it makes sense.

drdpov:
Hello @rstoyanchev, hope you are doing well. I've come across this PR and noticed, that there is one minor issue. I've created a PR, which should resolve it, could you please take a look? https://github.com/spring-projects/spring-framework/pull/33568

lucky8987:
@drdpov Hello, as we currently have no plans to upgrade to version 6.1. x, this issue has triggered a high-risk vulnerability: https://spring.io/security/cve-2024-38816 Can you fix those issues specifically for version 5.3.39?

bclozel:
@lucky8987 all CVE fixes are already backported to 5.3.x, see [our announcement blog post](https://spring.io/blog/2024/09/12/spring-framework-releases-fixes-for-cve-2024-38816) and [the advisory you've linked to](https://spring.io/security/cve-2024-38816). [5.3.x is not OSS supported anymore](https://spring.io/projects/spring-framework#support) so you'll have to upgrade to a newer generation or consider commercial support.

lucky8987:
> @lucky8987 all CVE fixes are already backported to 5.3.x, see [our announcement blog post](https://spring.io/blog/2024/09/12/spring-framework-releases-fixes-for-cve-2024-38816) and [the advisory you've linked to](https://spring.io/security/cve-2024-38816). [5.3.x is not OSS supported anymore](https://spring.io/projects/spring-framework#support) so you'll have to upgrade to a newer generation or consider commercial support.

I understand, thanks !

luckymanbuddha:
@bclozel Hello, I would like to ask if [CVE-2024-38819 ](https://spring.io/security/cve-2024-38819) is the same as [CVE-2024-38816](https://spring.io/security/cve-2024-38816)  can  use Tomcat or Jetty as the web server to reject such malicious requests? Thank you.

nao-nomura:
Hello. It is stated at https://spring.io/security/cve-2024-38816 that there is no impact when run on Tomcat or Jetty. However, in the above discussion, there was an opinion that it is not protected.

[...]

PR #33568, opened 2024-09-19
Use encoded resource path  instead of input path validation in spring-webflux

drdpov:
This pull request adds minor fix, to the validation introduced here: https://github.com/spring-projects/spring-framework/commit/d86bf8b2056429edf5494456cffcb2b243331c49 when resolving issue https://github.com/spring-projects/spring-framework/issues/33434
It seems that in the initial fix for spring-webmvc was created correctly, but for spring-webflux validation method was used incorrectly


rstoyanchev (approved):
Thanks for catching that. Thankfully quite similar, but we should use the one that's intended and is more targeted.

rstoyanchev:
Those method are now re-used, see #33574.

PR #32097, opened 2024-01-24
Implement Eclipse Jetty core HTTP handler adapter

gregw:
This provides an implementation of a HTTP Handler Adaptor that is coded directly to the Eclipse Jetty core API, bypassing any servlet implementation.

Fixes #32035

PR #33569, opened 2024-09-19
Fix Javadoc `@since` tags in `JettyCoreRequestUpgradeStrategy`

izeye:
This PR fixes Javadoc `@since` tags in the `JettyCoreRequestUpgradeStrategy`.

See gh-32097


sbrannen:
Good catch!

Thanks

PR #33571, opened 2024-09-19
Polish

nastae:
Use constant for empty string like as ""


sbrannen:
Hi @nastae,

Regarding the empty string, we have over 3700 occurrences within the code base, and we don't see the need to define that as a constant simply because it's such a common value and the use of a constant would not provide much value in terms of readability or maintenance.

Regards,

Sam

PR #33572, opened 2024-09-20
Simplify hasText and containsElement methods for better clarity

dukbong:
- In the containsElement method, I replaced the null check with the isEmpty method to enhance readability.
- In the hasText method, I substituted the length check with isEmpty to reduce code duplication.


sbrannen:
Hi @dukbong,

Thanks for the proposal; however, we are happy with the current implementation of these internal utility methods.

Regards,

Sam

PR #33577, opened 2024-09-21
Consistently use `CharSequence.isEmpty()` for emptiness checks

ngocnhan-tran1996:


spring-core/src/main/java/org/springframework/asm/TypePath.java
```diff
@@ -117,7 +117,7 @@ public int getStepArgument(final int index) {
    * @return the corresponding TypePath object, or {@literal null} if the path is empty.
    */
   public static TypePath fromString(final String typePath) {
-    if (typePath == null || typePath.length() == 0) {
+    if (typePath == null || typePath.isEmpty()) {
```
sbrannen, line 120: Please refrain from modifying classes under `org.springframework.asm`, `org.springframework.cglib`, and `org.springframework.objenesis`. Those include repackaged forks of the third-party libraries ASM, CGLIB, and Objenesis. Any refactoring to those classes should take place upstream in the originating repository. The Spring Framework will then pick up the changes when syncing with official updates of the forked third-party libraries.

----

Note, however, that we can revert this change manually if we decide to merge this PR.

ngocnhan-tran1996: I think I should revert files in `org.springframework.asm` and `org.springframework.objenesis` to prevent any unexpected exception related third-party

sbrannen:
This has been merged into `main` in c85050eb43c37f084ae5bd31a708d0e2825ea43c.

Thanks

p.s. please note that I updated the copyright dates accordingly.

PR #33578, opened 2024-09-23
Store non-empty bean names in `ConcurrentMapCacheFactoryBean`

dukbong:
Typically, there are no bean names composed solely of whitespace, so the parameter for this method should be a meaningful value. To ensure the accuracy of that value, it seems more appropriate to use hasText rather than hasLength.


sbrannen:
Hi @dukbong,

> Typically, there are no bean names composed solely of whitespace, so the parameter for this method should be a meaningful value.

Indeed, it is unlikely (or at least undesirable) to have a bean name composed solely of whitespace; however, this check is not meant to guard against that scenario.

Rather, we only intend to ensure the bean/cache name is non-null and has length (because `""` is already the default for the cache name).

In light of that, we will leave this `hasLength` check as-is.

Thanks anyway for the PR.

Issue #33355, opened 2024-08-09
Add a hearbeat executor to SSE


reda-alaoui:
It looks like the general recommendation about SSE is to send messages at fixed interval to clients to probe the connection health (see https://github.com/spring-projects/spring-framework/issues/18523#issuecomment-470141717).

Could Spring add this feature?

rstoyanchev:
Superseded by #34878.

PR #33582, opened 2024-09-23
Add heartbeat functionality to SseEmitter

Torres-09:
This PR addresses issue #33355 by adding heartbeat (ping) functionality to the SseEmitter class.

Heartbeat Functionality
Added a new constructor to SseEmitter that accepts a heartbeatInterval parameter:

Heartbeat messages are sent as SSE comments (:heartbeat) at the specified interval.

Heartbeat messages automatically stop when the emitter is completed, times out, or encounters an error.

Testing
Added test cases to verify that heartbeat messages are sent at the correct intervals.

All tests have passed successfully.

Issue: #33355

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java
@@ -21,6 +21,10 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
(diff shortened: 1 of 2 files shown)


mdeinum:
I can see a risk here if many `SseEmitter`s are created, each will create a `Thread` and take up memory and cpu cycles which can lead to memory and performance issues.

Torres-09:
@mdeinum, thank you for your review.

When I added the new heartbeat feature, my primary considerations were: (1) not to disrupt any existing functionality, and (2) not to introduce performance issues.

Unfortunately, it appears that the current changes do not fully meet the second condition. I overlooked the fact that creating an executor for each new instance could lead to performance problems due to the overhead of additional threads.

To address this, I am considering new approaches, such as (1) injecting a shared static resource, or (2) utilizing a TaskScheduler.

bclozel:
I'm declining this in favour of #33355 due to the concerns expressed here. Let's find another way to tackle this.

PR #33597, opened 2024-09-25
Remove unused generic parameter declaration

ngocnhan-tran1996:
I think `<T>` in method is unused


jhoeller:
Good catch! Thanks for the pull request.

PR #33603, opened 2024-09-28
Add Hibernate NonUniqueException ResultSize Processing

le2sky:
```java
    private void validateAlreadyIssued(Long memberId, Long couponId) {
        try {
            // return multiple result when transaction confilcted
            memberCouponRepository.findByMemberIdAndCouponId(memberId, couponId);
        } catch (IncorrectResultSizeDataAccessException e) {
            log.info("expected size = {}, actualSize = {}", e.getExpectedSize(), e.getActualSize());
            log.info("mesage = {}", e.getMessage());

            throw new IllegalStateException("transaction conflict");
        }
    }
```

```
2024-09-28T14:18:12.153+09:00  INFO 6651 --- [ool-2-thread-28] coupon.application.MemberCouponService   : expected size = 1, actualSize = -1
2024-09-28T14:18:12.153+09:00  INFO 6651 --- [ool-2-thread-28] coupon.application.MemberCouponService   : mesage = Query did not return a unique result: 8 results were returned
```

when I checking the logs, actualSize is -1. It seems strange for me. because the message generated by hibernate (`hibernate.NonUniqueResultException`) contains the actual result size.

Since `hibernate.NonUniqueResultException` does not provide the result count, I changed `HibernateJpaDialect`
that actualSize extracted from message generated by hibernate.

If there's anything I've overlooked, I'd appreciate your feedback. 🙇🏻‍♂️

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR.

> when I checking the logs, actualSize is -1. It seems strange for me. because the message generated by hibernate (hibernate.NonUniqueResultException) contains the actual result size.

It's `-1` as we couldn't know for sure what it was. You can see it's used by `EmptyResultDataAccessException` when we know what the actual size is.

Checking for the presence of any number in the exception message is too brittle for us to consider it. You may want to move this code in your own arrangement if you really care about the actual size.

PR #33607, opened 2024-09-29
Fix incorrect access to field `beanName` in `RuntimeBeanNameReference#equals()`

Alioth4J:
The original code used `that.beanName` to access the private field `beanName`, which is not allowed. Changed to use getter method.


sbrannen:
Hi @Alioth4J,

Unfortunately, your analysis is incorrect: in Java it is permissible to access a private field of an instance of the same type within methods of that type.

Regards,

Sam

Issue #33542, opened 2024-09-15
Add javadoc to org.springframework.web.util.UrlParser to indicate that it should only be used with modern browsers, not anything else


joakime:
**Affects:** Any with UrlParser

The recently added `org.springframework.web.util.UrlParser` is not spec compliant outside of the limited scope of modern browsers.

The living URL document at whatwg is incompatible with the [IETF URI spec](https://datatracker.ietf.org/doc/html/rfc3986), Java itself, the Servlet spec, and various other non-browser use cases.

Users that want to use the new `UrlParser` should not be using it for non-browser use cases (eg: REST clients).

bclozel:
Thanks for this feedback @joakime 

Could you elaborate on the limitations of using this parser for REST clients? What kind of valid URLs wouldn't parsed correctly with this parser in the context of REST clients?

Thanks!

joakime:
Take for example a URL with no heir-part, or no authority.

`http:///path.png`

The UrlParser sees that as a host with `path.png`

Here's what java.net.URI does ...
```
$ jshell
|  Welcome to JShell -- Version 17.0.11
|  For an introduction type: /help intro

jshell> var uu = new URI("http:///path.png")
uu ==> http:///path.png

jshell> uu.getHost()
$2 ==> null

jshell> uu.getPath()
$3 ==> "/path.png"
```

Here's what java.net.URL does ...

```
$ jshell
|  Welcome to JShell -- Version 17.0.11
|  For an introduction type: /help intro

jshell> var uu = new URL("http:///path.png")
uu ==> http:/path.png

jshell> uu.getHost()
$2 ==> ""

jshell> uu.getPath()
$3 ==> "/path.png"
```

Those built-in parsers, along with the existing URL / URI parsers in various Servlet libraries follow the spec in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986), which parses that as a URL with no authority, just a path.

WhatWG is good, if you want to follow browser behaviors, but not good outside of that limited scope.

joakime:
There are many decisions in WhatWG that are designed to "clean up" bad behaviors from users, typos and whatnot (eg: eliminate duplicates, normalize away extra slashes, eliminate whitespace, etc)

Those are great choices for a browser dealing with HTML and Javascript, but not appropriate outside of a browser.

rstoyanchev:
[...]

PR #33608, opened 2024-09-30
Parse host from authority in UriComponentsBuilder.uri when unsuccessful

yhkee0404:
According to [rain.drop](https://tech.kakaopay.com/post/url-is-strange/#%EC%82%AC%EB%9D%BC%EC%A7%84-%EC%A0%95%EB%B3%B4%EB%8A%94-%EC%97%AC%EA%B8%B0%EC%97%90) in kakaopay, it appears that `UriComponentsBuilder#fromUri` has a bug where it fails to recognize hostnames containing underscores (_) due to a limitation of `java.net.URI`. In contrast, `UriComponentsBuilder#fromUriString` successfully parses these hostnames using `URI_PATTERN`.

For example, `UriComponentsBuilder.fromUri(URI("kakaopay://payweb_tab"))` results in a null `host` but `payweb_tab` becomes its `authority`, unlike `UriComponentsBuilder.fromUriString("kakaopay://payweb_tab")` says its `host` and `authority` are both `payweb_tab`.

This issue arises because [java.net.URI](https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/net/URI.html) adheres to RFC 2396 and RFC 2732 but does not conform to RFC 3986, which allows underscores in hostnames as per the `URI_PATTERN`. This inconsistency can lead to problems when parsing such URIs with the standard URI class, potentially resulting in the `getHost()` method returning null.

https://github.com/openjdk/jdk/blob/ae4d2f15901bf02efceaac26ee4aa3ae666bf467/src/java.base/share/classes/java/net/URI.java#L3513-L3520

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -482,6 +482,11 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 			}
 			if (uri.getHost() != null) {
 				this.host = uri.getHost();
+			} else if (uri.getRawAuthority() != null) {
+				Matcher matcher = URI_PATTERN.matcher(uri.getRawAuthority());
+				if (matcher.matches()) {
+					this.host = matcher.group(6);
+				}
 			}
 			if (uri.getPort() != -1) {
 				this.port = String.valueOf(uri.getPort());


bclozel:
Thanks for the proposal, but I don't think we should add a special case for this; `java.net.URI` has a specific behavior and we should not change it because it would create inconsistencies. `fromUriString` uses a different parsing algorithm. We will refine our documentation and stance on this matter in #33542

PR #33614, opened 2024-09-30
Reparse URI when host is missing

### LINKED ISSUE #15852: JwtDecoderProviderConfigurationUtils incorrectly handles issuer URI
**Describe the bug**
`JwtDecoderProviderConfigurationUtils` use `UriComponentsBuilder` to modify a `URI` in `oidc`, `oidcRfc8414` and `oauth`.
Due to a known issue in UriComponents builder (https://github.com/spring-projects/spring-framework/issues/27774), this produces invalid URIs in the cases where the URI contains an `authority`, but no `host`.
These still come up in practice though, e.g. see the following partial backtrace:

```
Caused by: java.lang.IllegalArgumentException: Unable to resolve the Configuration with the provided Issuer of "http://elated_sutherland:8080/auth/realms/my-realm"
        at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getConfiguration(JwtDecoderProviderConfigurationUtils.java:178)
        at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getConfigurationForIssuerLocation(JwtDecoderProviderConfigurationUtils.java:90)
        at org.springframework.security.oauth2.jwt.NimbusJwtDecoder.lambda$withIssuerLocation$2(NimbusJwtDecoder.java:226)
        at org.springframework.security.oauth2.jwt.NimbusJwtDecoder$JwkSetUriJwtDecoderBuilder.processor(NimbusJwtDecoder.java:389)
        at org.springframework.security.oauth2.jwt.NimbusJwtDecoder$JwkSetUriJwtDecoderBuilder.build(NimbusJwtDecoder.java:405)
        at org.springframework.boot.autoconfigure.security.oauth2.resource.servlet.OAuth2ResourceServerJwtConfiguration$JwtDecoderConfiguration.lambda$jwtDecoderByIssuerUri$3(OAuth2ResourceServerJwtConfiguration.java:159)
        at org.springframework.security.oauth2.jwt.SupplierJwtDecoder.lambda$new$0(SupplierJwtDecoder.java:36)
        ... 88 common frames omitted
Caused by: org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http:/auth/realms/my-realm/.well-known/openid-configuration": Failed to select a proxy
        at org.springframework.web.client.RestTemplate.createResourceAccessException(RestTemplate.java:915)
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:895)
        at org.springframework.web.client.RestTemplate.exchange(RestTemplate.java:740)
        at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getConfiguration(JwtDecoderProviderConfigurationUtils.java:167)
        ... 94 common frames omitted
```

**To Reproduce**
```java
import org.springframework.web.util.UriComponentsBuilder;

import java.net.URI;
import java.util.Collections;

String rawUrl = "http://elated_sutherland:8080/auth/realms/my-realm";
URI uri = URI.create(rawUrl);
System.out.println(uri.getAuthority());
System.out.println(uri.getHost());
URI newUrl = UriComponentsBuilder.fromUri(uri)
  .replacePath(uri.getPath() + "/.well-known/openid-configuration")
  .build(Collections.emptyMap());
System.out.println(newUrl);
```

shows what happens. It outputs:
```
elated_sutherland:8080
null
http:/auth/realms/my-realm/.well-known/openid-configuration
```

**Expected behavior**
The correct oidc URI should be generated: `http://elated_sutherland:8080/auth/realms/my-realm/.well-known/openid-configuration`.

### LINKED ISSUE #27774: UriComponentsBuilder silently drops some Registry-based Naming Authorities
In spring-web 5.3.13, [UriComponentsBuilder.fromUri](https://github.com/spring-projects/spring-framework/blob/v5.3.13/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java#L207) / [UriComponentsBuilder.uri](https://github.com/spring-projects/spring-framework/blob/v5.3.13/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java#L554) silently drops some [RFC 2396 3.2.1 Registry-based Naming Authorities ](https://datatracker.ietf.org/doc/html/rfc2396#section-3.2.1) even when they are accepted as [RFC 3986 3.2 authorities](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2).

```
    // I tested java 17.0.1-zulu and 8.0.312-zulu - with the same results.

    @Test
    public void testUriComponentBuilder() throws URISyntaxException {
        // Generic scheme to avoid any assumptions around HTTP
        String scheme = "scheme";

        // Matches
        // - RFC 2396 3.2.1. "reg_name" Registry-based Naming Authority
        // - RFC 3986 3.2.2 "host" Host
        // NOT an RFC 2396 3.2.2. "server" Server-based Naming Authority
        String authority = "rfc_2396_authority";


        String uriString = scheme + ":" + "//" + authority ;

        URI testUri = new URI(uriString);

        UriComponents uriComponentsFromUriString = UriComponentsBuilder.fromUriString(uriString).build(true);
        UriComponents uriComponentsFromUri = UriComponentsBuilder.fromUri(testUri).build(true);

        // Demonstrate that the URI is well formed according to RFC 2396 and parses as expected
        assertAll("testUriFromString should have a RFC 2396 Registry-based Naming Authority (3.2.1.)",
                () -> assertEquals(authority, testUriFromString.getAuthority()),
                () -> assertNull(testUriFromString.getUserInfo()),
                () -> assertNull(testUriFromString.getHost()),
                () -> assertEquals(-1, testUriFromString.getPort())
        );

        // Demonstrate that the URI is parseable under RFC 3986 from string form
        assertAll("uriComponentsFromUriString should be populated ",
                () -> assertEquals(authority, uriComponentsFromUriString.getHost())
        );

        // Demonstrate that the parsed URI is equivalent (according to java.net.URI)
        assertEquals(testUriFromString, uriComponentsFromUriString.toUri());

        // Unexpectedly fails as getHost() return a null even though testUri has an RFC 3986 host
        assertAll("uriComponentsFromUri should have populated host ",
                () -> assertEquals(authority, uriComponentsFromUri.getHost())
        );

    }
```

When `java.net.URI` encounters an [RFC 2396 3.2.1](https://datatracker.ietf.org/doc/html/rfc2396#section-3.2.1) authority (identified as any correct authority not parseable under [RFC 2396 3.2.2](https://datatracker.ietf.org/doc/html/rfc2396#section-3.2.2)), it returns null values for the server-based naming authority components, causing `UriComponentBuilder` to assume they are unset and silently dropping the data. However, this data is available via `URI.getAuthority()`.

I would expect that `UriComponentBuilder` would attempt to reparse the authority from `Uri.getAuthority` under the RFC 3986 3.2 specification.

The fact that these classes are written against different versions of the spec is obviously a complicating factor. The most straightforward option is to reparse the URI from scratch under the RFC 3986 spec, though I think that it would be preferable to only parse the authority component to limit the blast radius of this change. I am happy to contribute a PR, but would appreciate thumbs up on the approach before investing the time as this would substantially change existing behavior if `UriComponentBuilder.uri` is being used after the authority components have been set via another method.

bodograumann:
When a hostname contains an underscore, it is valid in RFC 3986, but only part of a registry-based authority under RFC 2396. Due to the fact, that `URI` only supports RFC 2396, we cannot simply use `getHost()` in these cases. Instead we need to parse the URI ourselves.

Only parsing the authority part seems too much effort, given the custom `UrlParser` we are using.

Closes https://github.com/spring-projects/spring-framework/issues/27774
Closes https://github.com/spring-projects/spring-security/issues/15852

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -471,6 +471,11 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 	 */
 	public UriComponentsBuilder uri(URI uri) {
 		Assert.notNull(uri, "URI must not be null");
+		if (uri.getHost() == null && uri.getAuthority() != null) {
+			// see gh-27774
+			uriComponents(UriComponentsBuilder.fromUriString(uri.toString()).build());
+			return this;
+		}
 		this.scheme = uri.getScheme();
 		if (uri.isOpaque()) {
 			this.ssp = uri.getRawSchemeSpecificPart();
--- a/spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
+++ b/spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
@@ -160,6 +160,17 @@ class UriComponentsBuilderTests {
 		assertThat(fromUriString).isEqualTo(fromUri);
 	}
 
+	@Test  // see gh-27774
+	void fromUriRegistryBasedAuthority() {
+		URI uri = URI.create("http://elated_sutherland:8080/auth/realms/my-realm");
+		UriComponents result = UriComponentsBuilder.fromUri(uri).build();
+
+		assertThat(result.getScheme()).isEqualTo("http");
+		assertThat(result.getHost()).isEqualTo("elated_sutherland");
+		assertThat(result.getPort()).isEqualTo(8080);
+		assertThat(result.getPath()).isEqualTo("/auth/realms/my-realm");
+	}
+
 	@Test
 	void fromUriString() {
 		UriComponents result = UriComponentsBuilder.fromUriString("https://www.ietf.org/rfc/rfc3986.txt").build();


bclozel:
I think this is duplicating #33608, but I missed the comments in #27774.
We will discuss this as a team, but my comment on #33608 my still stand.

bodograumann:
I was going based on this comment: https://github.com/spring-projects/spring-framework/issues/27774#issuecomment-1004757583, but maybe that is outdated with #33608 as you say.
My main goal is to get https://github.com/spring-projects/spring-security/issues/15852 fixed, so simply not using URI with UriComponentsBuilder would also achieve that goal as suggested here: https://github.com/spring-projects/spring-security/pull/15853
Unfortunately I wasn't yet able to provide proper tests there...

rstoyanchev:
I agree with Brian this could lead to surprises, and we should not quietly re-parse from within `fromUri(URI)`. Spring Security can do that in fallback mode as I commented under https://github.com/spring-projects/spring-security/pull/15853.

PR #33625, opened 2024-10-01
Replace hardcoded `"null"` with `NULL_STRING` constant in `ObjectUtils.nullSafeConciseToString()`

nomoreFt:
#### Description:
This pull request replaces the hardcoded "null" string in the nullSafeConciseToString method with the existing NULL_STRING constant to improve maintainability and readability.

#### Changes:
Replaced occurrences of the hardcoded "null" string with the already declared NULL_STRING constant.


sbrannen:
Good catch!

That was an oversight on my part.

Thanks

PR #33626, opened 2024-10-01
Improve ObjectUtils#nullSafeEquals method for readability and performance

nomoreFt:
### Description:
 This pull request simplifies the nullSafeEquals method to improve readability and performance.

### Changes:
Reorganized if conditions for clarity and reduced redundancy.
Optimized the array comparison logic for early termination when possible.
Removed unnecessary equals checks, making the flow more efficient.


bclozel:
Thanks for the proposal, but I don't think this will improve performance as advertized, quite the opposite. This skips the equals check in favor of array class checks.

PR #33631, opened 2024-10-01
Polish code for Optional, `String#isEmpty`

ngocnhan-tran1996:


bclozel:
Thanks for the proposal, but I don't think this makes the code more readable.

PR #33650, opened 2024-10-04
Polish `CorsConfiguration`

khyojun:
## Changes & Considerations

- The * symbol was already handled as a constant, but the / symbol was not. Handling this as a constant will improve readability.
- Additionally, the * symbol was replaced in other parts of the code, but it was not replaced in OriginPattern, which caused some confusion. It seems necessary to replace it there as well.


sdeleuze:
Thanks for your contribution, I only merged the replacement of `"*"` by the existing public static field `ALL`, since the other change does not bring much from my POV. Notice we do not add `@author` tag for such minor change.

khyojun:
@sdeleuze "As a student, this is my first time contributing to open source, so I'm a bit unsure! In this case, does this mean I've also contributed to the Spring code? I'm asking because it's my first experience, and I'm so excited!"

sdeleuze:
Yes, you did.

PR #33653, opened 2024-10-05
Refactor Locale logic and use Optional for null checks

SungbinYang:
- Extracted repeated Locale resolution logic into a private method to reduce duplication.
- Replaced traditional null checks with Optional to improve readability and make null handling more explicit.


sdeleuze:
If think we are prefer to keep the current `null` check logic as it is rather than paying the cost of wrapping/unwrapping with `Optional`, the logic in the extracted method is very simple and we tend to prefer to handle this kind of polish within the team, so I decline this PR.

PR #33659, opened 2024-10-06
Remove Javadoc `@since` tag from `MapAccessor` default constructor

izeye:
## Overview

This PR removes Javadoc `@since` tag from `MapAccessor()` as it seems to have been added accidentally.

## Related Issues

- #33222


sbrannen:
Hi @izeye,

Technically speaking, that explicit default constructor was introduced in 6.2; however, you're right: it always had an _implicit_ default constructor. So, the `@since` tag can indeed be misinterpreted, and I've merged your PR into `main` to avoid confusion.

Thanks

PR #33671, opened 2024-10-09
Fix handling of headers in reactive HTTP requests to ensure case-insensivity

### LINKED ISSUE #33666: `DefaultServerHttpRequestBuilder` can create case-sensitive `Headers` instances
In 6.2.0-M5, the `DefaultServerHttpRequestBuilder` was modified to ensure copying read-only headers would result in a mutable `HttpHeaders` instance. Unfortunately, the way this is implemented has a side effect: the case-insensitive nature of the headers is lost.

See https://github.com/spring-projects/spring-framework/pull/32097/files#r1791264218

encircled:
Fixes gh-33666


simonbasle:
Thanks for the PR, but I'm already on it (as indicated by the issue being assigned to me).

PR #33675, opened 2024-10-09
Refactor InvocationsRecorderClassTransformer

jukekxm:
Hello, This is why I opened PR.

1. if statement in the `isTransformationCandidate` operates normally without the `else` statement.
2.  `transform` is not throwing an `IllegalClassFormatException.`
3. I thought that using logger would be more performant than printStackTrace.

I would appreciate any feedback on this PR. Thank You.


bclozel:
Thanks but I don't think those cosmetic changes help here.

PR #33679, opened 2024-10-10
Refactor duplicate code improvements

Shinjinsik123:
- I separated them into common codes. This is good for code consistency.

this is my first pull-request.
I will be active in the future :)

with @J-unStiN


sdeleuze:
I think using directly the field state to check for the status is ok in that case. It is a little bit less readable, bit a little bit more optimized, so I will keep what we have and decline this PR.

PR #33681, opened 2024-10-10
Polishing

ngocnhan-tran1996:
This PR includes:
- Reduce `else if` condition. I think this make code more readable
- Polish diamond operator usage
- Polish `Map#isEmpty` usage


sdeleuze:
Merged except `FastByteArrayOutputStream` changes since it does not achieve consistency class-wide. Notice **I do not** ask a follow-up PR for `FastByteArrayOutputStream`.

I appreciate the contributions but we need to be mindful of reaching a certain level of added value, otherwise we could spend too much energy reviewing Spring Framework codebase polishes. So for upcoming PRs, I will ask more clear added value for refinements, and otherwise will reject the related PRs. Thanks for your undertanding.

PR #33698, opened 2024-10-14
Polish `ServletWebRequest`

smartandhandsome:
This PR includes the following changes to improve code readability:

1. **Code readability**:
   - Refactored an `if` statement to enhance readability.
   - Removed an unnecessary `else` block
   
2. **Javadoc improvements**:
   - Replaced `@code` with `@link` for the `checkNotModified(String)` reference to create clickable links in the Javadoc.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/context/request/WebRequest.java
+++ b/spring-web/src/main/java/org/springframework/web/context/request/WebRequest.java
@@ -223,7 +223,7 @@ public interface WebRequest extends RequestAttributes {
 	 * also with conditional POST/PUT/DELETE requests.
 	 * <p><strong>Note:</strong> The HTTP specification recommends
 	 * setting both ETag and Last-Modified values, but you can also
-	 * use {@code #checkNotModified(String)} or
+	 * use {@link #checkNotModified(String)} or
 	 * {@link #checkNotModified(long)}.
 	 * @param etag the entity tag that the application determined
 	 * for the underlying resource. This parameter will be padded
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/web/context/request/WebRequest.java
```diff
@@ -223,7 +223,7 @@ public interface WebRequest extends RequestAttributes {
 	 * also with conditional POST/PUT/DELETE requests.
 	 * <p><strong>Note:</strong> The HTTP specification recommends
 	 * setting both ETag and Last-Modified values, but you can also
-	 * use {@code #checkNotModified(String)} or
+	 * use {@link #checkNotModified(String)} or
```
sdeleuze, line 226: The use of `@code` is done on purpose, as a `@link` to the current method generates a warning (in IntelliJ IDEA at least), so I will keep this part as it is.

smartandhandsome: Thank you for your feedback!!

But,,, I think this is an overloaded method and not the same one, so there’s no warning in IntelliJ.

sdeleuze: Good catch, I checked the wrong one, I will bring back this change via an additional commit.

smartandhandsome: Thank you!!!

PR #33709, opened 2024-10-15
Update scheduling.adoc

boulce:
Change `executor` to 'executor' like "The 'scheduler' Element" or "The 'scheduled-tasks' Element" for unity.

### DIFF
--- a/framework-docs/modules/ROOT/pages/integration/scheduling.adoc
+++ b/framework-docs/modules/ROOT/pages/integration/scheduling.adoc
@@ -633,7 +633,7 @@ scheduled with a trigger.
 
 
 [[scheduling-task-namespace-scheduler]]
-=== The 'scheduler' Element
+=== The `scheduler` Element
 
 The following element creates a `ThreadPoolTaskScheduler` instance with the
 specified thread pool size:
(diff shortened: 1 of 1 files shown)


sdeleuze:
I am not sure that would be consistent, we are still using backticks for `task` for example. It looks like backticks are more often used and are rendering nicely, with a meaning that we are quoting something from the code which is the case here, so maybe update the PR the other way around consistently for the section of this file?

boulce:
> I am not sure that would be consistent, we are still using backticks for `task` for example. It looks like backticks are more often used and are rendering nicely, with a meaning that we are quoting something from the code which is the case here, so maybe update the PR the other way around consistently for the section of this file?

Then Is okay If I change "The 'scheduler' Element" to "The `scheduler` Element" and "The 'scheduled-tasks' Element" to "The `scheduled-tasks` Element".
Therefore, you mean,  "The 'scheduler' Element" and  "The 'scheduler' Element" is wrong?

boulce:
> I am not sure that would be consistent, we are still using backticks for `task` for example. It looks like backticks are more often used and are rendering nicely, with a meaning that we are quoting something from the code which is the case here, so maybe update the PR the other way around consistently for the section of this file?

I changed as you mentioned. Please check.

PR #33714, opened 2024-10-16
Fix the issue in MappingJackson2MessageConverter where deserializatio…

cooperlyt:
Fix the issue in MappingJackson2MessageConverter where deserialization returns null values for POJOs that do not have ```@JsonView``` annotations

When a POJO does not have the JsonView annotation set, and Jackson's DEFAULT_VIEW_INCLUSION is false (the default value in Spring framework #16793), deserialization fails and returns an object with all properties empty.

I added a test in ```MappingJackson2MessageConverterTests```  that will fail before I fix it.
```java
	@Test
	public void fromMessageToMessageWithPojoClass(){
		// #16793 https://github.com/spring-projects/spring-framework/issues/16793
		//ObjectMapper objectMapper = JsonMapper.builder().configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false).build();
		MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
		String payload = "{\"string\":\"foo\"}";
		Message<?> message = MessageBuilder.withPayload(payload.getBytes(StandardCharsets.UTF_8)).build();
		Object actual = converter.fromMessage(message, MyBean.class, MyBean.class);
		assertThat(actual).isInstanceOf(MyBean.class);
		assertThat(((MyBean) actual).getString()).isEqualTo("foo");
	}
```

PS:   Discussion with jackson-databind  [issues](https://github.com/FasterXML/jackson-databind/issues/4748).

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/converter/MappingJackson2MessageConverter.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/converter/MappingJackson2MessageConverter.java
@@ -316,7 +316,8 @@ public class MappingJackson2MessageConverter extends AbstractMessageConverter {
 		else if (conversionHint instanceof JsonView jsonView) {
 			return extractViewClass(jsonView, conversionHint);
 		}
-		else if (conversionHint instanceof Class<?> clazz) {
+		else if (conversionHint instanceof Class<?> clazz &&
+				objectMapper.getSerializationConfig().isEnabled(MapperFeature.DEFAULT_VIEW_INCLUSION)) {
 			return clazz;
 		}
 
--- a/spring-messaging/src/test/java/org/springframework/messaging/converter/MappingJackson2MessageConverterTests.java
+++ b/spring-messaging/src/test/java/org/springframework/messaging/converter/MappingJackson2MessageConverterTests.java
@@ -25,6 +25,9 @@ import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.core.MethodParameter;
@@ -160,6 +163,18 @@ class MappingJackson2MessageConverterTests {
 	}
 
 	@Test
+	public void fromMessageToMessageWithPojoClass(){
+		// #16793 https://github.com/spring-projects/spring-framework/issues/16793
+		//ObjectMapper objectMapper = JsonMapper.builder().configure(MapperFeature.DEFAULT_VIEW_INCLUSION, false).build();
+		MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
+		String payload = "{\"string\":\"foo\"}";
+		Message<?> message = MessageBuilder.withPayload(payload.getBytes(StandardCharsets.UTF_8)).build();
+		Object actual = converter.fromMessage(message, MyBean.class, MyBean.class);
+		assertThat(actual).isInstanceOf(MyBean.class);
+		assertThat(((MyBean) actual).getString()).isEqualTo("foo");
+	}
+
+	@Test
 	void toMessage() {
 		MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
 		MyBean payload = new MyBean();
@@ -233,7 +248,6 @@ class MappingJackson2MessageConverterTests {
 		assertThat(back.getWithView2()).isEqualTo("with");
 		assertThat(back.getWithoutView()).isNull();
 	}
-
 
 
 	@JsonView(MyJacksonView1.class)


sdeleuze:
Could you please share the original use case that ended-up passing this `MyBean.class` hint to the converter?

cooperlyt:
I Use Spring Streams connect to RabbitMQ,  send a Json message,  that happened

sdeleuze:
I would need a reproducer since I am not sure the use case is valid, it could be a Spring Cloud Stream bug. Please attach a zip archive or comment with a link to a repository that allow to reproduce this use case.

cooperlyt:
the case need a  RabbitMQ,  Can I provide a docker-compose.yml

sdeleuze:
Sure `docker-compose.yml` is perfectly fine, it will allow me to reproduce locally.

cooperlyt:
A test case [Project](https://github.com/cooperlyt/SpringStreamJsonMessageTest) 

just  run ```./gradlew test``` test will be fails.

add ```spring.jackson.mapper.default-view-inclusion = true``` test will be pass

RabbitMQ [docker-compose.yml](https://github.com/cooperlyt/SpringStreamJsonMessageTest/blob/master/docker-compose.yml)

sdeleuze:
Thanks for the reproducer, I will have a look.

sdeleuze:
Looks like https://github.com/spring-cloud/spring-cloud-function is using the `conversionHint` parameter as a way to pass the itemType as a `Class` [here](https://github.com/spring-cloud/spring-cloud-function/blob/e3b968d0ee41d78d7da240398729817723c75c16/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java#L1389) which is not supported by `MappingJackson2MessageConverter` or other `org.springframework.messaging.converter.MessageConverter` as far as I can tell.

@olegz Can you elaborate on this? I think I understand the intent, but that's not a supported use case so far unless I am mistaken.

olegz:
Consider the following Function
```
public Function<List<Product>, Integer> consumer() 
```
with the following payload as input
```
[{"name":"julien"},{"name":"ricky"},{"name":"bubbles"}]
```

If we don't provide/use `conversionHint` the List will come in as `List<Map>` instead of `List<Product>`

So, aside form passing a `targetClass` as `Class` to `convertFromInternal(Message<?> message, Class<?> targetClass, @Nullable Object conversionHint)` we need to pass `ParameterisedType` as hint and eventually invoke `ObjectMapper.mapper.readValue((String) json, constructType)` where  `constructType` is `ParameterisedType`, thus ensuring successful conversion

olegz:
And with regard to the original issue, i don't understand why @cooperlyt needs to register custom message converter - `CustomMessageMarshallingConverter`. Within SCF/Stream stack we provide `JsonMessageConverter` that is capable of converting complex types
Can you please elaborate?

@sdeleuze feel free to push this issue to s-c-stream/function if you determine it's on us.

sdeleuze:
@cooperlyt Can you please elaborate on what @olegz asked?

cooperlyt:
Sorry for the late reply.

Let me explain the reasons:

First, after Spring Boot 3.3.X, if I don't register a custom JSON converter, it won't be applicable to any other converters. For example, in my sample program, if I remove the registration of ```CustomMessageMarshallingConverter```, it will throw an error saying a binary array cannot be converted to an object. I'm not sure why JsonMessageConverter is not effective.

Second, I need to make some custom settings to the ObjectMapper, such as registering some custom modules.
@sdeleuze @olegz

cooperlyt:
I checked the code of JsonMessageConverter, and it seems that it does not have the ability to deserialize from byte[] to an object via JSON. Additionally, starting from Spring Boot 3.3.X, it appears that the original JSON conversion done in the Stream RabbitMQ module has been deprecated, with the output fixed as byte[] to be handled by the converter. Am I misunderstanding anything?
@olegz

olegz:
@sdeleuze I would love if such support was provided at the core and somewhat surprised that it has never came up especially in the context of Spring Messaging 
Wondering what @artembilan thinks of that as well as I suspect there is something to that extent in spring-integration as well, but i may be wrong.

olegz:
@cooperlyt I don't understand what you mean when you say `it does not have the ability to deserialize from JSON byte[] to an object` as it is exactly what it does and all our payloads in stream/function come as `byte[]` and there are applications and tests to validated that, so please clarify. . .
If you can provide a reproducible sample we can certainly take a look and see if any improvements are necessary.

Also, customisations to `ObjectMapper` . . . you can certainly do it via standard boot mechanisms

But as I said, a reproducible sample would help

cooperlyt:
In my sample [program](https://github.com/cooperlyt/SpringStreamJsonMessageTest), if I remove my custom CustomMessageMarshallingConverter, an exception occurs, and no deserialization converter is applied. Did I miss something?

olegz:
@cooperlyt Indeed there was an issue in spring-cloud-function 4.1.3 with the way we create `ObjectMapper` that is used by `JsonMessageConverter`, but it has been addressed, so please upgrade your dependencies to 
```
extra["springCloudVersion"] = "2023.0.4-SNAPSHOT"
```
. . . and you should be fine.
I just tested it with your project

olegz:
@cooperlyt we'll have a release on November 12

sdeleuze:
@cooperlyt Please test with the version shared by Oleg and confirm I can close this PR unmerged.

cooperlyt:
Thanks
I tested the new version, and it works well. Since the issue has been resolved, my PR is no longer needed. Please close it.

PR #33722, opened 2024-10-16
SSE: Make Handler and initialize method public to improve testing.

kkocel:
Motivation: When I have my own SseEmitter and populate it with events, then I would like to write a unit test with TestHandler to assert events being sent

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java
@@ -117,7 +117,7 @@ public class ResponseBodyEmitter {
 	}
 
 
-	synchronized void initialize(Handler handler) throws IOException {
+	public synchronized void initialize(Handler handler) throws IOException {
 		this.handler = handler;
 
 		try {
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks for raising this @kkocel !

What do you mean by "my own SseEmitter". Are you subclassing this type? Can you explain why?

kkocel:
@bclozel I don't subclass it; just create a new instance and initialize it with a custom `TestHandler` copied from SseEmitterTests.

Here is how I test:

```kotlin
val testHandler = TestHandler() // test handler copy & pasted from https://github.com/spring-projects/spring-framework/blob/main/spring-webmvc/src/test/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitterTests.java#L141
val emitter =
            SseEmitter().apply {
                initialize(testHandler)
            }

 MyClassThatEmitsEventsUsingEmitter(emmiter)
```

I need to put those tests in `org.springframework.web.servlet.mvc.method.annotation` package to make `initialize` and `Handler` visible.

bclozel:
Unless I'm missing something, you're testing `SseEmitter` itself - we don't intend to open up our API for this use case.

kkocel:
@bclozel Actually, no, I tested my class, and `SSeEmitter` is only a sink for generated events from my class. 

Do you know any other way to extract those events from `SSeEmitter`?

bclozel:
In that case, I think introducing a test utility for SseEmitter would be a better idea. I don't think we should make such methods public as it will confuse developers and make methods available that have no use besides testing. I've created #33746 for that.

PR #33737, opened 2024-10-17
Refactor DefaultRestClient Constructor for Improved Readability and Consistency

Torres-09:
### Summary
Refactored DefaultRestClient constructor and DefaultRestClientBuilder's build method.

### Changes
Improved readability by aligning the order of parameters and their initialization in the DefaultRestClient constructor. (The reason for changing the order is to match the field order of the class)
Ensured consistency by renaming the only parameter whose name didn’t match its corresponding field.

### Other
No functional changes.
All tests pass.


Torres-09:
Thank you for reviewing the proposal.

PR #33766, opened 2024-10-21
Fix incorrect regex rendering in MVC controller documentation

Torres-09:
### Summary
Fix incorrect regex in documentation

### Step
- [x] running ./gradlew antora
- [x] then opening the framework-docs/build/site/6.1-SNAPSHOT/web/webmvc/mvc-controller/ann-requestmapping.html file in your browser

image
image

### Other
All tests pass.


bclozel:
Thanks for your first contribution @Torres-09 , [this will be available shortly in our ref docs](https://docs.spring.io/spring-framework/reference/).

PR #33769, opened 2024-10-22
Refactor SourceFilteringListener use Assert.state for null check

kwondh5217:
### Description
This PR refactors the `SourceFilteringListener` class to replace the existing null check with `Assert.state`, in line with Spring's coding standards for handling null checks. The error message has been updated to follow the guideline: it starts with the identifier (in this case, "Delegate") and ends with "must not be null."

### Changes:
- Replaced manual null check with `Assert.state` in the `onApplicationEventInternal` method.
- Updated the exception message to: `"Delegate must not be null"` to improve clarity and ensure consistent error reporting across the codebase.

### Why this change?
Using `Assert.state` improves readability and ensures that error messages are consistent with Spring's coding style. This change also enhances the maintainability of the code by adhering to standard practices.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/event/SourceFilteringListener.java
+++ b/spring-context/src/main/java/org/springframework/context/event/SourceFilteringListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


snicoll:
> This PR refactors the SourceFilteringListener class to replace the existing null check with Assert.state, in line with Spring's coding standards for handling null checks. The error message has been updated to follow the guideline: it starts with the identifier (in this case, "Delegate") and ends with "must not be null."

Unfortunately, that's not on of those. What you're referring to is `Assert.notNull()` and it throws an `IllegalArgumentException`. The code there is fine as it is. 

We are a small team with limited time to review PRs. Please be considerate of that before submitting again.

PR #33770, opened 2024-10-22
Refactor ETag-related logic for improved readability

smartandhandsome:
- Refactored `isEligibleForEtag` to simplify method signature by removing unnecessary parameters.
- Refactored if statement for better readability.

I noticed that in the `isEligibleForEtag` method, it is validating the response status code for any successful range (200–299, `ShallowEtagHeaderFilter.java` > line 151).

 but in `checkNotModified`, it is strictly checking for 200 OK(`DefaultServerWebExchange.java` > 308 line, `ServletWebRequest.java` > 210 line). I’m wondering if these conditions should be aligned for consistency. Should `checkNotModified` also handle a broader range of successful status codes, or is there a reason to limit it to 200 OK?


bclozel:
Sorry for the late feedback, but I'm declining this change for the following reason: this changes a `ShallowEtagHeaderFilter` contract for no reason, and this protected method is probably extended by developers out there.
Unless there is a strong reason to break this contract, we'd rather avoid making the upgrade harder for people.

Thanks!

PR #33778, opened 2024-10-23
Relax the visibility of MockMVC DSL constructors

Hejow:
Hello, I added Kotlin DSL for RestDocs.

At first, I thought this PR might be more suitable for `spring-restdocs`, so I was confused where to make PR.
Due to the `internal constructor`, I’m submitting the PR here.

I positioned the package with the following three considerations in mind
1. The duplication issue that arises when placed in the same package as the existing Kotlin DSL.
2. Preventing confusion for users.
3. Reflecting the package structure of the existing Rest-docs module

I would appreciate any feedback on this PR. Thank You!!


sdeleuze:
Adding RestDocs specific code in Spring Framework is not ok, I would favor relaxing the visibility of the constructor. Could you please refine the PR accordingly and add related test, or create a related issue and close this PR?

Hejow:
Thank you for ur feedback!
I guess it's much better to relaxing visibility of the constructor and let user to extend by themselves!
I will create an issue and close this pr Thx!!! 👍

Hejow:
@sdeleuze Sorry that im lateee.. 😭 
I refined this PR by relaxing visibility of mockMvc DSL constructor for users to customize themselves.
So, previously implemented Restdocument DSL is no longer needed, i removed it!

sdeleuze:
Could we use `protected` constructors instead of `public` ones?

Hejow:
@sdeleuze I love to restrict visibility of the constructors. 
But in this case, when we restrict them as `protected` each of DSL cannot access lamda. 🥹

p.s. thx for ur feedbacks!!! 👍

sdeleuze:
Ah yeah, indeed.

sdeleuze:
Squashed and merged, thanks!

PR #33793, opened 2024-10-25
Optimize `BeanMap#entrySet`

gudrb33333:
Hello, I have been learning a lot while analyzing this open source project recently. During my analysis, I noticed a TODO message and came up with an simple idea.

- Enhancing memory efficiency by setting the initial capacity of a HashMap.

I have read the contributor guidelines in detail, but there may still be some mistakes.

### DIFF
--- a/spring-core/src/main/java/org/springframework/cglib/beans/BeanMap.java
+++ b/spring-core/src/main/java/org/springframework/cglib/beans/BeanMap.java
@@ -17,10 +17,11 @@
 package org.springframework.cglib.beans;
 
 import java.security.ProtectionDomain;
+import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


sdeleuze:
I see a remaining `TODO: optimize` on `BeanMap#entrySet`, couldn't we directly use `Collections.unmodifiableSet()` with a `Set` where we define the initial capacity and remove this TODO?

gudrb33333:
I’m not sure if I fully understood it correctly since my English isn’t very strong, but I didn’t remove the TODO because it’s just a simple idea, and I think there may still be room for further optimization.

gudrb33333:
Do you mean using a Set directly without copying it into a HashMap?

sdeleuze:
Yes

gudrb33333:
I researched and discovered SimpleImmutableEntry, then modified the code to directly use Collections.unmodifiableSet().

sdeleuze:
Thanks, make sense, but I initially missed those changes were done on the CGLIB fork that is included in Spring Framework, and we prefer avoiding doing such refinement on the fork, so I will decline. Thanks for your understanding.

gudrb33333:
Thank you for the review! It was very motivating for me.

PR #33804, opened 2024-10-28
Refactor MimeType BitSet initialization for improved readability and maintainability

gudrb33333:
Refactored BitSet initialization to remove repetitive set calls by introducing a char array (SEPARATORS).

- Enhanced readability: Separator characters are now defined in one place, making the code easier to understand.
- Improved maintainability: Adding or modifying separator characters is now simpler, requiring updates only to the SEPARATORS array.


bclozel:
Thanks for the proposal but I don't think this improves readability nor maintenance.

Issue #33712, opened 2024-10-15
Verify static resource location ends on slash


rstoyanchev:
A location that does not end on "/" does not actually work as intended, which is probably why it hasn't been reported as an issue. We should append the slash where we can, e.g. String location values, and assert `Resource` locations.

cagliostro92:
> A location that does not end on "/" does not actually work as intended, which is probably why it hasn't been reported as an issue. **We should append the slash where we can, e.g. String location values**, and assert `Resource` locations.

@rstoyanchev I know this is already closed, but I'm wondering why actual code doesn't try at all to append a slash, but it simply throws an exception if not already present.

PR #33817, opened 2024-10-30
Add trailing slash to static resource location if missing

### LINKED ISSUE #33815: ResourceHttpRequestHandler throwing IllegalArgumentException if resource doesn't end with slash breaks some third-party libraris
Originally raised in https://github.com/spring-projects/spring-boot/issues/42929 by @milazzo-g. This looks like a problem introduced by #33712, but I'm not sure if the fix should be in framework or the third-party library.

---

During context loading the `ResourceHttpRequestHandler` class scans all possible paths and if these do not end with "/" an `IllegalArgumentException` type exception is thrown with message "_Resource location does not end with slash: …_" (see `ResourceHttpRequestHandler@282` the call to `ResourceHttpRequestHandler::assertLocationPath`) this cause several issue with various jar included (in our case `camunda-webapp-webjar` **7.22.0** - included by `camunda-bpm-spring-boot-starter-root` **7.22.0)**.

The path provided by the library is "_classpath:/META-INF/resources/webjars/camunda_" which as you can see does not contain the final "/". makes it unusable with `spring-boot` **6.2.0-RC3**.

We temporary fixed creating a copy of `org.springframework.web.servlet.resource.ResourceHttpRequestHandler` in our src folder and modifying starting from line 282 from:

`ResourceHandlerUtils.assertLocationPath(location);`

to:

```
try {
 ResourceHandlerUtils.assertLocationPath(location);
} catch (Exception e) {
 if (org.apache.commons.lang3.StringUtils.containsIgnoreCase(e.getMessage(), "resource location does not end with slash")){
  location += "/";
 }else {
  throw e;
 }
}
```

this solved the issue. Could you please fix it in a next minor?

KR.

Giuseppe.

1_ResourceHttpRequestHandler
2_ResourceHandlerUtils
[STACKTRACE.txt](https://github.com/user-attachments/files/17559455/STACKTRACE.txt)

astrologic7:
As already stressed by @philwebb in #33815 and by my colleague @milazzo-g, the brand new assertion about the required trailing slash can lead to important breaking changes in both custom project and libraries. As mentioned by the author in #33712:

> A location that does not end on "/" does not actually work as intended, which is probably why it hasn't been reported as an issue. **We should append the slash where we can, e.g. String location values**, and assert Resource locations.

It seems to be clear that this constraint was intended to avoid unexpected behaviour, but the framework can still be accountable for adding the trailing slash without breaking other projects.

closes: #33815

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/resource/ResourceHandlerUtils.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/resource/ResourceHandlerUtils.java
@@ -38,6 +38,7 @@ import org.springframework.web.context.support.ServletContextResource;
  * {@link ResourceHttpRequestHandler} and {@link org.springframework.web.servlet.function}.
  *
  * @author Rossen Stoyanchev
+ * @author Edoardo Patti
  * @since 6.2
  */
 public abstract class ResourceHandlerUtils {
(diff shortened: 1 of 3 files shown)


cagliostro92:
Hi @bclozel, 
I would like to bring to your attention that the offending assertion is still present and invoked in [ResourceHttpRequestHandler](https://github.com/spring-projects/spring-framework/blob/main/spring-webmvc/src/main/java/org/springframework/web/servlet/resource/ResourceHttpRequestHandler.java#L184) and [PathResourceLookupFunction](https://github.com/spring-projects/spring-framework/blob/main/spring-webmvc/src/main/java/org/springframework/web/servlet/function/PathResourceLookupFunction.java#L50).
Let me know how you prefer handling such cases.
PS: I've rebased my branch onto your main a few minutes ago, so if you have already downloaded it, please drop it.

rstoyanchev:
Re-opening as per discussion under #33826.

rstoyanchev:
Thanks for the PR, and I will apply equivalent changes, but I need to start in 6.1.x where the code is different, and then forward merge to main.

PR #33824, opened 2024-10-31
Use `Map::computeIfAbsent` where feasible

quaff:


jhoeller:
We had some performance regressions in the past with the use of `computeIfAbsent`, so let's be rather careful there. There's also the overhead of the lambda instance but primarily it's the stronger locking of `computeIfAbsent` as opposed to separate `get` and `put` steps.

quaff:
> We had some performance regressions in the past with the use of `computeIfAbsent`, so let's be rather careful there. There's also the overhead of the lambda instance but primarily it's the stronger locking of `computeIfAbsent` as opposed to separate `get` and `put` steps.

I didn't aware of that, could someone helps to identify which parts should be reverted, or should I close this PR?

bclozel:
All calls to concurrent collections and ad hoc caches. Which is most of what this PR is.
The rest is mostly cosmetic changes and is not adding much.

PR #33825, opened 2024-10-31
Introduce default method `checked()` for throwable functional interface

quaff:
Allow throwing original checked exception instead of wrapping unchecked exception.

Use case:
https://github.com/spring-projects/spring-boot/blob/23fe3977d2d319f9f1bd0101f40dbb639ade7082/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/type/classreading/ConcurrentReferenceCachingMetadataReaderFactory.java#L71-L89

```java
	@Override
	public MetadataReader getMetadataReader(String className) throws IOException {
		MetadataReader metadataReader = this.classNameCache.get(className);
		if (metadataReader == null) {
			metadataReader = super.getMetadataReader(className);
			this.classNameCache.put(className, metadataReader);
		}
		return metadataReader;
	}

	@Override
	public MetadataReader getMetadataReader(Resource resource) throws IOException {
		MetadataReader metadataReader = this.resourceCache.get(resource);
		if (metadataReader == null) {
			metadataReader = createMetadataReader(resource);
			this.resourceCache.put(resource, metadataReader);
		}
		return metadataReader;
	}
```
can be rewritten to:
```java
	@Override
	public MetadataReader getMetadataReader(String className) throws IOException {
		return this.classNameCache.computeIfAbsent(className,ThrowingFunction.of((ThrowingFunction<String, MetadataReader>)super::getMetadataReader).checked());
	}

	@Override
	public MetadataReader getMetadataReader(Resource resource) throws IOException {
		return this.resourceCache.computeIfAbsent(resource,ThrowingFunction.of(this::createMetadataReader).checked());
	}
```

WDYT @philwebb

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/function/ThrowingBiFunction.java
+++ b/spring-core/src/main/java/org/springframework/util/function/ThrowingBiFunction.java
@@ -18,12 +18,15 @@ package org.springframework.util.function;
 
 import java.util.function.BiFunction;
 
+import static org.springframework.util.function.ThrowingFunction.throwAsUnchecked;
+
 /**
  * A {@link BiFunction} that allows invocation of code that throws a checked
  * exception.
  *
  * @author Stephane Nicoll
  * @author Phillip Webb
+ * @author Yanming Zhou
  * @since 6.0
  * @param <T> the type of the first argument to the function
  * @param <U> the type of the second argument to the function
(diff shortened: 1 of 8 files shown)


philwebb:
Personally, I'm not so keen on the (somewhat) hack that's being used to throw the checked exception from a signature that doesn't expose one.

```
@SuppressWarnings ("unchecked")
static <E extends Throwable> void throwAsUnchecked(Exception exception) throws E {
	throw (E) exception;
}
```

I think it somewhat goes against the spirit of the design.

snicoll:
Yeah I agree.

@quaff thanks for the suggestion, in any case.

PR #33827, opened 2024-10-31
Fail with full description for XML diff in `XmlExpectationsHelper`

Boiarshinov:
Only first difference is printed in assertion error when xml comparison is failed.
For example comparison of two xmls:
```xml
<root><a>a</a><b>b</b></root>
```
and
```xml
<root><b>b</b><c>c</c></root>
```
fails with message `Body content Expected child 'a' but was 'null' - comparing <a...> at /root[1]/a[1] to <NULL>`. There is nothing about child `c` in this message.

That is a strange behaviour of a xml-unit and it was updated with https://github.com/xmlunit/xmlunit/issues/232

To fail with full diff description we should use `fullDescription()` method instead of `toString()` on `Diff` class.

With the full description fail message would be:
```
Body content Expected child 'a' but was 'null' - comparing <a...> at /root[1]/a[1] to <NULL>
Expected child 'null' but was 'c' - comparing <NULL> to <c...> at /root[1]/c[1]
```

Full description is better for bug fixing.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/util/XmlExpectationsHelper.java
+++ b/spring-test/src/main/java/org/springframework/test/util/XmlExpectationsHelper.java
@@ -110,7 +110,7 @@ public class XmlExpectationsHelper {
 
 		@Override
 		public String toString() {
-			return this.diff.toString();
+			return this.diff.fullDescription();
 		}
 
 	}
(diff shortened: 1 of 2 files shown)


sbrannen:
This has been merged into `6.2.x` and `main`, for inclusion in Spring Framework 6.2.1.

Thanks 👍

PR #33828, opened 2024-10-31
Treat numbers as separate elements in JdbcUtils#convertPropertyNameToUnderscoreName

brieuc-le-faucheur:
Hello!

While using JdbcClient, I have encountered a specific case while querying a column containing a digit. The mapping method `JdbcUtils.convertPropertyNameToUnderscoreName` does not handle numbers as separate words.

Using PostgreSQL, given a table declared such as this:

```sql
create table demo (
    id text,
    name_with_1_number_in_label text
);
```

Querying it with a data class declared as below will result in a PSQLException:

```kotlin
data class DemoModel(val id: String, val nameWith1NumberInLabel: String)

@Repository
class SimpleRepository(private val jdbcClient: JdbcClient) {

    fun getBy(id: String): MutableList<DemoModel> {

        return jdbcClient.sql("""
            select * from demo where id = :id
        """.trimIndent())
            .param("id", id)
            .query(DemoModel::class.java)
            .list()
    }
}
```

I think isolating numbers with underscore would match naming usage more than having numbers stuck to the word preceding it in the column name.

Here is a sample project demonstrating the exception along with the PR.

[numbersInJdbcColumnNames.zip](https://github.com/user-attachments/files/17586667/numbersInJdbcColumnNames.zip)


sbrannen:
Please note that the [build failed](https://github.com/spring-projects/spring-framework/actions/runs/11610889414/job/32336061027?pr=33828) for your PR.

```
294 actionable tasks: 139 executed, 155 from cache

* What went wrong:
Execution failed for task ':spring-jdbc:checkstyleMain'.
> A failure occurred while executing org.gradle.api.plugins.quality.internal.CheckstyleAction
   > Checkstyle rule violations were found. See the report at: file:///home/runner/work/spring-framework/spring-framework/spring-jdbc/build/reports/checkstyle/main.html
     Checkstyle files with violations: 1
     Checkstyle violations by severity: [error:2]
```

Please make sure you run a full local build before submitting a PR.

Thanks

brieuc-le-faucheur:
Hello @sbrannen ! I updated the PR so that the full build passes. Sorry for the inconvenience.

snicoll:
Thanks for the suggestion @brieuc-le-faucheur but the input is in camelCase format in the aim to split words separated by underscores. As such, numbers should not be treated as you suggest. I'd also argue that this may break existing users that are relying on the number to be attached at the end of a word.

PR #33830, opened 2024-10-31
DataBufferUtils and BodyInserters: allow consumers of OutputStream to throw IOException

fedpet:
At the moment, users of DataBufferUtils.outputStreamPublisher are forced to catch and re-throw IOException whenever they actually write to the OutputStream.

Example:
```
DataBufferUtils.outputStreamPublisher(
    outputStream -> {
        try {
            outputStream.write(1);
            ...
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
    },
new DefaultDataBufferFactory(),
Runnable::run)
```

This PR aims to allow users to write cleaner code, example:
```
DataBufferUtils.outputStreamPublisher(
    outputStream -> {
        outputStream.write(1);
        ...
    },
new DefaultDataBufferFactory(),
Runnable::run)
```

As exceptions are already handled nicely under the hoods.

Same concept for BodyInserters.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
@@ -439,22 +439,22 @@ public abstract class DataBufferUtils {
 	 * @since 6.1
 	 */
 	public static Publisher<DataBuffer> outputStreamPublisher(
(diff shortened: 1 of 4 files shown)


snicoll:
Thanks for the suggestion but the change is breaking method signature and can't be accepted as such. I'd argue that manipulating an `OutputStream` means that you have to deal with the `IOException` per its contract.

PR #33831, opened 2024-11-01
Use optimistic locking where possible in `ResponseBodyEmitter`

lucky8987:
1. Use CAS instead of synchronized to set the complete state.
2. Reduce the synchronized range added by the callback function.


simonbasle:
Thanks for the PR @lucky8987, this looks like a good performance improvement. That close to the release, I'll review it a bit more in depth and discuss it with other team members, but I expect this should make the cut 👍

lucky8987:
> Thanks for the PR @lucky8987, this looks like a good performance improvement. That close to the release, I'll review it a bit more in depth and discuss it with other team members, but I expect this should make the cut 👍

OK，this is my first PR and I'm really looking forward to it being accepted.

simonbasle:
@lucky8987 good news, then: it's been merged 😉

Issue #29904, opened 2023-01-30
Document instrumentation for RestTemplate and/or WebClient


oppegard:
**Affects:** 6.0.x

---

The following callout was written in the [Spring Sleuth documentation for HTTP clients](https://docs.spring.io/spring-cloud-sleuth/docs/current/reference/html/integrations.html#sleuth-http-client-integration):

> You have to register RestTemplate as a bean so that the interceptors get injected. If you create a RestTemplate instance with a new keyword, the instrumentation does NOT work.

I can't figure out if the same guidance applies in 6.0.x for Micrometer Tracing at either:
- https://docs.spring.io/spring-framework/docs/6.0.4/reference/html/web.html#webmvc-client
- https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#actuator.micrometer-tracing

bclozel:
I'll create a follow-up issue in Spring Boot to do the same.

PR #33857, opened 2024-11-08
Document that Spring Boot will auto-configure `RestClient.Builder` beans with the observation registry

quaff:
See https://github.com/spring-projects/spring-boot/issues/38500


snicoll:
Thanks for the suggestion but if this needs to be described, it should be in the Spring Boot reference doc.

snicoll:
@bclozel see the above for context. I am reopening to get some more feedback as they are other places within our ref doc that have such references.

bclozel:
While we always work towards consistency, I find this PR very frustrating to deal with. The former sentence came from an actual discussion with a library maintainer in #29904. I don't think dedicating our time to discussing this is time well spent.

quaff:
One of this PR and #33866 should be merged to keep consistency. @bclozel

PR #33866, opened 2024-11-11
Remove "Spring Boot will auto-configure beans with the observation registry" from document

quaff:
>> it should be in the Spring Boot reference doc.

See https://github.com/spring-projects/spring-framework/pull/33857#issuecomment-2466596967


quaff:
cc @snicoll

snicoll:
That's really an odd way to collaborate. I've reopened the relevant PR

Issue #33888, opened 2024-11-15
When `Accept-Language` header is blank, the default locale should be returned


lllichen:

lllichen:
![image](https://github.com/user-attachments/assets/2e28dd27-bee2-4dc7-9bcd-a8b8d0468bd6)

bclozel:
Sending such a request header with an "empty text" value is I think invalid HTTP.
Can you name an HTTP client that does this?

ririnto:
I don't think this is a common case.

Is there a case where this change would be useful?

lllichen:
> Sending such a request header with an "empty text" value is I think invalid HTTP. Can you name an HTTP client that does this?


> Sending such a request header with an "empty text" value is I think invalid HTTP. Can you name an HTTP client that does this?

I think what you said is correct. However, sometimes the "accept_language" is manually added by the front-end colleagues. If they don't include this, I expect the return value to be the default value that I have set, rather than the default language of the machine.

lllichen:
Of course, this is just my suggestion. I've already overridden this class locally to solve this problem.

bclozel:
Thanks for letting us know. Let's consider this an invalid usage right now.

PR #33889, opened 2024-11-15
Return default locale when Accept-Language header is blank

lllichen:


bclozel:
Waiting for feedback in #33888

ririnto:
I don't think this is a common case.

Is there a case where this change would be useful?

PR #33900, opened 2024-11-16
Fix HtmlUnit dependency group id

durimkryeziu:
When using Spring Boot, this dependency is not managed nor works with `MockMvcWebClientBuilder.webAppContextSetup(context).build()` as described on this page


durimkryeziu:
I had a version mismatch! My bad :D

PR #33903, opened 2024-11-17
Simplify utility implementations in `spring-core`

polyglot-k:
You just received a pull request, so I changed the format similar to that.

- enhancement by adding early return
- delete the unnecessary 'else'

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
+++ b/spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
@@ -500,8 +500,10 @@ public class FastByteArrayOutputStream extends OutputStream {
 		 */
 		@Override
 		public void updateMessageDigest(MessageDigest messageDigest, int len) {
-			if (this.currentBuffer == null || len == 0) {
+			if (this.currentBuffer == null) {
 				// This stream doesn't have any data in it...
+				return;
+			}else if(len == 0){
 				return;
 			}
 			else if (len < 0) {
--- a/spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
+++ b/spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
@@ -98,13 +98,11 @@ public class FastByteArrayOutputStream extends OutputStream {
 		if (this.closed) {
 			throw new IOException("Stream closed");
 		}
-		else {
-			if (this.buffers.peekLast() == null || this.buffers.getLast().length == this.index) {
-				addBuffer(1);
-			}
-			// store the byte
-			this.buffers.getLast()[this.index++] = (byte) datum;
+		if (this.buffers.peekLast() == null || this.buffers.getLast().length == this.index) {
+			addBuffer(1);
 		}
+		// store the byte
+		this.buffers.getLast()[this.index++] = (byte) datum;
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
```diff
@@ -504,13 +500,10 @@ public void updateMessageDigest(MessageDigest messageDigest) {
 		 */
 		@Override
 		public void updateMessageDigest(MessageDigest messageDigest, int len) {
-			if (this.currentBuffer == null) {
+			if (this.currentBuffer == null || len == 0) {
```
sbrannen, line 503: Please revert the changes to this method, since the subsequent inline comment does not apply to the `len == 0` case.

spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
```diff
@@ -119,11 +119,10 @@ public boolean containsKey(Object key) {
 			}
 			@Override
 			protected boolean removeEldestEntry(Map.Entry<String, V> eldest) {
-				boolean doRemove = LinkedCaseInsensitiveMap.this.removeEldestEntry(eldest);
-				if (doRemove) {
+				if (LinkedCaseInsensitiveMap.this.removeEldestEntry(eldest)) {
```
sbrannen, line 122: Please revert the changes to the `removeEldestEntry()` method, since the proposal changes the logic.

sbrannen (changes requested):
I've requested that you revert a few changes.

KNU-K:
Thank you for your review. I've finished working on the revers.

spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java
```diff
@@ -507,8 +503,7 @@ public void updateMessageDigest(MessageDigest messageDigest, int len) {
 			if (this.currentBuffer == null) {
 				// This stream doesn't have any data in it...
 				return;
-			}
-			else if (len == 0) {
+			}else if(len == 0){
```
sbrannen, line 506: Please revert this unnecessary formatting change.

spring-core/src/main/java/org/springframework/util/MimeType.java
```diff
@@ -583,7 +581,6 @@ public int compareTo(MimeType other) {
 				}
 			}
 		}
-
```
sbrannen, line 586: Please revert this unnecessary formatting change.

sbrannen (changes requested):

KNU-K:
@sbrannen 
I'm not used to contributing, but thank you for your meticulous review

spring-core/src/main/java/org/springframework/util/MimeType.java
```diff
@@ -583,7 +581,7 @@ public int compareTo(MimeType other) {
...
+
```
sbrannen, line 584: I assume this will break the build, since we do not allow trailing whitespace.

In any case, please remove those tabs.

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/util/MimeType.java
KNU-K: I just removed it

sbrannen:
> When I checked the report because the build didn't pass, there was an error in the "reactive" part. However, I didn't touch that part, and it's not where I corrected it that depends.

Yep... `WebClientIntegrationTests.applyAttributesToNativeRequest()` is a flaky test that's been failing quite a lot lately, and the team is looking into how to address that.

So, you can safely ignore those failures in the interim.

KNU-K:
Thank you for your good review of the long work :) 
Have a nice day today.

sbrannen:
This has been merged into `6.2.x` and `main` in 2494ecb47bedb9c82c2a4c906005eb1cff8d5c22.

After merging, I noticed the copyright headers had not been updated, so I did that in 7f7819329c952a98a6e4ed81d376b5e8ca610e79.

Thanks for the PR and for making all of the requested changes.

PR #33905, opened 2024-11-17
Refactor ConcurrentLruCache

rlaehd62:
Forgot to include this change, my previous PR was declined but I create new one for it.
Sorry for inconvenience, I'm not used to contribution.

- Enhance by adding early return, so that It doesn't run remaining codes.
  - I think it's better to place non-null case first in terms of readability 'cause It's still convenience to follow the flow even when new error-catches added.
- Update copyright message (2023 to 2024)


bclozel:
I don't think that accepting cosmetic changes on performance-sensitive code is a good tradeoff for the project. Closing this issue.

PR #33906, opened 2024-11-17
Remove label usage for improved efficiency

polyglot-k:
As far as I know, explicitly using "Label" is simple, but it's disorienting for a debugger, which is why I tried to create a more intuitive format.

- Create a class for Context
- Isolation of functions matching SubPatterns

Through this, we made it more readable.


bclozel:
While labels are somewhat unusual in our codebase, we don't think that accepting cosmetic changes on performance-sensitive code is a good tradeoff for the project.

PR #33928, opened 2024-11-20
RestClient does not expose full URI template as attribute

sonallux:
This PR closes #33927 by using the same fix that has been applied to the `WebClient` in 2e07f9ab33d882876f46912fcea08030b2593d49 and 2b4ffe03915187746c30d76f69c1f27bf1af0895.


bclozel:
Thanks for your contribution @sonallux this is now merged.

PR #33944, opened 2024-11-22
Update URL for Contributor License Agreement in CONTRIBUTING.md

tofi86:
The URL for the Contributor License Agreement in `CONTRIBUTING.md` is outdated. I updated it with the latest URL.


snicoll:
@tofi86 Thanks for the suggestion but the redirect is on purpose as pivotal is a former entity.

Issue #32383, opened 2024-03-06
Use `ELContext` instead of deprecated `VariableResolver` API in `JspPropertyAccessor`


sbrannen:
The JSP `VariableResolver` API has been deprecated since JSP 2.1 in favor of the newer `ELContext` API.

We should therefore use `ELContext` instead of the deprecated `VariableResolver` API in `JspPropertyAccessor`.

PR #33945, opened 2024-11-22
Fix implicit variable resolution in JSP `EvalTag`

### LINKED ISSUE #33942: EvalTag implicit variables resolving broken
After upgrading to Spring Boot 3.4.0 (So Spring Framework 6.2.0) I noticed that for example, the following code on a JSP page throws an error: `<spring:eval expression="pageContext.toString()"/>` , while the same code does work on previous Spring Framework versions, for example 6.1.15.

After digging into the `EvalTag` code, I noticed that in the `resolveImplicitVariable` method the arguments for the `ELResolver#getValue` call are in the wrong order, thus causing the resolveImplicitVariable to always return null.

The signature for the method is:

```
public abstract Object getValue(ELContext context, Object base, Object property);
```

So I suppose the `name` parameter passed to `resolveImplicitVariable` should correspond to the `property`, not the `base`?

I could also try to provide a PR for this

micopiira:
## Overview

The order of parameters passed to `ELResolver#getValue` was incorrect.

The `name` should correspond to the `property` parameter of the `getValue` method instead `base`.

## Related Issues

- #32383
- #33942

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/tags/EvalTag.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/tags/EvalTag.java
@@ -259,7 +259,7 @@ public class EvalTag extends HtmlEscapingAwareTag {
 				return null;
 			}
 			try {
-				return this.elContext.getELResolver().getValue(this.elContext, name, null);
+				return this.elContext.getELResolver().getValue(this.elContext, null, name);
 			}
 			catch (Exception ex) {
 				throw new AccessException(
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @micopiira,

This has been merged into `6.2.x` and `main` in 71f872e8bbfda69375f434ae35bf53b1786c08aa and slightly revised in 3d0fffa8e4a7bcfc0a0e4c11ad863bb08ebd9e2e and a89db89fc0f60f659b4156f7d45a985eccc35f4d and will be available in Spring Framework 6.2.2.

For future PRs, please make sure you use `BDDMockito.given` instead of `Mockito.when`; otherwise, the build fails due to our Checkstyle rules. Running a local Gradle build should also help to catch that.

Thanks for your contribution!

Sam

Issue #33954, opened 2024-11-25
PathMatchingResourcePatternResolver should not log directory-skip messages at info level


CHOICORE:
I started seeing unexpected logs after upgrading Spring Boot from version 3.3.4 to 3.4.0.

```
Skipping search for files matching pattern [**/*.class]: directory [/Users/core/workspace/me/projects/samples/test/dependencies/build/classes/java/main/me/choicore/samples/test/api] does not exist
```
```
rootPorject
ㄴ applicaion-module
    ㄴ src
        ㄴmain
            ㄴ java
                ㄴ me/choicore/samples/test
                    ㄴ api
                        ㄴ Application.java (entrypoint)
ㄴ dependencies-module
    ㄴ src
        ㄴ main
            ㄴ java
                ㄴ me/choicore/samples/test
                    ㄴ dependencies
...
```

<img width="584" alt="image" src="https://github.com/user-attachments/assets/0acd431c-c938-447c-9cd8-ac2d0678822f">

 The project is set up as a multi-module structure, and as shown in the image, the entry point is located under the API package.
 

<img width="978" alt="image" src="https://github.com/user-attachments/assets/3c92629d-3389-4449-a98b-207d97261cdd">

After analyzing line 695 of PathMatchingResourcePatternResolver.class, I found a process that creates sub-resources. While the dependencies-module doesn't contain a package named 'api', related logs are being generated.
Although the application starts up normally, these logs keep appearing. If this is intended behavior, shouldn't we consider adjusting the logging level

If analysis is needed, I will attach a test project.

jhoeller:
Good point, we should switch to debug log level there.

CHOICORE:
> Good point, we should switch to debug log level there.

Submitted a PR for this simple contribution. 

refs:#33956

PR #33956, opened 2024-11-25
PathMatchingResourcePatternResolver should not log directory-skip messages at info level

CHOICORE:
Fixed as suggested in the https://github.com/spring-projects/spring-framework/issues/33954#issuecomment-2497550624
---

## Fixes:
- #33954: switch to debug log level


snicoll:
@CHOICORE thank you for making your first contribution to Spring Framework.

PR #33957, opened 2024-11-25
AsyncExecution fails to detect the return type of an annotated method from an interface with a generic

PaulNgo-BlueOC:
The last param of `AsyncExecutionAspectSupport.doSubmit` is a return type, which can be get from the `invocation`. But in case of interface with paramter type like:
```java
interface Service<O> {
  O doSomething();
}

class DefaultAsyncService implements Service<Future<String>> {
  @Override
  @Async
  public Future<String> doSomething() {
   return CompletableFuture.abc();
  }
}

@RestController
class Controller {
  @Autowire
  Service<Future<String>> service;

  @GetMapping
  public Future<String> api() {
    return service.doSomething();
  }
}
```
Return type retuned by `invocation.getMethod().getReturnType()` will be `Object` because we invoke `doSomething` by the interface, not by the impl.
My PR try to fix that

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/interceptor/AsyncExecutionInterceptor.java
+++ b/spring-aop/src/main/java/org/springframework/aop/interceptor/AsyncExecutionInterceptor.java
@@ -61,6 +61,7 @@ import org.springframework.lang.Nullable;
  * @author Juergen Hoeller
  * @author Chris Beams
  * @author Stephane Nicoll
+ * @author Bao Ngo
  * @since 3.0
  * @see org.springframework.scheduling.annotation.Async
  * @see org.springframework.scheduling.annotation.AsyncAnnotationAdvisor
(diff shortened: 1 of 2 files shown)


snicoll:
Can you please add a unit test that exercise the scenario you've described?

anaconda875:
> Can you please add a unit test that exercise the scenario you've described?

Sure

snicoll:
Isn't that the same thing? I'd like that we see the change covered by an actual test that shows the behavior you've described is now working.

anaconda875:
> Isn't that the same thing? I'd like that we see the change covered by an actual test that shows the behavior you've described is now working.

Ok I will add it soon. Thank you

anaconda875:
I added unit test

jhoeller:
Could we simply always use `userMethod.getReturnType()` instead of `invocation.getMethod().getReturnType()`?

snicoll:
@anaconda874 thanks very much for making your first contribution to Spring Framework.

snicoll:
@anaconda874 I am not sure what you mean. Please take the time to look first. I thanked you for your contribution and it was merged, you can see the commits above.

Issue #33748, opened 2024-10-18
Upgrade ServletHttpHandlerAdapter to rely on Servlet 6.1 ByteBuffer support


jhoeller:
As of Servlet 6.1, there is `ByteBuffer` support in `ServletInputStream`/`ServletOutputStream` which can be used in `ServletHttpHandlerAdapter` and inherited by `TomcatHttpHandlerAdapter`.

PR #33967, opened 2024-11-26
Update copyright headers

ngocnhan-tran1996:
See 
- gh-33748
- gh-33749
- gh-33929


bclozel:
Thanks but I don't think that processing PRs for updating copyright headers is a good use of our time.

PR #33978, opened 2024-11-27
fix: Ensure protocol slashes are preserved in `getSanitizedPath`

sanhee:
## Description
When using `WebClient` with `UriBuilder` to construct URIs, I encountered a problem where the protocol's double slashes (`://`) were incorrectly reduced to a single slash (`:/`).
As a result, malformed URIs such as **http:`/`jsonplaceholder.typicode.com/users** were generated, causing WebClient to throw a `Host is not specified` exception.
This prevented the request from reaching the server and resulted in a `500 error`.

This behavior occurred because the `getSanitizedPath` method in the `UriComponentsBuilder class` attempted to remove redundant slashes, but inadvertently changed the protocol structure by failing to preserve the `://` delimiter.

## Error Sample
```
org.springframework.web.reactive.function.client.WebClientRequestException: Host is not specified
	at org.springframework.web.reactive.function.client.ExchangeFunctions$DefaultExchangeFunction.lambda$wrapException$9(ExchangeFunctions.java:137) ~[spring-webflux-6.2.0.jar:6.2.0]
Assembly trace from producer [reactor.core.publisher.MonoErrorSupplied] :
	reactor.core.publisher.Mono.error(Mono.java:315)
	org.springframework.web.reactive.function.client.ExchangeFunctions$DefaultExchangeFunction.wrapException(ExchangeFunctions.java:137)
	*____________Mono.error ⇢ at org.springframework.web.reactive.function.client.ExchangeFunctions$DefaultExchangeFunction.wrapException(ExchangeFunctions.java:137)
	*____Mono.onErrorResume ⇢ at org.springframework.web.reactive.function.client.ExchangeFunctions$DefaultExchangeFunction.exchange(ExchangeFunctions.java:106)
	|_             Mono.map ⇢ at org.springframework.web.reactive.function.client.ExchangeFunctions$DefaultExchangeFunction.exchange(ExchangeFunctions.java:107)
	|_        Mono.doOnNext ⇢ at org.springframework.web.reactive.function.client.DefaultWebClient$ObservationFilterFunction.filter(DefaultWebClient.java:745)
	*____________Mono.defer ⇢ at org.springframework.web.reactive.function.client.DefaultWebClient$DefaultRequestBodyUriSpec.lambda$exchange$12(DefaultWebClient.java:467)
	|_           checkpoint ⇢ Request to GET http:/jsonplaceholder.typicode.com/users [DefaultWebClient]
	|_   Mono.switchIfEmpty ⇢ at 
...
```

## Changes
- Refined logic to ensure protocol slashes are preserved.
- Cleaned up redundant slashes in the path without affecting the protocol.

## Test
- To validate and fix this behavior, I tested the **private** `getSanitizedPath` method using **reflection**. 
   - Since the method is private and part of an internal class, reflection was necessary to access and evaluate its behavior in isolation.
- Verified with multiple inputs, including paths with and without protocols, to ensure correct behavior.

| AS-IS (Existing Issue) | TO-BE (Fixed Behavior) |
|-------|--------|
| 스크린샷 2024-11-27 오후 11 50 32 | image|

## Example Test Case
Input: `http://example.com//path/to//resource`  
Output: `http://example.com/path/to/resource`

Input: `https://example.com//path`  
Output: `https://example.com/path`

Input: `example.com//path`  
Output: `example.com/path`

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -908,16 +908,20 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 		}
 
 		private static String getSanitizedPath(final StringBuilder path) {
-			int index = path.indexOf("//");
-			if (index >= 0) {
-				StringBuilder sanitized = new StringBuilder(path);
-				while (index != -1) {
-					sanitized.deleteCharAt(index);
-					index = sanitized.indexOf("//", index);
-				}
-				return sanitized.toString();
+			String pathString = path.toString();
+			int protocolIndex = pathString.indexOf("://");
+			boolean hasProtocol = protocolIndex != -1;
+
+			if (hasProtocol) {
+				String protocol = path.substring(0, protocolIndex + 3);
+				String restOfPath = path.substring(protocolIndex + 3);
+
+				String sanitizedRestPath = restOfPath.replaceAll("/{2,}", "/");
+
+				return protocol + sanitizedRestPath;
 			}
-			return path.toString();
+
+			return pathString.replaceAll("/{2,}", "/");
 		}
 
 		public void removeTrailingSlash() {
--- a/spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
+++ b/spring-web/src/test/java/org/springframework/web/util/UriComponentsBuilderTests.java
@@ -16,6 +16,7 @@
 
 package org.springframework.web.util;
 
+import java.lang.reflect.Method;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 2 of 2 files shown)


sanhee:
Hi @bclozel 
Hope you're well! I noticed that the first build for this [PR failed](https://github.com/spring-projects/spring-framework/actions/runs/12077722126), but I've since fixed the issue and pushed an update. 
It seems that the actions need to be retriggered for the updated changes.

If you have some time, would you mind retriggering the actions? 
I'd also really appreciate it if you could take a look at the changes and let me know if there's anything else I should adjust or clarify! 😊

rstoyanchev (commented):
The FullPathComponentsBuilder expects to operate on a URL path, and as such the example test inputs is not something that's really expected. Please explain how the WebClient and UriBuilder are are used in order to arrive at this scenario.

rstoyanchev:

PR #33997, opened 2024-12-01
Refactor getHeaders() method to use early return

khyojun:
## Changes & Considerations

- In this case, the if-else if structure was used for the conditions, but since there aren’t too many scattered branches, I thought it might be better to refactor the code for readability using an early return approach. So, I’ve proposed the following changes!


bclozel:
I don't think this PR is improving things.

Issue #32384, opened 2024-03-06
serde framework detection should be centralized and should not be hard-coded


volkert-fastned:
In Spring Framework 6.1.4, found five different locations that have a hard-coded serde framework lookup mechanism like this, in a `static` block:

```java
	static {
		ClassLoader classLoader = [CLASS NAME].class.getClassLoader();
		jaxb2Present = ClassUtils.isPresent("jakarta.xml.bind.Binder", classLoader);
		jackson2Present = ClassUtils.isPresent("com.fasterxml.jackson.databind.ObjectMapper", classLoader) &&
						ClassUtils.isPresent("com.fasterxml.jackson.core.JsonGenerator", classLoader);
		jackson2XmlPresent = ClassUtils.isPresent("com.fasterxml.jackson.dataformat.xml.XmlMapper", classLoader);
		jackson2SmilePresent = ClassUtils.isPresent("com.fasterxml.jackson.dataformat.smile.SmileFactory", classLoader);
		gsonPresent = ClassUtils.isPresent("com.google.gson.Gson", classLoader);
		jsonbPresent = ClassUtils.isPresent("jakarta.json.bind.Jsonb", classLoader);
		kotlinSerializationCborPresent = ClassUtils.isPresent("kotlinx.serialization.cbor.Cbor", classLoader);
		kotlinSerializationJsonPresent = ClassUtils.isPresent("kotlinx.serialization.json.Json", classLoader);
		kotlinSerializationProtobufPresent = ClassUtils.isPresent("kotlinx.serialization.protobuf.ProtoBuf", classLoader);
	}
```

This is duplicated in the following classes, as far as I could find:

- AllEncompassingFormHttpMessageConverter.java
- BaseDefaultCodecs.java
- DefaultRestClientBuilder.java
- RestTemplate.java
- WebMvcConfigurationSupport.java

In my experience, this poses multiple problems:

- This behavior is very hard to intercept, override and/or customize
- Different Spring components require this to be overridden separately

The fact that these are hard-coded classpath lookups means that it's very hard to explicitly select a preferred serde framework, when more than one of these happen to live in the classpath, for instance, as a pulled-in transative dependency in a non-Spring library. As a result, we are dependent on whatever opinionated preference the Spring Framework itself ends up having, depending on what it finds in the classpath.

[...]

PR #34001, opened 2024-12-02
Auto-detect Jackson XML and JAXB in RestClient

### LINKED ISSUE #34000: RestClient should auto-detect MappingJackson2XmlHttpMessageConverter
Related to #33635.

xehpuk:
Closes gh-34000


bclozel:
I don't think this is an oversight. Requesting XML using HTTP clients is a bit out of fashion so we don't intend to add this by default.

bclozel:
Automatic classpath detection and general support are different things. You can already configure XML support in `RestClient` quite easily. Whether we automatically configure the converter in `RestClient` and the ordering matters. Community members have recently complained that this mechanism is too agressive and that libraries can sometimes be on your classpath against your will (see #32384). For more opinionated preferences, maybe we should defer to Spring Boot.

PR #34017, opened 2024-12-04
Fix ApplicationListener no longer invoked for generic ApplicationEvent

### LINKED ISSUE #33982: `ApplicationListener` no longer invoked for generic `ApplicationEvent` with 6.2.0
Hello!

I have simple interface like this:

```java
public interface ITest
{
}
```

and the simple event:

```java
public class TestEvent<T extends ITest> extends ApplicationEvent
{
    public TestEvent(Object source)
    {
        super(source);
    }
}
```

and two listeners:

1. 
```java
@Component
public class MethodListener
{
    @EventListener
    public void handleTestEvent(TestEvent<ITest> iTest)
    {
        System.out.println(this.getClass() + " DONE!");
    }
}
```
2.
```java
@Component
public class SimpleTestListener implements ApplicationListener<TestEvent<ITest>>
{
    @Override
    public void onApplicationEvent(TestEvent<ITest> event)
    {
        System.out.println(this.getClass() + " DONE!");
    }
}
```

and when i trying to send an event via the spring context like this:

```java
public class MySpringApplication
{
    public static void main(String[] args)
    {
        ApplicationContext context = new AnnotationConfigApplicationContext(AppConfig.class);
        context.publishEvent(new TestEvent<>(""));
    }
}
```
i have different result in depends on the version Spring. When i use Spring 6.2 i can see in the log this:

```
class ... MethodListener DONE!
```

when i use Spring 6.1.15 i see both results:

```
class ...SimpleTestListener DONE!
class ...MethodListener DONE!
```

I guess it's a bug.

PaulNgo-BlueOC:
Fix #33982
In case `typeBounds` WildcardBounds has `TypeVariable` bound, we should check whether `typeBounds` isAssignableFrom the current ResolvableType or not

NOTE:
1. I'm new contributor, so my code might be not perfect. But please give feebacks, I will try to address them!
2. Do you think we nee to handle case `typeBounds` WildcardBounds has `WildcardType` bound? If yes, I will add some more commits!

Thank you!!

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/ResolvableType.java
+++ b/spring-core/src/main/java/org/springframework/core/ResolvableType.java
@@ -342,15 +342,14 @@ public class ResolvableType implements Serializable {
 				return (ourBounds.isSameKind(typeBounds) &&
 						ourBounds.isAssignableFrom(typeBounds.getBounds(), matchedBefore));
 			}
-			else if (upUntilUnresolvable) {
+			if (upUntilUnresolvable || other.isUnresolvableTypeVariable()) {
 				return typeBounds.isAssignableFrom(this, matchedBefore);
 			}
-			else if (!exactMatch) {
+			if (!exactMatch) {
 				return typeBounds.isAssignableTo(this, matchedBefore);
 			}
-			else {
-				return false;
-			}
+
+			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


jhoeller:
Superseded by https://github.com/spring-projects/spring-framework/issues/33982#issuecomment-2520879941

Issue #33918, opened 2024-11-19
Enforce Servlet 6.1 and WebSocket 2.2 baselines


bclozel:
Spring Framework 7.0 has a Jakarta EE 11 baseline, including Servlet 6.1 and WebSocket 2.2.

This is an umbrella issue that references relevant tasks.

PR #34026, opened 2024-12-04
Upgrade to Jetty 12.1

m1ngyuan:
A part of #33918


snicoll:
Thanks for the PR but we don't need something that does only dependency upgrades. We'll look at #33918 in due course.

PR #34027, opened 2024-12-04
Update server.adoc for Jakarta EE 11

m1ngyuan:


bclozel:
Sorry but we will need to update way more that that in this section of the reference documentation.

PR #34034, opened 2024-12-05
 Add property accessors to `ExecutorConfigurationSupport`

lucky8987:
Sometimes I need to check if the `ExecutorConfigurationSupport` executor has enabled virtual threads, so I added the corresponding getter methods.


snicoll:
> Sometimes I need to check if the ExecutorStructureSupport executor has enabled virtual threads,

Why is that? And why do you need the bean's name?

lucky8987:
> > Sometimes I need to check if the ExecutorStructureSupport executor has enabled virtual threads,
> 
> Why is that? And why do you need the bean's name?

Because I want to print executor related information on the actuator endpoint, I need to obtain beanName as the key. Of course, this is not necessary, and I can also obtain it through other means.

snicoll:
@lucky8987 you need to reset your "dev" branch to what you intend to submit in this PR. Right now there are a number of unrelated commits in it.

lucky8987:
> @lucky8987 you need to reset your "dev" branch to what you intend to submit in this PR. Right now there are a number of unrelated commits in it.
Sorry, my other change should not have been submitted here. I have now reset this branch and I really hope that this PR can be merged.

snicoll:
Thanks for the suggestion but the flag is a configuration shortcut to enable VT. We can't ensure it has the parity that you expect (i.e. it's not reliable to rely on the internal boolean flag to determine that VT is enabled).

PR #34045, opened 2024-12-07
Fix `MediaType#ALL` javadoc

ngocnhan-tran1996:
Related https://github.com/spring-projects/spring-framework/issues/34047

Reference
https://docs.spring.io/spring-framework/docs/7.0.0-SNAPSHOT/javadoc-api/org/springframework/http/MediaType.html


bclozel:
Thanks for the PR, but we went with a more comprehensive update in https://github.com/spring-projects/spring-framework/commit/ff28d2d47af53a6fbf9fd04a15fdb9c920264412

PR #34053, opened 2024-12-08
Refactor ResourceRegionHttpMessageConverter

divcon:
Extract collection handling into a separate method and use a stream-based all-match check to improve readability and maintainability. 
Also introduce new test cases to verify functionality and reliability of the supportsRepeatableWrites function.


bclozel:
Thanks for the PR, but I don't think the code change really improves readability and maintanability (those are really subjective). Also we usually avoid introducing `Stream` on possibly hot code paths for performance reasons.

PR #34068, opened 2024-12-11
Clarify the behaviour of `@Bean(autowireCandidate=false)`

quaff:
See https://github.com/spring-projects/spring-boot/issues/43481


snicoll:
Thanks for the suggestion but the linked issue in the Spring Boot project hasn't been triaged. As such, this issue is not actionable and we'll update the documentation when tackling it accordingly. Besides, this is no longer accurate with recent changes in `ObjectProvider`.

PR #34069, opened 2024-12-11
Polishing ConfigurationClassBeanDefinitionReader

kse-music:
Reduce redundant code


snicoll:
Thanks but that kind of optimizations is unnecessary, in particular in such advanced code base. We prefer code branches to be explicit

PR #34074, opened 2024-12-11
Invalid `@AliasFor` declaration in `@RequestHeader`

enbraining:
Aren't the name and value reversed in the AliasFor annotation?


sbrannen:
Hi @enbraining,

> Sorry, my mistake

No problem.

In the future, please keep in mind that it's better to first discuss a potential bug in an issue before submitting a PR, and there is never a need to open both an issue and a PR at the same time.

Also, questions like this are better suited for places like [Stack Overflow](http://stackoverflow.com/).

Regards,

Sam

PR #34077, opened 2024-12-11
Prevent execution of Antora jobs on forks

scordio:
Prior to this commit, any fork that copied the `main` branch only experienced daily failures at 10am UTC due to the scheduled execution of the `update-antora-ui-spring` workflow. With this change, the workflow jobs are executed only when under the `spring-projects` GitHub organization.

The failures of my fork: https://github.com/scordio/spring-framework/actions/workflows/update-antora-ui-spring.yml

An alternative solution would be [disabling the workflow](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow#disabling-a-workflow) on each affected fork.

### DIFF
--- a/.github/workflows/update-antora-ui-spring.yml
+++ b/.github/workflows/update-antora-ui-spring.yml
@@ -13,6 +13,7 @@ permissions:
 jobs:
   update-antora-ui-spring:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks @scordio - that looks like an oversight on our end.

Issue #33968, opened 2024-11-26
Generic type resolving broken in 6.2.0 (regression of GH-24963)


rPraml:
Hello, I want to report a possible regression in 6.2.0

We have some rest controllers that use jackson with `@JsonTypeInfo` annotations.

With 6.1.x the attached code works and returns the expected JSON: `[ { "type" : "one" }, { "type" : "two" } ]`
With 6.2.0, the type properties are missing and `[ {}, {} ]` is returned.

I've tracked down, that this commit may have changed the behaviour: https://github.com/spring-projects/spring-framework/commit/e788aeb25baba51880b6c9cccbeb51fe5a6d401b

I'm unsure, if this is "works as designed" or if that code path got lost through this commit:
The method declaration  `public <T extends BaseType> List<T> get()` may be a bit strange, but valid java syntax.
When I use this method declaration `public List<BaseType> get()`, the controller will work again.  (@quaff, @jhoeller FYI)

```java
@RestController
public class TestController {
	@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
	@JsonSubTypes({
			@JsonSubTypes.Type(value = SubType1.class, name = "one"),
			@JsonSubTypes.Type(value = SubType2.class, name = "two")
	})
	public static class BaseType {

	}

	public static class SubType1 extends BaseType {
	}

	public static class SubType2 extends BaseType {
	}

	@GetMapping(value = "/test", produces = MediaType.APPLICATION_JSON_VALUE)
	public <T extends BaseType> List<T> get() {
		List<T> list = new ArrayList<>();
		list.add((T) new SubType1());
		list.add((T) new SubType2());
		return list;
	}
}
```

ThanksForAllTheFish:
I think this is similar to something I observed too, but tracked to https://github.com/spring-projects/spring-framework/issues/20727

In our case, a bean defined as `fun <V> kafkaProducerFactory(valueSerializer: Serializer<V>): ProducerFactory<String?, V>` (`org.springframework.kafka.core.ProducerFactory`) was not recognized as a valid candidate to https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java#L92 (expecting a `ProducerFactory<Object, Object> kafkaProducerFactory`

```
	@Bean
[...]

PR #34086, opened 2024-12-13
Generic with a base type is no longer resolved to its base if the generic is unresolved

rPraml:
I tried to make a unit test and a possible fix for #33968 

Someone who knows more about it than me should take another look at the fix ;)

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
+++ b/spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
@@ -180,7 +180,7 @@ public final class GenericTypeResolver {
 								generics[i] = resolvedTypeArgument;
 							}
 							else {
-								generics[i] = ResolvableType.forType(typeArgument);
+								generics[i] = ResolvableType.forType(typeArgument).resolveType();
 							}
 						}
 						else if (typeArgument instanceof ParameterizedType) {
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/core/GenericTypeResolver.java
```diff
@@ -180,7 +180,7 @@ else if (genericType instanceof ParameterizedType parameterizedType) {
 								generics[i] = resolvedTypeArgument;
 							}
 							else {
-								generics[i] = ResolvableType.forType(typeArgument);
+								generics[i] = ResolvableType.forType(typeArgument).resolveType();
```
rPraml, line 183: as far as I understand the code here, we want to resolve the `T` of a generic. If the `T` is not defined by the contextType, (because ist is no `class Foo<T extends Bar>`) we get to this line.

In my case, the T is bound by the function `public <T extends BaseType> List<T> get()` and means, that `T` is at least a `BaseType`

IMHO the type must be resolved here, so that the following code knows , that the elements in the list are `BaseType` or subtypes. This is later required for jackson, otherwise we have problems mentioned in this [SO](https://stackoverflow.com/questions/34193177/why-does-jackson-polymorphic-serialization-not-work-in-lists) post.

snicoll:
Thanks very much @rPraml

PR #34100, opened 2024-12-17
Update BeanUtils.java

JabborovEgamberdi:
Functional programming added, nested if logics are optimized


bclozel:
Thanks for your enthusiasm but we would rather apply cosmetic changes ourselves. We value community contributions but we would rather have them on actual bugs, documentation and enhancements.

Keep an eye [on "ideal for contribution" issues](https://github.com/spring-projects/spring-framework/labels/status%3A%20ideal-for-contribution).

Issue #34082, opened 2024-12-12
ReflectionUtils behavior change in 6.2.0


welovelain:
After upgrading to Spring 6.2.0 (spring boot 3.4.0) my tests with io.rest-assured mock mvc started to be flaky.

I discovered that the issue might be connected to `ReflectionUtils`.

In `RestAssuredMockMvc` there is call to
https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java through `ReflectionUtils`.

When

```java
public static Method findMethod(Class<?> clazz, String name, @Nullable Class<?>... paramTypes)
```

is called, then method returns.

Sometimes, `method.isVarArgs` returns `true`, sometimes `false` (when `method.isBridge()` is `true`).

Here's some test example to show the behavior:

```java
    @Test
    void testMethod() {
        var method = ReflectionUtils.findMethod(MockHttpServletRequestBuilder.class, "cookie", Cookie[].class);
        System.out.println("IsBridge: " + method.isBridge());
        System.out.println("IsVarArgs: " + method.isVarArgs());

        assertTrue(method.isVarArgs());
    }
 ```

If you run the test multiple times, it will fail, but not always. In previous spring version `method.isVarArgs` always returned `true`. (At least I think so, it would explain why tests never failed before)

facewise:
It's so intersting. I found that `java.lang.Class#getDeclaredMethods()` sometimes returns methods in a different order. `ReflectionUtils` is innocent because it just copies the result😥

simonbasle:
@welovelain I see that you have opened an issue in `rest-assured`. I think this is a better avenue to explore, perhaps evolving the code there to use `BridgeMethodResolver.findBridgedMethod(method)`?

Unless you've investigated further and found an issue in `ReflectionUtils`?

@facewise thanks for the offer, but filtering out bridge methods in `findMethod` is a no-go as it is supposed to include such methods.

welovelain:
@simonbasle Yes, I created it there also, but it looks like the behavior still was changed, it might affect other libs too.
Because earlier it worked.
If the behavior was changed and you think that it must  have been changed, then ok. But to me it looks like a bug, when such method returns undetermined results

sbrannen:
Hi @welovelain,

[...]

PR #34110, opened 2024-12-18
Fix ReflectionUtils#getDeclaredMethods to provide consistent result

facewise:
Hi. This PR is inspired by gh-34082.

As @welovelain pointed in it, the test below sometimes passes but sometimes doesn't.

```java
import jakarta.servlet.http.Cookie;
import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
import java.lang.reflect.Method;

class ReflectionTests {
    @Test
    void test() {
        Method cookie = ReflectionUtils.findMethod(MockHttpServletRequestBuilder.class, "cookie", Cookie[].class);
        assertThat(cookie.isVarArgs()).isTrue();
    }
}
```

I've tested with IntelliJ IDEA 2024.3 and on several JDKs such as Temurin 17.0.13, Temurin 21.0.3, Corretto 21.0.5, etc., but the problem occurred upon all these JDKs.

You can check it out with this reproducer.

[sample.zip](https://github.com/user-attachments/files/18175606/sample.zip)

The problem is, how Class#getDeclaredMethods() returns a Method[] in a different order on each run. However `ReflectionUtils#getDeclaredMethods()` just iterates through this array and returns the first matched method.

Even `ReflectionUtilsTests#findMethod()` sometimes fails if you remove the sorting line introduced in this changes.

FWIW [ReflectionUtils in junit5](https://github.com/junit-team/junit5/blob/main/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java#L1771) (heavily contributed  by @sbrannen) also sorts methods of `Class#getDeclaredMethods()` and I believe this is worth it.

My sorter would not be optimized but I intended that the methods declared in actual source come first.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ReflectionUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ReflectionUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,6 +41,7 @@ import org.springframework.lang.Nullable;
  * @author Costin Leau
  * @author Sam Brannen
  * @author Chris Beams
+ * @author Yong-Hyun Kim
  * @since 1.2.2
  */
 public abstract class ReflectionUtils {
@@ -463,6 +464,7 @@ public abstract class ReflectionUtils {
 		if (result == null) {
 			try {
 				Method[] declaredMethods = clazz.getDeclaredMethods();
+				Arrays.sort(declaredMethods, ReflectionUtils::methodDeclaringClassHierarchySorter);
 				List<Method> defaultMethods = findDefaultMethodsOnInterfaces(clazz);
 				if (defaultMethods != null) {
 					result = new Method[declaredMethods.length + defaultMethods.size()];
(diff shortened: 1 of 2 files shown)


spring-core/src/test/java/org/springframework/util/ReflectionUtilsTests.java
```diff
@@ -231,6 +232,8 @@ void findMethod() {
 		assertThat(ReflectionUtils.findMethod(B.class, "bar", String.class)).isNotNull();
 		assertThat(ReflectionUtils.findMethod(B.class, "foo", Integer.class)).isNotNull();
 		assertThat(ReflectionUtils.findMethod(B.class, "getClass")).isNotNull();
+		assertThat(ReflectionUtils.findMethod(B.class, "baz", String[].class).isBridge()).isFalse();
```
sbrannen, line 235: Since there are no generics involved here, there is never a bridged method for `baz(String... args)`.

spring-core/src/main/java/org/springframework/util/ReflectionUtils.java
```diff
@@ -486,6 +488,18 @@ private static Method[] getDeclaredMethods(Class<?> clazz, boolean defensive) {
 		return (result.length == 0 || !defensive) ? result : result.clone();
 	}
 
+	private static int methodDeclaringClassHierarchySorter(Method method1, Method method2) {
+		Class<?> clazz1 = method1.getDeclaringClass();
+		Class<?> clazz2 = method2.getDeclaringClass();
+		if (clazz1 == clazz2) {
```
sbrannen, line 494: For methods in the array returned from `java.lang.Class.getDeclaredMethods()`, the declaring class is **always** the same class. Thus, there is no point in comparing the declaring classes.

sbrannen (changes requested):
Hi @facewise,

> The problem is, how `Class#getDeclaredMethods()` returns a `Method[]` in a different order on each run.

The Javadoc for that method states the following.

_The elements in the returned array are not sorted and are not in any particular order._

Similarly, the Javadoc for `ReflectionUtils.getDeclaredMethods(Class<?>)` states that it is a variant of `Class.getDeclaredMethods()` without specifying different ordering guarantees, and we have no plans to deviate from that behavior.

In light of the above combined with reasons I outlined in the comments I left in this review, we unfortunately have to decline this PR.

Regards,

Sam

PR #34115, opened 2024-12-18
Fix broken links in the web reference documentation

ngocnhan-tran1996:
- Fix reference link `functional endpoints` 
- Add breakline to display `See equivalent in the Reactive stack` or `See equivalent in the Servlet stack`


sdeleuze:
Merged, thanks for those useful fixes.

PR #34116, opened 2024-12-18
Clarify documentation for the default uri client request observation metric

### LINKED ISSUE #34107: Unclear documentation for the uri observability metric in HTTP clients
# Description

The current documentation for the uri metric for HTTP clients is:
`URI template used for HTTP request, or "none" if none was provided. Only the path part of the URI is considered.`
(found here: https://docs.spring.io/spring-framework/reference/integration/observability.html#observability.http-client)

It was last updated in relation to #29885, where the second sentence was added as a description for removing the protocol, host, and port part of the template.

I find "path" to be a semi-incorrect description, since the current behavior is that it retains query values (which are separate from _path_ https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax).

# Example

### TL;DR:

When executing:
`RestTemplate.exchange(https://example.org/resource/{id}?queryKey={queryValue}, GET, null, Response.class, "1", "query")`
the metric that is register has `uri` as `/resource/{id}?queryKey={queryValue}`, which contains more than just the path of the uri.

### Longer version

When calling RestTemplate `execute` with one of the signatures that support uriTemplate (e.g., https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java#L797), then I would expect that only the _path_ part of the uriTemplate would be registered to the `uri metric.`

So if my template is `https://example.org/resource/{id}?queryKey0={queryValue0}&queryKey1={queryValue1}` then I would expect the registered value to be `/resource/{id}`. But the actual registered value is `/resource/{id}?queryKey0={queryValue0}&queryKey1={queryValue1}`.

This is since the default behavior of the observation convention is to only removes the protocol, host and port (it is done here: https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/http/client/observation/DefaultClientRequestObservationConvention.java#L109).

# The issue

The low-cardinality metric might get a quite high cardinality if the client endpoints have a large variety of query parameters, which is not apparent in the documentation since it states that only the path is considered.

# Suggested solution

I would find it beneficial if the description was more specific and explicitly stated what the uri metric will contain by default.

Preferably test(s) could be introduced as well to illustrate that it is intended that the query parameters remain as part of the default metric.

# Sidenote

Is it intentional that the low cardinality metrics should include the templated query parameters? I would have expected the base behavior to strip the uri of everything except the path, and that the user then would have to extend the templating if they were to be interested in specific queries and to what degree they were present? Especially since the metric also comes with an complete path metric too with its high cardinality value (http.url).

Mattias-Sehlstedt:
Change the description for the uri client request observation metric to clarify that only the protocol, host and port parts are removed from the template with the default observability convention.

Adds tests to illustrate that it is intentional that the query parameters are kept as part of the template.

closes #34107

### DIFF
--- a/framework-docs/modules/ROOT/pages/integration/observability.adoc
+++ b/framework-docs/modules/ROOT/pages/integration/observability.adoc
@@ -285,7 +285,7 @@ Instrumentation uses the `org.springframework.http.client.observation.ClientRequ
 |===
 |Name | Description
 |`method` _(required)_|Name of the HTTP request method or `"none"` if not a well-known method.
-|`uri` _(required)_|URI template used for HTTP request, or `"none"` if none was provided. Only the path part of the URI is considered.
+|`uri` _(required)_|URI template used for HTTP request, or `"none"` if none was provided. The protocol, host and port part of the URI are not considered.
 |`client.name` _(required)_|Client name derived from the request URI host.
 |`status` _(required)_|HTTP response raw status code, or `"IO_ERROR"` in case of `IOException`, or `"CLIENT_ERROR"` if no response was received.
 |`outcome` _(required)_|Outcome of the HTTP client exchange.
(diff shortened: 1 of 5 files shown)


bclozel:
Thanks for your PR! Just realized that you didn't sign the CLA yet @Mattias-Sehlstedt, I'll put this issue back in triage until this is done.

Mattias-Sehlstedt:
I keep getting "Reauthorization required" when trying to sign the agreement, and I cannot see it as a know issue on the pivotal page (https://cla.pivotal.io/about). I have tried revoking the access and then enabling it again but the issue persists. Could this be considered an "Obvious fix", or should I be in contact with Pivotal and ask them about it?

image

bclozel:
Thanks @Mattias-Sehlstedt for your contribution!
FYI [we're moving away from the CLA and using signoffs now](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring) so this should avoid such problems in the future.

Mattias-Sehlstedt:
Thanks for the quick feedback and assistance @bclozel, makes it a very pleasant experience to contribute to open source 👍

PR #34138, opened 2024-12-22
Release connection after transaction if using Hibernate and open-in-view is on (turned off by default)

poxu:
One of the issues with open-in-view is that connections are not released for a long time, which can lead to connection pool starvation.

Sometimes, controller method even makes no queries for seconds and doesn't really need to keep the connection for itself all that time. For example that is often the case, when application makes REST call after making database query.

This pull request gives user setting, which tells transaction manager, that it should release JDBC connection after transaction is finished (successfully or not). Connection release only happens if the setting is turned on and if open-in-view is enabled. If open-in-view is disabled, then entity manager is closed after transaction is finished and connection is released anyway.

# What this pull request changes in existing behaviour

1.  JpaDialect#releaseJdbcConnection is now only called if entity manager is pre-bound. This happens if open-in-view is turned on.
2.  JpaDialect#cleanupTransaction is now called *before* JpaDialect#releaseJdbcConnection ,
because if connection is released, transaction manager can't reset isolation level and read-only flag. I think the only reason nobody noticed that before is that nobody tried to override JpaDialect#releaseJdbcConnection for HibernateJpaDialect

# What this pull request adds

1.  releaseConnectionAfterTransaction setting for HibernateJpaDialect and which user can configure through HibernateJpaVendorAdapter (like showSql and generateDdl).
2.  HibernateJpaDialect#releaseJdbcConnection method now really releases JDBC connection if releaseConnectionAfterTransaction setting is enabled.

P. S. I think those changes are ok, because the behaviour is copied from HibernateTransactionManager. Hibernate 6 is officially only supported through JpaTransactionManager, so it would be nice to port this HibernateTransactionManager feature.

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
@@ -47,6 +47,7 @@ import org.hibernate.exception.DataException;
 import org.hibernate.exception.JDBCConnectionException;
 import org.hibernate.exception.LockAcquisitionException;
 import org.hibernate.exception.SQLGrammarException;
+import org.hibernate.resource.jdbc.spi.LogicalConnectionImplementor;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.dao.CannotAcquireLockException;
@@ -89,6 +90,8 @@ public class HibernateJpaDialect extends DefaultJpaDialect {
 
 	boolean prepareConnection = true;
 
+	boolean releaseConnectionAfterTransaction = false;
+
 	private @Nullable SQLExceptionTranslator jdbcExceptionTranslator;
 
 	private @Nullable SQLExceptionTranslator transactionExceptionTranslator = new SQLExceptionSubclassTranslator();
@@ -116,6 +119,34 @@ public class HibernateJpaDialect extends DefaultJpaDialect {
 	 */
 	public void setPrepareConnection(boolean prepareConnection) {
 		this.prepareConnection = prepareConnection;
+	}
+
+	/**
+	 * Set, whether to release connection after transaction is commited or rolled
+	 * back. Only works for pre-bound sessions.
+	 * <p> This setting is needed to work around the fact, that it is not possible
+	 * to use {@link org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode#DELAYED_ACQUISITION_AND_RELEASE_AFTER_TRANSACTION}
+	 * handling mode without sacrificing setting non-default isolation levels
+	 * or read-only flag. Releasing connection might prevent connection pool
+	 * starvation if open-in-view is on.
+	 * <p> Default is "false" for backward compatibility. If you turn this flag
+	 * on it still does nothing unless session is pre-bound (most likely if
+	 * open-in-view) is off. If session is pre-bound and the flag is on, then
+	 * after transaction is finished (successfully or not), underlying JDBC
+	 * connection will be released and acquired later according to {@link org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode}
+	 * (is set to {@link org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode#DELAYED_ACQUISITION_AND_HOLD}
+	 * with {@link HibernateJpaVendorAdapter#getJpaPropertyMap()} if you don't
+	 * specify it yourself).
+	 * <p> Please pay attention, that this setting doesn't affect how Hibernate
+	 * handles connections, which were acquired on-demand, to lazily load
+	 * collections outside of transaction context.
+	 * <p> Specifically, connections, acquired to serialize entities, returned
+	 * by rest controller method will only be closed, after serialization is
+	 * complete. Hibernate will not acquire and release connections for each
+	 * lazy field loading.
+	 */
+	public void setReleaseConnectionAfterTransaction(boolean releaseConnectionAfterTransaction) {
+		this.releaseConnectionAfterTransaction = releaseConnectionAfterTransaction;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
```diff
@@ -118,6 +121,34 @@ public void setPrepareConnection(boolean prepareConnection) {
...
+	 * starvation if open-in-view is on.
+	 * <p> Default is "false" for backward compatibility. If you turn this flag
+	 * on it still does nothing unless session is pre-bound (most likely if
+	 * open-in-view) is off. If session is pre-bound and the flag is on, then
+	 * after transaction is finished (successfully or not), underlying JDBC
+	 * connection will be released and acquired later according to {@link org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode}
+	 * (is set to {@link org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode#DELAYED_ACQUISITION_AND_HOLD}
+	 * with {@link HibernateJpaVendorAdapter#getJpaPropertyMap()} if you don't
+	 * specify it yourself).
+	 * <p> Please pay attention, that this setting doesn't affect how Hibernate
```
mipo256, line 140: Am I correct that here you're talking about [this specific property that allows loading associations when the original transaction is already commited](https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/cfg/TransactionSettings.html#ENABLE_LAZY_LOAD_NO_TRANS)? Just to clarify.

spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
```diff
@@ -118,6 +121,34 @@ public void setPrepareConnection(boolean prepareConnection) {
...
+	 * handles connections, which were acquired on-demand, to lazily load
+	 * collections outside of transaction context.
+	 * <p> Specifically, connections, acquired to serialize entities, returned
```
mipo256, line 143: I do not understand this sentence and its purpose. What connections "acquired to serialize entities" are you talking about here?

> Specifically, connections, acquired to serialize entities, returned by rest controller method will only be closed, after serialization is complete. Hibernate will not acquire and release connections for each lazy field loading.

spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaVendorAdapter.java
```diff
@@ -109,6 +109,34 @@ public void setPrepareConnection(boolean prepareConnection) {
...
+	 * specify it yourself).
+	 * <p> Please pay attention, that this setting doesn't affect how Hibernate
+	 * handles connections, which were acquired on-demand, to lazily load
+	 * collections outside of transaction context.
+	 * <p> Specifically, connections, acquired to serialize entities, returned
+	 * by rest controller method will only be closed, after serialization is
+	 * complete. Hibernate will not acquire and release connections for each
+	 * lazy field loading.
+	 */
+	public void setReleaseConnectionAfterTransaction(boolean releaseConnectionAfterTransaction) {
```
mipo256, line 136: I think we can just link the existing `releaseConnectionAfterTransaction` documentation via javadoc tag `@link`

mipo256 (commented):
I think, in general, it makes sense to separate a single commit in this PR onto two separate. Both, I think, make sense.

spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
poxu: Well, if controller method returns @Entity , it is then serialized to JSON with Jackson. To generate JSON, sometimes we need to load LAZY fields. And to load them Hibernate needs to acquire a connection. And that is what I am talking about. That connection will only be released after EntityManager is closed

spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
poxu: No, I am talking about loading LAZY fields.  I wanted to say, that  connection will not be released after loading LAZY field, even if releaseConnectionAfterTransaction is on

spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaDialect.java
```diff
@@ -118,6 +121,34 @@ public void setPrepareConnection(boolean prepareConnection) {
...
+	 * Set, whether to release connection after transaction is commited or rolled
+	 * back. Only works for pre-bound sessions.
+	 * <p> This setting is needed to work around the fact, that it is not possible
+	 * to use {@link org.hibernate.resource.jdbc.spi.PhysicalConnectionHandlingMode#DELAYED_ACQUISITION_AND_RELEASE_AFTER_TRANSACTION}
+	 * handling mode without sacrificing setting non-default isolation levels
+	 * or read-only flag. Releasing connection might prevent connection pool
+	 * starvation if open-in-view is on.
+	 * <p> Default is "false" for backward compatibility. If you turn this flag
+	 * on it still does nothing unless session is pre-bound (most likely if
+	 * open-in-view) is off. If session is pre-bound and the flag is on, then
```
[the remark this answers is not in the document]
poxu: I agree, but PhysicalConnectionHandlingMode is not as simple as it seems. 

One might suppose, that if you don't configure it manually, then Hibernate will use default value. Which is DELAYED_ACQUISITION_AND_RELEASE_AFTER_STATEMENT. But Spring overrides PhysicalConnectionHandlingMode to be DELAYED_ACQUISITION_AND_HOLD, because DELAYED_ACQUISITION_AND_RELEASE_AFTER_STATEMENT breaks read only transactions and custom isolation levels.

So I thought it would be better to explicitly specify Spring default for PhysicalConnectionHandlingMode . If you still think it is sufficient to say that JDBC connection is managed according to PhysicalConnectionHandlingMode, I don't mind, just wanted to give you the whole picture

Issue #34142, opened 2024-12-23
Partially correct pointcut example.


kumaraman528:
Page: 333 (334)

"Any join point (method execution only in Spring AOP) which takes a single parameter, and
where the runtime type of the argument passed has the @Classified annotation:"

The solution in the documentation is:

@args(com.xyz.security.Classified)

However, the actual solution should be:

@execution(* *(..)) && args(arg) && @args(com.xyz.security.Classified).

The example in the documentation does not account for the requirement of a single parameter.

fix- [pr](https://github.com/spring-projects/spring-framework/pull/34143)

snicoll:
Closing as duplicate of #34143 

@kumaraman528 going forward, please do not open an issue if you intend to send a PR.

PR #34143, opened 2024-12-23
Fix documentation for AOP pointcut matching methods with a single @Cl…

kumaraman528:
…asssified annoted argument

fix for [Issue:34142](https://github.com/spring-projects/spring-framework/issues/34142)


snicoll:
Thanks for the suggestion but the bit of the documentation is focused on explaining how `@args` work. It's not mean at describing a fully-working example that you can copy/paste.

PR #34146, opened 2024-12-24
Add error handler for transactional event

hyh1016:
Exceptions propagated from the transactional event listener are handled by `TransactionSynchronizationUtils` and are not rethrown.
However, the lack of a configurable error handler for these exceptions makes it difficult to customize exception handling during transactional event processing.
To address this, I added an abstract error handler that can be defined as follows:

```java
@Component
public abstract class TransactionalEventErrorHandlerImpl extends TransactionalEventErrorHandler {

	// can inject other beans

	@Override
	public void handle(ApplicationEvent event, @Nullable Throwable ex) {
		// custom logic
	}

}
```


snicoll:
@hyh1016 thanks for the suggestion but we can't consider this PR until you add some unit tests that demonstrates what this is meant to fix. Can you please amend your proposal with them?

hyh1016:
@snicoll Okay, thanks for the feedback. I added unit tests to demonstrate how the code I added works.
Also, while writing the unit tests, I found an issue in the existing logic, so I updated the original PR and pushed the changes again.

PR #34175, opened 2024-12-29
Improve readability in SpringServletContainerInitializer with method extraction

nuheajiohc:
While studying servlet containers, I explored how this is implemented in Spring. I realized that extracting this part into a separate method could enhance readability, so I have made this modification.


snicoll:
Thanks but whether this change improves readability is largely subjective. And if it did, then this should be addressed consistently rather than in an isolated place as you suggest.

PR #34181, opened 2024-12-31
Optimize byte array reading

TungSword:
if message length over Integer.MAX_VALUE, the message.getBody().readAllBytes() method will throw OutOfMemoryError. But we can throw this Error in advance. It can reduce message reading.

message.getBody().readAllBytes() method achieve: when reading bytes over  Integer.MAX_VALUE, it will throw OutOfMemoryError. Message is Required array size too large.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/ByteArrayHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/ByteArrayHttpMessageConverter.java
@@ -54,8 +54,10 @@ public class ByteArrayHttpMessageConverter extends AbstractHttpMessageConverter<
 	@Override
 	public byte[] readInternal(Class<? extends byte[]> clazz, HttpInputMessage message) throws IOException {
 		long length = message.getHeaders().getContentLength();
-		return (length >= 0 && length < Integer.MAX_VALUE ?
-				message.getBody().readNBytes((int) length) : message.getBody().readAllBytes());
+		if (length > Integer.MAX_VALUE) {
+			throw new OutOfMemoryError("message size too large");
+		}
+		return message.getBody().readAllBytes()
 	}
 
 	@Override


snicoll:
Thanks for the PR but that's not what the current code does. It attempts to read the byte array using the length if possible for performance reason. If that's not possible then it calls another method that's less performant. Your change would mean that if the size is above 2GB it'd fail immediately. Also, user code should never throw `OutOfMemoryError` as you suggest.

snicoll:
> But, message.getBody().readAllBytes() method will throw OutOfMemoryError. If byte array length is above 2G

Why? I think you're conflating the capacity of an int, and the ability to read data over 2GB. You can see that the length is a `long` and, while it should be very rare, a JVM (64bits that is) can read byte array over 2GB just fine.

TungSword:
> > But, message.getBody().readAllBytes() method will throw OutOfMemoryError. If byte array length is above 2G
> 
> Why? I think you're conflating the capacity of an int, and the ability to read data over 2GB. You can see that the length is a `long` and, while it should be very rare, a JVM (64bits that is) can read byte array over 2GB just fine as long as it has enough memory allocated.

I view the InputStream.java [https://github.com/openjdk/jdk21/blob/master/src/java.base/share/classes/java/io/InputStream.java](url) class, inside this class
readAllBytes() method will call readNBytes() method. It judge in line 419 - 421.
`
                if (MAX_BUFFER_SIZE - total < nread) {
                    throw new OutOfMemoryError("Required array size too large");
                }
`
It will throw OutOfMemoryError Exception.

TungSword:
> > > But, message.getBody().readAllBytes() method will throw OutOfMemoryError. If byte array length is above 2G
> > 
> > 
> > Why? I think you're conflating the capacity of an int, and the ability to read data over 2GB. You can see that the length is a `long` and, while it should be very rare, a JVM (64bits that is) can read byte array over 2GB just fine as long as it has enough memory allocated.
> 
> I view the InputStream.java [https://github.com/openjdk/jdk21/blob/master/src/java.base/share/classes/java/io/InputStream.java](url) class, inside this class readAllBytes() method will call readNBytes() method. It judge in line 419 - 421. `if (MAX_BUFFER_SIZE - total < nread) { throw new OutOfMemoryError("Required array size too large"); }` It will throw OutOfMemoryError Exception.

When InputStream throw OutOfMemoryError Exception. The Programe will spend many time to read 2G data. It will affect interface performance.

bclozel:
I still think we should decline this PR for the following reasons:

1. this completely removes the `readNBytes` variant where we use the known content length in advance to optimize the allocation.
2. In the JDK `MAX_BUFFER_SIZE` isn't `Integer.MAX_VALUE` but actually `Integer.MAX_VALUE - 8`. So Spring would throw the exception in some cases, and the JDK would in some others
3. `OutOfMemoryError` is typically thrown by the code in charge of memory allocation, which is not Spring

Overall I think we would somehow optimize for a rare case, make maintenance more complicated and implement an inconsistent and surprising behavior.

TungSword:
> I still think we should decline this PR for the following reasons:
> 
> 1. this completely removes the `readNBytes` variant where we use the known content length in advance to optimize the allocation.
> 2. In the JDK `MAX_BUFFER_SIZE` isn't `Integer.MAX_VALUE` but actually `Integer.MAX_VALUE - 8`. So Spring would throw the exception in some cases, and the JDK would in some others
> 3. `OutOfMemoryError` is typically thrown by the code in charge of memory allocation, which is not Spring
> 
> Overall I think we would somehow optimize for a rare case, make maintenance more complicated and implement an inconsistent and surprising behavior.

OK. I got it.

PR #34183, opened 2025-01-01
Add support for multidimensional arrays

remeio:
Resolve TODO: `TODO - only handles 2-dimensional arrays`.

For method `org.springframework.beans.AbstractNestablePropertyAccessor#newValue`:
- When arg `type` is `Integer[][][].class`, it will return a 2-dimensional array instance like`[[]]`.

After this PR, this method support handling multidimensional arrays:
- When arg `type` is `Integer[][][].class`, it will return a 3-dimensional array instance like `[[[]]]`.

But actually, **this final behavior is correct before this PR**. For example: 
- Test case `org.springframework.beans.BeanWrapperAutoGrowingTests#getPropertyValueAutoGrow3dArray`  had passed.

So I just fix this method `newValue`.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/AbstractNestablePropertyAccessor.java
@@ -893,16 +893,7 @@ public abstract class AbstractNestablePropertyAccessor extends AbstractPropertyA
 	private Object newValue(Class<?> type, @Nullable TypeDescriptor desc, String name) {
 		try {
 			if (type.isArray()) {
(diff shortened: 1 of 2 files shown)


snicoll:
@remeio thanks for the PR but there are no tests that showcase what the code has changed. We can't consider the PR until you amend it with additional unit tests.

remeio:
@snicoll Hi，I had add the test case for list of 3-dimensional array like `List<Bean[][][]>`. 

I think the default 3-dimensional array should be `[[[]]]`. But it is `[[]]` before this PR.

You can see the test method: `getPropertyValueAutoGrow3dArrayListForDefault3dArray`

Look forward to your reply.

Issue #32676, opened 2024-04-19
Support data binding from request headers


kalgon:
**Affects:** 6.1.5
---
A rest controller with the following method

```java
    @GetMapping("/{path}")
    public void get(Params params, @PathVariable long path, @RequestParam String foo, @RequestHeader String accept) {
        System.out.printf("path: %d%n", path);
        System.out.printf("foo: %s%n", foo);
        System.out.printf("accept: %s%n", accept);
        System.out.printf("params: %s%n", params);
    }
```
```java
    public record Params(@PathVariable long path, @RequestParam String foo, @RequestHeader String accept) {}
```
will produce the following log when called with `/1?foo=FOO`:

```
    path: 1
    foo: FOO
    accept: application/json, text/plain, */*
    params: Params[path=1, foo=FOO, accept=null]
```

`@PathVariable` and `@RequestParam` are correctly used to instantiate the record but the `@RequestHeader` remains null (although it works as a parameter of the controller method)

sbrannen:
Hi @kalgon,

Thanks for creating this issue.

> `@PathVariable` and `@RequestParam` are correctly used to instantiate the record but the `@RequestHeader` remains null (although it works as a parameter of the controller method)

I have confirmed that to be the case.

@rstoyanchev, is it intentional that `@RequestHeader` is not supported for this use case?

rstoyanchev:
This is regular data binding at work, but applied through constructor args, and the `@PathVariable`, `@RequestParam`, and `@RequestHeader` annotations on the record fields play no role whatsoever, so you can drop them unless you actually want them for clarity. 

For data binding we've long supported using both request parameters and path variables. We can add headers to that too to the map of values to bind.

PR #34193, opened 2025-01-04
Keep the - of header on init parameter names, otherwise the nested path will not work.

remeio:
Keep the `-` of header names when init parameter names, otherwise the nested path will not work.

~~@rstoyanchev Hi, for **property name** and **parameter name**, I think they should use the same transformed header name using "camelCase".~~

~~For example, A http request with headers:~~
- `User-Agent`: `something`
- `WWW-Authenticate`: `something`

~~Before this PR,  the method `ExtendedServletRequestValueResolver#initParameterNames` will return a name set:~~
- `UserAgent`,  `WWWAuthenticate`

~~After this PR, the method will return a name set:~~
- `userAgent`, `wWWAuthenticate`

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ExtendedServletRequestDataBinder.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ExtendedServletRequestDataBinder.java
@@ -124,7 +124,7 @@ public class ExtendedServletRequestDataBinder extends ServletRequestDataBinder {
 				String name = names.nextElement();
 				Object value = getHeaderValue(httpRequest, name);
 				if (value != null) {
-					name = StringUtils.uncapitalize(name.replace("-", ""));
+					name = transformedHeaderName(name);
 					addValueIfNotPresent(mpvs, "Header", name, value);
 				}
 			}
@@ -145,6 +145,10 @@ public class ExtendedServletRequestDataBinder extends ServletRequestDataBinder {
 		else {
 			mpvs.addPropertyValue(name, value);
 		}
+	}
+
+	private static String transformedHeaderName(String headerName) {
+		return StringUtils.uncapitalize(headerName.replace("-", ""));
 	}
 
 	private @Nullable Object getHeaderValue(HttpServletRequest request, String name) {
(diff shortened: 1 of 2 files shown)


bclozel:
@remeio Can you elaborate on which use case you're trying to solve here?
Could you share a sample controller and data class to be bound to explain what's missing here?

remeio:
> @remeio Can you elaborate on which use case you're trying to solve here? Could you share a sample controller and data class to be bound to explain what's missing here?

@bclozel Hi, I had changed the content of this PR , you can look the test case.

 I think it's not correct for replace the `-` of header names when init parameter names, because the nested path will not work.

bclozel:
I missed the fact that this was for nested parameters, indeed. Thanks for clarifying that and adding a test case.

bclozel:
I'm not sure about this change after all.

When binding bean properties, the following case will work:

```java
class Person() {

  // will bind HTTP header "First-Name" : "Jane"
  String firstName;

  public String getFirstName() {
    //
  }
  public String setFirstName(String firstName) {
    //
  }
}
```

Constructor binding requires the `@BindParam` for the same behavior

```java
// will bind HTTP header "First-Name" : "Jane"
record Person(@BindParam("First-Name") String firstName) {
}
```

This PR is about supporting the nested case like so:

```
record MyAddress(@BindParam("My-Address") Address myAddress) {
}

// will bind HTTP header "My-Address.Street-Name" : "Spring Street"
record Address(@BindParam("Street-Name") String streetName) {
}
```

[The ABNF of RFC7230](https://datatracker.ietf.org/doc/html/rfc7230#appendix-B) allows "." in header names:

> header-field = field-name ":" OWS field-value OWS
> field-name = token
> token = 1*tchar
> tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
>    "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA

But in practice, I'm wondering if we should rely on "." characters in header names. [Proxies](http://nginx.org/en/docs/http/ngx_http_core_module.html#ignore_invalid_headers) and [CDNs](https://developers.cloudflare.com/rules/transform/request-header-modification/reference/header-format/) might not accept it.

What do you think @rstoyanchev ?

rstoyanchev:
Indeed, in #32676 we didn't mean to make headers a mechanism for binding to any object structure. The main intent was to include headers in data binding, and those are usually simple, non-hierarchical names. 

@remeio is there a specific case with such headers in use that you've come across?

remeio:
> Indeed, in #32676 we didn't mean to make headers a mechanism for binding to any object structure. The main intent was to include headers in data binding, and those are usually simple, non-hierarchical names. It's similar to what we do for path variables.
> 
> @remeio is there a specific case with such headers in use that you've come across?

Hi, There don't have the specific case. I commit this PR just because my code review. In my opinion, I think the handle logic of header and param should be keep the same, it's nice for Spring user, not the coder.

You also can close this PR, it's OK.

Issue #34044, opened 2024-12-06
AbstractHandshakeHandler should support RFC 8441 WebSocket upgrade via CONNECT with  HTTP/2


jazdw:
If the web server advertises that it supports [RFC 8441](https://datatracker.ietf.org/doc/html/rfc8441) via the HTTP/2 `SETTINGS_ENABLE_CONNECT_PROTOCOL` parameter then some web browsers (Chrome, Firefox) will follow RFC 8441 to open a WebSocket over a HTTP/2 stream using the `CONNECT` HTTP method.

The `org.springframework.web.socket.server.support.AbstractHandshakeHandler#doHandshake` method however only supports the `GET` method and will respond with `405 Method Not Allowed.` and log an error message: `Handshake failed due to unexpected HTTP method: CONNECT`

The work around is to disable RFC 8441 in your web server, e.g. with Jetty you can set `org.eclipse.jetty.http2.server.AbstractHTTP2ServerConnectionFactory#setConnectProtocolEnabled` to false (the default is that it is enabled in Jetty 12 EE10).

This problem was observed in Spring 6.2.0.

https://github.com/spring-projects/spring-framework/blob/f8fd6da10b2264ac98ef1735a9205ad5dcc88224/spring-websocket/src/main/java/org/springframework/web/socket/server/support/AbstractHandshakeHandler.java#L172-L179

bclozel:
@jazdw As far as I understand the Servlet spec does not support CONNECT requests. Can you explain what kind of support you're asking here?

jazdw:
> [@jazdw](https://github.com/jazdw) As far as I understand the Servlet spec does not support CONNECT requests. Can you explain what kind of support you're asking here?

I'm asking for Spring to support RFC 8441 WebSocket upgrades. I noticed that out of the box on Jetty 12 EE 10 the WebSocket upgrades didn't work so I opened an issue.

Here's a link to the relevant section of the Servlet spec - https://github.com/markt-asf/servlet-api/blob/master/spec/src/main/asciidoc/servlet-spec-body.adoc#213-additional-methods

[...]

PR #34195, opened 2025-01-04
Add CONNECT constant to HttpMethod

v-perfilev:
While reviewing issue #34044, I noticed that the HttpMethod class does not include a constant for the HTTP method CONNECT, which was introduced in RFC 7231. This PR adds the CONNECT constant to the HttpMethod class and updates the test class HttpMethodTests.

Changes in other files:

- CONNECT method excluded in initAllowedHttpMethods of RequestMappingInfoHandlerMapping and in initAllowHeader of WebContentGenerator.

- CONNECT method is not supported by HttpComponentsClientHttpConnector and JdkClientHttpConnector, so ClientHttpConnectorTests was updated to dynamically skip CONNECT for these connectors.

- CONNECT method is primarily used for establishing tunnels, so it is not relevant for RequestMethod class and was excluded from RequestMethodTests.

These changes prepare the repository for a follow-up PR addressing issue #34044, where a conditional check will be introduced for WebSocket handshakes: if the HTTP version is HTTP/2 the method must be CONNECT, as specified in RFC 8441, otherwise it should be GET.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpMethod.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpMethod.java
@@ -72,6 +72,12 @@ public final class HttpMethod implements Comparable<HttpMethod>, Serializable {
 	public static final HttpMethod DELETE = new HttpMethod("DELETE");
 
 	/**
+	 * The HTTP method {@code CONNECT}.
+	 * @see <a href="https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.6">RFC 9110, section 9.3.6</a>
+	 */
+	public static final HttpMethod CONNECT = new HttpMethod("CONNECT");
+
+	/**
 	 * The HTTP method {@code OPTIONS}.
 	 * @see <a href="https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.7">RFC 9110, section 9.3.7</a>
 	 */
@@ -83,7 +89,7 @@ public final class HttpMethod implements Comparable<HttpMethod>, Serializable {
 	 */
 	public static final HttpMethod TRACE = new HttpMethod("TRACE");
 
-	private static final HttpMethod[] values = new HttpMethod[] { GET, HEAD, POST, PUT, PATCH, DELETE, OPTIONS, TRACE };
+	private static final HttpMethod[] values = new HttpMethod[] { GET, HEAD, POST, PUT, PATCH, DELETE, CONNECT, OPTIONS, TRACE };
 
 
 	private final String name;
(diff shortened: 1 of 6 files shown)


snicoll:
> Additionally the documentation references for existing constants in HttpMethod have been updated to point to the RFC 9110, as the previous links referenced a superseded document.

Please revert, this is unrelated to your proposal of adding `CONNECT`.

v-perfilev:
Hello @snicoll !

Done: the commit with link updates has been reverted as requested. 

1. Would you like me to create a separate PR for this change, or should I leave it for now?
2. Do the same 55-character length rule apply to revert commit messages?

v-perfilev:
@snicoll 
Okay, I got it.
This rule comes from CONTRIBUTING.md:
> Format commit messages using 55 characters for the subject line, 72 characters per line for the description.

bclozel:
@v-perfilev I understand that CONNECT is an HTTP method defined in specs, but is there a use case in Spring Framework where we can send or receive HTTP CONNECT requests? I'm asking because adding it in `HttpMethod` has little value if we're ignoring this value totally in our HTTP support.

v-perfilev:
Hi @bclozel !

I noticed that the CONNECT method was missing while trying to resolve the issue #34044. At the moment, WebSocket handshake over HTTP/2 is not supported in Spring Framework. According to RFC the CONNECT method must be used for HTTP/2 WebSocket handshakes instead of GET. The HttpMethod class is the most logical place to add this constant. Otherwise, we would need to define and use the CONNECT constant directly in AbstractHandshakeHandler in the following way:
```java
private static final String CONNECT_HTTP_METHOD = "CONNECT"; // DEFINITION

@Override
public final boolean doHandshake(ServerHttpRequest request, ServerHttpResponse response,
        WebSocketHandler wsHandler, Map<String, Object> attributes) throws HandshakeFailureException {
    String protocol = request.getHeaders().getFirst("x-http2");
    boolean isHttp2 = "h2".equals(protocol);

    // some code

    if (isHttp2 && CONNECT_HTTP_METHOD != request.getMethod()) { // USAGE
        // some code
    }

    if (!isHttp2 && HttpMethod.GET != request.getMethod()) {
        // some code
    }

    // some code
}
```

This is not the only case where the CONNECT method is used. CONNECT is indispensable for building proxy-based and tunneling architectures, Spring Framework can also be used for.

Finally, having the CONNECT method ensures consistency within the industry standards. Similar rarely used method TRACE is already part of HttpMethod, and libraries like Apache HttpClient and Netty already support the CONNECT method.

Regards,
Vladimir

bclozel:
I'm not sure how we could produce CONNECT requests with `RestTemplate` or `RestClient` as this method is meant to be used by transport protocols, not high level clients.

As for #34044, I've already commented there that I don't think we can use this method in our support as Servlets do not support this HTTP method at the moment.

rstoyanchev:
Thanks for the PR, but we are going to make a more targeted change in `AbstractHandshakeHandler`. It makes sense to support the method of upgrade in RFC 8441, but there is no intent to support HTTP CONNECT more widely.

PR #34207, opened 2025-01-07
Optimization by caching the map of `cache key` and `shouldSkip`

remeio:
I add the cache for method `AspectJAwareAdvisorAutoProxyCreator#shouldSkip`:
- the key of cache: the cache key for the given class and name
- the value of cache:  Boolean value of should skip

And resolve `TODO: Consider optimization by caching the list of the aspect names`.


bclozel:
Thanks for the proposal, but this "TODO" is 17+ years old and this particular point never showed in performance benchmarks. I'll remove this mention on the main branch.

PR #34212, opened 2025-01-08
Clarify that @Primary and @Fallback doesn't affect the sorting of beans

quaff:
Users may misunderstand that `@Primary` will be the first and `@Fallback` will be the last.


snicoll:
> Users may misunderstand that @Primary will be the first and @Fallback will be the last.

They may, but I am not sure why. Ordering is an orthogonal concept so I am not keen to refer to it there.

PR #34222, opened 2025-01-09
Refactor: Use HttpMethod and HttpHeaders instead of literal string.

remeio:
Just refactor.


snicoll:
Thanks for the PR but the constants are public and therefore we can't remove them as you suggest.

PR #34223, opened 2025-01-09
Polish HttpEntityMethodProcessor

remeio:
JUST refactor

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/HttpEntityMethodProcessor.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/HttpEntityMethodProcessor.java
@@ -31,6 +31,8 @@ import org.jspecify.annotations.Nullable;
 
 import org.springframework.core.MethodParameter;
 import org.springframework.core.ResolvableType;
+import org.springframework.http.HttpStatusCode;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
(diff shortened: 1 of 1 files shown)


spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/HttpEntityMethodProcessor.java
```diff
@@ -237,18 +239,18 @@ else if (returnValue instanceof ProblemDetail detail) {
 		}
 
 		if (httpEntity instanceof ResponseEntity<?> responseEntity) {
-			int returnStatus = responseEntity.getStatusCode().value();
-			outputMessage.getServletResponse().setStatus(returnStatus);
-			if (returnStatus == 200) {
+			HttpStatusCode httpStatusCode = responseEntity.getStatusCode();
+			outputMessage.getServletResponse().setStatus(httpStatusCode.value());
+			if (HttpStatus.OK.value() == httpStatusCode.value()) {
```
ciscoo, line 244:
```suggestion
			if (HttpStatus.OK == httpStatusCode) {
```

You can can compare the enum directly.

remeio: Thanks your advice. But:
`DefaultHttpStatusCode` and `HttpStatus`, they implement interface `HttpStatusCode`, use `#value` is correct. Otherwise a `DefaultHttpStatusCode` instance is NEVER `==` the `HttpStatus#OK`.

So I keep my code.

Issue #34227, opened 2025-01-09
Document custom HttpServiceArgumentResolver usage


ciscoo:
When an API requires multiple values, the number of arguments for HTTP service method can be a bit painful. For example:

```java
interface ExampleService {

    @PostMapping(path = "/example", consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
    void example(@RequestBody Object example, @RequestHeader("X-Custom-Requirement") String foo,
                    @RequestHeader("X-Another-Custom-Requirement") String bar,
                    @RequestParam String one,
                    @RequestParam String two,
                    @RequestParam String three);

}
```

The service method can be simplified to:

```java
@PostMapping(path = "/example", consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
void example(@RequestBody Object example, @RequestHeader HttpHeaders httpHeaders, @RequestParam MultiValueMap<String, Object> params);
```

Which is better, but I think it would be easier or nicer even to allow a single argument that encapsulates all of the request details. 

That argument could also override the predefined ones such as the `consumes` or `path` from the exchange annotation. This is more or less what `UriBuilderFactory` does today for the URL.

If I'm not mistaken, `RestTemplate` provided something similar through `RequestEntity`.

Note I know that you can drop down to the 'lower level' `RestClient` to fully customize the request as indicated in [this table](https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#_migrating_from_resttemplate_to_restclient) in the documentation. But providing that control at the HTTP interface level I think will be a welcome addition.

Looking at Framework's code, `HttpServiceArgumentResolver` is responsible for consuming the various annotations and applying those values to `HttpRequestValues`. So a `HttpServiceArgumentResolver` that accepts a `HttpRequestValues` or similar may work I think.

---

My company has variety of modern and legacy APIs and Web services. While we try to conform to some standard when modernizing legacy services or APIs, there still exists APIs that require a lot of information in their request.

To work around this, I started wrapping the request details into an intermediary object and then delegate to the service method. This leads to cleaner code (IMO). For example:

```java
record RequestSpecification(Object body, String param, String header) {}

interface ExampleService {
    @PostMapping(path = "/example", consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
[...]

Issue #33913, opened 2024-11-19
Revisit `HttpHeaders` implementing the `MultiValueMap` API


simonbasle:
See #33823 : the `MultiValueMap` methods that iterate over the whole header collection misalign with some underlying server implementations.

On the other hand having `HttpHeaders` implement `MultiValueMap` brings in some (marginal?) benefits: using `CollectionUtils`, asserting headers with map-specific assertions, putting all the headers into a plain `Map` (`map.putAll(headers)`), etc...

simonbasle:
Keeping track of notable areas where we check for generic `Map` for headers (and now should check for both `Map` and `HttpHeaders`):
 - `RequestHeaderMapMethodArgumentResolver`
 - `RequestHeaderMethodArgumentResolver`

simonbasle:
Notable possible changes that could be backported to 6.2.x:
 - `getOrEmpty` using a `String` key
 - `getOrDefault` (similarly using a `String` key)
 - `addAll(HttpHeaders)`
 - `containsHeader(String headerName)`
 - `containsEntry(String, List<String>)`
 - `putAll(HttpHeaders)`
 - `remove(String)`
 - ...

PR #34229, opened 2025-01-10
Add HttpHeaders argument resolver for HTTP service

quaff:


bclozel:
Thanks for reaching out, but we'd like to discuss first this enhancement proposal.
Similar to #34227, I'm not sure why we should consider this use case for HTTP interfaces.

We already support [`@RequestHeader` for a single header, or `Map`/`MultivalueMap` for many headers](https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-http-interface-method-parameters). The goal of HTTP interface services is to provide a Java interface for developers with a meaningful API. We don't want to turn this feature into another way to call HTTP clients.

Here, I don't see what `HttpHeaders` would bring here, besides getting/setting transport-specific headers or copying entire headers from another call. Both these use cases are wrong in my opinion.

quaff:
> Here, I don't see what `HttpHeaders` would bring here, besides getting/setting transport-specific headers or copying entire headers from another call. Both these use cases are wrong in my opinion.

I agree it's a bit redundant, but I think the semantic is more accurate compare to `@RequestHeader MultiValueMap<String, ?>`, IMO most developers may expect `HttpHeaders` is resolved as well as `HttpMethod`.

bclozel:
I think that this would confuse developers, since `HttpHeaders` contains the entire set of headers to be sent, including "Content-Length", "Content-Type" and more. This begs the question: which headers should be sent as-is or overwritten by the client on the way out?

Thanks for the proposal, but I'm declining this enhancement.

quaff:
> I think that this would confuse developers, since `HttpHeaders` contains the entire set of headers to be sent, including "Content-Length", "Content-Type" and more. 

`@RequestHeader MultiValueMap<String, ?>` may contains those headers also.


> This begs the question: which headers should be sent as-is or overwritten by the client on the way out?

Agree.

ciscoo:
I just noticed this PR on my issue. Note that `@RequestHeader HttpHeaders` works as is today without this PR since it implements `MultiValueMap` which in turn is a `Map`, so it is already handled:


https://github.com/spring-projects/spring-framework/blob/d5da602bc29c3a61bb0b3203414a21447ed0cbe5/spring-web/src/main/java/org/springframework/web/service/invoker/AbstractNamedValueArgumentResolver.java#L86

May or may not work as expected, but it works. 🤷

bclozel:
This won't work in the future, see #33913.

PR #34241, opened 2025-01-11
Improve consistency in StandardMultipartFile implementation

nuheajiohc:
This PR refactors the **StandardMultipartFile** class to enhance consistency

**Changes include:**

1. Removed the `transferTo(Path dest)` method from **StandardMultipartFile**, as it is already provided as a default method in the **MultipartFile** interface.
2. Updated the `transferTo(File dest)` method to use `getInputStream()` for consistency with the **MultipartFile** interface's default `transferTo(Path dest)` implementation.
3. Modified the `isEmpty` and `getBytes` methods to also utilize internal methods (`getSize()` and `getInputStream()` respectively) for better code consistency.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/multipart/support/StandardMultipartHttpServletRequest.java
+++ b/spring-web/src/main/java/org/springframework/web/multipart/support/StandardMultipartHttpServletRequest.java
@@ -21,7 +21,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Serializable;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR but I think the code is fine as it is. The internal methods call what is expected, irrespective of the fact they implement a public contract as well. For instance, we may return a different `InputStream` than the one we'd be using internally to copy to a byte array. 

There are other places in the codebase where we do this. If we decided to change that, then we should be doing that consistently, not in an isolated class.

PR #34242, opened 2025-01-11
Make multiple slash sanitization optional in web util

### LINKED ISSUE #34076: Multiple slash sanitisation should be optional in `UriComponentsBuilder`
The primary reason for this issue is https://github.com/spring-cloud/spring-cloud-gateway/issues/2795 which under the hood uses `UriComponentsBuilder`. See this comment specifically: https://github.com/spring-cloud/spring-cloud-gateway/issues/2795#issuecomment-2527587851 

In the context of https://github.com/spring-projects/spring-framework/issues/14904 and https://github.com/spring-projects/spring-framework/issues/17006 there have been changes made to sanitise multiple consecutive slashes and replace them via a single slash:
https://github.com/spring-projects/spring-framework/blob/78f28fda5481f28e0808cdfd85bd0c29a78f07c0/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java#L881-L891

Technically, there is no restriction on multiple consecutive slashes - those are valid urls, so I think that adding the option of preserving those in `UriComponentsBuilder` would be a good thing, and would solve the mentioned gateway issue. 

What do you think?

vishnugt:
Make multiple slash sanitization optional in UriComponentsBuilder and UrlPathHelper

No breaking changes introduced; users can continue using the default behavior without modification. 
Optional multi-slash sanitization is added to align with section 3.3 of [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt), which permits empty path segments.

Closes gh-34076

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
+++ b/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
@@ -284,7 +284,7 @@ public class UriComponentsBuilder implements UriBuilder, Cloneable {
 	 * @return the URI components
 	 */
 	public UriComponents build() {
-		return build(false);
+		return build(false, true);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


spring-web/src/main/java/org/springframework/web/util/UrlPathHelper.java
```diff
@@ -392,12 +410,16 @@ else if (index1 == requestUri.length()) {
...
 	/**
-	 * Sanitize the given path. Uses the following rules:
+	 * Sanitize the given path if {code shouldSanitizePath()} is true.
+	 * Uses the following rules:
 	 * <ul>
 	 * <li>replace all "//" by "/"</li>
 	 * </ul>
 	 */
-	private static String getSanitizedPath(final String path) {
+	private String getSanitizedPath(final String path) {
```
vishnugt, line 419: This was a **private** static method that had 2 usages in the class, made it non-static to use the flag.
I can abstract the logic into a static method if you think that makes sense.

spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java
```diff
@@ -911,7 +931,7 @@ public void append(String... pathSegments) {
 		}
 
 		@Override
-		public @Nullable PathComponent build() {
+		public @Nullable PathComponent build(boolean sanitizePath) {
```
vishnugt, line 934: In the PathSegmentComponentBuilder:build method, I couldn't come up with any other approach that requires minimal changes while addressing the sanitizePath flag, especially considering that two of the three implementations make use of it. If you have a better solution, I’d be happy to implement it.

bclozel:
Thanks for contributing to Spring Framework.
Unfortunately, #34076 has been declined so I'll decline this PR as well.

PR #34243, opened 2025-01-11
Introduce `@BeanAlias` to support adding bean aliases declaratively outside of bean definitions

TigerZCoder:
Sometimes when you integrate with other projects, you want to use beans created in that project in other code, but you need to follow the naming convention. Or just use the beans you created in other modules to replace beans with the same functionality in another project.
Here I added a @BeanAlias ​​annotation so that simply declaring the bean alias in the configuration can achieve the effect.


snicoll:
Thanks for the suggestion but we'd like to avoid creating an annotation for such advanced use case. `@Bean` does not reflect everything that can be set on a bean definition for the same reason. 

If you find this useful in your project, nothing prevents you from adding that there.

PR #34250, opened 2025-01-13
Add test for DataSourceTransactionManager constructor

kwondh5217:
Hello,

I noticed that the DataSourceTransactionManager constructor was missing unit tests. 
I added a test to ensure that the afterPropertiesSet method throws an IllegalArgumentException when a null DataSource is provided.

Any feedback on this PR would be appreciated.
Thank you!


sbrannen:
Hi @kwondh5217,

Since the `DataSource` in the `DataSourceTransactionManager(DataSource)` is not marked as `@Nullable`, we expect that the caller never supplies `null`, and we don't typically write unit tests for such scenarios.

Thanks anyway!

kwondh5217:
Thank you for the review!
I noticed that afterPropertiesSet() contains code that throws an IllegalArgumentException when the default constructor is used.
Wouldn't it also be worth testing this scenario?

PR #34255, opened 2025-01-13
Extend HTTP status assertions

cms04:
This is an extension for HTTP status assertions in the `AbstractHttpServletResponseAssert`, which was introduced in version 6.2. With this extension, it is possible to write test assertions like this

```java
MvcTestResult result = mockMvcTester.get().uri("/example").exchange();
assertThat(result).httpStatus().isUnauthorized();
```

which is inspired by the "old" `MockMvc` status assertions:

```java
mockMvc.perform(get("/example"))
    .andExpect(status().isUnauthorized());
```

No breaking changes introduced.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/http/HttpStatusAssert.java
+++ b/spring-test/src/main/java/org/springframework/test/http/HttpStatusAssert.java
@@ -1,0 +1,310 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


snicoll:
Thanks for the suggestion @cms04 - We're aware of the lack of support for HttpStatus and did not introduce them on purpose. One reason is to avoid exactly what you're suggesting. By having such assertions there, it breaks the flow if you're trying to assert simple use cases as `httpStatus()` returns a dedicated assert object.

The "root" assertion object has basic `hasStatus` methods. I am still not keen to add more there as it would become quite busy in such a high-level object. Taking a step back, why are such narrowed assertions necessary? `assertThat(result).httpStatus.isUnauthorized()` can be written as `assertThat(result).hasStatus(HttpStatus.UNAUTHORIZED)` right now and I prefer the latter over the former.

snicoll:

PR #34258, opened 2025-01-14
Ignore cache operations if generated key is an empty Optional

### LINKED ISSUE #34257: @Cacheable when keyBuilder return null, then throw error to main thread
https://github.com/spring-projects/spring-data-redis/issues/3089#issue-2782880207
Can we not handle the cache when return is null not throw error to main thread?

quaff:
Closes GH-34257


snicoll:
Please stop opening PRs before triage on the issue took place.

PR #34266, opened 2025-01-15
Refactor: Use HttpStatus.Series instead of magic value.

remeio:
JUST refactor


sdeleuze:
Hi thanks for your contribution. IMO it introduces some unnecessary coupling between `DefaultHttpStatusCode` and `HttpStatus`, and the eager parsing and storing of `HttpStatus.Series` is likely not something we want. As a consequence, I will decline this PR.

PR #34267, opened 2025-01-15
Fix depth for NullAway subsections in reference docs

izeye:
This PR fixes depth for NullAway subsections in reference docs.


sdeleuze:
Merged, thanks!

PR #34281, opened 2025-01-17
Optimization: Sort array when the length is greater than 1

remeio:
Sort array when the length is greater than 1.


bclozel:
Thanks for the contribution but we only tend to accept performance-related changes when there is a clear performance improvement.

Issue #27020, opened 2021-06-04
[Bug] YamlPropertiesFactoryBean incorrect flatten nested map to properties when map key contains escaped brackets


ian4hu:
**Affects:** since 4.1

---

Yaml file like bellow:

```yaml
root:
  webservices:
    "[domain.test:8080]":
      - username: me
        password: mypassword

```

should equals properties file like:

```properties
root.webservices[[domain.test\:8080]][0].username=me
root.webservices[[domain.test\:8080]][0].password=mypassword
```

or:

```properties
root.webservices[[domain.test\:8080]][0].username=me
root.webservices[[domain.test\:8080]][0].password=mypassword
```

Both properties file is parseable for Spring Boot [Binding Maps](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.external-config.typesafe-configuration-properties.relaxed-binding.maps)

ian4hu:
Reproduce Test Case is

```java
	@Test
	void loadNestedEscapedProperties() throws Exception {

		String yaml = "root:\n" +
				"  webservices:\n" +
				"    \"[domain.test:8080]\":\n" +
				"      - username: me\n" +
				"        password: mypassword\n";
		String propsStr = "root.webservices[[domain.test\\:8080]][0].username=me\n" +
				"root.webservices[[domain.test\\:8080]][0].password=mypassword";
		final PropertiesFactoryBean factory = new PropertiesFactoryBean();
		factory.setLocations(new ByteArrayResource(propsStr.getBytes(StandardCharsets.UTF_8)));
		factory.setSingleton(false);
		final Properties properties = factory.getObject();

		final YamlPropertiesFactoryBean yamlFactory = new YamlPropertiesFactoryBean();
		yamlFactory.setResources(new ByteArrayResource(yaml.getBytes(StandardCharsets.UTF_8)));
		assertThat(properties).isEqualTo(yamlFactory.getObject());
	}
```

ian4hu:
See https://github.com/spring-cloud/spring-cloud-config/issues/1595

ian4hu:
related #27022

snicoll:
@ian4hu you said you had some code to attempt to fix this issue. Would you mind sharing that?

PR #34285, opened 2025-01-19
YamlPropertiesFactoryBean incorrect flatten nested map to properties

remeio:
To fix issue #27020.

For the method `org.springframework.beans.factory.config.YamlProcessor#getFlattenedMap` before this PR:
- When the key is start with `[`, will return `path + key` as map key.
- Otherwise, will return `path + '.' + key` as map key.
```java
if (key.startsWith("[")) {
	key = path + key;
}
else {
	key = path + '.' + key;
}
```

For the method `org.springframework.beans.factory.config.YamlProcessor#getFlattenedMap` after this PR:
- When the key is start with `[` or end with `]`, will return `path + '[' + key + ']'` as map key.
- Otherwise, will return `path + '.' + key` as map key.
- And when the key is indexed key at `Collection` like `[0]`, will return `path + key` as map key.

For the method `org.springframework.beans.factory.config.YamlProcessor#asMap` after this PR:
Change `result.put("[" + key.toString() + "]", value);` to `result.put(key.toString(), value);` on `asMap`.

And the test case you can see:
```yaml
root:
  webservices:
    "[domain.test:8080]":
      - username: foo
        password: bar
# Before this PR:
# root.webservices[domain.test:8080][0].username=foo
# root.webservices[domain.test:8080][0].password=bar
# After this PR:
# root.webservices[[domain.test:8080]][0].username=foo
# root.webservices[[domain.test:8080]][0].password=bar

indexed:
  "[0]": foo
  "[1]": bar
indexed2:
  - "[a]": foo
    "[b]": bar
# Before this PR:
# indexed[0]=foo
# indexed[1]=bar
# indexed2[0][a]=foo
# indexed2[0][b]=bar
# After this PR:
# indexed[[0]]=foo
# indexed[[1]]=bar
# indexed2[0][[a]]=foo
# indexed2[0][[b]]=bar

only-left-bracket:
  "[/key1/": foo
only-right-bracket:
  "/key1/]": foo
special-bracket:
  "][/key1/][": foo
# Before this PR:
# only-left-bracket[/key1/=foo
# only-right-bracket./key1/]=foo
# special-bracket.][/key1/][=foo
# After this PR:
# only-left-bracket[[/key1/]=foo
# only-right-bracket[/key1/]]=foo
# special-bracket.][/key1/][=foo


number-key:
  1: foo
# Before this PR:
# number-key[1]=bar
# After this PR:
# number-key.1=foo
```

By the way, `org.springframework.beans.factory.config.YamlMapFactoryBean#getObject` result is:
```
# Before this PR & After this PR:
# root={webservices={[domain.test:8080]=[{username=foo, password=bar}]}}
# indexed={[0]=foo, [1]=bar}
# indexed2=[{[a]=foo, [b]=bar}]
# only-left-bracket={[/key1/=foo}
# only-right-bracket={/key1/]=foo}
# special-bracket={][/key1/][=foo}

# Before this PR:
# number-key={[1]=foo}
# After this PR:
# number-key={1=foo}
```

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java
@@ -56,6 +56,7 @@ import org.springframework.util.StringUtils;
  * @author Juergen Hoeller
  * @author Sam Brannen
  * @author Brian Clozel
+ * @author Mengqi Xu
  * @since 4.1
  */
 public abstract class YamlProcessor {
@@ -305,16 +306,19 @@ public abstract class YamlProcessor {
 	 */
 	protected final Map<String, Object> getFlattenedMap(Map<String, Object> source) {
 		Map<String, Object> result = new LinkedHashMap<>();
-		buildFlattenedMap(result, source, null);
+		buildFlattenedMap(result, source, null, false);
 		return result;
 	}
 
 	@SuppressWarnings({"rawtypes", "unchecked"})
-	private void buildFlattenedMap(Map<String, Object> result, Map<String, Object> source, @Nullable String path) {
+	private void buildFlattenedMap(Map<String, Object> result, Map<String, Object> source, @Nullable String path, boolean isIndexedKey) {
 		source.forEach((key, value) -> {
 			if (StringUtils.hasText(path)) {
-				if (key.startsWith("[")) {
+				if (isIndexedKey) {
 					key = path + key;
+				}
+				else if (key.startsWith("[") || key.endsWith("]")) {
+					key = path + '[' + key + ']';
 				}
 				else {
 					key = path + '.' + key;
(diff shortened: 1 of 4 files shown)


remeio:
This PR just is a try, I submit the test case for `[]`. It's important to explain what result is correct for this test case. Maybe you can fill it with:
```yaml
root:
  webservices:
    "[domain.test:8080]":
      - username: foo
        password: bar
# properties should be?
# map should be?

indexed:
  "[0]": foo
  "[1]": bar
indexed2:
  - "[a]": foo
    "[b]": bar
# properties should be?
# map should be?

only-left-bracket:
  "[/key1/": foo
only-right-bracket:
  "/key1/]": foo
special-bracket:
  "][/key1/][": foo
# properties should be?
# map should be?

number-key:
  1: foo
# properties should be?
# map should be?
```

PR #34295, opened 2025-01-21
Improve Javadoc for `SpringProperties.getFlag()`

remeio:
`getFlag` return `true` when the property is equal, ignoring case, to the string "true", not just "true", "TrUe" also means `true`.

See more at `Boolean#parseBoolean`


sbrannen:
This has been merged in 016a4c4260ee882807be51df89870e1bc7726485 and revised in 6863b904b56993020debfec8b127f241baae074d.

Thanks

Issue #34305, opened 2025-01-22
Constructor binding fails for simple types List/Map/Array and for nested container combinations


VladM-N1:
Hello,

It's first time I'm reporting bug here.

I've found bunch of bugs in recent spring version (6.2.2) related to query parameters. I can see that few of them were already fixed (https://github.com/spring-projects/spring-framework/issues/34121)

Here is 3 more that I've found:

-     param[]=value -> java.lang.NumberFormatException: For input string: ""
-     param[0]=123 -> java.lang.IllegalStateException: No primary or single unique constructor found for class java.lang.Integer
-     param[key]=123 -> java.lang.IllegalStateException: No primary or single unique constructor found for class java.lang.Integer

Thanks for contributing to project and best regards

Code to reproduce:
GH project: https://github.com/VladM-N1/spring-622-bugs/blob/main/src/test/java/org/example/queryparamsbugs/BrokenQueryParamsTest.java

As code:

```java
package org.example.queryparamsbugs;

import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.test.web.reactive.server.WebTestClient;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;
import reactor.core.publisher.Mono;

import java.util.List;
import java.util.Map;

@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
class BrokenQueryParamsTest {

    @Autowired
    private WebTestClient webTestClient;

    @Test
    void unindexedStringArray() {
        final var expected = new ArrayDto(
            List.of("asd"),
            null
        );
       webTestClient.get()
           .uri("/array?stringList[]=asd")
           .exchange()
           .expectStatus()
           .is2xxSuccessful()
           .expectBody(ArrayDto.class)
           .isEqualTo(expected);
    }

    @Test
    void indexedIntArray() {
        final var expected = new ArrayDto(
            List.of("asd"),
            List.of(123)
        );
        webTestClient.get()
            .uri("/array?stringList[0]=asd&intList[0]=123")
            .exchange()
            .expectStatus()
            .is2xxSuccessful()
            .expectBody(ArrayDto.class)
            .isEqualTo(expected);
    }

    @Test
    void mapWithIntValues() {
        final var expected = new MapDto(
            Map.of(
                "key1", 123
            )
        );
[...]

PR #34307, opened 2025-01-22
Polish DataBinder.

remeio:
Use IllegalArgumentException instead of NumberFormatException.

gh-34305.


rstoyanchev (commented):
Thanks for the suggestion, but `NumberFormatException` is a subclass of `IllegalArgumentException`, so it is even more specific. It is also consistent with the exception raised by setter binding.

PR #34311, opened 2025-01-23
HierarchicalUriComponents should format Collection URI var values in query param

remeio:
When expand `/hello?name={name}`:

1. with `uriVariables` map: 
- `name`: `new String[]{"foo", "bar"}`

We will get `/hello?name=foo,bar`.

2. with `uriVariables` map: 
- `name`: `List.of("foo", "bar")`
- Or other `Collection`.

We will get `/hello?name=[foo, bar]`.

This PR will keep the same behavior for array and collection when expanding query parameter.


rstoyanchev:
This was added some time ago in #14346, but indeed no reason it shouldn't work for collections, which otherwise would be formatted with square brackets.

PR #34321, opened 2025-01-25
Polish org.springframework.scheduling.support.CronExpression#isValidExpression

IsAurora6:
The ‘parse’ method called by the ‘isValidExpression’ method already uses ‘Assert.hasLength’ first to check if the input of the parameter being checked is empty or null, so it is not necessary to make the null judgement first in ‘isValidExpression’.


bclozel:
@IsAurora6 please run the build locally before submitting your next PR.

PR #34353, opened 2025-02-01
Minor update in WebSocket STOMP reference documentation

pirocraft:


rstoyanchev (commented):
Thanks for the suggestion, but the way it is phrased currently is as it is intended.

Receiving does not need ordering. It's the subsequent publication (to the clientOutboutChannel) that needs to be orderded.

pirocraft:
Ok, thanks for your answer and your time!
You mentioned the clientOutboutChannel but it is the clientInboundChannel in this case, I presume.
I think that the strange symmetry between the two channels disturbed me.


Anyway it is clearer to me reading the javadoc of `StompEndpointRegistry.setPreserveReceiveOrder(..)`  
```
By default messages sent to the "clientInboundChannel" may be handled in parallel [...]
When this flag is set to true messages within the same session will be sent to the "clientInboundChannel" one at a time in order to preserve the order in which they were received.
```

rstoyanchev:
You're right, it's about inbound messages in which case the suggestion makes more more sense. Apologies for missing that.

PR #34362, opened 2025-02-03
Support RFC 8441 WebSocket upgrade with HTTP/2 CONNECT

### LINKED ISSUE #34044: AbstractHandshakeHandler should support RFC 8441 WebSocket upgrade via CONNECT with  HTTP/2

jazdw:
f477c1653d2d8198bede242f4ea4d4a599727bcc which closed #34044 was an incomplete fix. Further changes are necessary to support HTTP/2 CONNECT WebSocket upgrades (RFC 8441).

I have manually tested this PR and the CONNECT WebSocket upgrade works as expected using Jetty 12 EE10.

Link to RFC:
https://datatracker.ietf.org/doc/html/rfc8441

The RFC explicitly says that the `Sec-WebSocket-Key` header is not required:

> Implementations using this extended CONNECT to bootstrap WebSockets
   do not do the processing of the Sec-WebSocket-Key and Sec-WebSocket-
   Accept header fields of [[RFC6455](https://datatracker.ietf.org/doc/html/rfc6455)] as that functionality has been
   superseded by the :protocol pseudo-header field.

The RFC explicitly says that the `Connection ` and `Upgrade`  headers are not required:

>  [RFC6455] requires the use of Connection and Upgrade header fields
   that are not part of HTTP/2.  They MUST NOT be included in the
   CONNECT request defined here.

## TODO
- Look into if we need to verify that we are using HTTP/2 and verify the `:protocol` pseudo-header
- Fix WebFlux implementation also

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/socket/server/support/HandshakeWebSocketService.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/socket/server/support/HandshakeWebSocketService.java
@@ -205,23 +205,25 @@ public class HandshakeWebSocketService implements WebSocketService, Lifecycle {
 		HttpMethod method = request.getMethod();
 		HttpHeaders headers = request.getHeaders();
 
-		if (HttpMethod.GET != method && CONNECT_METHOD != method) {
+		if (HttpMethod.GET != method && !CONNECT_METHOD.equals(method)) {
 			return Mono.error(new MethodNotAllowedException(
 					request.getMethod(), Set.of(HttpMethod.GET, CONNECT_METHOD)));
 		}
 
-		if (!"WebSocket".equalsIgnoreCase(headers.getUpgrade())) {
-			return handleBadRequest(exchange, "Invalid 'Upgrade' header: " + headers);
-		}
+		if (HttpMethod.GET == method) {
+			if (!"WebSocket".equalsIgnoreCase(headers.getUpgrade())) {
+				return handleBadRequest(exchange, "Invalid 'Upgrade' header: " + headers);
+			}
 
-		List<String> connectionValue = headers.getConnection();
-		if (!connectionValue.contains("Upgrade") && !connectionValue.contains("upgrade")) {
-			return handleBadRequest(exchange, "Invalid 'Connection' header: " + headers);
-		}
+			List<String> connectionValue = headers.getConnection();
+			if (!connectionValue.contains("Upgrade") && !connectionValue.contains("upgrade")) {
+				return handleBadRequest(exchange, "Invalid 'Connection' header: " + headers);
+			}
 
-		String key = headers.getFirst(SEC_WEBSOCKET_KEY);
-		if (key == null) {
-			return handleBadRequest(exchange, "Missing \"Sec-WebSocket-Key\" header");
+			String key = headers.getFirst(SEC_WEBSOCKET_KEY);
+			if (key == null) {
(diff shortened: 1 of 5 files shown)


rstoyanchev:
@jazdw thanks for testing and sending this. Could you check the DCO instructions on how to sign off your commit and update the PR branch? This is required for us to process this.

About checking for HTTP/2, `ServletRequest` has a `getProtocolVersion()` method that we could use, but would have to expose it first through `ServerHttpRequest`. I don't think that's truly necessary.

jazdw:
> About checking for HTTP/2, ServletRequest has a getProtocolVersion() method that we could use, but would have to expose it first through ServerHttpRequest. I don't think that's truly necessary.

I dug into this a little more, Jetty verifies the `:protocol` pseudo header internally, see `org.eclipse.jetty.websocket.core.server.internal.RFC8441Negotiation#validateHeaders`. It also verifies the method is `CONNECT` and the HTTP version is `HTTP/2.0` in `org.eclipse.jetty.websocket.core.server.internal.RFC8441Handshaker#isWebSocketUpgradeRequest`

I am wondering if we should add a `supportsRfc8441()` method to the `RequestUpgradeStrategy`? That way we wouldn't attempt the upgrade if we know the implementation doesn't support `CONNECT`.

rstoyanchev:
I don't mind a dedicated supports check for RFC 8441 as long as it is possible through the Servlet API. Jetty is at a lower level and can perform more checks I suspect.

jazdw:
@rstoyanchev I've pushed another commit, I don't know that it is necessary though. Happy for you to drop the second commit if you are happy with just the first. I've tested it on Jetty 12 EE10 with both GET `setConnectProtocolEnabled(false)` and CONNECT `setConnectProtocolEnabled(true)`

rstoyanchev:
I dropped the second commit. Out handshake checks are there to provide a consistent behavior, but Servlet containers will perform their own checks when upgrading. So we don't need this to be configurable with a flag.

Thanks for the feedback and the changes!

jazdw:
> I dropped the second commit. Out handshake checks are there to provide a consistent behavior, but Servlet containers will perform their own checks when upgrading. So we don't need this to be configurable with a flag.
> 
> Thanks for the feedback and the changes!

No problem, thanks Rossen

PR #34414, opened 2025-02-12
Fix Non-transient non-serializable instance field in serializable class

Joydevelop:
when execute org/springframework/test/web/UriAssertTests#isEqualToTemplate org/springframework/http/ProblemDetail#properties will throw Non-transient non-serializable instance field in serializable class


bclozel:
Thanks, this has been fixed in 5c094358164

Issue #34015, opened 2024-12-04
Task scheduler configured by XML is not eligible for getting processed by all BeanPostProcessors


ProgMiner:
Hello!

Seems like XML-based scheduling configuration from documentation produces well-known "is not eligible" warning.

### Minimal sample

Tested on Spring Core versions: 6.1.15, 6.2.0.

`applicationContext.xml`:
```xml
<beans xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:task="http://www.springframework.org/schema/task"
       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-2.5.xsd http://www.springframework.org/schema/task http://www.springframework.org/schema/task/spring-task-3.0.xsd">
    <task:scheduler id="scheduler" pool-size="5" />
    <task:annotation-driven scheduler="scheduler" />
</beans>
```

Reference: https://docs.spring.io/spring-framework/reference/integration/scheduling.html#scheduling-enable-annotation-support

`Main.java`:
```java
import org.springframework.context.ConfigurableApplicationContext;
import org.springframework.context.support.ClassPathXmlApplicationContext;

public class Main {

    public static void main(String[] args) {
        new ClassPathXmlApplicationContext("applicationContext.xml");
    }
}
```

Starting results in singular log message:
```
WARNING: Bean 'scheduler' of type [org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying). Is this bean getting eagerly injected/applied to a currently created BeanPostProcessor [org.springframework.context.annotation.internalScheduledAnnotationProcessor]? Check the corresponding BeanPostProcessor declaration and its dependencies/advisors. If this bean does not have to be post-processed, declare it with ROLE_INFRASTRUCTURE.
```

Unfortunately, I haven't found any way to mark this bean as infrastructure using only XML.

Related: https://github.com/spring-projects/spring-framework/issues/32055, https://github.com/spring-projects/spring-integration/issues/7602

Best regards!

snicoll:
Thanks for the report, I can reproduce the problem based on the code snippet you've shared. Going forward, it would be very much appreciated to share a small sample we can run ourselves (zip or link to a GitHub project) rather than having to piece things together from code in text.

ProgMiner:
[...]

PR #34429, opened 2025-02-15
Make scheduler bean as a infrastructure bean

remeio:
// gh-34015


jhoeller:
This has been covered in a slightly rather revision. Thanks for the PR, in any case!

PR #34434, opened 2025-02-17
Sync `checkstyle` on `buildSrc`

Pankraz76:
I was wondering why the only existing wildcard import was not flagged by Checkstyle. It turned out that the configuration was out of sync.

Considering the example documentation from Checkstyle itself, the default setup appears to work out of the box without requiring a fully qualified import.  

This can likely be reverted—I just wanted to point this out, as I’ve been using it the same way on other projects as in the example, and it worked effortlessly and identically.

- https://checkstyle.sourceforge.io/checks/imports/illegalimport.html#Example1-config
- https://checkstyle.sourceforge.io/checks/imports

### DIFF
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN" "https://checkstyle.org/dtds/configuration_1_3.dtd">
-<module name="com.puppycrawl.tools.checkstyle.Checker">
+<module name="Checker">
 
 	<!-- Suppressions -->
 	<module name="SuppressionFilter">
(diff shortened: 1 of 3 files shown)


src/checkstyle/checkstyle.xml
```diff
@@ -89,11 +89,9 @@
 		<module name="com.puppycrawl.tools.checkstyle.checks.coding.OneStatementPerLineCheck"/>
 
 		<!-- Imports -->
-		<module name="com.puppycrawl.tools.checkstyle.checks.imports.AvoidStarImportCheck"/>
-		<module name="com.puppycrawl.tools.checkstyle.checks.imports.RedundantImportCheck"/>
-		<module name="com.puppycrawl.tools.checkstyle.checks.imports.UnusedImportsCheck">
-			<property name="processJavadoc" value="true"/>
```
punkratz312, line 95: apply default param: https://checkstyle.sourceforge.io/checks/imports/unusedimports.html#Properties

bclozel:
Thanks @punkratz312 !

punkratz312:
Thank you as well @bclozel

PR #34447, opened 2025-02-19
Improve Javadoc of ObjectProvider to clarify what is unique

quaff:


jhoeller:
I put an explanation at the type-level javadoc instead since uniqueness primarily matters for {@link #getObject()} (or even direct injection point declarations without any `ObjectProvider` involved at all), with `getIfUnique` just being a special-purpose method that turns a non-unique case into a `null` return value. Also, I've worded it a bit differently. Thanks for the pull request, in any case!

PR #34456, opened 2025-02-20
Introduce `getMergedAnnotation()` in `TestContextAnnotationUtils`

quaff:
It's equivalent to findMergedAnnotation() but return MergedAnnotation instead of synthesized annotation.


quaff:
@sbrannen See the actual use case: https://github.com/spring-projects/spring-boot/pull/44354#discussion_r1961464347

PR #34492, opened 2025-02-25
feature: Add LogExecutionTimeAspect to log method execution times

SachinSudhirShinde:
**Overview**

This PR introduces a new aspect LogExecutionTimeAspect and an annotation @LogExecutionTime to facilitate logging method execution times in a Spring application. This feature aims to help developers monitor and optimize the performance of their applications by providing execution time metrics for annotated methods.

This change is backward compatible and does not affect existing functionalities.

**Changes Made**

    **_New Annotation_**: @LogExecutionTime

        Introduced a custom annotation @LogExecutionTime to mark methods whose execution times need to be logged.

    **_Aspect Implementation_**: LogExecutionTimeAspect

        Implemented LogExecutionTimeAspect that intercepts methods annotated with @LogExecutionTime and logs their execution times using System.out.println.

        Added logic to handle custom loggers if provided.

**Tests**:
Added comprehensive tests to verify the functionality of the LogExecutionTimeAspect:

1. Basic Execution Time Logging: Verify that the aspect logs execution time for methods annotated with @LogExecutionTime.
2. Different Method Signatures: Test the aspect with methods that have various signatures, parameters, and return values.
3. Exception Handling: Ensure the aspect logs execution times even when methods throw exceptions.
4. Multiple Annotations: Verify that the aspect works correctly when a method has multiple annotations, including @LogExecutionTime.

**How to Test**
1. Clone the Repository: 
git clone https://github.com/sachinsudhirshinde/spring-framework.git
cd spring-framework
2. git checkout -b my-feature-branch
 3. Run Tests: ./gradlew test

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/LogExecutionTime.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/LogExecutionTime.java
@@ -1,0 +1,43 @@
+package org.springframework.context.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * Annotation to log the execution time of methods.
+ *
+ * When applied to a method, the execution time of that method will be logged.
+ * Optionally, a custom logger method can be provided.
+ *
+ * Usage Example:
+ * {@code
+ * @Service
+ * public class MyService {
+ *     @LogExecutionTime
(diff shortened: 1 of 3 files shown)


bclozel:
Thanks for the proposal.
Logging is probably the canonical first aspect example people te d to write and there is nothing specific to Spring here. Additionally there are lots of opinions here on how this should be applied, whether method parameters should be logged, the log level, whether there should be a time threshold, etc.

I am declining this PR for these reasons. Thanks!

PR #34500, opened 2025-02-26
Simplify if condition in OncePerRequestFilter

ngocnhan-tran1996:


bclozel:
Thanks, but we prefer to handle those cosmetic changes ourselves.

PR #34501, opened 2025-02-27
Polish JdbcTemplate to use getter instead of field

quaff:
1. align with others in the same source file
2. allow subclass to override


bclozel:
Thanks @quaff !

PR #34502, opened 2025-02-27
Fix "rowsExpected" of SqlQuery is unused

quaff:


quaff:
On the other hand, I don't think it's valuable, we should deprecate it for removal.
See https://github.com/spring-projects/spring-framework/compare/main...quaff:patch-106?expand=1

bclozel:
I don't understand the goal of this PR. Maybe create a new issue explaining the problem before suggesting changes?

quaff:
I thought it's obvious, `rowsExpected` is never used, then we have two options:
1. use it for query.
2. deprecate it for removal.

quaff:
> I don't understand the goal of this PR. Maybe create a new issue explaining the problem before suggesting changes?

I created GH-34512 as you requested.

PR #34511, opened 2025-02-28
Adding Null Check for validateIfNoneMatch() method

dk185173:
Basic Null Check to avoid Null pointer exception if the header is not sent

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/context/request/ServletWebRequest.java
+++ b/spring-web/src/main/java/org/springframework/web/context/request/ServletWebRequest.java
@@ -236,6 +236,9 @@ public class ServletWebRequest extends ServletRequestAttributes implements Nativ
 
 	private boolean validateIfNoneMatch(@Nullable String etag) {
 		Enumeration<String> ifNoneMatchHeaders = getRequest().getHeaders(HttpHeaders.IF_NONE_MATCH);
+		if(ifNoneMatchHeaders==null){
+			return false;
+		}
 		if (!ifNoneMatchHeaders.hasMoreElements()) {
 			return false;
 		}


dk185173:
Please merge it so that this header is not mandatory

bclozel:
The [`HttpServletRequest` API](https://jakarta.ee/specifications/servlet/6.1/apidocs/jakarta.servlet/jakarta/servlet/http/httpservletrequest#getHeaders(java.lang.String)) states the following for the `getHeaders` method:

> If the request did not include any headers of the specified name, this method returns an empty Enumeration.

Did you encounter a `NullPointerException` in production? Can you provide a stacktrace?

bclozel:
Thanks for the feedback @dk185173 .

I believe the bug is somewhere else, but not in Spring. Maybe one of your custom filters is not wrapping the request properly and returns a `null` value here where it should return an empty enumeration.

I couldn't reproduce this behavior with a vanilla Spring application:

```java
@Controller
public class TestController {

	@GetMapping("/")
	@ResponseBody
	public String test(ServletWebRequest request) {
		if (request.checkNotModified("test")) {
			return null;
		}
		return "test";
	}
}
```

```
➜  ~ http :8080/ -vv
GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: localhost:8080
User-Agent: HTTPie/3.2.4

HTTP/1.1 200
Connection: keep-alive
Content-Length: 4
Content-Type: text/plain;charset=UTF-8
Date: Fri, 28 Feb 2025 10:36:07 GMT
ETag: "test"
Keep-Alive: timeout=60

test
```

I'm closing this as a result.

dk185173:
@bclozel We encountered this issue after we upgraded from Spring 5->Spring 6. It was working in Spring 5.

bclozel:
I guess all of your application dependencies changed as well, since the Spring 5 -> Spring 6 upgrade requires a javax->jakarta change.

From here, I think you can work on a minimal sample application, starting with the controller I've provided and add things bit by bit, starting witht the Servlet container you are using, then custom filters, etc. The bug is most probably on a Servlet request wrapper implementation. You can also use a debug break point and inspect the `HttpServletRequest` instance you are getting at this point: what's the concrete type, where does it come from, etc.

PR #34526, opened 2025-03-03
Deprecate `rowsExpected` property of `SqlQuery` for removal

### LINKED ISSUE #34512: Deprecate `rowsExpected` property of `SqlQuery`
`SqlQuery::setRowsExpected` is invoked somewhere, but it make no sense since `SqlQuery::getRowsExpected` is never invoked anywhere.

We have two options:
1. use it for query.
2. deprecate it for removal.

See GH-34502

quaff:
Fix GH-34512

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/object/SqlFunction.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/object/SqlFunction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -65,7 +65,6 @@ public class SqlFunction<T> extends MappingSqlQuery<T> {
 	 * @see #compile
 	 */
 	public SqlFunction() {
-		setRowsExpected(1);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


spring-jdbc/src/main/java/org/springframework/jdbc/object/SqlFunction.java
```diff
@@ -65,7 +65,6 @@ public class SqlFunction<T> extends MappingSqlQuery<T> {
 	 * @see #compile
 	 */
 	public SqlFunction() {
-		setRowsExpected(1);
 	}
```
sbrannen, line 68: Please revert all changes to `SqlFunction`.

We will remove these invocations when we remove that setter method in conjunction with #34530.

spring-jdbc/src/test/java/org/springframework/jdbc/object/SqlQueryTests.java
```diff
@@ -168,7 +168,6 @@ void testStringQueryWithResults() throws Exception {
 		given(resultSet.next()).willReturn(true, true, true, false);
 		given(resultSet.getString(1)).willReturn(dbResults[0], dbResults[1], dbResults[2]);
 		StringQuery query = new StringQuery(dataSource, SELECT_FORENAME);
-		query.setRowsExpected(3);
```
sbrannen, line 171: Please revert this for the time being as well, since we'll remove it in conjunction with #34530.

sbrannen (approved):
I requested a few changes, since we plan to split the deprecation and removal across 6.2.4 and 7.0 (see #34530).

Though, there's no need to make the requested changes: I'll simply omit them when merging.

PR #34532, opened 2025-03-04
Ensure that ConversionService.canConvert(Enum) no longer throws an exception

moonfruit:
In my project, I use the following code to find an appropriate method to execute for a given class. It throws an `IllegalArgumentException: The target type java.lang.Enum does not refer to an enum` when there is a more appropriate “method”: `asString(CharSequence)`. 

```java
public class MyTest {

    public static void main(String[] args) throws AccessException {
        EvaluationContext evaluationContext = SimpleEvaluationContext.forReadOnlyDataBinding().build();
        ReflectiveMethodResolver methodResolver = new ReflectiveMethodResolver();
        MethodExecutor asString = methodResolver.resolve(evaluationContext, MyTest.class, "asString",
                Collections.singletonList(TypeDescriptor.valueOf(String.class)));
        TypedValue result = asString.execute(evaluationContext, MyTest.class, "value");
        System.out.println(result);
    }

    public static String asString(Enum<?> left) {
        return left.name();
    }

    public static String asString(CharSequence left) {
        return left.toString();
    }
}

```

After doing some research I realized that the root cause was that `StringToEnumConverterFactory.getConverter(Enum.class)` threw an exception when it couldn't convert and resolve the concrete `Enum`. This directly causes `ConversionService.canConvert(String.class,Enum.class)` to throw an exception. I don't think the method `canConvert()` should throw an exception, it returns `true` if it can convert and `false` if it can't. That's what this method is for, to avoid throwing an exception when calling the `convert` method directly.

So I've modified `StringToEnumConverterFactory` and `IntegerToEnumConverterFactory` a little and hopefully this PR will be accepted.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/ConversionUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/ConversionUtils.java
@@ -21,8 +21,9 @@ import org.jspecify.annotations.Nullable;
 import org.springframework.core.convert.ConversionFailedException;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.core.convert.converter.ConditionalConverter;
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.core.convert.converter.GenericConverter;
-import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 
 /**
@@ -71,13 +72,30 @@ abstract class ConversionUtils {
 		return false;
 	}
 
-	public static Class<?> getEnumType(Class<?> targetType) {
+	public static @Nullable Class<?> getEnumType(Class<?> targetType) {
 		Class<?> enumType = targetType;
 		while (enumType != null && !enumType.isEnum()) {
 			enumType = enumType.getSuperclass();
 		}
-		Assert.notNull(enumType, () -> "The target type " + targetType.getName() + " does not refer to an enum");
 		return enumType;
 	}
 
+	static class NonConvertableToEnum<T> implements Converter<String, T>, ConditionalConverter {
+
+		private final Class<T> targetType;
+
+		public NonConvertableToEnum(Class<T> targetType) {
+			this.targetType = targetType;
+		}
+
+		@Override
+		public boolean matches(TypeDescriptor sourceType, TypeDescriptor targetType) {
+			return false;
+		}
+
+		@Override
+		public @Nullable T convert(String source) {
+			throw new IllegalArgumentException("The target type " + targetType.getName() + " does not refer to an enum");
+		}
+	}
 }
--- a/spring-core/src/main/java/org/springframework/core/convert/support/ConversionUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/ConversionUtils.java
@@ -22,7 +22,6 @@ import org.springframework.core.convert.ConversionFailedException;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.TypeDescriptor;
 import org.springframework.core.convert.converter.GenericConverter;
-import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 
 /**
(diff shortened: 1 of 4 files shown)


moonfruit:
Hi, @sbrannen. Is there something else I should change?

spring-core/src/main/java/org/springframework/core/convert/support/ConversionUtils.java
sbrannen: Since this is an internal, package-private utility, it's OK to change the nullness and behavior, especially since it's only used by `StringToEnumConverterFactory` and `IntegerToEnumConverterFactory`.

However, since we are slightly changing the semantics, I'd likely rename the method to `resolveEnumType(...)` and introduce simple Javadoc to point out that it returns `null` if the enum type cannot be resolved.

sbrannen: We shouldn't need this. Rather, we should make the converter factories implement `ConditionalConverter`.

spring-core/src/main/java/org/springframework/core/convert/support/IntegerToEnumConverterFactory.java
sbrannen:
```suggestion
		Class<?> enumType = ConversionUtils.resolveEnumType(targetType);
		Assert.notNull(enumType, () -> "The target type " + targetType.getName() + " does not refer to an enum");
		return new IntegerToEnum(enumType);
```

Let's try this here as well as in `StringToEnumConverterFactory`.

spring-core/src/test/java/org/springframework/core/convert/support/GenericConversionServiceTests.java
sbrannen:
```suggestion
	@Test  // gh-34532
	void canConvertToEnumDoesNotThrowForNonEnumTargetType() {
		conversionService.addConverterFactory(new StringToEnumConverterFactory());
		conversionService.addConverterFactory(new IntegerToEnumConverterFactory());

		assertThat(conversionService.canConvert(String.class, Enum.class)).isFalse();
		assertThat(conversionService.canConvert(Integer.class, Enum.class)).isFalse();

		assertThat(conversionService.canConvert(String.class, MyEnum.class)).isTrue();
		assertThat(conversionService.canConvert(Integer.class, MyEnum.class)).isTrue();
		assertThat(conversionService.convert("A", MyEnum.class)).isEqualTo(MyEnum.A);
		assertThat(conversionService.convert(0, MyEnum.class)).isEqualTo(MyEnum.A);
	}

```

How about this instead?

sbrannen (changes requested):
Hi @moonfruit,

And thanks for bringing this to our attention.

I agree that `canConvert()` ideally should not throw an exception. To address that, I've requested some changes in the PR (see other comments). In addition, let's have `IntegerToEnumConverterFactory` and `StringToEnumConverterFactory` implement `ConditionalConverter`. Their implementations could then look something like this:

```java
	@Override
	public boolean matches(TypeDescriptor sourceType, TypeDescriptor targetType) {
		return (ConversionUtils.resolveEnumType(targetType.getType()) != null);
	}
```

If you feel comfortable making these changes, go for it!

Otherwise, just let me know, and I can take it from here.

Cheers,

Sam

moonfruit:
Hi @sbrannen,

I've updated the PR with all the requested changes:

- Renamed `ConversionUtils.getEnumType(...)` to `resolveEnumType(...)` and added Javadoc pointing out that it returns `null` if the enum type cannot be resolved.
- Removed the `NonConvertableToEnum` converter. `StringToEnumConverterFactory` and `IntegerToEnumConverterFactory` now implement `ConditionalConverter`, with `matches(...)` implemented as you suggested, and their `getConverter(...)` methods keep the original `Assert.notNull(...)` semantics.
- Applied your suggested test, `canConvertToEnumDoesNotThrowForNonEnumTargetType()`.

`GenericConversionServiceTests` and `checkstyleMain`/`checkstyleTest` pass locally for `spring-core`.

sbrannen (changes requested):
I think this looks good now. Thanks for making the requested changes! 👍 

However, your commit is failing the [DCO check](https://github.com/spring-projects/spring-framework/pull/34532/checks?check_run_id=101605066736). So please address that.

moonfruit:
Sorry about that — the commit was missing the `Signed-off-by` trailer. I have amended the commit to add the sign-off and force-pushed the branch; the DCO check is passing now. No changes were made to the code itself.

sbrannen (approved):

sbrannen:

Thanks

PR #34538, opened 2025-03-05
DefaultSimpUserRegistry Use ReentrantLock to Prevent Pinning

csh0034:
modified the part where pinning occurs in DefaultSimpUserRegistry due to the use of synchronized when using VT
I would appreciate any feedback 😃

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/messaging/DefaultSimpUserRegistry.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/messaging/DefaultSimpUserRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,8 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


spring-websocket/src/main/java/org/springframework/web/socket/messaging/DefaultSimpUserRegistry.java
```diff
@@ -110,7 +113,9 @@ else if (event instanceof SessionConnectedEvent) {
 			if (user instanceof DestinationUserNameProvider destinationUserNameProvider) {
 				name = destinationUserNameProvider.getDestinationUserName();
 			}
-			synchronized (this.sessionLock) {
+
+			this.sessionLock.lock();
```
pankratz76, line 117:
```suggestion
		sessionLock.lock();
  ```
There is no overloaded field that makes the `this.` access required, right? Both are correct, but if there is no ambiguity, it's more concise to leave it out.

**Review by Pankraz76 — APPROVED** [review]
nicely done

File: spring-websocket/src/main/java/org/springframework/web/socket/messaging/DefaultSimpUserRegistry.java (hunk 1 above)
Comment (pankratz76) [inline spring-websocket/src/main/java/org/springframework/web/socket/messaging/DefaultSimpUserRegistry.java:117]:
```suggestion
			sessionLock.lock();
```
There is no overloaded field that makes the `this.` access required, right? Both are correct, but if there is no ambiguity, it's more concise to leave it out.

csh0034: Thank you for your valuable feedback.
I actually considered this when making the code changes.

Looking at other classes that use ReentrantLock, I noticed that they all include `this.`
so I decided to keep it for consistency. 
(e.g., `SubProtocolWebSocketHandler`, `InMemoryWebSessionStore.ExpiredSessionChecker`)

jhoeller:
Are you experiencing any actual pinning there? The code within those `synchronized` blocks is not performing any blocking operations as far as I can see. From that perspective, it should be fine as-is.

csh0034:
@jhoeller 
I modified the code based on an article suggesting replacing synchronized with ReentrantLock to prevent VT pinning in Spring Blog. If there are no blocking operations and it doesn't cause any issues, feel free to close the PR.

jhoeller:
Alright, let's keep it as-is then since we keep using such synchronization around local data structure operations in other places as well.

Thanks for your efforts, in any case!

PR #34542, opened 2025-03-05
Allow `ContentResultMatchersDsl` matchers for supertypes of the checked type

### LINKED ISSUE #34494: ContentResultMatchersDsl's matchers too strict
In `ContentResultMatchersDsl`, the functions that accept a `Matcher` don't allow matchers for supertypes of the checked type. This causes code like

```
    mockMvc
        .get("/books/{bookId}/blurb", BOOK_ID)
        .andExpect { status { isOk() } }
        .andExpect { content { string(hasLength(145)) } }
```

to not compile, since `hasLength` is a `Matcher<CharSequence>` and `string` wants a `Matcher<String>`. The original method does accept a `Matcher<? super String>`; in this case, changing the `string` method to accept a `Matcher<in String>` would also solve this compilation error.

The other functions that accept a matcher (`node` and `source`) exhibit the same problem.

dmitrysulman:
This PR allows accepting `Matcher<CharSequence>` for  [ContentResultMatchersDsl#string](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/main/kotlin/org/springframework/test/web/servlet/result/ContentResultMatchersDsl.kt#L74).

**Before:**
```kotlin
fun string(matcher: Matcher<String>)
```

**After:**
```kotlin
fun string(matcher: Matcher<in String>)
```

Fixes spring-projectsgh-34494

### DIFF
--- a/spring-test/src/main/kotlin/org/springframework/test/web/servlet/result/ContentResultMatchersDsl.kt
+++ b/spring-test/src/main/kotlin/org/springframework/test/web/servlet/result/ContentResultMatchersDsl.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


sdeleuze:
Merged, thanks.

PR #34555, opened 2025-03-07
Propagate `CoroutineContext` to the `CoExchangeFilterFunction`

dmitrysulman:
This PR introduces a new `WebClient.ResponseSpec.awaitEntityOrNull()` extension function to replace `WebClient.ResponseSpec.toEntity(...).awaitFirstOrNull()` and pass the `CoroutineContext` from the outer scope to the `CoExchangeFilterFunction#filter` method.

`CoroutineContext` propagation is implemented via `ReactorContext` and `ClientRequest` attribute.

Any feedback is more than welcome.

Fixes #32148

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
@@ -71,6 +71,9 @@ import org.springframework.web.util.UriBuilderFactory;
  */
 final class DefaultWebClient implements WebClient {
 
+	// Copy of CoExchangeFilterFunction.COROUTINE_CONTEXT_ATTRIBUTE value to avoid compilation errors in Eclipse
+	private static final String COROUTINE_CONTEXT_ATTRIBUTE = "org.springframework.web.reactive.function.client.CoExchangeFilterFunction.context";
+
 	private static final String URI_TEMPLATE_ATTRIBUTE = WebClient.class.getName() + ".uriTemplate";
 
 	private static final Mono<ClientResponse> NO_HTTP_CLIENT_RESPONSE_ERROR = Mono.error(
@@ -430,6 +433,8 @@ final class DefaultWebClient implements WebClient {
 				if (filterFunctions != null) {
 					filterFunction = filterFunctions.andThen(filterFunction);
 				}
+				contextView.getOrEmpty(COROUTINE_CONTEXT_ATTRIBUTE)
+						.ifPresent(context -> requestBuilder.attribute(COROUTINE_CONTEXT_ATTRIBUTE, context));
 				ClientRequest request = requestBuilder.build();
 				observationContext.setUriTemplate((String) request.attribute(URI_TEMPLATE_ATTRIBUTE).orElse(null));
 				observationContext.setRequest(request);
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/CoExchangeFilterFunction.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/CoExchangeFilterFunction.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,9 +17,13 @@
 package org.springframework.web.reactive.function.client
 
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.reactor.awaitSingle
 import kotlinx.coroutines.reactor.mono
 import reactor.core.publisher.Mono
+import kotlin.coroutines.CoroutineContext
+import kotlin.jvm.optionals.getOrNull
 
 /**
  * Kotlin-specific implementation of the [ExchangeFilterFunction] interface
@@ -31,10 +35,14 @@ import reactor.core.publisher.Mono
 abstract class CoExchangeFilterFunction : ExchangeFilterFunction {
 
 	final override fun filter(request: ClientRequest, next: ExchangeFunction): Mono<ClientResponse> {
-		return mono(Dispatchers.Unconfined) {
+		val context = request.attribute(COROUTINE_CONTEXT_ATTRIBUTE).getOrNull() as CoroutineContext?
+		return mono(context ?: Dispatchers.Unconfined) {
 			filter(request, object : CoExchangeFunction {
 				override suspend fun exchange(request: ClientRequest): ClientResponse {
-					return next.exchange(request).awaitSingle()
+					val newRequest = ClientRequest.from(request)
+						.attribute(COROUTINE_CONTEXT_ATTRIBUTE, currentCoroutineContext().minusKey(Job.Key))
+						.build()
+					return next.exchange(newRequest).awaitSingle()
 				}
 			})
 		}
@@ -58,6 +66,17 @@ abstract class CoExchangeFilterFunction : ExchangeFilterFunction {
 	 * @return the filtered response
 	 */
 	protected abstract suspend fun filter(request: ClientRequest, next: CoExchangeFunction): ClientResponse
+
+	companion object {
+
+		/**
+		 * Name of the [ClientRequest] attribute that contains the
+		 * [kotlin.coroutines.CoroutineContext] to be passed to the
+		 * [CoExchangeFilterFunction.filter].
+		 */
+		@JvmField
+		val COROUTINE_CONTEXT_ATTRIBUTE = CoExchangeFilterFunction::class.java.name + ".context"
+	}
 }
 
 
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
@@ -20,17 +20,18 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.reactive.asFlow
-import kotlinx.coroutines.reactor.asFlux
-import kotlinx.coroutines.reactor.awaitSingle
-import kotlinx.coroutines.reactor.awaitSingleOrNull
-import kotlinx.coroutines.reactor.mono
+import kotlinx.coroutines.reactor.*
+import kotlinx.coroutines.withContext
 import org.reactivestreams.Publisher
 import org.springframework.core.ParameterizedTypeReference
 import org.springframework.http.ResponseEntity
+import org.springframework.web.reactive.function.client.CoExchangeFilterFunction.Companion.COROUTINE_CONTEXT_ATTRIBUTE
 import org.springframework.web.reactive.function.client.WebClient.RequestBodySpec
 import org.springframework.web.reactive.function.client.WebClient.RequestHeadersSpec
 import reactor.core.publisher.Flux
 import reactor.core.publisher.Mono
+import reactor.util.context.Context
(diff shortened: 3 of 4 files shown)


spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/CoExchangeFilterFunction.kt
```diff
@@ -31,10 +35,14 @@ import reactor.core.publisher.Mono
 abstract class CoExchangeFilterFunction : ExchangeFilterFunction {
 
 	final override fun filter(request: ClientRequest, next: ExchangeFunction): Mono<ClientResponse> {
-		return mono(Dispatchers.Unconfined) {
+		val context = request.attribute(COROUTINE_CONTEXT_ATTRIBUTE).getOrNull() as CoroutineContext?
+		return mono(context ?: Dispatchers.Unconfined) {
 			filter(request, object : CoExchangeFunction {
 				override suspend fun exchange(request: ClientRequest): ClientResponse {
-					return next.exchange(request).awaitSingle()
```
sdeleuze, line 37: Could you please double check this specific change is required? I am able to get full green tests when I keep `return next.exchange(request).awaitSingle()`, so please either refine the tests or just keep `next.exchange(request).awaitSingle()` here if that's enough.

sdeleuze (changes requested):

spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
```diff
@@ -58,6 +58,8 @@
 import org.springframework.web.util.UriBuilder;
 import org.springframework.web.util.UriBuilderFactory;
 
+import static org.springframework.web.reactive.function.client.CoExchangeFilterFunction.COROUTINE_CONTEXT_ATTRIBUTE;
```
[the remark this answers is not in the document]
dmitrysulman: Not sure if this Eclipse issue is still present.

spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/CoExchangeFilterFunction.kt
dmitrysulman: Added an additional test for context propagation between multiple `CoExchangeFilterFunction`.

spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java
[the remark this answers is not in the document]
sbrannen: Yes, that's problematic in Eclipse IDE.

```
The import org.springframework.web.reactive.function.client.CoExchangeFilterFunction cannot be resolved

DefaultWebClient.java

/spring-webflux/src/main/java/org/springframework/web/reactive/function/client

line 61
```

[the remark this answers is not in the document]
dmitrysulman: @sbrannen Thank you for checking.
@sdeleuze Fixed.

spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
```diff
@@ -20,17 +20,17 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.reactive.asFlow
-import kotlinx.coroutines.reactor.asFlux
-import kotlinx.coroutines.reactor.awaitSingle
-import kotlinx.coroutines.reactor.awaitSingleOrNull
-import kotlinx.coroutines.reactor.mono
+import kotlinx.coroutines.reactor.*
```
pankratz76, line 23: maybe consider avoiding wildcard import.

In checkstyle there is `AvoidStarImport` in place.

https://github.com/spring-projects/spring-framework/blob/7d5b3892c48c57f5d05d3f2283f2d2d0dcd9cab5/buildSrc/config/checkstyle/checkstyle.xml#L18

I would expect the build to fail in this case.

Pankraz76 (approved):
nicely done

spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
dmitrysulman: Thanks for the feedback. 

There are multiple Kotlin files with wildcard import in the Spring Framework codebase, for example:
https://github.com/spring-projects/spring-framework/blob/7d5b3892c48c57f5d05d3f2283f2d2d0dcd9cab5/spring-test/src/main/kotlin/org/springframework/test/web/reactive/server/WebTestClientExtensions.kt#L22

As far as I know Checkstyle doesn't support Kotlin source files therefore the build is successful.

spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
```diff
@@ -203,3 +203,18 @@ inline fun <reified T : Any> WebClient.ResponseSpec.toEntityList(): Mono<Respons
...
+
+/**
+ * Extension for [WebClient.ResponseSpec.toEntity] providing a `toEntity<Foo>()` variant
+ * leveraging Kotlin reified type parameters and allows [kotlin.coroutines.CoroutineContext]
+ * propagation to the [CoExchangeFilterFunction]. This extension is not subject to type erasure
+ * and retains actual generic type arguments.
+ *
+ * @since 7.0.0
+ */
+suspend inline fun <reified T : Any> WebClient.ResponseSpec.awaitEntityOrNull(): ResponseEntity<T>? {
```
sdeleuze, line 215: The tests seems to pass with:
```kotlin
suspend inline fun <reified T : Any> WebClient.ResponseSpec.awaitEntity(): ResponseEntity<T> {
	val context = currentCoroutineContext().minusKey(Job.Key)
	return withContext(Context.of(COROUTINE_CONTEXT_ATTRIBUTE, context).asCoroutineContext()) { toEntity(T::class.java).awaitSingle() }
}
```
Would that simpler variant be ok from your POV?

sdeleuze, line 215: `Entity` related methods either return a value or throw an exception, so I think this should be `awaitEntity(): ResponseEntity<T>`.

sdeleuze, line 215: Shouldn't `awaitExchange`, `awaitExchangeOrNull`, `awaitBody` and `awaitBodyOrNull` be refined to manage `COROUTINE_CONTEXT_ATTRIBUTE` as well? If so, could you please add also related tests?

sdeleuze (changes requested):
Thanks for refining the PR, please check my latest additional feedback.

spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
dmitrysulman: Good point, will fix it

dmitrysulman: It totally makes sense, I will refine these methods and add tests.

dmitrysulman: @sdeleuze It's a little bit tricky.

The current `CoroutineContext` already contains a `ReactorContext` element with some propagated Reactor `Context`. The Reactor `Context`, for example, contains an `Observation` added by the [HttpWebHandlerAdapter](https://github.com/spring-projects/spring-framework/blob/d1a13642316c1dd937f55f0fd08814d4fc697fea/spring-web/src/main/java/org/springframework/web/server/adapter/HttpWebHandlerAdapter.java#L372-L375): 

Screenshot 2025-03-24 at 18 27 03


So, using `withContext(Context.of(COROUTINE_CONTEXT_ATTRIBUTE, context).asCoroutineContext())` will override the current `ReactorContext`, which will break observation in `WebClient`.

sdeleuze: Make sense, thanks for the feedback.

spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/CoExchangeFilterFunction.kt
sdeleuze: Would that make sense to do something like below to avoid mutating the request when unnecessary? 
```kotlin
return if (context == null) {
						next.exchange(request).awaitSingle()
					}
					else {
						val newRequest = ClientRequest.from(request)
							.attribute(COROUTINE_CONTEXT_ATTRIBUTE, currentCoroutineContext().minusKey(Job.Key))
							.build()
						return next.exchange(newRequest).awaitSingle()
					}
```

dmitrysulman: This is interesting. The only possible case where `context == null` is when there is no outer coroutine at all. However it's still possible to have multiple `CoExchangeFilterFunction` and `CoroutineContext` propagation between them won't work with the suggested fix. The question is do we want to support it?

Here is a test that won't pass if we make this change:

```kotlin
	@Test
	fun `ResponseSpec#awaitEntity with coroutine context propagation to multiple CoExchangeFilterFunctions`() {
		val exchangeFunction = mockk<ExchangeFunction>()
		val mockResponse = mockk<ClientResponse>()
		val mockClientHeaders = mockk<ClientResponse.Headers>()
		val foo = mockk<Foo>()
		val slot = slot<ClientRequest>()
		every { exchangeFunction.exchange(capture(slot)) } returns Mono.just(mockResponse)
		every { mockResponse.statusCode() } returns HttpStatus.OK
		every { mockResponse.headers() } returns mockClientHeaders
		every { mockClientHeaders.asHttpHeaders() } returns HttpHeaders()
		every { mockResponse.bodyToMono(Foo::class.java) } returns Mono.just(foo)
		val responseEntity = WebClient.builder()
			.exchangeFunction(exchangeFunction)
			.filter(object : CoExchangeFilterFunction() {
				override suspend fun filter(request: ClientRequest, next: CoExchangeFunction): ClientResponse {
					return withContext(FooContextElement(foo)) { next.exchange(request) }
				}
			})
			.filter(object : CoExchangeFilterFunction() {
				override suspend fun filter(request: ClientRequest, next: CoExchangeFunction): ClientResponse {
					assertThat(currentCoroutineContext()[FooContextElement.Key]!!.foo).isEqualTo(foo)
					return next.exchange(request)
				}
			})
			.build().get().uri("/path").retrieve().toEntity(Foo::class.java)
			.block(Duration.ofSeconds(10))
		val capturedContext = slot.captured.attribute(COROUTINE_CONTEXT_ATTRIBUTE).get() as CoroutineContext
		assertThat(capturedContext[FooContextElement.Key]!!.foo).isEqualTo(foo)
		assertThat(responseEntity!!.body).isEqualTo(foo)
	}
```

spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/client/WebClientExtensions.kt
dmitrysulman: Fixed

dmitrysulman: Done

dmitrysulman: Also refined `awaitBodilessEntity`

sdeleuze (approved):
Looks good, thanks for taking in account the feedback, I will merge it shortly.

dmitrysulman:
@sdeleuze Thanks for your feedback and guidance, I really appreciate it.

PR #34557, opened 2025-03-09
Allow chained BeanRegistry registration

philwebb:
Whilst looking at the new `BeanRegistry` API it struck me that it might be useful to allow several register calls to be chained together. For example, I might have some custom utility that is itself implements `BeanRegistrar`. If we add a `BeanRegistry.register(BeanRegistrar registar)` method I think it will help user to compose registartions.

E.g.:

```java
class MyBeanRegistrar implements BeanRegistrar {

    void register(BeanRegistry registry, Environment env) {
        registry.register(MyCustomScannerThatIsAlsoABeanRegistrar.scan("some.package.for.things.i.know"));
    }

}
```


sdeleuze:
Thanks Phil, merged, looks like a useful additional capability.

PR #34573, opened 2025-03-11
Refactor ContentDisposition type check

Tussle0410:
I noticed that there was duplicated logic when checking the definitions of DispositionType in ContentDisposition.java, so I extracted this logic into a private method.

After that, I realized there weren't any tests to verify this code was working properly, so I went ahead and added test code for it.
```
    public boolean isAttachment() {
	return (this.type != null && this.type.equalsIgnoreCase("attachment"));
    }
    ...
```
I've modified the existing code to follow the format shown below.
```
    public boolean isAttachment() {
	return isDispositionType("attachment");
    }

    private boolean isDispositionType(String type) {
	return this.type != null && this.type.equalsIgnoreCase(type);
    }

```

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
+++ b/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
@@ -99,7 +99,7 @@ public final class ContentDisposition {
 	 * @since 5.3
 	 */
 	public boolean isAttachment() {
-		return (this.type != null && this.type.equalsIgnoreCase("attachment"));
+		return isDepositionType("attachment");
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/ContentDisposition.java
```diff
@@ -555,6 +555,10 @@ private static char hexDigit(int b) {
 		return Character.toUpperCase(Character.forDigit(b & 0xF, 16));
 	}
 
+	private boolean isDepositionType(String type) {
```
bclozel, line 558: Is this a typo? Did you mean `isDispositionType`?

Tussle0410: I made a mistake. There's a typo. I meant isDispositionType. I'll fix it.

bclozel:
Thanks for your contribution @Tussle0410 , this is now merged in the main branch.

Issue #25838, opened 2020-09-29
AbstractJackson2Decoder.decodeToMono should use non-blocking parser


philsttr:
`AbstractJackson2Decoder.decodeToMono` [currently uses `DataBufferUtils.join` to join all DataBuffers](https://github.com/spring-projects/spring-framework/blob/a6bede22c113e69206c13735dbf820c7cb043e2e/spring-web/src/main/java/org/springframework/http/codec/json/AbstractJackson2Decoder.java#L158), [converts the result to an input stream, and then uses jackson's blocking parser](https://github.com/spring-projects/spring-framework/blob/a6bede22c113e69206c13735dbf820c7cb043e2e/spring-web/src/main/java/org/springframework/http/codec/json/AbstractJackson2Decoder.java#L168).   This approach reads all DataBuffers for the entire payload into memory at the same time.  For large payloads, this is not memory efficient.

On the other hand, `AbstractJackson2Decoder.decode` uses jackson's non-blocking parser.  This approach does not require reading all buffers into memory at the same time.  i.e. each buffer can be processed individually, and released as it is parsed.

For larger payloads, I think it would be better if `decodeToMono` also used jackson's non-blocking parser.  This would allow each data buffer to be processed individually and released, rather than having to read the entire payload into memory.

Using the non-blocking parser would allow the `maxInMemorySize` to be increased to accommodate occasional large payloads, and allow the server to be more memory efficient when reading them.

It would also bring the behavior of JSON parsing in WebFlux more inline with the behavior of JSON parsing in WebMVC.  WebMVC's `AbstractJackson2HttpMessageConverter` does not read the entire payload into memory and does not enforce a `maxInMemorySize`.

FWIW, I started looking into this when I hit a `DataBufferLimitException` for some occasional large payloads.  I'd like to increase the `maxInMemorySize`, but then I noticed the consequence of loading everything into memory at the same time, which was a bit surprising since WebMVC doesn't have this limit.

What are your thoughts?

rstoyanchev:
The non-blocking parser would parse and release buffers but only to build a parsed representation that will accumulate until the complete value for the Mono is fully aggregated. So I don't think it will make a difference since either way we're buffering the raw bytes or the parsed representation. 

[...]

PR #34586, opened 2025-03-13
AbstractJackson2Decoder#decodeToMono support non-blocking parsing

ShenFeng312:
see #25838


bclozel:
Have you tried building locally before submitting a PR? This doesn't pass the build.

ShenFeng312:
@bclozel  I'm sorry, I only ran JSON-related test cases locally and overlooked the implementation of `Jackson2CborDecoder` I will fix it and perform a complete local build.

bclozel:
I don't think we should consider this implementation before fully understanding the performance tradeoffs as discussed in https://github.com/spring-projects/spring-framework/issues/25838#issuecomment-702069362. This change could very well decrease throughput significantly for small JSON responses (which is probably the main case when deserializing a single object).

Thanks for your contribution, but we need to first get to the bottom of the performance tradeoffs before considering a concrete implementation.

PR #34587, opened 2025-03-13
add `ObjectUtils.getNonNull/getIfNull/getIfEmpty`

Pankraz76:
apply: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ObjectUtils.html#defaultIfNull(T,T)

to (maybe) use here: https://github.com/spring-projects/spring-framework/pull/34537#pullrequestreview-2679655042

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ObjectUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ObjectUtils.java
@@ -127,7 +127,7 @@ public abstract class ObjectUtils {
 	 * @return An Optional containing the first non-null object, or an empty Optional if all objects are null.
 	 */
 	@SafeVarargs
-	public static <T> Optional<T> getNonNull(@Nullable final T... objects) {
+	public static <T> Optional<T> firstNonNull(@Nullable final T... objects) {
 		return Optional.ofNullable(objects)
 				.map(arr -> Arrays.stream(arr).filter(Objects::nonNull).findFirst())
 				.orElse(Optional.empty());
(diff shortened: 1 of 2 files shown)


pankratz76:
is this any good @currenjin @bclozel ?

currenjin (approved):
Wow, this is amazing! Thank you for going above and beyond by adding the `getIfNull` utility method. This is exactly what we needed for more concise code.

I really appreciate your dedication to improving the codebase and making this PR even better. The method looks well-designed with good documentation and examples.

This demonstrates what makes the Spring community so great - collaborative improvement and attention to detail. I'll definitely use this in my implementation to make the code more elegant.

Thank you again for your exceptional contribution!

spring-core/src/main/java/org/springframework/util/ObjectUtils.java
```diff
@@ -118,6 +118,49 @@ public static boolean isEmpty(@Nullable Object @Nullable [] array) {
...
+
+	/**
+	 * Returns the first non-null element from the provided varargs.
+	 *
+	 * @param objects The objects to check for non-null values.
+	 * @param <T>     The type of the objects.
+	 * @return An Optional containing the first non-null object, or an empty Optional if all objects are null.
+	 */
+	@SafeVarargs
+	public static <T> Optional<T> getNonNull(@Nullable final T... objects) {
```
pankratz76, line 130: this would have the bigger impact imho @currenjin

bclozel:
Thanks for your enthusiasm, but we are limiting the scope of our `*Utils` classes on purpose. Our goal is to maintain a very limited number of methods that are widely useful in the Spring Framework codebase and do not intend to be a shared utility library like Apache commons. Those methods are public because they need to be used from various places in our codebase, but we don't think that application code should depend on it.

Ultimately, adding/removing methods in Utils classes is mainly a Framework team call. We know the community is using those anyway, but whenever we get contributions for performance improvements, we always consider the use cases in Spring Framework and ignore others. In that sense, Apache commons and similar libraries are a much better fit.

I would suggest focusing on the original PR in this case.
Thanks!

Issue #34584, opened 2025-03-13
FormHttpMessageConverter should throw more a more specific throwable when it encounters obviously malformed data.


Helmsdown:
Let's say I have a Spring Boot service, that is picking up `org.springframework.boot.autoconfigure.web.servlet.WebMvcAutoConfiguration`. This means that by default I will get `org.springframework.boot.web.servlet.filter.OrderedFormContentFilter` installed. An arguably good default behavior. The implementation of the filter ends up using `org.springframework.http.converter.FormHttpMessageConverter` to parse the incoming request body. Good so far.

The implementation of `org.springframework.http.converter.FormHttpMessageConverter#read` leverages `java.net.URLDecoder#decode(java.lang.String, java.nio.charset.Charset)`. Again why would you want to reimplement URL decoding functionality for no good reason?

Now imagine you work for a large engineering organization that has at least a thousand developers with 3+ thousand Spring Boot services. Each one has this filter installed doing an arguably good job of parsing HTTP form data.

But then one day, someone in the Security org is doing their job facilitating and pen-test across your entire fleet of services (many of which are Spring Boot services). The pen-testers are crafty folks. They like to contrive all sorts of naughty request payloads and URLs and try them on each and every one of your services. When they hit on a Spring Boot service, with a request payload that is obviously malicious `URLDecoder.decode` is going to do the following:
```
        while (((i + 2) < numChars) &&
               (c == '%')) {
            int v = Integer.parseInt(s, i + 1, i + 3, 16);
            if (v < 0)
                throw new IllegalArgumentException(
                        "URLDecoder: Illegal hex characters in escape "
                        + "(%) pattern - negative value");
            bytes[pos++] = (byte) v;
            i += 3;
            if (i < numChars)
                c = s.charAt(i);
        }
```

The good news is that the patently bad request has been rejected. The bad news is that `IllegalArgumentException` has been thrown, at a layer in your Spring Boot app that is likely before any custom exception handling is invoked. So the exception is going to ride on up to web server layer (e.g. Tomcat/Catalina) and get logged there. 

Now imagine the company has spent a lot of resources building a system that tells your thousand+ developers when their apps are logging errors. The idea is that they can get notified when a novel error occurs in their system so they can clean it up. This is a great way to encourage error/logging hygiene over a large workforce of engineers. Unfortunately, on pen-test day, every developer with the error logging feature enabled got an email (per service they operate) that said:

```
[...]

PR #34594, opened 2025-03-13
`FormHttpMessageConverter` should throw `HttpMessageNotReadableException` when the http form data is invalid

Helmsdown:
`HttpMessageNotReadableException` is a more specific exception that will allow users to better target and react to invalid request payloads.

@sdeleuze here is the change per our conversation per #34584


sdeleuze:
Polished and merged, thanks!

PR #34614, opened 2025-03-18
Add caching headers to unmodified static resources

yuzawa-san:
I was playing around with a CDN and noticed some interesting behavior. It was caching using the original Cache-Control max-age on the original 200, but when a refresh and 304 was returned by spring, the Cache-Control was missing, so the caching proxy used its default max-age (which was longer than my original max-age). I found this https://stackoverflow.com/questions/1587667/should-http-304-not-modified-responses-contain-cache-control-headers which pointed me at https://www.rfc-editor.org/rfc/rfc7232#section-4.1 which states:

> The server generating a 304 response MUST generate any of the following header fields that would have been sent in a 200 (OK) response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.


bclozel:
Thanks for your contribution @yuzawa-san !

PR #34623, opened 2025-03-20
Introduce Stream variant methods for SqlQuery

### LINKED ISSUE #34474: Introduce Stream variant methods for SqlQuery
`queryForStream()` was introduced for `JdbcOperations` since 5.3, It would be nice if `SqlQuery` could leverage it to process large result too.

I would like to prepare PR if the proposal is accepted.

quaff:
Closes GH-34474

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/object/SqlQuery.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/object/SqlQuery.java
@@ -18,6 +18,8 @@ package org.springframework.jdbc.object;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


jhoeller:
On review, could we simply name those methods `stream(...)` instead? And then also introduce `streamByNamedParam` equivalents next to the existing `executeByNamedParam` methods.

quaff:
> On review, could we simply name those methods `stream(...)` instead? And then also introduce `streamByNamedParam` equivalents next to the existing `executeByNamedParam` methods.

Updated, please review.

jhoeller:
Looks great! Thanks for your efforts.

PR #34636, opened 2025-03-22
More Macros added for flexibility

senocak:


bclozel:
Thanks for your contribution @senocak !

I don't think we will merge this for a couple of reasons. Half of those new macros are very specific ("every x minutes") and we don't intend to ship here a library of expressions. The other half seems highly localized and wouldn't apply to many countries and culture (for example, meal times wouldn't apply to Spain and even [weekend days are debatable](https://en.m.wikipedia.org/wiki/Workweek_and_weekend)).

Thanks!

PR #34645, opened 2025-03-25
Add support for "application/*+x-protobuf" media type

doroszkiewicz:
We need to version our API using protobuf. I added support for versioning via the media type header. The solution is analogous to versioning the JSON format.

Example header supported by this change: `application/vnd.examle.public.v1+x-protobuf`


bclozel:
Thanks @doroszkiewicz and congrats on your first contribution!

PR #34657, opened 2025-03-26
Add support for heic and heif mime types

yotam2010:
Add additional mime types to not default to application/octet-stream


yotam2010:
> I'm not sure we should merge this PR. We usually [keep this file in sync with Tomcat's defaults](https://github.com/apache/tomcat/blob/main/java/org/apache/catalina/startup/MimeTypeMappings.properties) and it doesn't seem to refer to heic/heif, even if those are quite popular.

Just looked why its not supported on Tomcat, and it seems that it is also not supported on browsers which is ultimately where I want to display it, so i guess we dont need it for our project anyhow.

Will just need to convert it.

Let me know if I should close this PR or if you still considering :)

PR #34665, opened 2025-03-28
Polish YamlProcessor

quaff:
The result of `yaml.loadAll(reader)` will never contains null element.


quaff:
https://github.com/spring-projects/spring-framework/blob/e41487492dc9d6b9aa644644ad4a0a975ac56cf0/spring-beans/src/main/java/org/springframework/beans/factory/config/YamlProcessor.java#L344

Not sure null check required here.

rstoyanchev (commented):
On closer look, not so easy to verify. `loadAll` calls `constructor.getData()` and that in turn calls `constructObject` which does a lookup into a Map, and even if the key is present, the value could be `null`. I didn't trace it further, but I think the null check is worth having as protection.

PR #34656, opened 2025-03-26
Upgrade AspectJ Transaction to 5 to Fix Proxy Issue

JoshuaChen:
After PR #34316, an issue occurs when using AspectJ for transaction management in Spring.

Test Versions
	•	Spring Boot: 3.4.4
	•	Spring Framework: 6.2.5

Enable the following settings in the Spring Boot application:

```java
@EnableTransactionManagement(mode = AdviceMode.ASPECTJ, proxyTargetClass = true)
@EnableAspectJAutoProxy(proxyTargetClass = true)
```

```yaml
logging:
  level:
    org.springframework.transaction: TRACE
    org.springframework.aop: TRACE
    org.springframework.cglib: TRACE
    org.springframework.aot: TRACE
```

The following error occurs:
```log
2025-03-26T16:24:32.800+08:00  INFO 17613 --- [           main] com.example.demo.DemoApplication         : Starting DemoApplication using Java 17.0.14 with PID 17613 (/Users/joshua/Code/github/JoshuaChen/demo/target/classes started by joshua in /Users/joshua/Code/github/JoshuaChen/demo)
2025-03-26T16:24:32.802+08:00  INFO 17613 --- [           main] com.example.demo.DemoApplication         : No active profile set, falling back to 1 default profile: "default"
2025-03-26T16:24:33.198+08:00  INFO 17613 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Bootstrapping Spring Data JPA repositories in DEFAULT mode.
2025-03-26T16:24:33.219+08:00  INFO 17613 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Finished Spring Data repository scanning in 16 ms. Found 1 JPA repository interface.
2025-03-26T16:24:33.298+08:00 DEBUG 17613 --- [           main] .s.a.a.a.ReflectiveAspectJAdvisorFactory : Found AspectJ method: public java.lang.Object org.springframework.transaction.aspectj.AbstractTransactionAspect.ajc$around$org_springframework_transaction_aspectj_AbstractTransactionAspect$1$2a73e96c(java.lang.Object,org.aspectj.runtime.internal.AroundClosure,org.aspectj.lang.JoinPoint$StaticPart)
[...]

PR #34316, opened 2025-01-24
Stop assuming that AspectJ Advice has JoinPoint as the first argument

JoshuaChen:
After enabling `@EnableTransactionManagement(mode = AdviceMode.ASPECTJ, proxyTargetClass = true)`, an exception occurs.

```
java.lang.IllegalStateException: Expecting to find 3 arguments to bind by name in advice, but actually found 2 arguments
```

The issue seems to be that AspectJ has changed its logic, and the first element in `aspectJAdviceMethod.getParameterTypes()` no longer returns `JoinPoint`, causing the exception.

```
2025-01-24T22:15:06.825+08:00 DEBUG 85696 --- [           main] .s.a.a.a.ReflectiveAspectJAdvisorFactory : Found AspectJ method: public java.lang.Object org.springframework.transaction.aspectj.AbstractTransactionAspect.ajc$around$org_springframework_transaction_aspectj_AbstractTransactionAspect$1$2a73e96c(java.lang.Object,org.aspectj.runtime.internal.AroundClosure,org.aspectj.lang.JoinPoint$StaticPart)
2025-01-24T22:15:06.825+08:00 DEBUG 85696 --- [           main] .s.a.a.a.ReflectiveAspectJAdvisorFactory : Ignoring incompatible advice method: public java.lang.Object org.springframework.transaction.aspectj.AbstractTransactionAspect.ajc$around$org_springframework_transaction_aspectj_AbstractTransactionAspect$1$2a73e96c(java.lang.Object,org.aspectj.runtime.internal.AroundClosure,org.aspectj.lang.JoinPoint$StaticPart)

java.lang.IllegalStateException: Expecting to find 3 arguments to bind by name in advice, but actually found 2 arguments.
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.bindExplicitArguments(AbstractAspectJAdvice.java:472) ~[spring-aop-6.2.1.jar:6.2.1]
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.bindArgumentsByName(AbstractAspectJAdvice.java:438) ~[spring-aop-6.2.1.jar:6.2.1]
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.calculateArgumentBindings(AbstractAspectJAdvice.java:389) ~[spring-aop-6.2.1.jar:6.2.1]
[...]

PR #34667, opened 2025-03-28
Enhance aspect to support superclass aspects

JoshuaChen:
Summary:

1. Continue to fix the missing logic of parameter binding in PR #34316
2. AspectJ correctly identifies the subclasses of abstract classes, fixing the failure of pointcut binding
3. Change the JointPoint in AbstractTransactionAspect.aj to ProceedingJoinPoint to fix the problem of parameter binding failure when calling @Transaction at runtime

For more background, see PR #34316 and #34656

After this PR, the following exceptions have been fixed

```log
2025-03-26T16:24:33.336+08:00 DEBUG 17613 --- [           main] o.s.a.aspectj.AspectJExpressionPointcut  : Pointcut parser rejected expression [transactionalMethodExecution(txObject)]: java.lang.IllegalArgumentException: Pointcut is not well-formed: expecting 'identifier' at character position 0

```

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
+++ b/spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java
@@ -29,6 +29,7 @@ import org.aopalliance.aop.Advice;
 import org.aopalliance.intercept.MethodInvocation;
 import org.aspectj.lang.JoinPoint;
 import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.runtime.internal.AroundClosure;
 import org.aspectj.weaver.tools.JoinPointMatch;
 import org.aspectj.weaver.tools.PointcutParameter;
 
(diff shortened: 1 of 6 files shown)


jhoeller:
While this is definitely possible, I'm afraid we do not intend to go there. `@EnableAspectJAutoProxy` is only meant to deal with common `@Aspect` beans where such named pointcut indirection via inheritance is not a problem to deal with. Our management of different aspect types and different AspectJ usage scenarios is complex enough already.

Last but not least, we actually mean to ignore those weaving-only aspects for proxying purposes; that's by design. If `@EnableTransactionManagement(mode=ASPECTJ)` and `@EnableAspectJAutoProxy` are both declared, the former uses the transaction aspect for weaving purposes whereas the latter just looks at all custom `@Aspect` beans for proxying purposes. This is what the common intention is in existing applications, so we need to keep them separate.

So to be clear, we actually do not want our transaction aspect (or our similarly designed caching aspect) to be picked up by `@EnableAspectJAutoProxy` in such a setup since this would cause double invocation of the functionality around the same method: once from weaving and once from proxying. That's what our total exclusion of ajc-compiled aspects was for, and while this has been relaxed in the meantime, we still need to keep the weaving scenarios separate in common application setups.

So thanks for your efforts but please note that we do not intend to accept further revisions in this area.

PR #34669, opened 2025-03-28
Add convenient method to get generic beans by type reference

quaff:
Feel free to close if the team think it's not worthy.


jhoeller:
Actually, I'd like to avoid such further overloads there. If at all, we'd introduce a `getBeanProvider(ParameterizedTypeReference)` method (since there is no `getBean(ResolvableType)` method either, just `getBeanProvider(ResolvableType)`), and such an `ObjectProvider`-returning method would then just have the `ResolvableType.forType` convenience without further value.

Issue #34570, opened 2025-03-11
Refine `UriUtils#decode` and `StringUtils#uriDecode` implementation and documentation


gotson:
I believe i found an issue with `UriUtils.decode(String source, Charset charset)`.

When the `source` string contains the character `’` (Right Single Quotation Mark - unicode 2019), and other URI characters (to trigger the rewrite of the string), the character is changed to `` (End of Medium - unicode 0019).

Here is a sample test in Kotlin that highlights the behaviour:

```kotlin
@Test
  fun test(){
    val c = '\u2019'
    val s = "%20$c"
    val expected = " $c"

    val d = UriUtils.decode(s, Charsets.UTF_8)

    assertThat(d).isEqualTo(expected)
  }
```

Here is the difference shown from the failing test:

```
" ’"
" "
```

I am using Spring 6.2.0.

sdeleuze:
I suspect an inconsistency between the provided UTF-16 hexadecimal value and the UTF-8 charset specified, see https://www.fileformat.info/info/unicode/char/2019/index.htm.

Also if you check the Javadoc of the underlying [`StringUtils#uriDecode`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StringUtils.html#uriDecode(java.lang.String,java.nio.charset.Charset)) method, if am not sure what you try to do is supported ("For all other characters (including those already decoded), the output is undefined").

Any thoughts?

nosan:
I expect the output to behave similarly to `URLDecoder`, except that the `+` sign is treated as a literal `+` instead of a space (' ').

```java
@Test
void uriDecode() {
    assertThat(URLDecoder.decode("%20\u2019", StandardCharsets.UTF_8)).isEqualTo(" ’"); // success
    assertThat(URLDecoder.decode("\u2019", StandardCharsets.UTF_8)).isEqualTo("’"); // success
    assertThat(StringUtils.uriDecode("\u2019", StandardCharsets.UTF_8)).isEqualTo("’"); // success
    assertThat(StringUtils.uriDecode("%20\u2019", StandardCharsets.UTF_8)).isEqualTo(" ’"); // fail
}
```

sdeleuze:
@nosan Despite its name, `URLDecoder ` is per its Javadoc an "utility class for HTML form decoding", it has a different purpose so those differences of behavior are expected.

[...]

PR #34673, opened 2025-03-28
Refine `UriUtils#decode` and `StringUtils#uriDecode` implementation and documentation

kilink:
Optimize the StringUtils.uriDecode method in the following ways:

- Use a StringBuilder instead of ByteArrayOutputStream, and only decode %-encoded sequences.
- Use HexFormat.fromHexDigits to decode hex sequences.
- Decode to a byte array that is only allocated if encoded sequences are encountered.

Using a StringBuilder and decoding to a temporary byte buffer are inspired by how JDK's URLDecoder works; HexFormat was added in JDK17 and should be more performant than Integer.parseInt as it uses lookup tables.

I initially looked at this as an easy optimization opportunity, but it appears this also resolves the issue raised in #34570; I've added a test case to exercise the problem described in that issue.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -16,7 +16,6 @@
 
 package org.springframework.util;
 
-import java.io.ByteArrayOutputStream;
 import java.nio.charset.Charset;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -25,6 +24,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.Enumeration;
+import java.util.HexFormat;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/util/StringUtils.java
```diff
@@ -816,38 +816,50 @@ public static boolean pathEquals(String path1, String path2) {
 	 * @see java.net.URLDecoder#decode(String, String)
 	 */
 	public static String uriDecode(String source, Charset charset) {
+		Assert.notNull(charset, "Charset must not be null");
 		int length = source.length();
```
nosan, line 820: I believe this code could be adjusted slightly to avoid using a `byte[]` array.

```java
	private static final HexFormat HEX_FORMAT = HexFormat.of();

		public static String uriDecode(String source, Charset charset) {
		Assert.notNull(charset, "Charset must not be null");
		Assert.notNull(source, "Source must not be null");
		int length = source.length();
		if (length == 0) {
			return source;
		}
		StringBuilder result = new StringBuilder(length);
		boolean changed = false;
		for (int i = 0; i < length; i++) {
			char ch = source.charAt(i);
			if (ch != '%') {
				result.append(ch);
				continue;
			}
			if (i + 2 >= length) {
				throw new IllegalArgumentException("Invalid encoded sequence \"%s\"".formatted(source.substring(i)));
			}
			result.append(new String(HEX_FORMAT.parseHex(source, i + 1, i + 3), charset));
			i += 2;
			changed = true;
		}
		return (changed ? result.toString() : source);
	}
```	

It would also be great to add tests for `StringUtils.uriDecode(...)` and update the JavaDoc if possible. Thank you so much!

kilink: That approach doesn't work for multi-byte sequences.

nosan: I’m not sure if it does not work for multi-byte sequences. Do you have a test for that?

kilink: There are already tests that cover this, see [here](https://github.com/spring-projects/spring-framework/blob/551f6c0c66c3168df7e8f368b83f1c43cbd887a9/spring-web/src/test/java/org/springframework/web/util/UriUtilsTests.java#L107) for example.

nosan: Oh, I see! Thanks!

nosan: What do you think about this option? 

```java
    public static String uriDecode(String source, Charset charset) {
		Assert.notNull(source, "Source must not be null");
		Assert.notNull(charset, "Charset must not be null");
		int length = source.length();
		if (length == 0) {
			return source;
		}
		StringBuilder result = new StringBuilder();
		ByteArrayOutputStream bos = new ByteArrayOutputStream();
		int index = 0;
		while (index < length) {
			char ch = source.charAt(index);
			if (ch != '%') {
				result.append(ch);
				index++;
				continue;
			}
			while (index < length && source.charAt(index) == '%') {
				try {
					bos.write(HexFormat.fromHexDigits(source, index + 1, index + 3));
				}
				catch (IllegalArgumentException | IndexOutOfBoundsException ex) {
					throw new IllegalArgumentException("Invalid encoded sequence \"%s\"".formatted(source.substring(index)),
							ex);
				}
				index += 3;
			}
			result.append(bos.toString(charset));
			bos.reset();
		}
		return result.toString();
	}
```

sdeleuze (changes requested):
@kilink Could you please add a Signed-off-by [trailer](https://git-scm.com/docs/git-interpret-trailers) to your commit message as described in https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring?

sdeleuze:
Merged with some additional optimization + a Javadoc update, thanks a lot @kilink.

PR #34679, opened 2025-03-29
Remove unnecessary closing curly brackets in Javadoc

ngocnhan-tran1996:


spring-core/src/main/java/org/springframework/asm/SymbolTable.java
```diff
@@ -1473,7 +1473,7 @@ private static final class LabelEntry {
 
     /**
      * Another entry (and so on recursively) having the same hash code (modulo the size of {@link
-     * SymbolTable#labelEntries}}) as this one.
+     * SymbolTable#labelEntries}) as this one.
```
sbrannen, line 1476: Please note that this class belongs to our fork of the ASM project.

Consequently, we do not modify the Javadoc. Changes to ASM Javadoc should be made in the upstream project.

Thus, I will revert that when merging this PR.

sbrannen:
This has been merged in 30fcaef81349e2239869c44acd72f0b9b5dac7b0 and revised in 9fd1d0c6a33e5b32cd50322b7cbfeebc6707d19c.

Thanks

spring-core/src/main/java/org/springframework/asm/SymbolTable.java
izeye: I created https://gitlab.ow2.org/asm/asm/-/merge_requests/424 to fix it.

PR #34681, opened 2025-03-29
Make TransactionalApplicationListenerAdapter support fallbackExecution

### LINKED ISSUE #34674: Make TransactionalApplicationListenerAdapter support fallbackExecution
I have the following use case.

I have a bean that should act as a multi node event router. 
The bean must listen for a particular event interface and broadcast a received event to the other nodes. If a transaction is running, event should be broadcast after commit. If there is no transaction, the event should be immediatly broadcast.

The event interface looks like this:
```java
interface MultiNodeClusterEvent extends Serializable {}
```

The router looks like this:
```java
/**
 * @author Réda Housni Alaoui
 */
@Component
class MultiNodeClusterEventRouter {

  private final ApplicationEventPublisher eventPublisher;

  MultiNodeClusterEventRouter(ApplicationEventPublisher eventPublisher) {
    this.eventPublisher = requireNonNull(eventPublisher);
  }

  @TransactionalEventListener(fallbackExecution = true)
  public void onLocalEvent(MultiNodeClusterEvent event) {
    // Broadcast to other nodes
  }

  // Triggered by JGroup
  private void onRemoteEvent(MultiNodeClusterEvent event) {
    eventPublisher.publishEvent(event);
  }

}

```

The obvious issue with this code is that the routers (plural because 1 router/node) may enter in an infinite broadcast loop. The easiest solution that comes to mind is to make the router check the event source and ignore the event if the source is itself. But since I am listening for a raw event, I don't have access to `ApplicationEvent` informations.

I know that under the cover, Spring use `PayloadApplicationEvent` for raw events. So I try to implement the infinite loop guard like this:

```java
/**
 * @author Réda Housni Alaoui
 */
@Component
class MultiNodeClusterEventRouter extends TransactionalApplicationListenerAdapter<PayloadApplicationEvent<MultiNodeClusterEvent>> {

  private final ApplicationEventPublisher eventPublisher;

  MultiNodeClusterEventRouter(ApplicationEventPublisher eventPublisher) {
    super(event -> {
      // Missing fallbackExecution !
      // If event.source == this, ignore.
      // Else, broadcast to other nodes
   });
    this.eventPublisher = requireNonNull(eventPublisher);
  }

  // Triggered by JGroup
  private void onRemoteEvent(MultiNodeClusterEvent event) {
    eventPublisher.publishEvent(new PayloadApplicationEvent<>(event));
  }

}
```

But when I do this, I loose `TransactionalApplicationListenerMethodAdapter#fallbackExecution` feature. 

Could we add this feature to `TransactionalApplicationListenerAdapter` and pass its activation as an argument to `TransactionalApplicationListener#forPayload` ?

reda-alaoui:
Fix #34674

### DIFF
--- a/spring-tx/src/main/java/org/springframework/transaction/event/TransactionalApplicationListener.java
+++ b/spring-tx/src/main/java/org/springframework/transaction/event/TransactionalApplicationListener.java
@@ -43,6 +43,7 @@ import org.springframework.core.Ordered;
  *
  * @author Juergen Hoeller
  * @author Oliver Drotbohm
+ * @author Réda Housni Alaoui
  * @since 5.3
  * @param <E> the specific {@code ApplicationEvent} subclass to listen to
  * @see TransactionalEventListener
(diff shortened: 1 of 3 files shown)


sbrannen:
Closing due to https://github.com/spring-projects/spring-framework/issues/34674#issuecomment-2815408440.

PR #34592, opened 2025-03-13
Recursively boxing/unboxing nested inline value classes

dmitrysulman:
Fixes #34458

Issue #34458, opened 2025-02-20
`IllegalArgumentException` thrown when using Kotlin unsigned integer `value class` as annotated method parameter


evo13th:
## Summary

`IllegalArgumentException` is thrown when Kotlin unsigned `value class` (`ULong`, `UInt`) is used as the request parameter. However it's working as expected when the `value class` is signed (`Long`, `Int`)

## Environment

- Spring Framework version 6.2.2
- JVM version 21
- Kotlin version 2.1.10

## Minimal Example Code

```kotlin
import org.springframework.web.bind.annotation.GetMapping
import org.springframework.web.bind.annotation.RequestParam
import org.springframework.web.bind.annotation.RestController

@RestController
class IntegerController {
    @GetMapping("/unsigned")
    suspend fun doUnsigned(@RequestParam param: UnsignedIntegerValueClass): String {
        // got exception
        return param.value.toString()
    }

    @GetMapping("/unsigned-path/{param}")
    suspend fun doUnsigned2(@PathVariable param: UnsignedIntegerValueClass): String {
        // got exception too
        return param.value.toString()
    }

    @GetMapping("/signed")
    suspend fun doSigned(@RequestParam param: SignedIntegerValueClass): String {
        // works
        return param.value.toString()
    }

    @GetMapping("/signed-path/{param}")
    suspend fun doSigned2(@PathVariable param: SignedIntegerValueClass): String {
        // also works
        return param.value.toString()
    }

    @JvmInline
    value class UnsignedIntegerValueClass(val value: ULong)

    @JvmInline
    value class SignedIntegerValueClass(val value: Long)
}
```

## Stacktrace

<details>

```
java.lang.IllegalArgumentException: object is not an instance of declaring class
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.checkReceiver(DirectMethodHandleAccessor.java:197) ~[na:na]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Assembly trace from producer [reactor.core.publisher.MonoCreate] :
	reactor.core.publisher.Mono.create(Mono.java:203)
	kotlinx.coroutines.reactor.MonoKt.monoInternal(Mono.kt:83)
Error has been observed at the following site(s):
	*_______________________Mono.create ⇢ at kotlinx.coroutines.reactor.MonoKt.monoInternal(Mono.kt:83)
[...]

Issue #33630, opened 2024-10-01
Introduce `KotlinUtils`


sdeleuze:
Kotlin support has grown organically both in Spring Framework and other Spring portfolio projects, the set of features required portfolio wide is now pretty stabilized but there is a lack of reuse with related maintainability challenges.

For example, the logic Kotlin logic in [`InvocableHandlerMethod`](https://github.com/spring-projects/spring-framework/blob/309897497a3c5802e8abc7aa5e09c32904aef9cb/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java#L298-L340) is involved, duplicated on  WebFlux side, and required on Spring Boot side for features like configuration property binding to a Kotlin value class as described in https://github.com/spring-projects/spring-boot/issues/41693.

Spring Data has a also lot of Kotlin related infrastructure that could be shared as well.

mp911de:
Spring Data has a set of utils ([`KotlinReflectionUtils`](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/java/org/springframework/data/util/KotlinReflectionUtils.java), [`KotlinBeanInfoFactory`](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/java/org/springframework/data/util/KotlinBeanInfoFactory.java), [`KotlinValueUtils`](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/java/org/springframework/data/mapping/model/KotlinValueUtils.java)) for various applications, such as value class usage and Property descriptor detection. Would it make sense to move some of these into Spring Framework to let other components also benefit from Value classes?

sdeleuze:
@mp911de The functional scope of those utils make sense to me, we may end up to refine some public API and implementation, especially if we are able to move forward on #21546 and collaborate closely with Kotlin team to simplify value class support and fix some bug you have reported like https://youtrack.jetbrains.com/issue/KT-20768. if that's ok for you I will have a deeper look to Spring Data Kotlin utils, and include related learnings in my discussion with the Kotlin team, and when we have a reasonably clear idea of the underlying capabilities we will use, we can start collaborating actively on that.

PR #34682, opened 2025-03-29
Recursively boxing Kotlin nested value classes in `CoroutinesUtils`

dmitrysulman:
This PR is a follow-up to #34592.

Unfortunately #34592 does not fully resolve the original issue (#34458), as the invocation of suspended functions occurs in `CoroutinesUtils` which was not addressed in that PR.

While #33630 may eventually eliminate this logic duplication, it's still worthwhile to add recursive boxing to `CoroutinesUtils` in the meantime.


sdeleuze:
Merged, thanks.

Issue #34654, opened 2025-03-25
Support Forwarded header "by" parameter in ForwardedHeaderFilter


expuss2000:
Spring doesn't parse the "by" parameter of the Forwarded header in [ForwardedHeaderFilter](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/filter/ForwardedHeaderFilter.java).

According to [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239#section-5.1):

> The "by" parameter is used to disclose the interface where the request came in to the proxy server.

I think it should override `getLocalAddr`/`getLocalPort` of the `HttpServletRequest`: https://jakarta.ee/specifications/servlet/6.0/apidocs/jakarta.servlet/jakarta/servlet/servletrequest#getLocalAddr()

The syntax of a "by" value is the same of the "for" parameter.

Support for the "for" parameter has been added as requested in https://github.com/spring-projects/spring-framework/issues/23260

bclozel:
Superseded by #34683

Issue #23260, opened 2019-07-09
Support for X-Forwarded-For and Forwarded for="..."


larsgrefer:
The `ForwardedHeaderFilter` seems to ignore the `X-Forwarded-For` header and it's `Forwarded` complement: https://tools.ietf.org/html/rfc7239#section-5.2

As user of this filter, I would expect `X-Forwarded-For` or `Forwarded for=` to be handled and removed from the request. If there is a reason to not handle `X-Forwarded-For` this should be documented.

luvarqpp:
I am trying not to make separate issue, so please look, if this is same problem: https://github.com/spring-projects/spring-security/issues/7081

rstoyanchev:
@luvarqpp, the `server.use-forward-headers` property in Spring Boot relies on the server, and does not use the `ForwardedHeaderFilter`. The two are actually mutually exclusive, and you should either enable `server.use-forward-headers` property or configure the `ForwardedHeaderFilter`, but not both.

Related to this, it does look like in Spring Boot 2.2 you'll be [able to choose](https://github.com/spring-projects/spring-boot/issues/5677#issuecomment-454864389) where you get your forwarded header (server or Spring Framework.

rstoyanchev:
@larsgrefer can you clarify what handling you would expect to see? Overrides to `getRemoteAddr`, `getRemoteHost`, and `getRemotePort` in `HttpServletRequest`?

larsgrefer:
I'd expect overrides to `getRemoteAttr` and `getRemoteHost` (the same way like Tomcat's `RemoteIpValve` does it)

We've migrated our application from Tomcat's `RemoteIpValve` to Spring's `ForwardedHeaderFilter`, because we need the support of `X-Forwarded-Host`. But then we noticed that `X-Forwarded-For` isn't handled anymore.

luvarqpp:
@rstoyanchev Thanks for clarification, but setting **server.use-forward-headers** to **true** does not help making simple login project to work correctly. Still first redirect does have wrong **Location** header. I have my experiment based on spring security example called [hello world](https://github.com/spring-projects/spring-security/tree/master/samples/boot/helloworld).

PS: I have made given project to work, when **Forwarded** header is in use (instead of X-Forwarded-* family). Does tomcat require standardized header for forwarding?

rstoyanchev:
Thanks for the comment but this isn't the place to ask about Tomcat.

yawo:
[...]

PR #34683, opened 2025-03-29
Add support for Forwarded By HTTP header

remeio:
See gh-34654, gh-23260


bclozel:
Thanks for your contribution @remeio 

I chose to improve a bit your changes by removing the `X-Forwarded-By` HTTP header support. While many non-standard `X-Forwarded-*` exist, I couldn't find any usage of that one on the Internet. Let's not introduce support for a new non-standard variant then!

PR #34697, opened 2025-04-01
Fix GenericConversionService selects incorrect converter

### LINKED ISSUE #34685: `GenericConversionService` selects incorrect converter for `String` to `List<String>` conversion
This issue is related to #34535 and #34298.

`Converter<String, List<? extends Map<String, ?>>>` is incorrectly selected when converting a `String` to a `List<String>`.

The following reproducer test fails on all `6.2.x` versions except `6.2.3`:

```java
package com.example;

import org.springframework.core.convert.TypeDescriptor;
import org.springframework.core.convert.converter.Converter;
import org.springframework.core.convert.support.GenericConversionService;

import java.util.List;
import java.util.Map;

class GenericConversionServiceTests {

	private final GenericConversionService conversionService = new GenericConversionService();

	@Test
	@SuppressWarnings("unchecked")
	void stringToListOfString() {
		conversionService.addConverter(new StringToCollectionConverter(conversionService));
		conversionService.addConverter(new StringToListOfMapConverter());

		List<String> result = (List<String>) conversionService.convert("foo,bar",
				TypeDescriptor.valueOf(String.class),
				TypeDescriptor.collection(List.class, TypeDescriptor.valueOf(String.class))
		);

		assertThat(result.get(0)).isEqualTo("foo");
	}

	private static class StringToListOfMapConverter implements Converter<String, List<? extends Map<String, ?>>> {

		@Override
		public List<? extends Map<String, ?>> convert(String source) {
			return List.of(Map.of("bar", source));
		}
	}

	// Using this version of the converter also results in a failure
	private static class StringToListOfMapConverterWithoutWildcard implements Converter<String, List<Map<String, ?>>> {

		@Override
		public List<Map<String, ?>> convert(String source) {
			return List.of(Map.of("bar", source));
		}
	}
}
```

The exception shows that the `StringToListOfMapConverter` converter was chosen instead of expected `StringToCollectionConverter`:
```
class java.util.ImmutableCollections$Map1 cannot be cast to class java.lang.String (java.util.ImmutableCollections$Map1 and java.lang.String are in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class java.util.ImmutableCollections$Map1 cannot be cast to class java.lang.String (java.util.ImmutableCollections$Map1 and java.lang.String are in module java.base of loader 'bootstrap')
	at org.springframework.core.convert.support.GenericConversionServiceTests.stringToListOfString(GenericConversionServiceTests.java:28)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
```

belljun3395:
Fix [gh-34685](https://github.com/spring-projects/spring-framework/issues/34685)

This method is implemented to determine type compatibility between ResolvableType instances,
not only by comparing raw types but also by recursively checking deep assignability of generic types.
An alternative approach could be using TypeDescriptor.isAssignableTo.
I would appreciate your review and feedback on this.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
@@ -342,10 +342,50 @@ public class GenericConversionService implements ConfigurableConversionService {
 		}
 
 		public boolean matchesFallback(TypeDescriptor sourceType, TypeDescriptor targetType) {
-			return (this.typeInfo.getTargetType() == targetType.getObjectType() &&
-					this.targetType.hasUnresolvableGenerics() &&
-					(!(this.converter instanceof ConditionalConverter conditionalConverter) ||
-							conditionalConverter.matches(sourceType, targetType)));
+			if (this.typeInfo.getTargetType() != targetType.getObjectType()) {
+				return false;
+			}
+
+			if (!this.targetType.hasUnresolvableGenerics()) {
+				return false;
+			}
+
+			boolean assignable = isAssignableTo(targetType.getResolvableType());
+			if (!assignable) {
+				return false;
+			}
+
+			return !(this.converter instanceof ConditionalConverter conditionalConverter) ||
+					conditionalConverter.matches(sourceType, targetType);
+		}
+
+		private boolean isAssignableTo(ResolvableType expected) {
+			return isAssignable(this.targetType, expected);
+		}
+
+		private boolean isAssignable(ResolvableType actual, ResolvableType expected) {
+			Class<?> actualResolved = actual.resolve(Object.class);
+			Class<?> expectedResolved = expected.resolve(Object.class);
+			if (!expectedResolved.isAssignableFrom(actualResolved)) {
+				return false;
+			}
+
+			ResolvableType[] actualGenerics = actual.getGenerics();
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/core/convert/support/GenericConversionService.java
```diff
@@ -342,10 +342,50 @@ public boolean matches(TypeDescriptor sourceType, TypeDescriptor targetType) {
...
+				return false;
+			}
+
+			return !(this.converter instanceof ConditionalConverter conditionalConverter) ||
+					conditionalConverter.matches(sourceType, targetType);
+		}
+
+		private boolean isAssignableTo(ResolvableType expected) {
+			return isAssignable(this.targetType, expected);
+		}
```
belljun3395, line 364: If use `TypeDescriptor.isAssignableTo`

```java
private boolean isAssignableTo(TypeDescriptor targetType) {
	TypeDescriptor converterTargetType = new TypeDescriptor(this.targetType, this.targetType.toClass(), null);
	return converterTargetType.isAssignableTo(targetType);
}
```

PR #34705, opened 2025-04-02
Fix broken link for Server-Sent Events

### LINKED ISSUE #34704: The link of Server-Sent Events is broken
The link of Server-Sent Events is broken.

https://github.com/spring-projects/spring-framework/blob/dd888ed8139ab1ae16d7ee6b5628bab202d3e649/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java#L38

Actually it's not exactly broken. It goes to https://html.spec.whatwg.org/multipage/. But i think it should go to https://html.spec.whatwg.org/multipage/server-sent-events.html

acktsap:
Closes gh-34704

I've searched with `/eventsource` text

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webmvc-functional.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webmvc-functional.adoc
@@ -276,7 +276,7 @@ ServerResponse.async(asyncResponse);
 ----
 ======
 
-https://www.w3.org/TR/eventsource/[Server-Sent Events] can be provided via the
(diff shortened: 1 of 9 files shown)


sdeleuze:
Thanks for submitting this PR, be aware that I will refine the proposed commit by removing the `@author` tag as we don't add it for such small changes, especially documentation ones.

sdeleuze:
Merged, thanks!

PR #34713, opened 2025-04-03
Fix inconsistent query parameter conversion for String params

### LINKED ISSUE #34613: Inconsistent Query Parameter Conversion When Mapping Multiple Values to a Single Parameter
There is a significant inconsistency when converting a query parameter from a collection to a single value. In the following example controller, the value received by the method varies unpredictably when multiple values are provided, leading to inconsistent behavior.

```java
@RestController
public class SampleController {
    
    @PostMapping("/myIntegerParam")
    public String printIntegerParameter(@RequestParam(required = false) Integer param) {
        System.out.println("Query parameter: " + param);
        return "Received param: " + param;
    }

    @PostMapping("/myStringParam")
    public String printStringParameter(@RequestParam(required = false) String param) {
        System.out.println("Query parameter: " + param);
        return "Received param: " + param;
    }
}
```
If you send a `/myIntegerParam?param=1&param=2` request, the `printIntegerParameter()` method will only receive the first value `1`.
But if you send a `/myStringParam?param=1&param=2` request, the `printStringParameter()` method will receive both values comma separated as a single string: `1,2`

The reason why this is happening is because in the 1st case the conversion is handled by the `org.springframework.core.convert.support.ArrayToObjectConverter` class, while in the 2nd by the `org.springframework.core.convert.support.ArrayToStringConverter` class.

I'm not sure whether this is a deliberate design decision or simply Spring delegating the conversion in its standard way. However, for a **query parameter**, this behavior introduces several issues:

1. **Inconsistency** – This approach is inconsistent with how other types are handled (e.g., integers in my example). For most types, Spring keeps only the first value, but for strings, it concatenates them. Why should a developer expect this discrepancy?

2. **Lack of Practical Use** – The resulting value is not meaningful. It is neither the first nor the last provided value, nor any actual value passed. This makes it unclear how a developer should handle it.

3. **Incorrect Assumption** – If a developer intended to get multiple values, they would have used a list rather than a single string.

4. **Unnecessary Complexity** – Given that string query parameters are extremely common, this unusual behavior forces developers to add extra validations and custom converters to ensure expected behavior.

To maintain consistency, the best approach when converting multiple query parameters into a single value is to always return the same value (either the first or last)

The above behavior was observed with:
- 'org.springframework.boot' version '3.4.3'
- 'org.springframework:spring-core' version '6.2.3'

kssumin:
## Issue

When converting multiple query parameters with the same name to a single value, Spring behaves inconsistently depending on the target type:

* For Integer parameters: `/endpoint?param=1&param=2` → first value is used (`1`)
* For String parameters: `/endpoint?param=1&param=2` → comma-separated string is created (`1,2`)

This inconsistency makes behavior unpredictable and forces developers to implement workarounds.

## Changes

This PR modifies `WebRequestDataBinder` to handle `String[]` to `String` conversion consistently with other types by only using the first value when binding web request parameters.

## Tests

Added unit tests to verify consistent behavior across different parameter types.

Fixes #34613

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/bind/support/WebRequestDataBinder.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/support/WebRequestDataBinder.java
@@ -21,7 +21,10 @@ import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.Part;
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.beans.ConfigurablePropertyAccessor;
+import org.springframework.beans.InvalidPropertyException;
 import org.springframework.beans.MutablePropertyValues;
+import org.springframework.beans.PropertyValue;
 import org.springframework.core.MethodParameter;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
@@ -73,6 +76,7 @@ import org.springframework.web.multipart.support.StandardServletPartUtils;
  *
  * @author Juergen Hoeller
  * @author Brian Clozel
+ * @author Sumin Kim
  * @since 2.5.2
  * @see #bind(org.springframework.web.context.request.WebRequest)
  * @see #registerCustomEditor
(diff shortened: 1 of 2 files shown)


bclozel:
Original issue has been declined. https://github.com/spring-projects/spring-framework/issues/34613#issuecomment-2782155637

Thanks for your contribution.

PR #34716, opened 2025-04-04
Introduce minimal retry functionality as a core framework feature

fmbenhassine:
This PR introduces a minimal core retry feature in Spring Framework. It is inspired by Spring Retry, but redesigned and trimmed to the bare minimum to cover most cases.

The main entry point is the `RetryTemplate` class, which is designed in a similar fashion to other templates provided by Spring. The package is minimal on purpose to keep its usage as simple as possible (see examples in `RetryTemplateTests`). It focuses only on stateless retries for now, but can be extended with other retry operations as well.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/retry/RetryCallback.java
+++ b/spring-core/src/main/java/org/springframework/core/retry/RetryCallback.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2002-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.retry;
+
+/**
+ * Callback interface for a retryable piece of code. Used in conjunction with {@link RetryOperations}.
+ *
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ * @param <R> the type of the result
+ * @see RetryOperations
+ */
+@FunctionalInterface
+public interface RetryCallback<R> {
+
+	/**
+	 * Method to execute and retry if needed.
+	 * @return the result of the callback
+	 * @throws Throwable if an error occurs during the execution of the callback
+	 */
+	R run() throws Throwable;
+
+	/**
+	 * A unique logical name for this callback to distinguish retries around
+	 * business operations.
+	 * @return the name of the callback. Defaults to the class name.
+	 */
+	default String getName() {
+		return getClass().getName();
+	}
+}
--- a/spring-core/src/main/java/org/springframework/core/retry/RetryExecution.java
+++ b/spring-core/src/main/java/org/springframework/core/retry/RetryExecution.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2002-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.retry;
+
+/**
+ * Strategy interface to define a retry execution.
+ *
+ * <p>Implementations may be stateful but do not need to be thread-safe.
+ *
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ */
+public interface RetryExecution {
+
+	/**
+	 * Specify if the operation should be retried based on the given throwable.
+	 * @param throwable the exception that caused the operation to fail
+	 * @return {@code true} if the operation should be retried, {@code false} otherwise
+	 */
+	boolean shouldRetry(Throwable throwable);
+
+}
--- a/spring-core/src/main/java/org/springframework/core/retry/RetryListener.java
+++ b/spring-core/src/main/java/org/springframework/core/retry/RetryListener.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2002-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.retry;
+
+import org.springframework.core.retry.support.CompositeRetryListener;
+
+/**
+ * An extension point that allows to inject code during key retry phases.
+ *
+ * <p>Typically registered in a {@link RetryTemplate}, and can be composed using
+ * a {@link CompositeRetryListener}.
+ *
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ * @see CompositeRetryListener
+ */
+public interface RetryListener {
+
+	/**
+	 * Called before every retry attempt.
+	 * @param retryExecution the retry execution
+	 */
+	default void beforeRetry(RetryExecution retryExecution) {
+	}
+
+	/**
+	 * Called after the first successful retry attempt.
+	 * @param retryExecution the retry execution
+	 * @param result the result of the callback
+	 */
+	default void onRetrySuccess(RetryExecution retryExecution, Object result) {
+	}
+
+	/**
+	 * Called every time a retry attempt fails.
+	 * @param retryExecution the retry execution
+	 * @param throwable the throwable thrown by the callback
+	 */
+	default void onRetryFailure(RetryExecution retryExecution, Throwable throwable) {
+	}
+
+	/**
+	 * Called once the retry policy is exhausted.
+	 * @param retryExecution the retry execution
+	 * @param throwable the last throwable thrown by the callback
+	 */
+	default void onRetryPolicyExhaustion(RetryExecution retryExecution, Throwable throwable) {
+	}
+
+}
--- a/spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
+++ b/spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2002-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.retry;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Main entry point to the core retry functionality. Defines a set of retryable operations.
+ *
+ * <p>Implemented by {@link RetryTemplate}. Not often used directly, but a useful
+ * option to enhance testability, as it can easily be mocked or stubbed.
+ *
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ * @see RetryTemplate
+ */
+public interface RetryOperations {
+
+	/**
+	 * Retry the given callback (according to the retry policy configured at the implementation level)
+	 * until it succeeds or eventually throw an exception if the retry policy is exhausted.
+	 * @param retryCallback the callback to call initially and retry if needed
+	 * @param <R> the type of the callback's result
+	 * @return the callback's result
+	 * @throws RetryException thrown if the retry policy is exhausted. All attempt exceptions
+	 * should be added as suppressed exceptions to the final exception.
+	 */
+	<R extends @Nullable Object> R execute(RetryCallback<R> retryCallback) throws RetryException;
+
+}
+
--- a/spring-core/src/main/java/org/springframework/core/retry/RetryPolicy.java
+++ b/spring-core/src/main/java/org/springframework/core/retry/RetryPolicy.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2002-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.retry;
+
+/**
+ * Strategy interface to define a retry policy.
+ *
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ */
+public interface RetryPolicy {
+
+	/**
+	 * Start a new retry execution.
+	 * @return a fresh {@link RetryExecution} ready to be used
+	 */
+	RetryExecution start();
+
+}
--- a/spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
+++ b/spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2002-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.retry;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.logging.LogFactory;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.core.log.LogAccessor;
+import org.springframework.core.retry.support.CompositeRetryListener;
+import org.springframework.core.retry.support.MaxRetryAttemptsPolicy;
+import org.springframework.util.Assert;
+import org.springframework.util.backoff.BackOff;
+import org.springframework.util.backoff.BackOffExecution;
+import org.springframework.util.backoff.FixedBackOff;
+
+/**
+ * A basic implementation of {@link RetryOperations} that uses a
+ * {@link RetryPolicy} and a {@link BackOff} to retry a
+ * {@link RetryCallback}. By default, the callback will be called
+ * 3 times with a fixed backoff of 1 second.
+ *
+ * <p>It is also possible to register a {@link RetryListener} to intercept and inject code
+ * during key retry phases (before a retry attempt, after a retry attempt, etc.).
+ *
+ * <p>All retry operations performed by this class are logged at debug level,
+ * using "org.springframework.core.retry.RetryTemplate" as log category.
+ *
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ * @see RetryOperations
+ * @see RetryPolicy
+ * @see BackOff
+ * @see RetryListener
+ */
+public class RetryTemplate implements RetryOperations {
+
+	protected final LogAccessor logger = new LogAccessor(LogFactory.getLog(getClass()));
+
+	protected RetryPolicy retryPolicy = new MaxRetryAttemptsPolicy();
+
+	protected BackOff backOffPolicy = new FixedBackOff();
+
+	protected RetryListener retryListener = new RetryListener() {
+	};
+
+	/**
+	 * Create a new retry template with default settings.
+	 */
+	public RetryTemplate() {
+	}
+
+	/**
+	 * Create a new retry template with a custom {@link RetryPolicy}.
+	 * @param retryPolicy the retry policy to use
+	 */
+	public RetryTemplate(RetryPolicy retryPolicy) {
+		Assert.notNull(retryPolicy, "Retry policy must not be null");
+		this.retryPolicy = retryPolicy;
+	}
+
+	/**
+	 * Create a new retry template with a custom {@link RetryPolicy} and {@link BackOff}.
+	 * @param retryPolicy the retry policy to use
+	 * @param backOffPolicy the backoff policy to use
+	 */
+	public RetryTemplate(RetryPolicy retryPolicy, BackOff backOffPolicy) {
+		this(retryPolicy);
+		Assert.notNull(backOffPolicy, "BackOff policy must not be null");
+		this.backOffPolicy = backOffPolicy;
+	}
+
+	/**
+	 * Set the {@link RetryPolicy} to use. Defaults to <code>MaxAttemptsRetryPolicy()</code>.
+	 * @param retryPolicy the retry policy to use. Must not be <code>null</code>.
+	 */
+	public void setRetryPolicy(RetryPolicy retryPolicy) {
+		Assert.notNull(retryPolicy, "Retry policy must not be null");
+		this.retryPolicy = retryPolicy;
+	}
+
+	/**
+	 * Set the {@link BackOff} to use. Defaults to <code>FixedBackOffPolicy(Duration.ofSeconds(1))</code>.
+	 * @param backOffPolicy the backoff policy to use. Must not be <code>null</code>.
+	 */
+	public void setBackOffPolicy(BackOff backOffPolicy) {
+		Assert.notNull(backOffPolicy, "BackOff policy must not be null");
+		this.backOffPolicy = backOffPolicy;
+	}
+
+	/**
+	 * Set the {@link RetryListener} to use. Defaults to a <code>NoOp</code> implementation.
+	 * If multiple listeners are needed, use a {@link CompositeRetryListener}.
+	 * @param retryListener the retry listener to use. Must not be <code>null</code>.
+	 */
+	public void setRetryListener(RetryListener retryListener) {
+		Assert.notNull(retryListener, "Retry listener must not be null");
+		this.retryListener = retryListener;
+	}
+
+	/**
+	 * Call the retry callback according to the configured retry and backoff policies.
+	 * If the callback succeeds, its result is returned. Otherwise, a {@link RetryException}
+	 * will be thrown to the caller having all attempt exceptions as suppressed exceptions.
+	 * @param retryCallback the callback to call initially and retry if needed
+	 * @param <R> the type of the result
+	 * @return the result of the callback if any
+	 * @throws RetryException thrown if the retry policy is exhausted. All attempt exceptions
+	 * are added as suppressed exceptions to the final exception.
+	 */
+	@Override
+	public <R extends @Nullable Object> R execute(RetryCallback<R> retryCallback) throws RetryException {
+		Assert.notNull(retryCallback, "Retry Callback must not be null");
+		String callbackName = retryCallback.getName();
+		// initial attempt
+		try {
+			logger.debug(() -> "About to execute callback '" + callbackName + "'");
+			R result = retryCallback.run();
+			logger.debug(() -> "Callback '" + callbackName + "' executed successfully");
+			return result;
+		}
+		catch (Throwable initialException) {
+			logger.debug(initialException, () -> "Execution of callback '" + callbackName + "' failed, initiating the retry process");
+			// retry process starts here
+			RetryExecution retryExecution = this.retryPolicy.start();
+			BackOffExecution backOffExecution = this.backOffPolicy.start();
+			List<Throwable> suppressedExceptions = new ArrayList<>();
+
+			Throwable retryException = initialException;
+			while (retryExecution.shouldRetry(retryException)) {
+				logger.debug(() -> "About to retry callback '" + callbackName + "'");
+				try {
+					this.retryListener.beforeRetry(retryExecution);
+					R result = retryCallback.run();
+					this.retryListener.onRetrySuccess(retryExecution, result);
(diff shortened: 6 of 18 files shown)


spring-core/src/main/java/org/springframework/core/retry/RetryPolicy.java
```diff
@@ -0,0 +1,33 @@
...
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ */
+public interface RetryPolicy {
+
+	/**
+	 * Return the maximum number of retry attempts.
+	 * @return the maximum number of retry attempts
+	 */
+	int getMaxAttempts();
```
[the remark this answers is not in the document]
sbrannen: How do you envision a user possibly being able to implement a retry policy conditionally based on the actual exception thrown?

sbrannen (changes requested):
Hi @fmbenhassine,

First and foremost, thanks for putting this PR together! 👍 

I took a quick look, added a few comments, and asked a few questions (as food for thought).

And we will also discuss the proposal within the team.

Cheers,

Sam

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
```diff
@@ -0,0 +1,136 @@
...
+ * using "org.springframework.core.retry.RetryTemplate" as log category.
+ *
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ * @see RetryOperations
+ * @see RetryPolicy
+ * @see BackOffPolicy
+ * @see RetryListener
+ */
+public class RetryTemplate implements RetryOperations {
```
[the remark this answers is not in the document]
fmbenhassine: Makes sense. What about a builder (similar to [the one in Spring Retry](https://github.com/spring-projects/spring-retry?tab=readme-ov-file#imperative-example))? Since we have at least 3 configuration options, a builder would be better than having several constructors IMO, do you agree?

spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
```diff
@@ -0,0 +1,45 @@
...
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ */
+public interface RetryOperations {
+
+	/**
+	 * Retry the given callback (according to the retry policy configured at the implementation level)
+	 * until it succeeds or eventually throw an exception if the retry policy is exhausted.
+	 * @param retryCallback the callback to call initially and retry if needed
+	 * @param <R>           the type of the callback's result
```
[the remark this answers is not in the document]
fmbenhassine: I will fix that. But shouldn't the code style check task catch that? It seems to pass currently with this violation in the code.

spring-core/src/main/java/org/springframework/core/retry/RetryPolicy.java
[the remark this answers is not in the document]
fmbenhassine: Yes, I don't exclude extending the contract of `RetryPolicy` with anything that we think essential for this first iteration. The list of exceptions to retry/exclude could be part of the retry policy contract.

By "can be extended with other retry operations" I was referring to stateful retries. Stateful retries require some sort of context (similar to the one currently in Spring Retry). But instead of overwhelming the `RetryOperations` interface with both stateless and stateful operations, I envision stateful retries to be defined in a specialization of `RetryOperations` (something like `StatefulRetryOperations` with additional methods dealing with a context). I am trying to follow the ISP here, as most users won't probably need stateful retires, and therefore should not be forced to implement those methods. Do you agree?

spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
```diff
@@ -0,0 +1,45 @@
...
+	 * @return the callback's result
+	 * @throws Exception if the retry policy is exhausted
+	 */
+	<R> @Nullable R execute(Callable<R> retryCallback) throws Exception;
```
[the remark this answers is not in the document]
fmbenhassine: The idea was to minimize the number of interfaces the user has to deal with. I did not see any added value of having a `RetryCallback` with a single method over a plain `Callable` object. But now that you [mentioned](https://github.com/spring-projects/spring-framework/pull/34716#discussion_r2028396733) the necessity to have a name or description of the callback for diagnosis purposes, I believe it is better to have a dedicated interface for it. (BTW, the `RetryCallback` from Spring Retry has a [label](https://github.com/spring-projects/spring-retry/blob/main/src/main/java/org/springframework/retry/RetryCallback.java#L41-L49), which can be used for logging). I will introduce such a callback.

Regarding the dedicated exception type, I agree as well (for consistency). I will update the PR accordingly.

As per the exception type, I believe it should be checked. The reason is that as a client of the template, I am expecting it to retry the code and eventually throw an exception if the retry policy is exhausted (there is no (need for a) recovery callback, the recovery is actually the code I am expecting to write in the catch block).

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
```diff
@@ -0,0 +1,136 @@
...
+	 * @throws Exception thrown if the retry policy is exhausted
+	 */
+	@Override
+	public <R> @Nullable R execute(Callable<R> retryCallback) throws Exception {
+		Assert.notNull(retryCallback, "Retry Callback must not be null");
+		int attempts = 0;
+		int maxAttempts = this.retryPolicy.getMaxAttempts();
+		while (attempts++ <= maxAttempts) {
+			if (logger.isDebugEnabled()) {
+				logger.debug("Retry attempt #" + attempts);
```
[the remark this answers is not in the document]
fmbenhassine: As mentioned [previously](https://github.com/spring-projects/spring-framework/pull/34716#discussion_r2028877670), I agree. I will add a callback with a label that is used to distinguish callbacks in log messages.

fmbenhassine:
Thank you for the review, @sbrannen !

I added a couple of inline comments and I will update the PR accordingly.

spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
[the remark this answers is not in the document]
fmbenhassine: Fixed.

spring-core/src/main/java/org/springframework/core/retry/RetryPolicy.java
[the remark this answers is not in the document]
fmbenhassine: I pushed a change set that enriches the retry policy with a method that allows user to provide a predicate on the exception and specify the retry condition. I believe a predicate is more flexible than two lists of exceptions to include/exclude as it is currently in Spring Retry.

spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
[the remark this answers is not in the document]
fmbenhassine: I introduced a `RetryCallback` interface to avoid being tied to `Callable`. The callback also provides a method to specify a logical name for diagnosis purposes.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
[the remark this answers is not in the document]
fmbenhassine: Done. I added two constructors with the retry policy and backoff policy. In hindsight, I see no need for a builder for now.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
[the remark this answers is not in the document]
fmbenhassine: Done

fmbenhassine:
I updated the PR with the following changes:

- Add/Use a specific `RetryException` in the API instead of the generic `Exception`
- Add `RetryCallback` interface with a logical name for diagnostics purposes
- Improve `RetryPolicy` with the ability to specify which exceptions to retry based on a `Predicate<Exception>`

I find this new version to be more coherent and consistent with other templates in terms of structure and API. Looking forward to the team's thoughts on this!

spring-core/src/main/java/org/springframework/core/retry/RetryCallback.java
```diff
@@ -0,0 +1,42 @@
...
+	 * @throws Throwable if an error occurs during the execution of the callback
+	 */
+	R run() throws Throwable;
+
+	/**
+	 * A unique logical name for this callback to distinguish retries around
+	 * business operations.
+	 * @return the name of the callback
+	 */
+	String getName();
```
artembilan, line 41: Not requirement, but I wonder if we can use a `default` convenient implementation here to the `getClass().getName()`?
Or generated lambda class name is not very useful?

spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * Retry the given callback (according to the retry policy configured at the implementation level)
+	 * until it succeeds or eventually throw an exception if the retry policy is exhausted.
+	 * @param retryCallback the callback to call initially and retry if needed
+	 * @param <R> the type of the callback's result
+	 * @return the callback's result
+	 * @throws RetryException thrown if the retry policy is exhausted. All attempt exceptions
+	 * should be added as suppressed exceptions to the final exception.
+	 */
+	<R> @Nullable R execute(RetryCallback<R> retryCallback) throws RetryException;
```
artembilan, line 42: I'm not very good with nullness, but feel like this contract has to `<R extends @Nullable Object>`.
So, the usage of this method would depend on the provided callback implementation.
I mean that is no up to us to say that `execute()` might return null.
The null could be only as a result of `RetryCallback` call.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
```diff
@@ -0,0 +1,188 @@
...
+		this.backOffPolicy = backOffPolicy;
+	}
+
+	/**
+	 * Set the {@link RetryListener} to use. Defaults to a <code>NoOp</code> implementation.
+	 * If multiple listeners are needed, use a {@link CompositeRetryListener}.
+	 * @param retryListener the retry listener to use. Must not be <code>null</code>.
+	 */
+	public void setRetryListener(RetryListener retryListener) {
+		Assert.notNull(retryListener, "Retry listener must not be null");
```
artembilan, line 113: @sbrannen , @jhoeller , do we still do these asserts if we promote nullness?

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
```diff
@@ -0,0 +1,188 @@
...
+	 * @throws RetryException thrown if the retry policy is exhausted. All attempt exceptions
+	 * are added as suppressed exceptions to the final exception.
+	 */
+	@Override
+	public <R> @Nullable R execute(RetryCallback<R> retryCallback) throws RetryException {
+		Assert.notNull(retryCallback, "Retry Callback must not be null");
+		String callbackName = retryCallback.getName();
+		// initial attempt
+		try {
+			if (logger.isDebugEnabled()) {
```
artembilan, line 133: Not important.
Can we use a `LogAccessor` API in Spring Core itself to make this code more clean?

spring-core/src/main/java/org/springframework/core/retry/support/MaxRetryAttemptsPolicy.java
```diff
@@ -0,0 +1,69 @@
...
+
+		protected int retryAttempts;
+
+		public MaxRetryAttemptsPolicyExecution() {
+			this.retryAttempts = 0;
+		}
+
+		@Override
+		public boolean shouldRetry(Throwable throwable) {
+			return this.retryAttempts++ < maxRetryAttempts;
```
artembilan, line 64: I'm not familiar with Spring Framework code style, but I find this way as useful:
```
MaxRetryAttemptsPolicy.this.maxRetryAttempts
```

spring-core/src/main/java/org/springframework/core/retry/support/PredicateRetryPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+	 * @return a fresh {@link PredicateRetryPolicyExecution} ready to be used
+	 */
+	public PredicateRetryPolicyExecution start() {
+		return new PredicateRetryPolicyExecution(this.predicate);
+	}
+
+	/**
+	 * A {@link RetryExecution} based on a predicate.
+	 */
+	public static class PredicateRetryPolicyExecution implements RetryExecution {
```
artembilan, line 53: Why `static`?
Why do you allow to use this instance outside of the policy itself?
Why we cannot use `PredicateRetryPolicy.this.predicate` property and lock-in the usage of this class?

artembilan (changes requested):
This is great even if I have left so many questions.
Thanks

spring-core/src/main/java/org/springframework/core/retry/RetryCallback.java
sbrannen: We have `ClassUtils.isLambdaClass()`, so we might be able to use that and base the name of the callback on the _enclosing class_ of the lambda.

But, TBH, I'm not sure what the enclosing class of a lambda is. I'd have to experiment with that.

spring-core/src/main/java/org/springframework/core/retry/RetryExecution.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.core.retry;
+
+/**
+ * Strategy interface to define a retry execution.
+ *
+ * <p>Implementations do not need to be thread-safe.
```
sbrannen, line 22: ... buy may be stateful?

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
sbrannen: Sometimes, but not always. 

We can discuss the pros and cons offline.

spring-core/src/main/java/org/springframework/core/retry/support/MaxRetryAttemptsPolicy.java
```diff
@@ -0,0 +1,69 @@
...
+	@Override
+	public MaxRetryAttemptsPolicyExecution start() {
+		return new MaxRetryAttemptsPolicyExecution();
+	}
+
+	/**
+	 * Set the maximum number of retry attempts.
+	 * @param maxRetryAttempts the maximum number of retry attempts. Must be greater than zero.
+	 */
+	public void setMaxRetryAttempts(int maxRetryAttempts) {
```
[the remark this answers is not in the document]
sbrannen: We might want to support both options.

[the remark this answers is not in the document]
fmbenhassine: I updated the code to support both options.

spring-core/src/test/java/org/springframework/core/retry/support/MaxRetryDurationPolicyTests.java
```diff
@@ -0,0 +1,54 @@
...
+	void testMaxRetryDuration() throws InterruptedException {
+		MaxRetryDurationPolicy retryPolicy = new MaxRetryDurationPolicy();
+		retryPolicy.setMaxRetryDuration(Duration.of(1000, ChronoUnit.MILLIS));
+		MaxRetryDurationPolicy.MaxRetryDurationPolicyExecution retryExecution = retryPolicy.start();
+
+		assertThat(retryExecution.shouldRetry(Mockito.any(Exception.class))).isTrue();
+		Thread.sleep(400);
+		assertThat(retryExecution.shouldRetry(Mockito.any(Exception.class))).isTrue();
+		Thread.sleep(400);
+		assertThat(retryExecution.shouldRetry(Mockito.any(Exception.class))).isFalse();
```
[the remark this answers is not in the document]
fmbenhassine: Indeed there is no guarantee for this test to be stable, so I removed it for now. If there is a better way to test this, please let me know and I will update accordingly.

spring-core/src/main/java/org/springframework/core/retry/RetryExecution.java
fmbenhassine: Yes, I updated the javadoc accordingly.

spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOffPolicy.java
```diff
@@ -0,0 +1,280 @@
...
+ *
+ * <p>Note that the default max elapsed time is {@link Long#MAX_VALUE}, and the
+ * default maximum number of attempts is {@link Integer#MAX_VALUE}.
+ * Use {@link #setMaxElapsedTime} to limit the length of time that an instance
+ * should accumulate before returning {@link BackOffExecution#STOP}. Alternatively,
+ * use {@link #setMaxAttempts} to limit the number of attempts. The execution
+ * stops when either of those two limits is reached.
+ *
+ * @author Stephane Nicoll
+ * @author Gary Russell
```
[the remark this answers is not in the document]
fmbenhassine: Done.

spring-core/src/main/java/org/springframework/core/retry/support/PredicateRetryPolicy.java
fmbenhassine: I replaced this with a Lambda.

spring-core/src/main/java/org/springframework/core/retry/support/MaxRetryAttemptsPolicy.java
fmbenhassine: I updated the code with that style, and I will let the team decide what is appropriate for this.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
fmbenhassine: I think it is not (only) about formatting, but rather about replacing all code blocks (`if (logger.isDebugEnabled()) { logger.debug(...); }`) with a single statement. I believe this is better for readability. Even if `LogAccessor` is not currently used in Spring Framework itself, may be this is an opportunity to start using it? I personally like that API and think to start using it in Batch.

For now, I updated the PR to use `LogAccessor`, but if the team thinks it's not appropriate, then I will revert that change.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
fmbenhassine: I kept the check for now. I will defer the decision to the team about conventions here.

spring-core/src/main/java/org/springframework/core/retry/RetryOperations.java
fmbenhassine: Good point, @artembilan ! And thank you for the pointer @sbrannen !

I updated the PR accordingly.

spring-core/src/main/java/org/springframework/core/retry/RetryCallback.java
fmbenhassine: Shall we keep `RetryCallback` "unnamed" to be a functional interface and introduce a `NamedRetryCallback` (This idea seems to be used in a couple of other places like `NamedThreadLocal` and `NamedBean`)? I believe most users won't need to name the callback. Those who use several callbacks in the same app can use `Named RetryCallback `s to distinguish them:

```java
@FunctionalInterface
public interface RetryCallback<R> {

	/**
	 * Method to execute and retry if needed.
	 * @return the result of the callback
	 * @throws Throwable if an error occurs during the execution of the callback
	 */
	R run() throws Throwable;

}

public interface NamedRetryCallback<R> extends RetryCallback<R> {

	/**
	 * A unique logical name for this callback to distinguish retries around
	 * business operations.
	 * @return the name of the callback
	 */
	String getName();
}
```

With that, there is no need to think about a default name as users of `NamedRetryCallback` would specify a name anyway. wdyt?

fmbenhassine:
Thank you for the review, @artembilan and @sbrannen !

I updated the PR accordingly and added a few comments. Please let me know if we need other updates.

The build of the PR is failing for tests that are not related to this change set.

spring-core/src/main/java/org/springframework/core/retry/support/CompositeRetryListener.java
```diff
@@ -0,0 +1,84 @@
...
+	public CompositeRetryListener() {
+		this.listeners = new LinkedList<>();
+	}
+
+	/**
+	 * Create a new {@link CompositeRetryListener} with a list of delegates.
+	 * @param listeners the delegate listeners to register
+	 */
+	public CompositeRetryListener(List<RetryListener> listeners) {
+		this.listeners = listeners;
```
artembilan: This is not correct.
We have to create a copy here internally.
Otherwise, any mutate to that external collection would effect our component as well.
And there is chance of opposite, more dangerous logic: we modify `this.listeners` internally here and everything else using that external collection would be effected.
I remember having this error a lot in our projects when we used SonarQube.

spring-core/src/test/java/org/springframework/core/retry/support/MaxRetryAttemptsPolicyTests.java
```diff
@@ -0,0 +1,54 @@
...
+	@Test
+	void testDefaultMaxRetryAttempts() {
+		// given
+		MaxRetryAttemptsPolicy retryPolicy = new MaxRetryAttemptsPolicy();
+
+		// when
+		RetryExecution retryExecution = retryPolicy.start();
+
+		// then
+		assertThat(retryExecution.shouldRetry(Mockito.any(Exception.class))).isTrue();
```
artembilan: You still use `Mockito.any()` in these tests while it is wrong context for this API.
Here the simple `Mockito.mock()` should be enough.
Or might be even better from performance perspective to use some real exception instance instead of reflection 😄

artembilan (changes requested):

spring-core/src/main/java/org/springframework/core/retry/support/CompositeRetryListener.java
fmbenhassine: Good catch! Fixed.

spring-core/src/test/java/org/springframework/core/retry/support/MaxRetryAttemptsPolicyTests.java
fmbenhassine: You are right, apologies I just missed that from our last review. Should be fixed now.

spencergibb:
👋🏻  Joining the conversation here as Spring Cloud is an advanced user of Spring Retry, so we would be interested in how the functionality compares to the Spring Retry project.

spencergibb:
Are they other PRs or issues I should look at?

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
sbrannen: > I think it is not (only) about formatting, but rather about replacing all code blocks (`if (logger.isDebugEnabled()) { logger.debug(...); }`) with a single statement.

You don't have to use `LogAccessor` to achieve that.

We achieve that in multiple places using only `LogMessage`.

For example:

https://github.com/spring-projects/spring-framework/blob/b699b65b40a0299d4845c2d0160ce1a0145d85d4/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractBeanFactory.java#L1746-L1748

spring-core/src/main/java/org/springframework/core/retry/support/PredicateRetryPolicy.java
```diff
@@ -0,0 +1,50 @@
...
+	 * Create a new {@link PredicateRetryPolicy} with the given predicate.
+	 * @param predicate the predicate to use for determining whether to retry the exception or not
+	 */
+	public PredicateRetryPolicy(Predicate<Throwable> predicate) {
+		this.predicate = predicate;
+	}
+
+	/**
+	 * Start a new retry execution.
+	 * @return a fresh ready to be used
```
sbrannen: A fresh what?

spring-core/src/main/java/org/springframework/core/retry/support/MaxRetryAttemptsPolicy.java
```diff
@@ -0,0 +1,88 @@
...
+	/**
+	 * A {@link RetryExecution} based on a maximum number of retry attempts.
+	 */
+	private class MaxRetryAttemptsPolicyExecution implements RetryExecution {
+
+		private int retryAttempts;
+
+		MaxRetryAttemptsPolicyExecution() {
+			this.retryAttempts = 0;
+		}
```
sbrannen:
```suggestion
		private int retryAttempts;
```

The constructor is unnecessary.

sbrannen:
> Are they other PRs or issues I should look at?

@spencergibb, this PR is the primary thing to look at at the moment.

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/core/retry/RetryListener.java
```diff
@@ -0,0 +1,62 @@
...
+ * @author Mahmoud Ben Hassine
+ * @since 7.0
+ * @see CompositeRetryListener
+ */
+public interface RetryListener {
+
+	/**
+	 * Called before every retry attempt.
+	 */
+	default void beforeRetry(RetryExecution retryExecution) {
```
sbrannen, line 37: Please document the `RetryExecution` parameter in all methods in this interface.

spring-core/src/main/java/org/springframework/core/retry/RetryListener.java
```diff
@@ -0,0 +1,62 @@
...
+
+	}
+
+	/**
+	 * Called after the first successful retry attempt.
+	 * @param result the result of the callback
+	 * @param <T> the type of the result
+	 */
+	default <T> void onRetrySuccess(RetryExecution retryExecution, T result) {
```
sbrannen: Does `<T>` provide real benefit here?

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
```diff
@@ -0,0 +1,178 @@
...
+	 * Call the retry callback according to the configured retry and backoff policies.
+	 * If the callback succeeds, its result is returned. Otherwise, a {@link RetryException}
+	 * will be thrown to the caller having all attempt exceptions as suppressed exceptions.
+	 * @param retryCallback the callback to call initially and retry if needed
+	 * @param <R> the type of the result
+	 * @return the result of the callback if any
+	 * @throws RetryException thrown if the retry policy is exhausted. All attempt exceptions
+	 * are added as suppressed exceptions to the final exception.
+	 */
+	@Override
```
sbrannen, line 127: Is this Javadoc different than the Javadoc declared in the interface?

If not, you can remove the duplicated Javadoc.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
```diff
@@ -0,0 +1,178 @@
...
+	public <R extends @Nullable Object> R execute(RetryCallback<R> retryCallback) throws RetryException {
+		Assert.notNull(retryCallback, "Retry Callback must not be null");
+		String callbackName = retryCallback.getName();
+		// initial attempt
+		try {
+			logger.debug("About to execute callback '" + callbackName + "'");
```
sbrannen:
```suggestion
			logger.debug(() -> "About to execute callback '" + callbackName + "'");
```

If we're going to use `LogAccessor`, we have to use the `Supplier` variants of the logging methods for any dynamic log message.

Otherwise, we should use a standard logger and `LogMessage`.

sbrannen (changes requested):

spring-core/src/main/java/org/springframework/core/retry/support/MaxRetryAttemptsPolicy.java
fmbenhassine: That's right. Removed.

spring-core/src/main/java/org/springframework/core/retry/support/PredicateRetryPolicy.java
fmbenhassine: I don't know which modification led to that. Fixed.

spring-core/src/main/java/org/springframework/core/retry/RetryListener.java
fmbenhassine: Done.

spring-core/src/main/java/org/springframework/core/retry/RetryListener.java
fmbenhassine: I think this was copied from the previous interface in Spring Retry. Indeed, I see no benefit for this generic type here, so I removed it.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
fmbenhassine: Yes, it's different. It doesn't mention "according to the retry policy configured at the implementation level" and also has more details about suppressed exceptions.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
fmbenhassine: Good catch. I updated the code to use the `Supplier` variants.

spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java
fmbenhassine: I am missing something here, because I still don't see how we can achieve the same result using the standard logger and `LogMessage` (especially the test `if (logger.isTraceEnabled()) {...}`). But we can discuss that offline.

spring-core/src/main/java/org/springframework/core/retry/RetryCallback.java
fmbenhassine: After our discussion, there is no need to introduce another interface.

> Not requirement, but I wonder if we can use a default convenient implementation here to the getClass().getName()?

I agree, @artembilan . I have used that as a default implementation for now. But I will let the team decide if we should return null and adapt the log messages accordingly.

artembilan: Well, the `name` of callback is not only about logging.
We have in Spring Retry a `MetricsRetryListener` which uses that like this:
```
		String label = Objects.requireNonNullElse(callback.getLabel(), callback.getClass().getName());
		Tags retryTags = Tags.of("name", label)
```
Honestly, I really anticipate similar request here, so we might need to think how to propagate this name down to the listener.
Probably as a part of that `RetryExecution` volatile object for the current retry cycle.

spring-core/src/main/java/org/springframework/core/retry/support/CompositeRetryListener.java
```diff
@@ -0,0 +1,85 @@
...
+		this.listeners = List.copyOf(listeners);
+	}
+
+	/**
+	 * Add a new listener to the list of delegates.
+	 * @param listener the listener to add. Must not be <code>null</code>.
+	 */
+	public void addListener(RetryListener listener) {
+		Assert.notNull(listener, "Retry listener must not be null");
+		this.listeners.add(listener);
```
artembilan, line 62:
```
this.listeners = List.copyOf(listeners);
```
So, this will fail.
See it implementation:
```
    static <E> List<E> copyOf(Collection<? extends E> coll) {
        return ImmutableCollections.listCopy(coll);
    }
```
I think something like:
```
private final List<RetryListener> listeners  = new LinkedList<>();
```
And then in that `CompositeRetryListener(List<RetryListener> listeners)` ctor:
```
this.listeners.addAll(listeners);
```
would be enough.

spring-core/src/main/java/org/springframework/core/retry/RetryCallback.java
fmbenhassine: Spring Retry passes the callback itself to the listener methods (just to get the label), which is not very convenient in my opinion. But I agree, the `RetryExecution` object can be enriched with the callback name to give some context about the execution if needed.

artembilan (changes requested):

spring-core/src/main/java/org/springframework/core/retry/support/CompositeRetryListener.java
sbrannen: Good catch! 👍 

I've fixed that locally.

sbrannen:
This has been merged into `main` in 3fb4a75ae4b34fc6ac80fcc01eaba5caf6ae78a5 and slightly revised in 02af9e5ceedccf2326b5123c59c59d100e0f9447. 🚀 

Thanks everybody for the collaborative effort to get this into 7.0!

PR #34718, opened 2025-04-04
Preventing unmodifiableList from being affected by changes to the original List

Tussle0410:
This PR replaces `Collections.unmodifiableList()` with `List.copyOf()`.

When using `Collections.unmodifiableList()`, if the original list changes, it can affect the unmodifiable list, potentially causing unexpected bugs.

Using `List.copyOf` creates an immutable list by copying the original list, which prevents unexpected bugs and makes the code more readable, improving maintainability.

```
Collections.unmodifiableList(values)
```

I changed the code from the format above to the one below.
```
List.copyOf(values)
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ReadOnlyHttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/ReadOnlyHttpHeaders.java
@@ -86,7 +86,7 @@ class ReadOnlyHttpHeaders extends HttpHeaders {
 	@Override
 	public @Nullable List<String> get(String headerName) {
 		List<String> values = this.headers.get(headerName);
-		return (values != null ? Collections.unmodifiableList(values) : null);
+		return (values != null ? List.copyOf(values) : null);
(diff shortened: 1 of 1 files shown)


bclozel:
This is also likely to make performance worse as we'll allocate list of headers for each request. Rather than discussing this solution, can describe first the problem you've encountered? Maybe you can share a minimal sample application that demonstrates this happening?

Thanks

Tussle0410:
This wasn't a specific problem I encountered, but since it was named ReadOnlyHttpHeaders, I determined there could be reliability issues when using unmodifiableList, which is why I created this PR.

I only considered the advantage of ensuring reliability through List.copyOf(), and I didn't consider the performance implications during the copying process. 

If there's a possibility of performance issues, 

I will close this PR. 

Thank you for your input on this matter."

bclozel:
Thanks for the proposal we will keep things as they are unless there is a concrete case that we can look into.

PR #34719, opened 2025-04-05
Use `implementation` Gradle configuration for `framework-docs` module

izeye:
This PR changes to use `implementation` configuration for the `framework-docs` module as it seems to be sufficient for a documentation module.


sbrannen:

Thanks

PR #34720, opened 2025-04-05
Test escape character support in `PropertySourcesPropertyResolver`

### LINKED ISSUE #34326: Escaped placeholders are evaluated if the PropertySource has placeholder support
This is a follow-up of #34315

Consider the following example:

```properties
prop1=value1
prop2=value2\\${prop1}
```

`prop2` should evaluate to `value2${prop1}` as the placeholder has been escaped.

This works for basic `PropertySource` implementations, but if the source is backed by `PropertySourcesPlaceholderConfigurer` the escaped `${prop1}` is still evaluated even though it should not be.

iifawzi:
Hello, this PR aims to resolve #34326.

Added a new utility method `resolveNestedPlaceholdersWithEscapePreservation`, to preserve escaped placeholders using a regex pattern `\\(\$\{[^}]+\})`, which then masks it and then restores the placeholder after the call to  `resolveNestedPlaceholders` 

This is my first contribution to the project, so I might be missing something, let me know if you think I can help in doing it in any different way.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/env/PropertySourcesPropertyResolver.java
+++ b/spring-core/src/main/java/org/springframework/core/env/PropertySourcesPropertyResolver.java
@@ -18,6 +18,11 @@ package org.springframework.core.env;
 
 import org.jspecify.annotations.Nullable;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 /**
  * {@link PropertyResolver} implementation that resolves property values against
  * an underlying set of {@link PropertySources}.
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @iifawzi,

Please note that your tests pass without any changes to `PropertySourcesPropertyResolver`.

As I have outlined in #34861, the core of the problem lies in the implementation of `PropertySourcesPlaceholderConfigurer`.

However, your proposed tests are still good to have in place. Thus, I merged them in 9b52cfd7d587eb3ce51c072e03aa5b202243746c without the changes to `PropertySourcesPropertyResolver` and revised them in 348b4cd067d42b637784e97d82c7d34236a3a2de.

Thanks for the contribution

PR #34721, opened 2025-04-05
Handle direct CanncelationException on timeout in JdkClientHttpRequest

### LINKED ISSUE #33973: Random CancellationException instead of ResourceAccessException after upgrading to Spring Boot 3.4.0
Hi Spring-Team,

After upgrading to Spring Boot 3.4.0, we have encountered an issue with RestTemplate where random CancellationExceptions are thrown instead of the expected ResourceAccessException during timeout scenarios.

Observations:
This behavior is inconsistent; sometimes a CancellationException is thrown, and other times a ResourceAccessException is thrown.
This issue did not occur in previous versions of Spring Boot.

Reproduction:
We have created a test case to demonstrate the issue: https://github.com/SimonPNorra/timeout-execption-case

Steps to Reproduce:
Run the test case included in the project and observe the exceptions thrown. You will notice either a CancellationException or a ResourceAccessException occurring randomly.

Expected Behavior:
Consistent behavior with ResourceAccessException thrown during timeout scenarios.

Actual Behavior:
Random exceptions are thrown: either CancellationException or ResourceAccessException.

Please let us know if further information is required.

Thank you.

Regards

giampa91:
CancellationExceptions are thrown instead of the expected ResourceAccessException during timeout scenarios. 
Handle the CancellationExceptions in order to throw an ResourceAccessException when timeout occurred

Closes gh-33973

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
@@ -37,6 +37,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Flow;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.jspecify.annotations.Nullable;
 
@@ -96,12 +97,13 @@ class JdkClientHttpRequest extends AbstractStreamingClientHttpRequest {
 	@Override
 	protected ClientHttpResponse executeInternal(HttpHeaders headers, @Nullable Body body) throws IOException {
 		CompletableFuture<HttpResponse<InputStream>> responseFuture = null;
+		TimeoutHandler timeoutHandler = null;
 		try {
 			HttpRequest request = buildRequest(headers, body);
 			responseFuture = this.httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofInputStream());
 
 			if (this.timeout != null) {
-				TimeoutHandler timeoutHandler = new TimeoutHandler(responseFuture, this.timeout);
+				timeoutHandler = new TimeoutHandler(responseFuture, this.timeout);
 				HttpResponse<InputStream> response = responseFuture.get();
 				InputStream inputStream = timeoutHandler.wrapInputStream(response);
 				return new JdkClientHttpResponse(response, inputStream);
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
```diff
@@ -96,12 +96,13 @@ public URI getURI() {
 	@Override
 	protected ClientHttpResponse executeInternal(HttpHeaders headers, @Nullable Body body) throws IOException {
 		CompletableFuture<HttpResponse<InputStream>> responseFuture = null;
+		TimeoutHandler timeoutHandler = null;
```
rstoyanchev, line 100: There seems to be no purpose to this change that I can see. Is this intended?

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
```diff
@@ -224,6 +228,7 @@ public ByteBuffer map(byte[] b, int off, int len) {
 	private static final class TimeoutHandler {
 
 		private final CompletableFuture<Void> timeoutFuture;
+		private boolean isTimeout=false;
```
rstoyanchev: Likewise here, this flag is never used.

rstoyanchev (changes requested):
There seems to be no purpose to moving the TimeoutHandler

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
giampa91: the flag is used in line 240 and is setted to true, and on default is false

giampa91: sorry my mistake, now this variable has a purpose

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
giampa91: this change was to make the TimeoutHandler readable outside the try

giampa91:
Thank you for your patience @rstoyanchev  ! After a careful review, I realized I hadn’t double-checked the code thoroughly enough.

I’ve now updated the implementation to better distinguish between a cancelled request and a genuine timeout by explicitly checking the timeout variable.

Additionally, I’ve added tests covering both scenarios. While there might still be room for optimization, I believe this approach reflects real-world usage and edge cases.

rstoyanchev (approved):
Okay the intent is more clear now.

Issue #34597, opened 2025-03-14
`@MockitoSpyBean` does not work in tests with `@ContextHierarchy`


tvahrst:
Using `@MockitoSpyBean` in tests with parent-child contexts defined with  `@ContextHierarchy` fails with an `IllegalStateException`:

```
java.lang.IllegalStateException: Unable to select a bean to wrap: there are no beans of type ...
	at org.springframework.test.context.bean.override.BeanOverrideBeanFactoryPostProcessor.wrapBean(BeanOverrideBeanFactoryPostProcessor.java:273)
	at org.springframework.test.context.bean.override.BeanOverrideBeanFactoryPostProcessor.registerBeanOverride(BeanOverrideBeanFactoryPostProcessor.java:116)
	at org.springframework.test.context.bean.override.BeanOverrideBeanFactoryPostProcessor.postProcessBeanFactory(BeanOverrideBeanFactoryPostProcessor.java:100)
```

It seems that the `BeanOverrideBeanFactoryPostProcessor` tries to wrap the target bean for each context in the hierarchy - failing for the context in which the target bean does not exist.

Code to reproduce this issue:

```java
@ContextHierarchy({
	@ContextConfiguration(classes = Config1.class),
	@ContextConfiguration(classes = Config2.class)
})
@ExtendWith(SpringExtension.class)
class TestMockitoSpyBeanJUnit5 {

	// @SpyBean // works
	@MockitoSpyBean
	private MyService myService;

	@Test
	void test() {
		System.out.print(myService.getValue());
	}


	@Configuration
	public static class Config1 {
		@Bean
		public MyService myService() {
			return new MyService();
		}
	}

	@Configuration
	public static class Config2 {
	}


	public static class MyService {
		public String getValue() {
			return "New";
		}
	}
}
```

The exception occurs regardless of the order of `Config1.class` and `Config2.class`.

Issue #33293, opened 2020-04-23
`ApplicationContext` reloaded when `@MockitoBean` and `@ContextHierarchy` are mixed on single class


jonenst:
Hi,
attached is a testcase showing a testcontest that is wrongly reloaded when using ```@MockBean``` in conjunction with using ```@ContextHierarchy``` in single class mode.

The testcase is in the attached zip, in the "notworking" folder. the "working" folder is present only for comparison to illustrate the bug.

run $ ```mvn package``` at the root to run the testcases.
[mockbean-hiearchy-singleclass-bug.zip](https://github.com/spring-projects/spring-boot/files/4522524/mockbean-hiearchy-singleclass-bug.zip)
```
[INFO] aggregator 0.0.1-SNAPSHOT .......................... SUCCESS [  0.559 s]
[INFO] working ............................................ SUCCESS [  2.830 s]
[INFO] notworking 0.0.1-SNAPSHOT .......................... FAILURE [  1.718 s]
```

the testcase shows that a context that should be cached is instead reloaded when multiple classes using the same context as the parent of a contexthiearchy, and using mockbean on a bean in a child context (so the parent should still be reusable):

*DOESN'T WORK*
```java
@ContextHierarchy({
    @ContextConfiguration(classes = ErrorIfContextReloaded.class),
    @ContextConfiguration(classes = DefaultFooService.class),
})
@ExtendWith(SpringExtension.class)
class Demo1ApplicationTests {

    @MockBean
    DefaultFooService fooService;

}
```
```java
@ContextHierarchy({
    @ContextConfiguration(classes = ErrorIfContextReloaded.class),
    @ContextConfiguration(classes = DefaultFooService2.class),
})
@ExtendWith(SpringExtension.class)
class Demo2ApplicationTests {

    @MockBean
    DefaultFooService2 fooService;

}
```

The bug does not happen when using @ContextHiearchy with subclassing:
*WORKS*
```java
@ContextHierarchy({
    @ContextConfiguration(classes = ErrorIfContextReloaded.class),
})
@ExtendWith(SpringExtension.class)
class Demo1ApplicationTests {
}

@ContextHierarchy({ @ContextConfiguration(classes = DefaultFooService.class), })
class Demo1ApplicationChildTests extends Demo1ApplicationTests {
    @MockBean
    DefaultFooService fooService;
}
```

```java
@ContextHierarchy({
    @ContextConfiguration(classes = ErrorIfContextReloaded.class),
})
[...]

PR #34723, opened 2025-04-06
Provide first-class support for Bean Overrides with `@ContextHierarchy`

sbrannen:
## Background

As explained by @wilkinsona in https://github.com/spring-projects/spring-framework/issues/33293#issuecomment-2259901889, the Spring TestContext Framework (TCF) creates a `MergedContextConfiguration` for each `ApplicationContext` that it may create. Furthermore, this applies to every `ApplicationContext` created for each level in a `@ContextHierarchy`.

Thus, by default, all context configuration mechanisms for a given test class will be applied to **all** application contexts created for that test class. Specifically, this means that features such as `@ActiveProfiles`, `@TestPropertySource`, and bean overrides (`@TestBean`, `@MockitoBean`, `@MockitoSpyBean`, etc.) will be applied to each application context.

For features such as `@ActiveProfiles` and `@TestPropertySource`, that makes sense. You probably always want the same bean definition profiles and test property sources to be applied to every application context in a context hierarchy.

However, with bean overrides that is typically not the case. Rather, when you configure a bean override, you expect Spring to override a specific bean in a specific application context in the context hierarchy. Phrased differently, you probably do **not** want Spring to attempt to override matching beans in _all_ application contexts in a context hierarchy.

If Spring does attempt to override matching beans in _all_ application contexts in a context hierarchy, that can lead to unexpected behavior or errors like those reported in #33293 and #34597.

## Proposal

This PR provides first-class support for Bean Overrides (`@MockitoBean`, `@MockitoSpyBean`, `@TestBean`, etc.) with `@ContextHierarchy`.

Specifically, bean overrides can now specify which `ApplicationContext` they target within the context hierarchy by configuring the `contextName` attribute in the annotation. The `contextName` must match a corresponding `name` configured via `@ContextConfiguration`.

For example, the following test class configures the name of the second hierarchy level to be "child" and simultaneously specifies that the `ExampleService` should be wrapped in a Mockito spy in the context named "child". Consequently, Spring will only attempt to create the spy in the "child" context and will not attempt to create the spy in the parent context.

```java
@ContextHierarchy({
    @ContextConfiguration(classes = Config1.class),
    @ContextConfiguration(classes = Config2.class, name = "child")
})
class MockitoSpyBeanContextHierarchyTests {

    @MockitoSpyBean(contextName = "child")
    ExampleService service;

    // ...
}
```

The above example demonstrates how this feature would help to resolve the issue reported in #34597.

A similar approach can be applied to the failing use case reported in #33293, and the `ReusedParentConfigV1Tests` and `ReusedParentConfigV2Tests` test classes in this PR demonstrate that.

## `@Autowired` with Magical Powers

The proposed feature also introduces functionality that is not possible with `@Autowired`. `@Autowired` will always inject a matching bean found in the lowest level of the context hierarchy. However, with bean overrides in different levels of the context hierarchy, you may need to be able to have all of those bean override instances injected into the test class in order to interact with them -- for example, to configure stubbing for a mock.

The following example demonstrates that is possible to inject an `ExampleService` mock from both the "parent" **and** the "child" context simultaneously.

```java
@ContextHierarchy({
    @ContextConfiguration(classes = Config1.class, name = "parent"),
    @ContextConfiguration(classes = Config2.class, name = "child")
})
class MockitoBeanContextHierarchyTests {

    @MockitoBean(contextName = "parent")
    ExampleService serviceInParent;

    @MockitoBean(contextName = "child")
    ExampleService serviceInChild;

    // ...
}
```

## Ramifications

In order to support reliable context caching and field injection, the `contextName` attribute contributes to `MergedContextConfiguration` cache key. Thus, bean overrides that would otherwise appear to be logically equivalent are not considered equivalent if they specify different context names. Consequently, specifying different context names can result in an application context being loaded multiple times.

## Related Issues

- #33293
- #34597
- #34726

### DIFF
--- a/framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-mockitobean.adoc
+++ b/framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-mockitobean.adoc
@@ -47,6 +47,21 @@ the same bean in several test classes, make sure to name the fields consistently
 creating unnecessary contexts.
 ====
 
+[WARNING]
+====
+Using `@MockitoBean` or `@MockitoSpyBean` in conjunction with `@ContextHierarchy` can
+lead to undesirable results since each `@MockitoBean` or `@MockitoSpyBean` will be
+applied to all context hierarchy levels by default. To ensure that a particular
+`@MockitoBean` or `@MockitoSpyBean` is applied to a single context hierarchy level, set
+the `contextName` attribute to match a configured `@ContextConfiguration` name – for
+example, `@MockitoBean(contextName = "app-config")` or
+`@MockitoSpyBean(contextName = "app-config")`.
+
+See
+xref:testing/testcontext-framework/ctx-management/hierarchies.adoc#testcontext-ctx-management-ctx-hierarchies-with-bean-overrides[context
+hierarchies with bean overrides] for further details and examples.
+====
+
 Each annotation also defines Mockito-specific attributes to fine-tune the mocking behavior.
 
 The `@MockitoBean` annotation uses the `REPLACE_OR_CREATE`
--- a/framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-testbean.adoc
+++ b/framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-testbean.adoc
@@ -31,6 +31,19 @@ same bean in several tests, make sure to name the field consistently to avoid cr
 unnecessary contexts.
 ====
 
+[WARNING]
+====
+Using `@TestBean` in conjunction with `@ContextHierarchy` can lead to undesirable results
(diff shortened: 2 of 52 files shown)


sbrannen:
This feature will be available in the upcoming 6.2.6 release and is available for early testing now in 6.2.6 snapshots.

Documentation and examples can be viewed here:

- [Context hierarchies with bean overrides](https://docs.spring.io/spring-framework/reference/6.2-SNAPSHOT/testing/testcontext-framework/ctx-management/hierarchies.html#testcontext-ctx-management-ctx-hierarchies-with-bean-overrides)
- [`@MockitoBean` and `@MockitoSpyBean`](https://docs.spring.io/spring-framework/reference/6.2-SNAPSHOT/testing/annotations/integration-spring/annotation-mockitobean.html) (see WARNING block)

PR #34732, opened 2025-04-09
Prevent cache pollution by storing only the factories

philwebb:
Update `SpringFactoriesLoader` so that the cache stores only the factories and not the complete loader.

Prior to this commit, if a cache entry was added with the thread conect classloader, the loader instance would a added and the classloader stored. If the thread context classloader subsequently changes, and a call is made to `forDefaultResourceLocation` with `null` for the classloader, the cached entry would be used which contains the older classloader.


sdeleuze:
Sorry for the delay @philwebb, I think we missed that one, I will work on merging it into `6.2.x`.

PR #34739, opened 2025-04-10
Make HttpServiceGroupConfigurer extend Ordered.

OlgaMaciaszek:


rstoyanchev (approved):
The `HttpServiceConfigurer`s introduced in #34739 are meant to be applied in order, so it does make sense for each to express its own order.

PR #34742, opened 2025-04-11
modified the repetitive get method calls

Tussle0410:
I've modified the `DefaultClientRequestObservationConvention.class` to reduce redundant calls to `getCarrier()` and `getResponse()` in `ClientRequestObservationContext.`

Previously, `getCarrier()` was called 8 times and `getResponse()` was called 3 times throughout the code.

I've modified the implementation to retrieve the ClientHttpRequest and ClientHttpResponse objects just once per method and then use these references for subsequent operations, making the code more efficient and easier to maintain.

```
	protected KeyValue clientName(ClientRequestObservationContext context) {
		if (context.getCarrier() != null && context.getCarrier().getURI().getHost() != null) {
			return KeyValue.of(LowCardinalityKeyNames.CLIENT_NAME, context.getCarrier().getURI().getHost());
		}
		return CLIENT_NAME_NONE;
	}
```

I've modified the existing code as shown below.
```
	protected KeyValue clientName(ClientRequestObservationContext context) {
		ClientHttpRequest request = getClientHttpRequest(context);
		if (request != null && request.getURI().getHost() != null) {
			return KeyValue.of(LowCardinalityKeyNames.CLIENT_NAME, request.getURI().getHost());
		}
		return CLIENT_NAME_NONE;
	}
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/observation/DefaultClientRequestObservationConvention.java
+++ b/spring-web/src/main/java/org/springframework/http/client/observation/DefaultClientRequestObservationConvention.java
@@ -88,7 +88,7 @@ public class DefaultClientRequestObservationConvention implements ClientRequestO
 
 	@Override
 	public @Nullable String getContextualName(ClientRequestObservationContext context) {
-		ClientHttpRequest request = context.getCarrier();
+		ClientHttpRequest request = getClientHttpRequest(context);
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks for the proposal, but I don't think it's worth introducing methods for single liners like this. Cheers!

PR #34746, opened 2025-04-12
Improve handling of `mainThreadPrefix` and remove redundant null check

zzoe2346:
This Pull Request refactors the `isCurrentThreadAllowedToHoldSingletonLock` method to improve the consistency of `mainThreadPrefix` variable usage and remove a redundant null check.

## Changes
**Consistent Variable Usage**
- After assigning the value of `this.mainThreadPrefix` to the local variable `mainThreadPrefix` at the beginning of the method, the subsequent `if` check now consistently uses the local variable (`if (mainThreadPrefix != null)` instead of if `(this.mainThreadPrefix != null)`).

**Removed Redundant Null Check**

- The `mainThreadPrefix != null` check within the `else if (this.strictLocking == null)` block has been removed.
- This check was redundant because `if (maainThreadPrefix != null)` condition has already passed, guaranteeing the variable is not null at that point.

---

 I hope my PR useful :) 
Thank you.


jhoeller:
Good catch! Thanks for spotting this.

PR #34776, opened 2025-04-17
refactored the toList method to collect(collectors.toList()

Adagedo:
Replaced usage of Stream.toList() with collect(Collectors.toList()) to ensure compatibility with Java 11. This change prevents build errors during compilation with older Java versions while maintaining the same functionality


bclozel:
Spring Framework requires Java 17 so there is no point in making this change. Thanks!

PR #34783, opened 2025-04-20
Correctly expand query param with same name from URI variables array 

### LINKED ISSUE #34781: Unexpected URI variables handling when expanded into already existing query parameter name (UriComponentsBuilder)
With spring-web 6.2.5, I get the following behavior :

```
var uriOk = UriComponentsBuilder.fromUriString(“http://myhost?{pk1}={pv1}&{pk2}={pv2}”).buildAndExpand(“v1”, “v2”, “v3”, “v4”);
System.out.println(uriOk);
```
Output :
```
http://myhost?v1=v2&v3=v4
```

But :
```
var uriKo = UriComponentsBuilder.fromUriString(“http://myhost?{pk1}={pv1}&{pk2}={pv2}”).buildAndExpand(“v1”, “v2”, “v1”, “v4”);
System.out.println(uriKo);
```
Output :
```
http://myhost?v1=v4
```

I was expecting the following output for `uriKo` :
```
http://myhost?v1=v2&v1=v4
```

whlit:
Fixes #34781 

For repeated fields, the original method is to overwrite them instead of adding them to the original array.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/HierarchicalUriComponents.java
+++ b/spring-web/src/main/java/org/springframework/web/util/HierarchicalUriComponents.java
@@ -443,7 +443,7 @@ final class HierarchicalUriComponents extends UriComponents {
 		UriTemplateVariables queryVariables = new QueryUriTemplateVariables(variables);
 		this.queryParams.forEach((key, values) -> {
 			String name = expandUriComponent(key, queryVariables, this.variableEncoder);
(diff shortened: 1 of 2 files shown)


bclozel:
@whlit could you add a unit test for that?

whlit:
> @whlit could you add a unit test for that?

OK, I'll add it

christophejan:
Thank you very much @whlit @bclozel @rstoyanchev
Best regards

PR #34785, opened 2025-04-20
Exclude the duplicate media types in AbstractMessageConverterMethodProcessor::determineCompatibleMediaTypes

lgxbslgx:
Hi all,

This trivial patch excludes the duplicate media types in the method `AbstractMessageConverterMethodProcessor::determineCompatibleMediaTypes`.

Thanks for your review.

Best Regards,
-- Guoxiong


rstoyanchev:
I'm not sure it's that simple as the selected media type may have a quality parameter, and removing some even if duplicate may change the result.

Is there a specific issue that this is causing?

lgxbslgx:
> Is there a specific issue that this is causing?

I debuged the code and found the duplicate items. It doesn't cause a bug locally.

rstoyanchev:
Thanks for clarifying. 

Eliminating duplicates could have a side effect since media types can have parameters. At the same duplicates do not affect the outcome since we expect a single media type to be chosen, so I'm going to decline this.

PR #34789, opened 2025-04-21
Adds option to set Principal in MockServerWebExchange

blake-bauman:
Add a Principal for MockServerWebExchange.Builder to allow for a mock Principal object to be used with unit tests.


blake-bauman:
Use case:  If the class you're testing makes a call to `exchange.getPrincipal()` the current `MockServerWebExchange` will only ever return `Mono.empty()`.

rstoyanchev (approved):
We made a similar change in 6.2.5 to allow setting the ApplicationContext. We can add this as well.

PR #34797, opened 2025-04-22
Use HexFormat in ContentDisposition

kilink:
Switch to using `HexFormat` in ContentDisposition for conversion to / from hex.


bclozel:
Thanks @kilink and sorry for the delay, this is now merged in main for 7.0.0-SNAPSHOTs.

PR #34798, opened 2025-04-22
Use HexFormat in CharacterEditor

kilink:
Use HexFormat to perform conversion for unicode-prefixed Strings in CharacterEditor.


bclozel:
Thanks @kilink and sorry for the delay, this is now merged in main for 7.0.0-SNAPSHOTs.

PR #34799, opened 2025-04-22
Add optimized DataBufferInputStream overrides

kilink:
Add optimized DataBufferInputStream overrides for readNBytes, skip, and transferTo; all of them allocate byte buffers which we can either avoid (in the case of skip) or size more precisely since the number of remaining bytes is known.

Note that `transferTo` could potentially be optimized further; for instance if the passed in OutputStream is a DataBufferOutputStream, it could potentially do some special handling.


bclozel:
Thanks @kilink and sorry for the delay. I merged this back into 6.2.x.

PR #34810, opened 2025-04-24
Pull #34810: align `xmx args`

Pankraz76:
Pull #34810: align `xmx args`

align to same (memory) unit from `mb` to `gb`, as its actually about `gb`.

like done in: https://github.com/checkstyle/checkstyle/pull/16608

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 version=7.0.0-SNAPSHOT
 
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx2048m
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -1,7 +1,7 @@
 version=7.0.0-SNAPSHOT
 
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx2048m
+org.gradle.jvmargs=-Xmx4g
```
Pankraz76, line 4:
```suggestion
org.gradle.jvmargs=-Xmx2g
```

PR #34812, opened 2025-04-24
HttpEntity.EMPTY headers should not be possible to mutate via HttpHeaders constructor

### LINKED ISSUE #34806: HttpEntity.EMPTY may reference mutable headers
Version: 6.2.6

## Problem
Leaky reference in `HttpHeaders` breaks assumptions of immutability surrounding `HttpEntity.EMPTY`.

## Description
When passing an existing instance of `HttpHeaders` to the `HttpHeaders` constructor, a reference to the original is maintained internally, and this is updated when setting headers on the new instance.

When creating a new `HttpHeaders` from `HttpEntity.EMPTY.getHeaders()`, the original headers are updated when updating the new ones, and the original `HttpEntity` is no longer "empty".

## Use case
It is desirable to pass an existing instance of `HttpEntity` to some method, for the purposes of centralising logic to add an `Authorization` header for requests during integration testing. The `HttpEntity` is constructed naively by the test, with the parameters that it defines, then delegates to the method to have a valid auth header added.

Imagine a `GET` request to the endpoint `/some-path`.

The test may call a method like

```
<T> ResponseEntity<T> getAuthenticatedForEntity(String path, Class<T> responseType) {
    return restTemplate
            .exchange(
                path,
                HttpMethod.GET,
                authenticated(HttpEntity.EMPTY),
                responseType
            );
} 
```

where the `authenticated` method will take an existing `HttpEntity` and add a token to the headers. There may be other cases where the `HttpEntity` is defined by the test itself and the test calls the `authenticated` method, so it becomes desirable to base the abstraction, ultimately, around `HttpEntity`.

## Proposal
Deprecate the constant `HttpEntity.EMPTY`, as the semantics are flawed. There cannot be a single instance, due to the way that `HttpHeaders` works.

Either the default constructor on `HttpEntity` should be public, or a new static method `HttpEntity.empty()` should return a new, empty instance.

## Example Repo

https://github.com/sean-hernon/empty-http-entity-issue/

Only a simplified version of the `authenticated` method described above is included, for brevity.

## Workaround

Instead of `HttpEntity.EMPTY`, use `new HttpEntity(){}`.

quaff:
Fix GH-34806

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpEntity.java
@@ -50,6 +50,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Arjen Poutsma
  * @author Juergen Hoeller
+ * @author Yanming Zhou
  * @since 3.0.2
  * @param <T> the body type
  * @see org.springframework.web.client.RestTemplate
@@ -61,7 +62,7 @@ public class HttpEntity<T> {
 	/**
 	 * The empty {@code HttpEntity}, with no body or headers.
 	 */
-	public static final HttpEntity<?> EMPTY = new HttpEntity<>();
+	public static final HttpEntity<?> EMPTY = new HttpEntity<>(HttpHeaders.EMPTY);
 
 
 	private final HttpHeaders headers;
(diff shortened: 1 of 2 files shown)


rstoyanchev (changes requested):
I don't think we should remove the unwrapping in the HttpHeaders constructor, which is an optimization. HttpHeaders already supports an EMPTY instance, and we can update the creation of HttpEntity.EMPTY to use it.

quaff:
> I don't think we should remove the unwrapping in the HttpHeaders constructor, which is an optimization. HttpHeaders already supports an EMPTY instance, and we can update the creation of HttpEntity.EMPTY to use it.

Updated as you suggested.

rstoyanchev (approved):

PR #34820, opened 2025-04-24
Support `@Import` on interfaces

### LINKED ISSUE #34805: Consider implemented interfaces during `@Import` processing
See https://github.com/spring-projects/spring-boot/issues/45252 for background.

`ConfigurationClassParser` does not consider implemented interfaces when finding `@Import` (meta-)annotations. It would enable the scenario described in https://github.com/spring-projects/spring-boot/issues/45252 if it did. I think it would also be more consistent with other areas of Framework where implemented interfaces are considered. The test context framework is one such area.

kwondh5217:
Closes gh-34805

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
@@ -98,6 +98,7 @@ import org.springframework.util.StringUtils;
  * @author Phillip Webb
  * @author Sam Brannen
  * @author Stephane Nicoll
+ * @author Daeho Kwon
  * @since 3.0
  * @see ConfigurationClassBeanDefinitionReader
  */
@@ -549,6 +550,9 @@ class ConfigurationClassParser {
 	 * <p>For example, it is common for a {@code @Configuration} class to declare direct
 	 * {@code @Import}s in addition to meta-imports originating from an {@code @Enable}
 	 * annotation.
+	 * <p>In addition, {@code @Import} annotations declared on interfaces implemented by
+	 * the configuration class are also considered. This allows imports to be triggered
+	 * indirectly via marker interfaces or shared base interfaces.
 	 * @param sourceClass the class to search
 	 * @param imports the imports collected so far
 	 * @param visited used to track visited classes to prevent infinite recursion
@@ -565,6 +569,10 @@ class ConfigurationClassParser {
 				}
 			}
 			imports.addAll(sourceClass.getAnnotationAttributes(Import.class.getName(), "value"));
+
+			for (SourceClass ifc : sourceClass.getInterfaces()) {
+				collectImports(ifc, imports, visited);
+			}
 		}
 	}
 
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
@@ -550,7 +550,7 @@ class ConfigurationClassParser {
 	 * <p>For example, it is common for a {@code @Configuration} class to declare direct
 	 * {@code @Import}s in addition to meta-imports originating from an {@code @Enable}
 	 * annotation.
-	 * <p>In addition, {@code @Import} annotations declared on interfaces implemented by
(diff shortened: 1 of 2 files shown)


spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
```diff
@@ -565,6 +569,10 @@ private void collectImports(SourceClass sourceClass, Set<SourceClass> imports, S
 				}
 			}
 			imports.addAll(sourceClass.getAnnotationAttributes(Import.class.getName(), "value"));
+
+			for (SourceClass ifc : sourceClass.getInterfaces()) {
+				collectImports(ifc, imports, visited);
+			}
```
sbrannen: Please move the interface search above the annotation search on the current class (i.e., directly under `if (visited.add(sourceClass)) {`).

Also, please add a test which verifies that locally declared `@Import` annotations are processed _after_ `@Import` annotations discovered on implemented interfaces. For example, a bean imported from a local `@Import` annotation should be able to override a bean imported via an `@Import` declaration on an implemented interface.

spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
```diff
@@ -549,6 +550,9 @@ private Set<SourceClass> getImports(SourceClass sourceClass) throws IOException
 	 * <p>For example, it is common for a {@code @Configuration} class to declare direct
 	 * {@code @Import}s in addition to meta-imports originating from an {@code @Enable}
 	 * annotation.
+	 * <p>In addition, {@code @Import} annotations declared on interfaces implemented by
```
sbrannen:
```suggestion
	 * <p>As of Spring Framework 7.0, {@code @Import} annotations declared on interfaces implemented by
```

spring-context/src/test/java/org/springframework/context/annotation/ImportSelectorTests.java
```diff
@@ -203,6 +204,38 @@ void invokeAwareMethodsInImportGroup() {
 		assertThat(TestImportGroup.environment).isEqualTo(context.getEnvironment());
 	}
 
+	@Test
+	void importAnnotationOnImplementedInterfaceIsRespected() {
+		AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext();
+		context.register(InterfaceBasedConfig.class);
+		context.refresh();
```
sbrannen:
```suggestion
		AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext(InterfaceBasedConfig.class);
```

sbrannen (changes requested):
Thanks for the PR!

I've requested a few changes.

kwondh5217:
@sbrannen Thanks for the review! 
I’ve updated the code as suggested and added the test. 
Please take another look when you have time 🙇

spring-context/src/test/java/org/springframework/context/annotation/ImportSelectorTests.java
```diff
@@ -203,6 +204,66 @@ void invokeAwareMethodsInImportGroup() {
...
+	@Import(ImportedConfig.class)
+	interface ConfigImportMarker {
+	}
+
+	@Configuration
+	static class InterfaceBasedConfig implements ConfigImportMarker {
+	}
+
+	@Configuration
+	static class OverridingConfig implements ConfigImportMarker {
```
sbrannen, line 237: `OverridingConfig` should be annotated with `@Import` (and import something that overrides the `ImportedBean`) in order for `localImportShouldOverrideInterfaceImport()` to test this use case.

sbrannen (changes requested):
Thanks for making the requested changes!

I've added an additional request to revise the "local `@Import`" use case.

Please also update the copyright header in `ImportSelectorTests`.

Thanks

kwondh5217:
Thanks for the guidance, @sbrannen.
I’ve pushed the update with `@Import` on `OverridingConfig`.
Please take another look when you have time

sbrannen (approved):
Looks good now.

Thanks for making all of the requested changes.

sbrannen:
This has been merged into `main` in a4d5800a6cb1c57c1113af0ba563b5c4f0c0005a and slightly revised in 4581324038094982b266255d4c994f0d7a04d8cb.

Thanks

PR #34821, opened 2025-04-24
Replace AssertionsForClassTypes with Assertions

izeye:
This PR replaces `AssertionsForClassTypes` with `Assertions`.

### DIFF
--- a/spring-test/src/test/java/org/springframework/test/context/support/BootstrapTestUtilsMergedConfigTests.java
+++ b/spring-test/src/test/java/org/springframework/test/context/support/BootstrapTestUtilsMergedConfigTests.java
@@ -21,7 +21,6 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-import org.assertj.core.api.AssertionsForClassTypes;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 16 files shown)


spring-test/src/test/java/org/springframework/test/context/support/BootstrapTestUtilsMergedConfigTests.java
```diff
@@ -271,10 +270,7 @@ public void buildMergedConfigForContextHierarchy() {
 
 		MergedContextConfiguration parent = mergedConfig.getParent();
 		assertThat(parent).as("parent config").isNotNull();
-		// The following does not work -- at least not in Eclipse.
```
izeye, line 274: I'm not sure why it didn't work, but considering that Spring Boot doesn't use the `AssertionsForClassTypes` at all, I expected that it's okay now.

rstoyanchev: @sbrannen would you mind taking a look at this in Eclipse?

sbrannen: Surprise! The Eclipse compiler no longer has a issue with that. 👍

sbrannen (approved):

sbrannen:

Thanks

PR #34828, opened 2025-04-25
Provide a working example instead of unclear placeholders

Artur-:


rstoyanchev (commented):
Thanks for the suggestions. 

We can't be too prescriptive about how the processing chain. These are generic operators and their use depends on the application needs.

That said we could incorporate some of this to provide general guidance and suggestions for how each operator might be used.

Artur-:
The challenge I faced with the javadoc was not having any clue what to do instead of "// ..." so some kind of comment or example would be better than the current one. Even though it's not a "you always want this" type of code snippet

rstoyanchev:
Understood, but this is more about understanding the (general purpose), Flux operators. Use doOnNext for anything imperative that doesn't block nor return a value. Use map for imperative transformations. Use concatMap for anything that requires potentially asynchronous handling. The rest is up to you. Again, I appreciate the suggestion and will incorporate something, but will also finetune a bit.

PR #34864, opened 2025-05-07
Add `@since` tags for new PatternMatchUtils.simpleMatchIgnoreCase() methods

izeye:
This PR adds Javadoc `@since` tags for `PatternMatchUtils.simpleMatchIgnoreCase()` methods.

See gh-34802


sbrannen:
`@since` tags should be manually backported to `6.1.x` as well, since the backport bot does not support PRs.

rstoyanchev:
I have applied this separately to 6.1.x and 6.2.x and linking to #34802. Thanks for spotting this @izeye.

PR #34869, opened 2025-05-08
Align ClassFileClassMetadata with other implementations

philwebb:
Return null for getSuperClassName() with package-info classes:

Upgrading Spring Boot to use Framework 7 hit an issue with `spring-hateoas` and AOT hints.

```
at org.springframework.aot.hint.SimpleTypeReference.of(SimpleTypeReference.java:47)	
	at org.springframework.aot.hint.TypeReference.of(TypeReference.java:85)	
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:215)	
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:215)	
	at java.util.Iterator.forEachRemaining(Iterator.java:133)	
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939)	
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570)	
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560)	
	at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:153)	
	at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:176)	
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:265)	
	at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:636)	
	at org.springframework.hateoas.aot.AotUtils.registerTypesForReflection(AotUtils.java:109)
```

This is ultimately caused by a change in behavior when running on Java 24. With earlier versions, the `registerTypesForReflection` uses a scanner that uses a `SimpleAnnotationMetadataReader`. This scanner always returns `null` for `getSuperClassName()` results from `package-info` classes. It appears that `ClassFileClassMetadata` returns `Object.class`.

This seems like a possible bug in the new class file Java API, but I think we should add a workaround in our code as well.

### DIFF
--- a/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileClassMetadata.java
+++ b/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileClassMetadata.java
@@ -72,7 +72,7 @@ class ClassFileClassMetadata implements AnnotationMetadata {
 		this.className = className;
 		this.accessFlags = accessFlags;
 		this.enclosingClassName = enclosingClassName;
-		this.superClassName = superClassName;
+		this.superClassName = (!className.endsWith(".package-info")) ? superClassName : null;
 		this.independentInnerClass = independentInnerClass;
 		this.interfaceNames = interfaceNames;
 		this.memberClassNames = memberClassNames;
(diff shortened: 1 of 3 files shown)


philwebb:
I found another quirk with member classes so I've added a second commit to this PR.

bclozel:
Thanks @philwebb !

Issue #34768, opened 2025-04-16
R2DBC `NamedParameterUtils` does not properly expand reused collection parameter


Dieken:
https://github.com/spring-projects/spring-framework/blob/v7.0.0-M3/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/NamedParameterUtils.java#L528

```java
if (parameter.getValue() instanceof Collection collection) {
				Iterator<Object> iterator = collection.iterator();
				Iterator<BindMarker> markers = bindMarkers.iterator();
				while (iterator.hasNext()) {
					Object valueToBind = iterator.next();
					if (valueToBind instanceof Object[] objects) {
						for (Object object : objects) {
							bind(target, markers, object);
						}
					}
					else {
						bind(target, markers, valueToBind);
					}
				}
			}
```

For example,  for `SELECT ... WHERE  fund_codeA in (:fundCodes) OR fund_codeB in (:fundCodes)`,   `fundCodes` references a List,   then the code above only fill the first occurrence of `:fundCodes`.

```java
parameter = ["00001"]
markers = [?, ?]    // two occurrence

// iterate on parameter won't fill second "?"
// SELECT .... WHERE fund_codeA in (?)   OR fund_codeB in (?)

```

Dieken:
Possible fix below,  sorry my network speed to download gradle and dependencies is too slow, I haven't test it.

```
$ git diff -w
diff --git a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/NamedParameterUtils.java b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/NamedParameterUtils.java
index f6696e8d9c..91ae7b35a4 100644
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/NamedParameterUtils.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/NamedParameterUtils.java
@@ -529,6 +529,7 @@ abstract class NamedParameterUtils {
                        if (parameter.getValue() instanceof Collection collection) {
                                Iterator<Object> iterator = collection.iterator();
                                Iterator<BindMarker> markers = bindMarkers.iterator();
+                               do {
                                        while (iterator.hasNext()) {
                                                Object valueToBind = iterator.next();
                                                if (valueToBind instanceof Object[] objects) {
@@ -540,6 +541,13 @@ abstract class NamedParameterUtils {
                                                        bind(target, markers, valueToBind);
                                                }
[...]

PR #34873, opened 2025-05-09
Fix for parameter position Error that occurs when binding the same Collection multiple times in R2DBC DatabaseClient

Tussle0410:
While developing with R2DBC DatabaseClient, an error occurred with the message 'Parameter at position 3 is not set' during the process of binding a Collection multiple times.

[Example SQL]
```
SELECT * 
FROM PERSON
WHERE name IN (:ids)
GROUP BY address

UNION

SELECT * 
FROM PERSON
WHERE name NOT IN (:ids)
```
[Execute Code]
```java
    List<String> ids = List.of("1", "2", "3");

    return databaseClient.sql(query)
        .bind("ids", ids)
        .map((row, rowMetadata) -> row.get("name", String.class))
        .all();

```

[Exception]
```
java.lang.IllegalStateException: Parameter at position 3 is not set
```
When debugging the related code, we confirmed that while it was retrieving the marking and index information, it was replacing the Collection values with information that had only been marked once.

We modified the code that replaces Collections according to marking and index information to ensure it doesn't execute only once, and we've written related test code.

[Modified]
When binding a Collection multiple times, n `BindMarkers` are created for the collection, so we modified the code to repeat the marking process if there is remaining mark information for the collection, ensuring the binding process executes n times.

### DIFF
--- a/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/NamedParameterUtils.java
+++ b/spring-r2dbc/src/main/java/org/springframework/r2dbc/core/NamedParameterUtils.java
@@ -526,17 +526,19 @@ abstract class NamedParameterUtils {
 				return;
 			}
 			if (parameter.getValue() instanceof Collection collection) {
-				Iterator<Object> iterator = collection.iterator();
 				Iterator<BindMarker> markers = bindMarkers.iterator();
-				while (iterator.hasNext()) {
-					Object valueToBind = iterator.next();
-					if (valueToBind instanceof Object[] objects) {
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @Tussle0410,

Thanks for raising the issue and simultaneously submitting a proposed fix.

However, this issue was already reported a few weeks ago in #34768, and I just pushed a fix in dd6eede243b92dbbbc3790516f2a439f2941a8f4.

In light of that, I am closing this as a ...

- Duplicate of #34768

PR #34883, opened 2025-05-12
Remove unnecessary thenReturn in R2dbcTransactionManager

Tussle0410:
The onErrorResume() method uses the con value, but this value is not data received from thenReturn(), and since it doesn't use the data after thenReturn() chaining but proceeds with then(), I determined it was unnecessary and removed it.


rstoyanchev (approved):
Indeed, `doBegin` is a `Mono<Void>` method and the `connectMono` chain returned from it [ends with then()](https://github.com/spring-projects/spring-framework/blob/41b1dff8bea2eea31fe14705f6fc76f98067b165/spring-r2dbc/src/main/java/org/springframework/r2dbc/connection/R2dbcTransactionManager.java#L231) that effectively ignores the value.

rstoyanchev:
@mp911de could you have a quick look at this? It seems rather obviuos.

PR #34893, opened 2025-05-13
Forward more methods to underlying InputStream in NonClosingInputStream

kilink:
NonClosingInputStream extends FilterInputStream, which does not forward some newer InputStream methods such as transferTo and readAllBytes. Specific InputStream implementations may have more optimized methods (e.g., FileInputStream).


bclozel:
Thanks @kilink , this will be in the next maintenance release.

PR #34917, opened 2025-05-19
Set the access modifier of the MultipartFileResource class to public

joerulau:


bclozel:
This is not needed. If you have a MultiPartFile instance already, you can call getResource on it. All methods implemented on this class are inherited.

joerulau:
Oh, sorry for the extra workload.

PR #34921, opened 2025-05-19
Trung | Added StringUtils::resizeConsecutiveChars() and related Unit Tests

trungpnt-byte:
Added a new string utility method to enrich the existing xD

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -1416,4 +1416,63 @@ public abstract class StringUtils {
 		return charSequence.toString();
 	}
 
+	/**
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal, but [as noted in our Javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StringUtils.html), this utils class in mainly for internal usage in Framework and we don't intend to provide here a general utility library for applications to use.

trungpnt-byte:
TYVM for the reply @bclozel , it was my first time contributing to open-source projects 😄

bclozel:
Thanks for your contribution @trungpnt-byte 

This has been declined not because the code isn't useful nor good quality, but rather because we need to consider our maintenance workload and we cannot support a general purpose utility library for this.

Issue #34861, opened 2025-05-07
`PropertySourcesPlaceholderConfigurer` placeholder resolution fails in several scenarios


sbrannen:
## Overview

In summary, the implementation of the placeholder resolution algorithm in `PropertySourcesPlaceholderConfigurer` fails in several scenarios, and the root cause for this category of failures has existed since `PropertySourcesPlaceholderConfigurer` was introduced in Spring Framework 3.1.

https://github.com/spring-projects/spring-framework/blob/7a7df6637478607bef0277bf52a4e0a03e20a248/org.springframework.context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java#L126-L135

In the original code above, we see that a `PropertySource` is implemented as an anonymous inner class which delegates to `Environment#getProperty`, which in turn delegates to an internal `PropertySourcesPropertyResolver`, which in turn performs placeholder resolution.

And... that anonymous `PropertySource` is added to the `MutablePropertySources` managed by `PropertySourcesPlaceholderConfigurer`, which wraps those `MutablePropertySources` in its own `PropertySourcesPropertyResolver`, which in turn performs placeholder resolution.

Thus, we effectively have always had a top-level `PropertySourcesPropertyResolver` that indirectly delegates to a nested `PropertySourcesPropertyResolver`, which results in double placeholder parsing and resolution.

And **that** is what leads to a whole category of bugs.

In #27947, I realized that we were lacking proper support for `ignoreUnresolvablePlaceholders` in the nested `PropertySourcesPropertyResolver`, and due to #34315 and #34326 it became apparent that we are currently (as of Spring Framework 6.2.6) lacking support for the configured `escapeCharacter` in the nested `PropertySourcesPropertyResolver` as well.

## Analysis

My research into #34315 and #34326 led me to realize that the reported bug for escaped placeholders being evaluated despite the escaping was in fact due to the same underlying flaw in the core logic of `PropertySourcesPlaceholderConfigurer`.

Namely, we should never have indirectly used or directly created a nested `PropertySourcesPropertyResolver`.

Instead, properties from property sources from the `Environment` should be accessed directly without duplicate/nested placeholder resolution, since the top-level `PropertySourcesPropertyResolver` already handles placeholder resolution.

## Related Issues

- #27947
- #34326
- #34862
- #34865 
- #34315

sbrannen:
Reopening to address build failures in Spring Boot.

[...]

PR #34936, opened 2025-05-23
`PropertySourcesPlaceholderConfigurer` no longer uses `ConversionService` from `Environment`

mbazos:
I noticed an issue with PropertySourcesPlaceholderConfigurer moving from spring framework 6.2.6 to 6.2.7.

Basically there is a new propertyResolver being created but it is not setting the environments configured conversion service. In the context of spring boot this could cause some issues if custom converters are registered as the propertyResolver if it doesn't have a conversion service set then uses the default and would not include any registered custom converters with the environment.

I don't think this PR is the exact correct fix and I didn't add any tests but this did resolve the issue with the propertyResolver having the environments conversion service. I figured this would be the best way to show the problem and how to fix it. If I have more time I can do a proper PR but might need some guidance if the fix is not in the right place.

If you look at PropertySourcesPlaceholderConfigurer in spring 6.2.6 you will see that the environments conversion service is set on the propertyResolver.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java
+++ b/spring-context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java
@@ -172,6 +172,10 @@ public class PropertySourcesPlaceholderConfigurer extends PlaceholderConfigurerS
 	protected void processProperties(ConfigurableListableBeanFactory beanFactoryToProcess,
 			ConfigurablePropertyResolver propertyResolver) throws BeansException {
 
+		if(this.environment instanceof ApplicationServletEnvironment) {
+			propertyResolver.setConversionService(((ApplicationServletEnvironment) this.environment).getConversionService());
+		}
+		
 		propertyResolver.setPlaceholderPrefix(this.placeholderPrefix);
 		propertyResolver.setPlaceholderSuffix(this.placeholderSuffix);
 		propertyResolver.setValueSeparator(this.valueSeparator);


spring-context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java
```diff
@@ -172,6 +172,10 @@ protected ConfigurablePropertyResolver createPropertyResolver(MutablePropertySou
 	protected void processProperties(ConfigurableListableBeanFactory beanFactoryToProcess,
 			ConfigurablePropertyResolver propertyResolver) throws BeansException {
 
+		if(this.environment instanceof ApplicationServletEnvironment) {
+			propertyResolver.setConversionService(((ApplicationServletEnvironment) this.environment).getConversionService());
+		}
+
```
sbrannen, line 178:
```suggestion
		if (this.environment instanceof ConfigurableEnvironment configurableEnvironment) {
			propertyResolver.setConversionService(configurableEnvironment.getConversionService());
		}
```

I looked into a possible fix before looking at your proposal, and I first came up with the analogous solution above.

We cannot use `ApplicationServletEnvironment` here, since that class is specific Spring Boot and Servlet deployments.

Thus, downcasting to `ConfigurableEnvironment` using `instanceof` pattern matching would be the correct way to do that.

sbrannen:
Hi @mbazos,

And thanks for reporting the issue, which appears to be a regression caused by the changes made in conjunction with #34861.

After having pondered it a bit more, I think your proposal (as well as the similar one I posted) may not be the best approach to restoring the original behavior in `6.2.x`.

If we set the `ConversionService` on the supplied `ConfigurablePropertyResolver` in `processProperties(...)`, that `ConversionService` would then apply to all `PropertySources` which would include _local_ properties.

However, that was not the previous behavior.

Previously, the `ConversionService` used by the `Environment` was only applied to `PropertySources` in the `Environment`, and I plan to reinstate that behavior.

We _could_ consider applying a custom `ConversionService` to `PropertySourcesPlaceholderConfigurer` in general, but that would be a new feature and a change in behavior that we would likely only want to apply in Spring Framework 7.0.

----

At this point, I will introduce tests to confirm that this is a regression, and if it's a regression I will reinstate the previous behavior (automatic conversion of values from `PropertySources` in the `Environment` using the `ConversionService` configured in the `Environment`).

In light of the above, I plan to decline this PR; however, I will leave it open until I have completed the aforementioned tasks.

Cheers,

Sam

spring-context/src/main/java/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.java
mbazos: Thanks something didn't feel right with the change I made

mbazos:
Thanks @sbrannen appreciate the thorough and prompt review. Let me know if you need anything else.

sbrannen:
@mbazos, I did in fact go with a different approach to solving this regression (see commit 90be94a4a548b11dc12be38f1d15f867ad90c611).

However, I didn't see any need to create a separate GitHub issue to track the regression you reported, so I repurposed this PR as an "issue" so to speak.

kmandeville:
> @mbazos, I did in fact go with a different approach to solving this regression (see commit [90be94a](https://github.com/spring-projects/spring-framework/commit/90be94a4a548b11dc12be38f1d15f867ad90c611)).
> 
> However, I didn't see any need to create a separate GitHub issue to track the regression you reported, so I repurposed this PR as an "issue" so to speak.

Is this commit part of an upcoming Spring 6.2.x release? Or will it only be part of the version 7 release?

sbrannen:
> Is this commit part of an upcoming Spring 6.2.x release? Or will it only be part of the version 7 release?

@kmandeville, as you can see on the right-hand side of this page, this was fixed and included in the 6.2.8 release.

PR #34939, opened 2025-05-24
Correct sample in multipart-forms.adoc

BowieDu:
When the file type in the command object is `MultipartFile`, an HTTP status code 400 will be returned. Should it be `FilePart` instead?


rstoyanchev (approved):
Indeed on the WebFlux side it's not `MultipartFile` but implementations of `org.springframework.http.codec.multipart.Part`.

Issue #34745, opened 2025-04-12
JAXB message converters ignore Content-Type charset


pax95:
Hi

I have created a Stackoverflow with the details here https://stackoverflow.com/questions/79568225/restclient-message-converters-unable-to-parse-iso-8859-1-response-from-remote-se

It seems that setting the DefaultCharset on RestClient messageconverters doesn't have any effeft when marshalling a iso-8859-1  response.

Currently the workaround is to take the response as a string and then manually unmarshall to jaxb objects

```
var xml = woRestClient.get()
            .uri("myuri")
            .retrieve()
            .body(String.class);
        var programs = JaxbUtils.unmarshall(xml, Programs.class);
```
 
There is a simelar question here https://stackoverflow.com/questions/77663426/restclient-how-to-correctly-unmarshall-text-xml-response

pax95:
I currently use the workaround described in stacoverflow, and will leave the issue open to see if there is any way to change the messageconverter behaviour to accept a iso bytestream, and convert it, if the defaultCharset has been set on it. 
If not then the workaround is acceptable.

pax95:
Great support. Thanks @bclozel

PR #34940, opened 2025-05-25
Add Javadoc since for gh-34745

izeye:
This PR adds Javadoc `@since` tags for the changes made in gh-34745.


bclozel:
Thanks @izeye !

PR #34944, opened 2025-05-26
replaced ex with exception for easy code understanding or reviews.

Adagedo:


rstoyanchev:
Thanks for the suggestion, but we use `ex` throughout the codebase. This is a long established convention.

PR #34949, opened 2025-05-26
Fix `REPLY_CHANNEL` header check in `MessageHeaderAccessor`

### LINKED ISSUE #34881: Question about MessageHeaders.REPLY_CHANNEL.endsWith(headerName) usage in MessageHeaderAccessor#verifyType
Hi Spring team,

While reading the source code of org.springframework.messaging.support.MessageHeaderAccessor, I noticed something that seems potentially unintended in the verifyType method:

```java
protected void verifyType(@Nullable String headerName, @Nullable Object headerValue) {
	if (headerName != null && headerValue != null) {
		if (MessageHeaders.ERROR_CHANNEL.equals(headerName) ||
				MessageHeaders.REPLY_CHANNEL.endsWith(headerName)) {
			if (!(headerValue instanceof MessageChannel || headerValue instanceof String)) {
						"'" + headerName + "' header value must be a MessageChannel or String");
			}
		}
	}
}
```

My question is about the use of MessageHeaders.REPLY_CHANNEL.endsWith(headerName) — is this intended to be endsWith, or should it be equals, similar to how ERROR_CHANNEL is compared? Using endsWith might lead to false positives when the header name is a suffix of "replyChannel".

Could you please clarify if this is intentional or if it might be a bug?

Thanks for your great work!

remeio:
Fix gh-34881.

When header name is `hannel` and header value is `0`. 
```java
headerAccessor.setHeader("hannel", 0);
```
It will throw `IllegalArgumentException` because `replyChannel` ends with `hannel` and header value is a integer.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/support/MessageHeaderAccessor.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/support/MessageHeaderAccessor.java
@@ -331,7 +331,7 @@ public class MessageHeaderAccessor {
 	protected void verifyType(@Nullable String headerName, @Nullable Object headerValue) {
 		if (headerName != null && headerValue != null) {
 			if (MessageHeaders.ERROR_CHANNEL.equals(headerName) ||
-					MessageHeaders.REPLY_CHANNEL.endsWith(headerName)) {
+					MessageHeaders.REPLY_CHANNEL.equals(headerName)) {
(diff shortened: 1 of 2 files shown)


sbrannen:
This has been merged into `6.2.x` and `main` in bbf61c74ab889788492e5b24acdd20042705fcd3 and polished in f207c0ed5cb82f71366050906714fca272712c6a.

Thanks

PR #34960, opened 2025-05-28
Add real-world use cases section to WebSocket docs

DhruvTheDev1:
- Added a new subsection under "When to use Websockets" that lists common real-world use cases for Web Sockets such as online gaming, collaboration tools and others.


bclozel:
Thanks for the proposal, but I don't think we need to mention uses cases for all the features we provide. In this case, developers should be in charge of deciding whether the features matches their requirements.

PR #34974, opened 2025-05-31
Fix AOT code generation for autowired inner class constructor

nosan:
Fix index calculation in `AutowiredArgumentsCodeGenerator` for Inner Class

See https://github.com/spring-projects/spring-framework/issues/34973


sbrannen:
Thanks for the PR!

Since the `startIndex` is already properly calculated for an inner class in `InstanceSupplierCodeGenerator.buildGetInstanceMethodForConstructor(...)`, there is indeed no need to adjust the argument indexes within `AutowiredArgumentsCodeGenerator.generateCode(...)`, and I updated the commit message to reflect that: c2d678879faf3f8af446c96d799530d56aff225a

PR #34976, opened 2025-05-31
SseEmitter supports custom MimeType

YunKuiLu:
- Supports using `PRODUCIBLE_MEDIA_TYPES_ATTRIBUTE` to modify the default MediaType(`text/event-stream`).

---
When I use `@GetMapping(produces = "text/event-stream;charset=UTF-8")`, the Content-Type in the response has not been replaced with "text/event-stream;charset=UTF-8".

---
Minimal Reproducible Example:
https://github.com/YunKuiLu/issues-spring-web-sse-utf8

---
This PR keeps the default behavior unchanged but allows developers to change `"text/event-stream"` to `"text/event-stream;charset=UTF-8"` if needed.

---
related:
https://github.com/spring-projects/spring-framework/pull/33529
https://github.com/spring-projects/spring-framework/pull/24632


bclozel:
Thanks for the proposal but UTF-8 is already the default and only allowed encoding for this MIME type.
[The official specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream) does not mention any charset parameter.

I'm declining this issue as a result since it seems everything works as designed already.

PR #34988, opened 2025-06-03
Add NullPointerException checking

ecernice:
Adding null checks to prevent potential NullPointerException. Ensures that classes is validated before calling ClassFileImporter to prevent unexpected runtime errors.


DhruvTheDev1 (commented):
Your checking for null early and throwing a GradleException right away stops the build and gives the user a good log message. However you could use Gradle's in build validation such as [skipWhenEmpty](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/SkipWhenEmpty.html) and [InputFiles](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/InputFiles.html) instead of manually checking.

bclozel:
Thanks for the proposal but the task is skipped when the file collection is empty, so there is no need to check for that when the task is executed.

ecernice:
I see, thank for your explanation.

Issue #34986, opened 2025-06-03
PathPattern#combine throws StringIndexOutOfBoundsException


stefan-hasse:
**Situation**
I've had the situation that a `@Controller` was defined with `@RequestMapping("/*")` to manage all request for that `Dispatcher` (that already defined `urlPatterns="/main/something/*"` to distinguish from other dispatcher).
Now, when the Dispatcher was initializing it was throwing `StringIndexOutOfBoundsException` for a method in that Controller with a `@GetMapping(value = "/x/{val}")` mapping defined.

**Root cause*
I could drill this down to [PathPattern:combine](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/util/pattern/PathPattern.java#L397) ... also reproducible by adding this case to [spring-web/src/test/java/org/springframework/web/util/pattern/PathPatternTests.java:combine](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/test/java/org/springframework/web/util/pattern/PathPatternTests.java#L697)

`assertThat(pathMatcher.combine("/*", "/x/y")).isEqualTo("/x/y");`

**Workaround**
Currently I've solved it by defining the `@Controller` with a `@RequestMapping(/**)` for matching multiple path segments (as I assume it's intended).

**Enhancement**
I am unsure whether it's 
- an enhancement request for a better exception OR 
- a bug in `combine` when instead of `endsWithSeparatorWildcard` it's equal to the wildcard

:)

DhruvTheDev1:
So PathPattern.combine is expecting a pattern ending with /** but doesn't handle /* properly which causes the StringIndexOutOfBoundsException

DhruvTheDev1:
So PathPattern.combine is expecting a pattern ending with /** but doesn't handle /* properly which causes the StringIndexOutOfBoundsException - which is a runtime exception.

Hence, possibly a bug as it crashs when combining /* and /x/y.

PR #34990, opened 2025-06-03
Improve error handling for invalid path pattern combination in PathPattern.combine

DhruvTheDev1:
Addresses #34986

Added an explicit check in PathPattern.combine() for cases where the base pattern ends with "/*" and the incoming pattern contains / (it is not a single segment).

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/pattern/PathPattern.java
+++ b/spring-web/src/main/java/org/springframework/web/util/pattern/PathPattern.java
@@ -397,6 +397,15 @@ public class PathPattern implements Comparable<PathPattern> {
 					this.patternString.substring(0, this.patternString.length() - 2),
 					pattern2string.patternString));
 		}
+		
+		if (this.patternString.endsWith("/*")) {
+			String base = this.patternString.substring(0, this.patternString.length() - 2);
+			if (pattern2string.patternString.contains("/")) {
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal but the issue is already assigned.

DhruvTheDev1:
Need help testing it was my computer was really slow but this was the test I was trying to use:

`	@Test
	void combineSingleSegmentWildcardWithMultiSegmentThrows() {
    PathPatternParser parser = new PathPatternParser();
    PathPattern pattern1 = parser.parse("/hotels/*");
    PathPattern pattern2 = parser.parse("/bookings/extra");
    assertThatThrownBy(() -> pattern1.combine(pattern2))
        .isInstanceOf(IllegalArgumentException.class)
        .hasMessageContaining("Cannot combine single-segment wildcard '/*' with multi-segment path");
}`

`@Test
	void combineSingleSegmentWildcardWithSingleSegmentWorks() {
    PathPatternParser parser = new PathPatternParser();
    PathPattern pattern1 = parser.parse("/hotels/*");
    PathPattern pattern2 = parser.parse("booking");
    PathPattern combined = pattern1.combine(pattern2);
    assertThat(combined.getPatternString()).isEqualTo("/hotels/booking");
}`
This test which is for the original behaviour when combining a pattern with a single segment wildcard '/*' works as expected

PR #34993, opened 2025-06-03
Add RFC 10008 (QUERY HTTP method) support

### LINKED ISSUE #32975: Add support for the HTTP QUERY method
It's probably rather early stages, but it would be nice if at some point Spring's web frameworks supported [the HTTP QUERY method](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html).

desiderantes:
This change introduces support for RFC 10008: The HTTP QUERY Method.

Although initially described as just a "safe GET with a body", the QUERY method unlocks the ability to send complex payloads that do not represent a mutation, and can be handled in a safe, idempotent way. Think of it as a replacement for usecases like complex query parameters in URLs, POST requests that do not mutate anything (but have to be treated as such by all other components in the network roundtrip, so no caching, etc.), and non-mutating HTML Form submission that can use an appropriate encoding for the body.

Fixes #32975

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
+++ b/spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
@@ -185,6 +185,16 @@ public final class MockServerHttpRequest extends AbstractServerHttpRequest {
 	}
 
 	/**
+	 * HTTP QUERY variant. See {@link #get(String, Object...)} for general info.
+	 * @param urlTemplate a URL template; the resulting URL will be encoded
+	 * @param uriVars zero or more URI variables
+	 * @return the created builder
+	 */
+	public static BodyBuilder query(String urlTemplate, @Nullable Object... uriVars) {
+		return method(HttpMethod.QUERY, urlTemplate, uriVars);
+	}
+
+	/**
 	 * Create a builder with the given HTTP method and a {@link URI}.
 	 * @param method the HTTP method (GET, POST, etc)
 	 * @param url the URL
--- a/spring-test/src/main/java/org/springframework/test/web/reactive/server/WebTestClient.java
+++ b/spring-test/src/main/java/org/springframework/test/web/reactive/server/WebTestClient.java
@@ -147,6 +147,13 @@ public interface WebTestClient {
 	RequestHeadersUriSpec<?> options();
 
 	/**
+	 * Prepare an HTTP QUERY request.
+	 * @return a spec for specifying the target URL
+	 * @since 7.1
+	 */
+	RequestBodyUriSpec query();
+
+	/**
 	 * Prepare a request for the specified {@code HttpMethod}.
 	 * @return a spec for specifying the target URL
 	 */
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/assertj/MockMvcTester.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/assertj/MockMvcTester.java
@@ -334,6 +334,21 @@ public final class MockMvcTester {
 	}
 
 	/**
+	 * Prepare an HTTP QUERY request.
+	 * <p>The returned builder can be wrapped in {@code assertThat} to enable
+	 * assertions on the result. For multi-statements assertions, use
+	 * {@link MockMvcRequestBuilder#exchange() exchange()} to assign the
+	 * result. To control the time to wait for asynchronous request to complete
+	 * on a per-request basis, use
+	 * {@link MockMvcRequestBuilder#exchange(Duration) exchange(Duration)}.
+	 * @return a request builder for specifying the target URI
+	 * @since 7.1
+	 */
+	public MockMvcRequestBuilder query() {
+		return method(HttpMethod.QUERY);
+	}
+
+	/**
 	 * Prepare a request for the specified {@code HttpMethod}.
 	 * <p>The returned builder can be wrapped in {@code assertThat} to enable
 	 * assertions on the result. For multi-statements assertions, use
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/client/RestTestClient.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/client/RestTestClient.java
@@ -120,6 +120,13 @@ public interface RestTestClient {
 	RequestHeadersUriSpec<?> delete();
 
 	/**
+	 * Prepare an HTTP QUERY request.
+	 * @return a spec for specifying the target URL
+	 * @since 7.1
+	 */
+	RequestBodyUriSpec query();
+
+	/**
 	 * Prepare an HTTP OPTIONS request.
 	 * @return a spec for specifying the target URL
 	 */
--- a/spring-web/src/main/java/org/springframework/http/RequestEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/RequestEntity.java
@@ -385,6 +385,27 @@ public class RequestEntity<T> extends HttpEntity<T> {
 	}
 
 	/**
+	 * Create an HTTP QUERY builder with the given url.
+	 * @param url the URL
+	 * @return the created builder
+	 * @since 7.1
+	 */
+	public static BodyBuilder query(URI url) {
+		return method(HttpMethod.QUERY, url);
+	}
+
+	/**
+	 * Create an HTTP QUERY builder with the given string base uri template.
+	 * @param uriTemplate the uri template to use
+	 * @param uriVariables variables to expand the URI template with
+	 * @return the created builder
+	 * @since 7.1
+	 */
+	public static BodyBuilder query(String uriTemplate, Object... uriVariables) {
+		return method(HttpMethod.QUERY, uriTemplate, uriVariables);
+	}
+
+	/**
 	 * Create an HTTP PUT builder with the given url.
 	 * @param url the URL
 	 * @return the created builder
--- a/spring-web/src/main/java/org/springframework/http/client/SimpleClientHttpRequestFactory.java
+++ b/spring-web/src/main/java/org/springframework/http/client/SimpleClientHttpRequestFactory.java
@@ -155,7 +155,8 @@ public class SimpleClientHttpRequestFactory implements ClientHttpRequestFactory 
 
 		boolean mayWrite =
 				("POST".equals(httpMethod) || "PUT".equals(httpMethod) ||
-						"PATCH".equals(httpMethod) || "DELETE".equals(httpMethod));
+						"PATCH".equals(httpMethod) || "DELETE".equals(httpMethod) ||
+						"QUERY".equals(httpMethod));
 
 		connection.setDoInput(true);
 		connection.setInstanceFollowRedirects("GET".equals(httpMethod));
--- a/spring-web/src/main/java/org/springframework/web/HttpMediaTypeNotSupportedException.java
+++ b/spring-web/src/main/java/org/springframework/web/HttpMediaTypeNotSupportedException.java
@@ -132,6 +132,9 @@ public class HttpMediaTypeNotSupportedException extends HttpMediaTypeException {
 		if (HttpMethod.PATCH.equals(this.httpMethod)) {
 			headers.setAcceptPatch(getSupportedMediaTypes());
 		}
+		if (HttpMethod.QUERY.equals(this.httpMethod)) {
+			headers.setAcceptQuery(getSupportedMediaTypes());
+		}
 		return headers;
 	}
 
--- a/spring-web/src/main/java/org/springframework/web/HttpMediaTypeNotSupportedException.java
+++ b/spring-web/src/main/java/org/springframework/web/HttpMediaTypeNotSupportedException.java
@@ -132,7 +132,7 @@ public class HttpMediaTypeNotSupportedException extends HttpMediaTypeException {
 		if (HttpMethod.PATCH.equals(this.httpMethod)) {
 			headers.setAcceptPatch(getSupportedMediaTypes());
 		}
-		if (HttpMethod.QUERY.equals(this.httpMethod)) {
+		else if (HttpMethod.QUERY.equals(this.httpMethod)) {
 			headers.setAcceptQuery(getSupportedMediaTypes());
 		}
 		return headers;
--- a/spring-web/src/main/java/org/springframework/web/bind/annotation/QueryMapping.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/annotation/QueryMapping.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-present the original author or authors.
+ * Copyright 2026-present the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,33 +25,33 @@ import java.lang.annotation.Target;
 import org.springframework.core.annotation.AliasFor;
 
 /**
- * Annotation for mapping HTTP {@code GET} requests onto specific handler
+ * Annotation for mapping HTTP {@code QUERY} requests onto specific handler
  * methods.
  *
- * <p>Specifically, {@code @GetMapping} is a <em>composed annotation</em> that
- * acts as a shortcut for {@code @RequestMapping(method = RequestMethod.GET)}.
+ * <p>Specifically, {@code @QueryMapping} is a <em>composed annotation</em> that
+ * acts as a shortcut for {@code @RequestMapping(method = RequestMethod.QUERY)}.
  *
  * <p><strong>NOTE:</strong> This annotation cannot be used in conjunction with
  * other {@code @RequestMapping} annotations that are declared on the same method.
  * If multiple {@code @RequestMapping} annotations are detected on the same method,
  * a warning will be logged, and only the first mapping will be used. This applies
  * to {@code @RequestMapping} as well as composed {@code @RequestMapping} annotations
- * such as {@code @PutMapping}, {@code @PostMapping}, etc.
+ * such as {@code @GetMapping}, {@code @PutMapping}, etc.
  *
- * @author Sam Brannen
- * @since 4.3
- * @see PostMapping
+ * @author Mario Ruiz
+ * @since 7.1
+ * @see GetMapping
  * @see PutMapping
+ * @see PostMapping
  * @see DeleteMapping
  * @see PatchMapping
- * @see QueryMapping
  * @see RequestMapping
  */
 @Target(ElementType.METHOD)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@RequestMapping(method = RequestMethod.GET)
-public @interface GetMapping {
+@RequestMapping(method = RequestMethod.QUERY)
+public @interface QueryMapping{
 
 	/**
 	 * Alias for {@link RequestMapping#name}.
@@ -85,7 +85,6 @@ public @interface GetMapping {
 
 	/**
 	 * Alias for {@link RequestMapping#consumes}.
-	 * @since 4.3.5
 	 */
 	@AliasFor(annotation = RequestMapping.class)
 	String[] consumes() default {};
--- a/spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java
+++ b/spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java
@@ -51,8 +51,8 @@ import org.springframework.core.annotation.AliasFor;
  * at the method level. In most cases, at the method level applications will
  * prefer to use one of the HTTP method specific variants
  * {@link GetMapping @GetMapping}, {@link PostMapping @PostMapping},
- * {@link PutMapping @PutMapping}, {@link DeleteMapping @DeleteMapping}, or
- * {@link PatchMapping @PatchMapping}.
+ * {@link PutMapping @PutMapping}, {@link DeleteMapping @DeleteMapping},
+ * {@link PatchMapping @PatchMapping}, or {@link QueryMapping @QueryMapping}.
  *
  * <p><strong>NOTE:</strong> This annotation cannot be used in conjunction with
  * other {@code @RequestMapping} annotations that are declared on the same element
@@ -75,6 +75,7 @@ import org.springframework.core.annotation.AliasFor;
  * @see PutMapping
  * @see DeleteMapping
  * @see PatchMapping
+ * @see QueryMapping
  */
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
@@ -125,7 +126,7 @@ public @interface RequestMapping {
 
 	/**
 	 * The HTTP request methods to map to, narrowing the primary mapping:
-	 * GET, POST, HEAD, OPTIONS, PUT, PATCH, DELETE, TRACE.
+	 * GET, POST, HEAD, OPTIONS, PUT, PATCH, DELETE, TRACE, QUERY.
 	 * <p><b>Supported at the type level as well as at the method level!</b>
 	 * When used at the type level, all method-level mappings inherit this
 	 * HTTP method restriction.
--- a/spring-web/src/main/java/org/springframework/web/client/RestClient.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestClient.java
@@ -126,6 +126,13 @@ public interface RestClient {
 	RequestHeadersUriSpec<?> options();
 
 	/**
+	 * Start building an HTTP QUERY request.
+	 * @return a spec for specifying the target URL
+	 * @since 7.1
+	 */
+	RequestBodyUriSpec query();
+
+	/**
 	 * Start building a request for the given {@code HttpMethod}.
 	 * @return a spec for specifying the target URL
 	 */
--- a/spring-web/src/main/java/org/springframework/web/context/request/ServletWebRequest.java
+++ b/spring-web/src/main/java/org/springframework/web/context/request/ServletWebRequest.java
@@ -50,7 +50,7 @@ import org.springframework.web.util.WebUtils;
  */
 public class ServletWebRequest extends ServletRequestAttributes implements NativeWebRequest {
 
-	private static final Set<String> SAFE_METHODS = Set.of("GET", "HEAD");
+	private static final Set<String> SAFE_METHODS = Set.of("GET", "HEAD", "QUERY");
 
 	/**
 	 * Date formats as specified in the HTTP RFC.
--- a/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
+++ b/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
@@ -63,9 +63,9 @@ public class CorsConfiguration {
 
 	private static final List<String> DEFAULT_PERMIT_ALL = Collections.singletonList(ALL);
 
-	private static final List<HttpMethod> DEFAULT_METHODS = List.of(HttpMethod.GET, HttpMethod.HEAD);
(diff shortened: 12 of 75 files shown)


desiderantes:
Will keep this one in draft until the standard is published, but I'd appreciate any review

desiderantes:
the [draft](https://queue.rfc-editor.org/final-review/rfc10008/) has been approved for publication, so I'm marking this PR as ready for review.

bclozel:
> the [draft](https://queue.rfc-editor.org/final-review/rfc10008/) has been approved for publication, so I'm marking this PR as ready for review.

That's a 404 now, we can use https://www.rfc-editor.org/info/rfc10008/ instead as a reference.

jonatasrd (approved):

lucaspaixao-dev (approved):

HyunsangHan (approved):

anupamsingh24:
In which spring boot version will this change be available?

anupamsingh24 (approved):

imonraj27 (approved):

spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java
```diff
@@ -51,8 +51,8 @@
  * at the method level. In most cases, at the method level applications will
  * prefer to use one of the HTTP method specific variants
  * {@link GetMapping @GetMapping}, {@link PostMapping @PostMapping},
- * {@link PutMapping @PutMapping}, {@link DeleteMapping @DeleteMapping}, or
- * {@link PatchMapping @PatchMapping}.
+ * {@link PutMapping @PutMapping}, {@link DeleteMapping @DeleteMapping},
+ * {@link PatchMapping @PatchMapping}, or {@link QueryMapping}.
```
JustRed23: Looks like this comment is inconsistent with the rest of the mappings. This would read as `@PutMapping, @DeleteMaping, @PatchMapping or QueryMapping`

bclozel:
We will do a thorough review of this PR and synchronize with other Spring projects. There are many aspects to consider (like the fact that `@QueryMapping` already exists in spring graphql) but we would like to support this in time for Spring Framework 7.1 in November. We will post updates here, no further changes or reviews are needed at this time on this PR.

makungaj1 (approved):

diogorloureiro (approved):

Ali2m (approved):

desiderantes:
I understand a lot of people want this to be merged, after all, it is a very useful addition if I do say so myself. However, at this point approvals from non-committers are becoming background noise, it is hard to know if they're showing support, or if it's a form of spam. You can leave a reaction on the PR description, or [track the ecosystem adoption](https://gist.github.com/desiderantes/2c7e657649cb92672d68e580fb69aa1d) in other projects and get involved in other areas. You can also leave a comment if you spot any issues with this implementation. Otherwise I think it'd be best to heed the comment from @bclozel and await for further updates. Thanks for your interest!

pythonuntermensh (approved):

spring-web/src/main/java/org/springframework/http/client/SimpleClientHttpRequestFactory.java
```diff
@@ -155,7 +155,8 @@ protected void prepareConnection(HttpURLConnection connection, String httpMethod
 
 		boolean mayWrite =
 				("POST".equals(httpMethod) || "PUT".equals(httpMethod) ||
-						"PATCH".equals(httpMethod) || "DELETE".equals(httpMethod));
+						"PATCH".equals(httpMethod) || "DELETE".equals(httpMethod) ||
```
bclozel: Please remove this. The JDK's `HttpURLConnection` does not support QUERY methods and will throw a protocol exception if tried.

spring-web/src/main/java/org/springframework/http/RequestEntity.java
```diff
@@ -384,6 +384,27 @@ public static BodyBuilder post(String uriTemplate, @Nullable Object... uriVariab
 		return method(HttpMethod.POST, uriTemplate, uriVariables);
 	}
 
+	/**
```
bclozel: This is not needed for now. The `method` variant is already enough. We will consider DSLs and syntactic sugar when usage picks up.

spring-web/src/main/java/org/springframework/web/bind/annotation/QueryMapping.java
```diff
@@ -0,0 +1,104 @@
...
+ * @see PostMapping
+ * @see DeleteMapping
+ * @see PatchMapping
+ * @see RequestMapping
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@RequestMapping(method = RequestMethod.QUERY)
+public @interface QueryMapping{
```
bclozel: Please remove all changes from this package. The `@QueryMapping` annotation is not needed for now, `@RequestMapping(method = HttpMethod.QUERY)` will be available with this change.

Note that there is a popular `@QueryMapping` annotation already in the Spring for GraphQL project and this is likely to cause issues when importing types in IDEs.

spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java
```diff
@@ -125,7 +126,7 @@
 
 	/**
 	 * The HTTP request methods to map to, narrowing the primary mapping:
-	 * GET, POST, HEAD, OPTIONS, PUT, PATCH, DELETE, TRACE.
```
bclozel: You can keep this mention of the QUERY method, `@QueryMapping` is not to be considered for now.

spring-web/src/main/java/org/springframework/web/client/RestClient.java
```diff
@@ -125,6 +125,13 @@ public interface RestClient {
 	 */
 	RequestHeadersUriSpec<?> options();
 
+	/**
```
bclozel: Please remove this method. There is a `method(HttpMethod)` variant already and we will consider extensions in the future.

spring-web/src/main/java/org/springframework/web/context/request/ServletWebRequest.java
```diff
@@ -50,7 +50,7 @@
  */
 public class ServletWebRequest extends ServletRequestAttributes implements NativeWebRequest {
 
-	private static final Set<String> SAFE_METHODS = Set.of("GET", "HEAD");
```
bclozel: I don't think we can mark QUERY as a safe method here. [As seen in the spec](https://www.rfc-editor.org/info/rfc10008/#section-2.6), the entire request body has to be considered when processing conditional requests.
There is much more to this, since the spec mentions "Location" headers for stable URLs that can be cached by intermediaries.

Let's focus on the essential for now, we have no guarantee that popular caches and CDNs support this already.

spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
```diff
@@ -63,9 +63,9 @@ public class CorsConfiguration {
 
 	private static final List<String> DEFAULT_PERMIT_ALL = Collections.singletonList(ALL);
 
-	private static final List<HttpMethod> DEFAULT_METHODS = List.of(HttpMethod.GET, HttpMethod.HEAD);
```
bclozel: I don't think we should change the defaults here to the CORS configuration. QUERY is a brand new HTTP method and this would silently support this new method without much advanced warning.

Please remove those changes for now. We can reconsider later.

spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java
```diff
@@ -63,9 +63,9 @@ public class CorsConfiguration {
...
+	private static final List<HttpMethod> DEFAULT_METHODS = List.of(HttpMethod.GET, HttpMethod.QUERY, HttpMethod.HEAD);
 
-	private static final List<String> DEFAULT_PERMIT_METHODS = List.of(HttpMethod.GET.name(),
```
bclozel: Same here. QUERY is not a fetch-spec simple method, and requires a preflight request. This does not belong here. We will consider changes to the CORS support in a separate issue if needed.

spring-web/src/main/java/org/springframework/web/filter/reactive/HiddenHttpMethodFilter.java
```diff
@@ -47,7 +47,7 @@
 public class HiddenHttpMethodFilter implements WebFilter {
 
 	private static final List<HttpMethod> ALLOWED_METHODS =
-			List.of(HttpMethod.PUT, HttpMethod.DELETE, HttpMethod.PATCH);
```
bclozel: This filter exists so HTML forms (which can only send GET/POST) can express PUT/DELETE/PATCH. I don't think we should allow clients to "tunnel" a POST into an idempotent method.

Please remove this change.

spring-web/src/main/java/org/springframework/web/filter/HiddenHttpMethodFilter.java
```diff
@@ -38,7 +38,7 @@
  * is to use a normal POST with an additional hidden form field ({@code _method})
  * to pass the "real" HTTP method along. This filter reads that parameter and changes
  * the {@link HttpServletRequestWrapper#getMethod()} return value accordingly.
- * Only {@code "PUT"}, {@code "DELETE"} and {@code "PATCH"} HTTP methods are allowed.
```
bclozel: Same as the reactive variant, please remove all changes.

spring-web/src/main/java/org/springframework/web/filter/FormContentFilter.java
```diff
@@ -47,7 +47,7 @@
 import org.springframework.util.StringUtils;
 
 /**
- * {@code Filter} that parses form data for HTTP PUT, PATCH, and DELETE requests
```
bclozel: I don't think this is needed for now nor has any credible use case at this point. Please remove changes.

spring-web/src/main/java/org/springframework/web/filter/ShallowEtagHeaderFilter.java
```diff
@@ -151,7 +151,7 @@ protected boolean isEligibleForEtag(HttpServletRequest request, HttpServletRespo
 
 		if (!response.isCommitted() &&
 				responseStatusCode >= 200 && responseStatusCode < 300 &&
-				HttpMethod.GET.matches(request.getMethod())) {
```
bclozel: I don't think the caching/conditional request situation is a easy as that for QUERY requests. Please remove this change.

spring-web/src/main/java/org/springframework/web/server/adapter/DefaultServerWebExchange.java
```diff
@@ -66,7 +66,7 @@
  */
 public class DefaultServerWebExchange implements ServerWebExchange {
 
-	private static final Set<HttpMethod> SAFE_METHODS = Set.of(HttpMethod.GET, HttpMethod.HEAD);
```
bclozel: Same as `ServletWebRequest`, please undo.

spring-web/src/main/java/org/springframework/web/server/UnsupportedMediaTypeStatusException.java
```diff
@@ -161,6 +161,9 @@ public HttpHeaders getHeaders() {
 		if (this.method == HttpMethod.PATCH) {
 			headers.setAcceptPatch(this.supportedMediaTypes);
 		}
+		if (this.method == HttpMethod.QUERY) {
```
bclozel: I think an `else if` would work here. Please update.

spring-web/src/main/java/org/springframework/web/service/annotation/HttpExchange.java
```diff
@@ -51,6 +51,7 @@
  * <li>{@link PutExchange}
  * <li>{@link PatchExchange}
  * <li>{@link DeleteExchange}
+ * <li>{@link QueryExchange}
  * </ul>
  *
  * <p>Supported method arguments:
```
bclozel: I think this can be supported with the `method` attribute. Please remove `@QueryExchange`.

spring-web/src/main/java/org/springframework/web/HttpMediaTypeNotSupportedException.java
```diff
@@ -132,6 +132,9 @@ public HttpHeaders getHeaders() {
 		if (HttpMethod.PATCH.equals(this.httpMethod)) {
 			headers.setAcceptPatch(getSupportedMediaTypes());
 		}
+		if (HttpMethod.QUERY.equals(this.httpMethod)) {
```
bclozel: Same, an `else if` would work better here.

spring-web/src/main/kotlin/org/springframework/web/client/RestOperationsExtensions.kt
```diff
@@ -291,3 +292,44 @@ inline fun <reified T : Any> RestOperations.exchange(url: URI, method: HttpMetho
 @Throws(RestClientException::class)
 inline fun <reified T : Any> RestOperations.exchange(requestEntity: RequestEntity<*>): ResponseEntity<T> =
 		exchange(requestEntity, object : ParameterizedTypeReference<T>() {})
+
```
bclozel: Similar feedback, we don't need extensions for now since this is already possible with other methods. Let's reconsider when the usage increases in the industry.

spring-web/src/test/java/org/springframework/web/bind/annotation/ControllerMappingReflectiveProcessorTests.java
```diff
@@ -213,6 +213,11 @@ Response get() {
 		void post(@RequestBody Request request) {
 		}
 
+		@QueryMapping
```
bclozel: changes in this file are not needed anymore.

spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClient.java
```diff
@@ -123,6 +123,13 @@ public interface WebClient {
 	 */
 	RequestHeadersUriSpec<?> options();
 
+	/**
```
bclozel: Please undo. Similar to previous feedback: we don't need this to enable the HTTP QUERY use case.

spring-webflux/src/main/java/org/springframework/web/reactive/function/server/DefaultServerResponseBuilder.java
```diff
@@ -288,7 +288,7 @@ public Mono<ServerResponse> render(String name, Map<String, ?> model) {
 	 */
 	abstract static class AbstractServerResponse implements ServerResponse {
 
-		private static final Set<HttpMethod> SAFE_METHODS = Set.of(HttpMethod.GET, HttpMethod.HEAD);
```
bclozel: QUERY is not a safe method for conditional checks in Spring Framework. Proper support requires more changes. Please undo.

spring-webflux/src/main/java/org/springframework/web/reactive/function/server/RouterFunctions.java
```diff
@@ -741,6 +741,59 @@ <T extends ServerResponse> Builder OPTIONS(
 		<T extends ServerResponse> Builder OPTIONS(
 				String pattern, RequestPredicate predicate, HandlerFunction<T> handlerFunction);
 
+		/**
```
bclozel: This is already possible with other variants like `add(RequestPredicates.method(HttpMethod.QUERY), function)`. Please remove all DSL additions for now.

spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/ResponseEntityResultHandler.java
```diff
@@ -56,7 +56,7 @@
  */
 public class ResponseEntityResultHandler extends AbstractMessageWriterResultHandler implements HandlerResultHandler {
 
-	private static final Set<HttpMethod> SAFE_METHODS = Set.of(HttpMethod.GET, HttpMethod.HEAD);
```
bclozel: This is being used for conditional requests; please undo.

spring-webmvc/src/main/java/org/springframework/web/servlet/function/RouterFunctions.java
```diff
@@ -655,6 +655,59 @@ <T extends ServerResponse> Builder OPTIONS(
 		<T extends ServerResponse> Builder OPTIONS(
 				String pattern, RequestPredicate predicate, HandlerFunction<T> handlerFunction);
 
+		/**
```
bclozel: Similar feedback to the reactive variant. Please undo.

spring-webmvc/src/main/java/org/springframework/web/servlet/function/AbstractServerResponse.java
```diff
@@ -45,7 +45,7 @@
  */
 abstract class AbstractServerResponse extends ErrorHandlingServerResponse {
 
-	private static final Set<HttpMethod> SAFE_METHODS = Set.of(HttpMethod.GET, HttpMethod.HEAD);
+	private static final Set<HttpMethod> SAFE_METHODS = Set.of(HttpMethod.GET, HttpMethod.QUERY, HttpMethod.HEAD);
```
bclozel: QUERY is not safe for conditional requests right now. Please undo.

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/HttpEntityMethodProcessor.java
```diff
@@ -243,7 +243,7 @@ else if (returnValue instanceof ProblemDetail detail) {
 			outputMessage.getServletResponse().setStatus(returnStatus.value());
 			if (returnStatus.value() == HttpStatus.OK.value()) {
 				HttpMethod method = inputMessage.getMethod();
-				if ((HttpMethod.GET.equals(method) || HttpMethod.HEAD.equals(method)) &&
```
bclozel: This is conditional request support; please undo.

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/HttpEntityMethodProcessor.java
```diff
@@ -292,7 +292,7 @@ private boolean isResourceNotModified(ServletServerHttpRequest request, ServletS
 		HttpHeaders responseHeaders = response.getHeaders();
 		String etag = responseHeaders.getETag();
 		long lastModifiedTimestamp = responseHeaders.getLastModified();
-		if (request.getMethod() == HttpMethod.GET || request.getMethod() == HttpMethod.HEAD) {
+		if (request.getMethod() == HttpMethod.GET || request.getMethod() == HttpMethod.QUERY || request.getMethod() == HttpMethod.HEAD) {
```
bclozel: Same.

spring-test/src/main/java/org/springframework/mock/http/server/reactive/MockServerHttpRequest.java
```diff
@@ -184,6 +184,16 @@ public static BaseBuilder<?> options(String urlTemplate, @Nullable Object... uri
 		return method(HttpMethod.OPTIONS, urlTemplate, uriVars);
 	}
 
+	/**
```
bclozel: Please undo, this is already possible with the `method` variant. No need to expand the API surface for now.

spring-test/src/main/java/org/springframework/test/web/reactive/server/WebTestClient.java
```diff
@@ -146,6 +146,13 @@ public interface WebTestClient {
 	 */
 	RequestHeadersUriSpec<?> options();
 
+	/**
+	 * Prepare an HTTP QUERY request.
+	 * @return a spec for specifying the target URL
+	 * @since 7.1
+	 */
+	RequestBodyUriSpec query();
```
bclozel: Same, this is not needed for now. Please undo.

spring-test/src/main/java/org/springframework/test/web/servlet/assertj/MockMvcTester.java
```diff
@@ -333,6 +333,21 @@ public MockMvcRequestBuilder options() {
 		return method(HttpMethod.OPTIONS);
 	}
 
+	/**
```
bclozel: Please revert.

spring-test/src/main/java/org/springframework/test/web/servlet/client/RestTestClient.java
```diff
@@ -119,6 +119,13 @@ public interface RestTestClient {
 	 */
 	RequestHeadersUriSpec<?> delete();
 
+	/**
```
bclozel: Please revert.

spring-webflux/src/main/java/org/springframework/web/reactive/result/method/RequestMappingInfoHandlerMapping.java
```diff
@@ -413,7 +424,7 @@ private static Set<HttpMethod> initAllowedHttpMethods(Set<HttpMethod> declaredMe
 			}
 			else {
 				Set<HttpMethod> result = new LinkedHashSet<>(declaredMethods);
-				if (result.contains(HttpMethod.GET)) {
+				if (result.contains(HttpMethod.GET) || result.contains(HttpMethod.QUERY)) {
```
bclozel: Why this change? I don't think there is the same equivalence between "HEAD/GET" and "HEAD/QUERY" in HTTP world. HEAD is identical to GET but replies without the body. I'm not seeing any equivalence in the QUERY spec. Please elaborate or revert.

bclozel:
Thanks @desiderantes for tackling this. As you can see, touching HTTP method support in Spring can involve quite a few different places.

Before we can merge, I'd like to cut the scope down substantially. The core of this change should be about teaching the framework that HTTP QUERY exists and that it carries a request body. Anything else should be considered in later stages. This is a brand new standard; we want to collect developer feedback and see how the industry adopts this before significantly expanding the API surface.

I started an inline review but given the many comments, I thought I would summarize my opinion here to make things easier on your side.

## Focus this PR on the following

Please keep the following changes:

* everything in `HttpMethod`, `HttpHeaders` and `RequestMethod` is needed
* `Accept-Query` support on the two "unsupported media type" exceptions, `HttpMediaTypeNotSupportedException`, `UnsupportedMediaTypeStatusException`
* support for reading request bodies from QUERY requests, so that `@RequestBody` and `HttpEntity` work. Keep `AbstractMessageConverterMethodArgumentResolver` and `AbstractMessageReaderArgumentResolver` changes.
* the client-side support in `HttpComponentsClientHttpRequestFactory` is correct
* The `Accept-Query` / OPTIONS wiring in both `RequestMappingInfoHandlerMapping` types
* I think the form body parsing for QUERY in `FormContentFilter` is useful
* all related test updates

## Please revert the CORS-related changes

The changes you made in `CorsConfiguration` are problematic. Expanding `DEFAULT_METHODS` is dangerous here because we're permitting additional HTTP methods for CORS by default, without letting our users know - all for a brand new HTTP method that isn't widely supported yet. Some might see this as a security problem. Definitely not minor release material.

The `DEFAULT_PERMIT_METHODS` part is the Fetch-spec simple method set. I'm not the CORS resident expert but I think that GET, HEAD and POST are the only ones allowed here. Other methods will be sent with a preflight request by browsers. Until the CORS spec evolves, we shouldn't change this.

## Please revert the caching / conditional-request changes

This PR adds QUERY to various lists of "safe methods".

While QUERY is technically a safe method, this isn't what those lists mean. They drive conditional request support in Spring Framework and only look at the actual URL for validation. For QUERY, the request body must be encoded in the caching key when considering conditional requests. This is why QUERY endpoints might want to issue Location headers with a stable URL for a particular resource. This allows clients to send further conditional requests without sending the body again.

Supporting conditional requests and caching in general for QUERY methods would require substantial changes. Right now this doesn't make sense since we don't have use cases for this and I'm not aware of QUERY support for CDNs and intermediaries.

## Please revert the new annotations

Namely `@QueryMapping` and `@QueryExchange`. I'd like to hold off on both.
We should aim to unlock the use of QUERY in Spring Framework, not make it a first class citizen right away. This is a new standard and we don't know much about adoption in the wider industry. We've been burnt by this recently with the "Forwarded" header support (well known standard, poorly adopted in reality). `@QueryMapping` is also already widely used by Spring for GraphQL applications; adding a new annotation with the same name would create IDE import issues in existing applications.

## Please revert the new DSL and builder shortcut methods

The use cases are already possible with `HttpMethod` and other method variants. There is no need to add public API for this right now.
This includes changes in (and related to) `RestClient`, `WebClient`, `WebTestClient`, `RestTestClient`, `RouterFunctions.Builder`, `RequestEntity`, `MockServerHttpRequest` (including in tests).

## Other reverts

* `SimpleClientHttpRequestFactory`, the url connection JDK client does not support QUERY methods and rejects them with protocol exceptions. It's not needed.
* `HiddenHttpMethodFilter` - this filter exists so that HTML forms can express PUT/DELETE/PATCH (using GET or POST). I don't think we should support tunnelling a POST into a safe method. I'd leave both lists alone, but I'm open to hearing the use case if you had one in mind.

## Last remarks

The changes in `RequestMappingInfoHandlerMapping` are questionable.
I understand that HEAD is considered as "a GET without a response body", but here I'm not sure there is such an equivalent for QUERY. With the current implementation, sending HEAD to a QUERY endpoint would probably return "405 method not allowed", whereas an OPTIONS request to that same endpoint would reply with QUERY as an allowed method. Maybe this needs to be refined? Can you explain the rationale here?

bclozel (changes requested):
Please apply the suggested changes in my comment above.

desiderantes:
Hi @bclozel ! Thanks for the review, I've been going through the code fro a while, after all, this PR is a bit old so I had to refresh myself on the changes. I'll address each point:

- CORS-related changes: No problem, and I agree, I'll remove them.
- caching / conditional-request changes: I think that in this case, it is the client who should be in charge of caching for most situations, but i can also envision caching in a QUERY /orders with a json body just like people expect it when they issue a GET /orders?from=2024&to=2026&orderBy=DATE&orderDirection=ASC. Will remove it anyways.
- revert the new annotations: While this certainly diminishes the appeal of the PR, I understand the concerns.
- revert the new DSL and builder shortcut methods: Same as before
- SimpleClientHttpRequestFactory: I guess this is the same situation as with PATCH which was added before. Will remove it.
- HiddenHttpMethodFilter: I disagree. The point is that it is reasonable for someone building a QUERY endpoint to want to support both non-web users as web as web users, and then deal with the same restriction that made this filter exist in the first place. I don't understand the issue with a POST request that does not produce effects, as this is certainly what most people have done before QUERY.
- RequestMappingInfoHandlerMapping: This is certainly an error on my side. According to [A.3. Discovery of QUERY Formats](https://www.rfc-editor.org/info/rfc10008/#name-discovery-of-query-formats) HEAD would be ideal to probe for the Accept-Query header and check the supported formats, so I'll instead implement this behaviour

bclozel:
Thanks for your feedback.

> HiddenHttpMethodFilter: I disagree. The point is that it is reasonable for someone building a QUERY endpoint to want to support both non-web users as web as web users, and then deal with the same restriction that made this filter exist in the first place. I don't understand the issue with a POST request that does not produce effects, as this is certainly what most people have done before QUERY.

Fair enough, let's keep that one in then.

bclozel:
This is now merged and scheduled for the next milestone. You can use our SNAPSHOTs in the meantime to test things out. Cheers @desiderantes for tackling this.

martinbonnin:
@bclozel any chance we could have this in a milestone release?

bclozel:
@martinbonnin this got merged yesterday and is scheduled for the next milestone.

PR #35013, opened 2025-06-08
Allow update of existing `WebSession` after max sessions limit is reached

msnsaeed71:
Previously, when saving a session, the system did not check whether the session ID already existed. As a result, even if the session being saved was an update to an existing one, it was incorrectly treated as a new session, and a "maximum sessions exceeded" error was triggered.

This fix ensures that if a session with the same ID already exists, it will be updated rather than counted as a new session, thereby preventing unnecessary session limit violations.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/session/InMemoryWebSessionStore.java
+++ b/spring-web/src/main/java/org/springframework/web/server/session/InMemoryWebSessionStore.java
@@ -281,7 +281,7 @@ public class InMemoryWebSessionStore implements WebSessionStore {
 		private void checkMaxSessionsLimit() {
 			if (sessions.size() >= maxSessions) {
 				expiredSessionChecker.removeExpiredSessions(clock.instant());
-				if (sessions.size() >= maxSessions) {
+				if (sessions.size() >= maxSessions && !sessions.containsKey(this.getId())) {
 					throw new IllegalStateException("Max sessions limit reached: " + sessions.size());
 				}
 			}
--- a/spring-web/src/test/java/org/springframework/web/server/session/InMemoryWebSessionStoreTests.java
+++ b/spring-web/src/test/java/org/springframework/web/server/session/InMemoryWebSessionStoreTests.java
@@ -30,6 +30,7 @@ import org.springframework.web.server.WebSession;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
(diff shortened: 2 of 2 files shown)


DhruvTheDev1 (commented):
This PR addresses a important logic in session management.
- The updated condition makes sure that only truly new sessions are blocked when the maximum limit is reached.
- If a session with the same ID already exists, it’s treated as an update, not a new session, and is allowed.
- The fix is minimal, targeted, and clearly improves correctness.

Solid improvement!

sbrannen:
In order to process this PR, we will need a test to verify whether the proposed change is necessary.

Thus, please introduce a test that fails before the change and passes after the change.

Thanks

msnsaeed71:
> This PR addresses a important logic in session management.
> 
> * The updated condition makes sure that only truly new sessions are blocked when the maximum limit is reached.
> * If a session with the same ID already exists, it’s treated as an update, not a new session, and is allowed.
> * The fix is minimal, targeted, and clearly improves correctness.
> 
> Solid improvement!

Thank you for the thoughtful review and kind words 🙏
I'm glad the logic refinement is clear and appreciated.

Appreciate your feedback and support!

msnsaeed71:
> In order to process this PR, we will need a test to verify whether the proposed change is necessary.
> 
> Thus, please introduce a test that fails before the change and passes after the change.
> 
> Thanks

This PR adds a test to confirm that updating an existing session (with the same session ID) is still allowed even when the maximum number of sessions has been reached.
Thanks for the guidance. 🙏

sbrannen:
This has been merged into `6.2.x` and `main` in c04902fefbe54e89d423addbf8d724870cf09213 and slightly revised in 4d2cc4ae9720ae70f6a2eacad14fb31cd2edf7dd.

This will also be backported to `6.1.x` in conjunction with #35018.

Thanks for finding this bug and fixing it! 👍

msnsaeed71:
> This has been merged into `6.2.x` and `main` in [c04902f](https://github.com/spring-projects/spring-framework/commit/c04902fefbe54e89d423addbf8d724870cf09213) and slightly revised in [4d2cc4a](https://github.com/spring-projects/spring-framework/commit/4d2cc4ae9720ae70f6a2eacad14fb31cd2edf7dd).
> 
> This will also be backported to `6.1.x` in conjunction with #35018.
> 
> Thanks for finding this bug and fixing it! 👍

Thank you very much for merging the code and for your kind words! I appreciate your careful review and the backport effort as well. Glad I could help fix the bug! 👍

PR #35030, opened 2025-06-11
Polish DataBufferInputStream.skip()

izeye:
The removed `Math.min()` seems to be redundant due to the above `Math.min()`.

See gh-34799


DhruvTheDev1 (commented):
Looks good - skiped is already capped by available(), so no need to check again so removing that extra Math.min cleans up the code.

bclozel:
Thanks @izeye !

PR #35046, opened 2025-06-13
use `try-with-resources` statement

Pankraz76:
Java 7 introduced the try-with-resources statement. This statement ensures that each resource is closed at the end of the statement. It avoids the need of explicitly closing the resources in a finally block. Additionally exceptions are better handled: If an exception occurred both in the try block and finally block, then the exception from the try block was suppressed. With the try-with-resources statement, the exception thrown from the try-block is preserved.

- https://github.com/openrewrite/rewrite-static-analysis/issues/591
- https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html#usetrywithresources
- https://www.baeldung.com/java-try-with-resources
- https://github.com/apache/maven/pull/2426
- https://github.com/quarkusio/quarkus/pull/48364

image

PR #35047, opened 2025-06-13
fix ignored exceptions

Pankraz76:
avoid round abound and just name it ignored like commonly suggested by static code analysis PMD and IDEA.

its already the pattern so just complain:

image

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/ResourceHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/ResourceHttpMessageConverter.java
@@ -158,20 +158,20 @@ public class ResourceHttpMessageConverter extends AbstractHttpMessageConverter<R
 				in.transferTo(out);
 				out.flush();
 			}
-			catch (NullPointerException ex) {
-				// ignore, see SPR-13620
+			catch (NullPointerException ignored) {
+				// see SPR-13620
 			}
 			finally {
 				try {
 					in.close();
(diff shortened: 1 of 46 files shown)


spring-web/src/main/java/org/springframework/http/converter/ResourceHttpMessageConverter.java
```diff
@@ -157,22 +157,17 @@ protected void writeContent(Resource resource, HttpOutputMessage outputMessage)
 				OutputStream out = outputMessage.getBody();
 				in.transferTo(out);
 				out.flush();
+				// need to str
 			}
-			catch (NullPointerException ex) {
```
Pankraz76, line 161: image

Pankraz76: thats why its important to have empty blocks inline.

Pankraz76: - https://github.com/spring-projects/spring-framework/pull/35046
- https://github.com/spring-projects/spring-framework/pull/35046#discussion_r2144695598

as this so important should give this special attention then. @bclozel  

its not about the close as thats the same but about the catch as this seems to differ.

Pankraz76:
thanks

bclozel:
Thanks for your contribution @Pankraz76

PR #35053, opened 2025-06-14
remove false comment about  `try-with-resources` statement in `StompSubProtocolHandler`

Pankraz76:


spring-websocket/src/main/java/org/springframework/web/socket/messaging/StompSubProtocolHandler.java
```diff
@@ -399,8 +399,6 @@ private void sendErrorMessage(WebSocketSession session, Throwable error) {
 		headerAccessor.setMessage(error.getMessage());
 
 		byte[] bytes = this.stompEncoder.encode(headerAccessor.getMessageHeaders(), EMPTY_PAYLOAD);
```
Pankraz76, line 401: primitive data type is not implementing interface `autocloseable`, seems impossible to use `try-with-resources`, right?.

not possible in `L:413` ether, so idk if we still need this comment, as misleading.

image

bclozel:
I don't understand the goal of this PR. The `WebSocketSession` is not a primitive type and the comment is very much on point. We want IO exceptions to be thrown during writing, but we want to silently swallow the ones happening during close. Again, [this link](https://mail.openjdk.org/pipermail/coin-dev/2009-April/001503.html) should tell you more about this pattern.

Pankraz76:
can we merge test then at least to improve cover?

PR #35055, opened 2025-06-15
Document intention of `toString()` in `HandlerMethod`

wonyongg:
The `formatErrorForReturnValue` method in `ServletInvocableHandlerMethod` previously included an unnecessary call to `toString()` and a dangling "in" at the end of the formatted string.

Since the surrounding context already provides sufficient information, both the `toString()` call and the `" in"` suffix have been removed to improve clarity.

No functional changes.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
+++ b/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
@@ -352,6 +352,12 @@ public class HandlerMethod extends AnnotatedMethod {
 		return (this.bean.hashCode() * 31 + super.hashCode());
 	}
 
+	/**
+	 * Returns the description of this handler method.
+	 * This method is used in log and error messages,
+	 * so the returned description should typically include the method signature
+	 * of the underlying handler method for clarity and debugging purposes.
+	 */
 	@Override
 	public String toString() {
 		return this.description;
--- a/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
+++ b/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
@@ -353,10 +353,10 @@ public class HandlerMethod extends AnnotatedMethod {
 	}
 
 	/**
-	 * Returns the description of this handler method.
-	 * This method is used in log and error messages,
-	 * so the returned description should typically include the method signature
-	 * of the underlying handler method for clarity and debugging purposes.
+	 * Returns a concise description of this {@code HandlerMethod}, which is used
+	 * in log and error messages.
+	 * <p>The description should typically include the method signature of the
+	 * underlying handler method for clarity and debugging purposes.
 	 */
 	@Override
 	public String toString() {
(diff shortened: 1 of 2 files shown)


wonyongg:
After revisiting the HandlerMethod implementation, I realized that I misunderstood the original purpose of the `toString()` call.

Initially, I thought the usage of `toString()` was redundant or unclear. However, I now understand that it returns the description string, which actually provides useful context like `ClassName#methodName(args...)`. sorry for the confusion.

That said, I'd like to propose a small improvement.

The method name `toString()` by itself doesn't make the intention very clear to someone unfamiliar with the HandlerMethod class. I believe replacing it with `getShortLogMessage()` could make the code more readable and self-explanatory, especially for contributors or maintainers not deeply familiar with this class.

The output remains similar, but the method name communicates its purpose more directly.

If this suggestion doesn't align with the direction of the project, please feel free to suggest an alternative approach or close this PR if it's unnecessary.

sbrannen:
> The method name `toString()` by itself doesn't make the intention very clear to someone unfamiliar with the HandlerMethod class. I believe replacing it with `getShortLogMessage()` could make the code more readable and self-explanatory, especially for contributors or maintainers not deeply familiar with this class.
> 
> The output remains similar, but the method name communicates its purpose more directly.
> 
> If this suggestion doesn't align with the direction of the project, please feel free to suggest an alternative approach or close this PR if it's unnecessary.

I'm not so sure that it's worth changing things here.

If we were to change something, I suppose we could either make the `HandlerMethod.description` field `protected` or introduce a `getDescription()` method in `HandlerMethod`.

We'll discuss it within the team and get back to you.

wonyongg:
@sbrannen 
Thank you for your feedback! I’ll be waiting to hear back after your team discussion.

sbrannen:
Instead of changing the code or introducing a dedicated method to retrieve the `description`, I think we should just update the Javadoc for `HandlerMethod.toString()` to document that it's used in log/error messages and should typically include the method signature of the handler method.

If you would like to repurpose this PR to do that, feel free to do so.

Otherwise, I'll just add the Javadoc myself.

wonyongg:
Thanks for the feedback! 🙏
I’ve removed the previous changes and updated this PR to only include the Javadoc for HandlerMethod.toString() as suggested.
I’d appreciate it if you could take a quick look and let me know if the updated Javadoc looks good to you.

spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
```diff
@@ -352,6 +352,12 @@ public int hashCode() {
 		return (this.bean.hashCode() * 31 + super.hashCode());
 	}
 
+	/**
+	 * Returns the description of this handler method.
+	 * This method is used in log and error messages,
+	 * so the returned description should typically include the method signature
+	 * of the underlying handler method for clarity and debugging purposes.
+	 */
```
sbrannen:
```suggestion
	/**
	 * Returns a concise description of this {@code HandlerMethod}, which is used
	 * in log and error messages.
	 * <p>The description should typically include the method signature of the
	 * underlying handler method for clarity and debugging purposes.
	 */
```

sbrannen (changes requested):
Thanks for the proposed Javadoc.

I've requested minor changes.

sbrannen (approved):

sbrannen:

Thanks

PR #35061, opened 2025-06-16
Update new-framework.adoc

DhruvTheDev1:
- Corrected xref paths to point to the correct locations.

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webflux/new-framework.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webflux/new-framework.adoc
@@ -82,7 +82,8 @@ Whenever feasible (for example, annotated controllers), WebFlux adapts transpare
 of RxJava or another reactive library. See xref:web-reactive.adoc#webflux-reactive-libraries[Reactive Libraries] for more details.
 
 NOTE: In addition to Reactive APIs, WebFlux can also be used with
-xref:languages/kotlin/coroutines.adoc[Coroutines] APIs in Kotlin which provides a more imperative style of programming.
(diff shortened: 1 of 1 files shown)


DhruvTheDev1:
Updated new-framework.adac to fix broken links to point to correct path. Not all of them have been fixed yet - there are still a few more. When you click on a link it shows 404 - page not work as they reference wrong paths

bclozel:
This is breaking the reference documentation generation. The links you are trying to fix are working fine on the live documentation ([for example, the Coroutines link here](https://docs.spring.io/spring-framework/reference/web/webflux/new-framework.html#webflux-reactive-api)).

Since the locations are already valid, I'm declining this PR.

DhruvTheDev1:
@bclozel Thank you for the clarification! I misunderstood how the xref links are resolved in the live docs. I appreciate the explanation and will be more careful with these in the future.

sbrannen:
@bclozel and I started typing at the same time, and he already covered the important parts.

But... I suppose you are trying the links from the view provided by GitHub:

https://github.com/spring-projects/spring-framework/blob/main/framework-docs/modules/ROOT/pages/web/webflux/new-framework.adoc

The links are **not** supposed to work in that view.

All that matters is that the links within the [generated documentation](https://docs.spring.io/spring-framework/reference/web/webflux/new-framework.html) work.

DhruvTheDev1:
@sbrannen Thank you for the explanation! I was indeed checking the links directly from GitHubs raw file view which I now understand is not the intended way to verify them.

Thank you for your guidance!

PR #35086, opened 2025-06-20
Support `@RequestBody` in interface method when computing "consumes" condition

### LINKED ISSUE #35085: @RequestBody annotation on interface methods ignored for required flag
When `@RequestBody(required = false)` is declared on a controller interface method, Spring currently ignores this annotation during request mapping. As a result, mappings incorrectly require a request body even when required = false is set on the interface.

This leads to incorrect matching behavior, especially when the implementation method does not override the @RequestBody annotation.

Expected behavior:
Spring should detect the @RequestBody annotation on interface methods and respect the required flag during request mapping, allowing requests without bodies when `required = false`.

Actual behavior:
Only the annotations on the implementation method are considered, causing required = false on interfaces to be ignored.

Steps to reproduce:

- Define a controller interface method annotated with @RequestBody(required = false)
- Implement the interface without overriding the annotation
- Register the handler and observe that Spring expects a body even if required = false
- Make a request with no body and no Content-Type and observe that the request fails

Suggested fix:
Use `AnnotatedMethod` to retrieve parameter annotations from both implementation and interface methods to correctly determine the required flag.

Additional context:
I have a patch that fixes this and adds tests to cover both interface and implementation overrides.

renatomameli:
### Problem

Spring currently ignores `@RequestBody` annotations declared on interface methods when computing the `ConsumesCondition`. As a result, the `required = false` flag is not respected, and requests without a body and without a Content-Type fail to match the mapping — even though null should be allowed.

This leads to situations where e.g. `application/json` is unnecessarily enforced in the ConsumesCondition, despite `@RequestBody(required = false)` being declared (e.g., on an interface method).

### Solution

This change switches to using `AnnotatedMethod` for inspecting method parameters, which ensures that annotations from both interface and implementation methods are considered. This allows Spring to correctly determine whether a request body is required and to avoid enforcing a Content-Type match when required = false.

### Tests

Includes tests verifying correct behavior for:

- Interface methods with `@RequestBody(required = false)`
- Implementations that override with `required = true`

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerMapping.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerMapping.java
@@ -19,7 +19,6 @@ package org.springframework.web.servlet.mvc.method.annotation;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
-import java.lang.reflect.Parameter;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -31,7 +30,9 @@ import jakarta.servlet.http.HttpServletRequest;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.context.EmbeddedValueResolverAware;
+import org.springframework.core.MethodParameter;
 import org.springframework.core.annotation.AnnotatedElementUtils;
+import org.springframework.core.annotation.AnnotatedMethod;
 import org.springframework.core.annotation.MergedAnnotation;
 import org.springframework.core.annotation.MergedAnnotationPredicates;
 import org.springframework.core.annotation.MergedAnnotations;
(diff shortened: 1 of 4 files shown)


sickinga:
Thank you @renatomameli. We really rely on this fix and it would be great if we can get this in an upcoming release! @sbrannen

sbrannen:
Please note that your `httpExchangeAnnotation*()` test methods do not involve `@HttpExchange` annotations.

In addition, if we were to support processing of `@RequestBody` annotations on parameters in interface methods, we would need to do so not only for Spring MVC (as in this PR) but also for Spring WebFlux.

In any case, we will discuss within the team whether we want to make this change, and we'll get back to you.

Cheers

renatomameli:
> Please note that your `httpExchangeAnnotation*()` test methods do not involve `@HttpExchange` annotations.
> 
> In addition, if we were to support processing of `@RequestBody` annotations on parameters in interface methods, we would need to do so not only for Spring MVC (as in this PR) but also for Spring WebFlux.
> 
> In any case, we will discuss within the team whether we want to make this change, and we'll get back to you.
> 
> Cheers

Thanks for the feedback!  I’ve renamed them to more accurately reflect their purpose. Looking forward to your team’s decision.

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerMapping.java
```diff
@@ -410,13 +411,17 @@ protected void registerHandlerMethod(Object handler, Method method, RequestMappi
...
+			return;
+		}
+
+		AnnotatedMethod annotatedMethod = new AnnotatedMethod(method);
+
+		for (MethodParameter parameter : annotatedMethod.getMethodParameters()) {
+			RequestBody requestBodyAnnotation = parameter.getParameterAnnotation(RequestBody.class);
+			if (requestBodyAnnotation != null) {
+				condition.setBodyRequired(requestBodyAnnotation.required());
+				break;
```
sbrannen:
```suggestion
			RequestBody requestBody = parameter.getParameterAnnotation(RequestBody.class);
			if (requestBody != null) {
				condition.setBodyRequired(requestBody.required());
				break;
```

sbrannen (changes requested):
Thanks for updating the test method names.

We have decided to support this feature.

Thus, please introduce analogous support in `org.springframework.web.reactive.result.method.annotation.RequestMappingHandlerMapping` as well, including appropriate tests.

renatomameli:
> Thanks for updating the test method names.
> 
> We have decided to support this feature.
> 
> Thus, please introduce analogous support in `org.springframework.web.reactive.result.method.annotation.RequestMappingHandlerMapping` as well, including appropriate tests.

Thanks for the feedback.

I’ve implemented the support in `org.springframework.web.reactive.result.method.annotation.RequestMappingHandlerMapping`, along with corresponding tests. Let me know if there’s anything else you’d like to adjust.

sbrannen:
This has been merged into `main` in b901afffd0cb02a7238232b88da6c9e7cba36e7a and slightly revised in b1222ccab78abc98aec19423202f8d117790892f.

The main additional work involved ensuring that the tests cover more of the actual expectations.

Thanks

Issue #34748, opened 2025-04-13
Add Support to CompletableFuture as return type in Spring interfaces clients with RestClient


ne3ma-ibraheem:
I suggest adding support for `CompletableFuture`  and `CompletionStage` as a valid return type for methods in Spring client interfaces backed by `RestClient` and  `RestClientAdapter`.

It would be great to provide the `ExecutorService` when creating the adapter or maybe leverage `Async` annotation

Example 

```java
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.web.client.RestClient;
import org.springframework.web.service.annotation.GetExchange;
import org.springframework.web.service.annotation.PostExchange;
import org.springframework.web.service.annotation.RequestBody;
import org.springframework.web.service.annotation.PathVariable;
import org.springframework.http.ResponseEntity;
import org.springframework.web.service.invoker.HttpServiceProxyFactory;

public interface MyApiClient {

    @GetExchange("/users/{id}")
    CompletableFuture<User> getUser(@PathVariable int id);

    @PostExchange("/users")
    CompletableFuture<ResponseEntity<User>> createUser(@RequestBody User user);

    @GetExchange("/items")
    CompletableFuture<List<Item>> getAllItems();
}

@Configuration
public class ClientConfig {

    @Bean
    public MyApiClient myApiClient(RestClient.Builder restClientBuilder) {
        RestClient restClient = restClientBuilder.baseUrl("https://api.example.com").build();
        HttpServiceProxyFactory factory = HttpServiceProxyFactory.builder(RestClientAdapter.forClient(restClient)).build();
        return factory.createClient(MyApiClient.class);
    }
}

@Service
public class MyService {

    private final MyApiClient apiClient;

    public MyService(MyApiClient apiClient) {
        this.apiClient = apiClient;
    }

    public void processUser(int userId) {
        apiClient.getUser(userId)
                 .thenAccept(user -> System.out.println("Retrieved user: " + user))
                 .exceptionally(ex -> {
                     System.err.println("Error fetching user: " + ex.getMessage());
                     return null;
                 });
    }
}
```

bclozel:
I am wondering whether we should even consider this. Applications are meant to use several HTTP interface clients and often scatter/gather requests to different services.

Futures and Executor services are quite limited in that space because: they don't propagate cancellation signals, they are harder to instrument for observability, harder to debug, it is harder to synchronize separate calls.

[...]

PR #35090, opened 2025-06-21
Add Support to CompletableFuture as return type in http service method.

remeio:
To fix the issue #34748, I had added the support to `CompletableFuture` as return type in http service method like this:
```java
@GetExchange
CompletableFuture<XXX> asyncRequest();

@GetExchange
CompletableFuture<Optional<XXX>> asyncRequestOptional();
```
It is worth mentioning that I have not yet supported setting `Executor` of `CompletableFuture`. Can anyone give some advice?


rstoyanchev:
Thanks for the suggestion, but this issue is not marked as open for contribution, and not even triaged yet.

PR #35092, opened 2025-06-22
Fix typo in exceptionVariableName in EmptyCatchBlockCheck

izeye:
This PR fixes a typo in `exceptionVariableName` in `EmptyCatchBlockCheck`.

This PR also changes to use "ignored" for exception variable names instead of "ignore".

See gh-35047

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/json/AbstractJsonContentAssert.java
+++ b/spring-test/src/main/java/org/springframework/test/json/AbstractJsonContentAssert.java
@@ -551,7 +551,7 @@ public abstract class AbstractJsonContentAssert<SELF extends AbstractJsonContent
 				read();
 				throw failure(new JsonPathNotExpected(this.json, this.path));
 			}
-			catch (PathNotFoundException ignore) {
+			catch (PathNotFoundException ignored) {
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


bclozel:
Should we apply this?
I configured the check like this because we had already legitimate usage of both "ignore" and "ignored" in the codebase.

I get that "ignore" is not a noun so maybe not very well suited for a variable name. Do we really need to go that far on everything? Should we also refactor the entire codebase to only allow nouns as variable names?

sbrannen:
> Should we apply this? I configured the check like this because we had already legitimate usage of both "ignore" and "ignored" in the codebase.

Indeed, it's a regular expression, not a typo.

It's just not immediately obvious from the Checkstyle configuration.

Perhaps changing the pattern to `expected|ignore(d)?` would make it clearer and more robust.

And perhaps limiting it to only `^expected|ignore(d)?$` would be even better.

> I get that "ignore" is not a noun so maybe not very well suited for a variable name. Do we really need to go that far on everything? Should we also refactor the entire codebase to only allow nouns as variable names?

"expected" is also not a noun, so we certainly don't need to enforce any kind of "nouns only" check.

However, we can still consider tightening the RegEx we use for `exceptionVariableName` to ensure that only "expected", "ignore", or "ignored" is used to permit an empty catch block.

izeye:
@bclozel @sbrannen Thanks for the feedback!

I just thought it would be better to use a consistent name for the same purpose.

bclozel:
Thanks @izeye
The original intent was not clear in the checkstyle configuration.

Issue #35074, opened 2025-06-18
Leniently tolerate null `@Aspect` bean


quaff:
I want to conditionally enable aspect like this:
```java
import org.aspectj.lang.ProceedingJoinPoint;
import org.aspectj.lang.annotation.Around;
import org.aspectj.lang.annotation.Aspect;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.context.ApplicationContext;
import org.springframework.context.annotation.Bean;

@SpringBootApplication
public class DemoApplication {

	public static void main(String[] args) {
		ApplicationContext ctx = SpringApplication.run(DemoApplication.class, args);
		TestService testService = ctx.getBean(TestService.class);
		testService.test();
	}

	@Bean
	TestService testService() {
		return new TestService();
	}

	@Bean
	TestAspect testAspect() {
		boolean enabled = false; // TODO
		return enabled ? new TestAspect() : null;
	}

	@Aspect
	public static class TestAspect {

		@Around("execution(public * test(..))")
		public Object test(ProceedingJoinPoint pjp) throws Throwable {
			return pjp.proceed();
		}
	}

	public static class TestService {

		public void test() {

		}
	}
}

```

It will raise exception when proxied method invoked:
```
Exception in thread "main" org.springframework.aop.AopInvocationException: Mismatch on arguments to advice method [public java.lang.Object com.example.demo.DemoApplication$TestAspect.test(org.aspectj.lang.ProceedingJoinPoint) throws java.lang.Throwable]; pointcut expression [org.aspectj.weaver.internal.tools.PointcutExpressionImpl@51e1e058]
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs(AbstractAspectJAdvice.java:647)
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethod(AbstractAspectJAdvice.java:632)
	at org.springframework.aop.aspectj.AspectJAroundAdvice.invoke(AspectJAroundAdvice.java:71)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
	at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:97)
[...]

PR #35097, opened 2025-06-23
Polish BeanFactoryAspectInstanceFactory

quaff:
See https://github.com/spring-projects/spring-framework/issues/35074#issuecomment-2994607881


jhoeller:
I prefer to handle the `getOrder` problem in #35074 in a different way: calling `getBean` with `Ordered.class` as required type and catching `BeanNotOfRequiredTypeException)`, avoiding internal delegation to `getAspectInstance()` which should be externally called only (through the `AspectInstanceFactory` interface. I'll roll that in as part of #35074.

PR #35109, opened 2025-06-25
Use `ExceptionTypeFilter` to filter includes and excludes for retry policies

remeio:
It's easy to filter `Throwable` with `ExceptionTypeFilter`. Otherwise, we would have to handle inclusions and exclusions everywhere.


sbrannen:
This has been merged into `main` in 489ebd24383040a28b059b5c5407e806696e6eec and revised in 58061ae2959146a1888dce8e3c084983d7c1a35b.

In addition, this PR inspired the following work.

- #35160
- #35161

Cheers!

PR #35114, opened 2025-06-25
Use jspecify Nullable

ngocnhan-tran1996:


bclozel:
Thanks @ngocnhan-tran1996 , I completely missed that bit!
I've added a checkstyle rule that prevents such imports (in this case this was an inline import, so it wouldn't have helped unfortunately).

PR #35135, opened 2025-07-01
Remove duplicate “rss” assertion in test

NeatGuyCoding:
There’s a redundant `.containsEntry("rss", MediaType.APPLICATION_RSS_XML)` in the test – everything else is correct.


sbrannen:
- duplicates #35139

PR #35137, opened 2025-07-01
Fix jitter handling in `AbstractRetryInterceptor`

### LINKED ISSUE #35136: Potentially lose precision and jitter is not well capped
The jitter ratio calculation could throw an `ArithmeticException` if `spec.delay().toMillis()` returns 0. Additionally, converting to milliseconds may lose precision for sub-millisecond delays.
```
Retry retry = Retry.backoff(spec.maxAttempts(), spec.delay())
	.jitter((double) spec.jitter().toMillis() / spec.delay().toMillis())
	.multiplier(spec.multiplier())
	.maxBackoff(spec.maxDelay())
```

NeatGuyCoding:
Fix #35136


sbrannen:
- duplicates #35140

Issue #35133, opened 2025-06-30
Introduce `@ConcurrencyLimit` annotation based on `ConcurrentThrottleInterceptor`


jhoeller:
Following up on `@Retryable` in #34529, we are considering a related feature for throttling method invocations: namely, a `@ConcurrencyLimit` annotation based on our good old `ConcurrencyThrottleInterceptor` arrangement (dating back to 1.0) which has recently been revised for use with `SimpleAsyncTaskExecutor` in Virtual Threads mode (#34727).

Since both `@Retryable` and `@ConcurrencyLimit` are effectively resilience features for method invocations, it is attractive to combine them into an `org.springframework.resilience` package in the `spring-context` module (next to `org.springframework.scheduling` which features `@Scheduling` and `@Async`) rather than keeping the retry interceptor in `org.springframework.aop.retry`. The goal is to unify these efforts for 7.0 M7.

PR #35140, opened 2025-07-01
Fix jitter handling in `AbstractRetryInterceptor`

NeatGuyCoding:
## Overview

Fix potential loss of precision, and test jitter handling with unit tests.

## Related Issues

- #35136


jhoeller:
Good catch! I'll merge this right away in preparation of #35133 where the `aop.retry` package is about to be moved to `org.springframework.resilience.retry` in the `spring-context` module.

sbrannen:
Changed to a `task` since this feature has not yet been released.

PR #35150, opened 2025-07-03
Allow any @Transactional propagation for @TransactionalEventListener with BEFORE_COMMIT phase

carsago:
Following discussions in #30679 and #31414, validation was added to restrict `@TransactionalEventListener` to only use `@Transactional(propagation = Propagation.REQUIRES_NEW)` or `@Transactional(propagation = Propagation.NOT_SUPPORTED)`.

However, this validation is unnecessary when the `@TransactionalEventListener` phase is `BEFORE_COMMIT` (excluding propagation `NEVER`). While removing the `@Transactional` would solve the issue, this creates unnecessary friction when upgrading Spring versions.


jhoeller:
Good point! I've implemented this a bit differently, instantiating the listener first and then taking the phase from it for validation purposes. Thanks for the PR, in any case!

PR #35186, opened 2025-07-11
Add explanation for @Valid container parameters in ann-validation.adoc

HeeChanN:
Added a section explaining that container parameters annotated with `@Valid`

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webmvc/mvc-controller/ann-validation.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webmvc/mvc-controller/ann-validation.adoc
@@ -20,6 +20,13 @@ directly on method parameters, or on the method (for the return value), then met
 validation must be applied, and that supersedes validation at the method argument level
 because method validation covers both method parameter constraints and nested constraints
 via `@Valid`. The exception raised in this case is `HandlerMethodValidationException`.
++
+Container parameters annotated with `@Valid`—for example,
+`@RequestBody List<Foo> items` or `@RequestBody @Valid List<@Valid Foo> items`—are
+likewise subject to method validation. They are detected by
+`HandlerMethod.MethodValidationInitializer.checkArguments(…)`, and any
+resulting constraint violations cause a `HandlerMethodValidationException`
+to be raised.
 
 Applications must handle both `MethodArgumentNotValidException` and
 `HandlerMethodValidationException` as either may be raised depending on the controller


framework-docs/modules/ROOT/pages/web/webmvc/mvc-controller/ann-validation.adoc
```diff
@@ -20,6 +20,13 @@ directly on method parameters, or on the method (for the return value), then met
 validation must be applied, and that supersedes validation at the method argument level
 because method validation covers both method parameter constraints and nested constraints
 via `@Valid`. The exception raised in this case is `HandlerMethodValidationException`.
++
+Container parameters annotated with `@Valid`—for example,
+`@RequestBody List<Foo> items` or `@RequestBody @Valid List<@Valid Foo> items`—are
+likewise subject to method validation. They are detected by
+`HandlerMethod.MethodValidationInitializer.checkArguments(…)`, and any
+resulting constraint violations cause a `HandlerMethodValidationException`
+to be raised.
```
rstoyanchev, line 29: Is `@RequestBody List<Foo> items` really validated? I would expect it needs `@Valid`.

What is the purpose of having `@Valid` in both places with `@RequestBody @Valid List<@Valid Foo> items`? Isn't `@RequestBody @Valid List<Foo> items` sufficient, or likewise I would expect `@RequestBody List<@Valid Foo> items` to work. It's unclear why both?

HeeChanN: Thanks for catching that! I actually made a typo in my description — I meant to write either
`@RequestBody @Valid List<Foo> items` or` @RequestBody List<@Valid Foo>` items.

It was my mistake to put both annotations together in the example. Thanks for pointing it out!

For context, the reason I opened this PR is that when using `@Valid` together with a container type,
I noticed that `@RequestBody @Valid List<Foo>` triggers a `HandlerMethodValidationException`,
whereas the official documentation led me to expect a `MethodArgumentNotValidException`.
I thought it would be helpful to clarify this behavior in the reference documentation so that others won’t run into the same confusion.

rstoyanchev: Thanks for clarifying. Now that the intent is more clear, I am going to address this differently, and that section generally can use a little improvement.

rstoyanchev:
Superseded by #35759.

PR #35192, opened 2025-07-12
Allow explicitly declared null in providedArgs when no argument resolver matches

### LINKED ISSUE #35067: HandlerMethod cannot be resolved in @ExceptionHandler
**Summary**
I have a Spring Boot service where custom JSON error responses are returned when controller methods are annotated with a custom annotation.

When accessing an endpoint that does not exist, the default Spring 404 JSON body is returned, but the following unexpected error detail appears in the response:

```
java.lang.IllegalStateException: Could not resolve parameter [1] in handleException(java.lang.Exception,org.springframework.web.method.HandlerMethod) throws java.lang.Exception: No suitable resolver
	at org.springframework.web.method.support.InvocableHandlerMethod.getMethodArgumentValues(InvocableHandlerMethod.java:224)
```

**How to Reproduce**:
You can reproduce this by running the test here:
https://github.com/cbsingh1/exception-handler-method/blob/main/src/test/java/com/example/exception_handler/ExceptionHandlerApplicationTests.java

**Expected Behavior**
If the HandlerMethod cannot be resolved (e.g., because no handler exists for the request), I would expect:

- handlerMethod to be null, allowing me to handle it explicitly in my exception handler.
- or cleaner logging/formatting of the error

```
@ExceptionHandler(Exception.class)
ResponseEntity<String> handleAllExceptions(Exception exception, HandlerMethod handlerMethod) throws Exception {
    if (handlerMethod == null) {
        // special handling
    }
    if (handlerMethod.hasMethodAnnotation(MyCustomAnnotation.class)) {
        return ResponseEntity.internalServerError().body("custom failure message");
    }
    throw exception;
}
```

**Suggestion**
It would be great if Spring could:
 
- Provide a null (or safe default) for HandlerMethod when no handler is resolved.
- Improve error formatting

yongkaikai:
### Overview

This PR addresses an issue where `InvocableHandlerMethod#getMethodArgumentValues` throws an `IllegalStateException`
when a method parameter is not supported by any `HandlerMethodArgumentResolver` and `null` is explicitly provided in `providedArgs`.

### What this PR does

- Adds a fallback check via `isParameterDeclaredButNull(...)` to tolerate explicit `null` values
- Ensures that null in `providedArgs` is interpreted as a deliberate argument, avoiding spurious resolution failures

### Why this is useful

In practice, this improves error handler behavior, especially for `@ExceptionHandler` where the HandlerMethod can be `null`.

### Related issue

Fixes [#35067](https://github.com/spring-projects/spring-framework/issues/35067)

### DIFF
--- a/spring-test/src/test/java/org/springframework/test/web/servlet/samples/standalone/ExceptionHandlerTests.java
+++ b/spring-test/src/test/java/org/springframework/test/web/servlet/samples/standalone/ExceptionHandlerTests.java
@@ -29,6 +29,7 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
+import org.springframework.web.method.HandlerMethod;
 
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.forwardedUrl;
@@ -62,6 +63,14 @@ public class ExceptionHandlerTests {
 				.andExpect(status().isOk())
 				.andExpect(forwardedUrl("globalErrorView"));
 		}
+
+		@Test
+		void nullHandlerMethod() throws Exception {
+			standaloneSetup(new PersonController()).setControllerAdvice(new GlobalExceptionHandler()).build()
+				.perform(get("/invalid"))
+				.andExpect(status().isOk())
+				.andExpect(forwardedUrl("null handlerMethod exception"));
+		}
 	}
 
 
@@ -91,6 +100,14 @@ public class ExceptionHandlerTests {
 		@ExceptionHandler
 		String handleException(IllegalStateException exception) {
 			return "globalErrorView";
+		}
+
+		@ExceptionHandler(Exception.class)
+		String handleAllExceptions(IllegalStateException exception, HandlerMethod handlerMethod) {
+			if (handlerMethod == null) {
+				return "null handlerMethod exception";
+			}
+			return "simulated exception";
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java
```diff
@@ -218,6 +218,10 @@ public void setMethodValidator(@Nullable MethodValidator methodValidator) {
 				continue;
 			}
 			if (!this.resolvers.supportsParameter(parameter)) {
+				if (isParameterDeclaredButNull(parameter, providedArgs)) {
+					args[i] = null;
```
mengnankkkk, line 222: The new isParameterDeclaredButNull method checks if a parameter is explicitly set to null in providedArgs and assigns null directly, bypassing the exception. While this improves compatibility in some cases, it may hide the real cause of argument resolution failures, reducing code robustness and maintainability.

spring-test/src/test/java/org/springframework/test/web/servlet/samples/standalone/ExceptionHandlerTests.java
```diff
@@ -62,6 +63,14 @@ void globalExceptionHandlerMethod() throws Exception {
 				.andExpect(status().isOk())
 				.andExpect(forwardedUrl("globalErrorView"));
 		}
+
+		@Test
+		void nullHandlerMethod() throws Exception {
+			standaloneSetup(new PersonController()).setControllerAdvice(new GlobalExceptionHandler()).build()
+				.perform(get("/invalid"))
```
mengnankkkk, line 70: The test only covers the scenario where handlerMethod is null, but does not cover multiple parameters, complex types, or nested parameters. Additional tests are suggested.

rstoyanchev (commented):
Thanks for giving this a try. However, the proposed solution targets much more than just the original scenario with a `HandlerMethod` parameter, and would pass `null` for any parameter type instead of flagging it as unresolved.

PR #35196, opened 2025-07-13
Update docs by adding use of static @Bean methods for BPP and BFPP

### LINKED ISSUE #34964: Document registration recommendations for `BeanPostProcessor` and `BeanFactoryPostProcessor`
I believe that it's generally recommended for a `BeanPostProcessor` that's defined as a `@Bean` to use a `static` method and for the method to, ideally, have no dependencies. The goal being to minimise the chances of the BPP triggering eager initialization that makes other beans ineligible for full post-processing.

I thought this was mentioned in the reference documentation but I have been unable to find it. https://docs.spring.io/spring-framework/reference/core/beans/factory-extension.html#beans-factory-extension-bpp talks a little about it, but only in the context of AOP auto-proxying as far as I can see. I think the advice should apply more broadly. There's also no Java example of registering a BPP, instead XML is used which missed the opportunity to recommend the use of `static` when a `@Bean` method is involved.

DeepDhamala:
After reviewing https://github.com/spring-projects/spring-boot/issues/45622 and gh-34964,
I updated the documentation to recommend declaring BeanPostProcessor and BeanFactoryPostProcessor
as static @Bean methods. This practice prevents early bean initialization and ensures proper post-processing.

Closes gh-34964

### DIFF
--- a/framework-docs/modules/ROOT/pages/core/beans/factory-extension.adoc
+++ b/framework-docs/modules/ROOT/pages/core/beans/factory-extension.adoc
@@ -197,6 +197,54 @@ other bean. (The preceding configuration also defines a bean that is backed by a
 script. The Spring dynamic language support is detailed in the chapter entitled
 xref:languages/dynamic.adoc[Dynamic Language Support].)
 
+The following example shows recommended to register a class-based configuration using @Configuration:
+
+[tabs]
+======
+Java::
++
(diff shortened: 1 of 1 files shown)


sbrannen:
Why did you submit a PR for an issue that is already assigned to someone?

DeepDhamala:
Closing this PR as it is already assigned.

PR #35212, opened 2025-07-17
Correct the default value of nestedTransactionAllowed in JpaTransactionManager javadoc

Dockerel:
**Correct the default value of nestedTransactionAllowed in Javadoc**

* Changed the default value in the Javadoc from `false` to `true`
* For smoother wording, replaced **"though, since" to "but"** and **“use” with “disallow”**

The [Javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/orm/jpa/JpaTransactionManager.html) for JpaTransactionManager currently states that the nestedTransactionAllowed flag defaults to false. However, the default constructor has been setting this flag to true.

```java
/**
* Create a new JpaTransactionManager instance.
* <p>An EntityManagerFactory has to be set to be able to use it.
* @see #setEntityManagerFactory
*/
public JpaTransactionManager() {
    setNestedTransactionAllowed(true);
}
```

This PR updates the documentation to align with the actual behavior, ensuring consistency between the code and its Javadoc.

Let me know if you have any concerns or suggestions for further improvement.
Thank you for your time and consideration!

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/JpaTransactionManager.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/JpaTransactionManager.java
@@ -98,9 +98,7 @@ import org.springframework.util.CollectionUtils;
  * context. You can manually set the flag to {@code false} if you want to disallow nested
  * transactions for JDBC access code which participates in JPA transactions (provided
  * that your JDBC driver supports Savepoints). <i>Note that JPA itself does not support
- * nested transactions! Furthermore, as most JPA providers (e.g., Hibernate)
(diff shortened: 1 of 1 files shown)


Dockerel:
Is there anything that i should do for this PR to be merged? @jhoeller

jhoeller:
@Dockerel this has been postponed to reconsider the intended default for `nestedTransactionAllowed`. Arguably, it should be as the javadoc days for JPA which we consider fixing in 7.0. In that case, we should add a note in 6.2.11 that the default is wrong.

Dockerel:
I see. Thank you for clarification!

PR #35221, opened 2025-07-19
Fix Checkstyle violations in SpringVersion.java

Naveenkumarsuk:
This pull request fixes two Checkstyle violations in the SpringVersion.java file:

Corrects the header to match the expected license format
Removes an unnecessary blank line before the Javadoc @see tag

The build now passes cleanly with no Checkstyle errors.


bclozel:
Thanks for the proposal but the build passes just fine right now and the PR description does not fit the actual change, which removes a single blank line.

PR #35225, opened 2025-07-20
Add compression support in JdkClientHttpRequestFactory

spicydev:
### Title
Implement custom BodyHandler for GZIP decompression in JdkClientHttpRequest

### Description

**Problem:** JDK HTTP Client doesn't automatically decompress responses.

**Solution:** creating a custom BodyHandler that checks Content-Encoding and applies GZIPInputStream when necessary.
**Key Changes:** 
- Added new boolean field `compressionEnabled` to `JdkClientHttpRequestFactory` and `JdkClientHttpRequest`.
- Modified `buildRequest()` method to add `Accept-Encoding` header with `gzip` value when compression is enabled.
- Added a custom `DecompressingBodyHandler` that implements `BodyHandler<InputStream>`. The apply() method in the custom handler checks the `Content-Encoding` header.
- If the encoding is `gzip`, it uses BodySubscribers.mapping to wrap the InputStream with GZIPInputStream.
  Otherwise, it uses the default BodySubscribers.ofInputStream().

**Testing Steps:**
- Make an HTTP request to an endpoint that returns a gzip-encoded response (e.g., https://vpic.nhtsa.dot.gov/api/vehicles/getallmanufacturers?format=json).
- Verify that the response body is correctly decompressed and the data is accessible.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
@@ -25,6 +25,10 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.net.http.HttpTimeoutException;
+import java.net.http.HttpResponse.BodyHandler;
+import java.net.http.HttpResponse.BodySubscriber;
+import java.net.http.HttpResponse.BodySubscribers;
+import java.net.http.HttpResponse.ResponseInfo;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.Collections;
@@ -37,6 +41,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Flow;
 import java.util.concurrent.TimeUnit;
+import java.util.zip.GZIPInputStream;
 
 import org.jspecify.annotations.Nullable;
 
@@ -70,15 +75,18 @@ class JdkClientHttpRequest extends AbstractStreamingClientHttpRequest {
 
 	private final @Nullable Duration timeout;
 
+	private final boolean compressionEnabled;
+
 
 	public JdkClientHttpRequest(HttpClient httpClient, URI uri, HttpMethod method, Executor executor,
-			@Nullable Duration readTimeout) {
+			@Nullable Duration readTimeout, boolean compressionEnabled) {
 
 		this.httpClient = httpClient;
 		this.uri = uri;
 		this.method = method;
 		this.executor = executor;
 		this.timeout = readTimeout;
+		this.compressionEnabled = compressionEnabled;
 	}
 
 
@@ -98,7 +106,11 @@ class JdkClientHttpRequest extends AbstractStreamingClientHttpRequest {
 		CompletableFuture<HttpResponse<InputStream>> responseFuture = null;
 		try {
 			HttpRequest request = buildRequest(headers, body);
-			responseFuture = this.httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofInputStream());
+			if (compressionEnabled) {
+				responseFuture = this.httpClient.sendAsync(request, new DecompressingBodyHandler());
+			} else {
+				responseFuture = this.httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofInputStream());
+			}
 
 			if (this.timeout != null) {
 				TimeoutHandler timeoutHandler = new TimeoutHandler(responseFuture, this.timeout);
@@ -140,6 +152,10 @@ class JdkClientHttpRequest extends AbstractStreamingClientHttpRequest {
 
 	private HttpRequest buildRequest(HttpHeaders headers, @Nullable Body body) {
 		HttpRequest.Builder builder = HttpRequest.newBuilder().uri(this.uri);
+		
+		if (compressionEnabled) {
+			headers.add(HttpHeaders.ACCEPT_ENCODING, "gzip");
+		}
 
 		headers.forEach((headerName, headerValues) -> {
 			if (!DISALLOWED_HEADERS.contains(headerName.toLowerCase(Locale.ROOT))) {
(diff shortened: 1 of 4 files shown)


bclozel:
Should we also support the "deflate" compression algorithm?

Also, I am wondering if we really need an option that enables compression. Should we enable this no matter what?

What do you think?

spicydev:
Hi @bclozel, First off let me start off by apologizing for raising the PR in such bad state. I’ll work on unit tests and follow style guide once we can finalize on the requirements.

coming to your questions, Yes we can add support for multiple encoding types but following other clients like apache httpclient & netty, they only support gzip by default out of the box. I think at-least for now we can add support for gzip to JDK http client.

I think it’s better to have the control to the users but by default we can enable compression as it causes no harm but improves performance for all users.

Let me know your thoughts, based on that i can go ahead with changes.

bclozel:
@spicydev no worries, we are here to help.

I think Apache HttpClient [supports both](https://github.com/apache/httpcomponents-client/tree/9d38e5fb4438e8fb6442ac8e5ff304142938cf83/httpclient5/src/main/java/org/apache/hc/client5/http/entity). And [so does Netty](https://github.com/netty/netty/blob/4.2/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecompressor.java).

spicydev:
Hi @bclozel , Need your help here. When testing the Inflater with JDK http client I'm seeing error like below
`java.util.zip.ZipException: incorrect header check`. Can you check the unit test once

driverpt:
Good stuff!

Suggestion for the future for Spring Maintainers: Make Default Body Handler overridable from the outside, so we can implement workarounds

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
```diff
@@ -98,7 +110,7 @@ protected ClientHttpResponse executeInternal(HttpHeaders headers, @Nullable Body
 		CompletableFuture<HttpResponse<InputStream>> responseFuture = null;
 		try {
 			HttpRequest request = buildRequest(headers, body);
-			responseFuture = this.httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofInputStream());
+			responseFuture = this.httpClient.sendAsync(request, new DecompressingBodyHandler());
```
driverpt, line 113: Can we make this Handler a parameter ? This will make Workarounds easier

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequestFactory.java
```diff
@@ -43,6 +43,8 @@ public class JdkClientHttpRequestFactory implements ClientHttpRequestFactory {
 
 	private @Nullable Duration readTimeout;
 
+	private boolean compressionEnabled;
+
```
driverpt, line 47: Add Optional `BodyRequestHandler` parameter, defaulting to the new one you just created

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequestFactory.java
```diff
@@ -43,6 +43,8 @@ public class JdkClientHttpRequestFactory implements ClientHttpRequestFactory {
 
 	private @Nullable Duration readTimeout;
```
driverpt, line 44: Suggestion: Create `JdkClientHttpRequestFactoryConfig` with all parameters

spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequestFactory.java
```diff
@@ -96,10 +98,18 @@ public void setReadTimeout(Duration readTimeout) {
...
+	 */
+	public void setCompressionEnabled(boolean compressionEnabled) {
+		this.compressionEnabled = compressionEnabled;
+	}
+
 
 	@Override
 	public ClientHttpRequest createRequest(URI uri, HttpMethod httpMethod) throws IOException {
-		return new JdkClientHttpRequest(this.httpClient, uri, httpMethod, this.executor, this.readTimeout);
+		return new JdkClientHttpRequest(this.httpClient, uri, httpMethod, this.executor, this.readTimeout, this.compressionEnabled);
```
driverpt, line 112: Suggestion: With above suggestion, inject `JdkClientHttpRequestFactoryConfig`

spring-web/src/test/java/org/springframework/http/client/AbstractMockWebServerTests.java
```diff
@@ -106,6 +112,26 @@ else if(request.getTarget().startsWith("/header/")) {
 					String headerName = request.getTarget().replace("/header/","");
 					return new MockResponse.Builder().body(headerName + ":" + request.getHeaders().get(headerName)).code(200).build();
 				}
+				else if(request.getTarget().startsWith("/compress/")) {
+					String encoding = request.getTarget().replace("/compress/","");
+					ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+					if (encoding.equals("gzip")) {
```
driverpt, line 118: Use improved switch?

bclozel:
Thanks a lot @spicydev for your first contribution! Tests were initially failing because the test server was writing non-printable characters as a String instead of bytes directly.

@driverpt I don't think we should open the API more here. A `BodyHandler` is very much a JDK client concern. If a custom body handler is needed, developers should use the `HttpClient` directly.

driverpt:
Will this be backported into current Spring Boot 3.5.x?

> @driverpt I don't think we should open the API more here. A `BodyHandler` is very much a JDK client concern. If a custom body handler is needed, developers should use the `HttpClient` directly.

@bclozel, Actually IMHO you do, given that `JdkClientHttpRequest` Class is internal to Spring and there's no way of customizing it.

If you check the call

```java
	@Override
	@SuppressWarnings("NullAway")
	protected ClientHttpResponse executeInternal(HttpHeaders headers, @Nullable Body body) throws IOException {
                // ....
			HttpRequest request = buildRequest(headers, body);
			responseFuture = this.httpClient.sendAsync(request, 
                                          new DecompressingBodyHandler() // <--- This part is not customizable
                         ); 
                // ....
        }
```

bclozel:
Yes, that's exactly the point 👍 
We do not expose all Spring Framework classes as public on purpose. If your application needs to further customize `BodyHandler`, you should use `HttpClient` directly as our goal here is not to replace it, but rather provide a convenient common API for many HTTP clients.

driverpt:
Agree, but IMHO you should make if configurable, defaulting to `new DecompressingBodyHandler()`. If we just want to add some additional flavor to the `BodyHandler` or apply a workaround for an issue, We would need to implement a whole `JdkClientHttpFactory` of our own. Which kind of goes against the purpose of re-usable Spring Components, this way you are making an opinionated choice, forcing everyone to use Spring's own implementation of a `BodyHandler`

bclozel:
Reopening to disable decompression completely if the compression option is not enabled.

spicydev:
Hi @bclozel and @driverpt , Thanks for your valuable insights. What can we do here to add support going forward but not making this feature more complex. I can raise fresh PR with changes if we can decide on the feature set and depth needed thats sensible.

bclozel:
We are all good now I think @spicydev

PR #35234, opened 2025-07-23
Add provider variant of GroupRegistry.forGroup

philwebb:
Update `GroupRegistry` with an alternative `forGroup` method that uses provider functions for the group name and client type.

This update allows more advanced group registration, for example where the group name is provided is obtained from an annotation of the HTTP Service type. It also allows registration to be skipped by returning `null` as the group name.


rstoyanchev:
Superseded by #35244.

PR #35243, opened 2025-07-25
Use Collections#isEmpty or String#isEmpty

iddeepak:
**Summary**  
This PR improves readability and consistency across the codebase by replacing manual size/length checks with more idiomatic methods:

Replaces collection.size() == 0 or collection.size() != 0 with collection.isEmpty() / !collection.isEmpty()

Replaces string.length() == 0 with string.isEmpty()

**Motivation**
These changes align with modern Java best practices:

isEmpty() is more expressive and self-documenting than comparing .size() or .length() to 0.

This is similar to the [MNG-8060](https://github.com/apache/maven/pull/1420) change made in Apache Maven and [#10945](https://github.com/apache/maven/pull/10945), which improved code clarity.

Please check : 
[[Experimental] Add rewrite support for errorprone.refasterrules.StringRulesRecipes](https://github.com/diffplug/spotless/pull/2576)
[Convert value.trim().length() < 1 to value.isBlank() < 1](https://github.com/openrewrite/rewrite-apache/issues/90)
https://error-prone.picnic.tech/refasterrules/StringRules

These patterns are common and can be enforced or auto-corrected via CI using static analysis and refactoring tools.

### DIFF
--- a/integration-tests/src/test/java/org/springframework/beans/factory/xml/ComponentFactoryBean.java
+++ b/integration-tests/src/test/java/org/springframework/beans/factory/xml/ComponentFactoryBean.java
@@ -35,7 +35,7 @@ public class ComponentFactoryBean implements FactoryBean<Component> {
 
 	@Override
 	public Component getObject() {
-		if (this.children != null && this.children.size() > 0) {
+		if (this.children != null && !this.children.isEmpty()) {
 			for (Component child : children) {
 				this.parent.addComponent(child);
 			}
--- a/spring-context/src/main/java/org/springframework/context/support/ReloadableResourceBundleMessageSource.java
+++ b/spring-context/src/main/java/org/springframework/context/support/ReloadableResourceBundleMessageSource.java
@@ -379,18 +379,18 @@ public class ReloadableResourceBundleMessageSource extends AbstractResourceBased
 		StringBuilder temp = new StringBuilder(basename);
 
 		temp.append('_');
-		if (language.length() > 0) {
+		if (!language.isEmpty()) {
 			temp.append(language);
 			result.add(0, temp.toString());
 		}
 
 		temp.append('_');
-		if (country.length() > 0) {
+		if (!country.isEmpty()) {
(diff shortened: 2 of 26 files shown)


Pankraz76:
well done, like always!

If not done by some plugin this is kind of waste as it will occure any time again.

If spring really want to solve this issue, by having solution like in check:

- https://github.com/diffplug/spotless/pull/2576


Kindly let me know @bclozel

iddeepak:
This is best if rewrite not accepted can we still add a regex in checkstyle to fail the build. Your thoughts @Pankraz76

Pankraz76:
nice idea, yes give it a try in dedicated pr.

mdeinum:
IIRC the checks `collection.size() == 0` and `collection.size() > 0` are deliberate in there. Although it is tempting to use `collection.isEmpty()` and `!collection.isEmpty()` they are actually hard to distinguish from one another. The `!` is quite easily missed or added by accident, hence usage of `.size()` instead for readability and to reduce mistakes. 

If you want to replace them you could introduce an `isNotEmpty` method  (next to the `isEmpty`) on the `CollectionUtils` class and delegate to that. The same could apply for the `String` checks.

bclozel:
All classes under the `asm.*` and `cglib.*` packages are repackaged classes from the original projects. We track the original projects, update and maintain local patches. Changing the code in this namespace for stylistic reasons is just likely to make our maintenance process harder and error-prone. Readability is not really an issue here, because we only maintain the patches, not the rest of the code. In my opinion those are the only good instances of changes where we have `.size() == 0` and `.length() == 0`. I would suggest sending PRs to the original projects here if you believe this improves readability and maintenance.

As for all the other instances, we're really turning `something.size() > 0` into `!something.isEmpty()`, which I don't think improves readability.

As a result, I'm declining this PR because I don't think there are other cases than the two types listed above.
Thanks!

iddeepak:
Thank you for detailed feedback, @bclozel

PR #34519, opened 2025-02-28
Add CollectionUtils.isNotEmpty() method

darckyn:
The isNotEmpty method was created to making the code more readable and standardized.

Thank you.

PR #35247, opened 2025-07-28
Add isNotEmpty Utility Methods to Improve Readability and Safety

iddeepak:
Summary
This PR introduces new utility methods to improve readability and reduce the risk of subtle bugs related to negated emptiness checks:

Adds CollectionUtils.isNotEmpty(Collection<?>)
Adds CollectionUtils.isNotEmpty(Map<?, ?>)

Motivation
As noted by @mdeinum in #35243 , negated expressions like !collection.isEmpty() can be error-prone:

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/CollectionUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/CollectionUtils.java
@@ -81,6 +81,27 @@ public abstract class CollectionUtils {
 	}
 
 	/**
+	 * Return {@code true} if the supplied Collection is not {@code null} and not empty.
+	 * Otherwise, return {@code false}.
+	 * @param collection the Collection to check
+	 * @return whether the given Collection is not empty
+	 */
+	public static boolean isNotEmpty(@Nullable Collection<? extends @Nullable Object> collection) {
+		return (collection != null && !collection.isEmpty());
+	}
+
+	/**
+	 * Return {@code true} if the supplied Map is not {@code null} and not empty.
+	 * Otherwise, return {@code false}.
+	 * @param map the Map to check
+	 * @return whether the given Map is not empty
+	 */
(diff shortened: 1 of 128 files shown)


spring-core/src/main/java/org/springframework/util/CollectionUtils.java
```diff
@@ -80,6 +80,27 @@ public static boolean isEmpty(@Nullable Map<?, ? extends @Nullable Object> map)
...
+	/**
+	 * Return {@code true} if the supplied Map is not {@code null} and not empty.
+	 * Otherwise, return {@code false}.
+	 * @param map the Map to check
+	 * @return whether the given Map is not empty
+	 */
+	public static boolean isNotEmpty(@Nullable Map<?, ? extends @Nullable Object> map) {
+		return (map != null && !map.isEmpty());
+	}
+
```
iddeepak, line 102: @mdeinum 
Added isNotEmpty methods to CollectionUtils as suggested to improve readability and avoid !isEmpty() usage

spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
```diff
@@ -244,7 +244,7 @@ else if (ch == '"') {
 				nextIndex++;
 			}
 			String parameter = mimeType.substring(index + 1, nextIndex).trim();
-			if (parameter.length() > 0) {
+			if (StringUtils.hasLength(parameter)) {
```
iddeepak, line 247: use StringUtils already has a way to check 
@mdeinum

iddeepak:
I see StringUtils#isEmpty is replaced with !StringUtils.hasLength everywhere, Can't we just use if string.isEmpty() ?
@mdeinum

bclozel:
Thanks for the proposal, but this has been declined multiple times, see #34519 and linked issues.

Issue #35248, opened 2025-07-29
`ScriptUtils.executeSqlScript()` does not support multiple results per statement


mixaaaa:
I actually faced an issue for a Sybase ASE database using `DatabasePopulator.execute()` and `ScriptUtils.executeSqlScript()`.

I run a small script which looks similar to this example:

```sql
declare @id int

select @id = id from test where value = '123'

delete from other_one where test_id = @id
delete from other_two where test_id = @id
go
```

The reading of the script and splitting it into statements work absolutely fine when setting "go\n" as separator.

But when it comes to execution, the script will just be executed every third time which is very weird.

My expectation would be that the the select and delete statements are executed every time.

So when debugging and adjusting the `executeSqlScript()` method I found that there are multiple result sets and `updateCounts` which are delivered by the statement and Sybase needs explicitly all to be called here.

This problem seems already to be known for SQL server as well, as I found on [Stackoverflow](https://stackoverflow.com/a/14829425/5387984).

So my adjustment would be to add explicit handling for more results at this spot here:

https://github.com/spring-projects/spring-framework/blob/321a804449e78451f8ed22703ae93accb3a5139f/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/init/ScriptUtils.java#L261-L262

This could look like this for example:

```java
try {
	boolean results = stmt.execute(statement);
	int rowsAffected = stmt.getUpdateCount();
	while (results || rowsAffected != -1) {
		rowsAffected = stmt.getUpdateCount();
		// debug logging
 		results = stmt.getMoreResults(Statement.CLOSE_CURRENT_RESULT);
	}
}
```

sbrannen:
Thanks for reporting the issue, @mixaaaa. 👍 

This has been fixed in 37b076be5121edbe0412f6b8ef190d595692b0e0 for inclusion in `6.2.10` and 7.0.

We would be grateful if you could try out an upcoming 6.2.10 [snapshot](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Artifacts#snapshot-dependency-configuration) and let us know before the release (this coming Thursday) whether the change works for you with Sybase.

Cheers,

Sam

mixaaaa:
Hey @sbrannen,

Thanks for fixing the issue and I can completely confirm that the fix is working.
Tested it multiple times now without any problems.

Thanks,

Uwe

sbrannen:
Hi @mixaaaa,

[...]

PR #35249, opened 2025-07-29
Fix sybase script execution to handle multiple results

DhruvTheDev1:
Addresses issue #35248 
Updates `executeSqlScript()` to handle multiple result set returned by the statement execution. It uses a loop to process all the result and updates the count.


sbrannen:
- Duplicates #35250

PR #35250, opened 2025-07-29
Fix sybase script execution handling

DhruvTheDev1:
Addresses issue https://github.com/spring-projects/spring-framework/issues/35248
Updates executeSqlScript() to handle multiple result set returned by the statement execution. It uses a loop to process all the result and updates the count.

Reopened PR with sign off - had some issues with git bash so just made a new branch


sbrannen:
Hi @DhruvTheDev1,

I decided to address this differently in 37b076be5121edbe0412f6b8ef190d595692b0e0.

Thanks for the PR anyway.

Issue #35256, opened 2025-07-30
ApiVersionConfigurer misconfiguration is hard to detect when no version resolvers are added


philwebb:
Currently [this code](https://github.com/spring-projects/spring-framework/blob/4ae5d0d1fe54cea5372dca4ad1e3f4ba5a09391b/spring-webflux/src/main/java/org/springframework/web/reactive/config/ApiVersionConfigurer.java#L191-L193) in `ApiVersionConfigurer` means that if you call `setVersionRequired(true)` but forget to call a `use...` method API versioning is disabled and the required version is not enforced.

DhruvTheDev1:
Hi - I will open a PR for this issue
- I am thinking of logging a warning such as API version is required but no version is configured and adding a default resolver `.useRequestHeader("X-API-Version");`

philwebb:
@DhruvTheDev1 I would wait for someone from the team to provide a direction. I suspect we'll want an exception instead of a warning.

DhruvTheDev1:
@philwebb Is throwing an exception better in this case? If versionRequired = true but no resolvers are added then it will immediately fail - hence it's impossible to overlook and ensures the developer fixes the problem.
However for logging and falling to a default resolver is more of a safety net.

philwebb:
I don't know, that's why I'd suggest waiting before submitting a PR.

PR #35258, opened 2025-07-30
Add exception for required versioning without resolvers

DhruvTheDev1:
Addresses Issue #35256
- Throws an exception if API versioning is required but no resolvers are configured
- Alternative approach would be to log a warning and add a default resolver `.useRequestHeader("X-API-Version");`


DhruvTheDev1:
- Build failed as trailing whitespace hence removed it and pushed

PR #35266, opened 2025-07-31
Remove wrong Javadoc since tags in MockMvcWebTestClientSpecs

izeye:
This PR removes wrong Javadoc `@since` tags in the `MockMvcWebTestClientSpecs` as they seem to have been introduced in 336a5d0ac848853220a14c39101c4e4ef44e67b8 accidentally.


rstoyanchev (commented):
They are not wrong actually. They are correct original dates. All that is done is to wrap them in a new container class for 7.0.

Issue #35272, opened 2025-08-02
Support HTTP method-restricted interceptors in InterceptorRegistration


SRIRAM9487:
## Overview

Spring MVC currently allows interceptors to be conditionally applied using path patterns via `InterceptorRegistration`. However, there is no built-in mechanism to restrict an interceptor's execution based on HTTP methods (e.g., only for `POST` and `PUT` requests).

## Use Case

This feature would allow the configuration of interceptors that apply only to specific HTTP methods, which is useful for:

- Applying auditing logic only to mutating requests (e.g., `POST`, `DELETE`)
- Logging or metrics collection while ignoring idempotent requests (e.g., `GET`)
- Separating concerns declaratively without polluting the interceptor body with manual checks

## Current Workaround

At present, developers must implement method checks inside the interceptor itself, for example:

```java
if (!request.getMethod().equals("POST")) {
    return true;
}
```

This leads to repetitive and manual handling in every interceptor where such conditional logic is needed.

## Proposed Solution

Introduce an `allowedMethods(HttpMethod... methods)` method in `InterceptorRegistration` that wraps the original `HandlerInterceptor` with a delegating wrapper (`MethodInterceptor`) which conditionally invokes:

- `preHandle`
- `postHandle`
- `afterCompletion`

Only if the request method matches one of the specified methods.

Example configuration:

```java
registry.addInterceptor(new AuditInterceptor())
        .addPathPatterns("/api/**")
        .allowedMethods(HttpMethod.POST, HttpMethod.PUT);
```

This would ensure that the interceptor executes only for `POST` and `PUT` requests while skipping other methods such as `GET`, `OPTIONS`, etc., without interrupting request flow.

## Status

An implementation has been prepared and tested. It includes:

- A new `MethodInterceptor` class
- Integration with `InterceptorRegistration`
- Assertion to enforce that at least one HTTP method is specified
- No breaking changes to existing behavior
- Full test coverage

I am happy to submit a pull request for this enhancement and respond to any review feedback from the maintainers.

Please advise if this feature aligns with the direction of the framework.

bclozel:
Superseded by #35273

PR #35273, opened 2025-08-02
HTTP method support in MappedInterceptor

SRIRAM9487:
# Support HTTP method-restricted interceptors in `InterceptorRegistration`

This PR implements [#35272](https://github.com/spring-projects/spring-framework/issues/35272), adding support for restricting `HandlerInterceptor` registration to specific HTTP methods.

## Motivation

Spring MVC currently allows interceptors to be applied based on path patterns, but not HTTP methods. When method-based filtering is needed, it must be manually implemented within the interceptor itself:

```java
@Override
public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
    if (!"POST".equals(request.getMethod())) {
        return true;
    }
    // Logic here
    return true;
}
```

This approach clutters interceptor logic with method checks and duplicates similar logic across interceptors when needed. Centralizing HTTP method filtering in the registration API provides a cleaner and more declarative configuration style.

## Solution

This PR introduces a new method to `InterceptorRegistration`:

```java
public InterceptorRegistration allowedMethods(HttpMethod... methods)
```

This allows restricting the interceptor to specific HTTP methods like `GET`, `POST`, etc. Under the hood, the framework wraps the given `HandlerInterceptor` in a new `MethodInterceptor` that delegates calls only when the request's HTTP method matches one of the allowed methods.

```java
registry.addInterceptor(new MyInterceptor())
        .addPathPatterns("/api/**")
        .allowedMethods(HttpMethod.POST, HttpMethod.PUT);
```

The interceptor will now be invoked **only** for POST and PUT requests matching `/api/**`.

## Implementation Details

- `InterceptorRegistration` stores allowed HTTP methods as a `Set<String>`.
- If any methods are configured, it wraps the given interceptor using the new `MethodInterceptor`.
- `MethodInterceptor` checks the request's method before delegating to the original interceptor.
- The implementation is fully backward compatible; if no methods are provided, all HTTP methods are allowed (default behavior).

## Tests

A dedicated test class `MethodInterceptorTests` is included to validate:

- Delegation occurs only for allowed HTTP methods.
- Interceptor is skipped for disallowed methods.
- All interceptor lifecycle methods (`preHandle`, `postHandle`, `afterCompletion`) are covered.

## Example

```java
@Configuration
public class WebConfig implements WebMvcConfigurer {

    @Override
    public void addInterceptors(InterceptorRegistry registry) {
        registry.addInterceptor(new AuditInterceptor())
                .addPathPatterns("/admin/**")
                .allowedMethods(HttpMethod.GET);
    }
}
```

In this example, `AuditInterceptor` will apply only for `GET` requests under `/admin/**`.

## Compatibility

- No breaking changes introduced.
- The feature is opt-in.
- Reuses existing Spring MVC components (`MappedInterceptor`, `HandlerInterceptor`) without exposing new public types.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/InterceptorRegistration.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/InterceptorRegistration.java
@@ -22,6 +22,7 @@ import java.util.List;
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.http.HttpMethod;
 import org.springframework.util.AntPathMatcher;
 import org.springframework.util.Assert;
 import org.springframework.util.PathMatcher;
@@ -47,6 +48,10 @@ public class InterceptorRegistration {
 	private @Nullable List<String> includePatterns;
 
 	private @Nullable List<String> excludePatterns;
+
+	private @Nullable List<HttpMethod> includeHttpMethods;
+
+	private @Nullable List<HttpMethod> excludeHttpMethods;
 
 	private @Nullable PathMatcher pathMatcher;
 
@@ -107,6 +112,46 @@ public class InterceptorRegistration {
 	}
 
 	/**
+	 * Add HTTP methods the interceptor should be included for.
+	 * <p>Only requests with these HTTP methods will be intercepted.
+	 * @since 7.0.x
+	 */
+	public InterceptorRegistration includeHttpMethods(HttpMethod... httpMethods) {
+		return includeHttpMethods(Arrays.asList(httpMethods));
+	}
+
+	/**
+	 * List-based variant of {@link #includeHttpMethods(HttpMethod...)}.
+	 * @since 7.0.x
+	 */
+	public InterceptorRegistration includeHttpMethods(List<HttpMethod> httpMethods) {
+		this.includeHttpMethods = (this.includeHttpMethods != null ?
+				this.includeHttpMethods : new ArrayList<>(httpMethods.size()));
+		this.includeHttpMethods.addAll(httpMethods);
+		return this;
+	}
+
+	/**
+	 * Add HTTP methods the interceptor should be excluded from.
+	 * <p>Requests with these HTTP methods will be ignored by the interceptor.
+	 * @since 7.0.x
+	 */
+	public InterceptorRegistration excludeHttpMethods(HttpMethod... httpMethods){
+		return this.excludeHttpMethods(Arrays.asList(httpMethods));
+	}
+
+	/**
+	 * List-based variant of {@link #excludeHttpMethods(HttpMethod...)}.
+	 * @since 7.0.x
+	 */
+	public InterceptorRegistration excludeHttpMethods(List<HttpMethod> httpMethods){
+		this.excludeHttpMethods = (this.excludeHttpMethods != null ?
+				this.excludeHttpMethods : new ArrayList<>(httpMethods.size()));
+		this.excludeHttpMethods.addAll(httpMethods);
+		return this;
+	}
+
+	/**
 	 * Configure the PathMatcher to use to match URL paths with against include
 	 * and exclude patterns.
 	 * <p>This is an advanced property that should be used only when a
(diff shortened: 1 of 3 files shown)


SRIRAM9487:
@bclozel Thanks for the feedback on #35272. Understood that this feature is not currently planned.

I'll consider maintaining it as an external utility or extension module, since I still believe this helps reduce boilerplate and improves clarity in interceptor setups.

Appreciate your time and the review!

bclozel:
We have not declined nor reviewed this yet. We don't need both an issue and a PR so we closed the other one.

SRIRAM9487:
Thanks for the clarification, @bclozel.
Looking forward to your review once it's in the queue. Let me know if anything needs to be adjusted in the meantime!

rissay2015:
Until then (the review by those grannies) : 
   ->You can work on exclude methodes  : applying that interceptor to all except this methodes ....
   -> feels like we must fall back to manual logic inside the interceptor for anything more complex.
Otherwise that was impressive young man.

spring-aop/.factorypath
```diff
@@ -0,0 +1,25 @@
...
+    <factorypathentry kind="EXTJAR" id="/home/sriram/.gradle/caches/modules-2/files-2.1/com.google.auto.service/auto-service-annotations/1.0.1/ac86dacc0eb9285ea9d42eee6aad8629ca3a7432/auto-service-annotations-1.0.1.jar" enabled="true" runInBatchMode="false"/>
+    <factorypathentry kind="EXTJAR" id="/home/sriram/.gradle/caches/modules-2/files-2.1/com.google.protobuf/protobuf-java/3.25.5/5ae5c9ec39930ae9b5a61b32b93288818ec05ec1/protobuf-java-3.25.5.jar" enabled="true" runInBatchMode="false"/>
+    <factorypathentry kind="EXTJAR" id="/home/sriram/.gradle/caches/modules-2/files-2.1/org.pcollections/pcollections/4.0.1/59f3bf5fb28c5f5386804dcf129267416b75d7c/pcollections-4.0.1.jar" enabled="true" runInBatchMode="false"/>
+    <factorypathentry kind="EXTJAR" id="/home/sriram/.gradle/caches/modules-2/files-2.1/com.uber.nullaway/nullaway/0.12.7/1a813b7d156768204b0c8d52ba0632a8db6fbe12/nullaway-0.12.7.jar" enabled="true" runInBatchMode="false"/>
+    <factorypathentry kind="EXTJAR" id="/home/sriram/.gradle/caches/modules-2/files-2.1/com.google.guava/failureaccess/1.0.3/aeaffd00d57023a2c947393ed251f0354f0985fc/failureaccess-1.0.3.jar" enabled="true" runInBatchMode="false"/>
+</factorypath>
```
rstoyanchev: These seem to have been committed by mistake. Could you please remove them from the pull request?

spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/InterceptorRegistration.java
```diff
@@ -61,10 +65,10 @@ public InterceptorRegistration(HandlerInterceptor interceptor) {
 		this.interceptor = interceptor;
 	}
 
-
 	/**
 	 * Add patterns for URLs the interceptor should be included in.
-	 * <p>For pattern syntax see {@link PathPattern} when parsed patterns
+	 * <p>
+	 * For pattern syntax see {@link PathPattern} when parsed patterns
```
rstoyanchev: There are a number of Javadoc style changes, probably applied by the IDE. Could you please undo those? They are not related and in any case inconsistent with the rest of the codebase.

spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/InterceptorRegistration.java
```diff
@@ -75,18 +79,19 @@ public InterceptorRegistration addPathPatterns(String... patterns) {
 
 	/**
 	 * List-based variant of {@link #addPathPatterns(String...)}.
+	 * 
 	 * @since 5.0.3
 	 */
 	public InterceptorRegistration addPathPatterns(List<String> patterns) {
-		this.includePatterns = (this.includePatterns != null ?
-				this.includePatterns : new ArrayList<>(patterns.size()));
+		this.includePatterns = (this.includePatterns != null ? this.includePatterns : new ArrayList<>(patterns.size()));
```
rstoyanchev: Likewise, a number of code style changes. Please, undo those.

rstoyanchev (changes requested):
Aside from the comments below, generally the request seems reasonable with the following implementation related points:

- The existing `MappedInterceptor` could be enhanced instead to support mapping by HTTP method.
- The terminology `allowedMethods` could be just `httpMethods`, or `include/excludeHttpMethods` for consistency with `include/excludePatterns`.

SRIRAM9487:
@rstoyanchev Thanks for the detailed feedback. I understand the requested changes — I’ll work on enhancing MappedInterceptor with HTTP method support, update the naming to includeHttpMethods/excludeHttpMethods, and remove the unintended commits. I should have an updated PR ready in about 2 days. Thanks again for your feedback.

spring-aop/.factorypath
SRIRAM9487: I've removed the unintended .factorypath entries from the pull request.

spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/InterceptorRegistration.java
SRIRAM9487: I've reverted the unintended Javadoc formatting changes in InterceptorRegistration.java to keep the style consistent with the rest of the codebase.

spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/InterceptorRegistration.java
SRIRAM9487: I've reverted the unintended code style changes in InterceptorRegistration.java to match the existing formatting in the codebase.

SRIRAM9487:
@rstoyanchev Thank you for reviewing and integrating this feature into Spring Framework. I appreciate the refinements in the final implementation,. I initially thought it might be merged directly, but I understand it’s a relatively small feature. Looking forward to contributing again in the future.

rstoyanchev:
We don't merge directly unless it's straight forward. In this case, simplifications were needed. Also note the constructor changes where we can't simply add arguments and break compatibility on a long existing class.

SRIRAM9487:
Got it, thanks for explaining. I understand the need for simplifications and maintaining backward compatibility, especially for long-standing classes. I’ll keep this in mind for future contributions.

Issue #35261, opened 2025-07-30
Avoid creating byte[] multiple times in StringHttpMessageConverter


brucelwl:
When calculating the ``ContentLength`` and writing data to the `OutputStream` in `StringHttpMessageConverter`, calling `str.getBytes(charset)` repeatedly will result in unnecessary array objects occupying memory

https://github.com/spring-projects/spring-framework/blob/09917fad7bca9b3997522f0a75d6319203f2127f/spring-web/src/main/java/org/springframework/http/converter/StringHttpMessageConverter.java#L103-L106

https://github.com/spring-projects/spring-framework/blob/09917fad7bca9b3997522f0a75d6319203f2127f/spring-web/src/main/java/org/springframework/http/converter/StringHttpMessageConverter.java#L122-L129

bclozel:
Superseded by #35276

We can reopen this issue if we can find a way to improve performance without failing the correct behavior.

brucelwl:
@bclozel I have identified the reason for the build failure, but my submission signature was not carried. 
It has now been fixed. see https://github.com/spring-projects/spring-framework/pull/35280 , 
Please open this issue,  
Thanks

kilink:
I've looked into ways around this in the past, one easy approach is to special case ASCII Strings when the target Charset is ASCII or an ASCII superset (like UTF-8). Proof of concept here: #35290.

UTF-8 length can be completely handled without allocating the byte array if we added a utility like the [one](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Utf8.java#L48) in Guava.

For ASCII and ISO_8859_1, I believe we can actually just use `codePoints()` to calculate the length and handle the case of surrogate pairs if we really wanted to.

IMO it would probably be best to decide whether it's worth adding a utility for UTF-8 length calculation specifically, and special case that, since I imagine it's the most commonly used encoding.

Anyway, the only way I can see to preserve backwards compatibility and also avoid the extra allocation is to special case certain character sets and just do the calculations ourselves.

rstoyanchev:
@kilink thanks for the suggestion, but it becomes then a trade-off between CPU vs memory allocation. Also optimized for the ascii case, but as a side effect less optimized for non-ascii which would require both approaches.

I think we can try to bring  closer together setting the content-length and the actual write, or at least make it possible where an optimization can be made.

[...]

PR #35276, opened 2025-08-03
Avoid creating byte[] multiple times

brucelwl:
for issue  https://github.com/spring-projects/spring-framework/issues/35261


bclozel:
Thanks for the proposal but this PR fails the build and does not implement the content length contract anymore.
I see that it could make performance better but we should not sacrifice correctness.

PR #35280, opened 2025-08-03
Avoid creating byte[] multiple times

brucelwl:
for issue https://github.com/spring-projects/spring-framework/issues/35261


bclozel:
This still breaks the contract here. Returning null means we cannot determine the content length for a String.
This is a regression and we cannot accept this change.

brucelwl:
@bclozel Do you have a better optimization solution for the issue https://github.com/spring-projects/spring-framework/issues/35261 of repeated calling of `calling str.getBytes(charset)` ?

Issue #33820, opened 2024-10-30
Remove the `pure` attribute from `@Contract`


sdeleuze:
The `@Contract` annotation introduced via #32475 has a `pure` attribute that is coming from its `org.jetbrains.annotations.Contract` counterpart. It was historically used to trigger warnings when the return value is not used, but it looks like [it is not the case anymore with IDEA 2023.1+](https://youtrack.jetbrains.com/issue/IDEA-315247) and the `@CheckReturnValue` annotation introduced in #33818 provides a more explicit signal, so we prefer to remove the `pure` attribute from `@Contract` in Spring Framework 6.2.0, letting to door open to a future reintroduction only if strong use cases arise.

PR #35285, opened 2025-08-04
Refine `@Contract` Javadoc

scordio:
This commit removes references to the `pure` attribute.

Relates to gh-33820.


DhruvTheDev1 (commented):
Removed the depreciated `pure` attribute and replaced with a more general and clearer description.

PR #35290, opened 2025-08-05
Optimize getContentLength for ASCII Strings

kilink:
If the provided String contains only ASCII characters, and the target encoding is ASCII or an ASCII superset, avoid calling getBytes on the String and allocating a byte array unnecessarily.


rstoyanchev:
Closing for now as we're likely to try a different approach as per discussion under #35261.

PR #35294, opened 2025-08-06
Make `type` in `ProblemDetail` nullable

cw-dimedis:
### Background

According to [RFC 9457 Section 3.1.1](https://www.rfc-editor.org/rfc/rfc9457.html#name-type) the "type" member is optional and when not present, its value is assumed to be "about:blank". The current Spring implementation explicitly sets this default value, which results in all problem responses containing the unhelpful `"type": "about:blank"` field even when no specific problem type is needed.

### Changes

This change makes the `type` field nullable in `ProblemDetail`:
- Updated `setType(@Nullable URI type)` to accept null values
- Updated `getType()` to return `@Nullable URI`
- Removed the `Assert.notNull()` validation that prevented null types

### Compatibility

**This change is completely non-breaking.** Existing users will not notice any difference in behavior:

- The `type` field is still initialized to `BLANK_TYPE` ("about:blank") by default
- Existing code that relies on the type always being present continues to work unchanged
- JSON serialization will continue to show `"type": "about:blank"` for existing usage patterns

The only difference is that developers can now **optionally** set the type to `null` if they want cleaner JSON output without the default "about:blank" value.

### Benefits

1. **Optional cleaner JSON output**: Developers can now choose to omit the type field entirely by setting it to null
2. **RFC compliance**: Aligns with the RFC 9457 specification where the type field is truly optional
3. **Zero impact on existing code**: All current usage patterns continue to work exactly as before

### Example

For users who want cleaner output, they can now do:
```java
ProblemDetail problem = ProblemDetail.forStatus(400);
problem.setType(null); // Now possible - results in cleaner JSON
```

This change provides more flexibility while maintaining complete backward compatibility.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ProblemDetail.java
+++ b/spring-web/src/main/java/org/springframework/http/ProblemDetail.java
@@ -107,15 +107,14 @@ public class ProblemDetail implements Serializable {
 	 * <p>By default, this is {@link #BLANK_TYPE}.
 	 * @param type the problem type
 	 */
-	public void setType(URI type) {
-		Assert.notNull(type, "'type' is required");
+	public void setType(@Nullable URI type) {
 		this.type = type;
 	}
 
 	/**
 	 * Return the configured {@link #setType(URI) problem type}.
 	 */
-	public URI getType() {
+	public @Nullable URI getType() {
 		return this.type;
 	}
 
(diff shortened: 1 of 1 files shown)


rstoyanchev (commented):
Thanks for the suggestion. 

The `@Nullable` on the getter can cause issues for existing code. In 6.2.x we could avoid adding those and simply remove the `Assert.notNull`. It would put the burden on code setting a `null` value, but at least  make it possible.

For 7.0 in the main branch we can accept the changes as is.

cw-dimedis:
@rstoyanchev Thanks for the feedback and the pragmatic approach for 6.2.x vs 7.0!

For 7.0, I'd suggest we could go further and remove the default `BLANK_TYPE` initialization entirely. The field would be `null` by default, giving cleaner JSON output out of the box and full RFC 9457 compliance without requiring developers to explicitly set `type` to null.

Is that something the team would consider? Shall I add another commit to prepare that change?

Cheers, Christoph

rstoyanchev:
That makes sense. Please, add the extra change.

PR #35301, opened 2025-08-09
Allow `Type` reference instead of `Class` reference in `RestTemplate`'s convenience methods.

sam-hieken:
## Issue

The convenience methods `*ForEntity` and `*ForObject` in `RestTemplate` don't allow generic types to be specified in the result body. For example: 

```
public class Main {
	static class ExampleType {
		...
	}

		RestTemplate restTemplate = ...;
		URI url = ...;
		
                // Type mismatch: cannot convert from ResponseEntity<List> to ResponseEntity<List<Main.ExampleType>>
		ResponseEntity<List<ExampleType>> result = restTemplate.getForEntity(url, List.class);
        }
}
```

Using `Type` as the type of the `responseType` parameters instead of `Class<T>` in these methods would allow generic types to be used as well (and since `Class` extends `Type`, changing it wouldn't be a breaking change):

```
public class Main {
	static class ExampleType {
		...
	}
	
		RestTemplate restTemplate = ...;
		URI url = ...;
		
		// Example of how we can now allow generic types using com.fasterxml.jackson.core.type.TypeReference
		ResponseEntity<List<ExampleType>> result = restTemplate.getForEntity(url, new TypeReference<List<ExampleType>>() {}.getType());
}
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
+++ b/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java
@@ -477,7 +477,7 @@ public class RestTemplate extends InterceptingHttpAccessor implements RestOperat
 	// PATCH
 
 	@Override
-	public <T> @Nullable T patchForObject(String url, @Nullable Object request, Class<T> responseType,
+	public <T> @Nullable T patchForObject(String url, @Nullable Object request, Type responseType,
 			@Nullable Object... uriVariables) throws RestClientException {
 
 		RequestCallback requestCallback = httpEntityCallback(request, responseType);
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks for the proposal but this is well supported already with ParameterizedTypeReference, see https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-resttemplate

Also I don't think this change would have the desired behavior as the generic type information is probably not handled by messages readers in this way.

Thanks!

sam-hieken:
Thanks for taking a look Brian

PR #35303, opened 2025-08-11
Add test for resolving RequestParam through SpEL

qchole:
Add test case in `RequestParamMethodArgumentResolverTests` resolving request parameter naming through SpEL


rstoyanchev:
The functionality is in a base class `AbstractNamedValueMethodArgumentResolver` that supports quite a few of the resolver implementations. We already have a test `ExpressionValueMethodArgumentResolverTests`, and don't need to test it for every resolver. Thanks for the suggestion in any case.

PR #35307, opened 2025-08-11
Only add `httpServiceProxyRegistry` bean when necessary

philwebb:
Update `AbstractHttpServiceRegistrar` so that the
`httpServiceProxyRegistry` bean  is only added when registrations are found.


philwebb:
This update helps Spring Boot since we always have a `AbstractClientHttpServiceRegistrar` but it might not find any `@HttpServiceClient` interfaces. Another option would be to make the `registerBeanDefinitions` non-final so that we can override them.

PR #35329, opened 2025-08-14
Restore mixed usage support with uri() and queryParam() for query parameters in AbstractMockHttpServletRequestBuilder.buildRequest()

izeye:
I'm not sure if it was intentional, but 8ec0c21b0a0ef6c725c27d787c4e27ae5fe328f4 has changed the behavior when both `uri()` and `queryParam()` are used for the same query parameters by overwriting the former with the latter rather than merging both.

See gh-35210


rstoyanchev:
The source `parameters` is a map, but indeed parameters can be added from the `queryString` as well, so thanks for catching that. For both of these we need to ensure correct handling of params without a value.

rstoyanchev:
Thanks very much for spotting this so quickly. I've applied some further updates related to the original goal of #35210.

PR #35335, opened 2025-08-17
Improve documentation for `ApplicationEvents` to clarify recommended usage

### LINKED ISSUE #35297: Improve Javadoc of `ApplicationEvents`
`ApplicationEvents`' Javadoc talks about using it via an autowired field *before* it talks about its use as method parameter. While not explicitly stated, this seems to create the impression that `ApplicationEvents` is a Spring bean generally eligible for dependency injection. Thus, developers seem to be puzzled to learn that they can not constructor-inject `ApplicationEvents` (see the (German) description [here](https://github.com/denwae/ma_problems_and_solutions/blob/main/other/constructor_injection.md)).

I think we should direct developers to use `ApplicationEvents` as a method parameter primarily as its  lifecycle is per-method anyway.

khj68:
## Description
This PR improves the Javadoc of `ApplicationEvents` to better guide developers on its proper usage.

## Problem
The current Javadoc mentions field injection with `@Autowired` before method parameter injection, which creates the impression that `ApplicationEvents` is a regular Spring bean eligible for dependency injection. This confuses developers who then attempt to use constructor injection and find it doesn't work.

## Solution
Reordered and clarified the usage instructions to:
1. **Recommend method parameter injection as the primary approach** - since `ApplicationEvents` has a per-method lifecycle, this is the most appropriate usage pattern
2. **Clarify that `ApplicationEvents` is not a general Spring bean** - explicitly state that it's specifically designed for use within test methods
3. **Position field injection as an alternative** - rather than the primary approach

## Changes
- Reordered the bullet points to show method parameter injection first
- Added clarification about the per-method lifecycle
- Added a note that `ApplicationEvents` is not a general Spring bean

This helps prevent the confusion described in the linked German documentation where developers were puzzled by the inability to use constructor injection.

Closes gh-35297

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/event/ApplicationEvents.java
+++ b/spring-test/src/main/java/org/springframework/test/context/event/ApplicationEvents.java
@@ -33,12 +33,14 @@ import org.springframework.context.ApplicationEvent;
  * to be manually registered if you have custom configuration via
  * {@link org.springframework.test.context.TestExecutionListeners @TestExecutionListeners}
  * that does not include the default listeners.</li>
- * <li>Annotate a field of type {@code ApplicationEvents} with
+ * <li>With JUnit Jupiter, declare a parameter of type {@code ApplicationEvents}
+ * in a test or lifecycle method. Since {@code ApplicationEvents} is scoped to the
+ * lifecycle of the current test method, this is the recommended approach.</li>
+ * <li>Alternatively, you can annotate a field of type {@code ApplicationEvents} with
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @khj68,

Please note that #35297 was assigned to me, and we typically do not accept PRs for issues assigned to someone.

In any case, this PR served as a good starting point.

Thus I have merged this into `main` in c0b71f8999c8c171f26d3dbbba9668f3bef3ad3a and completed the work in 19d5ec67811e2f74bf08b1c909a609ead4380acc.

Cheers,

Sam

PR #35336, opened 2025-08-17
Optimize CacheOperation hashCode() performance

### LINKED ISSUE #35317: @Cacheable performance: CacheOperation spends significant time when computing hashCode
The benchmark is to call `@Cacheable` method: https://github.com/vlsi/aop-benchmarks/blob/f304fb811910aca8dd97c0fc0ced8d1b45d8a62b/src/main/java/io/github/vlsi/benchmarks/aop/cacheable/PlainCalculator.java#L6

Spring 6.2.9 + Java 21 results in ~500 bytes allocated per call.

I see notable time is spent in `hashCode`: https://github.com/spring-projects/spring-framework/blob/8ec0c21b0a0ef6c725c27d787c4e27ae5fe328f4/spring-context/src/main/java/org/springframework/cache/interceptor/CacheOperation.java#L115C10-L115C18

I believe the root cause is that `CacheOperation` itself is not cached, so it results in building the new string every time, thus JVM has to recompute the string every time.

Of course, it would be nice to reuse `CacheOperation`, however, it might be a more significant change, so I suggest a simpler approach instead:

1) Build tostring lazily in case it would really be needed. I assume `CacheOperation#toString()` should never really be called in production
2) Compute the hashcode in the constructor based on the hashcodes of the components
3) Use subcomponents for `equals` instead of `toString().equals(that.toString())`

Note: I believe the issue was not fixed with https://github.com/spring-projects/spring-framework/pull/957 as #957 did not remove `toString()` from the hot path.

## Profiling results

This was profiled with async-profiler 4, macOS, Java 21

Image

Async profiler results (in JFR format): [cacheable_jfrs.zip](https://github.com/user-attachments/files/21757450/cacheable_jfrs.zip)

khj68:
## Description
This PR optimizes the performance of `CacheOperation.hashCode()` by caching the hash code value during construction.

## Problem
As reported in gh-35317, `@Cacheable` method calls result in ~500 bytes allocated per call due to repeated `hashCode()` computation. The profiling showed that significant time is spent in `CacheOperation.hashCode()` which calls `toString().hashCode()` on every invocation.

While the `toString` value is already cached as a field, the hash code was being recomputed on every call, causing unnecessary overhead in high-throughput scenarios.

## Solution
Cache the hash code value as a final field during object construction, similar to how the `toString` value is already cached. This makes `hashCode()` calls O(1) instead of depending on string length.

## Changes
- Added a `private final int hashCode` field to `CacheOperation`
- Compute the hash code once in the constructor based on the cached `toString` value
- Modified `hashCode()` to return the cached value
- Updated JavaDoc to reflect the caching behavior

## Performance Impact
- Reduces allocations by ~500 bytes per `@Cacheable` method call
- Eliminates `toString().hashCode()` from the hot path
- Makes `hashCode()` calls constant time O(1)
- Improves cache operation performance in high-throughput scenarios

## Testing
The change is backward compatible and doesn't affect the behavior, only the performance. Existing tests should continue to pass.

Closes gh-35317

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/CacheOperation.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/CacheOperation.java
@@ -50,6 +50,8 @@ public abstract class CacheOperation implements BasicOperation {
 
 	private final String toString;
 
+	private final int hashCode;
+
 
 	/**
 	 * Create a new {@link CacheOperation} instance from the given builder.
@@ -64,6 +66,7 @@ public abstract class CacheOperation implements BasicOperation {
 		this.cacheResolver = b.cacheResolver;
 		this.condition = b.condition;
 		this.toString = b.getOperationDescription().toString();
+		this.hashCode = this.toString.hashCode();
 	}
 
 
@@ -107,12 +110,13 @@ public abstract class CacheOperation implements BasicOperation {
 	}
 
 	/**
-	 * This implementation returns {@code toString()}'s hash code.
+	 * This implementation returns the cached hash code computed
+	 * during construction based on {@code toString()}'s hash code.
 	 * @see #toString()
 	 */
 	@Override
 	public int hashCode() {
-		return toString().hashCode();
+		return this.hashCode;
 	}
 
 	/**


sbrannen:
Hi @khj68,

Note that #35317 is already assigned to @jhoeller, which indicates that he will resolve the issue.

In the future, please do not submit a PR for an issue assigned to someone, before asking that person if a PR is desired.

In addition, this PR does not implement the change the way we plan to.

PR #35344, opened 2025-08-19
Fall back to JDK proxy if CGLIB proxy creation fails

### LINKED ISSUE #31634: Improvement: Fall back from cglib to JDK proxies when possible
I am using Spring Boot 2.7, which defaults to cglib proxies. I have a JavaConfig bean definition:

```
interface ServiceInterface {
  @Async void log()
}

@Bean
ServiceInterface myBean() {
  new Impl()
}
```

If the `Impl` class is not subclassable by cglib (in this case, it uses a static factory method and has a private constructor), context startup fails with a "cglib can't subclass" error. However, if the _declared bean type is an interface_, in my view the ProxyFactory ought to use a JDK proxy instead (in the first place, but certainly if cglib fails).

YongGoose:
Prior to this change, if the default CGLIB proxy creation failed for a non-subclassable target class (e.g., due to a private constructor), the ApplicationContext would fail to start.

fixes #31634

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java
@@ -79,6 +79,7 @@ import org.springframework.util.ObjectUtils;
  * @author Chris Beams
  * @author Dave Syer
  * @author Sebastien Deleuze
+ * @author Yongjun Hong
  * @see org.springframework.cglib.proxy.Enhancer
  * @see AdvisedSupport#setProxyTargetClass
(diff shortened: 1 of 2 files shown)


YongGoose:
@sbrannen 
I think this feature could be helpful for the Spring Framework. 
Could you give me some feedback on it?

- https://github.com/spring-projects/spring-framework/issues/31634#issuecomment-3200350838

YongGoose:
@sbrannen 
When you have time, could you please take a look at this PR? 🙇🏻‍♂️🙇🏻‍♂️

ps. Happy new year!!

PR #35350, opened 2025-08-19
class level support for @JsonView in JsonViewResponseBodyAdvice

asifebrahim:
Made the necessary changes with a test file
### Closes 35343
Signed-off-by: Asif Ebrahim <asifebrahim13@gmail.com>

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/JsonViewResponseBodyAdvice.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/JsonViewResponseBodyAdvice.java
@@ -33,11 +33,14 @@ import org.springframework.util.Assert;
 /**
  * A {@link ResponseBodyAdvice} implementation that adds support for Jackson's
  * {@code @JsonView} annotation declared on a Spring MVC {@code @RequestMapping}
- * or {@code @ExceptionHandler} method.
+ * or {@code @ExceptionHandler} method, or at the controller class level.
  *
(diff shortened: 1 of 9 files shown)


sdeleuze:
Hey @asifebrahim, I am not sure to understand why you closed this PR which seems to fix #35343 before we had a chance to review it. Could you explain?

asifebrahim:
i thought someone else was assigned to it so that's why..I'll reopen

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/JsonViewResponseBodyAdvice.java
```diff
@@ -53,7 +56,9 @@ public class JsonViewResponseBodyAdvice extends AbstractMappingJacksonResponseBo
 
 	@Override
 	public boolean supports(MethodParameter returnType, Class<? extends HttpMessageConverter<?>> converterType) {
-		return super.supports(returnType, converterType) && returnType.hasMethodAnnotation(JsonView.class);
+		return super.supports(returnType, converterType) && 
+			(returnType.hasMethodAnnotation(JsonView.class) || 
+			 returnType.getDeclaringClass().isAnnotationPresent(JsonView.class));
```
quaff, line 61: Should `@JsonView` on super type be considered?

sdeleuze:
After more thoughts and discussion with the team, we prefer to decline this feature originally proposed in #35343 because the `@JsonView` class attribute is strongly related to the return type of the method, so specifying that at class level looks confusing from a programming model perspective.

PR #35358, opened 2025-08-20
ServerRequestObservationConvention compliant with OpenTelemetry semantic conventions for Servlet HTTP server

shakuzen:
Based on the stable semantic conventions for HTTP server in version 1.36.0: https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/README.md.

Users who wish to use this can pass an instance of the convention to the constructor of `ServerHttpObservationFilter`.


bclozel:
Thanks @shakuzen !

PR #35366, opened 2025-08-21
Optimize WebFlux multipart upload performance

### LINKED ISSUE #34651: Slow webflux multipart upload
Hello,

We are developing a document submission application using WebFlux. However, we have noticed that handling multipart requests is slower with WebFlux compared to MVC. For example, when using MultipartFile in MVC, sending a 200MB file takes about 700ms, whereas it takes 4.2s with WebFlux (FilePart) and 4.5s with PartEvent.

Did we miss something?

Here is a repository reproducing the issue: [Repo](https://github.com/sbxz/WebfluxSlowMultipart)

Spring boot version : 3.4.4
Java 21

xyraclius:
## 🚀 Overview

This PR improves performance in WebFlux multipart upload processing by optimizing how `AbstractNestedMatcher` scans `DataBuffer` instances for delimiters.

---

## 🔥 Motivation

Multipart uploads in WebFlux currently suffer from slower performance compared to Spring MVC, especially with large files. A significant bottleneck was found in the delimiter matching logic, which processed buffers one byte at a time and caused unnecessary overhead.

---

## 🔧 Changes

- Reworked `AbstractNestedMatcher` to use:
  - ~~A **thread-local buffer** (`LOCAL_BUFFER`) to avoid per-call allocations.~~  
  Replaced with an **instance-local buffer** (`localBuffer`) to simplify buffer management.
  - **Chunk-based scanning** instead of byte-by-byte iteration.
- Extracted helper methods (`processChunk`, `findNextCandidate`,
  `updateMatchIndex`) to reduce complexity and improve readability.
- Preserved existing API and behavior for backward compatibility.
- Added clear Javadoc to explain usage and limitations.

---

## ✅ Benefits

- Reduced allocation overhead during large multipart uploads.
- Improved throughput when scanning large `DataBuffer` instances.
- Better maintainability through clearer structure and documentation.
- **No breaking changes**. Existing API and behavior remain intact.
- **No new tests required**. Code is already fully covered (100%) by
  existing unit tests.

---

## 📈 Performance Impact

- 200MB file via `MultipartFile` (Spring MVC): ~700 ms  
- 200MB file via `FilePart` (WebFlux): ~4.1 s  
- 200MB file via `PartEvent` (WebFlux): ~4.2 s  

After this change, large multipart uploads in WebFlux `no longer` suffer from excessive overhead in delimiter scanning.

#### Before Optimization

```bash
curl -w "Elapsed: %{time_total}s\n" \
  -F file="@200mb.pdf" \
  http://localhost:8080/file-part \
  -o /dev/null -s
# Elapsed: 4.177509s

curl -w "Elapsed: %{time_total}s\n" \
  -F file="@200mb.pdf" \
  http://localhost:8080/part-event \
  -o /dev/null -s
# Elapsed: 4.224732s
```

#### After Optimization

```bash
curl -w "Elapsed: %{time_total}s\n" \
  -F file="@200mb.pdf" \
  http://localhost:8080/file-part \
  -o /dev/null -s
# Elapsed: 0.437359s

curl -w "Elapsed: %{time_total}s\n" \
  -F file="@200mb.pdf" \
  http://localhost:8080/part-event \
  -o /dev/null -s
# Elapsed: 0.317705s


#Test 19.53GB file
curl -w "\nElapsed: %{time_total}s\n" \
  -F file="@game.rar" \
  http://localhost:8080/file-part \
  -o /dev/null -s
# Elapsed: 27.586772s

curl -w "Elapsed: %{time_total}s\n" \
     -F file="@game.rar" \
     http://localhost:8080/part-event \
     -o /dev/null -s
# Elapsed: 18.328275s


#Test Multiple files
curl -w "Elapsed: %{time_total}s\n" \
  -F file="@200mb.pdf" \
  -F file="@game.rar" \
  http://localhost:8080/file-part \
  -o /dev/null -s
# Elapsed: 23.678042s

curl -w "Elapsed: %{time_total}s\n" \
     -F file="@200mb.pdf" \
     -F file="@game.rar" \
     http://localhost:8080/part-event \
     -o /dev/null -s
# Elapsed: 17.001363s
```

---
## Example
```java
@PostMapping(path = "/file-part", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
public Mono<Void> uploadFilePart(@RequestPart("file") FilePart file) {
    log.info("Received file part request {}", file.filename());
    return Mono.empty();
}

@PostMapping(path = "/part-event", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
public Mono<Void> uploadPartEvent(@RequestBody Flux<PartEvent> allPartsEvents) {
    return allPartsEvents.windowUntil(PartEvent::isLast)
        .concatMap(p -> p.switchOnFirst((signal, partEvents) -> {
            if (signal.hasValue()) {
                PartEvent event = signal.get();
                if (event instanceof FormPartEvent formEvent) {
                    return Mono.error(new RuntimeException("Unexpected event: " + event));
                } else if (event instanceof FilePartEvent fileEvent) {
                    log.info("Received part event request {}", fileEvent.filename());
                    return partEvents
                            .ofType(FilePartEvent.class)
                            .doOnNext(file -> DataBufferUtils.release(file.content()))
                            .then(Mono.empty());
                } else {
                    return Mono.error(new RuntimeException("Unexpected event: " + event));
                }
            } else {
                return partEvents;
            }
        })).then();
}

```
---

## Related Issue

Closes #34651

---

## Checklist

- [x] Code changes follow [Spring Framework contributing guidelines](https://github.com/spring-projects/spring-framework/blob/main/CONTRIBUTING.md).
- [x] Javadoc updated where needed.
- [x] Verified no breaking changes introduced.
- [x] Verified code is fully covered by existing unit tests (100%).


bclozel:
Thanks for raising this @xyraclius , but I'm not sure the approach is valid.

For WebFlux applications, there is no assumption about the processing of a single request. Unlike Servlet applications where the processing of a request is happening on a single thread, reactive apps can schedule work on many different threads. Isn't using a `ThreadLocal` likely to create concurrency issues if many multipart requests happen in parallel?

xyraclius:
Hi @bclozel,

I initially used ThreadLocal to reduce per-call allocations and improve CPU/memory usage. That said, I now understand that in WebFlux a single request can run across multiple threads, so using ThreadLocal could be unsafe.

The safest approach would be to switch to a per-request local buffer, like:

```
final byte[] chunk = new byte[8 * 1024];
```

This avoids any concurrency issues while keeping allocations reasonable. I can implement this change and test the performance to make sure we maintain the improvements.

sdeleuze:
Yeah please refine accordingly and we will review it.

xyraclius:
Hi @bclozel @sdeleuze,
Refined the change to use a per-instance buffer instead of ThreadLocal, as suggested. 
✅ Verified correctness and perf still look good.

Please let me know if further adjustments are needed. Thanks for the guidance! 🙏

spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
```diff
@@ -859,14 +860,28 @@ public void reset() {
...
+	 * @see NestedMatcher
+	 * @see DataBuffer
 	 */
 	private abstract static class AbstractNestedMatcher implements NestedMatcher {
 
 		private final byte[] delimiter;
 
 		private int matches = 0;
 
+		private final byte[] localBuffer = new byte[8 * 1024]; // Reusable buffer per matcher instance
```
DhruvTheDev1, line 884: Good switch from a threadLocal buffer to per-instance buffer as in WebFlux, where requests may span across threads using threadLocal can cause concurrency issues so the change is much better for multi-threaded environments.

spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java
```diff
@@ -882,14 +897,79 @@ protected int getMatches() {
...
+			final byte[] delimiterBytes = this.delimiter;
+			final int delimiterLength = delimiterBytes.length;
+			final byte delimiterFirstByte = delimiterBytes[0];
+
+			final byte[] chunk = localBuffer;
+			final int chunkSize = Math.min(chunk.length, length);
+
+			int matchIndex = this.matches;
+
+			try {
```
DhruvTheDev1, line 913: Nice approach to improve performance by using chunk based scanning as it reduces overhead of processing large buffers especially when uploading large files

DhruvTheDev1 (commented):
It is great to see the performance benchmarks clearly showing the significant improvements in WebFlux multipart upload performance especially for large files. The real-world file tests also show the optimisation can handle large-scale uploads seamlessly.

Also nice on making the PR so clear and easy to follow! It's evident that a lot of thought went into it.

bclozel:
Thanks for the proposal @xyraclius - I have tested it against the reproduced and it does bring important improvements. However, I don't think we can apply this for several reasons: this create one 8K buffer per matcher (and those can be nested multiple times), effectively trading throughput in favor of memory consumption for all multipart bodies no matter what their size. The proposed change is quite involved and not easy to maintain.

Let's take a step back in #34651 and see what kind of improvements we can make there.

Thanks again for your contribution!

PR #35370, opened 2025-08-22
Add support for relative ordering

### LINKED ISSUE #24844: Add support for relative ordering

YongGoose:
fixes #24844

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/CyclicOrderException.java
+++ b/spring-core/src/main/java/org/springframework/core/CyclicOrderException.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2002-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


YongGoose:
This PR is not fully complete yet, but I’ve marked it as ready for review to get some feedback.
I’d appreciate any feedback!! Also, I’ll continue the work based on that.

Issue #35369, opened 2025-08-22
`SingletonSupplier` is not singleton if `null` is returned on first supplier call


mhalbritter:
When `null` is returned from the supplier passed to `SingletonSupplier` and some non-null value on subsequent calls, it breaks the singleton nature.

It returns on first call of `singletonSupplier.get()` `null` and then afterwards `Hello`.

```java
private static int call = 0;

public static void main(String[] args) {
    SingletonSupplier<String> singletonSupplier = SingletonSupplier.ofNullable(() -> supply());
    System.out.println(singletonSupplier.get());
    System.out.println(singletonSupplier.get());
    System.out.println(singletonSupplier.get());
    System.out.println(singletonSupplier.get());
}

private static String supply() {
    call++;
    if (call % 2 == 1) {
        return null;
    }
    return "Hello";
}
```

prints

```
null
Hello
Hello
Hello
```

In Framework 7 it has this signature:

```java
public static <T> @Nullable SingletonSupplier<T> ofNullable(@Nullable Supplier<@Nullable T> supplier) {
```

suggesting that using a `Supplier<@Nullable String>` is fine.

sbrannen:
Indeed, the following line stores `null` in the `singleInstance` field in such use cases.

https://github.com/spring-projects/spring-framework/blob/8f4107953d7ff09f20fb4285c22b33f2ac6e37e0/spring-core/src/main/java/org/springframework/util/function/SingletonSupplier.java#L109

And for a subsequent `get()`, a `null` value indicates that another attempt should be made to retrieve the value from the `Supplier`.

In other words, `SingletonSupplier` does not track that `null` was actually returned from `instanceSupplier` or `defaultSupplier`.

Thanks for raising the issue and providing the reproducer.

We'll look into it.

sbrannen:
- Superseded by #35380

PR #35380, opened 2025-08-23
`SingletonSupplier` is not singleton if `null` is returned on first supplier call

nosan:
See https://github.com/spring-projects/spring-framework/issues/35369

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/function/SingletonSupplier.java
+++ b/spring-core/src/main/java/org/springframework/util/function/SingletonSupplier.java
@@ -47,6 +47,7 @@ public class SingletonSupplier<T> implements Supplier<@Nullable T> {
 
 	private volatile @Nullable T singletonInstance;
 
+	private volatile boolean initialized;
 	/**
 	 * Guards access to write operations on the {@code singletonInstance} field.
 	 */
@@ -63,6 +64,7 @@ public class SingletonSupplier<T> implements Supplier<@Nullable T> {
 		this.instanceSupplier = null;
 		this.defaultSupplier = defaultSupplier;
 		this.singletonInstance = instance;
+		this.initialized = (instance != null);
 	}
 
 	/**
@@ -85,6 +87,7 @@ public class SingletonSupplier<T> implements Supplier<@Nullable T> {
 		this.instanceSupplier = null;
 		this.defaultSupplier = null;
 		this.singletonInstance = singletonInstance;
+		this.initialized = (singletonInstance != null);
 	}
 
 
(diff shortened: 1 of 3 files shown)


spring-core/src/main/java/org/springframework/util/function/SingletonSupplier.java
```diff
@@ -41,6 +42,15 @@
  */
 public class SingletonSupplier<T> implements Supplier<@Nullable T> {
 
+	private static final Object NULL = new Object() {
+
+		@Override
+		public String toString() {
```
quaff: It seems that overriding this method is not necessary, what's the purpose?

nosan: The reason that the method is overridden isn't for functionality but for debugging purposes.

Screenshot 2025-08-25 at 10 08 17 

When overridden:
Screenshot 2025-08-25 at 10 07 57

spring-jdbc/src/test/java/org/springframework/jdbc/support/SQLErrorCodeSQLExceptionTranslatorTests.java
```diff
@@ -209,6 +209,7 @@ void dataSourceInitialization() throws Exception {
 
 		reset(dataSource);
 		given(dataSource.getConnection()).willReturn(connection);
+		translator = new SQLErrorCodeSQLExceptionTranslator(dataSource);
```
ronodhirSoumik, line 212: related to this issue/change?

nosan: yes

spring-core/src/main/java/org/springframework/util/function/SingletonSupplier.java
```diff
@@ -94,26 +97,24 @@ private SingletonSupplier(@Nullable T singletonInstance) {
 	 */
 	@Override
 	public @Nullable T get() {
-		T instance = this.singletonInstance;
-		if (instance == null) {
+		if (!this.initialized) {
```
ronodhirSoumik, line 100: as `initialized` has no external use, is this variable really necessary? cant we directly use `this.singletonInstance == null` or add a function->
```
	public boolean isInitialized() {
		return Objects.nonNull(this.singletonInstance);
     }
```

nosan: Using `this.singletonInstance == null` can't tell the difference between "**not initialized yet**" and "**initialized but supplier returned null**." Since the supplier can return null, a separate flag needs to be used to record that initialization was attempted. Otherwise, the supplier will keep getting re-invoked.

See https://github.com/spring-projects/spring-framework/issues/35369  for more details.

sbrannen:
This has been merged into `main` in a41af448ec23b124d8f5a450c4b43c67a4674771 and revised in 2d3c2e3feba276d56c7968fd087bef42ef973416.

Thanks

nosan:
Thanks, @sbrannen!

PR #35397, opened 2025-08-31
Add tests for applyRelativePath method in StringUtils

J-unStiN:
Documenting Test Scenarios:

- Basic Joins
- Relative paths start with a slash
- Include the root path
- No slashes in the base path
- Preserve parent directory tokens (..)
- Base paths end with a slash
- Empty relative path
Scope of Impact: Changes only to test code. No impact to production code.


bclozel:
Thanks for your contribution @J-unStiN !

Issue #29950, opened 2023-02-09
Violation of a unique index on insert in MSSQL is translated to DataIntegrityViolationException instead of DuplicateKeyException


simon-barbey-mo:
**Affects:** 6.0.4 (spring-r2dbc)

When inserting a row in MSSQL, using `spring-r2dbc`, that results in a unique constraint violation, the thrown exception is `DataIntegrityViolationException`.

I would have expected a `DuplicateKeyException`, for its Javadoc documentation states:
> Exception thrown when an attempt to insert or update data results in violation of a primary key **or unique constraint**.

The SQL state and code returned by MSSQL are the following:
```
sqlState: 23000
errorCode: 2601
```

The method `ConnectionFactoryUtils.indicatesDuplicateKey()` does not seem to take into account the `2601` error code.

The exception thrown has the following stack trace:
```
org.springframework.dao.DataIntegrityViolationException:
    executeMany; SQL [INSERT INTO MY_TABLE (my_column) VALUES (@P0_mycolumn)]; 
    Cannot insert duplicate key row in object 'dbo.MY_TABLE' with unique index 'IDX_MY_TABLE_MY_COLUMN'.
    The duplicate key value is (-1).
	at org.springframework.r2dbc.connection.ConnectionFactoryUtils.convertR2dbcException(ConnectionFactoryUtils.java:237) ~[spring-r2dbc-6.0.4.jar:6.0.4]
```

Note that it also affects `spring-jdbc` when using `JdbcTemplate`, for the error translation done in `SQLStateSQLExceptionTranslator.indicatesDuplicateKey()` is pretty much the same.

This might also relates to the following issue: https://github.com/spring-projects/spring-framework/issues/29699

Issue #29699, opened 2022-12-16
Conflict on insert gets translated to DataIntegrityViolationException instead of DuplicateKeyException in Spring 6


lukas-krecan:
Followup from https://github.com/spring-projects/spring-framework/issues/29511

If there is primary key conflict on insert, Spring 6.0.3 returns `DataIntegrityViolationException` instead of `DuplicateKeyException` as it did in Spring 5. It would make sense to me to implement similar fix as you did for https://github.com/spring-projects/spring-framework/issues/29511. The SQL state seems to be 23000.

jhoeller:
I suppose this is coming from MySQL/MariaDb, MS SQL and Oracle, as per your comment on the other issue?

23000 is a general constraint violation error. I suspected that Oracle might report that SQL state already, with its old error code 1 attached. But for MySQL, I thought I saw it supporting 23505 but must have mis-read that. MariaDb inherited that legacy from MySQL, I suppose. All of them really should be reporting 23505 instead, there is an old enhancement request for MySQL for it that never got addressed.

Anyway, it would helpful to analyze the exact exception reported by the JDBC driver for those cases, ideally with exception class, SQL state, error code and exception message. If you have the chance to check those remaining databases for those exception details, that would be super helpful. We'll try to integrate corresponding checks then since those databases are unlikely to fix it on their end.

lukas-krecan:
Hi,

* MariaDB - state=23000 errorCode=1062 java.sql.SQLIntegrityConstraintViolationException: (conn=9) Duplicate entry 'my-lock' for key 'PRIMARY'
* MySQL  - state=23000 errorCode=1062 java.sql.SQLIntegrityConstraintViolationException: Duplicate entry 'my-lock' for key 'shedlock.PRIMARY'
* MsSQL - state=23000 errorCode=2627 com.microsoft.sqlserver.jdbc.SQLServerException: Violation of PRIMARY KEY constraint 'PK__shedlock__72E12F1A058F9D72'. Cannot insert duplicate key in object 'dbo.shedlock'. The duplicate key value is (my-lock).
* Oracle - state=23000 errorCode=1 java.sql.SQLIntegrityConstraintViolationException: ORA-00001: unique constraint (TEST.SYS_C008305) violated

Postgres, H2, HSQL, DB2 look good (return 23505)

jhoeller:
Thanks, @lukas-krecan! I've got a change ready that specifically looks at those error codes within SQL state 23000, along the following lines:

[...]

PR #35400, opened 2025-09-01
Detect Informix error codes as `DuplicateKeyException`

sgflt:
This fix is related to https://github.com/spring-projects/spring-framework/issues/29950 and https://github.com/spring-projects/spring-framework/issues/29699 .

After migration from Spring Boot 2 to Spring Boot 3 spring-jdbc changed behavior and now throws `DataIntegrityViolationException` instead of `DuplicateKeyException`.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLStateSQLExceptionTranslator.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLStateSQLExceptionTranslator.java
@@ -95,7 +95,8 @@ public class SQLStateSQLExceptionTranslator extends AbstractFallbackSQLException
 			301,   // SAP HANA
 			1062,  // MySQL/MariaDB
 			2601,  // MS SQL Server
-			2627   // MS SQL Server
+			2627,  // MS SQL Server
+			-268   // Informix
 		);
 
 
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLStateSQLExceptionTranslator.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/support/SQLStateSQLExceptionTranslator.java
@@ -96,6 +96,7 @@ public class SQLStateSQLExceptionTranslator extends AbstractFallbackSQLException
 			1062,  // MySQL/MariaDB
 			2601,  // MS SQL Server
 			2627,  // MS SQL Server
+			-239,  // Informix
 			-268   // Informix
 		);
 
--- a/spring-jdbc/src/test/java/org/springframework/jdbc/support/SQLStateSQLExceptionTranslatorTests.java
+++ b/spring-jdbc/src/test/java/org/springframework/jdbc/support/SQLStateSQLExceptionTranslatorTests.java
@@ -91,6 +91,16 @@ class SQLStateSQLExceptionTranslatorTests {
 	}
 
 	@Test
+	void translateDuplicateKeyInformix1() {
+		assertTranslation("23000", -239, DuplicateKeyException.class);
+	}
+
+	@Test
+	void translateDuplicateKeyInformix2() {
+		assertTranslation("23000", -268, DuplicateKeyException.class);
+	}
+
+	@Test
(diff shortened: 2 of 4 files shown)


sbrannen:
Hi @sgflt,

This is indeed related to #29950, especially this comment (https://github.com/spring-projects/spring-framework/issues/29950#issuecomment-1424180744).

In `sql-error-codes.xml`, we support the following for duplicate keys.

https://github.com/spring-projects/spring-framework/blob/942fbf30320a52f2f28e46961176e13cf8576a4d/spring-jdbc/src/main/resources/org/springframework/jdbc/support/sql-error-codes.xml#L154-L156

The first two are still [documented as error codes](https://www.ibm.com/docs/en/informix-servers/15.0.0?topic=informix-error-messages) for Informix 15.

```
-239	Could not insert new row - duplicate value in a UNIQUE INDEX
column.

The row that is being inserted (or being updated to have a new primary key)
contains a duplicate value of some row that already exists, in a column or
columns that are constrained to have unique values. Another cause of this
error might be a locking conflict if the table lock mode is page. The new
or updated row is not inserted.

Roll back the current transaction and execute it again without any
duplicate rows or with the locking conflict resolved.

If you are using repeatable read isolation, then the error could be due
to a unique constraint being violated. Refer to error -268.
```

```
-268	Unique constraint <constraint-name> violated.

During execution of this statement, a duplicate value was introduced
into a column or columns that a unique constraint protects. The row
with the duplicate value was not allowed into the table (not inserted
or not updated). For IBM Informix Dynamic Server, any changes that this statement
made prior to the discovery of the duplicate value have been rolled back. (The
effects of preceding statements in the transaction, if there were any, remain
in effect. They must be explicitly rolled back or committed.)

This error occurs in a logging database. However, if you are using
repeatable read isolation, error -239 will display instead.
```

Thus, if we were to make this change, I believe we should do it for `-239` and `-268`.

In addition, we would want to do this not only for `SQLStateSQLExceptionTranslator` (in `spring-jdbc`) but also for `ConnectionFactoryUtils` (in `spring-r2dbc`), and we would need appropriate tests/assertions in `SQLStateSQLExceptionTranslatorTests` and `ConnectionFactoryUtilsTests`, respectively.

We will discuss within the team whether we wish to make these changes to `SQLStateSQLExceptionTranslator` and `ConnectionFactoryUtils`, and we will let you know whether to proceed with this PR.

----

Note, however, that we would only implement this change for `6.2.x` and `7.0`.

For `6.0.x` and `6.1.x`, see the recommendations in [Upgrading From Spring Framework 5.3](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-6.0-Release-Notes#data-access-and-transactions).

> Spring's default JDBC exception translator is the JDBC 4 based SQLExceptionSubclassTranslator now, detecting JDBC driver subclasses as well as common SQL state indications (without database product name resolution at runtime). As of 6.0.3, this includes a common SQL state check for DuplicateKeyException, addressing a long-standing difference between SQL state mappings and legacy default error code mappings.
> 
> For full backwards compatibility with database-specific error codes, consider re-enabling the legacy SQLErrorCodeSQLExceptionTranslator. This translator kicks in for user-provided sql-error-codes.xml files. It can simply pick up Spring's legacy default error code mappings as well when triggered by an empty user-provided file in the root of the classpath.

sbrannen:
We have decided to proceed with this.

@sgflt, please update this PR as outlined below.

Thanks

> Thus, if we were to make this change, I believe we should do it for `-239` and `-268`.
> 
> In addition, we would want to do this not only for `SQLStateSQLExceptionTranslator` (in `spring-jdbc`) but also for `ConnectionFactoryUtils` (in `spring-r2dbc`), and we would need appropriate tests/assertions in `SQLStateSQLExceptionTranslatorTests` and `ConnectionFactoryUtilsTests`, respectively.

sbrannen:
This has been merged into `6.2.x` and `main` in db9e938ec452510e4c33680ce95f3de4228a4743 and slightly revised in 02f0f92a7296bbbbd8506fbdc07c076baaff78d2.

Thanks

PR #35405, opened 2025-09-02
Refine Task Javadoc about Runnable wrapping

### LINKED ISSUE #35394: Improve Task Javadoc about Runnable wrapping
As seen in https://github.com/spring-projects/spring-framework/issues/35320#issuecomment-3233859012, we should refine the `Task` Javadoc to better explain that this is not simply a holder to a `Runnable` and that extra wrapping of that runnable instance should be expected.

trollfaceiv:
Clarify that the runnable associated with a Task may be wrapped by the framework (for scheduling or observability). The runnable returned by getRunnable() is the one to execute, and it may differ from the original instance supplied by the user.

Closes #35394


bclozel:
Thanks for the proposal, but the issue is already assigned and I find the javadoc changes a bit too verbose, mentioning non-public types in the javadoc.

PR #35423, opened 2025-09-04
Avoid thread pinning in SseEmitter, ResponseBodyEmitter

### LINKED ISSUE #35422: Avoid pinning in SSEEmitter & ResponseBodyEmitter
https://github.com/spring-projects/spring-framework/issues/30996 fixed thread pinning issue. But only for SSEEmitter. Still there are several synchronized block in ResponseBodyEmitter.  

https://github.com/spring-projects/spring-framework/blob/e5b58effa3b74569ca6d3b091fa44f47bdc6ccea/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java#L183

https://github.com/spring-projects/spring-framework/blob/e5b58effa3b74569ca6d3b091fa44f47bdc6ccea/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java#L211

To prevent thread pinning when using virtual thread, it have to replaced with `ReentrantLock`.

Many of us cannot move to JDK 24 yet (which reduces synchronized pinning), so we need a framework-level mitigation.

acktsap:
Closes gh-35422

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java
@@ -21,6 +21,8 @@ import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 
 import org.jspecify.annotations.Nullable;
@@ -63,6 +65,7 @@ import org.springframework.util.ObjectUtils;
  * @author Rossen Stoyanchev
  * @author Juergen Hoeller
  * @author Brian Clozel
+ * @author Taeik Lim
  * @since 4.2
  */
 public class ResponseBodyEmitter {
@@ -86,6 +89,8 @@ public class ResponseBodyEmitter {
 
 	private final DefaultCallback completionCallback = new DefaultCallback();
 
+	/** Guards access to write operations on the response. */
+	protected final Lock writeLock = new ReentrantLock();
 
 	/**
 	 * Create a new ResponseBodyEmitter instance.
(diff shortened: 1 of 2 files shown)


spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java
```diff
@@ -41,16 +41,13 @@
...
  * @author Brian Clozel
+ * @author Taeik Lim
  * @since 4.2
  */
 public class SseEmitter extends ResponseBodyEmitter {
 
 	private static final MediaType TEXT_PLAIN = new MediaType("text", "plain", StandardCharsets.UTF_8);
 
-	/** Guards access to write operations on the response. */
-	private final Lock writeLock = new ReentrantLock();
```
acktsap, line 51: I just moved this field to the upper level & changed it to `protected`

bclozel:
@acktsap please run the build locally before submitting a PR: `./gradlew check`.
You can then `git push --force` to your branch to update this PR.

acktsap:
> @acktsap please run the build locally before submitting a PR: `./gradlew check`. You can then `git push --force` to your branch to update this PR.

Completed

bclozel:
Thanks for raising this @acktsap , this is now merged.

spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitter.java
```diff
@@ -180,22 +197,28 @@ public void send(Object object) throws IOException {
 	 * @throws IOException raised when an I/O error occurs
 	 * @throws java.lang.IllegalStateException wraps any other errors
 	 */
-	public synchronized void send(Object object, @Nullable MediaType mediaType) throws IOException {
+	public void send(Object object, @Nullable MediaType mediaType) throws IOException {
 		Assert.state(!this.complete, () -> "ResponseBodyEmitter has already completed" +
```
kilink, line 201: This is reading plain fields now with no synchronization

kilink: To clarify, I think it has introduced a bug, since the `complete` and `failure` fields aren't read while holding the lock.

Issue #35134, opened 2025-06-30
KotlinReflectionParameterNameDiscoverer doesn't support Kotlin Context Parameters


IRus:
## Background
Context Parameters were introduced in Kotlin 2.1.20 and are currently in Beta status in Kotlin 2.2.0 under the flag `-Xcontext-parameters` ([announcement](https://blog.jetbrains.com/kotlin/2025/04/update-on-context-parameters/))

## Issue
Spring's `KotlinReflectionParameterNameDiscoverer` cannot detect Context Parameters in Kotlin methods. The `kotlinFunction` property returns incomplete results - context parameters are completely absent from the reflection data.

## Use Case
I'm migrating from Context Receivers to Context Parameters for implementing security via `HandlerMethodArgumentResolver`. My implementation:
- Uses a custom `UserContext` class as a context parameter
- Injects it into controller handler methods
- Propagates it through service method calls

## Technical Details
- Current implementation relies on: [KotlinReflectionParameterNameDiscoverer.java#L46](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/KotlinReflectionParameterNameDiscoverer.java#L46)
- Related Kotlin issues:
  - [KT-78752: Context Parameters missing in reflection](https://youtrack.jetbrains.com/issue/KT-78752/Context-Parameters-missing-in-reflection) - I'm asking to fix `kotlinFunction`, hopefully in 2.1.22/2.2.10
  - [KT-74529: Context parameters support in reflection](https://youtrack.jetbrains.com/issue/KT-74529/Context-parameters-support-in-reflection) - Card about support of context parameters targeting 2.2.20

## Expected

1. Track support for Context Parameters in Spring Framework (if KT-78752 will be resolved)
2. Add support for the new parameter kind coming in Kotlin 2.2.20 when reflection API is available

sdeleuze:
Let's explore that when Kotlin reflection will be fixed.

IRus:
Reflection support landed in 2.2.20

sdeleuze:
I think I need a bit more time to fully understand all the use cases in Spring, we need the support in other places like like `CoroutinesUtils` (see #35428 related PR), and in Kotlin 2.2.20 it is still experimental (`Kind.CONTEXT` is annotated with `@ExperimentalContextParameters`) so think that's probably more reasonable to not rush it for 7.0 and plan it for 7.1.

[...]

PR #35428, opened 2025-09-05
Support Kotlin Context Parameters

SIMULATAN:
*Note: finding a commit message for this change is quite difficult, the current one is likely to cause confusion. I'd appreciate better suggestions*

Originally introduced as Beta in 2.2.0, Context Parameters are a new Kotlin feature that I wanted to use in my project.
Those parameters have a different `KParameter.Kind` (`CONTEXT`) and are therefore not supported by [the current CoroutinesUtils implementation](https://github.com/spring-projects/spring-framework/blob/d2bdf11b3973cc7173a52c33ae7bbae2002442bd/spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java). This resulted in the following error message:

```
java.lang.IllegalArgumentException: No argument provided for a required parameter: context parameter tenant of context(tenant: com.application.TenantContext)
```

## Status
~~[kotlin-reflect's Context Parameters support](https://github.com/JetBrains/kotlin/commit/38842eefbec957507162974806b3f48bcbb8d9a4) was first released in 2.2.20-Beta1. Therefore, I had to target the latest 2.2.20 RC with this PR.~~
~~Once 2.2.20 is formally released, I'll await the update on Spring's main branch, rebase onto it, and set the PR as ready to merge.~~
With 86fb62c05998f776e42b7960d9c039f00645ebb2, main is using the latest Kotlin release. I've done the rebase and marked the PR as ready to review. Thanks in advance for your considerations!

### DIFF
--- a/spring-core/spring-core.gradle
+++ b/spring-core/spring-core.gradle
@@ -130,6 +130,12 @@ jar {
 	}
 }
 
+kotlin {
+    compilerOptions {
+        freeCompilerArgs.addAll("-Xcontext-parameters")
+    }
+}
+
 test {
 	// Make sure the classes dir is used on the test classpath (required by ResourceTests).
 	// When test fixtures are involved, the JAR is used by default.
--- a/spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
@@ -130,7 +130,7 @@ public abstract class CoroutinesUtils {
 					for (KParameter parameter : function.getParameters()) {
 						switch (parameter.getKind()) {
 							case INSTANCE -> argMap.put(parameter, target);
-							case VALUE, EXTENSION_RECEIVER -> {
+							case VALUE, EXTENSION_RECEIVER, CONTEXT -> {
 								Object arg = args[index];
 								if (!(parameter.isOptional() && arg == null)) {
 									KType type = parameter.getType();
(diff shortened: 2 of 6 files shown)


sdeleuze:
Could you please add related test(s)?

SIMULATAN:
Added tests in a fixup commit, analogous to the existing Kotlin Extension Function tests. Had to force-push once as I messed up the GPG signing on my work device.

Let me know in case any further modifications are wanted.

sdeleuze:
As mention in #35134, I prefer not rushing a support in 7.0, so I have planned merging this PR for 7.1, along with the `KotlinReflectionParameterNameDiscoverer` changes, documentation, etc.

@SIMULATAN Could you please share a reproducer for an application using this feature? I would like to fully understand the various use cases of Kotlin context parameters with Spring, so that would help to have for example a controller with suspending functions using this feature, in addition to the tests you provided.

SIMULATAN:
> I prefer not rushing a support in 7.0, so I have planned merging this PR for 7.1

Totally fair. Perhaps more people will see and try out this PR, hopefully providing additional feedback in the process.

> Could you please share a reproducer for an application using this feature? I would like to fully understand the various use cases of Kotlin context parameters with Spring, so that would help to have for example a controller with suspending functions using this feature, in addition to the tests you provided.

I've implemented a simple Todo-like application showcasing our use case: multitenancy.
Unfortunately, there's still some nasty glue involved. Thankfully, it is only necessary on the two ends - the controller, using a `CoWebFilter`, and Hibernate, by using a `CurrentTenantIdentifierResolver` backed by a `ThreadLocal` - the most dangerous part of the entire application.
Nevertheless, the risk of losing values through thread context switches and other tomfoolery in our business logic is greatly reduced and, quite frankly, it just feels way more reliable and explicit.

Here's the repo: https://github.com/SIMULATAN/spring-context-parameter-demo
Although my creativity was limited, it should function as a comprehensible PoC regardless.

PR #35430, opened 2025-09-05
Docs: Clarify default prefix/suffix of `registry.jsp()`

### LINKED ISSUE #35429: Docs: Clarify default prefix/suffix of registry.jsp()
Hello,

The current Spring MVC documentation for View Resolvers shows the following example:
```java
@Override
public void configureViewResolvers(ViewResolverRegistry registry) {
    registry.enableContentNegotiation(new MappingJackson2JsonView());
    registry.jsp();
}
```
However, the documentation does not explain what defaults are applied by registry.jsp().

### Problem

- Readers may assume they must always configure a prefix and suffix manually.

- In reality, registry.jsp() registers an InternalResourceViewResolver with sensible defaults.

- The absence of this detail can cause confusion when JSP views are not found.

### Suggested Improvement

Add a short [NOTE] block right after the JSP example clarifying:

- Default prefix: /WEB-INF/

- Default suffix: .jsp

- These defaults are applied automatically, and explicit configuration is only needed if you use a different location or extension.

### Benefits

- Prevents confusion about JSP view resolution.

- Makes the framework’s default behavior explicit.

- Helps users quickly diagnose view resolution issues without unnecessary configuration.

I’d appreciate your thoughts on this suggestion!

Dongnyoung:
Hello, 

This PR clarifies the Spring MVC documentation for *View Resolvers*.  
The current example shows `registry.jsp();` but does not explain what defaults are applied.

In this file: https://github.com/spring-projects/spring-framework/blob/main/framework-docs/modules/ROOT/pages/web/webmvc/mvc-config/view-resolvers.adoc

## Problem
- Readers may assume they must always configure a prefix and suffix manually.  
- In reality, `registry.jsp()` registers an `InternalResourceViewResolver` with sensible defaults.  
- The absence of this detail can cause confusion when JSP views are not resolved.

## Changes Introduced
- Added a `[NOTE]` block immediately after the JSP example.  
- The note clarifies that:
  - Default prefix: `/WEB-INF/`
  - Default suffix: `.jsp`
  - Explicit configuration is only required if JSP files are located in a different directory or use a different extension.  
- The `application.properties` example was intentionally omitted here, since it is more relevant to Spring Boot documentation.

Closes #35429

Thank you for reviewing.


bclozel:
I think this is already covered in the Javadocs: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/ViewResolverRegistry.html#jsp()

Dongnyoung:
I understand this is already covered in the Javadoc, thanks for pointing it out.

Would it still make sense to keep a short [NOTE] or at least a cross-reference in the reference docs?  
While the defaults can indeed be found in the Javadoc, a brief mention in the documentation could make it easier for readers to quickly grasp the behavior without having to jump into the API docs.

Happy to adjust the PR accordingly if you think it adds value.

sdeleuze:
@Dongnyoung I don't think this is the right place for documenting such details, but I would suggest to rework your PR as an update of https://docs.spring.io/spring-framework/reference/web/webmvc-view/mvc-jsp.html to replace the XML bean declaration by a more high level JavaConfig `configureViewResolvers(ViewResolverRegistry registry)` code snippet displayed by default and a complementary `<mvc:view-resolvers>` XML one with a related note.

Dongnyoung:
Thanks a lot for the valuable feedback 🙏  
I’ve reworked this change according to your suggestions (JavaConfig example as the default, XML `<mvc:view-resolvers>` as a note, and no duplication of defaults).  

Opened a new PR at #35444  with the updated version.

PR #35444, opened 2025-09-08
Docs: Rework mvc-jsp section with JavaConfig example (per review feedback)

Dongnyoung:
Supersedes #35430
### Summary
This PR reworks the [mvc-jsp](https://docs.spring.io/spring-framework/reference/web/webmvc-view/mvc-jsp.html) section following the review feedback:

- Replace the XML `<bean>` declaration with a higher-level JavaConfig `configureViewResolvers(ViewResolverRegistry registry)` snippet as the default example.
- Add a complementary `<mvc:view-resolvers>` XML example within a note for legacy configuration.
- Remove duplication of default prefix/suffix values, instead pointing to the Javadoc.

### Motivation
As suggested in the review, this section is not the right place to document low-level defaults.  
By focusing on configuration style and best practices, while still covering XML for legacy users, the documentation stays concise and aligned with Spring’s reference doc guidelines.


Dongnyoung:
The maintainer moved this into #35450 and closed it there.  
Thanks for handling this!

PR #35448, opened 2025-09-09
Add @Nullable annotation to AbstractJsonPathAssertions.isEqualTo

### LINKED ISSUE #35445: Missing @Nullable on JsonPathAssertions.isEqualTo
Hello.

Unless I am very confused, the parameter expectedValue of JsonPathAssertions.isEqualTo should be Nullable, but it is not. If it is not supposed to be, then a isNull assertion would be helpful.

Full story on reddit https://www.reddit.com/r/Kotlin/comments/1nbnmrh/tell_kotlin_a_java_library_does_accept_null_values/ , when what I first thought was a problem with Kotlin turned out to be a missing Nullable in spring-test-6.2.10. To sum it up :

`webTestClient.get()

    // ...

    .jsonPath("$.phone_number").isEqualTo(testUserData.account.phoneNumber)`

in a kotlin project will prompt Intellij to send the warning "Java type mismatch: inferred type is 'String?', but 'Any' was expected." because phoneNumber can be null. The helpfull people on Reddit helped me figure it out : Kotlin thinks isEqualTo cannot accept null because the package is annotated with NonNullApi, but isEqualTo is not annotated with Nullable, even though opening isEqualTo show null is handled.

rlaope:
fix https://github.com/spring-projects/spring-framework/issues/35445

This PR adds the @Nullable annotation to the parameter of 
AbstractJsonPathAssertions.isEqualTo(Object expectedValue).

Also adds a test to verify that passing null works as expected.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/support/AbstractJsonPathAssertions.java
+++ b/spring-test/src/main/java/org/springframework/test/web/support/AbstractJsonPathAssertions.java
@@ -58,7 +58,7 @@ public abstract class AbstractJsonPathAssertions<B> {
 	/**
 	 * Applies {@link JsonPathExpectationsHelper#assertValue(String, Object)}.
 	 */
-	public B isEqualTo(Object expectedValue) {
+	public B isEqualTo(@Nullable Object expectedValue) {
 		this.pathHelper.assertValue(this.content, expectedValue);
 		return this.bodySpec;
 	}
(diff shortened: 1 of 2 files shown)


sdeleuze:
This PR is hardly usable as not against the branch where we are asked to fix the issue originally, multiple draft commits not squashed, etc. We appreciate contributions but maybe for your next potential try, please take the time to polish it before sending it, and for straightforward changes where we mention in the issue that we will take care, not needed to send a PR that will take more time for us to integrate than directly do the fix. Better to choose non assigned issues that we don't have the time to process. As a consequence, I will close this PR unmerged and reopen the issue.

PR #35458, opened 2025-09-10
Add MediaType.IMAGE_WEBP and IMAGE_WEBP_VALUE

Dongnyoung:
Hello, 
### Motivation
`image/webp` is a widely adopted media type across browsers, CDNs, and mobile platforms.  
Currently, developers often hardcode `"image/webp"` in `produces` declarations.  
Adding `MediaType.IMAGE_WEBP` and `IMAGE_WEBP_VALUE` improves consistency with other image media types and avoids string literals.

### Changes
- Added `MediaType.IMAGE_WEBP` and `IMAGE_WEBP_VALUE` in `spring-web`.
- Javadoc aligned with existing constants (e.g. `IMAGE_PNG_VALUE`).
- Added `MediaTypeTests#imageWebpConstants` for verification.

### Compatibility
- Additive change only.
- No behavior changes, no removals, fully backwards compatible.

Thanks for the review!


bclozel:
Thanks for the suggestion, but these days we're trying to keep the number of values low here - even if the media type is well known, we tend to add values only if Spring uses it for building a specific feature or support.

See #29613 #26087 #25491 #2055 #1589 #1054 and more

PR #35465, opened 2025-09-11
useCaches option in PathMatchingResourcePatternResolver not applied in special case

reckart:
Set useCaches on single resources obtained directly from the resource loader before returning it from the PathMatchingResourcePatternResolver

Targets main, but should actually go to 6.2.x IMHO.


reckart:
Hm, this might actually not build since Resource doesn't seem to expose `setUseCaches()`... Sorry, I tried using a codespace for this and code completion there seems to be guessing more than knowing.

reckart:
I think this should do it now.

jhoeller:
Good catch!

In terms of addressing this, we can also simply call our local `getResource` method which does the `UrlResource` handling on top of `ResourceLoader` access already. I've resolved this issue with a corresponding commit on `6.2.x`.

PR #35491, opened 2025-09-17
Document placeholder and Ant-style pattern support for `@ComponentScan`

Rockernun:
## Overview

`@ComponentScan(basePackages)` already supports:

- Resolving ${…} property placeholders against the Spring Environment.
- Ant-style package patterns (e.g., com.example.**, org.acme.*).

However, this behavior wasn’t explicitly called out in the JavaDoc or the reference docs, making it easy to miss. This PR documents those capabilities and clarifies where pattern support applies.

JavaDoc: clarify that basePackages/value resolve ${…} via the Environment and support Ant-style package patterns (e.g., com.example.**); note patterns do not apply to basePackageClasses.

Reference docs: add a short subsection in classpath-scanning.adoc with Java/Kotlin + properties examples.

## Related Issues

- #35288

### DIFF
--- a/framework-docs/modules/ROOT/pages/core/beans/classpath-scanning.adoc
+++ b/framework-docs/modules/ROOT/pages/core/beans/classpath-scanning.adoc
@@ -323,6 +323,45 @@ sure that they are 'opened' (that is, that they use an `opens` declaration inste
 `exports` declaration in your `module-info` descriptor).
 ====
 
+==== Property placeholders and Ant-style patterns
+
+`@ComponentScan(basePackages)` supports `${…}` property placeholders resolved
+against the `Environment` and Ant-style package patterns such as `com.example.**`.
+Multiple packages and/or patterns may be specified.
+
+[tabs]
+======
+Java::
++
+[source,java,indent=0,subs="verbatim,quotes"]
+----
+@Configuration
+@ComponentScan(basePackages = "${app.scan.packages}")
+public class AppConfig {
+	// ...
+}
+----
+
+Kotlin::
++
+[source,kotlin,indent=0,subs="verbatim,quotes"]
+----
+@Configuration
+@ComponentScan(basePackages = ["\${app.scan.packages}"])
+class AppConfig {
+	// ...
+}
+----
+======
+
+[source,properties,indent=0,subs="verbatim,quotes"]
+----
+app.scan.packages=com.example.**,org.acme.*
+----
+
+NOTE: Ant-style patterns do not apply to `basePackageClasses`, which accepts concrete
+classes and derives packages from those classes.
+
 Furthermore, the `AutowiredAnnotationBeanPostProcessor` and
 `CommonAnnotationBeanPostProcessor` are both implicitly included when you use the
 component-scan element. That means that the two components are autodetected and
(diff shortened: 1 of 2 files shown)


framework-docs/modules/ROOT/pages/core/beans/classpath-scanning.adoc
```diff
@@ -323,6 +323,45 @@ sure that they are 'opened' (that is, that they use an `opens` declaration inste
 `exports` declaration in your `module-info` descriptor).
 ====
 
+==== Property placeholders and Ant-style patterns
```
sbrannen, line 326: Please note that this breaks the build.

```
[16:49:49.179] WARN (asciidoctor): section title out of sequence: expected level 2, got level 3
    file: /XXX/spring-framework/framework-docs/modules/ROOT/pages/core/beans/classpath-scanning.adoc:335
```

However, I have addressed this locally, thus there is no need to update this PR.

framework-docs/modules/ROOT/pages/core/beans/classpath-scanning.adoc
```diff
@@ -323,6 +323,45 @@ sure that they are 'opened' (that is, that they use an `opens` declaration inste
...
+======
+
+[source,properties,indent=0,subs="verbatim,quotes"]
+----
+app.scan.packages=com.example.**,org.acme.*
+----
+
+NOTE: Ant-style patterns do not apply to `basePackageClasses`, which accepts concrete
+classes and derives packages from those classes.
+
```
sbrannen, line 364: FYI: the following paragraph and note belong to the previous section about XML configuration.

However, I have addressed this locally, thus there is no need to update this PR.

spring-context/src/main/java/org/springframework/context/annotation/ComponentScan.java
```diff
@@ -86,6 +90,13 @@
 	 * attribute.
 	 * <p>Use {@link #basePackageClasses} for a type-safe alternative to
 	 * String-based package names.
+	 * <p>Supports {@code ${...}} placeholders resolved against the
+	 * {@link org.springframework.core.env.Environment Environment} as well as
+	 * Ant-style package patterns (for example, {@code com.example.**}).
+	 * Multiple packages and/or patterns may be specified.
+	 * <p><strong>Note:</strong> Ant-style patterns are <em>not</em> applicable to
+	 * {@link #basePackageClasses()}, which accepts concrete classes for type-safe
+	 * package selection.
```
sbrannen, line 99: I don't think there is any need to point this out, since one can only supply `Class` references via the `basePackageClasses` attribute.

Thus, I have removed this and other such text locally.

sbrannen (approved):
I left a few notes, but I am approving this PR as-is, since I have made the necessary changes locally on my machine.

framework-docs/modules/ROOT/pages/core/beans/classpath-scanning.adoc
sbrannen, line 326: For future reference, please note that we do our best to introduce custom anchors that align with other anchors used in the same file.

For example, I've added the following that I will apply after merging this PR.

```asciidoc
[[beans-scanning-placeholders-and-patterns]]
```

sbrannen:
This has been merged into `main` in fe04bfcadb5d68a66c74f7975a7ee01aeccd0fc9 and revised in dbb9bf939c47405e5cd7089a23a8181ba7438d62

Thanks! 👍

PR #35501, opened 2025-09-18
Adapt to Reactor 3.8 JSpecify annotations

chemicL:
reactor-core 3.8.0-SNAPSHOT includes the migration from JSR-305-based nullability annotations to JSpecify.
This change adapts the spring-framework codebase to the changes caused by better nullability markers.

@sdeleuze please review and if something needs to change on the reactor-core side, I'd be happy to incorporate those recommendations.

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
@@ -1170,7 +1171,7 @@ public abstract class CacheAspectSupport extends AbstractCacheInvoker
 						if (invokeFailure.get()) {
 							return Mono.error(ex);
 						}
-						return (Mono) invokeOperation(invoker);
+						return (Mono) Objects.requireNonNull(invokeOperation(invoker));
 					}
 					catch (RuntimeException exception) {
 						return Mono.error(exception);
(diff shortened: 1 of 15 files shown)


spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
```diff
@@ -1201,8 +1202,8 @@ private class ReactiveCachingHandler {
 				}
 				if (adapter.isMultiValue()) {
 					return adapter.fromPublisher(Flux.from(Mono.fromFuture(cachedFuture))
-							.switchIfEmpty(Flux.defer(() -> (Flux) evaluate(null, invoker, method, contexts)))
-							.flatMap(v -> evaluate(valueToFlux(v, contexts), invoker, method, contexts))
+							.switchIfEmpty(Flux.defer(() -> (Flux) Objects.requireNonNull(evaluate(null, invoker, method, contexts))))
```
chemicL, line 1205: For code on the hot path, in reactor-core, I used assertions to silence the warnings where I knew nullable API is not returning a null value. I wasn't sure what strategy you'd like to take so I left a TODO somewhere near `requireNonNull` to address this.

sdeleuze (changes requested):
I think some of those `Objects#requireNonNull` invocation are related to https://github.com/uber/NullAway/issues/1290 so I would recommend to revise the PR by using `@SupressWarnings("NullAway") // see https://github.com/uber/NullAway/issues/1290` instead for such lambdas.

sdeleuze:
We need to analyze each usage where there is a nullability mismatch, but it looks correct from a reactor-core POV, so you can follow-up with reactor-core 3.8.0-RC1 release and we will take this PR from there. Thanks for submitting it.

spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
sdeleuze: That's fine, we may refine potentially later.

PR #35510, opened 2025-09-19
Enhance toString method in SockJsFrame

polyglot-k:
## **Description:**

* Removed usage of external libraries (`StringUtils`) and `Math.min()`.
* Minimized intermediate `String` objects to reduce GC pressure.
* Used `StringBuilder` with pre-allocated capacity for better memory efficiency.
* Implemented manual handling of `\n` and `\r` replacements and truncated output after 80 characters.
* Ensured the method works efficiently for both short and long content strings.
* This change improves performance for high-frequency logging without affecting readability for standard logs.

**Key Changes:**

```java
@Override
public String toString() {
  int maxLength = 80;
  int contentLength = this.content.length();
  int truncatedContentLength  = Math.min(contentLength, maxLength);
  
  int extra = (contentLength > maxLength ? truncatedSuffix.length() : 0);
  
  StringBuilder sb = new StringBuilder(truncatedContentLength  + extra);
  
  for (int i = 0; i < truncatedContentLength; i++) {
	  char c = this.content.charAt(i);
	  switch (c) {
		  case '\n' -> sb.append("\\n");
		  case '\r' -> sb.append("\\r");
		  default -> sb.append(c);
	  }
  }
  
  if (contentLength > maxLength) {
	  sb.append(truncatedSuffix);
  }
  
  return "SockJsFrame content='" + sb + "'";
}
```

**Benefits:**

* Reduced memory allocation and GC overhead.
* Avoids dependency on external libraries.
* Efficient handling for large strings.
* Preserves readability for logs.

### DIFF
--- a/spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
+++ b/spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
@@ -16,6 +16,7 @@
 
 package org.springframework.web.socket.sockjs.frame;
 
+
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import org.jspecify.annotations.Nullable;
@@ -80,6 +81,30 @@ public class SockJsFrame {
 		}
 	}
 
+	public static SockJsFrame openFrame() {
+		return OPEN_FRAME;
+	}
+
+	public static SockJsFrame heartbeatFrame() {
+		return HEARTBEAT_FRAME;
+	}
+
+	public static SockJsFrame messageFrame(SockJsMessageCodec codec, String... messages) {
+		String encoded = codec.encode(messages);
+		return new SockJsFrame(encoded);
+	}
+
+	public static SockJsFrame closeFrameGoAway() {
+		return CLOSE_GO_AWAY_FRAME;
+	}
+
+	public static SockJsFrame closeFrameAnotherConnectionOpen() {
+		return CLOSE_ANOTHER_CONNECTION_OPEN_FRAME;
+	}
+
+	public static SockJsFrame closeFrame(int code, @Nullable String reason) {
+		return new SockJsFrame("c[" + code + ",\"" + (reason != null ? reason : "") + "\"]");
+	}
 
 	/**
 	 * Return the SockJS frame type.
@@ -116,7 +141,6 @@ public class SockJsFrame {
 		}
 	}
 
-
 	@Override
 	public boolean equals(@Nullable Object other) {
 		return (this == other || (other instanceof SockJsFrame that &&
@@ -149,32 +173,6 @@ public class SockJsFrame {
 		}
 
 		return "SockJsFrame content='" + sb + "'";
-	}
-
-
-	public static SockJsFrame openFrame() {
-		return OPEN_FRAME;
-	}
-
-	public static SockJsFrame heartbeatFrame() {
-		return HEARTBEAT_FRAME;
-	}
-
-	public static SockJsFrame messageFrame(SockJsMessageCodec codec, String... messages) {
-		String encoded = codec.encode(messages);
-		return new SockJsFrame(encoded);
-	}
-
-	public static SockJsFrame closeFrameGoAway() {
-		return CLOSE_GO_AWAY_FRAME;
-	}
-
-	public static SockJsFrame closeFrameAnotherConnectionOpen() {
-		return CLOSE_ANOTHER_CONNECTION_OPEN_FRAME;
-	}
-
-	public static SockJsFrame closeFrame(int code, @Nullable String reason) {
-		return new SockJsFrame("c[" + code + ",\"" + (reason != null ? reason : "") + "\"]");
 	}
 
 }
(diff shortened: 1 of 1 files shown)


spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
```diff
@@ -133,15 +130,27 @@ public int hashCode() {
...
 	@Override
 	public String toString() {
-		String result = this.content;
-		if (result.length() > 80) {
-			result = result.substring(0, 80) + "...(truncated)";
+		int maxLen = 80;
+		int contentLength = this.content.length();
+		int len = Math.min(contentLength, maxLen);
+
+		StringBuilder sb = new StringBuilder(len + 20);
```
rstoyanchev: Why +20? The truncated suffix is 14, and it depends on whether there is truncation. Couldn't this be more precise (if truncation expected + 14, or otherwise just `length`)?

spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
```diff
@@ -133,15 +130,27 @@ public int hashCode() {
 
 	@Override
 	public String toString() {
-		String result = this.content;
-		if (result.length() > 80) {
-			result = result.substring(0, 80) + "...(truncated)";
+		int maxLen = 80;
```
rstoyanchev: For consistency with the codebase, this should be maxLength (we generally don't abbreviate).

rstoyanchev (commented):
Thanks for the pull request.

The title says "for better content representation", but it looks like the output doesn't change, or does it? In other words it's about performance and streamlined implementation?

spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
polyglot-k: Yes, I'll change it to "maxLength"

polyglot-k:
@rstoyanchev 
Actually, the result is not changing. However, we changed it because using the "String" object directly is not good for GC.

spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
polyglot-k: You're right, the `+20` is just a rough safety margin to avoid potential `StringBuilder` resizing, not an exact calculation. 
It was intended to cover the truncated suffix and a few extra characters like quotes and delimiters. 
But I agree it could be made more precise by checking if truncation occurs and only adding the actual suffix length. 
I'll update the code to reflect that for better clarity.


How should I improve it?
@rstoyanchev

spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
```diff
@@ -133,15 +130,27 @@ public int hashCode() {
...
-		result = StringUtils.replace(result, "\n", "\\n");
-		result = StringUtils.replace(result, "\r", "\\r");
-		return "SockJsFrame content='" + result + "'";
-	}
 
+		if (contentLength > maxLen) {
+			sb.append("...(truncated)");
+		}
+
+		return "SockJsFrame content='" + sb + "'";
```
bclozel: Why not use the String Builder for this part as well? This PR is trying to optimize for memory allocation after all.

polyglot-k:
@rstoyanchev 
I'm done reflecting the content

bclozel (changes requested):
I think the overall changes is making readability worse.
Instead of actively trying to guess the builder size in advance with multiple operations, I would favor using a default `StringBuilder` instance and use it all the way (instead of performing concatenations on the last line).

This would cause probably some more allocation, but in the end this code is not on a hot path so treating `StringUtils` as an "external library" and making maintenance worse is not worth it.

spring-websocket/src/main/java/org/springframework/web/socket/sockjs/frame/SockJsFrame.java
polyglot-k: Thank you for the review.

Regarding the other `toString()` methods, they also use `StringBuilder`, so I thought there’s no need for the unnecessary String objects to become GC targets in the JVM. Of course, I agree that the original code is more readable.

My perspective is that the readability of this part isn’t particularly high, and concatenating immutable Strings in this way is generally considered an anti-pattern, so I opted to use a mutable `StringBuilder`.

I will revise the `return` part as you suggested.

@bclozel

rstoyanchev:
In the end I've gone for the simpler version of a fixed initialCapacity. The use of StringBuilder is good as it eliminates re-creating the String potentially multiple times. Given that SockJS frames typically have a `\n` that guarantees at least one replace and potentially more. The use of StringBuilder ensures the String will never be re-created more than once.

polyglot-k:
@rstoyanchev thank u :)

PR #35514, opened 2025-09-19
Nullability fixes for AtomicReference

msridhar:
Make the contents of some `AtomicReference` variables `@Nullable`.  This is in preparation for uber/NullAway#1298 which causes new errors to be reported for these variables.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/ReactorClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/ReactorClientHttpRequest.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.net.URI;
 import java.time.Duration;
+import java.util.Objects;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -131,13 +132,13 @@ final class ReactorClientHttpRequest extends AbstractStreamingClientHttpRequest 
 			return Mono.empty();
 		}
 
-		AtomicReference<Executor> executorRef = new AtomicReference<>();
+		AtomicReference<@Nullable Executor> executorRef = new AtomicReference<>();
 
 		return outbound
 				.withConnection(connection -> executorRef.set(connection.channel().eventLoop()))
 				.send(FlowAdapters.toPublisher(new OutputStreamPublisher<>(
 						os -> body.writeTo(StreamUtils.nonClosing(os)), new ByteBufMapper(outbound),
-						executorRef.getAndSet(null), null)));
+						Objects.requireNonNull(executorRef.getAndSet(null)), null)));
 	}
 
 	static IOException convertException(RuntimeException ex) {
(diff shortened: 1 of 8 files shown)


msridhar:
FYI @sdeleuze

spring-web/src/main/java/org/springframework/http/client/ReactorClientHttpRequest.java
```diff
@@ -131,13 +132,13 @@ private Publisher<Void> send(
...
 
-		AtomicReference<Executor> executorRef = new AtomicReference<>();
+		AtomicReference<@Nullable Executor> executorRef = new AtomicReference<>();
 
 		return outbound
 				.withConnection(connection -> executorRef.set(connection.channel().eventLoop()))
 				.send(FlowAdapters.toPublisher(new OutputStreamPublisher<>(
 						os -> body.writeTo(StreamUtils.nonClosing(os)), new ByteBufMapper(outbound),
-						executorRef.getAndSet(null), null)));
+						Objects.requireNonNull(executorRef.getAndSet(null)), null)));
```
msridhar, line 141: This is the only logic change in this PR (all other changes are just adding `@Nullable` annotations). I'm pretty confident in the change based on the surrounding code but I'm not an expert so it should be reviewed

sdeleuze: LGTM, but I would welcome a confirmation from @jhoeller and/or @rstoyanchev.

sdeleuze (approved):

spring-web/src/main/java/org/springframework/http/client/ReactorClientHttpRequest.java
sdeleuze: After discussing with Juergen and Rossen, looks fine but we would like @violetagg confirmation.

violetagg: There is already  assert in `OutputStreamPublisher`, so do we need this here or it is to suppress the warnings?
basically we will check twice.

https://github.com/spring-projects/spring-framework/blob/20aac6dd8d0819f953e71a89ebc2be31e1955726/spring-web/src/main/java/org/springframework/http/client/OutputStreamPublisher.java#L76

sdeleuze: Good point. That's not a hot code path but conceptually this double check is annoying.

@msridhar Would declaring a `AtomicReference<@NullUnmarked Executor> executorRef ` field be acceptable here with your upcoming changes to not trigger warnings without having to do the double check?

msridhar: `@NullUnmarked` cannot be written on a type argument (only methods and classes).  Seems like the assertion in `OutputStreamPublisher` is a defense against unchecked code, since the parameter is already non-null?  In any case, the only other alternative I can think of here is `@SuppressWarnings("NullAway")`, which unfortunately would have to be on the entire method.  If it were me I would keep the duplicate null check for a more localized suppression, but I'm happy to modify to whatever the maintainers think is best.

sdeleuze:
Merged, thanks @msridhar!

Issue #35511, opened 2025-09-19
Reinstantiate checks for `kotlin-reflect`


odrotbohm:
`BeanUtils.instantiateClass(Constructor, Object…)` is not checking for the presence of Kotlin Reflect before calling `KotlinDelegate.instantiateClass(…)` anymore. In case of the `kotlin-reflect` missing from the classpath, this will lead to the following exception:

```
java.lang.NoClassDefFoundError: kotlin/reflect/jvm/ReflectJvmMapping
    at org.springframework.beans.BeanUtils$KotlinDelegate.instantiateClass(BeanUtils.java:909)
```

While it might not be common for the reflection artifact to be missing if folks are using Kotlin, one might run into this if they use a library depending on `kotlin-stdlib` only. Currently, one has to add `kotlin-reflect` to avoid the exception. It would be cool if the code path was just skipped if the artifact is missing.

wonyongg:
Hi,

I created a PR for this issue since the change seemed fairly simple. If I misunderstood anything or this isn’t appropriate, please feel free to close it. Thank you for your time and consideration.🙏🏼

sdeleuze:
@wonyongg Thanks for creating the PR, but I will have to cover much more other usages by reverting some of the changes done by https://github.com/spring-projects/spring-framework/commit/1763334180ece33aca8ff9d5a456563cb2eaffe3, so I will handle it myself via this issue.

PR #35515, opened 2025-09-20
Fix NoClassDefFoundError when kotlin-reflect is missing from classpath

wonyongg:
This pull request addresses the issue described in #35511

## Changes
Added `KotlinDetector.isKotlinReflectPresent()` checks before using `KotlinDelegate`:

1. **`BeanUtils.instantiateClass(Constructor, Object...)`**: Now checks both `isKotlinType()` and `isKotlinReflectPresent()` before delegating to `KotlinDelegate`
2. **`BeanUtils.findPrimaryConstructor(Class)`**: Same check added for consistency

When `kotlin-reflect` is not available, the methods fall back to regular Java instantiation instead of throwing `NoClassDefFoundError`.

## Testing

- All existing tests pass
- Added new test cases covering both Java and Kotlin scenarios
- Verified that existing Kotlin functionality still works when kotlin-reflect is present

## Testing without kotlin-reflect
To simulate an environment without kotlin-reflect, I added the following configuration locally:
```md
configurations {
    testWithoutKotlinReflect {
        extendsFrom testRuntimeClasspath
        exclude group: "org.jetbrains.kotlin", module: "kotlin-reflect"
    }
}
```

Using this setup, I confirmed that the code behaves correctly and falls back without errors.

This configuration was used for local verification only and is not part of the commit. If integrating this scenario into the test suite is desirable, I would appreciate your guidance, as I’m not sure how to approach it.

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
+++ b/spring-beans/src/main/java/org/springframework/beans/BeanUtils.java
@@ -69,6 +69,7 @@ import org.springframework.util.StringUtils;
  * @author Rob Harrop
  * @author Sam Brannen
  * @author Sebastien Deleuze
+ * @author Wonyong Hwang
  */
 public abstract class BeanUtils {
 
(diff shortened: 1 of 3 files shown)


sdeleuze:
Thanks for this PR but I will have to cover much more other usages by reverting some of the changes done by https://github.com/spring-projects/spring-framework/commit/1763334180ece33aca8ff9d5a456563cb2eaffe3, so I will handle it myself via the issue.

PR #35533, opened 2025-09-23
Fix string array decode

Prabal864:
### Context / Problem

In Spring Framework 6 / WebFlux, using `WebClient` to decode a JSON array of strings 
(e.g., ["hello", "world"]) into a `List<String>` sometimes resulted in unexpected behavior: 
the JSON array was decoded as a **single String** containing the raw JSON ("[\"hello\",\"world\"]") 
instead of a proper `List<String>`.

This issue also occurs for empty arrays: decoding "[]" into a `List<String>` could produce 
a list containing a single string "[]" rather than an empty list.

This behavior is confusing and can lead to subtle bugs when consuming JSON arrays 
from reactive endpoints.

---

### Solution

This PR adds **self-contained reproducer tests** to the `spring-webflux` module 
demonstrating the correct way to decode JSON arrays into `List<String>` using 
`ParameterizedTypeReference`.  

Key points:

1. **Embedded Netty server for tests**  
   - The tests spin up a lightweight embedded Netty server on a dynamic port.
   - This ensures the tests are **self-contained** and do not rely on external services.

2. **Correct JSON decoding**  
   - `bodyToMono(new ParameterizedTypeReference<List<String>>() {})` correctly converts 
     the JSON array into a `List<String>`.

3. **Covers both empty and non-empty arrays**  
   - `decodeNotEmptyList()` → decodes `["hello", "world"]` into a list of two strings.  
   - `decodeEmptyList()` → decodes `[]` into an empty list.

4. **Assertions**  
   - Uses AssertJ `assertThat()` to verify the decoded lists exactly match expected values.

---

### Benefits

- Ensures `WebClient` correctly decodes JSON arrays into collections.  
- Provides a reproducible test case for this edge case in WebFlux.  
- Improves reliability for developers consuming JSON arrays in reactive applications.  
- Fully self-contained; can run in CI without external dependencies.

---

### Test Instructions

1. Run the tests locally:

   ```bash
   ./gradlew :spring-webflux:test --tests "org.springframework.web.reactive.function.client.StringArrayDecodeTests"

### DIFF
--- a/spring-webflux/src/test/java/org/springframework/web/reactive/function/client/StringArrayDecodeTests.java
+++ b/spring-webflux/src/test/java/org/springframework/web/reactive/function/client/StringArrayDecodeTests.java
@@ -1,0 +1,44 @@
+package org.springframework.web.reactive.function.client;
+
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.Test;
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.http.MediaType;
+import reactor.core.publisher.Mono;
+import reactor.netty.http.server.HttpServer;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class StringArrayDecodeTests {
+
+	@Test
+	void decodeNotEmptyList() {
+		HttpServer server = HttpServer.create()
+				.port(0)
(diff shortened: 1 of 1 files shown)


rstoyanchev:
This is expected behavior documented at https://docs.spring.io/spring-framework/reference/web/webflux/reactive-spring.html#webflux-codecs-jackson (note at the end of section).

PR #35535, opened 2025-09-24
Add `serialVersionUID` in `LinkedCaseInsensitiveMap`

quaff:
It's common to cache result of `JdbcTemplate::queryForMap` which is type of `LinkedCaseInsensitiveMap`, and Spring Cache Abstraction use Java serialization by default which rely on `serialVersionUID`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
+++ b/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
@@ -44,11 +44,13 @@ import org.jspecify.annotations.Nullable;
  *
  * @author Juergen Hoeller
  * @author Phillip Webb
+ * @author Yanming Zhou
  * @since 3.0
  * @param <V> the value type
  */
-@SuppressWarnings("serial")
 public class LinkedCaseInsensitiveMap<V> implements Map<String, V>, Serializable, Cloneable {
+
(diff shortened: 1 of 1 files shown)


spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
```diff
@@ -44,12 +44,14 @@
  *
  * @author Juergen Hoeller
  * @author Phillip Webb
+ * @author Yanming Zhou
```
sbrannen, line 47: If you look at the [history](https://github.com/spring-projects/spring-framework/commits/9cfe835a45ef2353b4f7e53e081b03e3fef3598e/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java), you'll see that there have actually been several people who have modified that file over the years with more significant changes.

Thus, I don't think adding a generated serialization version UID warrants being listed as an author, and I'll remove that when merging.

quaff: It's OK.

sbrannen:
This has been merged into `main` in 871076ef97d6f4f8b71a63ad9e053b2bd8ffe43d and revised in 5df082132dbb2881d600e002d5dfe4ba8f31a9a1.

Thanks

PR #35540, opened 2025-09-25
Fix: Null Pointer Exception when Using RestTemplate with MinimalHttpClient from HttpComponents

xxx-tea:
When using the RestTemplate, the executeOpen method of the link to the HttpComponentsClientHttpRequest is called, and the null HttpHost is passed in, resulting in subsequent errors. Use RoutingSupport.determineHost to solve


rstoyanchev:
Closing as per discussion under #35539.

PR #35546, opened 2025-09-25
refactor(messaging): remove redundant else blocks after early return

barrackkk:
This PR simplifies control flow by removing redundant `else` blocks that follow early `return` statements in `org.springframework.messaging.core`.

Files:
- AbstractMessageReceivingTemplate.receiveAndConvert
- GenericMessagingTemplate.headerToLong

Built locally with :spring-messaging:check.


bclozel:
Thanks but we don't usually accept cosmetic changes. The code reads well like it is.

PR #35567, opened 2025-10-02
Make StringUtils::pathEquals Null-Safe

sairohith2605:
This PR introduces a tiny patch for the method `StringUtils.pathEquals()` that short-circuits into a `false` result if any or both of the parameters of path are `null`. This aims to prevent a `NullPointerException` when `cleanPath()` returns `null` when the supplied path String does not have any length.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StringUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/StringUtils.java
@@ -825,7 +825,7 @@ public abstract class StringUtils {
 	 * @return whether the two paths are equivalent after normalization
 	 */
 	public static boolean pathEquals(String path1, String path2) {
-		return (path1 == null || path2 == null) ? false : cleanPath(path1).equals(cleanPath(path2));
+		return (path1 != null && path2 != null) && cleanPath(path1).equals(cleanPath(path2));
 	}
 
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @sairohith2605,

Unfortunately, I need to decline the proposal for the following reasons.

1. Supplying a `null` path to `StringUtils.pathEquals(...)` is not a supported use case.
2. The `org.springframework.util` package (see `package-info.java`) is annotated with `@NullMarked`.
3. Consequently, `null` is not a supported argument value unless a given parameter is annotated with `@Nullable`.
4. Neither `path1` nor `path2` is annotated with `@Nullable`.

It is therefore the responsibility of the caller to ensure that `pathEquals(...)` is not invoked with a `null` argument.

Note that we rely on the NullAway build plugin for Gradle as well as IDE support to ensure that we do not supply a `null` value for one of those arguments. For example, in `WebUtils` you can see that we verify that both arguments are not `null`.

https://github.com/spring-projects/spring-framework/blob/8a719117e308b5fe5928a91b2bc63b7ed3681e62/spring-web/src/main/java/org/springframework/web/util/WebUtils.java#L241-L252

Thanks anyway

PR #35570, opened 2025-10-05
Remove unused application startup call in AnnotationConfigApplicationContext

Wars:
AbstractApplicationContext.applicationStartup can only be injected after the instance construction. This metrics point has no practical effect at present. I think it can be removed.


bclozel:
Thanks @Wars this has been merged in main.

PR #35587, opened 2025-10-08
Retain order of `produces` media types in `@ExceptionHandler`

IMurzich:
Fix error when corrupt order of `produces` media types for `@ExceptionHandler`.

For example, the following returns xml by default.

```java
@ExceptionHandler(exception = Exception.class,
    produces = {"application/json", "application/xml"})
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/annotation/ExceptionHandlerMethodResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/method/annotation/ExceptionHandlerMethodResolver.java
@@ -21,7 +21,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @IMurzich,

However, before we process this PR, we need to take care of the following.

1. Please explain the error scenario you are attempting to fix.
2. Please introduce a unit or integration test which fails before the fix and passes after the fix.

Regarding point # 1, I realize you added some text to this issue's description, but it is not fully clear to me what you mean. So, perhaps you can just expound on that.

Thanks

IMurzich:
Hi @sbrannen,

I've added test. Please check it.

Thanks.

IMurzich:
Purpose of change:

I have a controller and exception handler in it.

My REST service can return JSON or XML replies.

Therefore I want to return exceptions in JSON or XML too.

Exception handler is annotated with:

```java
@ExceptionHandler(exception = Exception.class, produces = {"application/json", "application/xml"})
```

And I expect to get JSON response with error (`accept=*/*`), but I receive XML response because `HashSet` doesn't retain the order of the `produces` media types and returns `application/xml` as the first media type.

sbrannen:
@IMurzich, can you please sign off your commits so that the [Developer Certificate of Origin](https://github.com/apps/dco) (DCO) check passes?

Thanks

bclozel:
@IMurzich please sign the DCO. You can do that with by issuing the following commands on your fork:

```
git amend -s
git push --force
```

sbrannen (approved):

sbrannen:
This has been merged into `6.2.x` and `main` in 7699b4af9c11737391efde8465ae8cc80632ea87 and slightly revised in 8456cd1e74d8670f1620c10a10d9f8b77a0e9458.

Thanks for your contribution! 👍

PR #35588, opened 2025-10-08
MockMvc request builders should expose HttpHeaders

mieseprem:
Does it fix #35576 ?

If that's the case: here you go — enjoy

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/request/AbstractMockHttpServletRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/request/AbstractMockHttpServletRequestBuilder.java
@@ -31,6 +31,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.function.Consumer;
 
 import jakarta.servlet.ServletContext;
 import jakarta.servlet.ServletRequest;
@@ -109,7 +110,7 @@ public abstract class AbstractMockHttpServletRequestBuilder<B extends AbstractMo
 
 	private @Nullable String contentType;
 
-	private final MultiValueMap<String, Object> headers = new LinkedMultiValueMap<>();
+	private final HttpHeaders headers = new HttpHeaders();
 
 	private final MultiValueMap<String, String> parameters = new LinkedMultiValueMap<>();
 
@@ -342,7 +343,7 @@ public abstract class AbstractMockHttpServletRequestBuilder<B extends AbstractMo
 	 * @param values one or more header values
 	 */
 	public B header(String name, Object... values) {
-		this.headers.addAll(name, Arrays.asList(values));
+		this.headers.addAll(name, Arrays.stream(values).map(String::valueOf).toList());
 		return self();
 	}
 
@@ -352,6 +353,20 @@ public abstract class AbstractMockHttpServletRequestBuilder<B extends AbstractMo
 	 */
 	public B headers(HttpHeaders httpHeaders) {
 		httpHeaders.forEach(this.headers::addAll);
+		return self();
+	}
+
+	/**
+	 * Manipulate this builder's headers with the given consumer. The
+	 * headers provided to the consumer are "live", so that the consumer can be used to
+	 * {@linkplain HttpHeaders#set(String, String) overwrite} existing header values,
+	 * {@linkplain HttpHeaders#remove(String) remove} values, or use any of the other
+	 * {@link HttpHeaders} methods.
+	 * @param headersConsumer a function that consumes the {@code HttpHeaders}
+	 * @return this builder
+	 */
+	public B headers(Consumer<HttpHeaders> headersConsumer){
+		headersConsumer.accept(this.headers);
 		return self();
 	}
 
@@ -665,12 +680,11 @@ public abstract class AbstractMockHttpServletRequestBuilder<B extends AbstractMo
 			this.contentType = parentBuilder.contentType;
 		}
 
-		for (Map.Entry<String, List<Object>> entry : parentBuilder.headers.entrySet()) {
-			String headerName = entry.getKey();
-			if (!this.headers.containsKey(headerName)) {
-				this.headers.put(headerName, entry.getValue());
+		parentBuilder.headers.forEach((name, values) -> {
+			if (!this.headers.containsHeader(name)) {
+				this.headers.put(name, values);
 			}
-		}
+		});
 		for (Map.Entry<String, List<String>> entry : parentBuilder.parameters.entrySet()) {
 			String paramName = entry.getKey();
 			if (!this.parameters.containsKey(paramName)) {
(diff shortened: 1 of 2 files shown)


spring-test/src/main/java/org/springframework/test/web/servlet/request/AbstractMockHttpServletRequestBuilder.java
```diff
@@ -342,7 +343,7 @@ public B accept(String... mediaTypes) {
 	 * @param values one or more header values
 	 */
 	public B header(String name, Object... values) {
-		this.headers.addAll(name, Arrays.asList(values));
+		this.headers.addAll(name, Arrays.stream(values).map(String::valueOf).toList());
```
rstoyanchev: The `MultiValueMap<String, Object>` matches to how `MockHttpServletRequest` manages header values, allowing Collections, arrays, as well as date and number values, which it then formats when headers are obtained. So this change is unlikely to go without causing regressions.

The code in this method could match the behavior of `getDateHeader` and `getIntHeader` of `MockHttpServletRequest`, in reverse by formatting to String values in the same way, to ensure backwards compatibility.

rstoyanchev (changes requested):

spring-test/src/main/java/org/springframework/test/web/servlet/request/AbstractMockHttpServletRequestBuilder.java
mieseprem: To avoid heading in the wrong direction, I’ll briefly confirm what I’ve understood:

Working internally with `HttpHeaders` isn’t the best idea because of potential regressions. It’s better to continue using `MultiValueMap` internally.

So, if we still want to offer a method `headers(Consumer<HttpHeaders>)`, we should consume the `HttpHeaders` and then call the existing `headers(HttpHeaders)` method. I can change it in that direction.

sbrannen: > So, if we still want to offer a method `headers(Consumer<HttpHeaders>)`, we should consume the `HttpHeaders` and then call the existing `headers(HttpHeaders)` method.

That unfortunately does not work.

If you create a new instance of `HttpHeaders` and provide that to the `Consumer<HttpHeaders>`, the consumer can only add new headers. It cannot remove or process existing headers.

mieseprem: > If you create a new instance of `HttpHeaders` and provide that to the `Consumer<HttpHeaders>`, the consumer can only add new headers. It cannot remove or process existing headers

Ah, I think I’ve got it now. Yes — my approach indeed won’t work.

The challenge is "extracting" the `HttpHeaders` from the Consumer without first converting the existing `MultiValueMap` to `HttpHeaders` and then converting back after consumption. That seems cumbersome. I’ll think about possible solutions.

mieseprem: Do you think there’s any sensible way to implement this? As long as the internal headers map is a `MultiValueMap<String, Object>`, I can’t see how to make it compatible with the `MultiValueMap<String, String>` used by `HttpHeaders`. I can’t simply change the value type of objects already placed into the builder’s headers map into their string representations — that would almost certainly cause regressions.

For a moment I considered an implementation along these lines, but that was based on the incorrect assumption that the Consumer would only contain `HttpHeaders` instructions for purely String‑based header values (to be clear, this code produces the wrong result):
```java
public B headers(Consumer<HttpHeaders> headersConsumer){
	HttpHeaders intermediateHeaders = new HttpHeaders();
	
	var stringHeaders = this.headers.entrySet().stream().
		filter(entry -> entry.getValue().stream().allMatch(obj -> obj instanceof String))
		.peek(entry -> intermediateHeaders.addAll(entry.getKey(), entry.getValue().stream().map(String.class::cast).toList()))
		.map(Map.Entry::getKey)
		.toList();
	stringHeaders.forEach(this.headers::remove);

	headersConsumer.accept(intermediateHeaders);
	return headers(intermediateHeaders);
}
```

If I'm not (again) totally wrong, this should be a test that need to pass:
```java
@Test
void headersConsumer() {
	this.builder.header("X-Foo-String", "bar");
	this.builder.header("X-Foo-Date", LocalDate.now());
	this.builder.header("X-Foo-List-Int", List.of(1, 2, 3));

	this.builder.headers(httpHeaders -> {
		httpHeaders.put("X-Baz", Arrays.asList("qux", "quux"));
		httpHeaders.remove("X-Foo-Date");
	});

	MockHttpServletRequest request = this.builder.buildRequest(this.servletContext);
	List<String> headerNames = Collections.list(request.getHeaderNames());

	assertThat(headerNames).containsExactly("X-Foo-String", "X-Foo-List-Int", "X-Baz");
}
```
I think for this to work at all, roughly the following would need to happen:
* create from each `MultiValueMap<String, Object>` entry a `MultiValueMap<String, String>` one (this step alone seems very complex — I’m thinking of your point about `getDateHeader(String)` and `getIntHeader(String)`)
* it’s easy to create an `HttpHeaders` object from a `MultiValueMap<String, String>` (call that X)
* consume the Consumer with the created object X
* afterwards, add, remove or modify entries in the original headers map (we must not copy back untouched entries that we converted to String in X)

Either I’m overthinking this, I’ve gone completely off track, or I’m perhaps not the right person to implement this task :shrug:

rstoyanchev: Sorry if this wasn't clear, but my suggestion was not to change to `MultiValueMap`. Keep `HttpHeaders` internally, as you have done, and improve the way values are formatted. So instead of a simple `String#valueOf`, we would need to match the formatting that would be done otherwise in `MockHttpServletRequest`. 

For this you can trace the code to add and obtain headers in `MockHttpServletRequest`. You will see that `doAddHeaderValue` recognizes collections and arrays. Also `getDateHeader` and `getIntHeader` recognize various additional value types.

mieseprem: Okay, got it. So I will give it another try 👍

spring-test/src/main/java/org/springframework/test/web/servlet/request/AbstractMockHttpServletRequestBuilder.java
```diff
@@ -741,6 +765,57 @@ private boolean containsCookie(Cookie cookie) {
...
+	 * @return the object's {@code toString()} value by default; if {@code o} is a {@link TemporalAccessor},
+	 *         it is formatted to an RFC7231-compliant string.
+	 */
+	private static String objectToString(Object o) {
+		Assert.notNull(o, "'o' must not be null");
+		if (o instanceof TemporalAccessor ta) {
+			return temporalToString(ta);
+		} else {
+			return o.toString();
+		}
```
mieseprem, line 781: `o` could itself be a Collection or Array -> need to handle (see CollectionUtils)

rstoyanchev:
@mieseprem, this issue has proven to be more involved than it appeared, and there are also further changes I'd like to make to align the MockMvc request builder with that of RestTestClient and WebTestClient. At the same time we're running out of time for RC1, so I'm going to take over this request. Thanks in any case for the effort to help!

PR #35599, opened 2025-10-10
Optimize repeated annotation stream processing in cache annotation parsing

kamilkrzywanski:
Optimize repeated annotation stream processing in cache annotation parsing
Closes: #35598

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/annotation/SpringCacheAnnotationParser.java
+++ b/spring-context/src/main/java/org/springframework/cache/annotation/SpringCacheAnnotationParser.java
@@ -44,6 +44,7 @@ import org.springframework.util.StringUtils;
  * @author Phillip Webb
  * @author Stephane Nicoll
  * @author Sam Brannen
+ * @author Kamil Krzywański
  * @since 3.1
  */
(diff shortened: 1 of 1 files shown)


jhoeller:
After seeing the PR build failure (see `AnnotationCacheOperationSourceTests.multipleAnnotation`), this reminded me why we do those individual streams: Those annotations need to be processed in a specific order of annotation types for the resulting operations to semantically align, not in the order of annotation declaration. 

A simple loop over the declared annotations would have to collect the annotations into different categories first (probably one collection per category to hold), and then once again iterate over each category to actually process those annotations (resulting in a sequence of operations). Not sure whether we can make the resulting code look nicer than the current stream-based code...

kamilkrzywanski:
@jhoeller 
You’re absolutely right — in this case, any attempt to “optimize” the iteration isn’t meaningful.
Since the operations must be processed in a strict order by annotation type, we’d either have to sort the collected results afterward or buffer them into four separate lists before processing. Both approaches would introduce extra overhead and likely make the code less readable than the current stream-based implementation.

So I agree — it’s best to keep the existing approach as-is.

Issue #35497, opened 2025-09-18
Add support for recording and reusing the outcome of classpath scanning


snicoll:
We want Spring AOT to be able to replace runtime classpath scanning by an already-prepared index. This is similar to what the component indexer meant to be achieved, but at the application level.

To help with that, we need two new hook points on `ClassPathScanningCandidateComponentProvider`:

* A way to mark interest in what an instance does. That could be a callback whenever scanning occurs (and the outcome) for instance. This is used by Spring AOT to record and then make the index available
* A way to configure `ClassPathScanningCandidateComponentProvider` so that it uses an index, rather than doing the scanning at runtime.

PR #35601, opened 2025-10-11
Fix non-pattern package prefix check in `CandidateComponentsIndex`

NeatGuyCoding:
the non-pattern package prefix check uses packageName.startsWith(basePackage) which incorrectly treats "com.example2.foo" as a subpackage of "com.example"


sbrannen:
Since this is a follow-up to #35497 (which has not yet been released), I've labeled this as as `task` instead of `bug`.

sbrannen:
This has been merged into `main` in 113b9d131c3f2c0f4b652f30bd5930d3184ac5b5, and I added missing tests in b727dbb8022e8980c09908ac5a4b04a51e276609.

Thanks

PR #35611, opened 2025-10-12
#33856 emit warning when thread pool task scheduler is unable to meet task delay

### LINKED ISSUE #33856: Emit warning when ThreadPoolTaskScheduler is unable to meet task delay
It seems like a pretty common gotcha for applications with more than one type of scheduled task to run into issues when those tasks are long running and the default value for `spring.task.scheduling.pool.size` is `1`.

Virtual threads in Spring Boot obviously helps mitigate this, but for other uses if `ThreadPoolTaskScheduler` had a dedicated thread to peek the queue and determine if the next `RunnableScheduledFuture` delay had expired and log a warning, that would make it far easier to diagnose when thread starvation is preventing tasks from running on schedule.

alexeyleping:
# Add delay monitoring to ThreadPoolTaskScheduler

  Fixes #33856

  ## Problem

  `ThreadPoolTaskScheduler` with default pool size of 1 silently suffers from thread starvation when multiple tasks are scheduled, leading to production issues that are hard to diagnose.

  ## Solution

  Added automatic delay monitoring that:
  - Detects when tasks are delayed due to pool exhaustion
  - Logs actionable warnings with diagnostic metrics
  - Includes circuit breaker for self-protection
  - Exposes JMX management interface for runtime control

  ## Key Features

  -  Lock-free design (CAS operations, atomic variables)
  -  Zero contention with scheduler threads
  -  Configurable thresholds and intervals
  -  Structured logging support (JSON)
  -  CPU self-monitoring
  -  Fully backward compatible

  ## Example Warning

  WARN o.s.s.c.ThreadPoolTaskScheduler - 5 scheduled tasks are delayed
  (max delay: 1488ms) due to thread pool exhaustion. Pool size: 1,
  Active threads: 1, Queue size: 5. Consider increasing the pool size
  via ThreadPoolTaskScheduler.setPoolSize() or
  spring.task.scheduling.pool.size property.

  ## Testing

  Verified on production-like workload with demo application. All edge cases covered including concurrent modifications, circuit breaker transitions, and error recovery.

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/concurrent/ThreadPoolTaskScheduler.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/concurrent/ThreadPoolTaskScheduler.java
@@ -19,11 +19,13 @@ package org.springframework.scheduling.concurrent;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Delayed;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.RejectedExecutionHandler;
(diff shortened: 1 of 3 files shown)


bclozel:
Thanks for the contribution, but I think that this proposal goes way overboard. The initial request is about a simple WARN logging message - this change 2k LOC, new public API, new concepts and a complete JMX monitoring stack.
If we ever get to this enhancement request, we would rather opt in for a minimal approach and a simple WARN log.

PR #35612, opened 2025-10-12
Introduce `shouldFilter` flag in `OncePerRequestFilter`

sic2:
The introduction of `shouldFilter` allows for intentional filtering.

Usage example:
```java
public class BookingsFilter extends OncePerRequestFilter {

    @Override
    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response,
				FilterChain filterChain) { ... }

    @Override
    protected boolean shouldFilter(HttpServletRequest request) throws ServletException {
        return request.getRequestURI().startsWith("/booking");
    }
}
```

Closes gh-35605

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/filter/OncePerRequestFilter.java
+++ b/spring-web/src/main/java/org/springframework/web/filter/OncePerRequestFilter.java
@@ -121,7 +121,8 @@ public abstract class OncePerRequestFilter extends GenericFilterBean {
 				// Remove the "already filtered" request attribute for this request.
 				request.removeAttribute(alreadyFilteredAttributeName);
 			}
-		} else {
+		}
+		else {
 			// Proceed without invoking this filter...
 			filterChain.doFilter(request, response);
 		}
--- a/spring-web/src/main/java/org/springframework/web/filter/OncePerRequestFilter.java
+++ b/spring-web/src/main/java/org/springframework/web/filter/OncePerRequestFilter.java
@@ -121,6 +121,9 @@ public abstract class OncePerRequestFilter extends GenericFilterBean {
 				// Remove the "already filtered" request attribute for this request.
 				request.removeAttribute(alreadyFilteredAttributeName);
 			}
+		} else {
+			// Proceed without invoking this filter...
+			filterChain.doFilter(request, response);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Filters are already mapped to a specific URL or set of URLs, and it makes sense to narrow that down with an exclude list. So in the example you show, the `Filter` could be mapped to `"/booking/**"`, and it would only be invoked for that subpath. 

Could you expand on what you're trying to achieve, including how the `Filter` needs to be mapped?

sic2:
> Filters are already mapped to a specific URL or set of URLs, and it makes sense to narrow that down with an exclude list. So in the example you show, the `Filter` could be mapped to "/booking/**" and it would only be invoked for that subpath.
> 
> Could you expand on what you're trying to achieve, including how the Filter needs to be mapped?

@rstoyanchev I take it you are referring to the use of `AbstractFilterRegistrationBean` (and implementations) from spring-boot for the mapping of URLs via `setUrlPatterns`? 
Looking at it, I see what you mean. This is a miss from me.

In my mind, `shouldNotFilter` function in the `OncePerRequestFilter` is missing a counter part at the same level of abstraction, but I could be interpreting this all wrong.

### Examples

The way I see it, we would still be mapping the filter using `FilterRegistrationBean` or (similar), but this would be complemented by logic from the filters themselves.

#### Via URI (as in tests)

```java
@Override
protected boolean shouldFilter(HttpServletRequest request) {
	return request.getRequestURI().startsWith("/do_not_skip");
}
```

#### Via other bits of request (e.g. method)

```java
@Override
protected boolean shouldFilter(HttpServletRequest request) {
	return "POST".equals(request.getMethod());
}
```

### Moving forward

However, I get that there is already a way to perform filtering, even if it is by URL patterns only.

Feel free to close this PR if you think this doesn't fit your architectural design.

rstoyanchev:
Thanks for the extra comments. For mapping Filter's, I was referring to the Servlet API, see for example [FilterRegistration](https://javadoc.io/doc/jakarta.servlet/jakarta.servlet-api/latest/jakarta.servlet/jakarta/servlet/FilterRegistration.html), but yes in Boot this is exposed as a FilterRegistrationBean.

I will close this for now since OncePerRequestFilter has quite a few options already, and the idea generally is, map the Filter in the Servlet container, and then narrow via exclusions. This should work well even for HTTP methods as there are very few, so you can filter them either as exclusions or as inclusions. 

We can always reconsider for concrete uses cases.

PR #35619, opened 2025-10-13
Normalize JAR entry paths to strip BOOT-INF/classes prefix in Spring-Boot fat JARs

ashishranacse:
- Fixes #35617


jhoeller:
Superseded as per my comment on the original issue: https://github.com/spring-projects/spring-framework/issues/35617#issuecomment-3400916031

PR #35627, opened 2025-10-13
Consider defaultCandidate for scoped proxies

hosea:
PR for https://github.com/spring-projects/spring-framework/issues/35626

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/scope/ScopedProxyUtils.java
+++ b/spring-aop/src/main/java/org/springframework/aop/scope/ScopedProxyUtils.java
@@ -83,11 +83,15 @@ public abstract class ScopedProxyUtils {
 		proxyDefinition.setPrimary(targetDefinition.isPrimary());
 		if (targetDefinition instanceof AbstractBeanDefinition abd) {
 			proxyDefinition.copyQualifiersFrom(abd);
+			proxyDefinition.setDefaultCandidate(abd.isDefaultCandidate());
 		}
 
 		// The target bean should be ignored in favor of the scoped proxy.
 		targetDefinition.setAutowireCandidate(false);
 		targetDefinition.setPrimary(false);
+		if (targetDefinition instanceof AbstractBeanDefinition abd) {
+			abd.setDefaultCandidate(false);
+		}
 
 		// Register the target bean as separate bean in the factory.
 		registry.registerBeanDefinition(targetBeanName, targetDefinition);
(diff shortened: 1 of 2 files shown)


spring-beans/src/main/java/org/springframework/beans/factory/config/BeanDefinition.java
```diff
@@ -177,6 +177,21 @@ public interface BeanDefinition extends AttributeAccessor, BeanMetadataElement {
...
+	default void setDefaultCandidate(boolean defaultCandidate) {}
+
+	/**
+	 * Return wheter this bean is a candidate for getting autowired into some other bean based on
+	 * the plain type, without any further indications such as a qualifier match.
+	 */
+	default boolean isDefaultCandidate() {
+		return true;
+	}
+
```
hosea: This lines could have an impact. Declaration as "default" should be maximal transparent, but this is a central interface. Not necessarily required, could also be solved another way (see my other comments). But I think, this is straight and clear.

spring-aop/src/main/java/org/springframework/aop/scope/ScopedProxyUtils.java
```diff
@@ -81,13 +81,15 @@ public static BeanDefinitionHolder createScopedProxy(BeanDefinitionHolder defini
 		// Copy autowire settings from original bean definition.
 		proxyDefinition.setAutowireCandidate(targetDefinition.isAutowireCandidate());
 		proxyDefinition.setPrimary(targetDefinition.isPrimary());
+		proxyDefinition.setDefaultCandidate(targetDefinition.isDefaultCandidate());
```
hosea: If the extension fo the BeanDefinition-Interface is not acceptable, then setting the "defaultCandidate-Property" could also take place by checking/casting as in line 85 below:
proxyDefinition.setDefaultCandidate(abd.isDefaultCandidate());

spring-aop/src/main/java/org/springframework/aop/scope/ScopedProxyUtils.java
```diff
@@ -81,13 +81,15 @@ public static BeanDefinitionHolder createScopedProxy(BeanDefinitionHolder defini
...
 		if (targetDefinition instanceof AbstractBeanDefinition abd) {
 			proxyDefinition.copyQualifiersFrom(abd);
 		}
 
 		// The target bean should be ignored in favor of the scoped proxy.
 		targetDefinition.setAutowireCandidate(false);
 		targetDefinition.setPrimary(false);
+		targetDefinition.setDefaultCandidate(false);
```
hosea: Same with this line:
Alternative: add after line 86/87: abd.setDefaultCandidate(false)

Issue #35628, opened 2025-10-13
URIComponentsBuilder should not clear query parameters in method designed for adding


hosea:
I observed that UriComponentsBuilder#query unexpectedly clears all existing query parameters, when passing "null" as argument. In this case the method behaves like UriComponentsBuilder#replaceQuery that clearly should clear all query parameters when "null" is passed.

The semantic of UriComponentsBuilder#query is unclear: intentionally this method is build for adding query parameters, not for clearing / removing. But it can also be used for clearing all parameters as UriComponentsBuilder#replaceQuery when passing "null".

The path inside the method that executes "queryParams.clear()" should be a noop in my opinion.

PR added as proposal: https://github.com/spring-projects/spring-framework/pull/35629

jhoeller:
Worth noting: In the `UriBuilder` interface, `query(String)` does not declare a nullable argument. It's just the `UriComponentsBuilder` implementation of the method that declares it nullable. In contrast to `path(String)` , for example, which is non-null in both places.

In practice, `query(String)` should never be called with a `null` argument to begin with. If actually called with `null`, I guess it could simply ignore it indeed, like `path(String)` does - however, there might be occasional existing usage that accidentally relies on clearing.

In any case: If we remove the clearing there, we should also officially redeclare `UriComponentsBuilder.query(String)` for a non-null argument. Or the other way round, make `query(String)` officially accept a nullable argument in `UriBuilder`. Some overall alignment seems worth doing for 7.0.

jhoeller:
@hosea so from a backwards compatibility perspective, we can only really revise this in 7.0. Feel free to update your PR rebasing it to `main`, and removing the `@Nullable` from the `UriComponentsBuilder.query(String)` argument while you're at it.

Within the `query(String)` method body, we should check the given argument through `if (StringUtils.hasText(query))` rather than just against `null`, for defensiveness and for agreement with the `path(String)` implementation. That would give us clean nullability declarations but still defensive processing for null/empty arguments at runtime.

rstoyanchev:
Superseded by #35629.

PR #35629, opened 2025-10-13
UriComponentsBuilder#query should not clear queryParams if null passed in

hosea:
… which is intended for adding values

Issue: https://github.com/spring-projects/spring-framework/issues/35628


rstoyanchev:
@hosea, thanks for reporting the issue and sending a pull request. As we are due to release tomorrow, I will handle it from here, taking into account https://github.com/spring-projects/spring-framework/issues/35628#issuecomment-3405128001.

PR #35653, opened 2025-10-16
Fix `@Import` with multiple bean registrars

### LINKED ISSUE #35652: @Import multiple BeanRegistrar does not work
When using `@Import` to import multiple BeanRegistrars, not all beans are registered

when using an ApplicationContextInitializer to call `applicationContext.register(...)` it works

Reproducer: https://github.com/wakingrufus/spring-repro

wakingrufus:
fix ConfigurationClassParser so that `@Import`ed BeanRegistrars use their own class name rather than the class they are imported to

This allows multiple BeanRegistrars to be imported from the same Configuration class

fixes #35652

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java
@@ -615,7 +615,7 @@ class ConfigurationClassParser {
 						if (registrar instanceof ImportAware importAware) {
 							importAware.setImportMetadata(metadata);
 						}
-						configClass.addBeanRegistrar(metadata.getClassName(), registrar);
+						configClass.addBeanRegistrar(candidateClass.getName(), registrar);
 					}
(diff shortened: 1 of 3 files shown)


sdeleuze:
@wakingrufus I kept your test from the PR but fixed the bug [in a different way](https://github.com/spring-projects/spring-framework/commit/5fd3c9c449e3eea72baff29b9c85413f4b72d9ad), as we want to keep exposing the configuration class metadata, not the registrar one.

Thanks for catching this pretty serious bug before the GA!

wakingrufus:
> @wakingrufus I kept your test from the PR but fixed the bug [in a different way](https://github.com/spring-projects/spring-framework/commit/5fd3c9c449e3eea72baff29b9c85413f4b72d9ad), as we want to keep exposing the configuration class metadata, not the registrar one.
> 
> Thanks for catching this pretty serious bug before the GA!

Thanks for the fix!

Issue #35624, opened 2025-10-13
Add a method to `RestTestClient` API that provides a `JsonContentAssert`


donalmurtagh:
The `RestTestClient` API provides a [.json(String expected)](https://docs.spring.io/spring-framework/reference/7.0/testing/resttestclient.html#resttestclient-json) method that allows the entire JSON response to be verified via [JSONAssert](https://jsonassert.skyscreamer.org/) e.g.

```java
restTestClient.get().uri("/persons/1")
            .exchange()
            .expectStatus().isOk()
            .expectBody()
            .json("""
                {
                  "data": [
                    {"id": "123", "name": "John Doe",  "email": "john@example.com"}
                  ]
                }
            """)
```

This approach is fine when verifying simple JSON responses, but it becomes unwieldy for more complex JSON responses and/or when we want to compare the JSON properties with Java object properties.

By contrast, the `MockMvcTester` API provides a `bodyJson()` method that returns a `org.springframework.test.json.JsonContentAssert` object that supports more fine-grained assertions, e.g.

```java
mockMvcTester.get().uri("/persons/1")
            .assertThat()
            .hasStatusOk()
            .bodyJson()
            .extractingPath("$.data")
            .asArray().singleElement()
            .hasFieldOrPropertyWithValue("id", expectedUser.getId())
            .hasFieldOrPropertyWithValue("name", expectedUser.getName())
            .hasFieldOrPropertyWithValue("email", expectedUser.getEmail())
```

## Proposal
Add a method to`RestTestClient.BodyContentSpec` that allows the response to be consumed with a `JsonContentAssert`, e.g.

```java
restTestClient.get().uri("/persons/1")
            .exchange()
            .expectStatus().isOk()
            .expectBody() 
            .consumeWithJsonAssert((JsonContentAssert jsonContentAssert) -> {
                jsonContentAssert.extractingPath("$.data")
                        .asArray().singleElement()
                        .hasFieldOrPropertyWithValue("id", expectedUser.getId())
                        .hasFieldOrPropertyWithValue("name", expectedUser.getName())
                        .hasFieldOrPropertyWithValue("email", expectedUser.getEmail())
            })
```

## Current Workaround
In the absence of this method, I create the `JsonContentAssert` as below:

```java
restTestClient.get().uri("/persons/1")
    .exchange()
    .expectStatus().isOk()
    .expectBody()
    .consumeWith(response -> {
[...]

PR #35657, opened 2025-10-17
Add a method to `RestTestClient` API that provides a `JsonContentAssert`

donalmurtagh:
See issue #35624


vpelikh:
I very much miss JsonContent assertions, and this is an awesome change. @donalmurtagh, while you are on it, can you also add the same method to WebTestClient? I think the reactive and imperative test clients should have symmetrical API.

donalmurtagh:
> I very much miss JsonContent assertions

You can still use them (without this new method). The "Current Workaround" section of #35624 shows how.

> can you also add the same method to WebTestClient?

That doesn't belong in this PR because #35624 is specifically about adding a `JsonContentAssert` method to `RestTestClient`. I'd rather wait and see whether this PR is accepted before adding a similar method to `WebTestClient`.

> I think the reactive and imperative test clients should have symmetrical API.

If that's the case, they should implement a common interface, but they don't.

rstoyanchev:
@donalmurtagh thanks for the suggested change, but I'll close this for now, and we'll keep the discussion under the issue. What you're bringing up goes beyond JSON content and needs to be considered more broadly.

This PR and the branch is still here for us to refer to.

Issue #34788, opened 2025-04-21
Unexpected query parameter order with UriComponentsBuilder


christophejan:
With spring-web 6.2.5, I get the following behavior :
```
System.out.println(UriComponentsBuilder.fromUriString(“http://myhost?a={p1}&b={p2}&a={p3}”).buildAndExpand(“a1”, “b1”, “a2”));
```
Output :
```
http://myhost?a=a1&a=b1&b=a2
```

I was expecting the following output :
```
http://myhost?a=a1&b=b1&a=a2
```

It seems that query params are reordered.

Even if it’s not a good idea to rely on query parameter order, it seems to be legal (see [stackoverflow > Is it legal or safe to depend on the ordering of URL query parameters?](https://stackoverflow.com/a/25687581))… So I think that things like RestClient may require proper query parameter order handling.

Furthermore, query params reordering seems to happen before uri template variables are expanded so values get swapped when given from an array (ie not from map) as the order of variables is then significant.

rstoyanchev:
We generally preserve the order of query parameters and their values. The only exception is when multiple values for the same query parameter are not specified together (and have other query parameters in between). This is due to the use of a MultiValueMap for internal storage that groups by query parameter name.

We do [support the WhatWG URL living standard](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-uri-building.html#uri-parsing) as of the parsing algorithms, but that's just the parsing side.

Arguably, we could do better with this, but there are options for you to eliminate ambiguity if this is a URI template that's in your control. Either declare query parameter values together, or use the `buildAndExpand` variant with a `Map`, or initialize the query parameters in a MultiValueMap.

christophejan:
There is indeed workarounds but the current behavior is :
* error prone (require some knowledge of the implementation storage)
* limiting (some api may require to not follow the current params grouping)

Leaving the MultiValueMap storage for a list of tuple would solve that and be closer to the specification.

I think that it could be done without breaking the backward compatibility and without introducing any drawbacks.

UriComponent is widely used so please consider it may worth this effort.

rstoyanchev:
Indeed there is room for improvement so I will mark the issue accordingly.

Mamun-Al-Babu-Shikder:
[...]

PR #35658, opened 2025-10-17
#34788 -   HierarchicalUriComponents can handle repeated params in a query

FBibonne:
Use List<QueryParam> instead of MultiValueMap in  HierarchicalUriComponents to record query params


FBibonne:
Solves #34788

FBibonne:
While working on the PR I had to fix some tests, not directly related to URI compononents, like [RequestLoggingFilterTests::queryStringIncluded](https://github.com/spring-projects/spring-framework/pull/35658/commits/6d75001e78f1259c317727d3f3ae8444e8987215#diff-1da6721fc22dfd4d39c09e8ff62cdeee5e04b08536c1a8ec90cc50286ec831fe)
So the changes in URI components, as they modify the order of query params, can fail tests which assert an URI : it is true  inside the spring framework tests but this can also be true outside : this may be a breaking change.

PR #35660, opened 2025-10-17
Use empty array constants instead of repeatedly creating new ones

kamilkrzywanski:
Refactor: use predefined empty arrays instead of creating new ones in places identified during runtime where unnecessary objects were frequently created.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
@@ -76,7 +76,8 @@ public class AdvisedSupport extends ProxyConfig implements Advised {
 	 * supplied by the advisors.
 	 */
 	public static final TargetSource EMPTY_TARGET_SOURCE = EmptyTargetSource.INSTANCE;
-
+	/** Empty advisor array constant. */
+	public static final Advisor[] EMPTY_ADVISORS = new Advisor[0];
 
 	/** Package-protected to allow direct access for efficiency. */
 	@SuppressWarnings("serial")
@@ -288,7 +289,7 @@ public class AdvisedSupport extends ProxyConfig implements Advised {
 
 	@Override
 	public final Advisor[] getAdvisors() {
-		return this.advisors.toArray(new Advisor[0]);
+		return this.advisors.toArray(EMPTY_ADVISORS);
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


donalmurtagh (commented):
I don't see the point in this myself. I doubt it will ever make any noticeable difference to performance and the empty array declarations are avoidable clutter.

kamilkrzywanski:
@donalmurtagh Your suggested approach will produce a ClassCastException.
Here is a benchmark to compare these two approaches. I know it’s not a game-changer, but it seems to be a good practice, too.
```
ToArrayBenchmark.toArrayWithEmptyConst                     thrpt   20  67323.479 ± 515.563  ops/ms
ToArrayBenchmark.toArrayWithEmptyConst:gc.alloc.rate       thrpt   20   3081.736 ±  23.608  MB/sec
ToArrayBenchmark.toArrayWithEmptyConst:gc.alloc.rate.norm  thrpt   20     48.000 ±   0.001    B/op
ToArrayBenchmark.toArrayWithEmptyConst:gc.count            thrpt   20   1268.000            counts
ToArrayBenchmark.toArrayWithEmptyConst:gc.time             thrpt   20    882.000                ms
ToArrayBenchmark.toArrayWithNew                            thrpt   20  62202.810 ± 102.498  ops/ms
ToArrayBenchmark.toArrayWithNew:gc.alloc.rate              thrpt   20   3796.461 ±   6.257  MB/sec
ToArrayBenchmark.toArrayWithNew:gc.alloc.rate.norm         thrpt   20     64.000 ±   0.001    B/op
ToArrayBenchmark.toArrayWithNew:gc.count                   thrpt   20   1671.000            counts
ToArrayBenchmark.toArrayWithNew:gc.time                    thrpt   20   1090.000                ms
```

sbrannen:
We do in fact tend to _cache_ empty arrays in static fields if the empty arrays are reused.

For example, if you search the code base for `[] EMPTY_` you'll see that we have over 45 such constants in use in production code (i.e., excluding tests and test fixtures).

In other words, there is precedent for this form of micro optimization within Spring Framework.

sbrannen:
This has been merged into `main` in 948367092ca10baf3edfce9c8319dcbee18cbb2e and revised in 0fbebd856fd0b9d6b5e7d332f6a8556cf2a57e6a.

Thanks

PR #35665, opened 2025-10-20
Unnecessary `else` statements

jonathan-up:


bclozel:
Thanks for the proposal, but this is more of a cosmetic change and the PR itself is not signing the DCO. I don't think it's worth spending time on this change for no meaningful impact.

Issue #35213, opened 2025-07-17
Support wildcard path elements at the start of path patterns


bclozel:
While `AntPathMatcher` usage in MVC applications has been deprecated for a long time at this point, we recently received additional feedback about migration efforts.

`PathPattern` currently only allows path elements that match multiple path segments at the end of a path pattern; for example:
1. `/spring-framework/file/**` matches `/spring-framework/file/some/file.txt`
2. `/spring-framework/file/{*filename}` matches `/spring-framework/file/some/file.txt` and yields `filename=/some/file.txt`
3. `/**/file/index.html` is not allowed, pattern compilation fails
4. `/{*project}/file/index.html` is not allowed, pattern compilation fails
5. `/spring-framework/**/index.html` is not allowed, pattern compilation fails
6. `/spring-framework/{*folders}/index.html` is not allowed, pattern compilation fails

We would like to improve this situation by allowing 3) and 4). Cases 5) and 6) have been requested in the past (see #22847) but they will still be disallowed as the matching process is hard to define properly and creates undesired behavior.

Issue #34018, opened 2024-12-04
Deprecate use of PathMatcher in web modules


rstoyanchev:
`PathMatcher` has been used broadly in the Spring Framework since version 1.2, in core for Spring config, in web for mapping requests, and in messaging for routing messages.

After a long experience, `PathPatternParser` was introduced in Spring Framework 5 designed specifically for the needs of web applications, see #19112. Initially for WebFlux, but later for Spring MVC as well from 5.3. In addition to syntax adjustments, a key benefit is the use of parsed patterns for repeated matching at runtime.

`PathPatternParser` has been the default in Spring MVC since 6.0, and in Spring Boot since 2.6. Going forward, we don't intend to keep two competing abstractions for request path matching. In 7.0 we should deprecate use of `PathMatcher` in web modules.

PR #35672, opened 2025-10-21
Fix 6.2.x PathPatternRequestMatcher Migration Issues

Jyosua:
`antMatcher` and `AntPathMatcher` are deprecated with the expectation of users switching to `PathPatternRequestMatcher` and `PathPattern` instead, in preparation for migration to Spring Framework 7/Spring Boot 4. However, there was a problem in that a common pattern with a wildcard at the start of path patterns was not supported in `PathPattern`, but _**was**_ in `AntPathMatcher`. This was fixed by the resolution of #35213, but these changes were only merged into main and not backported into 6.2.X.

As a result, you cannot smoothly deprecate all `AntPathMatcher` usages prior to switching to Spring Framework 7 if you were matching on wildcards at the start of paths.

This PR seeks to fix this by backporting these changes and the sequential regression fix found in #35339

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webflux/controller/ann-requestmapping.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webflux/controller/ann-requestmapping.adoc
@@ -93,6 +93,10 @@ You can map requests by using glob patterns and wildcards:
 |===
 |Pattern |Description |Example
 
+| `spring`
+| Literal pattern
+| `+"/spring"+` matches `+"/spring"+`
+
 | `+?+`
 | Matches one character
 | `+"/pages/t?st.html"+` matches `+"/pages/test.html"+` and `+"/pages/t3st.html"+`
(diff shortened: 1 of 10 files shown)


bclozel:
Thanks for the proposal @Jyosua 

Do you encounter such an issue right now in your upgrade experience? Can you share the use case and path pattern being used? Maybe there is another way to support this.

The feedback that triggered #35213 was very late in the process, literally 8 years after the introduction of `PathPattern` and many years after making it the default in Spring Boot and Spring Framework. At this stage of the 6.2.x line, I think that backporting this change is more likely to introduce regressions than help with upgrades. This change has not been released to a production-grade version so far and it could lead to runtime bugs or performance issues.

Maybe we can reconsider a backport after 7.0 has been adopted by the community and if we get significant demand from the community.

Jyosua:
Yes, actually! We typically try to keep up with deprecations and move to the newer way of doing things as we have time. I think antMatcher has only been marked as deprecated for around a yearish, if I'm recalling correctly, and as the current client we're working with runs on more of a waterfall style of development, we only got around to trying to clean up this specific one relatively recently when upgrading the spring version. We're writing a backend-for-frontend type of application that sits in front of several other backend services. As a result, there are times that we don't want certain filters to run on certain types of endpoints, which we may have the same type of endpoint under multiple paths. To give a genericized example, we might have `pathA/foo/exclude-me` and `pathB/bar/exclude-me`, and we want to exclude any endpoints ending in `/exclude-me` from `ExampleFilter`, which is your typical run once filter implementation. The typical way we've been doing this is with an antMatcher using the pattern `**/exclude-me`.

bclozel:
Thanks for the info @Jyosua 

At this stage, `AntPathMatcher` usage in the web stack is deprecated in 7.0 but not removed (see #34018), so it's still possible to use 7.0 as a step towards `PathPattern` usage in your project.

Jyosua:
Hi @bclozel, the problem is that [AntPathRequestMatcher](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java) no longer exists in the `main` branch of Spring-Security, which I believe is also the version that corresponds to Spring Boot 4 and Spring Framework 7. So if you try to upgrade, it will immediately break, rather than being able to switch to `PathPatternRequestMatcher` while still on Spring Boot 3, verify everything is working, and then trying to upgrade.

> As a result, there are times that we don't want certain filters to run on certain types of endpoints, which we may have the same type of endpoint under multiple paths. To give a genericized example, we might have pathA/foo/exclude-me and pathB/bar/exclude-me, and we want to exclude any endpoints ending in /exclude-me from ExampleFilter, which is your typical run once filter implementation. The typical way we've been doing this is with an antMatcher using the pattern **/exclude-me.

I was out of the house when I wrote this. To be a little more explicit, we're doing something like:
```kotlin
val requestMatcher = OrRequestMatcher(
  antMatcher(POST, "/**/exclude-me-1"),
  antMatcher(POST, "/**/exclude-me-2"),
)
```

This `antMatcher` call is a convenience method for instantiating an `AntPathRequestMatcher` instance from `org.springframework.security.web.util.matcher` and the documentation for [Web Migrations](https://docs.spring.io/spring-security/reference/migration-7/web.html#use-path-pattern) tells us to replace these with `PathPatternRequestMatcher`, a class from the core Spring Framework... but it doesn't support all the same use cases as `AntPathRequestMatcher` in the last version available before `AntPathRequestMatcher` is completely removed 😅

bclozel:
Thanks for the feedback @Jyosua , I have backported changes in #35679

Jyosua:
Thank you!

PR #35698, opened 2025-10-28
RestTestClient.mutate() should not have side effects

UpStreamThomas:
Before this change, `DefaultRestTestClient.mutate()` was reusing the underlying builder all calls.

Building a new builder for each call clones the `RestTestClientBuilder`, protecting from side effects.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/client/DefaultRestTestClient.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/client/DefaultRestTestClient.java
@@ -132,7 +132,7 @@ class DefaultRestTestClient implements RestTestClient {
 	@SuppressWarnings("unchecked")
 	@Override
 	public <B extends Builder<B>> Builder<B> mutate() {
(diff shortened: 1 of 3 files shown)


RemiVangrevelynghe (approved):
Thanks !

rstoyanchev:
Thanks for the pull request .This is done in `.build()` currently, but should be done in `.mutate()` instead.

UpStreamThomas:
> Thanks for the pull request .This is done in `.build()` currently, but should be done in `.mutate()` instead.

I did not notice that that builder was cloned in the `build()` method, do you want me to remove it ?

rstoyanchev:
Yes, it should be removed from `.build()`, effectively moved to `.mutate()`.

UpStreamThomas:
> Yes, it should be removed from `.build()`, effectively moved to `.mutate()`.

I added a commit with clone removed from `build()`.
I also rebased the branch against main.

rstoyanchev:
On closer look, we need to keep the cloning in build() to protect against changes to the original builder. I've added an additional test for that.

PR #35738, opened 2025-11-01
Add support for detecting only local beans for DispatcherServlet

filiphr:
The goal of this PR is to be able to configure the `DispatcherServlet` to only look for beans in the local application context, i.e. do not look in all ancestors.

The PR is currently only doing this for the `HandlerMapping`(s). However, if this is something that is acceptable for the Spring Team then I can adjust the rest of the beans in the dispatcher servlet:

* `handlerAdapters`
* `handlerExceptionResolvers`
* `viewResolvers`

The reason for this is that we are using different `DispatcherServlet`(s) to expose different REST APIs and with Spring Boot Actuator the actuator endpoints were being exposed in our servlets as well. We have remedied that using Spring Security and the `AntPathMatcher` to block access to `**/actuator/*`. However, the `AntPathMatcher` will no longer be there in Spring 7 and I wanted to find a better solution.

I did manage to find [`WebMvcEndpointChildContextConfiguration`](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-webmvc/src/main/java/org/springframework/boot/webmvc/autoconfigure/actuate/web/WebMvcEndpointChildContextConfiguration.java) from Spring Boot, that is achieving exactly what we are looking for. However, I wanted to see if we can avoid the need of all those `CompositeHandlerMapping`, `CompositeHandlerAdapter`, etc. and let the `DispatcherServlet` itself handle this.

If this gets accepted then the different composites in Spring Boot can be removed, and the Actuator management `DispatcherServlet` would only need to change 

```java
dispatcherServlet.setDetectAllHandlerAdapters(false);
dispatcherServlet.setDetectAllHandlerExceptionResolvers(false);
dispatcherServlet.setDetectAllHandlerMappings(false);
dispatcherServlet.setDetectAllViewResolvers(false);
```

to something like

```java
dispatcherServlet.detectLocalHandlerMappingsOnly()
dispatcherServlet.detectLocalHandlerExceptionResolversOnly();
dispatcherServlet.detectLocalHandlerMappingsOnly();
dispatcherServlet.detectLocalViewResolversOnly();
```

----

I'm also open to adjusting the naming of the methods as well. It does not have to be exactly like I've done in this PR.

Thanks for taking the time and considering this PR

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
@@ -149,6 +149,7 @@ import org.springframework.web.util.WebUtils;
  * @author Chris Beams
  * @author Rossen Stoyanchev
  * @author Sebastien Deleuze
+ * @author Filip Hrisafov
  * @see org.springframework.web.HttpRequestHandler
  * @see org.springframework.web.servlet.mvc.Controller
  * @see org.springframework.web.context.ContextLoaderListener
@@ -261,8 +262,8 @@ public class DispatcherServlet extends FrameworkServlet {
 	/** Store default strategy implementations. */
 	private static @Nullable Properties defaultStrategies;
 
-	/** Detect all HandlerMappings or just expect "handlerMapping" bean?. */
-	private boolean detectAllHandlerMappings = true;
+	/** Detect all HandlerMappings, only local HandlerMappings or just expect "handlerMapping" bean?. */
+	private DetectionStrategy detectHandlerMappingsStrategy = DetectionStrategy.ALL_BEANS;
 
 	/** Detect all HandlerAdapters or just expect "handlerAdapter" bean?. */
 	private boolean detectAllHandlerAdapters = true;
@@ -374,9 +375,21 @@ public class DispatcherServlet extends FrameworkServlet {
 	 * just a single bean with name "handlerMapping" will be expected.
 	 * <p>Default is "true". Turn this off if you want this servlet to use a single
 	 * HandlerMapping, despite multiple HandlerMapping beans being defined in the context.
+	 * @see #detectLocalHandlerMappingsOnly()
 	 */
 	public void setDetectAllHandlerMappings(boolean detectAllHandlerMappings) {
-		this.detectAllHandlerMappings = detectAllHandlerMappings;
+		this.detectHandlerMappingsStrategy =
+				(detectAllHandlerMappings ? DetectionStrategy.ALL_BEANS : DetectionStrategy.SINGLE_BEAN);
+	}
+
+	/**
+	 * Configures the servlet to only detect HandlerMapping beans from the local context.
+	 * <p>The default is to detect HandlerMappings from all context ancestors.
+	 * @since 7.0
+	 * @see #setDetectAllHandlerMappings
+	 */
+	public void detectLocalHandlerMappingsOnly() {
+		this.detectHandlerMappingsStrategy = DetectionStrategy.LOCAL_BEANS;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


bclozel:
Hello @filiphr - independently of this request, I'm wondering if you noticed that `PathPattern` has been improved recently to closed that gap in security matchers. See https://github.com/spring-projects/spring-framework/issues/35213

filiphr:
Thanks for sharing that @bclozel, yes I did notice the improvements in the `PathPattern`. However, that is not enough for us, since for us the pattern we need is `/**/actuator/**`. We use that to do a `denyAll` for all actuator endpoints which are not in the default dispatcher servlet.

We had something like:

```java
public class ActuatorRequestMatcher extends ApplicationContextRequestMatcher<WebApplicationContext> {

    private static final RequestMatcher EMPTY_MATCHER = (request) -> false;

    private volatile RequestMatcher delegate;

    public ActuatorRequestMatcher() {
        super(WebApplicationContext.class);
    }

    @Override
    protected final void initialized(Supplier<WebApplicationContext> context) {
        this.delegate = createDelegate(context.get());
    }

    @Override
    protected final boolean matches(HttpServletRequest request,
        Supplier<WebApplicationContext> context) {
        return this.delegate.matches(request);
    }

    private RequestMatcher createDelegate(WebApplicationContext context) {
        try {
            String pathPrefix = getPathPrefix(context);
            RequestMatcherFactory requestMatcherFactory = new RequestMatcherFactory(pathPrefix);
            return createDelegate(context, requestMatcherFactory);
        } catch (NoSuchBeanDefinitionException ex) {
            return EMPTY_MATCHER;
        }
    }

    private String getPathPrefix(WebApplicationContext context) {
        try {
            return context.getBean(DispatcherServletPath.class).getPrefix();
        } catch (NoSuchBeanDefinitionException ex) {
            return "";
        }
    }

    protected RequestMatcher createDelegate(WebApplicationContext context, RequestMatcherFactory requestMatcherFactory) {
        WebEndpointProperties properties = context.getBean(WebEndpointProperties.class);
        if (StringUtils.isNotEmpty(properties.getBasePath())) {
            return new OrRequestMatcher(
                requestMatcherFactory.antPath(properties.getBasePath() + "/**"),
                requestMatcherFactory.antPath("/**" + properties.getBasePath() + "/**")
            );
        }
        return EMPTY_MATCHER;
    }

    private static class RequestMatcherFactory {

        private final String servletPath;

        RequestMatcherFactory(String servletPath) {
            this.servletPath = servletPath;
        }

        public RequestMatcher antPath(String part) {
            String pattern = (this.servletPath.equals("/") ? "" : this.servletPath);
            return new AntPathRequestMatcher(pattern + part);
        }

    }
}

```

Which was inspired by the Spring Boot `EndpointRequest`. And the way we would use it was:

```java
http
        .securityMatcher(new ActuatorRequestMatcher())
        .authorizeHttpRequests(configurer -> configurer
                .requestMatchers(EndpointRequest.to(InfoEndpoint.class, HealthEndpoint.class)).permitAll()
                .requestMatchers(EndpointRequest.toAnyEndpoint()).hasAuthority(SecurityConstants.ACCESS_ACTUATORS)
                .anyRequest().denyAll()
        )
        .httpBasic(Customizer.withDefaults());
```


We did this long time ago, and back then I was not that well versed in the lower levels of the `DispatcherServlet`. For us, we don't even want to have the actuator endpoints in the child dispatcher servlets. By the way, I am also now aware that we could have use the `applicationContext` in the matcher to get all of our custom `ServletRegistrationBean`(s) that use `DispatcherServlet` and just use the url mappings defined from there, but that was long time ago and I was that well versed in these low level things.

bclozel:
Ah, thanks for the additional information. `PathPattern` does not support such patterns because they are costly and tend to cause expectation issues when it comes to sorting matching patterns. We'll consider this PR in due time.

rstoyanchev:
What about a custom `RequestMatcher` that obtains the path as a `PathContainer` along the lines of [PathPatternRequestMatcher](https://github.com/spring-projects/spring-security/blob/27ae318992739b607a185048778383c32119fd74/web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java)? After that it would be a matter of iterating the path segments, and checking if any is equal to "actuator".

filiphr:
Thanks @rstoyanchev, that might work as well, I didn't think about that. My other approach, was something like:

```java
protected RequestMatcher createDelegate(WebApplicationContext context, PathPatternRequestMatcher.Builder pathMatcherBuilder) {
        WebEndpointProperties properties = context.getBean(WebEndpointProperties.class);
        if (StringUtils.isNotEmpty(properties.getBasePath())) {
            Collection<String> urlMappings = new ArrayList<>();
            for (ServletRegistrationBean<?> registrationBean : context.getBeansOfType(ServletRegistrationBean.class).values()) {
                if (registrationBean instanceof DispatcherServletRegistrationBean) {
                    continue;
                }
                if (registrationBean.getServlet() instanceof DispatcherServlet) {
                    urlMappings.addAll(registrationBean.getUrlMappings());
                }
            }

            List<RequestMatcher> matchers = new ArrayList<>(urlMappings.size() + 1);
            matchers.add(pathMatcherBuilder.matcher(properties.getBasePath() + "/**"));

            for (String urlMapping : urlMappings) {
                String url = urlMapping.endsWith("*") ? urlMapping.substring(0, urlMapping.length() - 1) : urlMapping;
                matchers.add(pathMatcherBuilder.matcher(url + properties.getBasePath() + "/**"));
            }

            return new OrRequestMatcher(matchers);
        }
        return EMPTY_MATCHER;
    }
```

In any case, for us the preferred way is now to go with making sure to only get handler mappings from the child context. I think that something like this would be beneficial in general for the `DispatcherServlet`.

filiphr:
Hey @rstoyanchev, have you had the time to look into the PR and decide whether or not it is something that is acceptable to have in Spring Web?

If this is something that you don't want to have in Spring Web, then maybe we should close the PR.

rstoyanchev:
@filiphr we have looked at this. The change makes sense, but haven't quite decided on the best way to change the properties, which go a long way back and are due for a revisit. It's not clear for example, if they need to be separate by strategy, or if they would typically mirror each other.

A question on your setup. Boot uses a child context for actuator endpoints when the management port is different, but you mention they are in the parent context. Could you provide more details on your setup, for example if we were to try and replicate it?

rstoyanchev (changes requested):
The proposal makes sense, but the implementation approach will be a little different.

The `detectAll*` properties can remain as they are. They are orthogonal to the request.

We can add a new boolean property, e.g. `detectInLocalContext` to switch from the default mode of checking the full context hierarchy.

While it makes sense for the existing `detectInAll*` to be more fine-grained, the local lookup is more cross-cutting and can be just a single property that applies to all lookups.

Would you like to update the PR accordingly?

filiphr:
Hey @rstoyanchev, thanks a lot for looking into this. Sorry that it took me a while to reply here.

>A question on your setup. Boot uses a child context for actuator endpoints when the management port is different, but you mention they are in the parent context. Could you provide more details on your setup, for example if we were to try and replicate it?

Yes you are right that Boot does use a child context when the management port is set. However, that only works when used as a standalone JAR, not when used as a WAR in a Tomcat and you have to explicitly configure it. Unfortunately, we don't control how our users would deploy this. Also more often then not, our users take our starters and build their own apps.

Our real configuration is in https://github.com/flowable/flowable-engine/blob/main/modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/main/java/org/flowable/spring/boot/RestApiAutoConfiguration.java.

What we basically have is a global application context then for different parts of the application we create different servlets:

* `process-api` - For Process Engine related REST APIs
* `cmmn-api` - For CMMN Engine related REST APIs
* `dmn-api` - For DMN Engine related REST APIs

each one of those registers a Spring Boot `ServletRegistrationBean` with its own `DispatcherServlet` and it's own base `@Configuration` class. I could create a simpler demo application if you want to have a better look at it.


>While it makes sense for the existing `detectInAll*` to be more fine-grained, the local lookup is more cross-cutting and can be just a single property that applies to all lookups.

That doesn't quite fit all are requirements. If we have a single property for detect local, then this would be applicable for all the lookups. However, we still want to expose shared exception handles and / or to allow users to expose additional exception handlers by registering them in the root application context.

For convenience we could offer `public void detectLocalOnly()` which can then set the lookup strategy for each one to local.

filiphr:
Hey @rstoyanchev, did you by any chance had the time to look at my last comment? If you still disagree with me, then I'll go ahead and do the changes you proposed in https://github.com/spring-projects/spring-framework/pull/35738#pullrequestreview-3924064382, we can still use the workaround I've found.

PR #35760, opened 2025-11-05
Improve documentaton on controller return values for file downloads

verils:
I've updated the documentation to indicate that a ResponseEntity<Resource> can be returned from a WebFlux controller.

On a separate note, since file uploads and downloads are common requirements in the Spring framework, I believe it would be beneficial to create a dedicated page to guide developers through the process. I'd be happy to create one. What are your thoughts?

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webflux/controller/ann-methods/responseentity.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webflux/controller/ann-methods/responseentity.adoc
@@ -33,6 +33,8 @@ Kotlin::
 ----
 ======
 
+A `ResponseEntity<Resource>` can be returned for file content, as same as in Spring MVC,
+
 WebFlux supports using a single value xref:web-reactive.adoc#webflux-reactive-libraries[reactive type] to
 produce the `ResponseEntity` asynchronously, and/or single and multi-value reactive types
 for the body. This allows a variety of async responses with `ResponseEntity` as follows:
--- a/framework-docs/modules/ROOT/pages/web/webflux/controller/ann-methods/return-types.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webflux/controller/ann-methods/return-types.adoc
@@ -84,6 +84,13 @@ Reactor provides a dedicated operator for that, `Flux#collectList()`.
   to be written (however, `text/event-stream` must be requested or declared in the mapping
   through the `produces` attribute).
 
+| `Flux<DataBuffer>`
+| Write raw data buffers to the response, for example, when handling file
+  downloads or proxying data.
+
+| `Resource`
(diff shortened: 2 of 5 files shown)


rstoyanchev (changes requested):
This is a worthwhile improvement to consider, but not how I would address it. 

Both Spring MVC and WebFlux list [supported return types](https://docs.spring.io/spring-framework/reference/web/webflux/controller/ann-methods/return-types.html). Neither of them mentions `Resource` explicitly, but that has more to do with the supported codecs (WebFlux) and message converters (WebMvc).

My suggestion is to begin by adding a subsection on `Resource`s under [Codecs](https://docs.spring.io/spring-framework/reference/web/webflux/reactive-spring.html#webflux-codecs) and [Message Converters](https://docs.spring.io/spring-framework/reference/web/webmvc/message-converters.html#message-converters). Then this can be mentioned as well as referenced from the list of return values for `ResponseEntity` and `@ResponseBody`.

Let me know if you plan to update the PR along those lines, or if you only update the WebFlux side, we can do the rest.

rstoyanchev:
> On a separate note, since file uploads and downloads are common requirements in the Spring framework, I believe it would be beneficial to create a dedicated page to guide developers through the process. 

We have [Multipart Content](https://docs.spring.io/spring-framework/reference/web/webflux/controller/ann-methods/multipart-forms.html) for WebFlux and similar for WebMvc that cover uploads. Renaming that to Multipart Uploads, and creating another on Multipart Downloads would be fine.

verils:
Initially, the issue I encountered was that I didn't know how to handle content downloads in WebFlux, as there was no clear explanation in the documentation. I did some search online and finally found out that returning `ResponseEntity<Flux<DataBuffer>>` is the best option for me. Therefore, my main goal is to clarify it, the [DataBuffer](https://docs.spring.io/spring-framework/reference/web/webflux/reactive-spring.html#webflux-codecs-buffers) section is very close to the truth but isn't explicit enough.

As @rstoyanchev pointed out, the [Return Values](https://docs.spring.io/spring-framework/reference/web/webflux/controller/ann-methods/return-types.html) section is the best place to address this, I'm gonna add a `Flux<DataBuffer>` line to that table.

I also used `Resource` for a while but encountered a side effect: the `ResourceHttpMessageWriter` (in WebFlux) and the `ResourceHttpMessageConverter` (in WebMvc) automatically override the response's Content-Type with media type from the `Resource`. This can break an `application/octet-stream` download, and this behavior should also be documented.

rstoyanchev:
@verils, I don't think we should avoid documenting returning `Resource` as a way to handle downloads just because of an issue you ran into. The issue should be considered to decide whether it is intentional, a limitation, something that needs fixing, and so on.

I will say briefly that if the returned ResponseEntity has the content-type set then I would expect it to get used. However, I would also want to know why the content type of the resource isn't the right one to begin with. This is best discussed in a separate issue.

verils:
@rstoyanchev Yes, I agree that.

So I'll just deal with the return type list first.

PR #35765, opened 2025-11-06
#35621 Preserve Original Produces Media Types in Exception Handlers

IMurzich:
Fix issue #35621


bclozel:
Thanks for the proposal, but this is a long standing behavior and we want to keep it, see #20865.

The error handling method does not necessarily produce the same content-type as the handler that errored in the first place. Content negotiation has to be considered again.

Issue #35767, opened 2025-11-06
Fix spring-core multirelease jar


Sineaggi:
The multi-release jar for spring-core doesn't validate when running `jar --validate --file spring-core.jar`. Instead, the following warnings are printed.

```
jar --validate --file /Users/cwalker/public-repos/spring-framework/spring-core/build/libs/spring-core-7.0.0-SNAPSHOT.jar
entry: META-INF/versions/21/org/springframework/core/task/VirtualThreadDelegate.class, contains a class with different api from earlier version
entry: META-INF/versions/24/org/springframework/core/type/classreading/ClassFileMetadataReaderFactory.class, contains a new public class not found in base entries
```

These can be fixed by adding an empty constructor to `VirtualThreadDelegate` and by removing the `public` modifier in `ClassFileMetadataReaderFactory`.

Sineaggi:
I've added an example fix + multi-release jar validation task in this pr https://github.com/spring-projects/spring-framework/pull/35768.

bclozel:
Superseded by the PR

PR #35768, opened 2025-11-06
Add jar validation and fix multi-release jar

Sineaggi:
As per https://github.com/spring-projects/spring-framework/issues/35767, the existing multi-release jar contains two incorrect entries.

This mr fixes those entries and adds a validation task to avoid this in the future.


bclozel:
Thanks for your contribution @Sineaggi !

PR #35782, opened 2025-11-08
Add @AcceptableExtension validation for file uploads

lifefire1:
## Description

Hi Spring team!

I'm a frequent Spring user and I really enjoy working with it. However, I often find myself writing boilerplate code to validate file extensions when handling file uploads. 

To solve this, I've created an `@AcceptableExtension` annotation that allows easy declarative validation of file extensions.

### Example Usage

**Before:**
```java
@PostMapping("/upload")
public String upload(@RequestParam("file") MultipartFile file) {
    String filename = file.getOriginalFilename();
    String extension = filename.substring(filename.lastIndexOf(".") + 1);
    if (!List.of("jpg", "png", "pdf").contains(extension.toLowerCase())) {
        throw new IllegalArgumentException("Invalid file extension");
    }
    // process file...
    return "success";
}
```

**After:**
```java
@PostMapping("/upload")
public String upload(
        @AcceptableExtension(extensions = {"jpg", "png", "pdf"})
        @RequestParam("file") MultipartFile file) {
    // file is already validated
    return "success";
}
```

### What's Included
- `@AcceptableExtension` annotation with customizable extensions and error messages
- `AcceptableExtensionMethodArgumentResolver` for validation logic
- Comprehensive test suite (15+ test cases)
- Case-insensitive extension matching

All tests pass successfully.

Thank you for considering this contribution!

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/annotation/AcceptableExtensionMethodArgumentResolver.java
+++ b/spring-web/src/main/java/org/springframework/web/method/annotation/AcceptableExtensionMethodArgumentResolver.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2002-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.method.annotation;
+
+import java.util.Arrays;
+
+import jakarta.servlet.http.HttpServletRequest;
+import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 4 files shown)


spring-web/src/main/java/org/springframework/web/method/annotation/AcceptableExtensionMethodArgumentResolver.java
```diff
@@ -0,0 +1,132 @@
...
+	private boolean isAcceptableExtension(String extension, String[] acceptableExtensions) {
+		if (acceptableExtensions.length == 0) {
+			return true;
+		}
+		for (String acceptable : acceptableExtensions) {
+			if (acceptable.equalsIgnoreCase(extension)) {
+				return true;
+			}
+		}
+		return false;
```
ronodhirSoumik:
```
    if (acceptableExtensions.length == 0) {
        return true;
    }

    return Arrays.stream(acceptableExtensions)
            .anyMatch(acceptable -> acceptable.equalsIgnoreCase(extension));
```

lifefire1: The comment has been corrected

bclozel:
Thanks for this proposal, but I don't think this use case is common enough that it should be included in Spring Framework by default. Maybe try and share this as a library first for the community? We can reconsider if there is enough demand for this enhancement.

Thanks!

PR #35817, opened 2025-11-13
Fix `Spliterator` characteristics in `ConcurrentReferenceHashMap`

kilink:
The Spliterators returned by values, entrySet, and keySet incorrectly reported the SIZED characteristic, instead of CONCURRENT. This could lead to bugs when the map is concurrently modified during a stream operation.

For keySet and values, the incorrect characteristics are inherited from AbstractMap, so to rectify that the respective methods are overridden and custom collections are provided that report the correct Spliterator characteristics.

The same sort of bug was [reported and fixed](https://github.com/ben-manes/caffeine/issues/1883) in Caffeine, and the issue seems to exist in several other ConcurrentMap implementations in the wild.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ConcurrentReferenceHashMap.java
+++ b/spring-core/src/main/java/org/springframework/util/ConcurrentReferenceHashMap.java
@@ -20,8 +20,10 @@ import java.lang.ref.ReferenceQueue;
 import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
 import java.lang.reflect.Array;
+import java.util.AbstractCollection;
 import java.util.AbstractMap;
 import java.util.AbstractSet;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


spring-core/src/main/java/org/springframework/util/ConcurrentReferenceHashMap.java
```diff
@@ -101,7 +105,17 @@ public class ConcurrentReferenceHashMap<K, V> extends AbstractMap<K, V> implemen
 	/**
 	 * Late binding entry set.
 	 */
-	private volatile @Nullable Set<Map.Entry<K, V>> entrySet;
+	private @Nullable Set<Map.Entry<K, V>> entrySet;
```
kilink, line 108: I remove `volatile` here, and didn't make the new fields `volatile`, since this is just a benign race; [ConcurrentHashMap](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java#L832) does the same thing. Seems preferable to paying the cost of the volatile field access.

jhoeller:
Good point, let's fix this for 7.0.1. This is also worth backporting to 6.2.14.

sbrannen:
This has been merged into `main` and backported to `6.2.x`.

Thanks 👍

Issue #33813, opened 2024-10-29
Deregister empty `Cache` from `CacheManager`


trung:
I use `CaffeineCacheManager` and create temporary caches dynamically for each request. When a request completes, the cache items are evicted, leaving the cache instances empty but still in memory.

`CaffeineCacheManager` has an internal `cacheMap` which keeps growing as caches are dynamically created. 

Is there away to remove/deregister the empty cache instances from `cacheMap`?

jhoeller:
Good point, I'll add a `removeCache(String name)` for 6.2.

trung:
Please also add `removeCache(String name)` for `ConcurrentMapCacheManager` as well. I'm not using it but I do see it has similar need.

ben-manes:
@jhoeller is instead `@RequestScoped CacheManager` the intended idiom?
(that was my best guess when asked in https://github.com/ben-manes/caffeine/discussions/1790)

jhoeller:
@ben-manes that would indeed work but I can also see the need for runtime removal of certain caches in a shared CacheManager scenario. As long as the cache names are distinct enough, I can see that working for short-lived request caches as well.

PR #35821, opened 2025-11-14
Fix memory leak in CaffeineCacheManager static mode

NYgomets:
## Description
This PR fixes a memory leak in `CaffeineCacheManager` where dynamically created caches were not removed when switching to static mode via `setCacheNames()`.

### Core Fix
- Modified `CaffeineCacheManager.setCacheNames()` to clear dynamically created caches while preserving custom caches
- Updated Javadoc to clarify the behavior

### Tests Added
- `setCacheNamesShouldRemovePreviousDynamicCaches()`: Verifies dynamic caches are cleared
- `setCacheNamesShouldPreserveCustomCaches()`: Ensures custom caches are not affected
- `switchingBetweenDynamicAndStaticMode()`: Tests mode transitions

## Backward Compatibility

This is a **potentially breaking change** for users who rely on the (undocumented) behavior of dynamic caches persisting after `setCacheNames()`.

However:
1. The current behavior violates the documented contract
2. The memory leak justifies the fix
3. The typical use case (setting static caches at startup) is unaffected

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
+++ b/spring-context-support/src/main/java/org/springframework/cache/caffeine/CaffeineCacheManager.java
@@ -106,9 +106,16 @@ public class CaffeineCacheManager implements CacheManager {
 	 * with no creation of further cache regions at runtime.
 	 * <p>Calling this with a {@code null} collection argument resets the
 	 * mode to 'dynamic', allowing for further creation of caches again.
+	 * <p><b>Note:</b> Switching to static mode will remove all dynamically created
+	 * caches, while preserving custom caches registered via
+	 * {@link #registerCustomCache(String, com.github.benmanes.caffeine.cache.Cache)}.
 	 */
 	public void setCacheNames(@Nullable Collection<String> cacheNames) {
 		if (cacheNames != null) {
+			// Remove all non-custom caches before setting up static caches
+			this.cacheMap.keySet().retainAll(this.customCacheNames);
+
+			// Add the specified static caches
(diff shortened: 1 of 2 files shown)


jhoeller:
On review, `setCacheNames` is really a configuration-time method and not meant to be called at runtime when caches already exist. That said, point taken that it can effectively be used for re-configuration purposes as well, so we need to take that into account to some degree.

For the specific case of "leaking" dynamic cache regions when a new set of fixed cache names is being configured, these formerly registered dynamic regions keep being accessible, so they are not really leaking. Also, if a specific set of cache names is set through `setCacheNames` initially, and then later on a smaller or different set of cache names is set on yet another `setCacheNames` call, we are not removing the old caches either, rather just adding the newly specified ones. The `registerCustomCache` facility makes the semantics even less clear there since those are always worth retaining.

Overall, I don't see us adding official clean-up semantics to `setCacheNames`. Instead, we're introducing a `resetCaches()` method (#35840) which can be called before `setCacheNames` if such clean-up is desirable: That method can generally be used for cache clearing at runtime, removing dynamic caches or simply clearing static caches. This goes nicely with the existing `removeCache` method that we introduced in 6.1.15 (#33813).

NYgomets:
Thanks for the detailed explanation, Juergen.

I understand the design philosophy regarding setCacheNames as a configuration-time method and the risk of implicit cleanup. Introducing an explicit resetCaches() method seems like a much safer and more flexible approach to handle runtime reconfiguration scenarios.

I'm glad that my report contributed to improving the cache management capabilities. Thanks for addressing this!

PR #35822, opened 2025-11-14
Fix single-check idiom in `UnmodifiableMultiValueMap`

kilink:
Read the respective fields only once in values, entrySet, and keySet methods.

See: https://javaagile.blogspot.com/2013/05/the-racy-single-check-idiom.html


sbrannen:
This has been merged into `main` and `6.2.x`.

Thanks

PR #35823, opened 2025-11-15
Annotate CGLIB `Transformer` interface with `@FunctionalInterface`

devch37:
### Summary

Add the `@FunctionalInterface` annotation to `Transformer`.

### Details

`Transformer` already has a single abstract method, so marking it as a functional interface clarifies its intended use and provides better compile-time checks and IDE support.

### Compatibility

No behavior changes; fully backward-compatible.

### DIFF
--- a/spring-core/src/main/java/org/springframework/cglib/core/Transformer.java
+++ b/spring-core/src/main/java/org/springframework/cglib/core/Transformer.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.cglib.core;
 
+@FunctionalInterface
 public interface Transformer {
     Object transform(Object value);
 }


bclozel:
@behappyleee do you have a particular use case needing this change? Can you share a code snippet of what you are trying to achieve?

I am asking because this code is being shaded in spring and only apply strictly necessary patches there.

devch37:
> @behappyleee do you have a particular use case needing this change? Can you share a code snippet of what you are trying to achieve?
> 
> I am asking because this code is being shaded in spring and only apply strictly necessary patches there.

@bclozel 
Hi, thanks for checking this!

The reason I applied this change is that I wanted to use Transformer as a functional interface in a small utility module, mainly for passing lambda expressions.
Without the @FunctionalInterface annotation, the interface works, but the annotation helps enforce its intended usage and prevents accidental method additions during local development.

Nothing beyond this is required on Spring’s side, and there’s no need for any additional patches in the shaded code.
My change was only for my own usage outside of Spring’s internal shading logic.

If this causes any concern, feel free to point it out — I can adjust my approach accordingly.

Here’s a small snippet showing how I intended to use the Transformer interface as a functional interface:

```
Transformer trimTransformer = value -> {
    if (value instanceof String s) {
        return s.trim();
    }
    return value;
};

Object result = trimTransformer.transform("hello  ");

```


Thanks again for the clarification!

bclozel:
@jhoeller do you think it's worth applying this change to our cglib fork? It's adding maintenance cost so we must consider the tradeoffs here.

sbrannen:
> Here’s a small snippet showing how I intended to use the Transformer interface as a functional interface:
> 
> ```
> Transformer trimTransformer = value -> {
>     if (value instanceof String s) {
>         return s.trim();
>     }
>     return value;
> };
> 
> Object result = trimTransformer.transform("hello  ");
> ```

In that example, it appears that you are using `Transformer` without using anything else from CGLIB.

Whereas, as far as I know, `Transformer` is only intended to be used with the `transform(Collection, Transformer)` and `bucket(Collection, Transformer)` methods in `org.springframework.cglib.core.CollectionUtils`.

And I don't consider that to be a supported use case with Spring's fork of CGLIB.

Instead, if you want to use something like `Transformer` I would recommend the use of standard functional interfaces like `java.util.function.Function<T, R>`. Or if you use Apache Commons Collections, you could use `org.apache.commons.collections4.Transformer<I, O>`.

In summary, I don't see any value in annotating `org.springframework.cglib.core.Transformer` with `@FunctionalInterface`.

sbrannen:
Closing due to the rationale provided in https://github.com/spring-projects/spring-framework/pull/35823#issuecomment-3541900770.

devch37:
Thank you for the clarification and for pointing me in the right direction. I appreciate it.

PR #32345, opened 2024-02-29
Add web support for YAML via Jackson

Flamme1004K:
Hello. This is my first spring framework contribution code.

I created this PR to add an application/yaml MediaType.

Recently, I noticed that application/yaml has been officially added to MediaType. The link I checked is below.

Link
- https://www.rfc-editor.org/rfc/rfc9512.html
- https://www.iana.org/assignments/media-types/media-types.xhtml

Thanks a lot.

PR #35843, opened 2025-11-18
Remove deprecated MappingJackson2YamlHttpMessageConverter

Flamme1004K:
Hello.

I heard that starting from Spring 7.x, Jackson 3.x will be used.

Therefore, I plan to delete the previously prepared Jackson 2 YAML source code. 
- https://github.com/spring-projects/spring-framework/pull/32345

Afterward, if possible, I would also like to submit a pull request to delete all Jackson 2-related source code. If this is acceptable, please let me know in the comments.

Thank you.


bclozel:
Thanks for the propsal, but we did deprecate this code and plan to remove in the future. There is no need to remove right now as this is the typical bulk operation we perform when we remove all the deprecated code when starting to work on a new generation.

Flamme1004K:
@bclozel Thank you for letting me know.

Issue #35777, opened 2025-11-07
`./gradlew :buildSrc:test` failing on JDKs before JDK 25


msridhar:
On JDK 21:

```
$ ./gradlew --version

------------------------------------------------------------
Gradle 9.2.0
------------------------------------------------------------

Build time:    2025-10-29 13:53:23 UTC
Revision:      d9d6bbce03b3d88c67ef5a0ff31f7ae5e332d6bf

Kotlin:        2.2.20
Groovy:        4.0.28
Ant:           Apache Ant(TM) version 1.10.15 compiled on August 25 2024
Launcher JVM:  21.0.9 (Azul Systems, Inc. 21.0.9+10-LTS)
Daemon JVM:    /Library/Java/JavaVirtualMachines/zulu-21.jdk/Contents/Home (no JDK specified, using current Java home)
OS:            Mac OS X 15.7.1 aarch64
$ ./gradlew assemble --rerun-tasks

> Task :buildSrc:test

MultiReleaseJarPluginTests > validateJar() FAILED
    java.lang.AssertionError at MultiReleaseJarPluginTests.java:143

4 tests completed, 1 failed
```

I see the same failure on JDK 17.  On JDK 25 `./gradlew assemble` passes.

This is with the main branch, git SHA 386c6cafc9d68e89af79dd9b8cb10940ca81a0fb.  I'm seeing the same failure [in CI on Linux](https://github.com/uber/NullAway/actions/runs/19181903007/job/54840415408?pr=1334).

sdeleuze:
That also prevents me to run tests in IDEA.
> Task :validateMultiReleaseJar FAILED
module jdk.jartool does not have a ModuleMainClass attribute, use -m <module>/<main-class>

Maybe related to recent changes from #35773 in this area @bclozel ?

sbrannen:
I also just encountered this in a local Gradle build, where the build previously worked with JDK 17 configured for the current shell.

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':validateMultiReleaseJar'.
> Process 'command '/Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home/bin/java'' finished with non-zero exit value 1

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to generate a Build Scan (powered by Develocity).
> Get more help at https://help.gradle.org.

BUILD FAILED in 374ms
[...]

PR #35848, opened 2025-11-19
Fix buildSrc:test failing on JDKs before JDK 25 with Non-English locale

quaff:
See https://github.com/spring-projects/spring-framework/issues/35777#issuecomment-3551591336

### DIFF
--- a/buildSrc/src/test/java/org/springframework/build/multirelease/MultiReleaseJarPluginTests.java
+++ b/buildSrc/src/test/java/org/springframework/build/multirelease/MultiReleaseJarPluginTests.java
@@ -43,10 +43,13 @@ public class MultiReleaseJarPluginTests {
 
 	private File buildFile;
 
+	private File propertiesFile;
+
 	@BeforeEach
 	void setup(@TempDir File projectDir) {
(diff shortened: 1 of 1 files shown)


ChildrenGreens:
I am seeing the same `MultiReleaseJarPluginTests.validateJar()` failure on a non-English Windows environment.

Environment:
- OS: Windows 11 (zh_CN)
- JDK: 25 (C:\Program Files\Java\jdk-25)
- Console encoding: GBK

The error message printed by the `validateMultiReleaseJar` task is localized and garbled, for example:

> Task :validateMultiReleaseJar FAILED
> ��Ŀ META-INF/versions/17/Main.class ����������������ڰ汾��ͬ�� api

Because of this, the assertion

    assertThatThrownBy(...).hasMessageContaining(
        "entry: META-INF/versions/17/Main.class, contains a class with different api from earlier version"
    );

does not match, and the test fails.

This PR makes sense to me since it forces `user.language=en` for the TestKit build and should fix the issue for non-English locales as well.

bclozel:
Thanks @quaff !

PR #35864, opened 2025-11-21
docs: add spring startup roadmap and debug app

aotemiao:


bclozel:
Thanks for the proposal but we only accept contributions in English with the DCO signed.
Also, we don't intend to add this level of detail as this would get outdated very fast.

PR #35868, opened 2025-11-22
Refine nullability of `Assert#noNullElements`

ngocnhan-tran1996:
From reference
>`@Nullable Object @Nullable [] array means both individual elements and the array can be null.`

https://docs.spring.io/spring-framework/reference/core/null-safety.html

I think we should to add `@Nullable` because both the array and its elements may be null.


sdeleuze:
Merged, thanks!

PR #35869, opened 2025-11-22
refactor(StaxStreamXMLReader): remove duplicate ifs

RealTake:
**What was done**
- Removed unnecessary duplicate if conditions in the codebase.
- Ensured that no existing functionality was affected by the changes.


bclozel:
Build fails and the DCO is not signed, we cannot accept this contirbution and I don't it's worth spending time on such a minimal change.

PR #35870, opened 2025-11-22
refactor: DispatcherServlet Bean Initialization

RealTake:
- Unified single bean initialization methods into initSingleBean, removing duplicate logic.
- Consolidated list bean initialization methods into initListBeans, streamlining flag handling and default strategy.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
@@ -28,6 +28,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


mikelhamer (commented):
My overall impression is that while reducing duplication is good in theory, the changes here combine several focused, purpose-specific methods into one general method. Over time this will make the behavior harder to modify, because a single method now has to handle every scenario. It also takes a large number of parameters and still requires passing in a function to customize behavior, which suggests the method is doing too much. Sometimes duplication is the price we willingly pay for clarity, isolation, stability, and ease of change. Over generalizing can end up being more costly than a few repeated lines.

bclozel:
I agree that while reducing duplication, this makes the code harder to read and debug here. `DispatcherServlet` is an important piece of infrastructure and I'd rather not introduce complexity there for cosmetic changes.

Also, DCO is not signed so we cannot accept this contribution anyway.

PR #35899, opened 2025-11-26
Fix broken Javadoc links to methods

ngocnhan-tran1996:


sdeleuze:
Good catch, should be applied to `6.0.x` branch too.

PR #35924, opened 2025-11-28
Upgrade to json-path 2.10.0

ivonaest:
## Summary
 
1. Why:
To remove CVEs:
     - [CVE-2024-57699](https://avd.aquasec.com/nvd/2024/cve-2024-57699/)
     - This CVE affects `json-smart` library and can lead to denial-of-service through stack exhaustion when parsing malicious JSON

2. What:

     - Upgrade  `json-path` to 2.10.0 to remove [CVE-2024-57699](https://avd.aquasec.com/nvd/2024/cve-2024-57699/)
     - `json-smart` is a transitive dependency of `json-path` and it needed 2.5.2+ version
     - By upgrading `json-path` to 2.10.0, `json-smart` was upgraded to 2.6.0
 
 ## Additional evidence
 
 Partial output from security scanner Trivy:
spring-framework cves json-smart

 ## Categorization

- [x] security/CVE

### DIFF
--- a/framework-platform/framework-platform.gradle
+++ b/framework-platform/framework-platform.gradle
@@ -33,7 +33,7 @@ dependencies {
 		api("com.google.code.gson:gson:2.13.2")
 		api("com.google.protobuf:protobuf-java-util:4.32.1")
 		api("com.h2database:h2:2.3.232")
-		api("com.jayway.jsonpath:json-path:2.9.0")
+		api("com.jayway.jsonpath:json-path:2.10.0")
 		api("com.networknt:json-schema-validator:1.5.3")
 		api("com.oracle.database.jdbc:ojdbc11:21.9.0.0")
 		api("com.rometools:rome:1.19.0")


bclozel:
[CVE-2024-57699](https://avd.aquasec.com/nvd/2024/cve-2024-57699/) is irrelevant here as Spring Framework optionnally compiles against this dependency. Only applications depend on json-path.

ivonaest:
Hi @bclozel, while Spring is not vulnerable merging it is a small upgrade that would remote CVE from future security scans. Are you sure you don't want to merge it?

bclozel:
@ivonaest I will merge it. I am just stating that this dependency will not show up in security scans because it is not declared in any of our artifacts. Only "spring-test" compiles against it and you can check that it does not depend on it in the published POM: https://central.sonatype.com/artifact/org.springframework/spring-test

Upgrading dependencies is an important concern, but it's even more important to understand how dependency management works and whether a CVE is applicable. Here it is not.

ivonaest:
@bclozel Thank you for taking the time to explain CVE in this project. From my understanding, it is a false positive reported by Trivy scanner.

bclozel:
Can you share the part where Trivy points to spring-test as the culprit?

Trivy might be looking at the Gradle module information (where this dependency might show up) but is maybe missing the fact that it's not on the compile classpath.

ivonaest:
It showed up only here

image

bclozel:
I don't understand why Trivy thinks spring-test has guava and other dependencies as transitive dependencies. Neither [the POM file](https://repo1.maven.org/maven2/org/springframework/spring-test/6.2.14/spring-test-6.2.14.pom) nor the [Gradle module file](https://repo1.maven.org/maven2/org/springframework/spring-test/6.2.14/spring-test-6.2.14.module) lists anything like this. The `commons-beanutils` mention is also very strange.

If you can reach out to the Trivy team and get some answers, please let us know.

bclozel:
Thanks for your contribution @ivonaest

While I'm not sure it makes a real difference to applications, we are upgrading to json-path 2.10 in our maintenance releases due next week.

ivonaest:
Hi @bclozel , sorry for the late reply, I have been talking to my colleagues about this. The lockfiles were generates with multiple scopes and Trivy was looking at all of them. For example checkstyle and other scopes, that's why guava was there.
Lockfiles have been modified so that they only contain compileClasspath and runtimeClasspath.
This is the Trivy output I am getting now:
image

And I appreciate the merge, thank you

bclozel:
@ivonaest It's still not clear why json-smart is pointing to spring-test as a dependent. Is your application depending on json-smart directly? How is the version managed? It looks like the problem lies in the lockfile generation and not in Trivy after all.

Note that depending on how this version is managed, the json-smart upgrade in Spring Framework might not solve your problem. Especially if json-smart is used outside of the test scope: Framework does not use json-smart outside of the test classpath.

ivonaest:
@bclozel  , with command 
`./gradlew spring-test:dependencyInsight --dependency json-smart --configuration runtimeClasspath`
I get this output

json-smart 2 6 0

From my understanding and this output, it looks like `json-path` uses `json-smart` as default parser and that's why it's pulling it in transitively through `framework-platform`. And `spring-test` has `framework-platform` on it's runtimeClasspath so that's why Trivy was showing it as CVE related to `spring'test`.

bclozel:
@ivonaest Wait a minute, so you're running Trivy against the spring-framework project directly? You shouldn't be doing that. Running security scanners against your own application is the only thing that matters. Unless you have forked spring-framework and you are re-building it internally?

ivonaest:
We are using it internally but I am just contributing to open-source projects by remediating CVEs at the moment.

PR #35958, opened 2025-12-03
Introduce `MultiValueMapCollector` for use with streams

florianhof:
Stream's collector for building a MultiValueMap. 

Mostly copied from spring-data-commons, see https://github.com/spring-projects/spring-data-commons/issues/3420

I add the type parameters description, the static method `indexingBy` and the unit test.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/MultiValueMapCollector.java
+++ b/spring-core/src/main/java/org/springframework/util/MultiValueMapCollector.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *	  https://www.apache.org/licenses/LICENSE-2.0
+ *      https://www.apache.org/licenses/LICENSE-2.0
  *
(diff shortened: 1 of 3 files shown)


florianhof:
Try to fix the the style problems...
I don't understand the problem of missing _Signed-off-by_ because my 2 commits have them.
Would it be ok if I backport it in the 6.2.x branch?

sbrannen:
> I don't understand the problem of missing _Signed-off-by_ because my 2 commits have them.

I set the DCO to pass. 👍 

> Would it be ok if I backport it in the 6.2.x branch?

Since we are now very late in the 6.2.x line, we typically do not introduce new features there. So I'll process this for `main` (7.0.2).

sbrannen:
This has been merged into `main` in bfc02cda0c1cf7daa2a77a8fd5e65aefd4750505 and revised in 3731fed4cabed534c9e52e36f1d8f30c6ae25238.

Thanks 👍

PR #35961, opened 2025-12-04
Upgrade apache derby, derbyclient and derbytools to 10.17.1.0 

ivonaest:
## Summary
 
1. Why:
To remove CVEs:
     - [CVE-2022-46337](https://avd.aquasec.com/nvd/2022/cve-2022-46337/) is a vulnerability that lets attackers bypass LDAP login checks using specifically crafted username which gives them access to the database where they can run database functions without proper authorisation.

2. What:

     - Upgrade  `derby`, `derbyclient` and `derbytools` to 10.17.1.0 to remove [CVE-2022-46337](https://avd.aquasec.com/nvd/2022/cve-2022-46337/)
 
 ## Additional evidence
 
 Partial output from security scanner Trivy:
spring-framework cves derby

 ## Categorization

- [x] security/CVE


bclozel:
Thanks but we'd rather deal with those updates ourselves. In #35924, I initially thought that this CVE warning was reported in your application, not against our own build. As those are compile-only dependency and never published as dependencies to our POMs, CVEs should not be the main motivation but rather compatibility.

ivonaest:
@bclozel I understand. I'll let you handle these CVEs then. Please feel free to reach out if you'll need any support with future CVE handling.

Issue #35966, opened 2025-12-05
RestClient cannot make HEAD requests when the response declares gzip Content-Encoding


shartte:
**Versions:**
spring-web: 7.0.1
spring-boot: 4.0.0

**Reproducer:**
```
RestClient.create().head()
    .uri("https://maven.neoforged.net/releases/net/neoforged/neoforge/0.25w14craftmine.3-beta/neoforge-0.25w14craftmine.3-beta.pom")
    .retrieve()
    .toBodilessEntity();
```

This code will crash with the stack-trace shown further down below.

The custom decompressing body handler doesn't seem to handle responses with `Content-Encoding: gzip`, but without a response body. This is the case for responses to `HEAD` requests. `GET` requests will work normally, even when using `toBodilessEntity()`.

What I don't know is if the body handler should generally handle 0-byte responses (outside of HEAD) as empty content, even with that encoding.

**Stacktrace:**
```
org.springframework.web.client.ResourceAccessException: I/O error on HEAD request for "https://maven.neoforged.net/releases/net/neoforged/neoforge/0.25w14craftmine.3-beta/neoforge-0.25w14craftmine.3-beta.pom": null
	at org.springframework.web.client.DefaultRestClient$DefaultRequestBodyUriSpec.createResourceAccessException(DefaultRestClient.java:763) ~[spring-web-7.0.1.jar:7.0.1]
	at org.springframework.web.client.DefaultRestClient$DefaultRequestBodyUriSpec.exchangeInternal(DefaultRestClient.java:615) ~[spring-web-7.0.1.jar:7.0.1]
	at org.springframework.web.client.DefaultRestClient$DefaultRequestBodyUriSpec.exchange(DefaultRestClient.java:567) ~[spring-web-7.0.1.jar:7.0.1]
	at org.springframework.web.client.RestClient$RequestHeadersSpec.exchange(RestClient.java:750) ~[spring-web-7.0.1.jar:7.0.1]
	at org.springframework.web.client.DefaultRestClient$DefaultResponseSpec.executeAndExtract(DefaultRestClient.java:896) ~[spring-web-7.0.1.jar:7.0.1]
	at org.springframework.web.client.DefaultRestClient$DefaultResponseSpec.toBodilessEntity(DefaultRestClient.java:860) ~[spring-web-7.0.1.jar:7.0.1]
[...]

PR #35968, opened 2025-12-05
Fix gzip handling for HEAD requests in JdkClientHttpRequest (Issue #35966)

AtharvUrunkar:
This PR fixes a bug where making a HEAD request through Spring’s RestClient fails when the response contains:

Content-Encoding: gzip

An empty body (Content-Length: 0)

Previously, Spring attempted to wrap the empty body in a GZIPInputStream, causing an exception.

Fix implemented

Skip gzip/deflate decompression when:

The HTTP method is HEAD

OR the response declares Content-Length: 0

This prevents GZIPInputStream from being created for empty bodies.

Tests

Added a new test:

headRequestWithGzipContentEncodingShouldNotFail

This test verifies that gzip headers on a HEAD response with an empty body no longer cause exceptions.

Linked issue

Fixes: [https://github.com/spring-projects/spring-framework/issues/35966](https://github.com/spring-projects/spring-framework/issues/35966?utm_source=chatgpt.com)

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
+++ b/spring-web/src/main/java/org/springframework/http/client/JdkClientHttpRequest.java
@@ -113,7 +113,7 @@ class JdkClientHttpRequest extends AbstractStreamingClientHttpRequest {
 		TimeoutHandler timeoutHandler = null;
 		try {
 			HttpRequest request = buildRequest(headers, body);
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal @AtharvUrunkar but I went with a different approach in 12c3dc0. Cheers!

PR #35978, opened 2025-12-08
SubscriberInputStream#resume misuses parked thread reference

izeye:
This PR tries to fix the `SubscriberInputStream.resume()`.

See gh-35468


bclozel:
Thanks @izeye !

PR #35992, opened 2025-12-10
Add ThrowingRunnable alongside ThrowingConsumer et al

hosamaly:
A need often arises for a version of `ThrowingConsumer` and `ThrowingFunction` that can be used as a `Runnable`.

This merge request introduces a subclass of `Runnable` that is a copy of `ThrowingConsumer` with the necessary adaptations.


bclozel:
Thanks for the suggestion @hosamaly but we don't have much use for this in the Spring Framework codebase. We'd rather add types and utilities only if they are being used in Framework, because as a new public type, this adds maintenance work to the team for no immediate benefit to the project.

PR #35993, opened 2025-12-10
Polish async test by removing sleep in DefaultAsyncServerResponseTests

hydrationn:
This PR removes Thread.sleep from async tests in DefaultAsyncServerResponseTests
and replaces it with a non-blocking CompletableFuture pattern. This makes the
test more deterministic and avoids timing-based flakiness.

### 🔍 What was changed
- Removed `Thread.sleep(500)` inside `CompletableFuture.supplyAsync`
- Replaced it with a non-blocking test pattern:
  - Created an incomplete `CompletableFuture`
  - Completed it asynchronously using `new Thread(() -> future.complete(...))`
- This makes the test more deterministic and avoids relying on timing-based behavior

### 💡 Why this change
Using `Thread.sleep` in tests is generally discouraged because:
- It increases test execution time
- It may lead to unstable test results depending on system timing
- It does not reflect real asynchronous behavior

The updated approach ensures:
- Faster test execution
- More reliable async behavior
- Clearer test intent

### 📌 Notes
No functional changes were made to production code.  
This update improves the internal test quality and stability.

### DIFF
--- a/spring-webmvc/src/test/java/org/springframework/web/servlet/function/DefaultAsyncServerResponseTests.java
+++ b/spring-webmvc/src/test/java/org/springframework/web/servlet/function/DefaultAsyncServerResponseTests.java
@@ -39,16 +39,10 @@ class DefaultAsyncServerResponseTests {
 	@Test
 	void blockNotCompleted() {
 		ServerResponse wrappee = ServerResponse.ok().build();
-		CompletableFuture<ServerResponse> future = CompletableFuture.supplyAsync(() -> {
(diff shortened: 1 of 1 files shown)


bclozel:
I don't think this change really improves things.
This change starts a new Thread that immediately completes the future. I don't understand why a new Thread should be started here, and completing immediately the future means that the tested component will never wait for the value since it will be present from the start.

We might be able to improve tests here, but this change doesn't look right to me.

PR #36004, opened 2025-12-12
Provide request-scoped cache support for WebFlux coroutines

Brack93:
## Background
Coroutines support for `@Cacheable` was introduced in [466c8d8](https://github.com/spring-projects/spring-framework/commit/466c8d8f23a982cb53f1cfad0143afd28e9c27ab#diff-d4fad4bb92c9b2ba887c9f83cb36b39a1d60737b6aa4f16d34db8c773789f21a) and refined in [6becfe2](https://github.com/spring-projects/spring-framework/commit/6becfe2508eeb576aa220c20a412edb002f4cab4#diff-d4fad4bb92c9b2ba887c9f83cb36b39a1d60737b6aa4f16d34db8c773789f21a) (following Spring AOP coroutines support in [c8169e5](https://github.com/spring-projects/spring-framework/commit/c8169e5cad0bd616e251aa9c004eb2cdca081a59#diff-20c0a3a0cd86c44481294464284ee232d64132a32ea2227243ebf5bc6959c7de)).

While this effectively solves long-lived cache use cases for suspend methods, it does not easily translate to request-scoped caching.

Although there is no direct support for standard methods as well, request-scoped caching is typically achieved by defining a `@RequestScope`  `CacheManager` bean. However, since `@RequestScope` mechanism relies on thread local, the same approach is not suitable for coroutines, that can suspend and resume on different threads.

For a coroutine-based WebFlux server, a potential workaround involves explicitly adding to the request coroutine context the required resolved data (or a lazy resolver), using a custom coroutine context element, or manually passing it down the call stack as a method argument. However, this explicit approach strictly ties the implementation of a service - that may need this shared data at a deep request execution level - with the request context. It may also require a huge refactor to pass all the way down the required shared data, if the caching need emerges late in development. Additionally, for Kotlin WebFlux based GraphQL frameworks, clients can batch independent queries within the same request to save network resources, preventing an explicit design on server side that is able to cache common data for each possible queries combination.

For a better separation of concerns, we might prefer instead to inject the data provider as a service dependency, but still avoid duplicated executions within the same request.

## Proposal
This PR provides request-scoped, proxy-based, cache management capability to Spring WebFlux servers leveraging Kotlin coroutines.

Specifically, suspend methods of Spring beans can now be annotated with `@CoRequestCacheable`. 
```kotlin
class MyServiceBean {

	@CoRequestCacheable(key = "#userName")
	suspend fun fetchUserAgeFromDownstreamService(userName: String, authHeader: String): Int {
		// prepare request and fetch the user info
		return userInfo.age
	}

}
```
The cached result is tied to the web request coroutine context, ensuring that each new incoming web request triggers a new execution of the annotated method.

The required infrastructure can be enabled by using `@EnableCoRequestCaching` on a configuration class.
```kotlin
@Configuration
@EnableCoRequestCaching
class AppConfig {

	@Bean
	fun myService(): MyService {
		// configure and return a class having @CoRequestCacheable suspend methods
		return MyService()
	}

}
```

Although the proposed implementation could not fit in the already existing `@Cacheable` infrastructure, individual components have been reused or extended whenever possible to maintain consistency.

### DIFF
--- a/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/cache/CoRequestCacheable.kt
+++ b/spring-webflux/src/main/kotlin/org/springframework/web/reactive/function/server/cache/CoRequestCacheable.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2002-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.web.reactive.function.server.cache
+
+/**
+ * Annotation indicating that the result of invoking a *suspend method* can be cached
+ * for the lifespan of the underlying web request coroutine.
+ *
+ * Example:
+ *
+ * ```
+ * class MyServiceBean {
+ *
+ * 	@CoRequestCacheable(key = "#userName")
+ * 	suspend fun fetchUserAgeFromDownstreamService(userName: String, authHeader: String): Int {
+ * 		// prepare request and fetch the user info
+ * 		return userInfo.age
+ * 	}
+ *
+ * }
+ * ```
+ *
+ * Each time an advised suspend method is invoked, caching behavior will be applied,
+ * checking whether the method has been already invoked for the given arguments *within the same web request execution*.
+ * A sensible default simply uses the method parameters to compute the key, but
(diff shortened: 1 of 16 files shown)


sdeleuze:
Thanks for this PR, but we prefer to decline it as it is unclear how useful would be such request-scoped caching, as we typically build Coroutines features on top of Reactive ones, and as this kind of `@CoRequestCacheable` annotation does not seems a good fit in Spring programming model.

Brack93:
Thanks for your feedback @sdeleuze.

This PR was inspired by the following kotlin-based open source GraphQL framework, in particular to support something similar to what dataloaders achieve, avoiding duplicated network calls within the same GraphQL request when it includes multiple queries all requiring the same data, but without the need to switch to Futures:

https://opensource.expediagroup.com/graphql-kotlin/docs/server/data-loader/#dataloaders-and-coroutines

Since the above framework is based on Spring Webflux, I thougth it could fit and maybe be useful in general when using Spring WebFlux with Kotlin. Thanks again for the feedback.

sdeleuze:
Thanks for the additional context and your understanding, I guess my recommendation would be to implement that logic in a programmatic way instead of declarative with annotations. Looks too opinionated for the annotation-based programming model.

Issue #35906, opened 2025-11-26
Include "since" info in generated Javadoc for new and deprecated APIs


sbrannen:
## Overview

Recent versions of the `javadoc` command support [`-since`](https://docs.oracle.com/en/java/javase/24/docs/specs/man/javadoc.html#option-since) and [`-since-label`](https://docs.oracle.com/en/java/javase/24/docs/specs/man/javadoc.html#option-since-label) options which allow `@since` and `@Deprecated(since=...)` information to be included in the generated Javadoc HTML.

## Examples for JDK 24

- [New APIs](https://docs.oracle.com/en/java/javase/24/docs/api/new-list.html)
- [Deprecated APIs](https://docs.oracle.com/en/java/javase/24/docs/api/deprecated-list.html)

## Examples for Spring Framework.

Screenshots taken from local build.

<img width="822" height="499" alt="Image" src="https://github.com/user-attachments/assets/c680d3b1-956e-468d-ade7-e85e9d7fa0e8" />

<img width="825" height="477" alt="Image" src="https://github.com/user-attachments/assets/52da8fb0-cc39-4ec8-bbf5-4d5d8cc4fcca" />

sbrannen:
To try this out for yourself in a local build, use this branch.

https://github.com/spring-projects/spring-framework/compare/main...sbrannen:spring-framework:issues/gh-35906-javadoc-since-new-and-deprecated-api-info

sbrannen:
FWIW, we already have a [Deprecated APIs](https://docs.spring.io/spring-framework/docs/7.0.1/javadoc-api/deprecated-list.html) page included in the Javadoc, but you cannot filter by "since" by default.

sdeleuze:
The `6.2` part is confusing, if we enable this feature we should move `@since` tags from `@since 6.2` to `@since 6.2.0` (and same for other majors).

I am also puzzled by the too many releases displayed, the UI looks pretty bad with that amount of versions. If wildcards are not supported, maybe we should focus on the latest major.

sbrannen:
> The `6.2` part is confusing, if we enable this feature we should move `@since` tags from `@since 6.2` to `@since 6.2.0` (and same for other majors).

That's certainly worth considering.

> I am also puzzled by the too many releases displayed, the UI looks pretty bad with that amount of versions. If wildcards are not supported, maybe we should focus on the latest major.

I merely selected all of those manually as an experiment and to show how the feature works.

[...]

PR #36006, opened 2025-12-12
Add clarifying Note for limitation issue #35906

albonidrizi:
Issue: #35906

## Description
This PR adds a clarifying [NOTE] in the TestContext Framework documentation (`fixture-di.adoc`) regarding the usage of `@LocalServerPort`.

It highlights that `@LocalServerPort` (from Spring Boot) may not work correctly when used with nested test classes and inheritance, as discussed in issue #35906.

Location: `framework-docs/modules/ROOT/pages/testing/testcontext-framework/fixture-di.adoc`


bclozel:
Obvious hallucination. This is mixing random subjects with correlation at all.

Issue #35942, opened 2025-12-01
Problem with Generic typing extending AbstractHttpMessageConverter<List<XXX>


cedricbarbin:
I try to extend `AbstractHttpMessageConverter` and my return type is a `List`. I'd like to have different converters depending on the object type of my list (Generic typing).

Super class:

```java
public abstract class PrometheusCommonHttpMessageConverter<T> extends AbstractHttpMessageConverter<List<T>> {
...
}
```

Implementation:

```java
public class PrometheusMonitoringHttpMessageConverter extends PrometheusCommonHttpMessageConverter<MonitoringLineDTO> {
...
    @Override
    protected boolean supports(Class<?> clazz) {
        return ???   // Here if I test with List.class it always return true, no matter which generic typing is used
    }
...
}

```

I couldn't make it works because reflection and generic typing are not compatible.

Can you think of a way to make it possible?

I did it by modifying `AbstractMessageConverterMethodProcessor`... but my way is a workaround, compatible for my specific project and I'm pretty sure you can find a better way to do it.

My workaround:

```java
public class PrometheusMonitoringHttpMessageConverter extends PrometheusCommonHttpMessageConverter<MonitoringLineDTO> {
...
    @Override
    protected boolean supports(Class<?> clazz) {
        return MonitoringLineDTO.class.isAssignableFrom(clazz);
    }
...
}
```

And (the idea was to send the class of the objects in the list) :

```java
protected <T> void writeWithMessageConverters(@Nullable T value, MethodParameter returnType, ServletServerHttpRequest inputMessage, ServletServerHttpResponse outputMessage) throws IOException, HttpMediaTypeNotAcceptableException, HttpMessageNotWritableException {
...
                //Patch
                //If List type, use a converter only a data is present and override class type with an element of the list
                else if (List.class.isAssignableFrom(valueType)) {
                   if(value instanceof List<?> && !((List) value).isEmpty() && converter.canWrite(((List) value).get(0).getClass(), selectedMediaType)) {
                        converterTypeToUse = ConverterType.BASE;
                    }
                //Patch end
                } else if(converter.canWrite(valueType, selectedMediaType)) {
                    converterTypeToUse = ConverterType.BASE;
                }
...
}
```

PaulNgo-BlueOC:
Please use `AbstractGenericHttpMessageConverter` and override those methods
[...]

PR #36007, opened 2025-12-12
Document limitation of multiple @BootstrapWith instances

albonidrizi:
Issue: #35942

## Description
This PR adds a clarifying note to the `@BootstrapWith` Javadoc.

It documents that multiple instances of this annotation on the same test class are not supported, and caution should be exercised when using nested test classes or inheritance to avoid unexpected behavior.


bclozel:
Obvious hallucination. This is mixing random subjects with no correlation at all.

PR #36009, opened 2025-12-12
Document semantics for combining @Retryable with other proxy-based features (#35586)

albonidrizi:
Issue: #35586

## Description
Adds a clarifying note in the documentation about combining `@Retryable` with other proxy-based features to ensure correct ordering of proxies and expected retry behavior.

Location: `framework-docs/modules/ROOT/pages/testing/testcontext-framework/tx.adoc`


bclozel:
Suggested changes are wrong and will break documentation rendering. This also contains unrelated changes.

PR #36011, opened 2025-12-12
Revamp documentation to prioritize Java/Annotation-Based configuration over XML (#35446)

albonidrizi:
Issue: #35446

## Description
This PR adds a recommendation note to the "Configuration Metadata" section in `core/beans/basics.adoc`.

It explicitly recommends Java/Annotation-based configuration over XML for new projects, highlighting benefits like type safety and better IDE support, while noting that XML remains supported for legacy scenarios.


sdeleuze:
This PR is pretty far from what we want to achieve, and the related issue is already assigned, meaning I plan to take care. Also this Copilot review bring more noise that help. As a consequence, I will decline this PR.

PR #36018, opened 2025-12-13
Extend and improve some test cases

vincentstradiot:
Hi, here is a Pull Request that extends and improves some test cases.

- In the `CollectionUtilsTest`, it seems important to me to run each case/assert with a fresh enumeration. Otherwise, the test cases are not independent as we might already have enumerated some elements in case 1 when running case 2.
- In general, I extended test cases to be more exhaustive and cover uncovered lines/branches/functionalities.


bclozel:
Thanks for the proposal, but this PR is now stale and spreads changes over multiple classes, not focusing on a concrete issue or bug. I'll decline this for now.

PR #36021, opened 2025-12-13
SimpleAliasRegistry: extend test cases + small fix around alias overriding

vincentstradiot:
Hi, here is a Pull Request that:

1. extends the existing `SimpleAliasRegistryTests` to cover the `allowAliasOverriding` functionality.
2. updates the implementation to fix a corner case issue in the `SimpleAliasRegistry` (explanation below).

**Old behaviour was:**
- _register alias A -> B:_ :white_check_mark: OK
- _register alias A -> A_: :white_check_mark: OK
   - ⚠️ But alias A -> B was gone, which is not the expected behaviour as alias A -> B was overridden.

To my opinion, this should only be allowed if `allowAliasOverriding = true`, in exactly the same way as the existing implementation would only allow registering alias [A -> C] if `allowAliasOverriding = true`.

**New behaviour:**
- _register alias A -> B_: OK :white_check_mark:
- _register alias A -> A_:
   - :white_check_mark: OK if `allowAliasOverriding = true`
   - :x: Otherwise throws exception

**Behaviour covered the following test cases in this PR:** 
- `SimpleAliasRegistryTests.registerAliasWithSameName_GivenAliasOverridingAllowed_ThenReplacesExistingAlias `
- `SimpleAliasRegistryTests.registerAliasWithSameName_GivenAliasOverridingNotAllowed_ThenThrowsException`

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/SimpleAliasRegistry.java
+++ b/spring-core/src/main/java/org/springframework/core/SimpleAliasRegistry.java
@@ -59,6 +59,11 @@ public class SimpleAliasRegistry implements AliasRegistry {
 		Assert.hasText(alias, "'alias' must not be empty");
 		synchronized (this.aliasMap) {
 			if (alias.equals(name)) {
+				String registeredName = this.aliasMap.get(alias);
+				if (registeredName != null && !registeredName.equals(name) && !allowAliasOverriding()) {
+					throw new IllegalStateException("Cannot define alias '" + alias + "' for name '" +
+							name + "': It is already registered for name '" + registeredName + "'.");
+				}
 				this.aliasMap.remove(alias);
 				this.aliasNames.remove(alias);
 				if (logger.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


sbrannen:
Hi @vincentstradiot,

Congratulations on submitting your first PR for the Spring Framework! 👍🏻

And sorry for taking so long to get back to you.

I took a look at your proposal, and unfortunately I don't think the implementation change is correct. The unconditional removal in the `alias.equals(name)` branch is long-standing, intentional behavior, not a bug: `resolveAliases()` handles the analogous self-reference case (a placeholder resolving an alias to its own registered name) the same way, and `removeAlias()` never checks `allowAliasOverriding()` either. The consistent design principle here is that override permission gates *redirecting* an alias to a _different_ target, not *removing* an existing one. So, changing this case would make `registerAlias()` inconsistent with `resolveAliases()` for the same underlying scenario, and could break existing usage that relies on the current tolerant handling.

That said, three of your new tests are valuable independent of the implementation change, since they cover paths in `registerAlias()` that currently have no coverage at all:

- `overrideAlreadyExistingAlias_GivenAliasOverridingAllowed_ThenReplacesExistingAlias`
- `overrideAlreadyExistingAlias_GivenAliasOverridingNotAllowed_ThenThrowsException`
- `registerSameAliasTwice_DoesNothing`

Would you be open to submitting a new, smaller PR with just those three tests? If so, I'd be happy to merge that.

In light of the above, I'm closing this PR, but thanks again for the effort.

Cheers,

Sam

PR #36028, opened 2025-12-15
StopWatchTests: improve test cases

vincentstradiot:
Hi, here is a Pull Request that improves the test cases for `StopWatch` by:

- restoring and replacing flaky asserts by a deterministic implementation
- removing the need of running `Thread.sleep` in unit tests

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/StopWatch.java
+++ b/spring-core/src/main/java/org/springframework/util/StopWatch.java
@@ -76,7 +76,7 @@ public class StopWatch {
 	private long totalTimeNanos;
 
 	/**
-	 * The {@link NanoClock} to compute execution times of tasks.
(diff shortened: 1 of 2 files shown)


bclozel:
Test hasn't been flaky over the last 6 months.

image

vincentstradiot:
@bclozel They aren't flaky because all the asserts are commented out on the main branch, which is not ideal (see image below).
The purpose of the PR was to uncomment these, to actually assert meaningful things in the tests.

Up to you to decide. Let met know if you want me to reactivate/recreate this PR.

Screenshot 2025-12-17 115038

bclozel:
I don't think it's worth it in this current form. We could consider completely rewriting the class and its tests but it's definitely not our priority. Thanks for the suggestion.

Issue #36030, opened 2025-12-16
Replace String concatenation with StringBuilder in BatchUpdateException handling


jher235:
In `JdbcTemplate.batchUpdate(String... sql)`, I noticed that String concatenation is used inside the `BatchUpdateException` catch block to construct the error message.

## Current Implementation

```java

catch (BatchUpdateException ex) {
    String batchExceptionSql = null;
    for (int i = 0; i < ex.getUpdateCounts().length; i++) {
        if (ex.getUpdateCounts()[i] == Statement.EXECUTE_FAILED) {
            // String concatenation in loop
            batchExceptionSql = appendSql(batchExceptionSql, sql[i]);
        }
    }
    if (StringUtils.hasLength(batchExceptionSql)) {
        this.currSql = batchExceptionSql;
    }
    throw ex;
}

```

## Issue

Using String concatenation in a loop creates unnecessary temporary objects. While this happens in the exception path, using `StringBuilder` is a standard optimization practice in Java.

## Proposed Solution

Replace the concatenation logic with `StringBuilder`:

```java
catch (BatchUpdateException ex) {
    int[] updateCounts = ex.getUpdateCounts();
    StringBuilder failedSql = new StringBuilder();

    int limit = Math.min(updateCounts.length, sql.length);
    for (int i = 0; i < limit; i++) {
        if (updateCounts[i] == Statement.EXECUTE_FAILED) {
            if (failedSql.length() > 0) {
                failedSql.append("; ");
            }
            failedSql.append(sql[i]);
        }
    }

    if (failedSql.length() > 0) {
        this.currSql = failedSql.toString();
    }
    throw ex;
}

```

## Benefits

- Reduces memory allocation during batch failures
- Follows standard Java best practices
- No functional changes / Fully backward compatible
- Handles edge case where `updateCounts.length < sql.length`

jher235:
If this approach looks reasonable, I can submit a small PR with this change.
Happy to adjust based on any feedback.

bclozel:
Please submit a PR. I'll close this issue as superseded by your upcoming PR.

PR #36032, opened 2025-12-16
Use StringBuilder in JdbcTemplate for batch updates

jher235:
### Description
As discussed in issue #36030, this PR refactors the exception handling logic within
`JdbcTemplate.batchUpdate(String... sql)`.

Currently, String concatenation is used inside a loop when constructing the error
message for a `BatchUpdateException`. This PR replaces it with `StringBuilder` to
avoid unnecessary temporary object allocation in the failure path.

### Changes
- **Replaced String concatenation with `StringBuilder`**: avoids unnecessary temporary
  object allocation for loop-based SQL accumulation.
- **Added Safety Check**: uses `Math.min(updateCounts.length, sql.length)` to safely
  handle edge cases where JDBC drivers may return fewer update counts than the number
  of SQL statements.
- **Preserved Existing Behavior**: the happy path logic remains unchanged, and the
  resulting error message format is identical to the previous implementation.

### Verification
Verified by the existing CI test suite.


bclozel:
Thanks for your contribution @jher235 this is now merged in main.

PR #36049, opened 2025-12-19
gh-36043 - configureMessageConvertersList

### LINKED ISSUE #36043: Adding a HttpMessageConverter to the end was possible with deprecated extendMessageConverters(…) but not with its replacement configureMessageConverters(…)
With `WebMvcConfigurer.extendMessageConverters(List<HttpMessageConverter<?>> converters)` it was possible to add a custom HttpMessageConverter to the front, back or anywhere in the middle. This method has been deprecated and the documented replacement is `WebMvcConfigurer.configureMessageConverters(HttpMessageConverters.ServerBuilder builder)`.

However, in `configureMessageConverters` the builder only offers one method `addCustomConverter` which always adds the custom converter to the front of the list. In my application I have a custom HttpMessageConverter that supports `MediaType.ALL` and `canWrite(Class<?> clazz, @Nullable MediaType mediaType)` returns true if the clazz is an instance of my custom error object. This converter is supposed to be a fallback in case the other converters failed to handle the response which can happen if the client sends an accept-header that only accepts unusual media types.

This is only supposed to be a fallback and therefore should be the last converter in the list. Otherwise it will handle the error objects before, for example, JacksonJsonHttpMessageConverter gets a chance to handle it.

How do I achieve this in a non-deprecated way?
There was already a feature request for this in issue https://github.com/spring-projects/spring-framework/issues/35177 which was closed after implementing the current behavior. However, the current behavior does not actually allow ordering the message converters. Can this be added as an enhancement to the ServerBuilder?

haydenrear:
Closes gh-36043                               

Adds ability to change message converters list before it's returned by registering consumer using configureMessageConvertersList

Signed-off-by: Hayden Rear hayden.rear@gmail.com

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/DefaultHttpMessageConverters.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/DefaultHttpMessageConverters.java
@@ -122,6 +122,8 @@ class DefaultHttpMessageConverters implements HttpMessageConverters {
 
 		@Nullable Consumer<HttpMessageConverter<?>> configurer;
 
+		@Nullable Consumer<List<HttpMessageConverter<?>>> convertersListConfigurer;
+
 		@Nullable HttpMessageConverter<?> kotlinJsonConverter;
 
 		@Nullable HttpMessageConverter<?> jsonConverter;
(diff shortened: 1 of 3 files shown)


bclozel:
@haydenrear thanks for this PR, but we cannot process it at the moment.
Can you:
* squash and sign the commits into one?
* run the build locally and ensure it passes?

Thanks

haydenrear:
> @haydenrear thanks for this PR, but we cannot process it at the moment. Can you:
> 
> * squash and sign the commits into one?
> * run the build locally and ensure it passes?
> 
> Thanks

Hi @bclozel 

I have squashed and ran the build locally and pushed.

Thanks!
Hayden

bclozel:
Thanks @haydenrear but it seems that [your commits aren't signed with the DCO](https://github.com/spring-projects/spring-framework/pull/36049/checks?check_run_id=59123982418) (even though it looks like they are in the commit message). At this stage, I'm not sure how we will resolve #36043 so I'll close this PR to avoid spending time processing a PR that we cannot merge.

PR #36054, opened 2025-12-19
Update to NullAway 0.12.15 and fix new warnings

msridhar:
The new warnings are a consequence of https://github.com/uber/NullAway/pull/1371, which better detects annotations on type arguments in return types.  

Fortunately, the fixes are straightforward.  Note that writing `@Nullable Void` is required by JSpecify (see https://github.com/jspecify/jspecify/issues/51).  This is not yet checked exhaustively by NullAway, but some instances were caught here.

### DIFF
--- a/gradle/spring-module.gradle
+++ b/gradle/spring-module.gradle
@@ -105,6 +105,10 @@ tasks.register('javadocJar', Jar) {
 	from javadoc
 }
 
+nullability {
+    nullAwayVersion = "0.12.15"
+}
+
 publishing {
 	publications {
 		mavenJava(MavenPublication) {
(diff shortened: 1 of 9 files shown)


msridhar:
FYI @sdeleuze

snicoll:
I don't think we should upgrade just yet even if this works for framework, see https://github.com/spring-gradle-plugins/nullability-plugin/issues/25

msridhar:
Hi @snicoll thanks, I commented on the NullAway issue you opened.  It might be reasonable to change the nullability-plugin to not enable `RequireExplicitNullMarking` checking by default, if it's too surprising for users; people could opt in to it instead.  But I think that's independent of the NullAway 0.12.15 release.

For context, I'm interested in getting Spring Framework upgraded to NullAway 0.12.15 since we use it as an integration test for NullAway, and that integration test has been failing due to the new warnings that this PR fixes.  Nothing critical, though.

sdeleuze:
Thanks @msridhar, the code changes LGTM, but I will wait the release of https://github.com/spring-gradle-plugins/nullability-plugin/milestone/9 to merge it instead of the NullAway version override.

I have tagged this issue as en enhancement as it will refine the nullability of some of our public APIs.

sdeleuze:
@rstoyanchev While changing how we check `@NullMarked` packages to leverage https://github.com/spring-gradle-plugins/nullability-plugin/issues/23, I found that the `org.springframework.test.web.support` package was not annotated, so I changed that. Please give it a look to verify everything looks ok from a nullability perspective.

PR #36056, opened 2025-12-19
Fix case-insensitive semantics for LinkedCaseInsensitiveMap entrySet

TerryTaoYY:
### Summary
Align `LinkedCaseInsensitiveMap`’s `entrySet()` view with the map’s case-insensitive key contract by making `entrySet().contains(..)` and `entrySet().remove(..)` perform case-insensitive key resolution.

### Background
`LinkedCaseInsensitiveMap` provides case-insensitive behavior for `containsKey/get/remove`. However, `entrySet()` operations previously delegated to the backing map’s entry set, which performs case-sensitive key comparisons. As a result, entry-based operations could miss existing mappings when callers used different key casing.

### Example
```java
LinkedCaseInsensitiveMap<String> map = new LinkedCaseInsensitiveMap<>();
map.put("Key", "value");

assertThat(map.containsKey("KEY")).isTrue();
assertThat(map.entrySet()).contains(Map.entry("KEY", "value")); // previously failed
assertThat(map.entrySet().remove(Map.entry("key", "value"))).isTrue();
assertThat(map).isEmpty();
```

### Changes
- Implement case-insensitive key lookup for EntrySet.contains(Object) and EntrySet.remove(Object) via containsKey/get/remove
- Preserve Map.Entry semantics by requiring value equality (using ObjectUtils.nullSafeEquals) before removing

### Tests
- `./gradlew :spring-core:test --tests org.springframework.util.LinkedCaseInsensitiveMapTests`
- `./gradlew :spring-core:test`

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
+++ b/spring-core/src/main/java/org/springframework/util/LinkedCaseInsensitiveMap.java
@@ -448,7 +448,18 @@ public class LinkedCaseInsensitiveMap<V> implements Map<String, V>, Serializable
 
 		@Override
 		public boolean contains(Object o) {
-			return this.delegate.contains(o);
+			if (!(o instanceof Map.Entry<?, ?> entry)) {
+				return false;
+			}
+			Object key = entry.getKey();
+			if (!(key instanceof String stringKey)) {
+				return false;
+			}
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for your contribution @TerryTaoYY

PR #36062, opened 2025-12-21
Revert change to JUnit's Javadoc URL

marcphilipp:
Now that `current` redirects to `6.0.1` rather than the other way around.

Sorry for breaking this again. Shouldn't happen again!


bclozel:
Thanks @marcphilipp , I just noticed this has been fixed already last night with fc2e1dd2e5565a9

Issue #36061, opened 2025-12-21
Wrong exception handling for @RequestParam of type boolean


mzatko:
If an invalid value is passed to @RequestParam of type boolean, an IllegalArgumentException is thrown, which is indistinguishable from other IllegalArgumentException exceptions.
This then leads to a 500 InternalServerError, which is clearly incorrect, as it is a clear example of a 400(BadRequest) error.

example
```java
@GetMapping
ResponseEntity<Void> getException(@RequestParam(name = "bool", defaultValue = "false", required = false) boolean bool) {
   ...
}
```

then pasing GET http://server:port/?bool=dummy

causes 500 internalServerError

Gautam-aman:
Hi, I’d like to work on this issue.

I plan to add a failing MockMvc test demonstrating that an invalid boolean
@RequestParam currently results in a 500 response, and then update the
argument resolution logic so it correctly returns 400 Bad Request.

Please let me know if there’s any preferred approach before I open a PR.

bclozel:
@mzatko can you share a minimal sample? I cannot reproduce the behavior you are describing. Please find my sample attached.

[illegalarg.zip](https://github.com/user-attachments/files/24314145/illegalarg.zip)

bclozel:
@Gautam-aman it looks like you went ahead without confirming the problem first. I'll close the other issue and the 2 PRs as we're not sure we need anything at the moment.

mzatko:
@bclozel I had "deeper" look at the problem, it looks like it is `@ExceptionHandler` problem. In fact, there is `@MethodArgumentTypeMismatchException` thrown however I had no `@ExceptionHandler(MethodArgumentTypeMismatchException.class)` defined so exception was handled by `@ExceptionHandler(IllegalArgumentException.class)` as it is a cause of `@MethodArgumentTypeMismatchException`. `@MethodArgumentTypeMismatchException` is cut out completely from "causes chain" and ignored as if never existed. It is really strange behavior of `@ExceptionHanler`
I am closing this issue.

PR #36066, opened 2025-12-22
Fix wrong exception handling for boolean @RequestParam

Gautam-aman:
### Issue
Invalid boolean values supplied to `@RequestParam` could result in
`IllegalArgumentException` escaping argument resolution, leading to
inconsistent error handling.

### Changes
- Catch and normalize `IllegalArgumentException` during argument conversion
- Re-throw as `MethodArgumentTypeMismatchException`
- Add regression test to `RequestMappingHandlerAdapterTests`

### Result
Boolean request parameter conversion errors are now handled consistently
with other type mismatches.


bclozel:
See https://github.com/spring-projects/spring-framework/issues/36061#issuecomment-3687027499

PR #36068, opened 2025-12-22
Fix primitive boolean @RequestParam handling and invalid value behavior

Gautam-aman:
### Summary

This PR improves Spring MVC request parameter binding for primitive boolean
types in two edge cases:

1. Invalid boolean values for `@RequestParam boolean` now consistently
   result in `MethodArgumentTypeMismatchException`, leading to HTTP 400.
2. Missing primitive boolean parameters declared with
   `@RequestParam(required = false)` now default to `false` instead of
   raising a type mismatch exception.

### Motivation

Previously:
- Invalid boolean values could fail inconsistently.
- Primitive boolean parameters with `required=false` would still raise
  `TypeMismatchException` when missing, despite being optional.

This behavior was surprising and inconsistent with developer expectations.

### Changes

- Updated `AbstractNamedValueMethodArgumentResolver` to correctly handle
  missing primitive boolean parameters when `required=false`.
- Ensured invalid boolean values fail fast with a consistent exception.
- Added regression tests in `RequestMappingHandlerAdapterTests`.

### Impact

- Backward-compatible
- Improves correctness and developer experience
- No public API changes


bclozel:
See https://github.com/spring-projects/spring-framework/issues/36061#issuecomment-3687027499

PR #36075, opened 2025-12-26
Consistently declare `@Nullable` on  parameter in `equals()` implementations

MukjepScarlet:
String found: org.springframework.cglib.core.EmitUtils/KeyFactory
I don't know if they should be changed.

Reason: https://youtrack.jetbrains.com/projects/KT/issues/KT-83314/JSpecify-NullMarked-changes-Java-equalsObject-to-equalsAny-causing-override-conflict-in-Kotlin-2.3

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/AdvisedSupport.java
@@ -731,7 +731,7 @@ public class AdvisedSupport extends ProxyConfig implements Advised {
 		}
 
 		@Override
-		public boolean equals(Object other) {
+		public boolean equals(@Nullable Object other) {
 			return (this == other || (other instanceof AdvisorKeyEntry that &&
 					this.adviceType == that.adviceType &&
 					ObjectUtils.nullSafeEquals(this.classFilterKey, that.classFilterKey) &&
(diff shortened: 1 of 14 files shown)


sbrannen:
> String found: org.springframework.cglib.core.EmitUtils/KeyFactory
> I don't know if they should be changed.

No, those are used to create a `Signature` which is used for bytecode generation.

----

More importantly, why did you add `@Nullable` to types in the `cglib` package and its subpackages?

As far as I can tell, none of those packages are `@NullMarked`, and we have never added nullability declarations to our fork of CGLIB.

Also, why did you introduce `@Nullable` in test classes?

MukjepScarlet:
I used text replacement... That's why there are unneeded replacements (in not `@NullMarked` packages this will be `Any!` in Kotlin). I can revert some of them of course.

sbrannen:
> I used text replacement... That's why there are unneeded replacements (in not `@NullMarked` packages this will be `Any!` in Kotlin). I can revert some of them of course.

Yes, please remove `@Nullable` declarations in packages that are not `@NullMarked`.

MukjepScarlet:
`Nullable` in `cglib` package and test modules are removed now.

sbrannen:

Thanks

PR #36077, opened 2025-12-27
Use String instead of long for JFR event IDs to support JMC filtering.

youngledo:
Because "eventId" and "parentId" are of type "long", they cannot be selected in the context filter menu in "JMC":
image

But sometimes it is powerless when filtering by eventId or parentId is required.

Avoid this problem by modifying the field type:
image

7e5fefce-08dd-4085-b5f5-c1a8e1ebd29e

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/metrics/jfr/FlightRecorderStartupEvent.java
+++ b/spring-core/src/main/java/org/springframework/core/metrics/jfr/FlightRecorderStartupEvent.java
@@ -29,6 +29,7 @@ import jdk.jfr.Label;
  * as a single {@code String}, since Flight Recorder events do not support complex types.
  *
  * @author Brian Clozel
+ * @author Huang Xiao
  * @since 5.3
  */
 @Category("Spring Application")
@@ -36,9 +37,11 @@ import jdk.jfr.Label;
 @Description("Spring Application Startup")
 class FlightRecorderStartupEvent extends Event {
 
-	public final long eventId;
+	@Label("EventId")
+	public final String eventId;
 
-	public final long parentId;
+	@Label("ParentId")
+	public final String parentId;
 
 	@Label("Name")
 	public final String name;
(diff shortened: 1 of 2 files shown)


youngledo:
@bclozel Hello, could you take a look at this issue?

bclozel:
Thanks for the proposal, but it's not clear to me why we would need to turn a `Long` value into a `String` for filtering reasons in the mission control tool. Both `eventId` and `parentId` columns work with natural sorting; maybe there is a reason for not supporting this feature with `Long` values?

Have you tried suggesting supporting this type or asking this to JMC maintainers?

bclozel:
Additional note: what happens if we change to a `String` type in the `eventId` and `parentId` columns? If natural ordering is used, I would think that sorting by id would yield: "1", "10", "11", "2", "3"... Wouldn't this be a regression in the UI? Have you tried your change locally?

youngledo:
> Thanks for the proposal, but it's not clear to me why we would need to turn a `Long` value into a `String` for filtering reasons in the mission control tool. Both `eventId` and `parentId` columns work with natural sorting; maybe there is a reason for not supporting this feature with `Long` values?
> 
> Have you tried suggesting supporting this type or asking this to JMC maintainers?

As I described at the beginning, it appears that JMC’s context menu supports only the String type, not the Number type.

youngledo:
> Additional note: what happens if we change to a `String` type in the `eventId` and `parentId` columns? If natural ordering is used, I would think that sorting by id would yield: "1", "10", "11", "2", "3"... Wouldn't this be a regression in the UI? Have you tried your change locally?

You're right. After changing to String type, natural sorting is indeed not supported anymore, but for these two properties, sorting is not that important. Besides, perhaps we should look at the source code of JMC to see if the context menu should support Number type selection?

bclozel:
Thanks for checking. We won't merge this change because this will cause a regression. I think you should consider this on the JMC side then.

youngledo:
This issue has been resolved via JMC. see: https://github.com/openjdk/jmc/pull/718

PR #36084, opened 2025-12-30
Allow WebFlux ApiVersionResolver to return a Mono

JKaplanEmpty-Nes:
The current WebFlux API resolution returned a nullable String, this prevented custom API resolvers to be based on values which would need to be blocked. 
For example, creating an API version resolver based on the Authorized Principal. Since this is returned as a Mono in a server exchange.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
@@ -17,6 +17,7 @@
 package org.springframework.web.reactive.accept;
 
 import org.jspecify.annotations.Nullable;
+import reactor.core.publisher.Mono;
 
 import org.springframework.web.accept.InvalidApiVersionException;
 import org.springframework.web.accept.MissingApiVersionException;
@@ -27,6 +28,7 @@ import org.springframework.web.server.ServerWebExchange;
  * to manage API versioning for an application.
  *
  * @author Rossen Stoyanchev
+ * @author Jonathan Kaplan
  * @since 7.0
  * @see DefaultApiVersionStrategy
  */
@@ -37,9 +39,23 @@ public interface ApiVersionStrategy {
 	 * @param exchange the current exchange
 	 * @return the version, if present or {@code null}
 	 * @see ApiVersionResolver
+	 * @deprecated as of 7.0.3, in favor of
+	 * {@link #resolveVersionAsync(ServerWebExchange)}
 	 */
+	@Deprecated(forRemoval = true, since = "7.0.3")
 	@Nullable
 	String resolveVersion(ServerWebExchange exchange);
+
+
+	/**
+	 * Resolve the version value from a request asynchronously.
+	 * @param exchange the current server exchange containing the request details
+	 * @return a {@code Mono} emitting the resolved version as a {@code String},
+	 *         or an empty {@code Mono} if no version is resolved
+	 */
+	default Mono<String> resolveVersionAsync(ServerWebExchange exchange) {
+		return Mono.justOrEmpty(this.resolveVersion(exchange));
+	}
 
 	/**
 	 * Parse the version of a request into an Object.
@@ -59,6 +75,16 @@ public interface ApiVersionStrategy {
 	void validateVersion(@Nullable Comparable<?> requestVersion, ServerWebExchange exchange)
 			throws MissingApiVersionException, InvalidApiVersionException;
 
+	private Mono<Comparable<?>> validateVersionAsync(@Nullable Comparable<?> requestVersion, ServerWebExchange exchange) {
+		try {
+			this.validateVersion(requestVersion, exchange);
+			return Mono.justOrEmpty(requestVersion);
+		}
+		catch (MissingApiVersionException | InvalidApiVersionException ex) {
+			return Mono.error(ex);
+		}
+	}
+
 	/**
 	 * Return a default version to use for requests that don't specify one.
 	 */
@@ -70,6 +96,8 @@ public interface ApiVersionStrategy {
 	 * @param exchange the current exchange
 	 * @return the parsed request version, or the default version
 	 */
+	@SuppressWarnings({"DeprecatedIsStillUsed", "DuplicatedCode"})
+	@Deprecated(forRemoval = true, since = "7.0.3")
 	default @Nullable Comparable<?> resolveParseAndValidateVersion(ServerWebExchange exchange) {
 		String value = resolveVersion(exchange);
 		Comparable<?> version;
@@ -86,6 +114,32 @@ public interface ApiVersionStrategy {
 		}
 		validateVersion(version, exchange);
 		return version;
+	}
+
+	/**
+	 * Convenience method to return the API version from the given request exchange, parse and validate
+	 * the version, and return the result as a reactive {@code Mono} stream. If no version
+	 * is resolved, the default version is used.
+	 * @param exchange the current server exchange containing the request details
+	 * @return a {@code Mono} emitting the resolved, parsed, and validated version as a {@code Comparable<?>},
+	 * or an error in case parsing or validation fails
+	 */
+	@SuppressWarnings("Convert2MethodRef")
+	default Mono<Comparable<?>> resolveParseAndValidateVersionAsync(ServerWebExchange exchange) {
+		return this.resolveVersionAsync(exchange)
+				.switchIfEmpty(Mono.justOrEmpty(this.getDefaultVersion())
+									.mapNotNull(comparable -> comparable.toString()))
+				.<Comparable<?>>handle((version, sink) -> {
+					try {
+						sink.next(this.parseVersion(version));
+					}
+					catch (Exception ex) {
+						sink.error(new InvalidApiVersionException(version, null, ex));
+					}
+				})
+				.flatMap(version -> this.validateVersionAsync(version, exchange))
+				.switchIfEmpty(this.validateVersionAsync(null, exchange));
+
 	}
 
 	/**
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/handler/AbstractHandlerMapping.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/handler/AbstractHandlerMapping.java
@@ -185,38 +185,51 @@ public abstract class AbstractHandlerMapping extends ApplicationObjectSupport
 
 	@Override
 	public Mono<Object> getHandler(ServerWebExchange exchange) {
-		ApiVersionHolder versionHolder = initApiVersion(exchange);
-		return getHandlerInternal(exchange).map(handler -> {
-			if (logger.isDebugEnabled()) {
-				logger.debug(exchange.getLogPrefix() + "Mapped to " + handler);
-			}
-			if (versionHolder.hasError()) {
-				throw versionHolder.getError();
-			}
-			ServerHttpRequest request = exchange.getRequest();
-			if (hasCorsConfigurationSource(handler) || CorsUtils.isPreFlightRequest(request)) {
-				CorsConfiguration config = (this.corsConfigurationSource != null ?
-						this.corsConfigurationSource.getCorsConfiguration(exchange) : null);
-				CorsConfiguration handlerConfig = getCorsConfiguration(handler, exchange);
-				config = (config != null ? config.combine(handlerConfig) : handlerConfig);
-				if (config != null) {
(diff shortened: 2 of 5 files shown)


bclozel:
It is unfortunately too late for this type of change. This would break existing public contracts in a spectacular fashion. I will leave this open for now so we can discuss the use case.

JKaplanEmpty-Nes:
> It is unfortunately too late for this type of change. This would break existing public contracts in a spectacular fashion. I will leave this open for now so we can discuss the use case.

What if I changed the public contracts so that the old return was there, but added a new reactive one and by default it would enclose the old result in a mono?

bclozel:
We can always consider solutions like that. Let us discuss this matter first.

JKaplanEmpty-Nes:
> We can always consider solutions like that. Let us discuss this matter first.

Absolutely I would love to discuss my use case, just in case I made the change to not break the previous contracts.

The specific case is that we set the version (alpha, beta, GA) based on the Authorized Principal of the request. 
However using WebFlux we get that wrapped in a Mono, which we obviously can't block to resolve.

rstoyanchev (changes requested):
This was unfortunately an oversight, but we need to consider the best way to get there with the least disruption.

I do like the idea of allowing an `ApiVersionResolver` to be sync or async, so the `resolveVersionReactively` method with a default implementation is good, but I would call it `resolveVersionAsync`.

However, it doesn't make sense to keep both sync and async variants at the level of `ApiVersionStrategy`, so we should deprecate the sync version for removal, and again call the new method `resolveVersionAsync`. A similar pattern for `resolveParseAndValidate`.

During the deprecation phase, we do need to continue to call the sync `resolveParseAndValidate` first, suppressing any `MissingApiVersionException`, and then conditionally call the new async variant if the sync version did not produce a version.

Let me know if you plan to make those updates.

JKaplanEmpty-Nes:
I pushed the following changes:
ApiVersionResolver.resolveVersionReactively renamed to resolveVersionAsync
ApiVersionStrategy.resolveVersionReactively renamed to resolveVersionAsync

Deprecated ApiVersionStrategy.resolveVersion since 7.0.3

Instead of defaulting the async resolver to sync resolver in the ApiVersionStrategy, both methods are being called when calling resolveParseAndValidate. If the sync method throws a MissingApiVersionException then it is suppressed.

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
```diff
@@ -65,27 +81,44 @@ void validateVersion(@Nullable Comparable<?> requestVersion, ServerWebExchange e
...
-				version = parseVersion(value);
-			}
-			catch (Exception ex) {
-				throw new InvalidApiVersionException(value, null, ex);
-			}
-		}
-		validateVersion(version, exchange);
-		return version;
+	@SuppressWarnings("Convert2MethodRef")
+	default Mono<Comparable<?>> resolveParseAndValidateVersion(ServerWebExchange exchange) {
```
rstoyanchev: This is a breaking changing of the signature of a public API method. I mentioned "similar pattern", meaning deprecate the existing public method, and introduce a new one next to it.

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
```diff
@@ -65,27 +81,44 @@ void validateVersion(@Nullable Comparable<?> requestVersion, ServerWebExchange e
...
+		return Mono.defer(()-> {
+					try {
+						return Mono.justOrEmpty(this.resolveVersion(exchange));
```
rstoyanchev: This is a blocking call within a reactive chain with non-blocking assumptions. `defer` doesn't use a different thread, it is just a way to make the code "cold", not triggered until subscription. You need `.subscribeOn` for this.

That said, related to the previous comment, I don't think the blocking version should be called here. Calling both sync and async should be done in AbstractHandlerMapping.

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
```diff
@@ -65,27 +81,44 @@ void validateVersion(@Nullable Comparable<?> requestVersion, ServerWebExchange e
...
+					}
+				}).<Comparable<?>>handle((version, sink) -> {
+					try {
+						this.validateVersion(version, exchange);
+						sink.next(version);
+					}
+					catch (MissingApiVersionException | InvalidApiVersionException ex) {
+						sink.error(ex);
+					}
+				})
```
rstoyanchev, line 139: Not sure this needs to be split into two `handle` operators. I think it makes it less readable.

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
```diff
@@ -65,27 +81,44 @@ void validateVersion(@Nullable Comparable<?> requestVersion, ServerWebExchange e
...
+				.switchIfEmpty(Mono.error(new MissingApiVersionException()));
```
rstoyanchev: I don't think this is correct. It's up to the validate method to decide on this depending on whether the version is required for example.

spring-webflux/src/main/java/org/springframework/web/reactive/handler/AbstractHandlerMapping.java
```diff
@@ -210,19 +210,19 @@ public Mono<Object> getHandler(ServerWebExchange exchange) {
 				}
 			}
 			return handler;
-		});
+		}));
```
rstoyanchev: Before this statement is where the blocking version should be called for backwards compatibility.

rstoyanchev (changes requested):

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
JKaplanEmpty-Nes: Just have the async one called in ApiVersionStrategy?

rstoyanchev: ApiVersionStrategy should have both sync and async versions (same patters as in ApiVersionResolver). The async version can call only the async resolve methods, while the sync calls the sync resolve methods.

rstoyanchev:
@JKaplanEmpty-Nes, thinking further about the use case:

> The specific case is that we set the version (alpha, beta, GA) based on the Authorized Principal of the request.

Could you clarify why the version depends on the logged in user? It seems to me the version to use should match the one requested by the client, which knows best what their capability and compatibility level is.

JKaplanEmpty-Nes:
> @JKaplanEmpty-Nes, thinking further about the use case:
> 
> > The specific case is that we set the version (alpha, beta, GA) based on the Authorized Principal of the request.
> 
> Could you clarify why the version depends on the logged in user? It seems to me the version to use should match the one requested by the client, which knows best what their capability and compatibility level is.

Sure, I can go into a bit more detail. My use case is very specific, but allowing async API version resolution will solve for it and other possible cases.

We use a 3rd party service called LaunchDarkly. It allows us to select the correct version of an API based on a supplied context. I created a Launch Darkly API Version resolver for WebMvc just fine however I was unable to do it for WebFlux. The context we send to Launch Darkly includes information from the JWT.

This allows us to switch versions for specific segments of users and allows us to change the routing at any time.

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
JKaplanEmpty-Nes: I put back the old resolveParseAndValidateVersion, deprecated it, and renamed the new one to resolveParseAndValidateVersionAsync

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
JKaplanEmpty-Nes: there is now an async version and the previous version so the resolveVersion is no longer being called in the async version

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
JKaplanEmpty-Nes: When the stream is empty we are now calling the validateVersionAsync with a null version.

spring-webflux/src/main/java/org/springframework/web/reactive/handler/AbstractHandlerMapping.java
JKaplanEmpty-Nes: The blocking version is now being called before the return

rstoyanchev:
Thanks for elaborating.

spring-webflux/src/main/java/org/springframework/web/reactive/accept/ApiVersionStrategy.java
```diff
@@ -59,6 +75,25 @@ public interface ApiVersionStrategy {
...
+	default Mono<Comparable<?>> validateVersionAsync(@Nullable Comparable<?> requestVersion, ServerWebExchange exchange) {
+		try {
+			this.validateVersion(requestVersion, exchange);
+			return Mono.justOrEmpty(requestVersion);
+		}
+		catch (MissingApiVersionException | InvalidApiVersionException ex) {
+			return Mono.error(ex);
+		}
+	}
+
```
rstoyanchev, line 87: I'm processing this locally now, but wondering is there a a specific reason to expose this an additional, separate, new method?

rstoyanchev: Is it for re-use purely? I'm not sure the implementation of a default method should lead to a new method on the contract.

JKaplanEmpty-Nes: I did it just in case someone needs to validate the version using an async process.
I removed it from the contract and make it private

rstoyanchev:
@JKaplanEmpty-Nes as I mentioned before, I'm already making changes locally. Please, don't push any more changes. All of those will not be included.

JKaplanEmpty-Nes:
> @JKaplanEmpty-Nes as I mentioned before, I'm already making changes locally. Please, don't push any more changes. All of those will not be included.

Sure, I misunderstood. I thought you said you were processing it locally not updating the code.

rstoyanchev:
Yeah, processing as in processing the PR, not digesting it.

I've merged the changes and also evolved the solution a bit further. The main thing to point out are the changes in `ApiVersionResolver`, deprecating the sync method and making the async method the main one instead (for `@FunctionalInterface`), and creating an `SyncApiVersionResolver` extension. This is a better setup since otherwise if you implement the async method, the sync one remains without an obvious implementation. It also matches a pattern we've used elsewhere in WebFlux. This does break lambda implementations, but that's inevitable with the new setup after the deprecation phase, so it's better to absorb that now. In any case the expectation is for concrete implementations so I think this should have limited impact at this point in time early in the 7.0.x phase.

Hopefully the rest is self explanatory and makes sense, but feel free to comment.

JKaplanEmpty-Nes:
> Yeah, processing as in processing the PR, not digesting it.
> 
> I've merged the changes and also evolved the solution a bit further. The main thing to point out are the changes in `ApiVersionResolver`, deprecating the sync method and making the async method the main one instead (for `@FunctionalInterface`), and creating an `SyncApiVersionResolver` extension. This is a better setup since otherwise if you implement the async method, the sync one remains without an obvious implementation. It also matches a pattern we've used elsewhere in WebFlux. This does break lambda implementations, but that's inevitable with the new setup after the deprecation phase, so it's better to absorb that now. In any case the expectation is for concrete implementations so I think this should have limited impact at this point in time early in the 7.0.x phase.
> 
> Hopefully the rest is self explanatory and makes sense, but feel free to comment.

As long as we have an async resolver then I am absolutely delighted. My first commit also changed the defaults to the asynced version but I changed it to the synced version based on the first suggestion.

rstoyanchev:
I've done one more iteration, restoring `ApiVersionResolver` to be exactly the way it was, and having `AsyncApiVersionResolver` as the sub-interface that can be implemented when necessary. In effect, the overall version resolution is still async first, but the sync resolver contract remains front and center as the more common way to resolve a version.

PR #36095, opened 2026-01-05
Optimize single-char wildcard path matching performance

playous:
Uses lazy evaluation in SingleCharWildcardedPathElement to avoid unnecessary Character.toLowerCase() calls. 
Best-case scenarios (lowercase URLs, wildcard matches) now require zero conversion calls.

Resolves performance TODO from 2017.


bclozel:
Thanks for your contribution @playous this has been merged in main.

PR #36097, opened 2026-01-05
Fix double encoding in `DefaultApiVersionInserter`

### LINKED ISSUE #36080: `DefaultApiVersionInserter` encodes already encoded URI
The `DefaultApiVersionInserter` encodes an already encoded URI when inserting a path segment version.
This way the URI is double encoded.

The following code:
```java
ApiVersionInserter.usePathSegment(0).insertVersion("1", URI.create("http://localhost/test?foo=%20"))
```

yields:
```
http://localhost/1/test?foo=%2520
```

expected is:
```
http://localhost/1/test?foo=%20
```

Attached is a sample project. Executing the included `ApiVersionInserterTest` reproduces the error.
[demo.zip](https://github.com/user-attachments/files/24371972/demo.zip)

xyraclius:
Fixes gh-36080

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultApiVersionInserter.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultApiVersionInserter.java
@@ -33,6 +33,7 @@ import org.springframework.web.util.UriComponentsBuilder;
  * Default implementation of {@link ApiVersionInserter}.
  *
  * @author Rossen Stoyanchev
+ * @author Nabil Fawwaz Elqayyim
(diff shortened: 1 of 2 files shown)


rstoyanchev (changes requested):
Could you add a test? Currently this is tested at a higher level, in `RestClientVersionTests`, but we can start a test class specifically for `DefaultApiVersionInserter`.

rstoyanchev:
Thanks for the pull request. Note I removed the author tag on the main class, while we don't have precise rules, a single line change is very minimal.

PR #36101, opened 2026-01-06
Defer the API version validation error

### LINKED ISSUE #36059: When no API version is provided, static resources fail to load
I am experimenting with the API Versioning feature introduced in Spring 7.

I have an image file named `foo.png` under the `resources/static` directory.
However, when I try to access `http://localhost:8080/foo.png` in the browser, the image cannot be loaded.

In my case, the `HandlerMapping`s are evaluated in the following order:
1. `RouterFunctionMapping` (with `ApiVersionStrategy`)
2. `RequestMappingHandlerMapping` (with `ApiVersionStrategy`)
3. `WelcomePageHandlerMapping`
4. `BeanNameUrlHandlerMapping`
5. `WelcomePageNotAcceptableHandlerMapping`
6. `SimpleUrlHandlerMapping`

I expect the image resource to be handled by `SimpleUrlHandlerMapping`. However, due to the ordering, the request fails earlier in the `initApiVersion(HttpServletRequest)` method of `RouterFunctionMapping`.

ghoonch4:
This change defers the API version validation error.

As a result, the `versionRequired` option (e.g., `spring.mvc.apiversion.required=true`) is evaluated for each `HandlerMapping`, and only applies when a `HandlerMapping` both configures an `ApiVersionStrategy` and is matched by the URL.

Closes gh-36059

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/HandlerMapping.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/HandlerMapping.java
@@ -90,6 +90,13 @@ public interface HandlerMapping {
 	 */
 	String API_VERSION_ATTRIBUTE = HandlerMapping.class.getName() + ".apiVersion";
 
+	/**
+	 * Name of the {@link ServerWebExchange#getAttributes() attribute} containing
+	 * the {@link RuntimeException} raised during API version validation.
+	 * @since 7.0.3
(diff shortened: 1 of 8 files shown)


rstoyanchev:
I opted for a slightly different approach. Thanks for the pull request in any case, and for the report.

PR #36104, opened 2026-01-07
ResolvableType.getNested typeProvider was missing

TAKETODAY:
This causes JacksonJsonHttpMessageConverter.read() 's contextClass to resolve to null

> Determine why this fails with JacksonJsonHttpMessageConverter but passes with MappingJackson2HttpMessageConverter

Because the `MappingJackson2HttpMessageConverter` is a `GenericHttpMessageConverter` that explicitly provides the `contextClass` during reading, whereas the `JacksonJsonHttpMessageConverter` obtains it through `(ResolvableType.getSource() as MethodParameter).getContainingClass`.

https://github.com/spring-projects/spring-framework/blob/253eb28458123fd0d0e1959c594c50fd7eed2a70/spring-web/src/main/java/org/springframework/http/converter/GenericHttpMessageConverter.java#L69-L71

but type.getSource() is not instanceof MethodParameter

https://github.com/spring-projects/spring-framework/blob/253eb28458123fd0d0e1959c594c50fd7eed2a70/spring-web/src/main/java/org/springframework/http/converter/AbstractJacksonHttpMessageConverter.java#L311-L320

https://github.com/spring-projects/spring-framework/blob/253eb28458123fd0d0e1959c594c50fd7eed2a70/spring-webmvc/src/test/java/org/springframework/web/servlet/mvc/method/annotation/HttpEntityMethodProcessorTests.java#L154-L176

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webmvc/mvc-servlet/localeresolver.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webmvc/mvc-servlet/localeresolver.adoc
@@ -83,4 +83,4 @@ that contain a parameter named `siteLanguage` now changes the locale. So, for ex
 a request for the URL `https://domain.com/home.view?siteLanguage=nl` changes the site
 language to Dutch. The following example shows how to intercept the locale:
 
-include-code::./WebConfiguration[tag=snippet,indent=0]
+include-code::./WebConfiguration[tag=snippet,indent=0,chomp=-tags]
--- a/framework-docs/package.json
+++ b/framework-docs/package.json
@@ -1,11 +1,11 @@
 {
   "dependencies": {
-    "antora": "3.2.0-alpha.4",
-    "@antora/atlas-extension": "1.0.0-alpha.2",
-    "@antora/collector-extension": "1.0.0-alpha.3",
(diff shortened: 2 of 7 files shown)


bclozel:
@TAKETODAY please sign your commit; no need to close/create a new PR, you can do this with:

```
git amend -s
git push --force
```

See [this blog post for more information](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring/)

sdeleuze:
Thanks for this PR, we are going to fix this but I tend to think it will be better (less risk of regression due to the changes on utility methods, more robust) to handle that using `SmartHttpMessageConverter` hints. As a consequence, I will decline this PR and create a related issue that I will tentativaly fix for 7.0.3.

TAKETODAY:
> Thanks for this PR, we are going to fix this but I tend to think it will be better (less risk of regression due to the changes on utility methods, more robust) to handle that using `SmartHttpMessageConverter` hints. As a consequence, I will decline this PR and create a related issue that I will tentativaly fix for 7.0.3.

OK，Thank you for your reply.

PR #36108, opened 2026-01-08
Polishing AOT cache documentation

mipo256:
Added an example of AOT cache with Java 25 (simplified AOT cache creation) + added a note on aot cache recomendation


sdeleuze:
Merged and refined, thanks!

PR #36115, opened 2026-01-08
Fix Spring-ORM SpringBeanContainer::getBeanClass returns proxy-class instead of target-class

lbsekr:
**Problem**
Since Hibernate ORM `7.2.x`, `ManagedBeanRegistryImpl` enforces that the class returned by `ContainedBean::getBeanClass` must be identical to the requested bean type.
Spring’s `SpringContainedBean` currently returns `bean.getClass()`, which for AOP-proxied beans is a synthetic CGLIB subclass rather than the user-defined type. This mismatch causes Hibernate to throw:

```
IllegalArgumentException: Bean is not of requested type
```

**Root cause**
`SpringContainedBean` misinterprets the contract of `ContainedBean::getBeanClass()`: it must return the original target class, not the runtime proxy class.

**Solution**
Use `AopUtils.getTargetClass(bean)` instead of `bean.getClass()` to defer the actual user-defined type of CGLIB/Java-proxy wrapper.

### DIFF
--- a/spring-orm/src/main/java/org/springframework/orm/jpa/hibernate/SpringBeanContainer.java
+++ b/spring-orm/src/main/java/org/springframework/orm/jpa/hibernate/SpringBeanContainer.java
@@ -27,6 +27,7 @@ import org.hibernate.resource.beans.spi.BeanInstanceProducer;
 import org.hibernate.type.spi.TypeBootstrapContext;
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.aop.framework.AopProxyUtils;
 import org.springframework.beans.BeansException;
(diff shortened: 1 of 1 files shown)


lbsekr:
For context see this conversation with the maintainers of Hibernate-ORM
https://discourse.hibernate.org/t/managedbeanregistryimpl-getbean-in-7-2-0-throwing-exception/12057

quaff:
@lbsekr I don't think it's a correct fix, I created #36116 including test case.

lbsekr:
Yes, your fix looks more straightforward

lbsekr:
I didn't take into account that ultimateTargetClass only defers from proxy-classes but does not retain abstract super type information.

PR #36124, opened 2026-01-10
Clarify unversioned handler matching precedence

parkhojeong:
The documentation is updated to clarify that a handler with no version attribute matches any version only when it is not superseded by a more specific version match.

As noted in [#36082](https://github.com/spring-projects/spring-framework/issues/36082)


parkhojeong:
Superseded by https://github.com/spring-projects/spring-framework/pull/36125.

Issue #36143, opened 2026-01-14
Backoff for `DefaultMessageListenerContainer` through `DefaultJmsListenerContainerFactory` is not applied consistently


joshdcu:
## Description

A `BackOff` policy can be configured for the JMS message listener container using the `setBackOff(BackOff backOff)` method ([ref-1](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jms/config/DefaultJmsListenerContainerFactory.html#setBackOff(org.springframework.util.backoff.BackOff)), [ref-2](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jms/listener/DefaultMessageListenerContainer.html#setBackOff(org.springframework.util.backoff.BackOff))). However, this policy is not applied consistently when a `JMSException` is encountered and the listener tries to recover.

## Reproduction

The reproduction is based on the [JMS messaging guide](https://spring.io/guides/gs/messaging-jms) on the Spring Boot documentation website with the following changes:
- no use of custom `Email` class for simplicity
- the listener container factory is configured with a fixed backoff policy of 3 maximum attempts with 1s between each attempt
- Artemis is used in the native mode to reproduce the issue

1. Ensure Java 21, a recent version of Maven, and Docker/Podman are installed on your machine.
2. Download the project ZIP ([`demo.zip`](https://github.com/user-attachments/files/24612171/demo.zip)) to your machine and extract it.
3. Build the app by running `mvn clean package` at the root of the project.
4. Launch Artemis as a container:
```
podman run -d --name artemis -p 61616:61616 docker.io/apache/activemq-artemis:2.44.0-alpine
```
5. Create a test user in Artemis:
```
podman exec artemis /var/lib/artemis-instance/bin/artemis user add --user artemis --password artemis --user-command-user test-user --user-command-password test-pswd --role test-role
```

Sample output:
```
Connection brokerURL = tcp://localhost:61616
test-user added successfully.
```

### Expected Behaviour 1

Run the app and see that the message is sent and received successfully:

```
java -jar target/demo-0.0.1-SNAPSHOT.jar
```

Sample output:
```
[...]

PR #36146, opened 2026-01-14
Fix backoff handling for listener setup failures

seekskyworld:
Issue: https://github.com/spring-projects/spring-framework/issues/36143

Summary:
- Reuse a single BackOffExecution for consecutive listener setup failures and stop the container when the back-off is exhausted.
- Reset the listener setup back-off after a successful receive.

Motivation:
- Listener setup failures currently restart the BackOffExecution each time, so max attempts are never reached.
- This aligns listener setup recovery behavior with connection recovery behavior.

Validation:
- .\\gradlew.bat :spring-jms:test --tests "org.springframework.jms.listener.DefaultMessageListenerContainerTests.listenerSetupFailureBackOffStopsContainer" --no-build-cache --no-daemon


bclozel:
The issue is still being triaged.

PR #36174, opened 2026-01-19
Upgrade to Apache POI 5.5

shub-est:
**1. Why:**
To remove CVEs

**2. What:**
Upgraded Constraints for Apache POI OOXML from 5.2.5 to 5.5.1 to remediate CVE-2024-25710, CVE-2024-26308 and CVE-2025-31672
Upgraded Apache Commons IO from 2.15.0 to 2.21.0 for compatibility with Apache POI


jhoeller:
Note that these are just optional compilation dependencies at the Spring Framework level, with no immediate relevance for CVE purposes. Only the versions brought in by Spring Boot's dependency management matter in that regard. Also, we do not usually rely on PRs for dependency upgrades but rather self-manage those before releases.

That said, POI is nevertheless worth upgrading, also for compatibility testing against the latest. For that reason, I'll merge the PR. Thanks for pointing out that we were several feature releases behind there already!

shub-est:
Thank you for your prompt review and explanation.

PR #36177, opened 2026-01-19
Add selector type support for sendAndReceive methods in JmsOperations and JmsTemplate

pisek:
For explicit response queue we need correlationId to be used.
There are 2 main approaches depending on JMS provider:
- CORRELATION_ID - correlationId set on request message that JMS provider copies to response message
- MESSAGE_ID - after sending request, JMS provider assigns messageId to request message and this value is also being copied to correlationId on response message

### DIFF
--- a/spring-jms/src/main/java/org/springframework/jms/core/JmsOperations.java
+++ b/spring-jms/src/main/java/org/springframework/jms/core/JmsOperations.java
@@ -407,6 +407,7 @@ public interface JmsOperations {
 	 * <p>The {@link MessageCreator} callback creates the message given a Session,
 	 * and the specified {@code responseQueue} is set in the {@code JMSReplyTO}
 	 * header of the message.
+	 * <p>A default {@link SelectorType#MESSAGE_ID} is used to correlate request and reply messages.
 	 * @param destination the destination to send the message to
 	 * @param responseQueue the destination to receive the reply from
 	 * @param messageCreator callback to create a message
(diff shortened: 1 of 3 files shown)


jhoeller:
Frankly, this adds quite a bit of complexity to the `sendAndReceive` overloads. At this point I wonder whether the previously introduced `sendAndReceive` methods without a `SelectorType` argument (#36162) are even useful on their own? If not, I'd rather revert that addition to 7.0.4, revisiting the topic for 7.1 instead.

If the selector type is commonly configured per provider, it should rather be a JmsTemplate-level setter rather than yet another `sendAndReceive` argument. Also, it is unusual for `JmsTemplate` to compose a selector expression - that might be an indication that the functionality should better live on top of `JmsTemplate` itself.

pisek:
Well, they would kindof work - only when you use one thread to call JMS... so yeap - it might be misleading and unusable in common scenarios :/
I forgot to include these changes in first PR.

The setter thing - sure, i will move it to setter then.

Composing selector expression - it is rather hard to compose it on top of JmsTemplate as you need messageId that is only set after the message is sent.

jhoeller:
In light of this, let's drop the "none" option and use the messageId approach by default for those `sendAndReceive` overloads with an explicit response queue. And for a custom correlation id, the provided `MessageCreator` callback can call `setJMSCorrelationID` if it prefers that - so we could simply check for `getJMSCorrelationID` in the message and use that when given, otherwise use `getJMSMessageID`. No need for additional overloads then and no need for a configurable enum either, and the correlation id can be built in a custom way in the `MessageCreator` rather than baking a UUID strategy into an enum.

jhoeller:
@pisek I'm experimenting with my approach above, using a variation of your unit tests. No need to update the PR right now, I'll keep you posted. Let me know if you see a fundamental problem with this approach though.

jhoeller:
I've pushed the implicit JMSCorrelationID/JMSMessageID usage as suggested above. While a JMSCorrelationID (if desired) has to be manually set in each `MessageCreator`, I find that flexibility preferable over a hard-coded enum mode.

From my perspective, this removes the need for the additional method overloads with the enum in this PR. Please give this a try against the latest 7.0.4 snapshot. If anything remains for this to be usable in the 7.0.4 release, let me know.

PR #36178, opened 2026-01-19
Replace System.err/printStackTrace with logger

VladimirKozulin:
Replace System.err.println() in ServletContextPropertyUtils and printStackTrace() in InvocationsRecorderClassTransformer with proper logging using Apache Commons Logging. This improves error handling and follows Spring Framework logging conventions.

Also fix typo in JsonConverterDelegate javadoc: 'tyep' -> 'type'.

### DIFF
--- a/spring-core-test/src/main/java/org/springframework/aot/agent/InvocationsRecorderClassTransformer.java
+++ b/spring-core-test/src/main/java/org/springframework/aot/agent/InvocationsRecorderClassTransformer.java
@@ -21,6 +21,8 @@ import java.lang.instrument.IllegalClassFormatException;
 import java.security.ProtectionDomain;
 import java.util.Arrays;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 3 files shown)


bclozel:
The InvocationsRecorderClassTransformer should not use a logging framework. This is done on purpose. This is part of a Java agent that needs to avoid calling the logging framework and initialize it during the class tranformation process.

As for the Servlet Utils one, maybe this is done in purpose as well for lifecycle reasons? I'm not sure.

jhoeller:
Actually, on review, those places should not use Commons Logging to begin with:
* The `aot.agent` package is a JVM agent for testing purposes (and even deprecated in the meantime).
* `ServletContextPropertyUtils` is a low-level utility that is only meant to be used in external setup code (analogous to `SystemPropertyUtils`, this may even be called in setup code for the logging system itself).

Thanks for the pull request, in any case! I'll make sure to fix the javadoc typo that you spotted there, and might replace the `ex.printStackTrace()` with a `System.err.println` call (just to satisfy the requirement of never calling `ex.printStackTrace()` in the framework codebase).

P.S.: @bclozel having written this at the same time as you, I concur. I missed that part at first glance.

PR #36179, opened 2026-01-20
Make SendToMethodReturnValueHandler and SubscriptionMethodReturnValueHandler customizable, to allow for pass-through of message headers

### LINKED ISSUE #36168: Make `SendToMethodReturnValueHandler` and  `SubscriptionMethodReturnValueHandler` customizable, to allow for pass-through of message headers
`SendToMethodReturnValueHandler`, `SubscriptionMethodReturnValueHandler` shall be customizable, in order to be able to pass through configured message headers of handled input messages.

Now: all message headers of input messages are discarded by mentioned ReturnValueHandlers.

Assuming a possibility to configure these handlers, it may be feasible to let message headers carried by input messages pass through into the messaging system while still using mentioned ReturnValueHandlers.
For example, this comes in handy for implementing receipt-header-assisted communication between producers and consumers.
Also, pass-through of other proprietary headers may enhance application implementations. 

References: 
> @JohnHaze I do see how copying all headers fits well for a controller method that intercepts a message before forwarding it to the broker, but I can also see a case where the controller method handles the message, and sends a new one. Therefore I'm not sure this should be done always, but it does make sense to provide control and allow you to do this.
> 
> Could you create a new issue, and we'll look for a way to make `SendToMethodReturnValueHandler` and `SubscriptionMethodReturnValueHandler` customizeable?  

 _Originally posted by @rstoyanchev in [#16664](https://github.com/spring-projects/spring-framework/issues/16664#issuecomment-3755830970)_

https://github.com/spring-projects/spring-framework/issues/16664#issuecomment-3712348328

catturtle123:
This PR addresses issue https://github.com/spring-projects/spring-framework/issues/36168

## Summary

* Added an opt-in mechanism to selectively propagate headers from an input `Message` to the outbound `Message` created by:

  * `SendToMethodReturnValueHandler`
  * `SubscriptionMethodReturnValueHandler`
* Header propagation is controlled via a configurable `Predicate<String>`.
* Default behavior remains unchanged: no headers are propagated unless explicitly configured.

## Resolved Issue

* Allows applications to forward selected headers (e.g. correlation or receipt-related headers) while avoiding accidental propagation of internal or protocol-specific headers.
* Provides a simple and explicit customization point instead of always copying or discarding headers.

## Scope

* This change applies **only at the Spring Messaging layer**.
* It affects how outbound `MessageHeaders` are populated by handler return value processing.
* It does **not** change STOMP native header serialization or WebSocket frame encoding.

### Discussion Points

* Whether the public getter (getHeaderPropagationPredicate) should be retained, or removed to keep the API strictly write-only.
* The current scope of this PR is limited to propagating headers at the Spring Messaging level.
A discussion point is whether this is sufficient for the intended use cases, or if extending the mechanism to STOMP native headers should be considered separately rather than as part of this change.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
+++ b/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
@@ -91,6 +91,8 @@ public class HandlerMethod extends AnnotatedMethod {
 
 	private @Nullable HandlerMethod resolvedFromHandlerMethod;
 
+	private @Nullable HandlerMethod resolvedBeanHandlerMethod;
+
 	private final String description;
 
 
@@ -298,8 +300,10 @@ public class HandlerMethod extends AnnotatedMethod {
 	}
 
 	/**
-	 * Return the HandlerMethod from which this HandlerMethod instance was
-	 * resolved via {@link #createWithResolvedBean()}.
+	 * Return the original HandlerMethod instance that from which the current
+	 * HandlerMethod instance was created via either
+	 * {@link #createWithValidateFlags()} or {@link #createWithResolvedBean()}.
+	 * The original instance is needed for cached CORS config lookups.
 	 */
 	public @Nullable HandlerMethod getResolvedFromHandlerMethod() {
 		return this.resolvedFromHandlerMethod;
@@ -322,6 +326,10 @@ public class HandlerMethod extends AnnotatedMethod {
 	 * <p>If the {@link #getBean() handler} is not String, return the same instance.
 	 */
 	public HandlerMethod createWithResolvedBean() {
+		if (this.resolvedBeanHandlerMethod != null) {
+			return this.resolvedBeanHandlerMethod;
+		}
+
 		if (!(this.bean instanceof String beanName)) {
 			return this;
 		}
@@ -329,7 +337,13 @@ public class HandlerMethod extends AnnotatedMethod {
 		Assert.state(this.beanFactory != null, "Cannot resolve bean name without BeanFactory");
 		Object handler = this.beanFactory.getBean(beanName);
 		Assert.notNull(handler, "No handler instance");
-		return new HandlerMethod(this, handler, false);
+
+		HandlerMethod handlerMethod = new HandlerMethod(this, handler, false);
+		if (this.beanFactory.isSingleton(beanName)) {
+			this.resolvedBeanHandlerMethod = handlerMethod;
+		}
+
+		return handlerMethod;
 	}
 
 	/**


spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
```diff
@@ -74,6 +75,13 @@ public class SendToMethodReturnValueHandler implements HandlerMethodReturnValueH
 	private @Nullable MessageHeaderInitializer headerInitializer;
 
 
+	/**
+	 * Predicate to determine which header names from the input message should be propagated.
+	 * If null, no headers are propagated (default behavior).
+	 */
+	private @Nullable Predicate<String> headerPropagationPredicate;
```
rstoyanchev: Simplify the name to `headerFilter`, and no need for Javadoc on the field as well.

spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
```diff
@@ -133,6 +141,24 @@ public void setHeaderInitializer(@Nullable MessageHeaderInitializer headerInitia
 		return this.headerInitializer;
 	}
 
+	/**
+	 * Set a predicate to filter which headers from the input message should be propagated to the output message.
+	 * <p><b>Warning:</b> The predicate should avoid propagating or overwriting well-known protocol headers
+	 * (e.g. headers starting with "simp", "content-type", etc.) to prevent breaking internal messaging semantics.
+	 * </p>
+	 * <p>If not set, no input headers are propagated (default behavior).</p>
```
rstoyanchev: Instead of such a warning, which requires further investigation on its own, how about applying header propagation first, and then setting standard headers (like session id) second, so there is no possibility for overwriting.

spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
```diff
@@ -133,6 +141,24 @@ public void setHeaderInitializer(@Nullable MessageHeaderInitializer headerInitia
...
+	 */
+	public void setHeaderPropagationPredicate(@Nullable Predicate<String> predicate) {
+		this.headerPropagationPredicate = predicate;
```
rstoyanchev: We could allow multiple predicates to be provided, and combine those with `Predicate#or`, also changing the method name to `add~` instead of `set~`.

rstoyanchev (changes requested):

spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
catturtle123: Agreed. I’ve renamed it to headerFilter and removed the field-level Javadoc to keep it consistent with existing style.

spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
catturtle123: That makes sense. I’ve removed the warning and updated the order so propagated headers are applied first, followed by standard headers, to avoid any possibility of overwriting.

spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
catturtle123: Good suggestion. I’ve changed the API to addHeaderFilter and combined multiple predicates using Predicate#or.

catturtle123:
I've applied the requested changes (renaming to headerFilter, switching to add~ method with Predicate#or logic, and adjusting the execution order). I've squashed the changes into the main commit to keep the history clean as per the contribution guide. Sorry for making the delta review slightly harder.

rstoyanchev:
No problem, I added an additional property on `SimpAnnotationMethodMessageHandler` to make it a little easier to configure.

catturtle123:
Hi @rstoyanchev,

When I was working on this PR, I found that the propagated headers didn't actually reach the STOMP client. There are two issues:

Input side: STOMP custom headers (e.g. x-correlation-id) are stored inside the nativeHeaders sub-map, not as top-level MessageHeaders entries. So when headerFilter iterates over top-level keys, it never matches them.

Output side: Even if headers were copied to the top-level MessageHeaders, StompEncoder.writeHeaders() only serializes entries from the nativeHeaders sub-map, so they would never be written to the outbound STOMP frame.

That's why I had left it as a discussion point in the PR. Is the current scope what you intended, or should we extend propagation to STOMP native headers as well? If it makes sense to extend, I'd be happy to follow up with a PR.

rstoyanchev:
@catturtle123 thanks for the reminder on that, I did miss the note on native headers. One option is to apply the predicate at both levels, but ultimately the original request is about propagating native headers to the broker, so we should amend the the solution to apply the predicate on native headers.

I am re-opening in order to append the extra commits here and keep this as the main issue, but if you would like to submit the changes, you can send a new PR, and I will just pick the commits from there. Please, keep in mind 7.0.4 is due this week so do let me know if you're able to submit the changes before then.

catturtle123:
@rstoyanchev 

I've updated the implementation to apply the predicate on native headers instead of Spring Messaging-level headers. The createHeaders() method now uses SimpMessageHeaderAccessor.wrap() to extract native headers from the input message and propagates matching ones via setNativeHeaderValues().

The changes have been pushed. Let me know if anything else needs to be adjusted.

PR address: https://github.com/spring-projects/spring-framework/pull/36282

(I accidentally closed this PR while updating the branch.....)

rstoyanchev:
This is in now. I simplified the access to the native headers map to avoid creating copies of the map.

catturtle123:
Thanks for the review and polishing!

PR #36181, opened 2026-01-20
Migrate tests to `@ParameterizedClass` in `spring-core`

CodingMasterLSW:
This PR migrates parameterized tests to `@ParameterizedClass` in the `spring-core` module.

To prioritize code cohesion, I actively use `@Nested` test classes.

- See #35833

### DIFF
--- a/spring-core/src/test/java/org/springframework/core/io/ResourceTests.java
+++ b/spring-core/src/test/java/org/springframework/core/io/ResourceTests.java
@@ -44,7 +44,7 @@ import mockwebserver3.RecordedRequest;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.ParameterizedClass;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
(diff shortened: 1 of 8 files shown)


sbrannen:
Please provide the [DCO sign-off](https://probot.github.io/apps/dco/) in all commits.

CodingMasterLSW:
@sbrannen 
I have updated the commits to include the DCO sign-off. Thanks for pointing that out!

spring-core/src/test/java/org/springframework/core/io/buffer/support/DataBufferTestUtilsTests.java
```diff
@@ -29,12 +33,16 @@
  * @author Arjen Poutsma
  * @author Sam Brannen
  */
+@ParameterizedClass
+@MethodSource("org.springframework.core.testfixture.io.buffer.AbstractDataBufferAllocatingTests#dataBufferFactories()")
```
sbrannen: Can't this be declared directly on `AbstractDataBufferAllocatingTests`?

spring-core/src/test/java/org/springframework/core/io/buffer/support/DataBufferTestUtilsTests.java
```diff
@@ -29,12 +33,16 @@
...
 class DataBufferTestUtilsTests extends AbstractDataBufferAllocatingTests {
 
-	@ParameterizedDataBufferAllocatingTest
-	void dumpBytes(DataBufferFactory bufferFactory) {
-		this.bufferFactory = bufferFactory;
+	DataBufferTestUtilsTests(DataBufferFactory bufferFactory) {
+		DataBufferTestUtilsTests.this.bufferFactory = bufferFactory;
+	}
```
sbrannen: Similarly, instead of declaring such constructors in each subclass, I think it would be much better to annotate the `bufferFactory` field in `AbstractDataBufferAllocatingTests` with `@Parameter`.

spring-core/src/test/java/org/springframework/core/io/buffer/DataBufferUtilsTests.java
```diff
@@ -82,1210 +85,1140 @@ class DataBufferUtilsTests extends AbstractDataBufferAllocatingTests {
 		this.tempFile = Files.createTempFile("DataBufferUtilsTests", null);
 	}
 
+	@Nested
+	@ParameterizedClass
+	@MethodSource("org.springframework.core.testfixture.io.buffer.AbstractDataBufferAllocatingTests#dataBufferFactories()")
+	class ParameterizedDataBufferUtilsTests {
```
sbrannen: I think the `@Nested` test classes like this one should extend `AbstractDataBufferAllocatingTests`, and the enclosing classes should no longer extend that.

spring-core/src/test/java/org/springframework/core/io/ResourceTests.java
```diff
@@ -68,82 +68,88 @@
...
+			assertThat(relative4.isReadable()).isFalse();
+			assertThatExceptionOfType(FileNotFoundException.class).isThrownBy(relative4::contentLength);
+			assertThatExceptionOfType(FileNotFoundException.class).isThrownBy(relative4::lastModified);
+			assertThatExceptionOfType(FileNotFoundException.class).isThrownBy(relative4::getInputStream);
+			assertThatExceptionOfType(FileNotFoundException.class).isThrownBy(relative4::readableChannel);
+			assertThatExceptionOfType(FileNotFoundException.class).isThrownBy(relative4::getContentAsByteArray);
+			assertThatExceptionOfType(FileNotFoundException.class).isThrownBy(() -> relative4.getContentAsString(UTF_8));
+		}
+
+		private static Stream<Arguments> resource() throws URISyntaxException {
```
sbrannen:
```suggestion
		private static Stream<Arguments> resources() throws URISyntaxException {
```

While we're at it, we might as well rename that method to "resources".

spring-core/src/test/java/org/springframework/util/MultiValueMapTests.java
```diff
@@ -45,98 +44,120 @@
...
-		assertThat(map).isEqualTo(map2);
+	@Nested
+	@ParameterizedClass
+	@MethodSource("mapsUnderTest")
+	class ParameterizedMultiValueMapTests {
+
+		private final MultiValueMap<String, String> map;
+
+		public ParameterizedMultiValueMapTests(MultiValueMap<String, String> map) {
+			this.map = new LinkedMultiValueMap<>(map);
```
sbrannen, line 55:
```suggestion
			this.map = map;
```

sbrannen (commented):
Thanks for the first PR in this "series"! 👍 

I think we'll "learn as we go" regarding how best to achieve this migration.

So, I've requested a few changes as a first step.

spring-core/src/test/java/org/springframework/core/io/ResourceTests.java
CodingMasterLSW: Reflected your feedback in 84d677bb1a2a3341a1f1963bd19015342aaa0dfa

spring-core/src/test/java/org/springframework/core/io/buffer/support/DataBufferTestUtilsTests.java
CodingMasterLSW: I moved these annotations to the `AbstractDataBufferAllocatingTests` class.

spring-core/src/test/java/org/springframework/core/io/buffer/support/DataBufferTestUtilsTests.java
CodingMasterLSW: Oh, That's great point!
I'm new to `@parameter`, so I appreciate you pointing that out!

spring-core/src/test/java/org/springframework/core/io/buffer/DataBufferUtilsTests.java
CodingMasterLSW: That makes sense to me.
Reflected your feedback in 672157b4e05c9fa54f01c26000fb0f04dcd16988.

spring-core/src/test/java/org/springframework/util/MultiValueMapTests.java
CodingMasterLSW: I found a potential issue with using `@ParameterizedClass` in this case.                                                                                                                                                                
                                                                                                                                                                                                                                        
Since the parameter object from @MethodSource is created only once and shared across all test methods,
modifying it (e.g., this.map = map;) causes state to leak between tests. 

To ensure test isolation, I've considered the following options:   

1. Create a defensive copy in the constructor (current approach) 
Since the constructor is called for each test method, instantiating a new object ensures each test gets its own isolated instance. 

2. Use @ParameterizedTest instead
Each test method receives a fresh parameter instance, but this requires repeating the parameter declaration for every test method.  

Please let me know if there's a better way to handle this!

CodingMasterLSW:
Hi @sbrannen,

I've applied your feedback. Additionally, I’ve updated `ServerSentEventHttpMessageWriterTests` and `WebClientDataBufferAllocatingTests` (which inherit from AbstractDataBufferAllocatingTests) to use the @Test annotation as well.

Also, I’ve left a comment regarding a potential issue with one of the suggested changes. I’d appreciate it if you could share your thoughts on that.

Thanks!

CodingMasterLSW:
Hi @sbrannen,
I hope you are doing well. I wanted to gently follow up on this, as I am still interested and available to work on it. I would appreciate any feedback or thoughts when you have a chance.

sbrannen:
Hi @CodingMasterLSW,

I apologize for the belated reply. I was on vacation for a bit and have been tied up with other tasks.

I do still want to migrate some tests to `@ParameterizedClass`; however, this task currently has a rather low priority.

In any case, I'll do my best to look into this again over the summer.

Thanks for your understanding,

Sam

CodingMasterLSW:
@sbrannen 
No worries at all, I completely understand — please take your time. I'm happy to pick this up whenever it fits your schedule.

Feel free to reach out if there's anything I can help with in the meantime. Wishing you all the best!

Thanks,
Lee

PR #36195, opened 2026-01-22
Improve XJC configuration for `spring-oxm` in the Gradle build

Sineaggi:
Existing solution used a filter to remove the generated sources from the main source set, but we shouldn't have xjc sources on main at all. This also still leaves the sources associated with main, just excluded. Additionally, this code relies on the `xjc` task already being a dependency of `compileJava`, configuring sources when the `xjc` task is realized.

We can work around this by resetting the main and resources source sets to their standard dirs.

Once we've done that, we can hook up the output (here `flatMap`ed out from the generation task) so that the output directories are properly associated with the task that generates them.

This simplifies the overall configuration and enables the use of configuration-cache for test compile and debug tasks.

### DIFF
--- a/spring-oxm/spring-oxm.gradle
+++ b/spring-oxm/spring-oxm.gradle
@@ -1,3 +1,5 @@
+import com.github.bjornvester.xjc.XjcTask
+
 plugins {
 	id "com.github.bjornvester.xjc"
 }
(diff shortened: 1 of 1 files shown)


Sineaggi:
I don't use eclipse but it's possible https://github.com/spring-projects/spring-framework/commit/5aa38833dddf427c3d702930bf1678a126806191 is no longer necessary to work around Gradle's eclipse integration.

Sineaggi:
Any interest in merging this?

sbrannen:
Hi @Sineaggi,

Thanks for the PR. I'll look into this in the coming days.

sbrannen:
This has been merged into `main` in e102bc49332810b45b61ec1113bea981f15df366 and revised in 0f990e44a80fa98211db0fa8435db865a0cfd5c5 and 9b366b28b7ecfed300847ce09a54085de9d0b97f.

So, thanks for helping us improve our Gradle build scripts! 👍 

----

However, although the changes align the build scripts with how the XJC plugin is supposed to be used, those changes do not resolve all issues related to Gradle's configuration cache.

For example, if you now run `./gradlew :spring-oxm:check --configuration-cache`, you will encounter the following error.

```
Reusing configuration cache.

* What went wrong:
Could not load the value of field `val$prepareCondition` of `com.tngtech.archunit.lang.syntax.ObjectsShouldInternal$AddMode$1` bean found in field `addMode` of `com.tngtech.archunit.lang.syntax.ObjectsShouldInternal$ConditionAggregator` bean found in field `conditionAggregator` of `com.tngtech.archunit.lang.syntax.ClassesShouldInternal` bean found in field `__rules__` of task `:spring-oxm:checkArchitectureMain` of type `org.springframework.build.architecture.ArchitectureCheck`.
> Class 'java.util.function.Function$$Lambda/0x00007f8001167868' not found in Gradle runtime class loader 'MixInLegacyTypesClassLoader(legacy-mixin-loader)' of type 'org.gradle.initialization.MixInLegacyTypesClassLoader'.
```

In light of that, I changed the title of this issue and the commit to reflect that the changes do not address all configuration cache issues.

Sineaggi:
> In light of that, I changed the title of this issue and the commit to reflect that the changes do not address all configuration cache issues.

I've created an issue an issue for that here https://github.com/spring-projects/spring-framework/issues/36483

PR #36196, opened 2026-01-23
Use Reader.transferTo in FileCopyUtils

kilink:
Use the transferTo method in FileCopyUtils#copy(Reader, Writer); Reader subclasses may have optimized implementations of `transferTo`.


bclozel:
Thanks @kilink !

PR #36212, opened 2026-01-26
Polish RetryTask

quaff:
1. Fix wrong type name in assertion message.
2. Rename `retryTemplate` to `retryOperations` to match type better.


jhoeller:
Well spotted - however, while this is debatable, it is actually intentional. `JdbcClient` and `JmsClient` follow the same pattern of accepting `Operations` interfaces while suggesting the `Template` classes both in the parameter name and the assertion message. The rationale is that the common thing to pass in there is a template instance, with the operations interface primarily declared for testability. In other words, the operations interface is just a facade for a specific template rather than an independent contract.

PR #36221, opened 2026-01-28
Use Java Files utilities in FileCopyUtils

kilink:
Use helpers from java.nio.file.Files for some methods in FileCopyUtils.

Additionally, add unit tests for the various file-related methods.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/FileCopyUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/FileCopyUtils.java
@@ -16,7 +16,6 @@
 
 package org.springframework.util;
 
-import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


bclozel:
@jhoeller should we deprecate some of those utility methods as a result? We've done that in the past when Java upgrades turn those into one-liners.

jhoeller:
For completeness, I've also applied consistent stream closing through try-with-resources in #36223 and aligned 6.2.16 in #36224.

As for deprecating some methods, most of them do not provide much additional anymore indeed. However, they are quite old and potentially called from many parts of the ecosystem, so I'd rather leave them in place - not least of it all, for their consistent closing of streams in `FileCopyUtils` and the flushing after copying in `StreamUtils` (which some callers may effectively rely on).

FWIW, once we have a JDK 25 baseline, we can use `Reader#readAllAsString` in our `FileCopyUtils#copyToString` implementation.

PR #36225, opened 2026-01-29
Allow customization of the `ProxyFactory` in `HttpServiceProxyFactory` used to create HTTP service proxies

dungdm93:
The Spring Data project provides support for `RepositoryProxyPostProcessor`, which allows customization of the underlying `ProxyFactory` (e.g, registering additional interceptors) prior to proxy bean creation.

In contrast, `HttpServiceProxyFactory`, which is responsible for creating client proxies from HTTP service interfaces annotated with `@HttpExchange`, currently does not offer a straightforward extension point for registering interceptors (e.g., for tracing or cross-cutting concerns).

Example usage of `HttpServiceProxyFactory.ProxyFactoryCustomizer`:
```java
public class OpenTelemetryHttpServiceProxyFactoryCustomizer implements HttpServiceProxyFactory.ProxyFactoryCustomizer {
    private final SingletonSupplier<OpenTelemetry> openTelemetry;
    private final HttpServiceGroup group;

    public OpenTelemetryHttpServiceProxyFactoryCustomizer(SingletonSupplier<OpenTelemetry> openTelemetry, HttpServiceGroup group) {
        this.openTelemetry = openTelemetry;
        this.group = group;
    }

    @Override
    public void customize(ProxyFactory factory, Class<?> serviceType) {
        var interceptor = new OpenTelemetryMethodInterceptor(openTelemetry.obtain(), group);
        factory.addAdvice(0, interceptor);
    }
}

@Component
public class OpenTelemetryRestClientHttpServiceGroupConfigurer implements RestClientHttpServiceGroupConfigurer {
    private final SingletonSupplier<OpenTelemetry> openTelemetry;

    public OpenTelemetryRestClientHttpServiceGroupConfigurer(SingletonSupplier<OpenTelemetry> openTelemetry) {
        this.openTelemetry = openTelemetry;
    }

    @Override
    public void configureGroups(Groups<RestClient.Builder> groups) {
        groups.forEachProxyFactory(this::configureProxyFactory);
    }

    private void configureProxyFactory(HttpServiceGroup group, HttpServiceProxyFactory.Builder builder) {
        var customizer = new OpenTelemetryHttpServiceProxyFactoryCustomizer(openTelemetry, group);
        builder.proxyCustomizer(customizer);
    }
}
```

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/service/invoker/HttpServiceProxyFactory.java
+++ b/spring-web/src/main/java/org/springframework/web/service/invoker/HttpServiceProxyFactory.java
@@ -63,16 +63,19 @@ public final class HttpServiceProxyFactory {
 
 	private final @Nullable StringValueResolver embeddedValueResolver;
 
+	private final List<ProxyFactoryCustomizer> proxyCustomizers;
 
 	private HttpServiceProxyFactory(
 			HttpExchangeAdapter exchangeAdapter, List<HttpServiceArgumentResolver> argumentResolvers,
 			List<HttpRequestValues.Processor> requestValuesProcessor,
-			@Nullable StringValueResolver embeddedValueResolver) {
+			@Nullable StringValueResolver embeddedValueResolver,
+			List<ProxyFactoryCustomizer> proxyCustomizers) {
 
 		this.exchangeAdapter = exchangeAdapter;
 		this.argumentResolvers = argumentResolvers;
 		this.requestValuesProcessor = new CompositeHttpRequestValuesProcessor(requestValuesProcessor);
 		this.embeddedValueResolver = embeddedValueResolver;
+		this.proxyCustomizers = proxyCustomizers;
 	}
 
 
@@ -84,7 +87,6 @@ public final class HttpServiceProxyFactory {
 	 * @return the created proxy
 	 */
 	public <S> S createClient(Class<S> serviceType) {
-
 		List<HttpServiceMethod> httpServiceMethods =
 				MethodIntrospector.selectMethods(serviceType, this::isExchangeMethod).stream()
 						.map(method -> createHttpServiceMethod(serviceType, method))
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/web/service/invoker/HttpServiceProxyPostProcessor.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.aop.framework.ProxyFactory;
+import org.springframework.web.service.registry.HttpServiceGroup;
+
+/**
+ * Callback interface used during HttpService proxy creation. Allows manipulating the {@link ProxyFactory} creating the
+ * HttpService.
+ *
+ * @author Dung Dang Minh
+ */
+public interface HttpServiceProxyPostProcessor {
```
rstoyanchev: I would make it  a nested interface in `HttpServiceProxyFactory` called `ProxyFactoryCustomizer`. I don't think it needs to be a top-level contract.

spring-web/src/main/java/org/springframework/web/service/invoker/HttpServiceProxyPostProcessor.java
```diff
@@ -0,0 +1,19 @@
...
+	/**
+	 * Manipulates the {@link ProxyFactory}, e.g. add further interceptors to it.
+	 *
+	 * @param factory will never be {@literal null}.
+	 */
+	void postProcess(ProxyFactory factory, Class<?> serviceType);
```
rstoyanchev: Also, update the method to match, i.e. `customize(...)`

rstoyanchev (changes requested):
Thanks for the suggestion. The request makes sense, however I would update the implementation a bit as suggested below. Having at least one test would also be a good idea.

Please, let me know if you will make those updates.

dungdm93:
@rstoyanchev I update PR as your suggestion, could you please take a look.

dungdm93:
@rstoyanchev I update PR as your suggestion, could you please take a look.

rstoyanchev (approved):
Thanks @dungdm93.

Issue #36236, opened 2026-01-31
ResponseEntity body nullability issue with Kotlin


alfonsoristorato:
Hi there, since Spring 4 and the introduction of JSpecify I am no longer able to declare some nullable types in `ResponseEntity`.

For example, such an endpoint:
```kotlin
@RestController
class HelloController {
    @GetMapping("/hello")
    fun hello(): ResponseEntity<String?> {
        val maybeHello :String? = "hey"
        return ResponseEntity.ok(maybeHello)
    }
}
```
would return a `ResponseEntity` with body `String`, or a `ResponseEntity` with `null` body (according to whatever logic determines whether the body is null or not).
Leaving alone as to whether or not this is a recommended Rest endpoint, I would expect this to compile properly, especially due to `ResponseEntity` contract, which is as follows (current version 7.0.2):
```java
	public static <T> ResponseEntity<T> ok(@Nullable T body) {
		return ok().body(body);
	}
```
However, although the IDE does not recognise this as an error (cause I assume it should not be, given `@Nullable`), when I compile I get the following error:
```
exception: src/main/kotlin/.../Application.kt:19:33: error: type argument is not within its bounds: must be subtype of 'Any'.
exception:     fun hello(): ResponseEntity<String?> {
exception:                                 ^^^^^^^
exception: src/main/kotlin/.../Application.kt:21:16: error: return type mismatch: expected 'ResponseEntity<String?>', actual 'ResponseEntity<String>'.
exception:         return ResponseEntity.ok(maybeHello)
exception:                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I understand that I can completely omit `ResponseEntity` and just return the nullable body, but I am confused as to why `Jsepcify` is not recognising this kotlin nullability.
I have read that there are still teething issues when it comes to `Jspecify` and generics, but I wonder if this would fall under this category, and if so what is your recommendation for the short term.

ngocnhan-tran1996:
You can try this


```kotlin
@RestController
class HelloController {
    @GetMapping("/hello")
    fun hello(): ResponseEntity<String> {
        val maybeHello :String? = null
        return ResponseEntity.ok(maybeHello)
    }
}
```

alfonsoristorato:
Yeah, this is what I am doing at the moment, and it compiles property and handles the nullability correctly when the body is indeed null.
But that is very much counter intuitive to me, plus for whatever openApi reader it will be quite unclear that the body can indeed be null

[...]

PR #36238, opened 2026-01-31
Revise nullability in HttpEntity and ResponseEntity

ngocnhan-tran1996:
Based on the IDE warning shown in the screenshot

image

This change adds `@Nullable` to the generic bound of ResponseEntity to make the nullability contract.

Related to gh-36236

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpEntity.java
@@ -56,7 +56,7 @@ import org.springframework.util.ObjectUtils;
  * @see #getBody()
  * @see #getHeaders()
  */
-public class HttpEntity<T> {
+public class HttpEntity<T extends @Nullable Object> {
 
 	/**
 	 * An {@code HttpEntity} instance with a {@code null} body and
--- a/spring-web/src/main/java/org/springframework/http/HttpEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpEntity.java
@@ -112,7 +112,7 @@ public class HttpEntity<T extends @Nullable Object> {
 	 * @param headers the entity headers
 	 * @deprecated in favor of {@link #HttpEntity(HttpHeaders)}
 	 */
-	@SuppressWarnings("NullAway")
+	@SuppressWarnings({"removal", "NullAway"})
 	@Deprecated(since = "7.0", forRemoval = true)
 	public HttpEntity(MultiValueMap<String, String> headers) {
 		this(null, headers);
--- a/spring-web/src/main/java/org/springframework/http/HttpEntity.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpEntity.java
@@ -67,12 +67,13 @@ public class HttpEntity<T extends @Nullable Object> {
 
 	private final HttpHeaders headers;
 
-	private final @Nullable T body;
+	private final T body;
 
 
 	/**
 	 * Create a new, empty {@code HttpEntity}.
 	 */
+	@SuppressWarnings("NullAway")
 	protected HttpEntity() {
 		this(null, (HttpHeaders) null);
 	}
@@ -90,6 +91,7 @@ public class HttpEntity<T extends @Nullable Object> {
 	 * @param headers the entity headers
 	 * @since 7.0
 	 */
+	@SuppressWarnings("NullAway")
 	public HttpEntity(HttpHeaders headers) {
 		this(null, headers);
 	}
@@ -100,7 +102,7 @@ public class HttpEntity<T extends @Nullable Object> {
 	 * @param headers the entity headers
 	 * @since 7.0
 	 */
-	public HttpEntity(@Nullable T body, @Nullable HttpHeaders headers) {
+	public HttpEntity(T body, @Nullable HttpHeaders headers) {
 		this.body = body;
 		this.headers = (headers != null) ? headers : new HttpHeaders();
 	}
@@ -110,7 +112,7 @@ public class HttpEntity<T extends @Nullable Object> {
 	 * @param headers the entity headers
 	 * @deprecated in favor of {@link #HttpEntity(HttpHeaders)}
 	 */
-	@SuppressWarnings("removal")
+	@SuppressWarnings("NullAway")
 	@Deprecated(since = "7.0", forRemoval = true)
 	public HttpEntity(MultiValueMap<String, String> headers) {
 		this(null, headers);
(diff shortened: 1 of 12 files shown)


spring-web/src/main/java/org/springframework/http/HttpEntity.java
sdeleuze: Here, you should just leverage the nullability defined at type level, so please remove the explicit `@Nullable` override here.

sdeleuze: Here, you should just leverage the nullability defined at type level, so please remove the explicit `@Nullable` override here and in other similar use cases.

spring-web/src/main/java/org/springframework/http/ResponseEntity.java
```diff
@@ -261,7 +261,7 @@ public static BodyBuilder ok() {
 	 * @return the created {@code ResponseEntity}
 	 * @since 4.1
 	 */
-	public static <T> ResponseEntity<T> ok(@Nullable T body) {
+	public static <T extends @Nullable Object> ResponseEntity<T> ok(@Nullable T body) {
```
sdeleuze: Here I think you should just use `public static <T> ResponseEntity<T> ok(T body)` since the nullability is defined at type level. Please adapt similar use cases.

sdeleuze (changes requested):
Please refine as suggested for all similar use cases and check with a full build that NullAway does not trigger errors.

spring-web/src/main/java/org/springframework/http/ResponseEntity.java
sbrannen: The `T` here is for a `static` method, which is not directly tied to the `T` in the `ResponseEntity` type declaration -- even though both types are called "T".

So, unless I'm confusing things, I think we need to explicitly declare this as follows.

```java
public static <T extends @Nullable Object> ResponseEntity<T> ok(T body) {
```

Otherwise, the supplied `body` cannot be `null`.

sdeleuze: Ah good point, so my remark is not relevant.

spring-web/src/main/java/org/springframework/http/HttpEntity.java
ngocnhan-tran1996: image

I have removed but the compileJava task build fail

>Within @NullMarked, remember, unannotated types are the same as if they were annotated with @NonNull. Since the bound of E is the same as @NonNull Number, and not @Nullable Number, that means the type argument for E can't be a type that includes null. @Nullable Integer can't be the type argument, then, since that can include null. (In other words: @Nullable Integer is not a subtype of Number.)

https://jspecify.dev/docs/user-guide/#declaring-generics

ngocnhan-tran1996: DITTO

sdeleuze: You need to update the rest of the codebase to fix NullAway errors, with for example here changing the return value to `HttpEntity<? extends @Nullable Object> build()` for example.

spring-web/src/main/java/org/springframework/http/ResponseEntity.java
```diff
@@ -78,7 +78,7 @@
  * @see org.springframework.web.client.RestOperations#getForEntity(URI, Class)
  * @see RequestEntity
  */
-public class ResponseEntity<T> extends HttpEntity<T> {
+public class ResponseEntity<T extends @Nullable Object> extends HttpEntity<T> {
```
sdeleuze, line 81: Can you check how `ResponseEntity` of "empty value" translates in a sample application checked with NullAway, using https://github.com/sdeleuze/jspecify-nullaway-demo as a basis maybe?

I guess that would be `ResponseEntity<@Nullable Void>` based on https://github.com/jspecify/jspecify/issues/51 but worth to check.

spring-web/src/test/kotlin/org/springframework/http/ResponseEntityKotlinTests.kt
```diff
@@ -43,4 +43,21 @@ class ResponseEntityKotlinTests {
 		assertThat(responseEntity.body).isNull()
 	}
 
+
+	@Test
+	fun ofNullNullableType() {
```
sdeleuze, line 39: Please add a test for empty value, ie. how `ResponseEntity<@Nullable Void>` translates to Kotlin. Is it `ResponseEntity<Unit>`?

sdeleuze (commented):
After discussing with the team, we would like to clarify how entity of empty value translates on both Java and Kotlin signature level before deciding if we move forward.

spring-web/src/main/java/org/springframework/http/ResponseEntity.java
```diff
@@ -117,7 +119,7 @@ public ResponseEntity(HttpHeaders headers, HttpStatusCode status) {
 	 * @param rawStatus the status code value
 	 * @since 7.0
 	 */
-	public ResponseEntity(@Nullable T body, @Nullable HttpHeaders headers, int rawStatus) {
+	public ResponseEntity(T body, @Nullable HttpHeaders headers, int rawStatus) {
```
PaulNgo-BlueOC, line 122: Why do we remove the annotation here? Does it mean the T body is not null?

sdeleuze: We remove this annotation there because with this PR, the nullability of `T` with be defined at type level and does not needs to be overridden at method level.

PaulNgo-BlueOC: Please correct me if i'm wrong
IMO, these are 2 completely different stories
- `public class HttptEntity<T extends @Nullable Object>` indicates that the `final T body` is nullable (we just move `@Nullable` from `final @Nullable T body` to the `<....>`
-The `@Nullable T body` in  `public ResponseEntity(@Nullable T body, @Nullable HttpHeaders headers, int rawStatus)`  indicates that the argument passed to the constructor is nullable. `final T body` is nullable vs ...passed to the constructor is nullable.
I would like to reopen the comment to get your feedback. Thanks

PaulNgo-BlueOC, line 122: Please correct me if i'm wrong
IMO, these are 2 completely different stories

    public class HttptEntity<T extends @Nullable Object> indicates that the final T body is nullable (we just move @Nullable from final @Nullable T body to the <....>
    -The @Nullable T body in public ResponseEntity(@Nullable T body, @Nullable HttpHeaders headers, int rawStatus) indicates that the argument passed to the constructor is nullable. final T body is nullable vs ...passed to the constructor is nullable.

PR #36241, opened 2026-02-01
Remove obsolete doc for spring.context.annotated-bean-reader.create

izeye:
This PR removes obsolete doc for `spring.context.annotated-bean-reader.create` that has been removed in #35570.


bclozel:
Thanks @izeye !

Issue #36248, opened 2026-02-03
Add opt-out for default RestClient StatusHandler


seungy0:
## Use case
  We need to treat some non-200 responses as valid business outcomes.
  Specifically, the API returns a JSON error payload with business error codes.
  For a known subset of error codes, the client should NOT throw and should return the response to the caller.

  We are using RestClient via RestClientAdapter (HttpServiceProxyFactory).

  ## What I tried
  I registered a custom ResponseErrorHandler via RestClient.Builder.defaultStatusHandler(...).
  The handler’s hasError(...) returns false for known business error codes so they can be handled as normal responses.

  However, RestClient still throws for 4xx/5xx because a default StatusHandler is always added and runs after my handler when hasError returns false.

  To fully bypass it, I had to add a handler that matches ALL status codes and conditionally runs the error handling inside it. That workaround is brittle and non-obvious.

  ## Problem
  - The default StatusHandler is added implicitly (DefaultResponseSpec constructor).
  - It is hard to discover why 4xx/5xx still throw even after registering a custom ResponseErrorHandler.
  - There is no official way to disable the default handler (NoOpResponseErrorHandler does not remove it).

  ## Enhancement request
  Provide an opt-out for the default StatusHandler, e.g.
  - `RestClient.Builder.disableDefaultStatusHandler()` or
  - `RestClient.Builder.defaultStatusHandlerEnabled(boolean)`

  Default behavior should remain unchanged.

  ## Related
  Issue: #33276

Thank you.

rstoyanchev:
Multiple questions here.

First, in the related issued, the [main takeaway](https://github.com/spring-projects/spring-framework/issues/33276#issuecomment-2257647017) is that `ResponseErrorHandler` was created for the `RestTemplate`, and while it is supported in `RestClient` to help with transitioning, it is not the best fit for it.

If you wanted to use it and take over, you do need to return `true` for an error response. However, as [explained in the Javadoc](https://github.com/spring-projects/spring-framework/blob/e3568a3f0a4b7b886b3e9790c2f79eca8261c548/spring-web/src/main/java/org/springframework/web/client/RestClient.java#L385-L388), if you want to take over status handling, you need to use the `Predicate` based alternative. The following works:

```java
RestClient.builder()
		.requestFactory(this.requestFactory)
		.defaultStatusHandler(statusCode -> true, (request,  response) -> {})
		.build();
```

[...]

PR #36251, opened 2026-02-04
Add opt-out for default RestClient status handler

seungy0:
## Summary
  Introduce an explicit opt-out for the built-in default RestClient status handler (4xx/5xx → exception).
  This allows users to fully control status handling without the “always-match” workaround.

  ## Motivation
  - Default StatusHandler is added implicitly, which is hard to discover.
  - NoOpResponseErrorHandler does not disable the built-in handler.
  - Treating some 4xx/5xx as normal outcomes is currently cumbersome.

  Related issue: https://github.com/spring-projects/spring-framework/issues/36248

  ## Changes
  - Add `RestClient.Builder#disableDefaultStatusHandler()` (default behavior unchanged).
  - Skip built-in handler registration when disabled.
  - Update Javadoc to document the opt-out.
  - Tests verifying default behavior vs opt-out.

  ## Tests
  - `DefaultRestClientTests.defaultStatusHandlerThrowsOnErrorStatus`
  - `DefaultRestClientTests.disableDefaultStatusHandlerAllowsErrorBody`
  - `RestClientBuilderTests.disableDefaultStatusHandler`

  ## Notes
  Backward compatible. Opt-in only.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
@@ -116,6 +116,8 @@ final class DefaultRestClient implements RestClient {
 
 	private final List<StatusHandler> defaultStatusHandlers;
 
+	private final boolean defaultStatusHandlerEnabled;
+
 	private final DefaultRestClientBuilder builder;
 
 	private final List<HttpMessageConverter<?>> messageConverters;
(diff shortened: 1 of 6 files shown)


LavaC3:
@seungy0 
Thanks for your feedback and fixing this issue perfectly!

🙏 I realize my previous implementation had an oversight (didn't consider the built-in StatusHandler logic in DefaultResponseSpec), and your solution with a dedicated disable flag passed to DefaultResponseSpec is definitely the right and complete approach. 

It's a great learning for me to dive into the RestClient source code and Spring contribution process through this issue.

rstoyanchev:
Closing as per https://github.com/spring-projects/spring-framework/issues/36248#issuecomment-3926068225.

PR #36282, opened 2026-02-09
Refactor SendTo and Subscription Method Return Value Handlers to use …

### LINKED ISSUE #36168: Make `SendToMethodReturnValueHandler` and  `SubscriptionMethodReturnValueHandler` customizable, to allow for pass-through of message headers

catturtle123:
### **PR Title**

**Propagate native STOMP headers to broker via HeaderFilter**

---

### **Summary**

This PR implements the propagation of native STOMP headers based on a predicate, as discussed in previous feedback. The core focus is ensuring that specific native headers (e.g., those with an `x-` prefix) are correctly carried over to the broker-bound messages.

**Key Implementation Details:**

* **Native Header Filtering:** Configured `addHeaderFilter` to propagate native headers starting with the `x-` prefix to response messages.
* Updated **`SendToMethodReturnValueHandler`**: Supports header propagation for `@SendTo` responses.
* Updated **`SubscriptionMethodReturnValueHandler`**: Supports header propagation for `@SubscribeMapping` responses.
* Updated unit tests to verify native header propagation via SimpMessageHeaderAccessor

---

### **Local Test (not included in this PR)**
스크린샷 2026-02-10 오전 4 11 40

---

### **Related Issues**

* Supersedes #36179
* Fixes #36168

---

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/annotation/support/SendToMethodReturnValueHandler.java
@@ -263,9 +263,10 @@ public class SendToMethodReturnValueHandler implements HandlerMethodReturnValueH
 			getHeaderInitializer().initHeaders(headerAccessor);
 		}
 		if (inputMessage != null && this.headerFilter != null) {
-			inputMessage.getHeaders().forEach((name, value) -> {
+			SimpMessageHeaderAccessor inputAccessor = SimpMessageHeaderAccessor.wrap(inputMessage);
+			inputAccessor.toNativeHeaderMap().forEach((name, values) -> {
 				if (this.headerFilter.test(name)) {
-					headerAccessor.setHeader(name, value);
+					headerAccessor.setNativeHeaderValues(name, values);
 				}
 			});
 		}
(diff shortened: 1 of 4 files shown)


rstoyanchev:
Thanks, I've incorporated merged the commit under #36179.

PR #36284, opened 2026-02-10
Remove obsolete code in `Redirecting to a resource` section

### LINKED ISSUE #36283: Incomplete example in "Redirecting to a resource"
In 6.2, the example snippet shown in the [Redirecting to a resource section](https://docs.spring.io/spring-framework/reference/6.2/web/webmvc-functional.html#webmvc-fn-resource) uses the `pathExtension()` request predicate to match requests for static resources.

In [the same section](https://docs.spring.io/spring-framework/reference/7.0/web/webmvc-functional.html#webmvc-fn-resource) in 7.0 (following the deprecation of the predicate), use of the `pathExtension()` predicate was removed, but the `extensions` list was not.

While the unused file extensions list can cause some confusion and should be removed, I'm more interested in the recommended way to achieve the same functionality as before, matching a request by its file extension. Perhaps the example could be fixed/updated to demonstrate that. In its current form, the code in the snippet would also serve the _index.html_ resource for a request to, for example, `/static/image.png`, which is what the previous version's snippet avoided.

ngocnhan-tran1996:
Closes gh-36283

### DIFF
--- a/framework-docs/modules/ROOT/pages/web/webmvc-functional.adoc
+++ b/framework-docs/modules/ROOT/pages/web/webmvc-functional.adoc
@@ -836,7 +836,6 @@ Java::
 [source,java,indent=0,subs="verbatim,quotes"]
 ----
 	ClassPathResource index = new ClassPathResource("static/index.html");
-	List<String> extensions = List.of("js", "css", "ico", "png", "jpg", "gif");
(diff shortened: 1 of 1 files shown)


bclozel:
Closed with 7a85ce959b

Thanks @ngocnhan-tran1996 !

PR #36325, opened 2026-02-13
Refactor calculateHashCode in RequestMappingInfo

AgilAghamirzayev:
This will:
1. Mathematical Distribution (Collision Reduction)
2. Pipelining and CPU Caching
3. Avoiding "Method Heavy" Expressions

1. Mathematical Distribution (Collision Reduction)
In your original version, the multiplier 31 is only applied to the first element.
-  Original: (A x 31) + B + C + D...
-  Optimized: ((((A x 31) + B) × 31) + C) × 31...
In the original, if field B changes by 1 and field C changes by -1, the hash remains exactly the same. When collisions happen, data structures like HashMap slow down significantly (shifting from O (1) to O(log n) or O(n) search time). By applying the multiplier at every step, you ensure that a small change in any field results in a massive change in the final hash.

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/RequestMappingInfo.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/RequestMappingInfo.java
@@ -541,11 +541,19 @@ public final class RequestMappingInfo implements RequestCondition<RequestMapping
 
 	@SuppressWarnings({"ConstantConditions", "NullAway", "removal"})
 	private int calculateHashCode() {
-		return (this.pathPatternsCondition != null ? this.pathPatternsCondition : this.patternsCondition).hashCode() * 31 +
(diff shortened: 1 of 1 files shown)


spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/RequestMappingInfo.java
```diff
@@ -541,11 +541,19 @@ public int hashCode() {
 
 	@SuppressWarnings({"ConstantConditions", "NullAway", "removal"})
 	private int calculateHashCode() {
```
amirpzk, line 543: In the `equals` method we consider `versionCondition` as part of identity, but here it became excluded. I'd suggest include `versionCondition` so when two `RequestMappingInfo` has everything the same but `versionCondition` is different then they won't produce the same hashcode.

amirpzk (changes requested):

PR #36328, opened 2026-02-13
Remove RFC 2047 encoding from Content-Disposition filename

tobifasc:
[Appendix C.1 of RFC 2066](https://www.rfc-editor.org/rfc/rfc6266.html#appendix-C.1) and [Section 5 of RFC 2047](https://www.rfc-editor.org/rfc/rfc2047.html#section-5) describe that an "encoded-word" (as described in RFC 2047) must not be used as parameter of a Content-Disposition header.

The current implementation in `ContentDisposition` does however encode the fallback `filename` parameter using the mechanism described in RFC 2047 (given that the charset is set to something other than `US_ASCII`).
Related discussion: https://github.com/spring-projects/spring-framework/issues/29861

This PR updates the Content-Disposition header creation logic to use only ISO-8859-1 characters for the fallback `filename` parameter instead. Non-compatible characters are replaced with `_`. The "full" filename is still present in the `filename*` parameter.

This does not remove the ability to parse RFC 2047 encoded headers.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
+++ b/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
@@ -17,11 +17,14 @@
 package org.springframework.http;
 
 import java.io.ByteArrayOutputStream;
+import java.nio.CharBuffer;
 import java.nio.charset.Charset;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CodingErrorAction;
+import java.nio.charset.CharacterCodingException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Base64;
-import java.util.BitSet;
 import java.util.HexFormat;
 import java.util.List;
 import java.util.Locale;
@@ -59,21 +62,7 @@ public final class ContentDisposition {
 	private static final String INVALID_HEADER_FIELD_PARAMETER_FORMAT =
 			"Invalid header field parameter format (as defined in RFC 5987)";
 
-	private static final BitSet PRINTABLE = new BitSet(256);
-
 	private static final HexFormat HEX_FORMAT = HexFormat.of().withUpperCase();
-
-
-	static {
-		// RFC 2045, Section 6.7, and RFC 2047, Section 4.2
-		for (int i=33; i<= 126; i++) {
-			PRINTABLE.set(i);
-		}
-		PRINTABLE.set(34, false); // "
-		PRINTABLE.set(61, false); // =
-		PRINTABLE.set(63, false); // ?
-		PRINTABLE.set(95, false); // _
-	}
 
 
 	private final @Nullable String type;
@@ -195,7 +184,7 @@ public final class ContentDisposition {
 			}
 			else {
 				sb.append("; filename=\"");
-				sb.append(encodeQuotedPrintableFilename(this.filename, this.charset)).append('\"');
+				sb.append(toIso88591(encodeQuotedPairs(this.filename))).append('\"');
 				sb.append("; filename*=");
 				sb.append(encodeRfc5987Filename(this.filename, this.charset));
 			}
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/ContentDisposition.java
```diff
@@ -446,44 +431,8 @@ else if (b == '=' && index < value.length - 2) {
...
-	}
-
-	private static boolean isPrintable(byte c) {
-		int b = c;
-		if (b < 0) {
-			b = 256 + b;
-		}
-		return PRINTABLE.get(b);
+	private static String toIso88591(String input) {
+		return new String(input.getBytes(StandardCharsets.ISO_8859_1));
```
shardt68: this would lead to Questionmarks "?" in the propose filename, which is not allowed in most filesystems,
I would rather propose an underline in place of non ascii characters
Suggetion:
        // NFD decomposition splits characters like ä into base character 'a' + combining diacritic.
        // Removing the combining diacritics (Unicode category Mn) gives a readable ASCII approximation
        // (e.g. "Schöne Äpfel" → "Schone Apfel") without resorting to '?' which is a forbidden
        // filename character on Windows.
        String decomposed = java.text.Normalizer.normalize(input, java.text.Normalizer.Form.NFD);
        return decomposed
                .replaceAll("\\p{InCombiningDiacriticalMarks}", "")
                .replaceAll("[^\\x20-\\x7E]", "_");

tobifasc: Fair point with questionmarks in filenames. I changed the replacement character to "_".

Regarding the removal of combining diacritics: 
Since this is targeting ISO 8859-1, characters like ä, ö, û, ñ,... are actually supported and shouldn't cause issues.
I wanted to "fail fast" by removing unavailable characters instead of trying to approximate them. Does that sound reasonable or would you still prefer removing diacritical marks of characters not in ISO 8859-1?

bclozel:
Thanks for your contribution @tobifasc , this is now merged and will be shipped for the next 7.0.x maintenance release.

The PR is marked as "closed" because I had to amend and rebase your commit. In the future, you can avoid that by always submitting a PR from a branch on your fork (i.e. not `main`).

pechanec-iba:
I do not think this change is valid because Mozilla reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition refers to:
- ASCII, not ISO-8859-1 (I understand ASCII as 7 bytes ASCII)
- it contains example to replace 'é' by 'e' - so not to '_'

bclozel:
@pechanec-iba my understanding here is that ASCII is used for the main field and that another encoding can be used for the fallback field.

As for the following, this is an advice from the MDN and is not specified in the RFC:

> you can convert filename* to filename by substituting non-ASCII characters with ASCII equivalents

Can you elaborate more and share some context? Are you seeing a problem in production? What exactly?

pechanec-iba:
> @pechanec-iba my understanding here is that ASCII is used for the main field and that another encoding can be used for the fallback field.
> 
> As for the following, this is an advice from the MDN and is not specified in the RFC:
> 
> > you can convert filename* to filename by substituting non-ASCII characters with ASCII equivalents
> 
> Can you elaborate more and share some context? Are you seeing a problem in production? What exactly?

The problem blocked us from SpringBoot upgrade in production.
We have API->proxy->browser.
When API started to return 'é' in `filename` inside Content-Disposition (not RFC 2047 encoded like before), it started to fail on proxy - it was not able to process Content-Disposition and ignored it - it then caused problem in browser.
See stack trace from proxy below. '￩' is originally 'é' in API.

```
The HTTP response header [Content-Disposition] with value [attachment; filename=\"DevTools_Sel￩￩￩￩_____￭tory_Prehled.pdf\"; filename*=UTF-8''DevTools_Sel%C3%A9%C3%A9%C3%A9%C3%A9%C4%8D%C5%99%C5%A1%C4%8D%C5%BE%C3%ADtory_Prehled.pdf] has been removed from the response because it is invalid","ecs.version": "1.2.0","process.thread.name":"http-nio-8080-exec-4","log.logger":"org.apache.coyote.http11.Http11Processor","error.type":"java.lang.IllegalArgumentException","error.message":"The Unicode character [￩] at code point [65,513] cannot be encoded as it is outside the permitted range of 0 to 255","error.stack_trace":"java.lang.IllegalArgumentException: The Unicode character [￩] at code point [65,513] cannot be encoded as it is outside the permitted range of 0 to 255\n\tat org.apache.tomcat.util.buf.MessageBytes.toBytesSimple(MessageBytes.java:310)\n\tat org.apache.tomcat.util.buf.MessageBytes.toBytes(MessageBytes.java:283)\n\tat org.apache.coyote.http11.Http11OutputBuffer.write(Http11OutputBuffer.java:389)\n\tat org.apache.coyote.http11.Http11OutputBuffer.sendHeader(Http11OutputBuffer.java:368)\n\tat org.apache.coyote.http11.Http11Processor.prepareResponse(Http11Processor.java:1052)\n\tat org.apache.coyote.AbstractProcessor.action(AbstractProcessor.java:376)\n\tat org.apache.coyote.Response.action(Response.java:209)\n\tat org.apache.coyote.Response.sendHeaders(Response.java:439)\n\tat org.apache.catalina.connector.OutputBuffer.doFlush(OutputBuffer.java:289)\n\tat org.apache.catalina.connector.OutputBuffer.close(OutputBuffer.java:248)\n\tat org.apache.catalina.connector.CoyoteOutputStream.close(CoyoteOutputStream.java:176)\n\tat
```

Thank you for any comment.

bclozel:
I have created https://github.com/spring-projects/spring-framework/issues/36805 and will follow up on that new issue next week.

Issue #36300, opened 2026-02-11
DispatcherServlet swallows HttpMediaTypeException during content negotiation in Spring Boot 3.5.9


connect-ankit:
**Description**
During request mapping, when the Accept header contains a large number of media types (e.g., 50+), Spring MVC throws HttpMediaTypeException while parsing the header inside content negotiation.

**Spring Boot version**
3.5.9

**Spring Framework version**
6.2.15 (managed by Boot 3.5.9)

However, this exception is swallowed inside ProducesRequestCondition#getMatchingCondition() and null is returned instead of propagating the exception.

Because of this:

@ExceptionHandler(HttpMediaTypeNotAcceptableException.class) is never invoked
@ControllerAdvice cannot catch the error
request mapping fails silently
This makes it impossible to handle malformed or oversized Accept headers using standard MVC exception handling.

**Observed behaviour**

When calling an endpoint with many Accept values:

Accept: application/type1,application/type2, ... (60+)

**Internally**:

```
MediaType.parseMediaTypes()
  → throws HttpMediaTypeException

ProducesRequestCondition#getMatchingCondition()
  catch (HttpMediaTypeException)
  → returns null

```

**Result:**

handler mapping fails

controller not executed

exception never surfaces
**Expected behaviour**

One of the following:

Propagate HttpMediaTypeException to the MVC exception chain so @ExceptionHandler can handle it

Return 406/400 directly from DispatcherServlet

Currently, there is no way to handle large/malformed Accept headers inside MVC.

**Reproduction**

**Request**
```
curl -X GET http://localhost:8080/test \
$(for i in $(seq 1 60); do printf ' -H "Accept: application/type%d"' "$i"; done)
```

**Result**

handler not invoked
exception handler not invoked
Root cause (code reference)
Inside Spring Framework:
ProducesRequestCondition#getMatchingCondition
```
try {
    acceptedMediaTypes = this.getAcceptedMediaTypes(request);
}
catch (HttpMediaTypeException ex) {
    return null; **//The exception is swallowed here**
}
```

**This swallowing prevents the exception from propagating to controller-level handlers.**

**Workaround**

Add a OncePerRequestFilter to validate the Accept header before it reaches MVC:

@Component
```
public class AcceptHeaderLimitFilter extends OncePerRequestFilter {
    private static final int MAX_ACCEPT_VALUES = 50;
    @Override
    protected void doFilterInternal(
            HttpServletRequest request,
            HttpServletResponse response,
[...]

PR #36329, opened 2026-02-14
Add TooManyHttpMediaTypesException and TooManyMimeTypesException

PaulNgo-BlueOC:
#36300
Reason I make this PR
1. We have a "secret" constraint of 50 mime types
```
	public static <T extends MimeType> void sortBySpecificity(List<T> mimeTypes) {
		Assert.notNull(mimeTypes, "'mimeTypes' must not be null");
		if (mimeTypes.size() > 50) {
			throw new InvalidMimeTypeException(mimeTypes.toString(), "Too many elements");
		}

		bubbleSort(mimeTypes, MimeType::isLessSpecific);
	}
```
and we just throw a generic `InvalidMimeTypeException`, then `HttpMediaTypeNotAcceptableException`. It is unclear for the developers and end-users why my `MediaType` are `NotAcceptable`. Their fault? Or framework limitation? Or configuration error? Not easy to know!
2. The framework default to return 406 http status without any infomation, making the end-users/client can not know what is happenning behind
3. Adding new Exception will help the developers to write their exception handler to provide useful infomation to the end-users. E.g let them know they must not provide more than 50 mime types.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java
@@ -351,7 +351,7 @@ public abstract class MimeTypeUtils {
 	public static <T extends MimeType> void sortBySpecificity(List<T> mimeTypes) {
 		Assert.notNull(mimeTypes, "'mimeTypes' must not be null");
 		if (mimeTypes.size() > 50) {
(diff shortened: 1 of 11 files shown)


bclozel:
Thanks for your proposal but the issue has been declined. See https://github.com/spring-projects/spring-framework/issues/36300#issuecomment-3902172370

PR #36359, opened 2026-02-20
Remove unintended space in Javadoc of `HandlerMethod`

ngocnhan-tran1996:


rstoyanchev:
Good point that the split word results in an unintended space.

PR #36360, opened 2026-02-20
fix: close connection to avoid resource leak

cuiweixie:


spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java
```diff
@@ -63,9 +63,11 @@ public boolean exists() {
 					httpCon.setRequestMethod("HEAD");
 					int code = httpCon.getResponseCode();
 					if (code == HttpURLConnection.HTTP_OK) {
```
amirpzk, line 66: I would suggest consider invoking `disconnect()` in a finally block to avoid missing it on new return paths.

amirpzk (changes requested):
Small change suggested.

PR #36361, opened 2026-02-20
Avoid invoking writeFunction under lock in ChannelSendOperator

l2yujw:
ChannelSendOperator transitions to FIRST_SIGNAL_RECEIVED on the first signal and invokes the provided writeFunction to start the actual write pipeline.

Previously, writer invocation could occur while holding the operator's monitor lock, which increases lock hold time and can block cross-thread access or lead to re-entrant deadlock scenarios depending on the writer implementation.

This change limits synchronization to state transitions and signal bookkeeping and moves writer invocation and subscription outside the synchronized block.

A regression test verifies that the writer function is not invoked while holding the monitor lock by asserting another thread can acquire the publisher's monitor during invocation.


rstoyanchev (changes requested):
Thanks for the suggestion. To make the review (and history) easier, could you eliminate stylistic differences, i.e. not switching from if-else to if with return?

l2yujw:
Thanks for the suggestion — agreed. I updated the change to keep the original if/else structure. The diff is now limited to moving writeFunction invocation (and subscription) outside the synchronized block, without additional stylistic restructuring.

l2yujw:
Friendly ping — changes from your feedback are already applied. Let me know if anything else is needed!

l2yujw:
Hi @rstoyanchev, just following up on this PR. The requested stylistic cleanup has been applied, and the branch is now up to date with the current main and remains conflict-free. The original locking issue is also still present on main.

Would you be open to taking another look, or is there anything else you’d like me to adjust?

PR #36364, opened 2026-02-20
Update Spring MVC functional endpoints to use Servlet response header adapters

amirpzk:
After #36334, Spring MVC controllers use optimized Servlet response header handling via `ServletServerHttpResponse` and `ServletResponseHeadersAdapter`.

This updates functional endpoints by switching `AbstractServerResponse` header writing to `ServletServerHttpResponse` as well.

A test is also added to verify existing response headers are replaced when writing a `ServerResponse`.

Closes gh-36358


amirpzk:
Superseded by #36365 (same change, clean commit metadata).

PR #36370, opened 2026-02-22
Propagate max frame length to WebSocket session

### LINKED ISSUE #36369: ReactorNettyWebSocketSession ignores maxFramePayloadLength from WebsocketClientSpec
In Spring Framework 7.x.x `ReactorNettyWebSocketClient` correctly applies `maxFramePayloadLength` to the underlying `HttpClient`, but it fails to propagate this value to the created `ReactorNettyWebSocketSession`.

Steps to Reproduce:
1. Configure `ReactorNettyWebSocketClient` with a `WebsocketClientSpec` that has `maxFramePayloadLength` set to **128KB**.
2. Connect to a server (e.g., Tomcat or Jetty) that sends a frame larger than **64KB**.
3. The client throws `io.netty.handler.codec.TooLongFrameException: content length exceeded 65536 bytes`.

The `ReactorNettyWebSocketSession` should respect the `maxFramePayloadLength` configured in the client's `WebsocketClientSpec`, but the session remains initialized with the default **64KB** limit, causing errors when receiving larger frames.

aavoronin93:
This PR ensures that `maxFramePayloadLength` from `WebsocketClientSpec` is passed to `ReactorNettyWebSocketSession`.

Previously, the session used the default 64KB limit regardless of client configuration, causing `io.netty.handler.codec.TooLongFrameException` when receiving larger frames from servers like Tomcat or Jetty.

Closes gh-36369

| Before| After |
|------------|-----------|
|largePayload_before_fix|largePayload_after_fix|

I have updated the PR to include a fix for `JettyWebSocketClient` as well.

The issue was consistent across both clients: session-level limits were not being synchronized with the client-level configuration. For Jetty, I've implemented this by configuring the native session during initialization within the `JettyWebSocketHandlerAdapter`. For Reactor Netty, the `maxFramePayloadLength` is now correctly propagated through the session constructor.

I've also expanded the integration tests to verify the fix for both Reactor Netty and Jetty. All tests are passing.

| Before| After |
|------------|-----------|
|largePayload_jetty_before_fix|largePayload_fixed|

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java
@@ -127,16 +127,18 @@ public class ReactorNettyWebSocketClient implements WebSocketClient {
 	@Override
 	public Mono<Void> execute(URI url, HttpHeaders requestHeaders, WebSocketHandler handler) {
 		String protocols = StringUtils.collectionToCommaDelimitedString(handler.getSubProtocols());
+		WebsocketClientSpec wsClientSpec = buildSpec(protocols);
 		return getHttpClient()
 				.headers(nettyHeaders -> setNettyHeaders(requestHeaders, nettyHeaders))
-				.websocket(buildSpec(protocols))
+				.websocket(wsClientSpec)
 				.uri(url.toString())
 				.handle((inbound, outbound) -> {
 					HttpHeaders responseHeaders = toHttpHeaders(inbound);
 					String protocol = responseHeaders.getFirst("Sec-WebSocket-Protocol");
 					HandshakeInfo info = new HandshakeInfo(url, responseHeaders, Mono.empty(), protocol);
 					NettyDataBufferFactory factory = new NettyDataBufferFactory(outbound.alloc());
-					WebSocketSession session = new ReactorNettyWebSocketSession(inbound, outbound, info, factory);
+					WebSocketSession session = new ReactorNettyWebSocketSession(inbound, outbound, info, factory,
+							wsClientSpec.maxFramePayloadLength());
 					if (logger.isDebugEnabled()) {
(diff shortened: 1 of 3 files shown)


aavoronin93:
Updated: While expanding the regression tests to cover all clients in `WebSocketIntegrationTests`, I discovered that `JettyWebSocketClient` suffers from the same issue.

Even when the native Jetty client (org.eclipse.jetty.websocket.client.WebSocketClient) is configured with higher limits, the `JettyWebSocketSession` still defaults to 64KB, leading to: `CloseStatus[code=1009, reason=Text message too large: 69,624 > 65,536]`

|Configuration|Log|State|
|---|---|---|
|`org.eclipse.jetty.websocket.client.WebSocketClient#setMaxFrameSize`, `org.eclipse.jetty.websocket.client.WebSocketClient#setMaxBinaryMessageSize`, `org.eclipse.jetty.websocket.client.WebSocketClient#setMaxTextMessageSize`, `org.eclipse.jetty.websocket.client.WebSocketClient#setMaxOutgoingFrames`|image|image|

rstoyanchev (changes requested):
For Reactor Netty, it looks like originally we had an `maxFramePayloadLength` field that was passed to `ReactorNettyWebSocketSession` that was then replaced in favour of the `WebSocketClientSpec` from Reactor Netty. However, the change 2ed281f6a89d50fe042335a1c88b515cb254afff to make the switch didn't pass the value into the session constructor, so that's a regression in 7.0.

For Jetty, however, I'm not sure yet why we need to pass the settings from the Jetty `WebSocketClient` to the Jetty `Session`, with the latter created by the former. Could you drop the Jetty changes, and submit them as a separate PR, please?

aavoronin93:
After closer investigation, I’ve confirmed that `JettyWebSocketClient` correctly propagates the limits when properly configured, so I have restored `JettyWebSocketClient` to its original state.
The initial issues I encountered were due to **local misconfiguration.**

Issue #36372, opened 2026-02-23
Refactor JettyDataBuffer to eliminate conditional branching and SRP violations


verhasi:
JettyDataBuffer currently serves two distinct purposes: wrapping a Jetty Content.Chunk and acting as a virtual wrapper for a DataBuffer. This "dual-mode" nature violates the Single Responsibility Principle (SRP).

```java
public final class JettyDataBuffer implements PooledDataBuffer {

	private final DefaultDataBuffer delegate;

	private final Content.@Nullable Chunk chunk;

	private final JettyDataBufferFactory bufferFactory;

	private final AtomicInteger refCount = new AtomicInteger(1);


	JettyDataBuffer(JettyDataBufferFactory bufferFactory, DefaultDataBuffer delegate, Content.Chunk chunk) {
		Assert.notNull(bufferFactory, "BufferFactory must not be null");
		Assert.notNull(delegate, "Delegate must not be null");
		Assert.notNull(chunk, "Chunk must not be null");

		this.bufferFactory = bufferFactory;
		this.delegate = delegate;
		this.chunk = chunk;
	}

	JettyDataBuffer(JettyDataBufferFactory bufferFactory, DefaultDataBuffer delegate) {
		Assert.notNull(bufferFactory, "BufferFactory must not be null");
		Assert.notNull(delegate, "Delegate must not be null");

		this.bufferFactory = bufferFactory;
		this.delegate = delegate;
		this.chunk = null;
	}
``` 
By having two constructors—one where chunk is vital and one where it is explicitly null—the class is essentially saying, "I am two different types of Buffer, but I’m wearing the same skin."

The contradiction is that first the chunk is `@Nullable`, later `Assert.notNull(chunk, "Chunk must not be null");` and finally in the other constructor `this.chunk = null;`
It has a consequence in the methods emerging branching based on whether the chunk is null or not.

I propose to split this class and introduce JettyVirtualDataBuffer that has no chuck at all. 
Sure I provide a PR as a solution.

bclozel:
Superseded by #36373.

PR #36373, opened 2026-02-23
Refactor JettyDataBuffer structure to separate concerns 

verhasi:
by extending `JettyVirtualDataBuffer`, simplifying code by consolidating shared logic and delegating operations to the parent class.
Resolves the #36372


bclozel:
Thanks @verhasi !

PR #36409, opened 2026-03-03
Repeat condition 'skipEquals'

cetf9h:
I found some code to simplify when using IDEA to inspect the code, so I submit a PR here.


bclozel:
Thanks but this code is not directly maintained by the Spring team, this is an internal copy of the cglib codebase. We'd rather not merge cosmetic changes if it makes it harder for us to track the changes.

You can [suggest this enhancement to the cglib project directly](https://github.com/cglib/cglib).

PR #36410, opened 2026-03-03
Incomplete debug message in ConfigurationClassBeanDefinitionReader

cetf9h:
I found some code to simplify when using IDEA to inspect the code, so I submit a PR here.

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassBeanDefinitionReader.java
+++ b/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassBeanDefinitionReader.java
@@ -368,8 +368,8 @@ class ConfigurationClassBeanDefinitionReader {
 					"@Bean definition illegally overridden by existing bean definition: " + existingBeanDef);
 		}
 		if (logger.isDebugEnabled()) {
-			logger.debug("Skipping bean definition for %s: a definition for bean '%s' already exists. " +
-					"This top-level bean definition is considered as an override.".formatted(beanMethod, beanName));
+			logger.debug(("Skipping bean definition for %s: a definition for bean '%s' already exists. " +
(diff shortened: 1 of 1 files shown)


sbrannen:
We consider the code in question fine as it is: we do not wrap concatenated strings within parentheses.

If you are wondering about our coding guidelines, please consult our [Code Style](https://github.com/spring-projects/spring-framework/wiki/Code-Style) wiki page.

----

In addition, your proposed change has absolutely nothing to do with "fixing a format string argument count". Did you perhaps use AI to generate this PR?

bclozel:
I think this PR is valid, the parentheses do have a meaning here and are actually fixing a bug. In Java, the method call `.` operator has a higher precedence than the string concatenation `+` operator. Here, the arguments are only applied to the last string, before concatenation - so arguments are in fact not replaced.

See the following:

```java
System.out.println("First %s" + ", second %s" + ", third".formatted("1", "2"));
// prints 'First %s, second %s, third'

System.out.println(("First %s" + ", second %s" + ", third").formatted("1", "2"));
// prints 'First 1, second 2, third'
```
Reopening.

sbrannen:
Indeed. I apologize: I glanced at that too quickly.

cetf9h:
image
@sbrannen 


I did not use AI.
My test results are shown in the figure.
In the original `str1 + str2.formatted(beanMethod, beanName)`, the `str2.formatted(beanMethod, beanName)` part is executed first, and then its result is concatenated with `str1`.

我没有使用AI
我的测试结果如图，原来的 `str1 + str2.formatted(beanMethod, beanName)`，其中 `str2.formatted(beanMethod, beanName)` 会先执行，然后其结果与 `str1` 进行连接

bclozel:
Thanks @cetf9h this is now merged and backported

PR #36448, opened 2026-03-11
Update exception message in `JacksonJsonMessageConverter`

gbouwen:
Fix typo, there was probably a copy-paste error made and the wrong classname shows in the exception.


sbrannen:
Good catch!

Thanks

----

p.s. I noticed two other spots that I addressed in 4e8acb9ef2ff43c6f841af2a2ad0a3de53cac914.

PR #36461, opened 2026-03-13
Document webflux native transports

vpelikh:
Users struggle to determine if native transports are supported and how to enable them. This documentation makes it clear and provides actionable steps, improving performance for high-concurrency applications.

### DIFF
--- a/framework-docs/antora.yml
+++ b/framework-docs/antora.yml
@@ -75,6 +75,7 @@ asciidoc:
     kotlin-issues: 'https://youtrack.jetbrains.com/issue'
     micrometer-docs: 'https://docs.micrometer.io/micrometer/reference'
     micrometer-context-propagation-docs: 'https://docs.micrometer.io/context-propagation/reference'
+    netty-native-transports: 'https://netty.io/wiki/native-transports.html'
     petclinic-github-org: 'https://github.com/spring-petclinic'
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal, unfortunately we cannot accept it at the moment since the DCO is not signed.

Also I think this documentation section would be a better fit [for the Reactor Netty docs](https://projectreactor.io/docs/netty/release/reference/about-doc.html) - Spring Framework is not really involved here and the Reactor Netty would be better positioned to keep this information up-to-date. Please suggest improvements in the Reactor Netty project directly. Thanks!

vpelikh:
Hello @bclozel.

I understand your point of view.

However, I think that given the performance improvements of native transports and the ease of adding them, information about them should be provided to Spring Framework users not from somewhere on the internet, but directly in the documentation, as it is a cost-free optimization for most applications.

P.S. I have added DCO to my commit, but the PR does not update, apparently because it is closed.

bclozel:
@vpelikh I can't find such documentation in Reactor Netty. Wouldn't be strange to document this here and not have it in Reactor Netty? I am sure other people would argue that this could be in Spring Boot because dependency management is done there.

We can link to the relevant Reactor Netty section once it's there.

vpelikh:
@bclozel, you're right, I'll make an MR in Reactor Netty, thanks!

PR #36462, opened 2026-03-13
Apply InstanceSupplier post-processing when bypassed by explicit args

vlsi:
## Summary

Fixes #35871.

When a bean definition has an `InstanceSupplier` with `andThen()` post-processing (as generated by AOT for `@Autowired` setter injection), calling `getBean(name, args)` bypasses the `InstanceSupplier` in `createBeanInstance()` but the `andThen()` post-processing was silently skipped. This meant that `@Autowired` setter injection would not be invoked for prototype beans retrieved with explicit constructor arguments in AOT mode.

This PR introduces `InstanceSupplier#postProcessInstance()` — a method that applies only the post-processing chain (registered via `andThen()`) to an already-created instance, without invoking the instance creation itself. `AbstractAutowireCapableBeanFactory#doCreateBean()` now calls this when it detects that the instance supplier was bypassed due to explicit args.

### Changed files

- **`InstanceSupplier`** — new `postProcessInstance(RegisteredBean, T)` default method; the anonymous class returned by `andThen()` overrides it to chain through the `after` function.
- **`AbstractAutowireCapableBeanFactory`** — new hook `applyInstanceSupplierPostProcessing()` (no-op by default), called in `doCreateBean()` when `args != null` and an `InstanceSupplier` is present.
- **`DefaultListableBeanFactory`** — overrides `applyInstanceSupplierPostProcessing()` to delegate to `InstanceSupplier#postProcessInstance()`.

### Note for reviewers

The post-processing runs **after** `addSingletonFactory()` (early singleton caching). For the typical case where `postProcessInstance()` returns the same instance (e.g. setter injection), this is safe. If a custom `andThen()` returns a **different** object (wrapping), the existing circular reference detection at the end of `doCreateBean()` will detect the identity mismatch and throw `BeanCurrentlyInCreationException` — consistent with how Spring handles any `BeanPostProcessor` that changes bean identity. Reviewers may want to evaluate whether moving the post-processing before `addSingletonFactory()` would be preferable for the wrapping case, though that would diverge from the established pattern for other post-processors.

## Test plan

- [x] `PrototypeWithArgsAotTests` — end-to-end AOT test: prototype bean with `@Autowired` setter + `BeanFactoryAware`, retrieved via `getBean(name, args)` in a fresh AOT-compiled context
- [x] `DefaultListableBeanFactoryTests#singletonWithInstanceSupplierAndThenPostProcessingAppliedWithExplicitArgs`  — singleton bean with `InstanceSupplier.andThen()` + explicit args
- [x] `DefaultListableBeanFactoryTests#prototypeWithInstanceSupplierAndThenPostProcessingAppliedWithExplicitArgs`
  — prototype bean with `InstanceSupplier.andThen()` + explicit args

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Opus 4.6)
🤖 Cross-reviewed with Codex (GPT 5.4)

### DIFF
--- a/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
+++ b/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java
@@ -598,6 +598,20 @@ public abstract class AbstractAutowireCapableBeanFactory extends AbstractBeanFac
 
 		// Initialize the bean instance.
 		Object exposedObject = bean;
+
+		// If the instance supplier was bypassed (e.g. explicit args were provided),
+		// apply any post-processing that was registered via InstanceSupplier.andThen().
+		// Note: this runs after early singleton caching. For the typical case where
+		// post-processing returns the same instance, this is safe. If post-processing
+		// wraps the instance, circular reference detection at the end of this method
+		// will detect the mismatch and throw BeanCurrentlyInCreationException.
+		if (args != null && mbd.getInstanceSupplier() instanceof InstanceSupplier<?>) {
+			exposedObject = applyInstanceSupplierPostProcessing(exposedObject, beanName, mbd);
+			if (exposedObject != instanceWrapper.getWrappedInstance()) {
+				instanceWrapper = new BeanWrapperImpl(exposedObject);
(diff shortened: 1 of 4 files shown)


ssserj:
Please review this PR.

PR #36475, opened 2026-03-16
Fix HttpServiceProxyFactory return LinkedHashMap with method have wildcard return type

### LINKED ISSUE #36474: HttpServiceProxyFactory returns LinkedHashMap instead of target type for method with Wildcard return type
Almost the same as https://github.com/spring-projects/spring-framework/issues/36326 but with Wildcard return type
Feign works correctly without any issue

Reproduce-able: From https://github.com/anaconda875/spring-framework please checkout commit 74f5b94f42454d95766d111f0817542d7c4d2f52 and run getWildcardReturnType test case

PaulNgo-BlueOC:
close #36474

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/ResolvableType.java
+++ b/spring-core/src/main/java/org/springframework/core/ResolvableType.java
@@ -620,6 +620,7 @@ public class ResolvableType implements Serializable {
 		ResolvableType[] generics = getGenerics();
 		for (ResolvableType generic : generics) {
 			if (generic.isUnresolvableTypeVariable() ||
+					generic.isWildcardWithoutBounds() ||
 					generic.isUnresolvableWildcard(currentTypeSeen(alreadySeen)) ||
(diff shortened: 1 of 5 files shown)


spring-core/src/main/java/org/springframework/core/ResolvableType.java
```diff
@@ -1628,6 +1695,58 @@ public Object getSource() {
 	}
 
 
+	private static final class WildcardTypeImpl	implements WildcardType, Serializable {
```
PaulNgo-BlueOC: Error: eckstyle] [ERROR] /home/runner/work/spring-framework/spring-framework/spring-core/src/main/java/org/springframework/core/ResolvableType.java:1698:57: Use a single space to separate non-whitespace characters. [SingleSpaceSeparator]
May this is a bug from checkstyle?

PaulNgo-BlueOC:
Close #36474

Issue #36483, opened 2026-03-17
Configuration-cache issue relating to ArchitectureCheck in buildSrc


Sineaggi:
As mentioned by @sbrannen in https://github.com/spring-projects/spring-framework/pull/36195#issuecomment-4063040073, spring-framework cannot be run with configuration-cache as it encounters the following issue: 
```
Reusing configuration cache.g

FAILURE: Build failed with an exception.

* What went wrong:
Could not load the value of field `val$prepareCondition` of `com.tngtech.archunit.lang.syntax.ObjectsShouldInternal$AddMode$1` bean found in field `addMode` of `com.tngtech.archunit.lang.syntax.ObjectsShouldInternal$ConditionAggregator` bean found in field `conditionAggregator` of `com.tngtech.archunit.lang.syntax.ClassesShouldInternal` bean found in field `__rules__` of task `:spring-oxm:checkArchitectureMain` of type `org.springframework.build.architecture.ArchitectureCheck`.
> Class 'java.util.function.Function$$Lambda/0x00007f8001167868' not found in Gradle runtime class loader 'MixInLegacyTypesClassLoader(legacy-mixin-loader)' of type 'org.gradle.initialization.MixInLegacyTypesClassLoader'.
```

Looking into this task it appears that `ListProperty<ArchRule>` is refusing to serialize, which is what configuration cache is attempting to do.

One fix might be to inline all of the rules into the `@TaskAction`, and avoid creating a `ListProperty<ArchRule>` at all.

Alternatively, enums are serializable by Gradle, so something like 
```java
public enum ArchitectureRules {

	ALL_PACKAGES_SHOULD_BE_FREE_OF_TANGLES(allPackagesShouldBeFreeOfTangles()),
	NO_CLASSES_SHOULD_CALL_STRING_TO_LOWER_CASE_WITHOUT_LOCALE(noClassesShouldCallStringToLowerCaseWithoutLocale()),
	NO_CLASSES_SHOULD_CALL_STRING_TO_UPPER_CASE_WITHOUT_LOCALE(noClassesShouldCallStringToUpperCaseWithoutLocale()),
	CLASSES_SHOULD_NOT_IMPORT_FORBIDDEN_TYPES(classesShouldNotImportForbiddenTypes()),
	JAVA_CLASSES_SHOULD_NOT_IMPORT_KOTLIN_ANNOTATIONS(javaClassesShouldNotImportKotlinAnnotations())
	;
...
```
[...]

PR #36487, opened 2026-03-17
Fix configuration-cache compatibility with ArchRule task

Sineaggi:
Example fix for configuration-cache issues relating to embedding ArchRule into task objects.

https://github.com/spring-projects/spring-framework/issues/36483


bclozel:
Thanks @Sineaggi !

PR #36488, opened 2026-03-17
Enforce strict @Import rules for transitive bean dependencies

pranavmanglik:
This commit introduces an optional 'strict mode' to detect when a `Configuration class` uses a bean from a transitive source without explicitly importing that source.

Fixes #36153


bclozel:
Original issue is not triaged yet, so I'll close this PR. Thanks.

PR #36493, opened 2026-03-18
Use computeIfAbsent in CommandLineArgs.addOptionArg

junhyung8795:
Replace the manual `containsKey`/`put`/`get` three-step pattern in
`CommandLineArgs.addOptionArg` with `Map.computeIfAbsent`, which expresses
the same intent more concisely and avoids the redundant second map lookup.

Before:
```java
if (!this.optionArgs.containsKey(optionName)) {
    this.optionArgs.put(optionName, new ArrayList<>());
}
if (optionValue != null) {
    this.optionArgs.get(optionName).add(optionValue);
}
```

After:
```java
List<String> values = this.optionArgs.computeIfAbsent(optionName, key -> new ArrayList<>());
if (optionValue != null) {
    values.add(optionValue);
}
```

Behavior is identical: the option name is always registered in the map (even
without a value), and the value is added only when non-null.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/env/CommandLineArgs.java
+++ b/spring-core/src/main/java/org/springframework/core/env/CommandLineArgs.java
@@ -46,11 +46,9 @@ class CommandLineArgs {
 	 * without an associated value &mdash; for example, "--foo" vs. "--foo=bar".
 	 */
 	public void addOptionArg(String optionName, @Nullable String optionValue) {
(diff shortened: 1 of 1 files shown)


bclozel:
Please [sign the DCO](https://github.com/spring-projects/spring-framework/pull/36493/checks?check_run_id=67541216670).

bclozel:
Thanks for your contribution @junhyung8795 !

PR #36494, opened 2026-03-18
Fix: Avoid updating to request "Content-Type" in ServletServerHttpRequest

PaulNgo-BlueOC:
Fix https://github.com/spring-projects/spring-framework/issues/36426
Hope that my PR again approved and merged.
Please comment if anything need to be adapted


rstoyanchev:
I'm sorry but this is not the right approach and the issue is not open for contribution.

Issue #36512, opened 2026-03-22
AcceptHeaderLocaleResolver ignores defaultLocale when Accept-Language header is present but blank


elgunshukurov:
## Bug Description

`AcceptHeaderLocaleResolver` correctly returns `defaultLocale` when the `Accept-Language` header is **absent** (`null`), but ignores it when the header is **present with a blank value** (`""`), falling back to `Locale.getDefault()` (JVM system locale) instead.

## Current Behaviour
```java
if (defaultLocale != null && request.getHeader("Accept-Language") == null) {
    return defaultLocale;
}
```

This check only guards against `null`. A blank `""` header value passes through, hits `request.getLocale()` which delegates to the JVM default locale — ignoring the configured `defaultLocale` entirely.

## Expected Behaviour

A blank `Accept-Language` header value represents the same intent as an absent header — "no language preference". Both should return `defaultLocale` when configured.

## Steps to Reproduce
```java
AcceptHeaderLocaleResolver resolver = new AcceptHeaderLocaleResolver();
resolver.setDefaultLocale(Locale.forLanguageTag("az"));
resolver.setSupportedLocales(List.of(Locale.forLanguageTag("az")));

MockHttpServletRequest request = new MockHttpServletRequest();
request.addHeader("Accept-Language", "");  // blank, not null

Locale result = resolver.resolveLocale(request);
// actual:   en  (JVM default)
// expected: az  (configured defaultLocale)
```

## Proposed Fix
```java
String acceptLanguage = request.getHeader("Accept-Language");
if (defaultLocale != null && (acceptLanguage == null || acceptLanguage.isBlank())) {
    return defaultLocale;
}
```

## Environment

- Spring Framework 6.x
- Reproducible via curl: `curl --header 'Accept-Language;'` sends the header with an empty string value, which Tomcat parses as `accept-language = ""` (confirmed via `MimeHeaders` debug logging)
- JVM default locale affects result — behaviour differs across environments

bclozel:
Superseded by #36513

PR #36513, opened 2026-03-22
Handle blank Accept-Language header in AcceptHeaderLocaleResolver

elgunshukurov:
## Problem

`AcceptHeaderLocaleResolver.resolveLocale()` correctly returns `defaultLocale` 
when the `Accept-Language` header is absent (`null`), but ignores it when the 
header is present with a blank value (`""`), falling back to `Locale.getDefault()` 
(JVM system locale) instead.

## Fix

Replace the `null` check with `!StringUtils.hasText()` which handles `null`, 
empty string, and whitespace-only values consistently.

## Root cause
```java
// before
if (defaultLocale != null && request.getHeader("Accept-Language") == null) {

// after  
if (defaultLocale != null && !StringUtils.hasText(request.getHeader("Accept-Language"))) {
```

Closes gh-36512


bclozel:
@elgunshukurov at this stage we try to avoid introducing behavior changes in the 6.2.x line. Can you rebase against 7.0.x please?

elgunshukurov:
@bclozel done, rebased against 7.0.x. Thanks for the guidance.

bclozel:
Thanks for your contribution @elgunshukurov this has been merged in 7.0.x and main.

Issue #36374, opened 2026-02-23
Make it easier to create custom HttpExchangeAdapter


rstoyanchev:
It can be useful to be able to customize response handling in `RestClientAdapter` or `WebClientAdapter`, e.g. for custom response handling via `.exchange((request, response) -> {...})` instead of the `.retrieve()` workflow. Currently both adapters are final. Rather than making them open of extension, we can provide access to new request, and that would make it easy to use the built-in adapters as a delegate.

rstoyanchev:
There is one other minor detail in `RestClientAdapter` related to `InputStream` support. I've added #36380 to improve that, so all methods in `RestClientAdapter` will truly be straight forward adaptation to `RestClient` calls.

PR #36514, opened 2026-03-22
Perserve default API version in RestClientAdapter

itsmevichu:
This PR fixes the regression reported in spring-boot - https://github.com/spring-projects/spring-boot/issues/49701.


philwebb:
As mentioned in the Boot issue the cause of this is commit b83cb4a9479ae8c3b2e37f2a797c2648c6c2f25a which relates to issue #36374

PR #36543, opened 2026-03-26
Refactor: ResponseExtractor nullability for better inference

Ivictors:
Move @Nullable from extractData return type to the generic type parameter T. Closes gh-36319


Ivictors:
I see that this interface is deprecated, is the issue still valid?

PR #36549, opened 2026-03-27
Fix incorrect method name in `TypeDescriptor.array()` Javadoc

bebeis:
The Javadoc example for `TypeDescriptor.array()` references `TypeDescriptor.value(String.class)`, which does not exist.
The correct method is `TypeDescriptor.valueOf(String.class)`.
(The neighboring `collection()` and `map()` Javadoc examples already use `valueOf()` correctly.)


sbrannen:

Thanks

PR #36559, opened 2026-03-28
Augment FieldError.java's javadoc

ottlinger:
to make clearer that defaultMessage is not a i18n key from your application bundle, but a static string. Resulting in a problem that setting defaultMessage means no i18n. If you want i18n you should use the code parameter.

Behaviour shown in:
* https://github.com/ottlinger/custom-validation-messages-1029
* reported and resolved via https://github.com/thymeleaf/thymeleaf/issues/1029


DhruvTheDev1 (commented):
Simple but effective change. By adding 'static' to the defaultMessage description it helps to prevent from accidentally passing translation keys into the wrong parameter.

ottlinger:
Is there anything missing here to allow easier PR processing? Thanks

PR #36577, opened 2026-03-31
Invalid method return type metadata for ClassFile variant on JDK 24+

bebeis:
## Problem
On JDK 24+, `@Bean` methods declared with a `void` return type are not rejected during configuration parsing, and no `BeanDefinitionParsingException` is raised.

### JDK 21 (expected behavior)
```java
package com.java21.asm.config;

@Configuration
public class AppConfig {

    @Bean
    public void doSetUp() {
        // invalid @Bean method
    }
}
```

```
org.springframework.beans.factory.parsing.BeanDefinitionParsingException: Configuration problem: @Bean method 'doSetUp' must not be declared as void; change the method's return type or its annotation.
```

```java
@Test
void should_throw_beanDefinitionParsingException() {
    AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext();
    ctx.scan("com.java21.asm.config");

    assertThatThrownBy(ctx::refresh)
            .isInstanceOf(BeanDefinitionParsingException.class)
            .hasMessageContaining("must not be declared as void");
}
```

### JDK 24+ (actual behavior)
```java
package com.java25.classfile.config;

@Configuration
public class AppConfig {

    @Bean
    public void doSetUp() {
		// no exception is raised
    }
}
```
No exception is thrown and the context starts normally.

**Test Failed**
```java
@Test
void should_throw_beanDefinitionParsingException() {
    AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext();
    ctx.scan("com.java25.classfile.config");

    assertThatThrownBy(ctx::refresh)
            .isInstanceOf(BeanDefinitionParsingException.class)
            .hasMessageContaining("must not be declared as void");
}
```

## Root cause
On JDK 24+, Spring uses `ClassFileMethodMetadata` (backed by the JDK 24 ClassFile API) to read method metadata. The `getReturnTypeName()` method was implemented as:

```java
// Before fix
String returnTypeName = returnType.packageName() + "." + returnType.displayName();
```

`ClassDesc.packageName()` is only defined for class and interface types. Using it unconditionally when building return type names causes malformed results for primitive and array types.

| Return type | Before fix   | After fix            |
|-------------|--------------|----------------------|
| `void`      | `".void"`    | `"void"`             |
| `int`       | `".int"`     | `"int"`              |
| `String`    | `"java.lang.String"` | `"java.lang.String"` |
| `String[]`  | `".String[]"` | `"java.lang.String[]"` |

The most visible symptom is that `BeanMethod.validate()` in `spring-context` checks:
```java
// spring-context: BeanMethod.java
if ("void".equals(getMetadata().getReturnTypeName())) {
    problemReporter.error(new VoidDeclaredMethodError());
}
```

Because `getReturnTypeName()` returns `".void"` instead of `"void"` on JDK 24+, this check silently passes — allowing `void @Bean` methods without any error.

## Fix
Replaced the inline string concatenation with a `resolveTypeName()` helper that handles each case correctly:
```java
// After fix — ClassFileMethodMetadata.java (java24 source set)
private static String resolveTypeName(ClassDesc type) {
    if (type.isPrimitive()) {
        return type.displayName();
    }
    if (type.isArray()) {
        return resolveTypeName(type.componentType()) + "[]";
    }
    String packageName = type.packageName();
    return (packageName.isEmpty() ? type.displayName() : packageName + "." + type.displayName());
}
```
This ensures:
- Correct handling of primitive types (`void`, `int`, etc.)
- Proper resolution of array types based on their component type
- Consistent behavior with ASM and reflection-based metadata
- Safe handling of types in the default package, avoiding malformed names such as `.MyClass`

This change aligns ClassFile-based metadata with existing ASM and reflection-based implementations.

## Tests
* Added `ClassFileMethodMetadataTests` (java24Test)
  * covers primitive, array, reference, and void return types
* Extended `AbstractMethodMetadataTests`
  * added explicit void return coverage

## Notes
Similar issues exist in other ClassFile-based implementations (e.g. ClassFileAnnotationDelegate and Source#toString()), where primitive, array, and default package types may also be rendered incorrectly. These are not addressed in this PR.


bclozel:
Please hold off further changes here, I'll take care of fixing the failing build and additional cases.

bebeis:
I realized that I mistakenly used spaces instead of a tab on one line (ClassFileMethodMetadata.java:165 only). Sorry about that.
I’ll hold off on pushing any changes and leave it to you as suggested.

bclozel:
Thanks for your contribution @bebeis !

sbrannen:
FYI: I avoided the recursive calls for arrays in d4cc273c312f6ff64f9cf2eb31f1cf8d6c242abf.

PR #36580, opened 2026-04-01
Support @HttpExchange in MvcUriComponentsBuilder

### LINKED ISSUE #36235: HttpExchange equivalent of MvcUriComponentsBuilder
We could use a HttpExchangeUriComponentsBuilder that has similar functionality as the MvcUriComponentsBuilder,

MvcUriComponentsBuilder looks for @RequestMapping and fails on Http Interface classes
`No @RequestMapping on: public abstract org.springframework.http.ResponseEntity<com.company.api.model.Account> `

The goal would be to have similar fromMethodCall and fromMethodName functionality.

kwondh5217:
MvcUriComponentsBuilder now recognizes `@HttpExchange` and composed 
annotations such as `@GetExchange` in addition to `@RequestMapping` 
when resolving class- and method-level mappings.

This allows controllers that implement `@HttpExchange` interfaces to use 
`fromController()`, `fromMethodName()`, and `fromMethodCall()` to build links.

Closes gh-36235


DhruvTheDev1 (commented):
Nice fix! Using `AnnotatedElementUtils` here seems right to make sure it picks up on the `httpExchange` mappings when they're defined on the parents interface which is how the docs suggest using them. Nice testing too especially on the `PersonHttpExchangeControllerImpl` to verify the inheritance works as expected. Also nice on  updating the error message to include `httpExchange` which will save time debugging.

DhruvTheDev1 (approved):

PR #36590, opened 2026-04-02
Fix typo in HttpRange error message

xxxxxxjun:
`then` → `than` in error message for invalid byte range positions.


sbrannen:
This has been applied to 7.0.x and main.

Thanks

Issue #36158, opened 2026-01-15
WebClient (Reactor) attributes on Netty channel do not clear after connection release


nicholashagen:
WebClient, when used with Reactor HttpClient and pooled connection provider, sets the WebClient attributes map as a key on the underlying `NioSocketChannel`. This key is never unset. It is overwritten on subsequent requests. However, if requests are periodic on that channel, then the memory lives past the response. These attributes should be released and cleared when the channel is returned to the pool.

WebClient, once the request is executed in `AbstractClientHttpRequest.doCommit`, calls `applyAttributes`.
This invokes `org.springframework.http.client.reactive.ReactorClientHttpRequest.applyAttributes` (reactive version) that saves the entire attributes map into the Netty channel: https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpRequest.java#L148

The following test demonstrates this behavior. This creates a WebClient instance with reactor, netty, and pooled connection provider. After a request is made, that connection is returned to the pool. However, the attributes remain set and are never cleared. If these types of requests are executed across large clusters with large pools and large attributes are set, this can lead to memory issues, especially in a batch style environment where connections are bursty and sit idle for a moment.

This was tested against 7.0.2 version of Spring Framework but for sure also exists in 6.x where we initially discovered the issue. We validated against the 7.x series to verify it didn't get fixed in newer version.

```java
package webclient.bug.report;

import io.netty.channel.socket.nio.NioSocketChannel;
import io.netty.util.Attribute;
import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Test;
import org.springframework.http.client.reactive.ClientHttpConnector;
import org.springframework.http.client.reactive.ReactorClientHttpConnector;
import org.springframework.util.ReflectionUtils;
import org.springframework.web.reactive.function.client.ClientRequest;
import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
import org.springframework.web.reactive.function.client.ExchangeFunction;
import org.springframework.web.reactive.function.client.ExchangeFunctions;
import org.springframework.web.reactive.function.client.WebClient;
import reactor.netty.Connection;
import reactor.netty.http.client.HttpClient;
import reactor.netty.http.client.HttpClientConfig;
import reactor.netty.internal.shaded.reactor.pool.Pool;
import reactor.netty.resources.ConnectionProvider;

import java.lang.reflect.Field;
[...]

PR #36592, opened 2026-04-03
Fix flaky reactorNettyAttributes test by using doFinally for synchronous cleanup

### LINKED ISSUE #36589: Determine why `reactorNettyAttributes()` in `WebClientIntegrationTests` is flaky
## Overview

The `reactorNettyAttributes()` test method fails intermittently as follows.

```
org.opentest4j.AssertionFailedError: 	
expected: null	
 but was: {"org.springframework.web.reactive.function.client.ClientRequest.LOG_ID"="2a2cdba0", "org.springframework.web.reactive.function.client.WebClient.uriTemplate"="http://localhost:45573/greeting"}	
at org.springframework.web.reactive.function.client.WebClientIntegrationTests.reactorNettyAttributes(WebClientIntegrationTests.java:1344)
```

This occurs on the [CI server](https://ge.spring.io/s/lv2juuku6pw7k/tests/task/:spring-webflux:test/details/org.springframework.web.reactive.function.client.WebClientIntegrationTests/reactorNettyAttributes()?top-execution=1) as well as locally.

In the interim I have `@Disabled` that test method.

## Related Issues

- #36158

Pratapchandradeo:
Use `doFinally` operator in the reactive chain for guaranteed synchronous cleanup:
- Store `Connection` in `AtomicReference`
- Use `doFinally(signal -> clearChannelAttribute(conn))` instead of async callback
- Ensures cleanup is part of reactive chain
- Remove `@Disabled` annotation from test

## Changes
- **ReactorClientHttpConnector.java**: Add `connectionRef`, `doFinally` operator, `clearChannelAttribute()` method
- **WebClientIntegrationTests.java**: Remove `@Disabled` annotation

Fixes #36589
Relates to #36158

Signed-off-by: Pratap Chandra Deo <pratapchandradeo@users.noreply.github.com>

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpConnector.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpConnector.java
@@ -162,22 +162,33 @@ public class ReactorClientHttpConnector implements ClientHttpConnector, SmartLif
 
 		requestSender = setUri(requestSender, uri);
 		AtomicReference<ReactorClientHttpResponse> responseRef = new AtomicReference<>();
+		AtomicReference<Connection> connectionRef = new AtomicReference<>();
 
-		return requestSender
+			return requestSender
 				.send((request, outbound) -> requestCallback.apply(adaptRequest(method, uri, request, outbound)))
 				.responseConnection((response, connection) -> {
 					ReactorClientHttpResponse clientResponse = new ReactorClientHttpResponse(response, connection);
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal, but I don't think the change here will work. The connection and the response don't have the same lifecycle and this is likely to clear attributes before it should. This change does not seem to be complete, an unused variable is left and the intent is not clear. The DCO is not signed so we can't accept this contribution. Cheers.

PR #36599, opened 2026-04-05
`AbstractMockHttpServletRequestBuilder#param` does not allow `null` in Kotlin

seregamorph:
## Problem statement
Since Spring Framework 7.0 it's not possible to use `null` value as an argument for MockMvc query parameters.
E.g. for such snippet of Kotlin code
```kotlin
mockMvc
    .perform(
        get("/url")
            .param("limit", "1"),
            .param("cursor", null) // <- compilation fails now
    ).andExpect(status().isOk)
```

While `null` value is a legal value for query parameters (it's serialized as "?paramName" in the query string without following `=` (and is expected in our test on the server side)

Original change: https://github.com/spring-projects/spring-framework/commit/bc5d771a0639c6e47eda2b610c221a4515384abb by @sdeleuze

## Proposed solution
Mark argument `@Nullable` for Java callers and Kotlin extension.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/request/AbstractMockHttpServletRequestBuilder.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/request/AbstractMockHttpServletRequestBuilder.java
@@ -447,7 +447,7 @@ public abstract class AbstractMockHttpServletRequestBuilder<B extends AbstractMo
 	 * @param name the parameter name
 	 * @param values one or more values
 	 */
-	public B param(String name, String... values) {
+	public B param(String name, @Nullable String... values) {
 		this.parameters.addAll(name, Arrays.asList(values));
 		return self();
 	}
--- a/spring-test/src/main/kotlin/org/springframework/test/web/servlet/MockHttpServletRequestDsl.kt
+++ b/spring-test/src/main/kotlin/org/springframework/test/web/servlet/MockHttpServletRequestDsl.kt
@@ -107,7 +107,7 @@ open class MockHttpServletRequestDsl(private val builder: AbstractMockHttpServle
 	/**
 	 * @see [MockHttpServletRequestBuilder.param]
 	 */
-	fun param(name: String, vararg values: String) {
+	fun param(name: String, vararg values: String?) {
 		builder.param(name, *values)
 	}
 


sdeleuze:
@seregamorph If we do this change, we are going to need to update the nullability in much more other places (fields, `UriComponentsBuilder`, etc.)

@rstoyanchev Could you please confirm query parameters value should be nullable?

seregamorph:
Well, there is alternative option to solve this. Allow passing empty vararg:
```kotlin
mockMvc.perform(get("/url")
    .param("limit", "1"),
    .param("cursor")
```
This is compilable, but fails during the runtime:
```
java.lang.IllegalArgumentException: 'values' must not be empty
	at org.springframework.util.Assert.notEmpty(Assert.java:323)
	at org.springframework.test.web.servlet.request.AbstractMockHttpServletRequestBuilder.addToMultiValueMap(AbstractMockHttpServletRequestBuilder.java:954)
	at org.springframework.test.web.servlet.request.AbstractMockHttpServletRequestBuilder.param(AbstractMockHttpServletRequestBuilder.java:378)
	at org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder.param(MockHttpServletRequestBuilder.java:154)
```

Why this could be reasonable: in case if you want to achieve consistency with `UriComponentsBuilder`:
```
var uri = UriComponentsBuilder.fromUriString("https://google.com/")
        .queryParam("param1")
        .build();
// Allowed and evaluated as "https://google.com/?param1"
```

WDYT? Should I submit PoC fix for it?

emthrm:
A related data point from the **reactive server read side**, since this PR currently focuses on the MockMvc write side (`AbstractMockHttpServletRequestBuilder#param`).

A similar `@NullMarked` inconsistency also exists in [`ServerHttpRequest.getQueryParams()`](https://github.com/spring-projects/spring-framework/blob/v7.0.8/spring-web/src/main/java/org/springframework/http/server/reactive/ServerHttpRequest.java#L63-L66), where it can surface as a runtime NPE in Kotlin application code, rather than as a test compilation issue.

`AbstractServerHttpRequest.initQueryParams()` stores `null` as the value for a valueless query parameter (`?flag`, with no `=`):

https://github.com/spring-projects/spring-framework/blob/9c64be98e4d2cb9a0ee75f27055a1c3a1dd22fdf/spring-web/src/main/java/org/springframework/http/server/reactive/AbstractServerHttpRequest.java#L174-L175

However, the public return type is `MultiValueMap<String, String>`, whose values are non-null under `@NullMarked`.
As a result, Kotlin treats the value as non-null and inserts no null check:

```kotlin
// GET /endpoint?flag
val values: List<String> = exchange.request.queryParams.getValue("flag")
val value: String = values.first() // declared as a non-null String, but actually null
println(value.length) // NPE at runtime
```

This reproduces on 7.0.x. Before the JSpecify migration, these values surfaced as platform types in Kotlin, so nullability was the caller's responsibility.
Now the API declares the value as strictly non-null while the framework can still produce `null`, which seems contradictory.

PR #36594, opened 2026-04-03
Document that `spring.profiles.active` is ignored by the TCF

Mohak-Nagaraju:
## Overview

The Test Context Framework does not honor the `spring.profiles.active` system property when determining active profiles for a test class.

This commit documents that behavior in the following places:

- `@ActiveProfiles` Javadoc
- `DefaultActiveProfilesResolver` Javadoc
- Reference manual `@ActiveProfiles` annotation section
- Reference manual Context Configuration with Environment
  Profiles section, including a new
  `SystemPropertyActiveProfilesResolver` example showing how
  to allow `spring.profiles.active` to override `@ActiveProfiles`

## Related Issues

- #36269

Issue #36269, opened 2026-02-06
Improve documentation: behaviour for overriding profiles defined by "@ActiveProfiles" through "-Dspring.profiles.active"


manueljordan:
Hello Spring Team

Currently and so far in SF 7.0

For a Test class when it uses  `@ActiveProfiles`  it works as expected by applying the profile(s) defined through the `profiles` element. But is **not** possible override this through the command by `-Dspring.profiles.active=...`

In other words

* Override `@ActiveProfiles({"dev","cache"})` by `-Dspring.profiles.active=prod,-no-cache`

Of course, if the test class comments  `@ActiveProfiles({"dev","cache"})` then `-Dspring.profiles.active=prod,-no-cache` works in peace

Is there an official _explicit_ indication and explanation about 

1. Why is not possible override the profiles `@ActiveProfiles` through `-Dspring.profiles.active=...`? 
2. How to accomplish the desired goal? 

I already did do a research in the official documentation at:

* [@ActiveProfiles](https://docs.spring.io/spring-framework/reference/testing/annotations/integration-spring/annotation-activeprofiles.html)
* [Annotation Interface ActiveProfiles](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/ActiveProfiles.html)

And there is nothing about to resolve 1 and 2

According with my understanding to accomplish (2) is

a. Implement the [ActiveProfilesResolver](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/ActiveProfilesResolver.html) interface to take in consideration the `spring.profiles.active` option over the `profiles` element.  Of course it only if the former exists, otherwise the latter works in peace (practically the current behavioir of the unique and default implementation: the [DefaultActiveProfilesResolver](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/support/DefaultActiveProfilesResolver.html) class)
b. Finally refer this implementation through the `resolver` element as `@ActiveProfiles(profiles={"dev","cache"}, resolver=ImprovedActiveProfilesResolver.class)`

Thanks for your understanding

Mohak-Nagaraju:
Hi, I'd like to work on this. Could a maintainer confirm this is accepted and point me to the right documentation file to edit? Thank you.

sbrannen:
> Is there an official _explicit_ indication and explanation about the following?

There are various answers available online; however, you are correct: we have not officially documented this clearly.

[...]

PR #36600, opened 2026-04-05
Document that `spring.profiles.active` is ignored by `@ActiveProfiles`

Mohak-Nagaraju:
## Overview

The Test Context Framework does not honor the `spring.profiles.active` system property when determining active profiles for a test class.

This commit documents that behavior in the following places:

- `@ActiveProfiles` Javadoc
- `DefaultActiveProfilesResolver` Javadoc
- Reference manual `@ActiveProfiles` annotation section
- Reference manual Context Configuration with Environment
  Profiles section, including a new
  `SystemPropertyActiveProfilesResolver` example showing how
  to allow `spring.profiles.active` to override `@ActiveProfiles`

## Related Issues

- #36269
- #36594

### DIFF
--- a/framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-activeprofiles.adoc
+++ b/framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-activeprofiles.adoc
@@ -72,6 +72,13 @@ bean definition profiles programmatically by implementing a custom
 xref:testing/testcontext-framework/ctx-management/env-profiles.adoc#testcontext-ctx-management-env-profiles-ActiveProfilesResolver[`ActiveProfilesResolver`]
 and registering it by using the `resolver` attribute of `@ActiveProfiles`.
 
+NOTE: When `@ActiveProfiles` is declared on a test class, the `spring.profiles.active`
+property (whether configured as a JVM system property or environment variable) is not
+taken into account by the Test Context Framework when determining active profiles. If
+you need to allow `spring.profiles.active` to override the profiles configured via
+`@ActiveProfiles`, you can implement a custom `ActiveProfilesResolver` as described in
+xref:testing/testcontext-framework/ctx-management/env-profiles.adoc[Context Configuration with Environment Profiles].
+
 See xref:testing/testcontext-framework/ctx-management/env-profiles.adoc[Context Configuration with Environment Profiles],
 xref:testing/testcontext-framework/support-classes.adoc#testcontext-junit-jupiter-nested-test-configuration[`@Nested` test class configuration], and the
 {spring-framework-api}/test/context/ActiveProfiles.html[`@ActiveProfiles`] javadoc for
--- a/framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
+++ b/framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
@@ -543,3 +543,75 @@ Kotlin::
 ----
 ======
 
+The following example demonstrates how to implement and register a custom
+`SystemPropertyOverrideActiveProfilesResolver` that allows the `spring.profiles.active`
+property (whether configured as a JVM system property or environment variable) to
+override profiles configured via `@ActiveProfiles`:
+
+[tabs]
+======
+Java::
++
+[source,java,indent=0,subs="verbatim,quotes",role="primary"]
+----
+	// profiles resolved programmatically via a custom resolver that
+	// allows "spring.profiles.active" to override @ActiveProfiles
+	@ActiveProfiles(
+			resolver = SystemPropertyOverrideActiveProfilesResolver.class,
+			inheritProfiles = false)
+	class TransferServiceTest extends AbstractIntegrationTest {
+		// test body
+	}
+----
+
+Kotlin::
++
+[source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
+----
+	// profiles resolved programmatically via a custom resolver that
+	// allows "spring.profiles.active" to override @ActiveProfiles
+	@ActiveProfiles(
+			resolver = SystemPropertyOverrideActiveProfilesResolver::class,
+			inheritProfiles = false)
+	class TransferServiceTest : AbstractIntegrationTest() {
+		// test body
+	}
+----
+======
+
+[tabs]
+======
+Java::
++
+[source,java,indent=0,subs="verbatim,quotes",role="primary"]
+----
+	public class SystemPropertyOverrideActiveProfilesResolver implements ActiveProfilesResolver {
+
+		@Override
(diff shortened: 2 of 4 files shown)


framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-activeprofiles.adoc
```diff
@@ -72,6 +72,12 @@ bean definition profiles programmatically by implementing a custom
 xref:testing/testcontext-framework/ctx-management/env-profiles.adoc#testcontext-ctx-management-env-profiles-ActiveProfilesResolver[`ActiveProfilesResolver`]
 and registering it by using the `resolver` attribute of `@ActiveProfiles`.
 
+NOTE: The `spring.profiles.active` system property is not taken into account by the
```
sbrannen: It actually is taken into account if you don't declare `@ActiveProfiles` -- right?

sbrannen: Although it is usually set as a JVM system property, it's also supported as an environment variable. So, we should revise the wording here.

framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
```diff
@@ -543,3 +543,74 @@ Kotlin::
 ----
 ======
 
+The following example demonstrates how to implement and register a custom
+`SystemPropertyActiveProfilesResolver` that allows the `spring.profiles.active`
+system property to override profiles configured via `@ActiveProfiles`:
```
sbrannen: Again JVM system property vs. Spring property.

It might be easiest to refer to it as a Spring property and link to the Spring Properties section in the appendix, like we do in other places.

framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
```diff
@@ -543,3 +543,74 @@ Kotlin::
...
+----
+======
+
+[tabs]
+======
+Java::
++
+[source,java,indent=0,subs="verbatim,quotes",role="primary"]
+----
+	public class SystemPropertyActiveProfilesResolver implements ActiveProfilesResolver {
```
sbrannen: Naming is hard. 😉 

Perhaps `SpringPropertyOverrideActiveProfilesResolver`?

Got any better ideas?

framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
```diff
@@ -543,3 +543,74 @@ Kotlin::
...
+----
+
+Kotlin::
++
+[source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
+----
+	class SystemPropertyActiveProfilesResolver : ActiveProfilesResolver {
+
+		override fun resolve(testClass: Class<*>): Array<String> {
+			val profiles = System.getProperty("spring.profiles.active")
```
sbrannen:
```kotlin
			val profiles = SpringProperties.getProperty("spring.profiles.active")
```

spring-test/src/main/java/org/springframework/test/context/ActiveProfiles.java
```diff
@@ -37,6 +37,14 @@
  * <p>This annotation will be inherited from an enclosing test class by default.
  * See {@link NestedTestConfiguration @NestedTestConfiguration} for details.
  *
+ * <p>Note that the {@code spring.profiles.active} system property is not taken
```
sbrannen: Same as my first comment.

> It actually is taken into account if you don't declare `@ActiveProfiles` -- right?

sbrannen (changes requested):
Thanks for the PR and for basing it on 7.0.x.

sbrannen:
> The Test Context Framework does not honor the `spring.profiles.active` system property when determining active profiles for a test class.

In line with comments I left in the review, I've modified the title of this PR to reflect that it's the default behavior of `@ActiveProfiles` which does not honor that Spring property.

framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
sbrannen: For some reason I was thinking this could also be set as a Spring property, but that's not the case. So please disregard related comments.

framework-docs/modules/ROOT/pages/testing/annotations/integration-spring/annotation-activeprofiles.adoc
Mohak-Nagaraju: Good catch! You're right on both points.

1. The wording is too broad: `spring.profiles.active` is honored by the TCF when `@ActiveProfiles` is not declared. I'll narrow it to clarify that this only applies when `@ActiveProfiles` is present on the test class.
2. I'll also broaden "system property" to cover the environment variable case as well.

Will update in the next push.

framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
Mohak-Nagaraju: That makes sense. Since this specifically targets the JVM system property, would `SystemPropertyOverrideActiveProfilesResolver` be more accurate than `SpringPropertyOverrideActiveProfilesResolver`? Happy to use whichever you prefer!

spring-test/src/main/java/org/springframework/test/context/ActiveProfiles.java
Mohak-Nagaraju: Agreed, same fix applies here, will narrow the statement to be specific to when `@ActiveProfiles` is declared, and broaden "system property" to also mention the environment variable form.

framework-docs/modules/ROOT/pages/testing/testcontext-framework/ctx-management/env-profiles.adoc
sbrannen: `SystemPropertyOverrideActiveProfilesResolver` is fine.

Mohak-Nagaraju:
Updated in the latest push:
Narrowed the wording to be specific to when `@ActiveProfiles` is declared, broadened 'system property' to cover environment variables, and renamed the resolver class to SystemPropertyOverrideActiveProfilesResolver across all files.

sbrannen (commented):

sbrannen (approved):

sbrannen:
This has been merged into `7.0.x` and `main`, backported to `6.2.x`, and slightly revised in 99b78adce337c001986b2f79bc0d510e4570b49d.

Thanks 👍

PR #36602, opened 2026-04-05
Map version from @HttpExchange to RequestMappingInfo

renatomameli:
The version attribute from @HttpExchange and its shortcut annotations such as @GetExchange was not passed to the RequestMappingInfo builder, which caused version-based routing to be silently ignored for controllers annotated with @HttpExchange.

This commit adds .version(httpExchange.version()) to the createRequestMappingInfo(HttpExchange, ...) method in both WebMvc and WebFlux RequestMappingHandlerMapping, and adds corresponding tests.


renatomameli:
Hi team. Just a friendly ping on this PR. All checks are passing and there are no conflicts with main. I'd appreciate any feedback on the approach — happy to adjust if needed. Thanks! @sbrannen @jhoeller

PR #36606, opened 2026-04-06
`MergedAnnotation` does not use `ClassLoader` for method or field

nameearly:
I checked carefully, something goes wrong without `return`.


DhruvTheDev1 (commented):
Hi, the logic seems to be correct as the missing return was definitely causing the classLoader to be lost. Maybe a quick unit test to verify the member source now correctly returns its loader instead of null.

sbrannen:
Thanks for bringing this to our attention! 👍 

Since the commit had not been signed off and also did not include a test, I went ahead and completed this myself in f3b6c222f98f152643a006d0295a4324050f5c3e.

PR #36618, opened 2026-04-08
Polish DisconnectedClientHelper

quaff:
Use `CollectionUtils::newLinkedHashSet` instead of `LinkedHashSet::new` to avoid resizing.


sbrannen:

Thanks

PR #36626, opened 2026-04-08
Use defensive `Date` copies for `SimpleMailMessage` `sentDate`

bebeis:
## Summary

This PR updates `SimpleMailMessage` to use defensive `Date` copies for `sentDate`.

Currently, `SimpleMailMessage` already copies its recipient arrays in copy paths, but `sentDate` can still be shared across instances if it is not copied consistently. As a result, external mutation of a shared `Date` reference can affect message state unexpectedly.

## Changes

This change makes `sentDate` handling consistent by using defensive `Date` copies in the relevant paths.

Specifically:
- `setSentDate(Date)` stores a defensive copy
- `getSentDate()` returns a defensive copy
- the copy constructor copies `sentDate` independently
- `copyTo(MailMessage)` passes an independent `Date` instance

## Rationale

This is a narrow change intended to preserve the existing behavior of `SimpleMailMessage` while avoiding accidental state mutation through shared mutable `Date` references.

## Tests

I also added regression tests covering:
- mutation of the input `Date` after calling `setSentDate`
- mutation of the `Date` returned by `getSentDate`
- copy constructor isolation for `sentDate`
- `copyTo(MailMessage)` isolation for `sentDate`

### DIFF
--- a/spring-context-support/src/main/java/org/springframework/mail/SimpleMailMessage.java
+++ b/spring-context-support/src/main/java/org/springframework/mail/SimpleMailMessage.java
@@ -79,7 +79,7 @@ public class SimpleMailMessage implements MailMessage, Serializable {
 		this.to = copyOrNull(original.getTo());
 		this.cc = copyOrNull(original.getCc());
 		this.bcc = copyOrNull(original.getBcc());
-		this.sentDate = original.getSentDate();
+		this.sentDate = copyOrNull(original.sentDate);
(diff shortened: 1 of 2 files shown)


sbrannen:
This has been merged into `main` in 4b0101a9dcdf7dbe1841dfc33992fa8eb27eaf2f and slightly revised in e654b2e41b376b5fdf5f356ad382a53dad962b09.

Thanks

PR #36641, opened 2026-04-10
Avoid redundant URI object creation in WebClientUtils

### LINKED ISSUE #36605: Long URL causes high memory allocation in `WebClient` calls
Hi team, 

We are sending GET request with long urls (often longer than 200 characters due to various parameters)

Recently we tried to migrate it from apache http client to Webclient but found a performance hotspot in
https://github.com/spring-projects/spring-framework/blob/2ee4c3a3631aadc452a52c3423ff48959c0ba3ee/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientUtils.java#L75-L83

This method is primarily used for debugging / logging
https://github.com/spring-projects/spring-framework/blob/2ee4c3a3631aadc452a52c3423ff48959c0ba3ee/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultWebClient.java#L464-L466

Could you help make a fix on them?

MintBee:
Resolve [issue 36605](https://github.com/spring-projects/spring-framework/issues/36605#issuecomment-4221403256)

  ## Problem

  `WebClientUtils.getRequestDescription()` constructed a new `URI` via the
  multi-arg constructor on every call just to strip query parameters for logging.
  This re-parses and re-validates what the input `URI` already parsed.
  https://github.com/spring-projects/spring-framework/blob/2ee4c3a3631aadc452a52c3423ff48959c0ba3ee/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientUtils.java#L75-L83

  ## Changes

  Replace `new URI(scheme, null, host, port, path, null, null)` with direct
  `StringBuilder` composition from the original URI's components. The
  reconstruction logic follows JDK's own [`URI.defineString()`][1], minus the
  `userInfo`, `query`, and `fragment` fields — including its IPv6 bracket
  handling. Since the URI spec (RFC 3986) has been stable since 2005, this
  component-to-string mapping is unlikely to break.

  Also switched from `StringUtils.hasText(uri.getQuery())` to
  `uri.getRawQuery() != null` for a more performant query presence check. (no decoding needed)

  [1]: https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/java/net/URI.java#L1957

  ## Performance

  JMH (JDK 21, `@Fork(2)`, 5w+10m iterations) on the full
  `WebClient.get().uri(...).retrieve().toBodilessEntity().block()` pipeline
  with a no-op connector and a 200+ char query-heavy URI:

  | Version | ops/s | 99.9% CI |
  |---|---|---|
  | Baseline (`new URI(...)`) | 79,985 ± 19,649 | [60,335 — 99,635] |
  | Patched (`StringBuilder`) | 100,942 ± 18,521 | [82,421 — 119,464] |
  | **Speedup** | **1.26×** | |

  JFR confirms: `URI$Parser` execution samples dropped 86%, and
  `getRequestDescription` itself drops to 0 samples (below ~10ms threshold).

  ## Tests

  Added `WebClientUtilsTests` for regression testing, covering various situations.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientUtils.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientUtils.java
@@ -17,7 +17,6 @@
 package org.springframework.web.reactive.function.client;
 
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.List;
 import java.util.function.Predicate;
 
@@ -28,7 +27,6 @@ import reactor.core.publisher.Mono;
 import org.springframework.core.codec.CodecException;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
-import org.springframework.util.StringUtils;
 
 /**
  * Internal methods shared between {@link DefaultWebClient} and
@@ -69,18 +67,43 @@ abstract class WebClientUtils {
 	}
 
 	/**
-	 * Return a String representation of the request details for logging purposes.
+	 * Return a String representation of the request details for logging purposes
+	 * in "METHOD URI" format.
+	 * For the Security purpose, URI is returned in encoded format,
+	 * while userInfo, query, and fragment is stripped out.
 	 * @since 6.0.16
 	 */
 	public static String getRequestDescription(HttpMethod httpMethod, URI uri) {
-		if (StringUtils.hasText(uri.getQuery())) {
-			try {
-				uri = new URI(uri.getScheme(), null, uri.getHost(), uri.getPort(), uri.getPath(), null, null);
+		StringBuilder sb = new StringBuilder()
+				.append(httpMethod.name()).append(" ");
+
+		// also handles Opaque URI, which has only schemeSpecificPart
+		if (uri.getRawUserInfo() == null && uri.getRawQuery() == null && uri.getRawFragment() == null) {
+			return sb.append(uri).toString();
+		}
+
+		if (uri.getScheme() != null) {
+			sb.append(uri.getScheme()).append(':');
+		}
+		if (uri.getHost() != null) {
+			sb.append("//");
+			String host = uri.getHost();
+			// IPv6 handling
+			if (host.indexOf(':') >= 0 && !host.startsWith("[") && !host.endsWith("]")) {
+				sb.append('[').append(host).append(']');
 			}
-			catch (URISyntaxException ignored) {
+			else {
+				sb.append(host);
+			}
+
+			if (uri.getPort() != -1) {
+				sb.append(':').append(uri.getPort());
 			}
 		}
-		return httpMethod.name() + " " + uri;
+		if (uri.getRawPath() != null) {
+			sb.append(uri.getRawPath());
+		}
+		return sb.toString();
 	}
 
 }
--- a/spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
+++ b/spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2002-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


spring-webflux/src/main/java/org/springframework/web/reactive/function/client/WebClientUtils.java
```diff
@@ -69,18 +67,37 @@ public static <T> Mono<ResponseEntity<List<T>>> mapToEntityList(ClientResponse r
...
+			
+			if (uri.getPort() != -1) {
+				sb.append(':').append(uri.getPort());
+			}
+		}
+		if (uri.getPath() != null) {
+			sb.append(uri.getPath());
 		}
-		return httpMethod.name() + " " + uri;
+		return httpMethod.name() + " " + sb;
```
bclozel: Could we consider adding `httpMethod.name()` and the `" "` at the beginning to the StringBuilder directly? Using a Srting builder to then perform two concatenations is a bit wasteful.

spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
```diff
@@ -0,0 +1,127 @@
...
+	@Test
+	void stripsUserInfoWithQuery() {
+		URI uri = URI.create("https://admin:secret@host/api?token=abc");
+		assertThat(WebClientUtils.getRequestDescription(HttpMethod.GET, uri))
+				.isEqualTo("GET https://host/api");
+	}
+
+	@Test
+	void preservesUserInfoWithoutQuery() {
+		URI uri = URI.create("https://admin:secret@host/api");
```
bclozel, line 71: in the original issue, you added a comment about this: I think we should indeed strip the user info from the logged URI. Can you update the implementation to check for the presence of raw user info?

spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
```diff
@@ -0,0 +1,127 @@
...
+	@Test
+	void preservesPort() {
+		URI uri = URI.create("https://api.example.com:8443/path?q=1");
+		assertThat(WebClientUtils.getRequestDescription(HttpMethod.GET, uri))
+				.isEqualTo("GET https://api.example.com:8443/path");
+	}
+
+	@Test
+	void decodesPathWhenStrippingQuery() {
+		URI uri = URI.create("https://host/hello%20world?q=1");
```
bclozel, line 57: Could explain why the path needs decoding? I think using the raw path would be consistent with URI#toString().

bclozel (changes requested):

spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
MintBee: Should I strip the fragment too when only the fragment is in presence for the URI? I guess we should, since this is more consistent with existing API, and the purpose of the method is purely for logging.

spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
MintBee: There were two reasons.
1. Since it was for logging, I though decoded path is more readable.
2. Controller annotations take decoded string, such as @GetMapping("/hello world"). I though it would be more consistent with the overall API.

I should've made a comment about it. I was too focused on only describing the performance parts. Thanks for pointing that @bclozel 

Should I rollback and return the encoded string?

spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
bclozel: Yes please, let's remove the user info, query and fragment if any of them are present. Thanks!

spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientUtilsTests.java
bclozel: Yes please, let's use the encoded form. It will be consistent with the URI toString and will avoid log injection / log forging vulnerabilities.

MintBee: Resolved all what you have said @bclozel. Thanks for helping me a lot! Really appreciate it.

bclozel:
Thanks for your contribution @MintBee this is now merged!

PR #36647, opened 2026-04-13
Add zstd content coding support to EncodedResourceResolver

making:
Add zstd (Zstandard, RFC 8878) to the default content codings supported by EncodedResourceResolver in both Spring MVC and Spring WebFlux.


bclozel:
Thanks for the proposal @making but we didn't get much demand for this so we'll decline this feature for now.

making:
@bclozel  I agree that there isn't much demand for it. I believe this feature is underrated, likely because there hasn't been an easy mechanism for pre-compression.

That is why I created the `compression-maven-plugin`:
https://github.com/making/compression-maven-plugin
Thanks to this, the feature has become a hidden gem.
Brotli is already powerful enough, but having Zstd support as well would make it even more attractive...

PR #36650, opened 2026-04-13
Fix inverted logic for boolean last flag in JettyWebSocketSession when sending binary message

mguiking:
`sendPartialBinary`'s `last` parameter controls the WebSocket FIN bit — true means "this is the final fragment." Spring sets it to `iterator.hasNext()` which is inverted: for a single-buffer binary message, last becomes false, so Jetty leaves the connection in a fragmented state (FIN=0). The next send (e.g. a text message) then fails.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/socket/adapter/JettyWebSocketSession.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/socket/adapter/JettyWebSocketSession.java
@@ -209,7 +209,7 @@ public class JettyWebSocketSession extends AbstractWebSocketSession<Session> {
 							}
 
 							ByteBuffer buffer = iterator.next();
-							boolean last = iterator.hasNext();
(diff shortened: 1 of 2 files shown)


DhruvTheDev1 (commented):
Hi, nice catch on the inverted logic here - by changing it to `!iteratior.hasNext()` flips the logic so the final buffer signals the end of the message (FIN = 1). Without this the last buffer was signalling more data was to come. Maybe adding a test case can help prove the fix works.

DhruvTheDev1:
I noticed there isn't a dedicated unit test for this webflux module yet maybe adding it to the existing WebSocketIntegrationTests.java or creating a new JettyWebSocketSessionTests in the adapter folder?

DhruvTheDev1 (approved):
Hi, the tests look solid - checking for the false, false, true sequence proves the logic flip is correct for the fragmented messages. Looks good to me!

PR #36664, opened 2026-04-16
Refactor code for improved quality and exception handling

leonardogaertner:
Refactor: Apply PMD rules improvements

- Replace generic Exception catch with specific exceptions
- Remove empty catch blocks
- Improve method naming for readability
- Simplify boolean returns

No functional changes intended.


bclozel:
Thanks but we cannot process 2M+ lines PR that are obviously breaking the build and adding output files to the source tree. We do have quality checks in place and broadly applying a different opinion on our entire codebase is not likely to help.

PR #36449, opened 2026-03-11
Handle Kotlin nullable value class param correctly in `CoroutineUtils`

T45K:
as-is: under some conditions, spring framework throws an exception when calling a suspend function that has nullable value class param with a non-null value class object arg
to-be: spring framework can handle it correctly

reproducer: https://github.com/T45K/Spring-CoroutineUtils-bug-reproducer

root cause:
when `CoroutineUtils` is called to invoke such suspend functions with a non-null value class object arg, it tries to `box` the arg even though the type of the arg is already boxed value class.
that causes `IllegalArgumentException`

```
java.lang.IllegalArgumentException: argument type mismatch
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:108) ~[na:na]
	Suppressed: The stacktrace has been enhanced by Reactor, refer to additional information below: 
Error has been observed at the following site(s):
	*__checkpoint ⇢ Handler io.github.t45k.trial.spring.repo.TrialController#get(Continuation) [DispatcherHandler]
	*__checkpoint ⇢ HTTP GET "/" [ExceptionHandlingWebHandler]
Original Stack Trace:
		at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:108) ~[na:na]
		at java.base/java.lang.reflect.Method.invoke(Method.java:565) ~[na:na]
		at kotlin.reflect.jvm.internal.calls.CallerImpl$Method.callMethod(CallerImpl.kt:97) ~[kotlin-reflect-2.2.21.jar:2.2.21-release-469]
		at kotlin.reflect.jvm.internal.calls.CallerImpl$Method$Static.call(CallerImpl.kt:106) ~[kotlin-reflect-2.2.21.jar:2.2.21-release-469]
		at kotlin.reflect.jvm.internal.calls.ValueClassAwareCaller.call(ValueClassAwareCaller.kt:209) ~[kotlin-reflect-2.2.21.jar:2.2.21-release-469]
		at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:151) ~[kotlin-reflect-2.2.21.jar:2.2.21-release-469]
		at org.springframework.core.CoroutinesUtils.box(CoroutinesUtils.java:177) ~[spring-core-7.0.5.jar:7.0.5]
[...]

PR #36665, opened 2026-04-17
Regression on value class parameter handling

sgerke-1L:
The fix in https://github.com/spring-projects/spring-framework/pull/36449 introduces a regression related to the handling of value class conversion. In particular, passing e.g. `UUID` for an argument of type `ValueClass(val value: UUID)` doesn't work anymore. This is required to make ValueClass parameters work, see added integration tests.

**Fix description**
- Partially reverts #36449 to fix the aforementioned regression
- Fixes the issue reported in #36449 by introducing a `!JvmClassMappingKt.getJavaClass(kClass).isInstance(arg)` guard in the appropriate places, and keeps the regression test in from #36449 
- Introduced integration tests that test the handling of Kotlin value classes and the interaction of optional request parameters with an optional value class argument

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
+++ b/spring-core/src/main/java/org/springframework/core/CoroutinesUtils.java
@@ -134,9 +134,10 @@ public abstract class CoroutinesUtils {
 								Object arg = args[index];
 								if (!(parameter.isOptional() && arg == null)) {
 									KType type = parameter.getType();
-									if (!type.isMarkedNullable() &&
+									if (!(type.isMarkedNullable() && arg == null) &&
 											type.getClassifier() instanceof KClass<?> kClass &&
-											KotlinDetector.isInlineClass(JvmClassMappingKt.getJavaClass(kClass))) {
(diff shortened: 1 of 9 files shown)


sdeleuze:
I will try to leverage this PR to fix this regression caused by #36449, the fix will have to be backported to 6.2.x.

sgerke-1L:
Thank you! Let me know if there is anything I can help with.

sdeleuze:
Were the integration tests required to exercise some broken code paths or were they just added as additional tests "in case of" ?

sgerke-1L:
`Suspending handler method with nullable value class request param` and `suspendingValueClass` exhibit the regression introduced in #36665 . The others were green already, and I added them to document what's currently working.

sdeleuze:
Please test `7.0.8-SNAPSHOT` to double check everything work as expected.

joaodias14:
@sdeleuze it works as expected on my side. Thanks!

sgerke-1L:
Today's 7.0.8-SNAPSHOT spring framework bom looks good to me. Thanks!

PR #36671, opened 2026-04-18
Align MockHttpServletRequest.isRequestedSessionIdValid() with Servlet 6.1

### LINKED ISSUE #36631: `MockHttpServletRequest.isRequestedSessionIdValid()` does not follow Servlet 6.1 Javadoc
### Affected version

All Spring Versions

### Summary

`MockHttpServletRequest.isRequestedSessionIdValid()` currently defaults to `true` instead of being calculated, which does not match [Jakarta Servlet 6.1 `HttpServletRequest.isRequestedSessionIdValid()`](https://jakarta.ee/specifications/servlet/6.1/apidocs/jakarta.servlet/jakarta/servlet/http/HttpServletRequest.html#isRequestedSessionIdValid()).

**Relevant API (Servlet 6.1):**

- [getRequestedSessionId()](https://jakarta.ee/specifications/servlet/6.1/apidocs/jakarta.servlet/jakarta/servlet/http/HttpServletRequest.html#getRequestedSessionId()) — `null` if the client did not specify a session ID.
- [isRequestedSessionIdValid()](https://jakarta.ee/specifications/servlet/6.1/apidocs/jakarta.servlet/jakarta/servlet/http/HttpServletRequest.html#isRequestedSessionIdValid()) — returns `false` if the client did not specify any session ID; otherwise reflects whether the requested id corresponds to a valid session in the current context.
- [changeSessionId()](https://jakarta.ee/specifications/servlet/6.1/apidocs/jakarta.servlet/jakarta/servlet/http/HttpServletRequest.html#changeSessionId()) — assigns a new id to the current session (the incoming requested id is then stale until the client is updated).

### Minimal sample

1. **No client session id** — expected per Javadoc: `isRequestedSessionIdValid()` is `false`; actual today: `true`.

```java
assertThat(request.isRequestedSessionIdValid()).isFalse(); // fails today
```
2. **Session id rotation** — after `changeSessionId()`, the requested id no longer matches the session; expected: `isRequestedSessionIdValid()` is `false`; actual today: still `true` unless manually toggled. This breaks Spring Security tests and we must work around it by extending `MockHttpServletRequest` so that `changeSessionId()` invokes `setRequestedSessionIdValid(false)`.

```java
request.setSession(session);
String previousRequestedId = request.getRequestedSessionId();
request.changeSessionId();
assertThat(session.getId()).isNotEqualTo(previousRequestedId);
assertThat(request.getRequestedSessionId()).isEqualTo(previousRequestedId);
assertThat(request.isRequestedSessionIdValid()).isFalse(); // fails today
```

### Additional Considerations

At times users are not using `MockHttpServletRequest` directly. For example, when using `MockMvc` users are leveraging the builders provided with `MockMvc` and thus cannot easily extend/override `MockHttpServletRequest`.

### Possible Solution

While it is going to cause breaking changes, one solution might be to change the member variable (leave the accessors) `isRequestedSessionIdValid` to be a `Boolean`. If it is `null` (default), then the logic for `isRequestedSessionIdValid` should align with the Javadoc. If`isRequestedSessionIdValid` is explicitly set, then use the explicitly set value.

tianhaocui:
Fixes #36631

## Summary
`MockHttpServletRequest.isRequestedSessionIdValid()` currently defaults to `true`, which does not match the Servlet 6.1 specification.

## Changes
- Changed `requestedSessionIdValid` field from `boolean` (default `true`) to `@Nullable Boolean` (default `null`)
- When not explicitly set via `setRequestedSessionIdValid()`, `isRequestedSessionIdValid()` now calculates the return value by checking if the requested session ID matches the current session's ID
- When explicitly set, the explicit value is used (backward compatible)

## Behavior
- No client session ID (`getRequestedSessionId()` returns `null`) -> `isRequestedSessionIdValid()` returns `false`
- After `changeSessionId()`, requested ID no longer matches -> returns `false`
- `setRequestedSessionIdValid(true/false)` still works as before for explicit control

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
@@ -250,7 +250,7 @@ public class MockHttpServletRequest implements HttpServletRequest {
 
 	private @Nullable HttpSession session;
 
-	private boolean requestedSessionIdValid = true;
+	private @Nullable Boolean requestedSessionIdValid;
 
 	private boolean requestedSessionIdFromCookie = true;
 
@@ -1348,9 +1348,25 @@ public class MockHttpServletRequest implements HttpServletRequest {
 		this.requestedSessionIdValid = requestedSessionIdValid;
 	}
 
+	/**
+	 * Return whether the requested session ID is still valid.
+	 * <p>If {@link #setRequestedSessionIdValid} has not been called explicitly,
+	 * this method calculates the value based on whether the
+	 * {@linkplain #getRequestedSessionId() requested session ID} matches the
+	 * ID of the {@linkplain #getSession(boolean) current session}, aligning
+	 * with the Servlet 6.1 specification.
+	 */
 	@Override
 	public boolean isRequestedSessionIdValid() {
-		return this.requestedSessionIdValid;
+		if (this.requestedSessionIdValid != null) {
+			return this.requestedSessionIdValid;
+		}
+		String requestedId = getRequestedSessionId();
+		if (requestedId == null) {
+			return false;
+		}
+		HttpSession currentSession = getSession(false);
+		return (currentSession != null && requestedId.equals(currentSession.getId()));
 	}
 
 	public void setRequestedSessionIdFromCookie(boolean requestedSessionIdFromCookie) {


sbrannen (changes requested):
This PR is missing tests, so please introduce appropriate tests which fail before the proposed changes and pass after the proposed changes.

Perhaps you'd like to introduce your own or collaborate with the author of #36671.

bclozel:

PR #36672, opened 2026-04-19
Include cross-parameter violations in hasErrors() and getAllErrors()

### LINKED ISSUE #36666: `MethodValidationAdapter` ignores cross-parameter violations when `adapt-constraint-violations` is enabled
**Description:**
In a Spring MVC REST Controller, when a cross-parameter validation constraint (annotated with `@SupportedValidationTarget(ValidationTarget.PARAMETERS)`) is applied to a method, the validation fails to interrupt the request flow if `spring.validation.method.adapt-constraint-violations` is set to `true`.

The validator is correctly invoked and returns `false`, but no exception is triggered, and the controller method is executed anyway.

**Analysis:**
The issue resides in how `MethodValidator` handles the `MethodValidationResult`.

When a cross-parameter constraint fails, `result.getParameterValidationResults()` is empty, while `result.getCrossParameterValidationResults()` contains the violations. 
However, the current logic only seems to trigger an exception based on individual parameter results.

Specifically, in the current Spring Framework 6.2.15 source `MethodValidator#applyArgumentValidation`, the `MethodValidationException` is triggered only when `result.getParameterValidationResults()` is not empty (checked through `result.hasError()`), the method `applyArgumentValidation`  never check `result.getCrossParameterValidationResults()`.

**Steps to reproduce:**
1. Create a REST Controller with a method-level cross-parameter constraint.
2. Set `spring.validation.method.adapt-constraint-violations=true` in `application.yml`.
3. Invoke the endpoint with arguments that trigger a validation failure.
4. **Observed:** The controller method is executed (returns HTTP 200).
5. **Expected:** A `MethodValidationException`  should be thrown before the controller method is invoked.

Please see https://github.com/zerikv/spring-issue-36666, this repo contains a minimal spring-boot project which reproduce the problem.

**Environment:**
- Spring Boot: 3.5.10
- Spring Framework: 6.2.15
- JDK: 21

tianhaocui:
Closes #36666

`MethodValidationResult.hasErrors()` only checked `getParameterValidationResults()`, completely ignoring cross-parameter constraint violations. Similarly, `getAllErrors()` only collected errors from per-parameter results.

When `adapt-constraint-violations` is enabled, cross-parameter violations are stored separately via `getCrossParameterValidationResults()`. This caused `hasErrors()` to return `false` and `getAllErrors()` to return an empty list even when cross-parameter violations were present, effectively swallowing these errors.

**Changes:**
- `hasErrors()` now also checks `getCrossParameterValidationResults()`
- `getAllErrors()` now includes cross-parameter errors via `Stream.concat`

### DIFF
--- a/spring-context/src/main/java/org/springframework/validation/method/MethodValidationResult.java
+++ b/spring-context/src/main/java/org/springframework/validation/method/MethodValidationResult.java
@@ -19,6 +19,7 @@ package org.springframework.validation.method;
 import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Stream;
 
 import org.springframework.context.MessageSourceResolvable;
 import org.springframework.validation.Errors;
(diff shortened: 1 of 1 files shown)


rstoyanchev:
Thanks for the change but this issue requires a deeper review, and is not open for contribution.

PR #36673, opened 2026-04-19
Set host header consistently in STOMP relay CONNECT frames

### LINKED ISSUE #36663: Consistently set host header in StompRelayMessageBrokerHandler

tianhaocui:
Closes #36663

`StompBrokerRelayMessageHandler` only set the `host` header in CONNECT frames when `virtualHost` was explicitly configured. Per STOMP 1.2 specification, the `host` header is required on CONNECT frames.

This change falls back to `relayHost` (the TCP connection target) when `virtualHost` is not configured, ensuring the `host` header is always present in both system session and client session CONNECT frames.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompBrokerRelayMessageHandler.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/stomp/StompBrokerRelayMessageHandler.java
@@ -456,10 +456,7 @@ public class StompBrokerRelayMessageHandler extends AbstractBrokerMessageHandler
 		accessor.setLogin(this.systemLogin);
 		accessor.setPasscode(this.systemPasscode);
 		accessor.setHeartbeat(this.systemHeartbeatSendInterval, this.systemHeartbeatReceiveInterval);
-		String virtualHost = getVirtualHost();
-		if (virtualHost != null) {
-			accessor.setHost(virtualHost);
-		}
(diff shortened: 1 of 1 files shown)


twiedow:
I don't know if this also lead to problem like mine in other cases: Due to this change, the client could not connect to the external broker (RabbitMQ in my case) and responded with error `Virtual host 'localhost' access denied`.

RabbitMQ has a standard virtual host configured with name `/`. The RabbitMQ STOMP docs says that if there is no host set in the headers it defaults to `/`.

In order to fix this error I have to set the virtual host in the `StompBrokerRelayRegistration` explicitly to `/`.

In my eyes this change is a breaking change and should have been mentioned somewhere.

rstoyanchev:
@twiedow could you create a separate issue, and we will address it.

lukas0krupa:
Not sure if this is relevant, but I just want to ask if this is intentional.

After upgrading to version Spring 7, client can't fails to connect to RabbitMQ with CONNECT frame as `host` header from Message is overridden by code below. In Spring 6 this worked. Am I missing something?

`
Line 582: stompHeaderAccessor.setHost(getVirtualHost() != null ? getVirtualHost() : null);
`

Because line above sets Host to NULL, defaults to '/', what is not good, as that user has restricted access to `host` send with CONNECT frame.

After couple days, I finally find it and I have workaround, but not sure if this is a way to go + don't want to end up with workarounds being in my code forever :-).

Help would be appreciated. I couldn't create an issue and this was the only way to get feedback.

PR #36675, opened 2026-04-20
Remove singleton cache in HandlerMethod.createWithResolvedBean()

### LINKED ISSUE #36612: Issue with HandlerMethod#createWithResolvedBean after bean destruction
## Bug Report

Following this commit 6162d89042db2d11f3d1fe1e5896f8ded65d96a1, my application does not behave as expected after destroying the bean.

Instead of recreating a new bean and a new `HandlerMethod`, the cached value `resolvedBeanHandlerMethod` with the old bean is returned.

## Enhancements request

`DefaultSingletonBeanRegistry#destroySingleton()` should reset this cached value.

tianhaocui:
Closes #36612

The `resolvedBeanHandlerMethod` cache introduced a regression: after a singleton bean is destroyed via `destroySingleton()`, the cached `HandlerMethod` still holds a reference to the old (destroyed) bean instance. Subsequent requests return the stale handler instead of resolving the bean again through the `BeanFactory`.

This removes the cache entirely. For singletons, `BeanFactory.getBean()` is a `ConcurrentHashMap` lookup with negligible overhead, so the cache provided no meaningful performance benefit while introducing a correctness issue.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
+++ b/spring-web/src/main/java/org/springframework/web/method/HandlerMethod.java
@@ -96,8 +96,6 @@ public class HandlerMethod extends AnnotatedMethod {
 
 	private @Nullable HandlerMethod resolvedFromHandlerMethod;
 
-	private volatile @Nullable HandlerMethod resolvedBeanHandlerMethod;
-
 	private final String description;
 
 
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks for the proposal but this issue is already assigned and we don't intend to revert the cache completely.

tianhaocui:
Closing this — understood the issue is already assigned internally and a more granular invalidation approach is planned. Thanks for the feedback!

PR #36678, opened 2026-04-20
Use `String#replace` instead of `String#replaceAll` where appropriate

shenjianeng:
Avoid using `String#replaceAll` when the pattern is not a regular expression.

Using `java.lang.String#replace(CharSequence, CharSequence)` will improve performance.


sbrannen:

Thanks

PR #36697, opened 2026-04-23
Expose validateIfApplicable and isBindExceptionRequired as protected extension points in ModelAttributeMethodArgumentResolver

### LINKED ISSUE #36690: ModelAttributeMethodArgumentResolver should expose validateIfApplicable and isBindExceptionRequired as protected extension points
## Describe the use case
When migrating from Spring MVC to WebFlux, a common customization pattern is to subclass `ModelAttributeMethodProcessor` and override `validateIfApplicable` to apply group-based or context-aware validation — for example, selecting different `@Validated` groups based on the HTTP method, tenant context, or feature flags. The same need exists in WebFlux, but `ModelAttributeMethodArgumentResolver` does not support it.

Similarly, `isBindExceptionRequired` is overridden in MVC subclasses to control whether a binding/validation failure should throw or be silently passed to a `BindingResult` parameter — this decision sometimes needs to depend on runtime state, not just parameter inspection.

## What is missing
`ModelAttributeMethodProcessor` (MVC) exposes both as protected extension points:
```java
protected void validateIfApplicable(WebDataBinder binder, MethodParameter parameter)
protected boolean isBindExceptionRequired(WebDataBinder binder, MethodParameter parameter)
protected boolean isBindExceptionRequired(MethodParameter parameter)
```
In `ModelAttributeMethodArgumentResolver` (WebFlux), `validateIfApplicable` is private and `isBindExceptionRequired` does not exist at all — the equivalent logic is inlined as a private `hasErrorsArgument` helper. Neither can be overridden by subclasses.

## Proposed solution
Change validateIfApplicable from private to protected:
```java
protected void validateIfApplicable(WebExchangeDataBinder binder,
                                     MethodParameter parameter,
                                     ServerWebExchange exchange)
```
Introduce isBindExceptionRequired as a protected method:
```java
protected boolean isBindExceptionRequired(MethodParameter parameter)
```
The existing private `hasErrorsArgument(MethodParameter)` can be replaced by this method with no behavioral change.

## Precedent in the same class
PR #24947 already established this pattern by making `bindRequestParameters` protected in `ModelAttributeMethodArgumentResolver`. The class currently exposes two protected reactive extension points:
```java
protected Mono<Void> constructAttribute(WebExchangeDataBinder binder, ServerWebExchange exchange)
protected Mono<Void> bindRequestParameters(WebExchangeDataBinder binder, ServerWebExchange exchange)
```
`validateIfApplicable` sits at the same level in the binding lifecycle and should follow the same convention. The visibility change for `validateIfApplicable` keeps the synchronous signature consistent with how it is called internally — inside a `.doOnSuccess()` — which matches the existing contract. A separate reactive variant can be considered later if async validation becomes a requirement.
No behavioral change
Both changes are purely additive. Existing behavior is unaffected for any code that does not subclass `ModelAttributeMethodArgumentResolver`.

Happy to open a PR about it if the proposed solution is fine and also if there is any other approach you may want to have for this case I will be happy to help

Arbaz4Sayyad:
Expose validateIfApplicable and isBindExceptionRequired as protected extension points in ModelAttributeMethodArgumentResolver

This change aligns WebFlux with Spring MVC by making validation-related
methods protected extension points, allowing subclasses to customize
validation behavior.

**Motivation:**
When migrating from Spring MVC to WebFlux, a common customization pattern is to
subclass ModelAttributeMethodProcessor and override validateIfApplicable to apply
group-based or context-aware validation — for example, selecting different @Validated
groups based on the HTTP method, tenant context, or feature flags. The same need
exists in WebFlux, but ModelAttributeMethodArgumentResolver does not support it.

Similarly, isBindExceptionRequired is overridden in MVC subclasses to control whether
a binding/validation failure should throw or be silently passed to a BindingResult
parameter — this decision sometimes needs to depend on runtime state, not just
parameter inspection.

**Changes:**
- Changed validateIfApplicable from private to protected with Javadoc
- Introduced isBindExceptionRequired as protected method with Javadoc
- Replaced private hasErrorsArgument helper with isBindExceptionRequired
- Updated call site to use isBindExceptionRequired
- Added test to verify protected methods can be overridden

**Precedent:**
This follows the same pattern established in PR #24947, which made bindRequestParameters
protected in ModelAttributeMethodArgumentResolver. The class currently exposes two
protected reactive extension points (constructAttribute and bindRequestParameters).
validateIfApplicable sits at the same level in the binding lifecycle and should
follow the same convention.

**Testing:**
- All existing tests pass
- Added new test [protectedMethodsCanBeOverridden](cci:1://file:///d:/Projects/spring-framework/spring-webflux/src/test/java/org/springframework/web/reactive/result/method/annotation/ModelAttributeMethodArgumentResolverTests.java:402:1-416:2) to verify the protected methods
  can be successfully overridden by subclasses

**Behavioral Impact:**
No behavioral changes - purely additive changes that enable extension points for
subclasses. Existing behavior is unaffected for any code that does not subclass
ModelAttributeMethodArgumentResolver.

Fixes #36690

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/ModelAttributeMethodArgumentResolver.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/method/annotation/ModelAttributeMethodArgumentResolver.java
@@ -144,7 +144,7 @@ public class ModelAttributeMethodArgumentResolver extends HandlerMethodArgumentR
 									return adapter.fromPublisher(mono);
 								}
 								else {
-									if (errors.hasErrors() && !hasErrorsArgument(parameter)) {
+									if (errors.hasErrors() && isBindExceptionRequired(parameter)) {
 										throw new WebExchangeBindException(parameter, errors);
 									}
 									return attribute;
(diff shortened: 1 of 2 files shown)


bclozel:
We didn't get much requests for this. I understand the appeal for consistency but we don't necessarily need to replicate the entire arrangement in webflux. We can reconsider later if the community needs it. Thanks.

PR #36715, opened 2026-04-27
Support HTTP Interface clients in test framework

### LINKED ISSUE #36674: Support HTTP Interface Clients in Test Framework
From: https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-http-service-client

[start quote]

On the server side, an `@Controller` class can implement the same interface to handle requests with [@HttpExchange](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller/ann-requestmapping.html#mvc-ann-httpexchange-annotation) controller methods.

[end quote]

At the moment, there are  `WebTestClient` and `RestTestClient` for REST Testing. However, it may be beneficial to use HTTP Interface clients for testing where  `WebTestClient` and `RestTestClient` assertions are not needed.

At the moment, I see no way to instantiate HTTP Interface Clients and bind them to test Web Server in conjunction with `SpringBootTest.WebEnvironment.RANDOM_PORT`.

Would you be interested in providing any kind of support for using HTTP Interface Clients within Test Context?

At least consider providing `HttpExchangeAdapter` for `WebTestClient` and `RestTestClient`.

Thanks in advance.

devareddy05:
## Summary

Adds `RestTestClientAdapter` and `WebTestClientAdapter` so an `@HttpExchange` interface can be invoked against a `RestTestClient` or `WebTestClient`, including mock-server setups. Reference docs updated with a short example.

Closes gh-36674

## Test plan

- New tests for both adapters covering GET, `ResponseEntity`, path variables, POST body, and headers.
- `./gradlew :spring-test:check` passes locally.

### DIFF
--- a/framework-docs/modules/ROOT/pages/integration/rest-clients.adoc
+++ b/framework-docs/modules/ROOT/pages/integration/rest-clients.adoc
@@ -1029,6 +1029,26 @@ to a client application, what input and output types to create, what endpoint me
 signatures are needed, what Javadoc to have, and so on. The resulting Java API guides and
 is ready to use.
 
+For testing, an HTTP Interface Client can also be backed by `WebTestClient` or
+`RestTestClient` so that an `@HttpExchange` interface can be invoked against a mock
+server (for example a controller bound via `WebTestClient.bindToController(...)` or
+`RestTestClient.bindToController(...)`):
+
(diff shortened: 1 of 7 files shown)


rstoyanchev:
Closing for now as per discussion under the issue.

PR #36719, opened 2026-04-28
Deprecate setAwaitTerminationSeconds in favor of setAwaitTerminationMillis

quaff:
We should encourage using milliseconds over seconds for better precision, see https://github.com/spring-projects/spring-boot/pull/50225


DhruvTheDev1 (commented):
This look like a solid move! Since we're seeing the precision loss it makes sense to deprecate `setAwaitTerminationSeconds` in favour of the millisecond based setter.

PR #36721, opened 2026-04-28
Fix incorrect Java version reference in overview docs

Margomvi:
Fix incorrect Java version reference in overview documentation

The overview section currently states:

"As of Spring Framework 6.0, Spring requires Java 17+."

This is incorrect for the current documentation version and should refer to Spring Framework 7.0 instead.

This PR updates the text accordingly.

This is a documentation-only change.


bclozel:
Thanks for your proposal @Margomvi 

The documentation is right, we require Java 17 as of Spring Framework 6. Applying this change would imply that this baseline change happened for Spring Framework 7.0, which is not true. We'll leave things as they are.

PR #36733, opened 2026-04-30
Fix MimeType parser to unquote quoted parameter values

### LINKED ISSUE #36730: Parsing failure for MIME type with quoted parameter values
As seen in #36729, the `MimeType` parser in `MimeTypeUtils` does not always parse properly parameter values when they are quoted.

Changes discussed in #36729 are too risky for maintenance releases, so we should fix that problem separately in existing branches.

husseinvr97:
Closes #36730

## Problem

`MimeTypeUtils.parseMimeTypeInternal()` stored parameter values verbatim, including surrounding double quotes. Per RFC 2045 §5.1, `charset=utf-8` and `charset="utf-8"` are semantically equivalent — but `getParameter("charset")` returned `"utf-8"` (with quotes), causing `Charset.forName()` to throw `UnsupportedCharsetException` for quoted charset values. Backslash-escaped characters inside quoted-strings (RFC 9110 §5.6.4) were also not unescaped.

## Fix

The fix is applied in `createParametersMap()` in `MimeType`. After`checkParameters()` validates the raw value — which still sees the quoted form, so `isQuotedString()` returns true and `checkToken()` is correctly skipped for values containing non-token characters — `unquote()` is called before `map.put()`. Values are stored clean. `getParameter()` and `getParameters()` both return unquoted values without any additional indirection.

`unquote()` is upgraded from a simple quote-stripper to a full RFC 9110 §5.6.4 implementation that also unescapes quoted pairs (`\"` → `"`, `\\` → `\`).

`appendTo()` is updated with a `requiresQuoting()` helper that detects values containing non-token characters, and a `quoteValue()` helper that re-quotes and re-escapes them on serialization. This preserves a valid RFC round-trip: `toString()` output is always a parseable MIME type string.

The three `resolvedCharset` initialization sites (main constructor, copy-constructor, `readObject()`) are updated to read directly from the already-clean stored value, removing the now-redundant `unquote()` call.

## Behavior changes

**`getParameter()` now returns unquoted values.** Any caller that was previously stripping quotes manually from the return value will now double-process the result. Given that the old behavior was incorrect, this is an intentional breaking change.

**Normalization.** Quoted-pairs are unescaped on parse and re-escaped on serialize. `parseMimeType("foo/bar;param=\"\\\\\"").toString()` produces `foo/bar;param="\\"` — semantically equivalent, not byte-for-byte identical.
This is consistent with how conforming HTTP implementations handle quoted-string normalization.

**Single-quoted values are not unquoted.** Single quotes are not a quoted-string delimiter per RFC 2045 or RFC 9110 — they are ordinary token characters. The existing leniency for single-quoted parameters (SPR-8917) is preserved: single-quoted values pass `checkToken()` via `isQuotedString()` returning true, and are stored and returned with single quotes intact. No downstream caller in `spring-web` or `spring-messaging` passes the result of `getParameter("charset")` directly to `Charset.forName()` — charset resolution routes entirely through `getCharset()` / `resolvedCharset`.

Note: `ContentDisposition` has its own independent quote-stripping logic for `Content-Disposition` header parsing — it was examined and is not affected by this change.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/MimeType.java
+++ b/spring-core/src/main/java/org/springframework/util/MimeType.java
@@ -154,11 +154,11 @@ public class MimeType implements Comparable<MimeType>, Serializable {
 		checkToken(type);
 		checkToken(subtype);
 		this.type = type.toLowerCase(Locale.ROOT);
-		this.subtype = subtype.toLowerCase(Locale.ROOT);
-		this.parameters = createParametersMap(parameters);
-		if (this.parameters.containsKey(PARAM_CHARSET)) {
-			this.resolvedCharset = Charset.forName(unquote(this.parameters.get(PARAM_CHARSET)));
-		}
+        this.subtype = subtype.toLowerCase(Locale.ROOT);
+        this.parameters = createParametersMap(parameters);
+        if (this.parameters.containsKey(PARAM_CHARSET)) {
+            this.resolvedCharset = Charset.forName(this.parameters.get(PARAM_CHARSET));
+        }
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the contribution but the issue is already assigned and I've got a different approach in the works.

PR #36597, opened 2026-04-04
Support JSON array decoding to `Flux` in `KotlinSerializationJsonDecoder`

dmitrysulman:
Fixes #36187

PR #36735, opened 2026-04-30
Add caching for Kotlin list serializers

dmitrysulman:
Changes made in 22bcac1eb5570dc4aac63b25a380c6f07fceeb0e introduced the creation of a new list serializer instance in `KotlinSerializationJsonDecoder` on each `decode()` invocation. This PR adds list serializer caching, similar to the existing element serializer caching, to address this.

See https://github.com/spring-projects/spring-framework/pull/36597

cc @sdeleuze


sdeleuze:
The creation of such list deserializer in straightforward and does not involve expensive logic as far as I can tell, so I don't think such caching is needed. If you have strong evidence of the opposite, please share.

PR #36739, opened 2026-05-01
Honor BeanDefinition order attribute for DispatcherServlet strategies

### LINKED ISSUE #36637: Setting order programmatically on OncePerRequestFilter or HandlerExceptionResolver beans using BeanRegistrar or GenericApplicationContext does not work
If you annotate your `OncePerRequestFilter` or `HandlerExceptionResolver` beans with `@Order` annotation that is registered programmatically using  `BeanRegistrar` or `GenericApplicationContext`, it works correctly. Order value provided in `@Order` annotation influences order of filter or resolver.

However if you instead provide order value in `registerBean`'s `order` parameter in case of `BeanRegistrar` or if you set `AbstractBeanDefinition.ORDER_ATTRIBUTE` in case of `GenericApplicationContext` , this value has no influence on filter or resolver order.

SpringBoot 4.0.3
JDK 25
Kotlin 2.3

daguimu:
## Problem

`DispatcherServlet` detects four kinds of strategy beans (`HandlerMapping`, `HandlerAdapter`, `HandlerExceptionResolver`, `ViewResolver`) by type and keeps each list sorted. The sort calls `AnnotationAwareOrderComparator.sort(list)` directly on the matched bean instances, which only consults `@Order` on the bean class and the `Ordered` interface.

Setting the order programmatically — via `BeanRegistrar.Spec#order(int)`, the `order` parameter on `GenericApplicationContext#registerBean(...)`, or `AbstractBeanDefinition#ORDER_ATTRIBUTE` directly on a bean definition — has no effect on these strategy lists, even though the same attribute is honored for sorted dependency injection through `DefaultListableBeanFactory.FactoryAwareOrderSourceProvider`.

## Root cause

`AnnotationAwareOrderComparator.sort(List)` runs without an `OrderSourceProvider`. With no source provider, the comparator never reaches the bean definition and cannot read `ORDER_ATTRIBUTE`.

## Fix

Add a private helper `sortStrategyBeans` that sorts the matched list using `AnnotationAwareOrderComparator.INSTANCE.withSourceProvider(...)`. The `BeanDefinitionOrderSourceProvider` mirrors `FactoryAwareOrderSourceProvider` exactly:

- Looks up the merged `BeanDefinition` via `ConfigurableApplicationContext.getBeanFactory().getMergedBeanDefinition(...)` (which already walks the parent BeanFactory chain).
- Adds the `ORDER_ATTRIBUTE` value as an `Ordered` source when present and `Integer`.
- Adds the factory method as a source for `RootBeanDefinition` instances (so `@Order` on a `@Bean` factory method works).
- Adds the declared target type as a source when distinct from the bean's runtime class.
- Throws `IllegalStateException` for a non-`Integer` `ORDER_ATTRIBUTE`, matching the bean factory's fail-fast behavior.

The helper is applied at all four strategy-detection sites. Beans without `ORDER_ATTRIBUTE` see no behavior change — the comparator still falls back to `@Order` / `Ordered`.

## Tests Added

| Change Point | Test |
|--------------|------|
| `sortStrategyBeans` honors `ORDER_ATTRIBUTE` | `detectsHandlerMappingsOrderedByBeanDefinitionOrderAttribute` |
| `ORDER_ATTRIBUTE` overrides class-level `@Order` | `beanDefinitionOrderAttributeOverridesAnnotationOrderForHandlerMappings` |
| `ORDER_ATTRIBUTE` resolved across parent contexts | `handlerMappingOrderFromBeanDefinitionInheritsAcrossParentContext` |
| Non-`Integer` `ORDER_ATTRIBUTE` is rejected | `nonIntegerOrderAttributeIsRejected` |

The tests exercise the change through `getHandlerMappings()` but cover the shared helper used by all four strategies. The full `:spring-webmvc:test` suite passes.

## Impact

- Scope: `spring-webmvc` only; private API, no public surface added.
- Behavior change is strictly additive for the common case.
- Fail-fast path is only reachable when user code sets `ORDER_ATTRIBUTE` to a non-`Integer` directly; public APIs only accept `int`/`Integer`.
- Default-strategy fallback (`DispatcherServlet.properties`) is unaffected.
- WebFlux's `DispatcherHandler` and `OncePerRequestFilter` ordering are out of scope.

Closes gh-36637

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java
@@ -17,12 +17,14 @@
 package org.springframework.web.servlet;
 
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -42,9 +44,14 @@ import org.jspecify.annotations.Nullable;
 import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.support.AbstractBeanDefinition;
+import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.i18n.LocaleContext;
+import org.springframework.core.OrderComparator;
+import org.springframework.core.Ordered;
 import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.support.PropertiesLoaderUtils;
@@ -513,7 +520,7 @@ public class DispatcherServlet extends FrameworkServlet {
 			if (!matchingBeans.isEmpty()) {
 				this.handlerMappings = new ArrayList<>(matchingBeans.values());
 				// We keep HandlerMappings in sorted order.
-				AnnotationAwareOrderComparator.sort(this.handlerMappings);
+				sortStrategyBeans(context, matchingBeans, this.handlerMappings);
 			}
 		}
 		else {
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Note that we've had to re-write commits in main and 7.0.x over the past 2 days after erroneous merge that this PR was based on. The PR needs to be handled with care focusing on just the one commit 7a05a9e97dbc3ddd7de525d072861cfacdc13927 and ignoring the other 4.

daguimu:
Thanks for the heads-up @rstoyanchev — yes, still in progress.

I've rebased onto the rewritten `main`. The four extraneous commits from the bad base were resolved automatically:
- `910117e` (webjars resource resolver) and `20059d3` (MIME quoted pairs) were detected as already-applied (they exist under new SHAs in the rewritten history) and skipped by `git rebase`.
- The two `Merge branch '7.0.x'` commits collapsed away as no-ops.

The branch now contains only the single contribution commit `3dcc741`, and CI should re-run on the new SHA. Diff vs. `main` is unchanged (`+194 / −4`, 2 files).

PR #36750, opened 2026-05-04
feat: Added an enum field for Content-Security-Policy response header…

PrittSpadeLord:
I have added the Content-Security-Policy response header as enum field. As per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy), the response header is considered baseline widely supported, so I figured it would make sense to get it added as an option to aid developers using this field.

Various Policy Directives are also defined in the same document with varying levels of support between browsers. I am not sure how to proceed with adding that kind of support into Spring Web at present moment, so I would like to use this PR as an opportunity to inquire about that as well!

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -225,6 +225,11 @@ public class HttpHeaders implements Serializable {
 	 */
 	public static final String CONTENT_RANGE = "Content-Range";
 	/**
+	 * The CORS {@code Content-Security-Policy} response header field name.
+	 * @see <a href="https://www.w3.org/TR/CSP3/">Content Security Policy Level 3 W3C recommendation</a>
+	 */
+	public static final String CONTENT_SECURITY_POLICY = "Content-Security-Policy";
+	/**
 	 * The HTTP {@code Content-Type} header field name.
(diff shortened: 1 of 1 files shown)


PrittSpadeLord:
Quick update, I was unsure if the header should be marked as a "HTTP", because this is not defined within [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231) nor [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110). I've set it as merely "Security" for now, since I don't know what else idiomatic way to phrase it would be.

bclozel:
Thanks for the proposal but we only add headers names here when we have a concrete use case in Spring Framework.
I believe Spring Security does support this aspect and I even think they do without having to expose this name as a public String!

In summary I don't think we need this. Thanks for this contribution!

PrittSpadeLord:
> Thanks for the proposal but we only add headers names here when we have a concrete use case in Spring Framework. I believe Spring Security does support this aspect and I even think they do without having to expose this name as a public String!
> 
> In summary I don't think we need this. Thanks for this contribution!

I see! I was not aware of the policy that Spring Framework only added http headers to this enum when it was explicitly used within the framework itself. When I cross referenced these enums with another application, pretty much everything was found here except Content-Security-Policy, which led me to the incorrect conclusion that Spring Framework intended to document every supported header here!

I will be more mindful when making proposals in the future, since I understand every new code adds maintenance burden to the project. Thanks for your swift reply!

PR #36763, opened 2026-05-08
Use CollectionUtils to initialize HashMap in DefaultUriBuilderFactory

shenjianeng:


DhruvTheDev1 (approved):
Solid micro-optimisation! 
`Map<String, Object> map = CollectionUtils.newHashMap(defaultUriVariables.size() + uriVars.size());`Ensures the map is pre-sized to handle the combined size of defaultUriVariables and uriVars without needing to rehash/resize. Logic remains consistent with the existing URI behaviour.

rstoyanchev (approved):

PR #36772, opened 2026-05-10
Replace thread-unsafe SimpleDateFormat with DateTimeFormatter in AbstractMockHttpServletRequestBuilder

SebTardif:
## Summary

Replace the static `SimpleDateFormat` instance with a thread-safe `DateTimeFormatter` in `AbstractMockHttpServletRequestBuilder`.

## Problem

A `static final SimpleDateFormat` is shared across all test request builder instances. `SimpleDateFormat.format()` mutates internal `Calendar` state and is not thread-safe. When tests run in parallel (e.g., with JUnit's `@Execution(CONCURRENT)`), concurrent calls to `header(name, dateValue)` can produce corrupt date strings or throw `ArrayIndexOutOfBoundsException`.

## Fix

Replace with `DateTimeFormatter.ofPattern(...)` using the same pattern (`EEE, dd MMM yyyy HH:mm:ss zzz`), locale (US), and timezone (UTC). `DateTimeFormatter` is immutable and thread-safe by design.


bclozel:
Thanks @SebTardif !

PR #36777, opened 2026-05-11
Allow specifying the charset to use in `ExchangeFilterFunctions#basicAuthentication`

kzander91:
This PR adds support for specifying the charset to use for encoding the credentials in `ExchangeFilterFunctions#basicAuthentication`.

I recently noticed that the existing method uses _ISO-8859-1_ to encode the credentials. In my app, this caused authentication failures when calling an API protected by Spring Security with a password that contains multi-byte characters. Spring Security uses _UTF-8_ by default to decode the credentials (see https://github.com/spring-projects/spring-security/blob/ee35c2c9ccd9f57d279d6bdef61d21c750bef15e/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationConverter.java#L49).

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ExchangeFilterFunctions.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/ExchangeFilterFunctions.java
@@ -17,9 +17,11 @@
 package org.springframework.web.reactive.function.client;
 
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
+import org.jspecify.annotations.Nullable;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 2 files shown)


sbrannen:
Indeed, `BasicAuthenticationInterceptor` and `HttpHeaders.encodeBasicAuth()` both already support a `Charset`.

kzander91:
@sbrannen I just noticed that the new `@since` tag was incorrect, I have just pushed a fix (7.0 -> 7.1).

sbrannen:
> @sbrannen I just noticed that the new `@since` tag was incorrect, I have just pushed a fix (7.0 -> 7.1).

We actually decided to include this in `7.0.x`. 😉

sbrannen:
This has been merged into `7.0.x` and `main` in 4bb20ea8db01ec9e8238dcd3ebc9d7f66f6e998b and slightly revised in 1e843fd3ecf21795153c8516dfc6f3e47b020ab5.

Thanks

PR #36807, opened 2026-05-14
Add dedicated exception for HttpStatus.PRECONDITION_FAILED

dominik-kovacs:
Add a dedicated `HttpClientErrorException.PreconditionFailed` subclass for HTTP 412 status code, consistent with the existing pattern for other 4xx status codes (NotFound, Gone, Conflict, etc.).

This allows callers to catch `HttpClientErrorException.PreconditionFailed` directly rather than inspecting the status code on a generic `HttpClientErrorException`.


rstoyanchev (commented):
An equivalent change to be made also in `WebClientResponseException`.

Issue #36779, opened 2026-05-11
KotlinConventions still has references to javadoc.io


snicoll:
I believe a decision was made to exclude javadoc.io from the build as it is unreliable. It is, however, still configuring in Kotlin conventions. This does not fail the build but produces a ton of warnings as it appears to be down at the moment.

pratx08:
Hey @snicoll , I have got rid off the javadoc.io. Please have a look at this [PR ](https://github.com/spring-projects/spring-framework/pull/36810)when you find time and let me know if any changes are needed. Thank you!

PR #36810, opened 2026-05-15
Remove javadoc.io links from Kotlin Conventions

pratx08:
Removed remaining `javadoc.io` links from Kotlin Dokka external documentation configuration.

Addresses [#36779](https://github.com/spring-projects/spring-framework/issues/36779)


bclozel:
Thanks for the proposal but this issue has not been discussed by the team and is not triaged.

PR #36816, opened 2026-05-20
Deprecate setDisallowedFields in DataBinder for removal

### LINKED ISSUE #36802: Deprecate disallowedFields property of DataBinder
As explained under #36789, for data binding we recommend objects that are one or both of:
- immutable (records or classes with a primary constructor)
- designed specifically for web binding with properties that match the expected inputs

For anything else it is essential to configure allowedFields or disallowedFields that control which setters can be used for binding. Use of the former is safer as it is an explicit opt-in while the latter is fragile (easy to missing something) especially over time.

We should deprecate the use of disallowedFields in 7.1 in favour of using either immutable objects with constructor binding, or object specifically designed for data binding (and therefore don't require such configuration), or the safer allowedFields config if necessary.

jhan0121:
This PR deprecates `setDisallowedFields()` and `getDisallowedFields()` in `DataBinder` as discussed in #36802.

## Changes

- Deprecate `setDisallowedFields()` and `getDisallowedFields()` in `DataBinder` with `@Deprecated(since = "7.1", forRemoval = true)`
- Remove the reference to `disallowedFields` as an alternative in `setAllowedFields()` Javadoc
- Update reference documentation to reflect the deprecated status
- Add `@SuppressWarnings("removal")` to affected test methods

Closes gh-36802

### DIFF
--- a/framework-docs/modules/ROOT/partials/web/web-data-binding-model-design.adoc
+++ b/framework-docs/modules/ROOT/partials/web/web-data-binding-model-design.adoc
@@ -28,8 +28,8 @@ For example:
 	}
 ----
 
-NOTE: It is also possible to configure `disallowedFields`, but that's fragile, and
-due to be https://github.com/spring-projects/spring-framework/issues/36802[deprecated] in Spring Framework 7.1.
+NOTE: It is also possible to configure `disallowedFields`, but that's fragile and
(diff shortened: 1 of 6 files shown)


bclozel:
Thanks for the help, but the issue is already assigned.

jhan0121:
Thanks for the quick response! I missed that the issue was already assigned. I will make sure to check the assignment status more carefully next time.

PR #36822, opened 2026-05-21
Fix client disconnect logged as error in WebFlux

### LINKED ISSUE #36811: AbortedException from client logged at ERROR level with WebFlux functional endpoint
**Version**: `org.springframework.boot:spring-boot-starter-webflux:3.5.8 / org.springframework:spring-web:6.2.14`

I have a service that uses Reactor Netty.

When client closes connection before sending full request body there is a log with ERROR level:
```log
2026-05-18T14:22:17.887+03:00 ERROR 9144 --- [tor-http-nio-11] o.s.w.s.adapter.HttpWebHandlerAdapter    : [e8392712-7] 500 Server Error for HTTP POST "/test"

reactor.netty.channel.AbortedException: Connection has been closed
	at reactor.netty.http.server.HttpServerOperations.onInboundClose(HttpServerOperations.java:931) ~[reactor-netty-http-1.2.12.jar:1.2.12]
	*__checkpoint ? HTTP POST "/test" [ExceptionHandlingWebHandler]
Original Stack Trace:
		at reactor.netty.http.server.HttpServerOperations.onInboundClose(HttpServerOperations.java:931) ~[reactor-netty-http-1.2.12.jar:1.2.12]
```

Which is a bit unexpected since it's a client-side issue related to connection, so there is seemingly no need for it to be logged as ERROR by default and have any status code. (Also undesirable because in my case there are alerts configured for any logs with ERROR level)

I'd expect it to be logged at WARN/DEBUG/TRACE and this [line](https://github.com/spring-projects/spring-framework/blob/0c25d817bdd7959f17a225cfdc92d6403284f13d/spring-web/src/main/java/org/springframework/web/server/adapter/HttpWebHandlerAdapter.java#L372) does exactly that, but `checkAndLogClientDisconnectedException` isn't executed since `response.setStatusCode(HttpStatus.INTERNAL_SERVER_ERROR)` returns `true`.

This looks like a bug since the code for "expected" behavior is there but is unreachable. If it's not a bug or a limitation of connectivity-related error handling, how can I safely provide my own handling for such case, via ExceptionHandler?

**Below are some snippets for reproducing this issue**

Main.java:
```java
@SpringBootApplication
public class Main {
        SpringApplication.run(Main.class, args);
    }

    @Bean
    RouterFunction<?> routerFunction() {
        return RouterFunctions
            .route()
            .POST(
                "/test",
                request -> request
                    .bodyToMono(String.class)
                    .flatMap(body -> ServerResponse.ok().build())
            ).build();
    }
}
```

build.gradle:
```gradle
plugins {
    id 'java'
    id 'org.springframework.boot' version '3.5.8'
    id 'io.spring.dependency-management' version '1.1.7'
}

java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(21)
    }
}

repositories {
    mavenCentral()
}

dependencies {
    implementation 'org.springframework.boot:spring-boot-starter-webflux'
}
```

python script for producing aborted request:
```python
import socket

HOST = "127.0.0.1"
PORT = 8080

request = (
    b"POST /test HTTP/1.1\r\n"
    b"Host: localhost\r\n"
    b"Content-Type: text/plain\r\n"
    b"Content-Length: 2\r\n"
    b"\r\n"
)

sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.connect((HOST, PORT))
sock.sendall(request)
sock.sendall(b"A")
sock.close()
```

suuuuuuminnnnnn:
In HttpWebHandlerAdapter.handleUnresolvedError(), when a client
disconnects before the response is committed (e.g., the client
sends a partial request body and closes the connection), the
resulting AbortedException arrives while setStatusCode(500)
still returns true. This causes the handler to immediately log
a 500 Server Error at ERROR level and return, making the
checkAndLogClientDisconnectedException branch unreachable.

This commit moves the disconnect check before the 500 status
assignment. DisconnectedClientHelper.isClientDisconnectedException
already excludes outbound/remote-call exception wrappers, so the
reordering does not affect server-side error handling.

Closes gh-36811

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/adapter/HttpWebHandlerAdapter.java
+++ b/spring-web/src/main/java/org/springframework/web/server/adapter/HttpWebHandlerAdapter.java
@@ -362,15 +362,13 @@ public class HttpWebHandlerAdapter extends WebHandlerDecorator implements HttpHa
 		ServerHttpResponse response = exchange.getResponse();
 		String logPrefix = exchange.getLogPrefix();
 
-		// Sometimes a remote call error can look like a disconnected client.
-		// Try to set the response first before the "isDisconnectedClient" check.
-
-		if (response.setStatusCode(HttpStatus.INTERNAL_SERVER_ERROR)) {
-			logger.error(logPrefix + "500 Server Error for " + formatRequest(request), ex);
+		// Check for client disconnect first; DisconnectedClientHelper already excludes remote-call wrappers.
+		if (disconnectedClientHelper.checkAndLogClientDisconnectedException(ex)) {
+			observationContext.setConnectionAborted(true);
 			return Mono.empty();
 		}
-		else if (disconnectedClientHelper.checkAndLogClientDisconnectedException(ex)) {
-			observationContext.setConnectionAborted(true);
+		else if (response.setStatusCode(HttpStatus.INTERNAL_SERVER_ERROR)) {
+			logger.error(logPrefix + "500 Server Error for " + formatRequest(request), ex);
 			return Mono.empty();
 		}
 		else {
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Thanks for the suggestion, but the fix is not as obvious as it seems, and also required deeper review.

PR #36823, opened 2026-05-21
Reject competing SpEL accessors

### LINKED ISSUE #32737: Reject competing read and write accessors for `PropertyAccessor` and `IndexAccessor` in SpEL
## Overview

While working on #26478 and related issues, it became apparent that a developer can inadvertently register one `PropertyAccessor` for reading a given property and a second competing accessor for writing the same property, which can lead to improper behavior when a cached accessor is expected to be able to support both read and write access for a given property.

The same can happen with multiple competing `IndexAccessor` implementations.

We have therefore decided to reject competing read and write accessors for the same property or the same index for both `PropertyAccessor` and `IndexAccessor`, respectively, starting with Spring Framework 6.2.

## Related Issues

- #26478

Will-thom:
This rejects competing read and write accessors for the same property or index when SpEL needs a read/write reference, such as compound assignments. It preserves read-only and write-only operations when there is no competing accessor.

Closes gh-32737

Tests:
- ./gradlew.bat --no-daemon :spring-expression:check

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/SpelMessage.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/SpelMessage.java
@@ -303,7 +303,11 @@ public enum SpelMessage {
 
 	/** @since 6.2 */
 	EXCEPTION_DURING_INDEX_WRITE(Kind.ERROR, 1084,
-			"A problem occurred while attempting to write index ''{0}'' in ''{1}''");
+			"A problem occurred while attempting to write index ''{0}'' in ''{1}''"),
+
+	/** @since 7.1 */
+	COMPETING_ACCESSORS(Kind.ERROR, 1085,
+			"Found competing read and write accessors for ''{0}'' in ''{1}''");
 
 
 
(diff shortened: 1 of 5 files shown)


Will-thom:
Moved this PR out of draft; it is ready for review now. Please let me know if there is anything else needed from my side.

sbrannen:
Hi @Will-thom,

Congratulations on submitting your first PR for the Spring Framework! 👍🏻

I sincerely appreciate the time and effort you put into this; however, after thoughtful consideration, I have decided not to introduce checks for competing read and write accessors in SpEL.

While reviewing your PR, I realized that your proposal would not address all use cases necessary to _properly_ implement such checks. To properly apply such checks, we would need to consistently check whether there are competing read and write accessors for every single access, since (in theory) accessors can come and go over the lifetime of an `EvaluationContext`. In addition, we'd ideally want to check for a competing write accessor for an expression that only reads, and vice versa.

But doing all of that would undermine the original goal associated with caching of accessors: to improve performance of SpEL expressions.

In light of that, I have decided to leave things as they are and instead to document the relationships between expressions, evaluation contexts, and accessors in order to better inform users, and I created the following issue to improve the documentation.

- #36968

Regards,

Sam

PR #36825, opened 2026-05-21
Close contexts when clearing test context cache

### LINKED ISSUE #26196: Properly cleanup ApplicationContexts in TestContext when all tests have been executed
When using ```@SpringBootTest``` with JUnit Jupiter a pitest user had a problem when executing tests with pitest (see https://github.com/hcoles/pitest/issues/827). As far as my analysis brought me the cause is Spring Frameworks TestContext caching.
Spring caches all application contexts loaded in a static variable and reused them among different tests and test classes (```TestContextManager```). Unfortunately Spring does not cleanup after all(!) tests have been run assuming the JVM will be terminated immediately after tests have been executed (which in turn triggers the automatic closing behavior for application contexts). This might hold for surefire or IDE starting those tests. But Pitest reuses the JVM instance for different test suite executions (multiple invocations of JUnit Jupiter's ```Launcher.execute()``` inside the same JVM) . Thus the cached application contexts stay there and interfere with future test runs.

On the other hand Testcontainers integrates with the JUnit Jupiter Engine and uses ClosableResource hook to cleanup containers afterwards.

I've created a JUnit extension cleaning up application contexts on shutdown as a proof of concept. See [SpringBootCleanup](https://github.com/StefanPenndorf/pitTestcontainers/blob/sprint-test-context-cleanup-poc/src/test/java/com/example/pitTestcontainers/pets/SpringBootCleanup.java) in the sample project originally provided by @jaguado-arima .

@DirtiesContext will have the same effect here and I didn't notice any differences in performance regarding this simple example. But I think it's a bad idea to add ```@DirtiesContext``` to all test classes only to be able to use pitest... if you're relying on application context caching to reduce overall testing time for a huge test suite adding ```@DirtiesContext``` to all tests for the sake of mutation testing will slow down test execution (Springs context cache has been implemented for a reason...).

@sbrannen  If you consider changing the behavior as suggested I can also provide a pull request implementing the necessary changes for ```SpringExtension```.

Will-thom:
This updates the TestContext Framework cache so that clearing the cache also closes cached ConfigurableApplicationContext instances instead of only dropping the internal references. The implementation reuses the existing removal path, preserving the hierarchy-aware close behavior already used by cache eviction/removal.

The ContextCache contract now documents the close behavior for clear(), and LruContextCacheTests covers both clear() and reset(), since reset() delegates to clear().

Closes gh-26196

Tests:
- `./gradlew.bat --no-daemon --max-workers=1 :spring-test:test --tests org.springframework.test.context.cache.LruContextCacheTests`
- `./gradlew.bat --no-daemon --max-workers=1 :spring-test:test --tests org.springframework.test.context.cache.ContextCacheTests`
- `./gradlew.bat --no-daemon --max-workers=1 :spring-test:check`

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/context/cache/DefaultContextCache.java
+++ b/spring-test/src/main/java/org/springframework/test/context/cache/DefaultContextCache.java
@@ -426,6 +426,9 @@ public class DefaultContextCache implements ContextCache {
 	@Override
 	public void clear() {
 		synchronized (this.contextMap) {
+			for (MergedContextConfiguration key : new ArrayList<>(this.contextMap.keySet())) {
+				remove(key, HierarchyMode.CURRENT_LEVEL);
+			}
 			this.contextMap.clear();
 			this.hierarchyMap.clear();
 			this.contextUsageMap.clear();
--- a/spring-test/src/test/java/org/springframework/test/context/cache/LruContextCacheTests.java
+++ b/spring-test/src/test/java/org/springframework/test/context/cache/LruContextCacheTests.java
@@ -81,6 +81,42 @@ class LruContextCacheTests {
 		assertThatIllegalArgumentException().isThrownBy(() -> new DefaultContextCache(0));
 	}
 
+	@Test
+	void clearClosesContexts() {
+		DefaultContextCache cache = new DefaultContextCache(4);
+
+		cache.put(fooConfig, key -> fooContext);
+		cache.put(barConfig, key -> barContext);
+		cache.put(bazConfig, key -> bazContext);
+		assertCacheContents(cache, "Foo", "Bar", "Baz");
+
+		cache.clear();
+		assertCacheContents(cache);
+
+		verify(fooContext, times(1)).close();
+		verify(barContext, times(1)).close();
+		verify(bazContext, times(1)).close();
+		verify(abcContext, never()).close();
+	}
+
+	@Test
+	void resetClosesContexts() {
+		DefaultContextCache cache = new DefaultContextCache(4);
+
+		cache.put(fooConfig, key -> fooContext);
+		cache.put(barConfig, key -> barContext);
+		cache.get(fooConfig);
+		assertThat(cache.getHitCount()).isEqualTo(1);
+		assertCacheContents(cache, "Bar", "Foo");
+
+		cache.reset();
+		assertCacheContents(cache);
+		assertThat(cache.getHitCount()).isZero();
+
+		verify(fooContext, times(1)).close();
+		verify(barContext, times(1)).close();
+	}
+
 
 	@Nested
 	@SuppressWarnings("deprecation")
--- a/spring-test/src/main/java/org/springframework/test/context/cache/ContextCache.java
+++ b/spring-test/src/main/java/org/springframework/test/context/cache/ContextCache.java
@@ -314,7 +314,10 @@ public interface ContextCache {
 	void reset();
 
 	/**
-	 * Clear all contexts from the cache, clearing context hierarchy information as well.
+	 * Clear all contexts from the cache, explicitly
+	 * {@linkplain org.springframework.context.ConfigurableApplicationContext#close() closing}
+	 * each context that is an instance of {@code ConfigurableApplicationContext}
+	 * and clearing context hierarchy information as well.
 	 */
 	void clear();
 


sbrannen:
Thanks for the PR, @Will-thom. 👍 

> Closes [gh-26196](https://github.com/spring-projects/spring-framework/issues/26196)

Please note that this change would only serve as a building block for #26196, since it would not address the broader topic raised in that issue.

In any case, it makes sense for `clear()` and `reset()` to actually close contexts instead of just removing them from the cache.

Also, please make sure you comply with the DCO requirement by _signing_ your commits. Otherwise, we cannot process this PR.

sbrannen:
@Will-thom, please address the [DCO sign-off issue](https://github.com/spring-projects/spring-framework/pull/36825/checks?check_run_id=77271493848).

> There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.
> 
> To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Will-thom:
Thanks for the reminder. The commit now includes the required Signed-off-by line, and the DCO check is passing on the latest PR head. Could you please take another look when you have a chance?

spring-test/src/test/java/org/springframework/test/context/cache/LruContextCacheTests.java
```diff
@@ -81,6 +81,42 @@ void maxCacheSizeZero() {
...
+		assertThat(cache.getHitCount()).isEqualTo(1);
+		assertCacheContents(cache, "Bar", "Foo");
+
+		cache.reset();
+		assertCacheContents(cache);
+		assertThat(cache.getHitCount()).isZero();
+
+		verify(fooContext, times(1)).close();
+		verify(barContext, times(1)).close();
+	}
```
sbrannen, line 118: There are a couple of things I'd like addressed on the test side before merging:

1. **Hierarchy coverage.** These two new tests only cover flat, non-hierarchical configs (`config()` always passes a `null` parent). Since this change goes through the hierarchy-aware `remove()` path, could you please add a test that builds an actual multi-level `@ContextHierarchy` (similar to the `removeContextHierarchyCacheLevel*()` tests in `ContextCacheTests`) and asserts that `clear()`/`reset()` closes every level correctly?

2. **Invariant coverage.** Neither of the new tests verifies that `hierarchyMap` and `contextUsageMap` actually end up empty — only `contextMap` is checked, via `assertCacheContents()`. Since those two structures are exactly what this change touches inside `remove()`, please also assert `getParentContextCount()` and `getContextUsageCount()` return to `0` after `clear()`/`reset()` – for example, using the existing `assertContextCacheStatistics()` / `assertParentContextCount()` helpers already used in `ContextCacheTests`. For `resetClosesContexts()`, it'd also be good to assert `getMissCount()` is `0` for symmetry with the existing `getHitCount()` check.

This would give us a real regression guard on these new code paths, rather than only checking that the mock contexts were closed.

----

As side note, I think using `Mockito.inOrder()` for the verification would also serve us well to ensure the contexts are closed in the expected order within a context hierarchy.

However, `inOrder` should only assert order along a single ancestor → descendant chain (e.g. child → parent). It shouldn't assert anything about the relative order of *sibling* contexts, since children of the same parent are stored in a `HashSet` and closed in unspecified order. In other words, asserting sibling order would result in a flaky test.

spring-test/src/main/java/org/springframework/test/context/cache/DefaultContextCache.java
```diff
@@ -426,6 +426,9 @@ public void reset() {
 	@Override
 	public void clear() {
 		synchronized (this.contextMap) {
+			for (MergedContextConfiguration key : new ArrayList<>(this.contextMap.keySet())) {
+				remove(key, HierarchyMode.CURRENT_LEVEL);
+			}
 			this.contextMap.clear();
```
sbrannen, line 432: After this new loop, the four subsequent calls to `this.contextMap.clear()`, `this.hierarchyMap.clear()`, `this.contextUsageMap.clear()`, and `this.unusedContexts.clear()` are now redundant: the loop's calls to `remove()` should already empty all four as an invariant, since every key in the pre-loop snapshot ends up removed either directly or as a descendant.

Could you please remove those `clear()` calls and ensure the tests cover those invariants (see other comment regarding tests)?

sbrannen (changes requested):
Thanks for the update, @Will-thom. The DCO check looks good now.

I went through the `clear()`/`reset()` change in detail, and the core approach is correct: reusing `remove(key, HierarchyMode.CURRENT_LEVEL)` over a snapshot of the cache's keys closes every cached context exactly once, in the correct bottom-up order, regardless of iteration order.

I've requested a minor change and improvements to the tests, after which I think this PR will be good to go.

sbrannen:
Hi @Will-thom,

Do you have time to address https://github.com/spring-projects/spring-framework/pull/36825#discussion_r3791824881 and https://github.com/spring-projects/spring-framework/pull/36825#discussion_r3791827524?

Please let us know in either case, so that we know how to proceed with this PR.

Thanks,

Sam

sbrannen:
Hi @Will-thom,

Any update on this?

sbrannen (approved):
I'll take care of the open issues after merging this PR.

sbrannen:
This has been merged into `main` in 97067f9c9cfbcbdf0e7dfd07295a061f53ccb6dd and revised in 4b5afaff5341c110fa5cfe4f853f7ff71f15e30e.

Thanks

Issue #35872, opened 2025-11-22
The content-type for SSE changed when the version was upgraded from 6.2.11 to 7.0.1


heowc:
I'm using `spring-webflux` and was returning responses like this:

```java
@GetMapping("test")
public Flux<ServerSentEvent<String>> test() {
    return Flux.interval(Duration.ofSeconds(1))
        .take(Duration.ofSeconds(3))
        .map(i -> ServerSentEvent.builder("Hello " + i).build());
}
```

The content-type changed when upgrading to 7.0.1.

AS-IS: `Content-Type: text/event-stream;charset=UTF-8`
TO-BE: `Content-Type: application/json`

So, I need to change it to the following to get the same behavior as before. Is this intended?

```java
@GetMapping(value = "test", produces = "text/event-stream")
public Flux<ServerSentEvent<String>> test() { 
    return Flux.interval(Duration.ofSeconds(1)) 
        .take(Duration.ofSeconds(3)) 
        .map(i -> ServerSentEvent.builder("Hello " + i).build());
}
```

sdeleuze:
This is a regression caused by the removal of `canSerialize` in Jackson 3, so Spring does not perform this check which was previously returning `false` for `ServerSentEvent`. We will fix it.

sdeleuze:
Please check with `7.0.2-SNAPSHOT` that it now works as expected.

PR #36828, opened 2026-05-22
Restore Jackson XML alongside JAXB2 in WebFlux codecs and exclude String from supported types

### LINKED ISSUE #36775: JacksonXmlEncoder.canEncode incorrectly returns true for String body with application/xml
When using Spring Boot 4's WebClient to send an XML request with a String body and Content-Type: application/xml, the request fails with:

```
Stream decoding is currently not supported
```

# Minimal Reproducer:

``` java
webClient
.post()
.uri(url)
.contentType(MediaType.APPLICATION_XML)
.bodyValue("<xml><a>test</a></xml>")
.retrieve();
```

# Root Cause Analysis

The failure is caused by `JacksonXmlEncoder#canEncode`, which returns true for MediaType.APPLICATION_XML + String.

However, Jackson’s XML encoder does not​ actually support encoding raw String values.

This behavior differs from `JacksonJsonEncoder`, which explicitly prevents this case via:

``` java
@Override
public boolean canEncode(ResolvableType elementType, @Nullable MimeType mimeType) {
    return super.canEncode(elementType, mimeType) && !String.class.isAssignableFrom(elementType.toClass());
}
```

JacksonXmlEncoder does not​ override canEncode, so it incorrectly signals support for String bodies.

# Expected Behavior
JacksonXmlEncoder.canEncode should return false for String types, similar to JacksonJsonEncoder. And falls back to a non-streaming encoder for String bodies with application/xml.

### LINKED ISSUE #36776: Jaxb2XmlEncoder exclusivity prevents JacksonXmlEncoder usage and hinders POJO serialization
In BaseDefaultCodecs (Spring framework 7.0), the registration logic for XML encoders is mutually exclusive: 

if Jaxb2XmlEncoder is detected, the framework automatically registers it and skips​ the registration of JacksonXmlEncoder.

https://github.com/spring-projects/spring-framework/blob/9a54f75d6de31e16219a389a6c1437318dc62ca1/spring-web/src/main/java/org/springframework/http/codec/support/BaseDefaultCodecs.java#L820-L825

Typically, having Jaxb2XmlEncoder available wouldn't be an issue. However, the JAXB API (jakarta.xml.bind:jakarta.xml.bind-api) is often transitively included by widely-used dependencies like spring-data-redis(via spring-oxm).

Because Jaxb2XmlEncoder is prioritized and exclusively registered whenever the JAXB API is present, projects with spring-data-redis(which is almost every Spring Boot project) are forced to use Jaxb2XmlEncoderfor XML serialization.

Noah3521:
## Summary

This PR updates WebFlux default XML codec registration so Jackson XML codecs are available when present even if JAXB is also on the classpath.

The ordering intentionally keeps JAXB XML codecs before Jackson XML codecs to preserve existing WebFlux behavior for JAXB-eligible types while restoring Jackson XML availability for supported cases that JAXB does not cover.

It also applies XML-level Jackson String checks for both encoder and decoder, matching the JSON codec-level checks moved in gh-35872, so raw String values remain on String codec paths.

Closes gh-36776
Closes gh-36775

## Verification

- `git diff --check`
- `./gradlew :spring-web:test --tests 'org.springframework.http.codec.support.CodecConfigurerTests' --tests 'org.springframework.http.codec.support.ClientCodecConfigurerTests' --tests 'org.springframework.http.codec.support.ServerCodecConfigurerTests' --tests 'org.springframework.http.codec.xml.JacksonXmlEncoderTests' --tests 'org.springframework.http.codec.xml.JacksonXmlDecoderTests' --configuration-cache`
- `./gradlew :spring-web:check --no-configuration-cache`

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/support/BaseDefaultCodecs.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/support/BaseDefaultCodecs.java
@@ -679,7 +679,7 @@ class BaseDefaultCodecs implements CodecConfigurer.DefaultCodecs, CodecConfigure
 			addCodec(this.objectReaders, new DecoderHttpMessageReader<>(this.jaxb2Decoder != null ?
 					(Jaxb2XmlDecoder) this.jaxb2Decoder : new Jaxb2XmlDecoder()));
 		}
-		else if(JACKSON_XML_PRESENT) {
+		if (JACKSON_XML_PRESENT) {
 			addCodec(this.objectReaders, new DecoderHttpMessageReader<>(getJacksonXmlDecoder()));
 		}
 		if (KOTLIN_SERIALIZATION_PROTOBUF_PRESENT) {
@@ -821,7 +821,7 @@ class BaseDefaultCodecs implements CodecConfigurer.DefaultCodecs, CodecConfigure
 			addCodec(writers, new EncoderHttpMessageWriter<>(this.jaxb2Encoder != null ?
 					(Jaxb2XmlEncoder) this.jaxb2Encoder : new Jaxb2XmlEncoder()));
 		}
-		else if (JACKSON_XML_PRESENT) {
+		if (JACKSON_XML_PRESENT) {
 			addCodec(writers, new EncoderHttpMessageWriter<>(getJacksonXmlEncoder()));
 		}
 		if (KOTLIN_SERIALIZATION_PROTOBUF_PRESENT) {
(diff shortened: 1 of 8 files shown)


rstoyanchev (changes requested):
We've discussed this and we'll go with Jackson XML first and JAXB2 second in an if-else condition matching to HttpMessageConverters and to the effective behavior for a very long time, all the way up until 7.0.

Jackson XML supports Jackson annotations and optionally JAXB annotations. If Jackson annotations are used that's clearly for Jackson XML. If JAXB annotations are used and both libraries are present, then Jackson XML should be preferred because it's presence is more likely intentional whereas JAXB2 can be present as a transitive dependency as is the case in the report #36776 where it's brought in by spring-oxm.

Please, let me know if you plan to make this update, or otherwise I can also take it from here.

rstoyanchev:
As mentioned above, we have chosen to resolve this differently. In addition, keep in mind we prefer to fix issues in independent commits.

Issue #36820, opened 2026-05-20
@Conditional inheritance on nested @Configuration is sensitive to discovery path


Dennis-Mircea:
## Use case

A common pattern when organizing related `@Configuration` classes is to group them as `static` nested classes inside a containing `@Configuration`, with a `@Conditional` on the outer that is intended to gate the whole group:

```java
@Configuration
@Conditional(MyFeatureCondition.class)
class MyFeatureConfig {

    @Configuration
    static class DataAccess {
        @Bean DataSource dataSource() { ... }
    }

    @Configuration
    static class Web {
        @Bean WebFilter myFilter() { ... }
    }
}
```

The developer's expectation: if `MyFeatureCondition` does not match, none of the nested configurations register their beans.

In practice this works when `MyFeatureConfig` is registered directly or imported via `@Import`, because the parser's recursion into member classes sets up an `importedBy` reference on each inner class, and the existing inheritance mechanism in `ConfigurationClassBeanDefinitionReader.TrackedConditionEvaluator.shouldSkip` (`spring-context/.../ConfigurationClassBeanDefinitionReader.java:509-531`) walks that chain at REGISTER_BEAN phase.

It silently breaks down when `MyFeatureConfig` happens to live in a package reached by `@ComponentScan`. The scan picks up both `MyFeatureConfig` and its nested classes as independent bean-definition candidates. The outer is correctly skipped by its condition, but each nested `@Configuration` is parsed using only its own metadata: it has no `importedBy` reference, so no inheritance applies, and its beans register regardless of whether `MyFeatureCondition` matched. The same lexical arrangement therefore behaves differently depending purely on how the inner happens to be discovered, with no diagnostic and no documentation that flags discovery path as relevant.

This surfaces in production code as "my bean appeared even though the feature was disabled" or, in tests, as fixtures that work when registered directly but stop working when scanned. Spring Boot's auto-configuration codebase is full of nested `@Configuration` classes for exactly this kind of grouping, which is why every Spring Boot auto-config defensively redeclares conditions on each nested class.

## What I have tried

1. **Redeclaring the conditions on every nested `@Configuration`.** This is the current idiomatic workaround and matches the pattern used everywhere in Spring Boot's `module/*`. It works, but it scales poorly when there are many nested classes, and the duplication is easy to drift on (someone adds a new nested class and forgets the condition; the bug becomes silent again).
[...]

PR #36831, opened 2026-05-23
Document @Conditional gating of nested @Configuration classes

Dennis-Mircea:
## Summary

The "With nested `@Configuration` classes" section of the `@Configuration` Javadoc previously mentioned `@Profile` working with nested classes but was silent on how `@Conditional` interacts with them. The corresponding reference-doc section ("Conditionally Include `@Configuration` Classes or `@Bean` Methods" in `composing-configuration-classes.adoc`) was similarly silent.

In practice, whether an enclosing class's `@Conditional` gates a nested `@Configuration` depends on how the nested class is discovered. Nested classes reached through parser recursion from their enclosing class or via `@Import` inherit the enclosing's conditions through the existing `importedBy` mechanism (`ConfigurationClassBeanDefinitionReader.TrackedConditionEvaluator.shouldSkip`). Nested classes discovered independently (via `@ComponentScan` or by directly registering them) are processed using only their own conditions. The rule is real, long-standing, and currently undocumented.

This PR adds a paragraph to the `@Configuration` Javadoc and a `[NOTE]` admonition to the reference-doc page documenting that rule and recommending the redeclare-or-meta-annotation workaround when the same gating is intended on independent discovery paths.

See also #36820 for a separate proposal to close the discovery-path asymmetry at the framework level. This PR is independent of that proposal; it documents current behavior.

## Test plan

- `./gradlew :spring-context:check` (includes javadoc with `-Werror`, checkstyle, runtime-hints, tests), all green
- `./gradlew :framework-docs:antora`, all green
- Locally rendered Antora site verified, admonition renders correctly under "Conditionally Include `@Configuration` Classes or `@Bean` Methods"
- Locally rendered Javadoc for `@Configuration` verified, new paragraph reads cleanly and `{@link}` references resolve

### DIFF
--- a/framework-docs/modules/ROOT/pages/core/beans/java/composing-configuration-classes.adoc
+++ b/framework-docs/modules/ROOT/pages/core/beans/java/composing-configuration-classes.adoc
@@ -620,7 +620,7 @@ independently of its enclosing class, for example via `@ComponentScan` or
 by directly registering it against the application context, it is processed
 using only its own `@Conditional` annotations. In that case, redeclare the
 relevant conditions on the nested class, or extract them into a composed
-meta-annotation applied to both, if the same gating is intended.
+annotation applied to both, if the same gating is intended.
 ====
 
 
--- a/framework-docs/modules/ROOT/pages/core/beans/java/composing-configuration-classes.adoc
+++ b/framework-docs/modules/ROOT/pages/core/beans/java/composing-configuration-classes.adoc
@@ -610,6 +610,19 @@ Kotlin::
 See the {spring-framework-api}/context/annotation/Conditional.html[`@Conditional`]
 javadoc for more detail.
 
+[NOTE]
+====
+A `@Conditional` declared on an enclosing `@Configuration` class gates the
+registration of nested `@Configuration` classes within it only when the
+nested class is reached through the parser's recursion from its enclosing
+class, or through `@Import`. When the nested class is discovered
+independently of its enclosing class, for example via `@ComponentScan` or
+by directly registering it against the application context, it is processed
+using only its own `@Conditional` annotations. In that case, redeclare the
+relevant conditions on the nested class, or extract them into a composed
+meta-annotation applied to both, if the same gating is intended.
+====
+
 
 [[beans-java-combining]]
 == Combining Java and XML Configuration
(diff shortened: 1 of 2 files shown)


framework-docs/modules/ROOT/pages/core/beans/java/composing-configuration-classes.adoc
```diff
@@ -610,6 +610,19 @@ Kotlin::
...
+====
+A `@Conditional` declared on an enclosing `@Configuration` class gates the
+registration of nested `@Configuration` classes within it only when the
+nested class is reached through the parser's recursion from its enclosing
+class, or through `@Import`. When the nested class is discovered
+independently of its enclosing class, for example via `@ComponentScan` or
+by directly registering it against the application context, it is processed
+using only its own `@Conditional` annotations. In that case, redeclare the
+relevant conditions on the nested class, or extract them into a composed
+meta-annotation applied to both, if the same gating is intended.
```
sbrannen: Instead of "composed meta-annotation", we refer to that as a "composed annotation" (or "custom composed annotation") since that composed annotation may or may not be used as a "meta-annotation".

See our documented [Terminology](https://github.com/spring-projects/spring-framework/wiki/Spring-Annotation-Programming-Model#terminology) for details.

spring-context/src/main/java/org/springframework/context/annotation/Configuration.java
```diff
@@ -341,6 +341,18 @@
...
+ * {@code @Configuration} class gate registration of any nested
+ * {@code @Configuration} classes within it when the nested class is reached
+ * through the parser's recursion from its enclosing class (the case shown
+ * above) or via {@link Import @Import}. When the nested class is discovered
+ * independently of its enclosing class, for example via
+ * {@link ComponentScan @ComponentScan} or by directly registering the nested
+ * class against the application context, it is processed using only its own
+ * {@code @Conditional} annotations. In that case, redeclare the relevant
+ * conditions on the nested class, or extract them into a composed
+ * meta-annotation applied to both, if the same gating is intended.
```
sbrannen: The same applies here regarding "composed annotations".

sbrannen (changes requested):
Thanks for the PR. I've requested a few changes.

framework-docs/modules/ROOT/pages/core/beans/java/composing-configuration-classes.adoc
Dennis-Mircea: Thank you for the review, @sbrannen! I fixed it now.

spring-context/src/main/java/org/springframework/context/annotation/Configuration.java
Dennis-Mircea: Fixed!

sbrannen (approved):

sbrannen:
This has been merged into `7.0.x` and `main` in f3bfe27445d3f5aa284ba211ca81e4bbf3d49b29 and revised in c17939ed5c177ecf4f848a5031b2e689ac908658.

Thanks

PR #36853, opened 2026-05-29
Document LocaleContextHolder inheritable behavior

### LINKED ISSUE #36687: Doc: LocaleContextHolder with inheritable only propagates to newly created threads
I noticed that the Locales when trying to use them inside threads are not consistent when using `LocaleContextHolder.setLocale(Locale.GERMAN, true);`.
It seems like the thread pools use the language it once was sent and uses it for all other requests.
You can reporduce it when using a threadpool of 1.

This is the Test I have:

```
@SpringBootTest
class AsyncContextPropagationTest {

    private ThreadPoolTaskExecutor delegate;
    private TaskExecutor taskExecutor;

    @BeforeEach
    void setUpExecutor() {
        delegate = new ThreadPoolTaskExecutor();
        delegate.setCorePoolSize(1);
        delegate.setMaxPoolSize(100);
        delegate.setQueueCapacity(50);
        delegate.setThreadNamePrefix("test-async-");
        delegate.initialize();

        taskExecutor = new DelegatingSecurityContextAsyncTaskExecutor(delegate);
    }

    @AfterEach
    void tearDownExecutor() {
        if (delegate != null) {
            delegate.shutdown();
        }
    }

    @Test
    void localeIsVisibleInAsyncTask() {
        LocaleContextHolder.setLocale(Locale.GERMAN, true);
        CompletableFuture<Locale> futureGerman = CompletableFuture.supplyAsync(
            LocaleContextHolder::getLocale,
            taskExecutor
        );
        var futureGermanResult = futureGerman.join();

        LocaleContextHolder.setLocale(Locale.ENGLISH, true);
        CompletableFuture<Locale> futureEnglish = CompletableFuture.supplyAsync(
            LocaleContextHolder::getLocale,
            taskExecutor
        );

        var futureEnglishResult = futureEnglish.join();

        assertThat(futureGermanResult).isEqualTo(Locale.GERMAN);
        assertThat(futureEnglishResult).isEqualTo(Locale.ENGLISH);
    }
}

```

Right now the workaround is to use an explicit TaskDecorator for that case to copy the locale to all child threads.

```
public class ContextCopyingDecorator implements TaskDecorator {

    @NonNull
    @Override
    public Runnable decorate(@NonNull Runnable runnable) {
        var locale = LocaleContextHolder.getLocale();

        return () -> { // code runs inside executor thread and binds context
            try {
                if (locale != null) {
                    LocaleContextHolder.setLocale(locale);
                }
                runnable.run();
            } finally {
                // Nothing to do.
            }
        };
    }
}
```

Then set it via `delegate.setTaskDecorator(new ContextCopyingDecorator());` and it works just fine.

Not sure if the `inheritable` on setLocale() is intended to work like that or its simply a bug.
At least I don't know why it would work like that.

cfcromn:
Closes #36687

This change documents that `LocaleContextHolder`'s `inheritable` flag relies on
JDK `InheritableThreadLocal` semantics, which only propagate values to newly
created child threads at the time of their creation.

In thread pool environments where threads are reused across tasks, the context
is not refreshed per task submission, so this mechanism is not suitable for
per-task context propagation to reused threads.

The documentation now points users toward `TaskDecorator` for explicit per-task
locale context propagation.

Verification:
- `./gradlew :spring-context:compileJava`
- `./gradlew :spring-context:test --tests "org.springframework.context.i18n.LocaleContextHolderTests"`

### DIFF
--- a/spring-context/src/main/java/org/springframework/context/i18n/LocaleContextHolder.java
+++ b/spring-context/src/main/java/org/springframework/context/i18n/LocaleContextHolder.java
@@ -30,6 +30,17 @@ import org.springframework.core.NamedThreadLocal;
  * by any child threads spawned by the current thread if the
  * {@code inheritable} flag is set to {@code true}.
  *
+ * <p><b>Note:</b> The {@code inheritable} flag relies on JDK
+ * {@link java.lang.InheritableThreadLocal} semantics, which only propagates
+ * the context value to <em>newly created</em> child threads at the time of
+ * their creation. In thread pool environments where threads are reused across
+ * tasks, the context is not refreshed per task submission; reused threads
+ * retain whatever value was inherited when they were originally created.
+ * This mechanism is therefore not suitable for per-task context propagation
+ * in thread pool environments. Instead, consider using a
+ * {@link org.springframework.core.task.TaskDecorator} that captures and
+ * restores the locale context around each task execution.
+ *
  * <p>Used as a central holder for the current Locale in Spring,
  * wherever necessary: for example, in MessageSourceAccessor.
  * DispatcherServlet automatically exposes its current Locale here.
@@ -42,6 +53,7 @@ import org.springframework.core.NamedThreadLocal;
  * @see LocaleContext
  * @see org.springframework.context.support.MessageSourceAccessor
  * @see org.springframework.web.servlet.DispatcherServlet
+ * @see org.springframework.core.task.TaskDecorator
  */
 public final class LocaleContextHolder {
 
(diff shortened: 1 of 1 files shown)


cfcromn:
Hi team, just checking whether this PR is aligned with the intended direction for #36687.

The change is intentionally limited to documenting the existing `InheritableThreadLocal` behavior and pointing users to `TaskDecorator` for per-task propagation in reused thread-pool scenarios.

Happy to adjust the wording or location if there is a preferred place for this documentation.

PR #36856, opened 2026-05-30
Log broker availability events as String messages

kodacme:
## Overview

`AbstractBrokerMessageHandler` logs `BrokerAvailabilityEvent` directly at INFO level when the broker becomes (un)available:

```java
logger.info(this.availableEvent);
```

With a structured JSON logging layout (e.g. Logstash/ECS encoders backed by Jackson), the logging framework may serialize the *event object* rather than its `toString()`. It then traverses `BrokerAvailabilityEvent.getSource()`, which is the `SimpleBrokerMessageHandler` itself.

That object graph contains a cyclic reference:

```
SimpleBrokerMessageHandler["clientInboundChannel"]
  -> ExecutorSubscribableChannel["executor"]
    -> ThreadPoolTaskExecutor["threadPoolExecutor"]
      -> ThreadPoolTaskExecutor$1["threadFactory"]   // anonymous inner class, holds outer ref
        -> ThreadPoolTaskExecutor["threadPoolExecutor"]
          -> ...
```

Because the thread factory is a non-static inner class, it keeps an implicit reference back to the enclosing `ThreadPoolTaskExecutor`, producing an effectively infinite graph. Jackson aborts at its maximum nesting depth:

```
com.fasterxml.jackson.databind.JsonMappingException:
Document nesting depth (1001) exceeds the maximum allowed (1000,
from `StreamWriteConstraints.getMaxNestingDepth()`)
```

The application keeps starting, but startup logs are flooded with a very large stack trace, which can hide real problems.

## Change

Log the event's `toString()` representation instead of the event object, in both `publishBrokerAvailableEvent()` and `publishBrokerUnavailableEvent()`:

```java
logger.info(this.availableEvent.toString());
```

`BrokerAvailabilityEvent.toString()` already produces a concise, safe message, so the operational signal is preserved while structured logging layouts no longer traverse framework internals.

## Notes

- Behavior with plain text layouts is unchanged (they already called `toString()`).
- Added a test verifying both lifecycle methods log the expected String messages.

### DIFF
--- a/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/AbstractBrokerMessageHandler.java
+++ b/spring-messaging/src/main/java/org/springframework/messaging/simp/broker/AbstractBrokerMessageHandler.java
@@ -334,7 +334,7 @@ public abstract class AbstractBrokerMessageHandler
 		boolean shouldPublish = this.brokerAvailable.compareAndSet(false, true);
 		if (this.eventPublisher != null && shouldPublish) {
 			if (logger.isInfoEnabled()) {
-				logger.info(this.availableEvent);
+				logger.info(this.availableEvent.toString());
 			}
 			this.eventPublisher.publishEvent(this.availableEvent);
 		}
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks @kodacme and sorry for the delayed feedback. I really appreciate the efforts you made to describe the use case, this is really helpful.

PR #36862, opened 2026-06-02
Apply auto-grow collection limit to direct field binding

### LINKED ISSUE #36861: DataBinder direct field access ignores auto-grow collection limit
`DataBinder` applies its `autoGrowCollectionLimit` to bean-property binding, but direct-field binding does not apply the same limit to the underlying `DirectFieldAccessor`.

With the default limit of `256`, binding `items[255].name` succeeds for both bean-property binding and direct-field binding, as expected.

Binding `items[256].name` is rejected for bean-property binding, but succeeds after calling `initDirectFieldAccess()`.

**Reproducer:**
https://github.com/mkurz/spring-direct-field-autogrow-repro
- This test should succeed but fails: https://github.com/mkurz/spring-direct-field-autogrow-repro/blob/f1731a500208493f615c30db97a14597b9ce1d09/src/test/java/example/DirectFieldAutoGrowLimitTests.java#L55-L63

**Expected behavior:**
Direct-field binding should reject `items[256].name` the same way bean-property binding does.

**Actual behavior:**
Direct-field binding grows the collection past the configured limit.

mkurz:
`DataBinder` applies its auto-grow collection limit to bean property access, but direct field access left `DirectFieldAccessor` at its default limit.

Pass DataBinder's configured limit into `DirectFieldBindingResult` and apply it to the `DirectFieldAccessor`.

Closes gh-36861

### DIFF
--- a/spring-context/src/main/java/org/springframework/validation/DataBinder.java
+++ b/spring-context/src/main/java/org/springframework/validation/DataBinder.java
@@ -341,7 +341,7 @@ public class DataBinder implements PropertyEditorRegistry, TypeConverter {
 	 */
 	protected AbstractPropertyBindingResult createDirectFieldBindingResult() {
 		DirectFieldBindingResult result = new DirectFieldBindingResult(getTarget(),
-				getObjectName(), isAutoGrowNestedPaths());
+				getObjectName(), isAutoGrowNestedPaths(), getAutoGrowCollectionLimit());
 
 		if (this.conversionService != null) {
 			result.initConversion(this.conversionService);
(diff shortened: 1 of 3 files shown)


jhoeller:
Good catch, thanks for the PR! I've refined the implementation to expose the autoGrowCollectionLimit flag in the ConfigurablePropertyAccessor interface, along the lines of autoGrowNestedPaths. Also, we'll fix the 6.1-era javadoc note in the older branches so that the javadoc there correctly reflects the original applicability: #36863.

Issue #34481, opened 2025-02-25
ExceptionHandlerExceptionResolver results in 200 for some special exceptions


NiFNi:
I stumbled upon https://github.com/spring-projects/spring-framework/issues/34264 which caused some bigger issues internally because some server side connection resets resulted in a 200 successful response. While the biggest issue was solved I think that it is quite unsettling that an exception results in a 200 response and this is still the case for the current spring version (6.2.3) shipped with spring boot 3.4.3.

Minimal example which returns a 200 even though an exception is thrown and not properly handled:

```
@RestController
@ControllerAdvice
public class DemoController {

    private static final Logger log = LoggerFactory.getLogger(DemoController.class);

    @GetMapping
    public String demo() {
        throw new RuntimeException("connection reset by peer");
    }

    @ExceptionHandler(RuntimeException.class)
    public ResponseEntity<Object> handleRuntimeException(RuntimeException e) {
        log.error(e.getMessage());
        throw e;
    }
}
```

While this usecase might not look to useful we definitely have use cases where we want to handle an exception in some custom way using an exception handler but only in specific cases. So something like this:

```
    @ExceptionHandler(RuntimeException.class)
    public ResponseEntity<Object> handleRuntimeException(RuntimeException e) {
        if (matchesSomeCriteria(e)) {
            return specialHandling(e);
        }
        throw e;
    }
```

While investigating I found that this is probably caused here: 
https://github.com/spring-projects/spring-framework/blob/75329e6d9d93180d9edceb81838b0d96a145d665/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ExceptionHandlerExceptionResolver.java#L463-L465

If I understand this correctly the idea is to not continue handling exceptions in case the client closed the connection. Since the check for this is just using some string comparison on the exception message it is quite easy to have an exception matching this as can be seen above.

If no one has ideas on how to fix this properly I would suggest to at least set a 500 error status on the ModelView that is created in the ExceptionHandlerExceptionResolver to make this a little less worrying. While the behaviour might still be unexpected it is at least not returning successful responses for cases where disconnectedClientHelper.checkAndLogClientDisconnectedException returns true even though some internal exception occurred.

If you agree with my findings I can create a PR for this.

rstoyanchev:
For "connection reset by peer" or "broken pipe" I would expect the exception type is an instance of `IOException`, and we could add that check to make it more specific. 

[...]

PR #36868, opened 2026-06-03
Set 500 status for disconnected client exceptions

king-407:
## Summary

This PR updates Spring MVC disconnected-client exception handling to set the response status to 500 instead of leaving it at the default 200 status.

The change covers both paths discussed in gh-34481:

- `DefaultHandlerExceptionResolver`, which handles default Spring MVC exception resolution
- `ExceptionHandlerExceptionResolver`, where an `@ExceptionHandler` rethrows a disconnected-client style exception

For an actual disconnected client, the client will not observe the response. For cases where the exception is incorrectly classified as a disconnected-client exception, this avoids silently reporting a failed request as successful.

This is based on the latest issue discussion; happy to adjust the scope or implementation if the maintainers prefer a different approach.

## Testing

- `./gradlew :spring-webmvc:test --tests org.springframework.web.servlet.mvc.support.DefaultHandlerExceptionResolverTests --tests org.springframework.web.servlet.mvc.method.annotation.ExceptionHandlerExceptionResolverTests`
- `git diff --check`

See gh-34481

### DIFF
--- a/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ExceptionHandlerExceptionResolver.java
+++ b/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ExceptionHandlerExceptionResolver.java
@@ -465,6 +465,9 @@ public class ExceptionHandlerExceptionResolver extends AbstractHandlerMethodExce
 		}
 		catch (Throwable invocationEx) {
 			if (disconnectedClientHelper.checkAndLogClientDisconnectedException(invocationEx)) {
+				if (!response.isCommitted()) {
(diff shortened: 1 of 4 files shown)


king-407:
This PR is based on the latest discussion around setting a 500 response status for disconnected-client exceptions instead of leaving the response at the default 200.

The change covers both Spring MVC paths discussed in the issue: `DefaultHandlerExceptionResolver` and `ExceptionHandlerExceptionResolver`. I also added focused tests for `IOException("Broken pipe")`, including the case where an `@ExceptionHandler` rethrows the exception.

Happy to adjust the scope or implementation if the maintainers prefer a different approach.

rstoyanchev:
This has been fixed already.

PR #36871, opened 2026-06-04
Include zone ID in CronTrigger's equals/hashCode implementations

zmovo:
By default, `CronTrigger#equals` and `#hashCode` consider only the wrapped
`CronExpression`. Since 5.3, `CronTrigger` also carries an optional `ZoneId`
that is used in `nextExecution()` when a time zone is configured explicitly.
Two triggers built with the same cron pattern but different zones can therefore
produce different next execution times while still comparing as equal and
sharing the same hash code.

This causes incorrect deduplication when triggers are stored in a `Set` or used
as `Map` keys, and breaks the general `Object` contract for a type whose runtime
behavior depends on both the expression and the zone.

This PR includes `zoneId` in `equals` and `hashCode`, using
`ObjectUtils.nullSafeEquals` and `ObjectUtils.nullSafeHash` so that triggers
with a `null` zone (the default constructor) remain equal to each other but
distinct from triggers with an explicit zone.

This does not change trigger scheduling behavior or how `nextExecution`
selects a zone at runtime; it only corrects value-based equality for triggers
that differ solely by time zone configuration.

It also adds a regression test in
`CronTriggerTests#equalsAndHashCodeConsidersZoneId`.

## Test plan
- [x] `./gradlew :spring-context:test --tests "org.springframework.scheduling.support.CronTriggerTests.equalsAndHashCodeConsidersZoneId"`

### DIFF
--- a/spring-context/src/main/java/org/springframework/scheduling/support/CronTrigger.java
+++ b/spring-context/src/main/java/org/springframework/scheduling/support/CronTrigger.java
@@ -19,6 +19,7 @@ package org.springframework.scheduling.support;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.util.Objects;
 import java.util.TimeZone;
 
 import org.jspecify.annotations.Nullable;
@@ -26,7 +27,6 @@ import org.jspecify.annotations.Nullable;
 import org.springframework.scheduling.Trigger;
 import org.springframework.scheduling.TriggerContext;
 import org.springframework.util.Assert;
-import org.springframework.util.ObjectUtils;
 
 /**
  * {@link Trigger} implementation for cron expressions. Wraps a
@@ -147,12 +147,12 @@ public class CronTrigger implements Trigger {
 	public boolean equals(@Nullable Object other) {
 		return (this == other || (other instanceof CronTrigger that &&
 				this.expression.equals(that.expression) &&
-				ObjectUtils.nullSafeEquals(this.zoneId, that.zoneId)));
+				Objects.equals(this.zoneId, that.zoneId)));
 	}
 
 	@Override
 	public int hashCode() {
-		return ObjectUtils.nullSafeHash(this.expression, this.zoneId);
+		return Objects.hash(this.expression, this.zoneId);
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


spring-context/src/main/java/org/springframework/scheduling/support/CronTrigger.java
```diff
@@ -145,12 +146,13 @@ Instant determineInitialTimestamp(TriggerContext triggerContext) {
 	@Override
 	public boolean equals(@Nullable Object other) {
 		return (this == other || (other instanceof CronTrigger that &&
-				this.expression.equals(that.expression)));
+				this.expression.equals(that.expression) &&
+				ObjectUtils.nullSafeEquals(this.zoneId, that.zoneId)));
```
sbrannen:
```suggestion
				Objects.equals(this.zoneId, that.zoneId)));
```

I think we can just use the `java.lang.Objects` API here.

spring-context/src/main/java/org/springframework/scheduling/support/CronTrigger.java
```diff
@@ -145,12 +146,13 @@ Instant determineInitialTimestamp(TriggerContext triggerContext) {
...
 	}
 
 	@Override
 	public int hashCode() {
-		return this.expression.hashCode();
+		return ObjectUtils.nullSafeHash(this.expression, this.zoneId);
```
sbrannen:
```suggestion
		return Objects.hash(this.expression, this.zoneId);
```

Same here.

spring-context/src/test/java/org/springframework/scheduling/support/CronTriggerTests.java
```diff
@@ -746,6 +747,16 @@ void daylightSavingMissingHour() {
 		assertThat(nextExecutionTime).isEqualTo(this.calendar.getTime());
 	}
 
+	@Test
+	void equalsAndHashCodeConsidersZoneId() {
```
sbrannen:
```suggestion
	void equalsAndHashCodeConsiderZoneId() {
```

sbrannen (changes requested):
Thanks for the PR.

zmovo:
> Thanks for the PR.
> 
> I've requested a few minor changes.

Thanks for the feedback! I've updated the PR accordingly. Could you please take another look?

sbrannen:

Thanks

Issue #29520, opened 2022-11-18
ConcurrentLruCache implementation is using too much heap memory


bclozel:
As seen in FasterXML/jackson-databind#3665, the approach taken in our `ConcurrentLruCache` implementation can result in an increase of memory heap consumption because of how the read operations queue is structured.

We've experimented with an alternate solution that "flattens" that queue, trading arrays of `AtomicReference` for `AtomicReferenceArray`. This results in a slight performance decrease but looks acceptable for our use case. We can also consider decreasing the default size of queues as well. They're currently calculated with "number of CPUs x fixed size" - the use cases present in Spring Framework probably don't need this much memory by default.

PR #36872, opened 2026-06-04
Simplify BUFFER_COUNT in ConcurrentLruCache to a constant

### LINKED ISSUE #36780: Suspicious BUFFER_COUNT value in ConcurrentLruCache
The `BUFFER_COUNT` computed in `ConcurrentLruCache.ReadOperations` seems suspicious, since it's basically always just 4; it seems like the intention was to size things based on number of CPUs, but using `Math.min` here with 4 means that we always just end up with the value 4 (unless it's a uniprocessor machine). If it was intentional then the logic in `detectNumberOfBuffers` can be simplified, and there's no reason to try to find the next largest power of two. It feels like maybe the intent was to clamp the value to be a min of 4, but to allow a higher max size?

I have yet to benchmark it yet, but it should be easy enough to test out changing the logic to allow for actual buffer count values > 4.

See: https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/util/ConcurrentLruCache.java#L453

s-chan-o:
## Problem

The `detectNumberOfBuffers()` method in `ConcurrentLruCache.ReadOperations` attempts to scale the read buffer count based on the number of available processors:

```java
private static int detectNumberOfBuffers() {
    int availableProcessors = Runtime.getRuntime().availableProcessors();
    int nextPowerOfTwo = 1 << (Integer.SIZE - Integer.numberOfLeadingZeros(availableProcessors - 1));
    return Math.min(4, nextPowerOfTwo);
}
```

However, the use of `Math.min(4, nextPowerOfTwo)` caps the result at 4 for any system with 3 or more processors. For systems with 5+ processors, `nextPowerOfTwo` would be 8 or more, but the `Math.min` always returns 4. The CPU-detection logic adds complexity without any actual effect for the vast majority of deployment environments.

As noted in [gh-36780](https://github.com/spring-projects/spring-framework/issues/36780), the value is *"basically always just 4"*.

## Fix

Simplify `BUFFER_COUNT` to a plain constant `4`, removing the `detectNumberOfBuffers()` method entirely. This is consistent with the feedback from [gh-29520](https://github.com/spring-projects/spring-framework/issues/29520) that informed the original cap at 4.

Closes gh-36780

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ConcurrentLruCache.java
+++ b/spring-core/src/main/java/org/springframework/util/ConcurrentLruCache.java
@@ -367,7 +367,7 @@ public final class ConcurrentLruCache<K, V> {
 
 	private static final class ReadOperations<K, V> {
 
-		private static final int BUFFER_COUNT = detectNumberOfBuffers();
+		private static final int BUFFER_COUNT = 4;
 
 		private static final int BUFFERS_MASK = BUFFER_COUNT - 1;
 
@@ -450,11 +450,6 @@ public final class ConcurrentLruCache<K, V> {
 			this.processedCount.lazySet(bufferIndex, writeCount);
 		}
 
-		private static int detectNumberOfBuffers() {
-			int availableProcessors = Runtime.getRuntime().availableProcessors();
-			int nextPowerOfTwo = 1 << (Integer.SIZE - Integer.numberOfLeadingZeros(availableProcessors - 1));
-			return Math.min(4, nextPowerOfTwo);
-		}
 	}
 
 


bclozel:
@s-chan-o [please sign off your commit](https://github.com/spring-projects/spring-framework/pull/36872/checks?check_run_id=79441480233), we cannot accept this contribution without it. Thanks

s-chan-o:
@bclozel Added Signed-off-by in 592ab02. Thanks!

bclozel:
Thanks for your contribution @s-chan-o !

PR #36876, opened 2026-06-05
Restore thread interrupt flag in `DefaultMvcResult` when catching `InterruptedException`

leestana01:
`DefaultMvcResult.awaitAsyncDispatch()` catches `InterruptedException` and returns `false` without restoring the thread's interrupt status, so the interruption is silently lost.

When `InterruptedException` is caught without being rethrown, the interrupt status should be reasserted, as the framework already does elsewhere. For example, `ConcurrencyThrottleSupport` re-interrupts the current thread on the same exception ("Re-interrupt current thread, to allow other threads to react."). This applies the same handling with `Thread.currentThread().interrupt()` before returning `false`.

There is no other behavior change: an interruption already makes `getAsyncResult(...)` throw `IllegalStateException` via the `false` return, so this only preserves the interrupt status as it propagates. The added test interrupts the current thread before calling `getAsyncResult(...)` and verifies the interrupt status is still set afterwards.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/DefaultMvcResult.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/DefaultMvcResult.java
@@ -153,6 +153,7 @@ class DefaultMvcResult implements MvcResult {
 			return this.asyncDispatchLatch.await(timeout, TimeUnit.MILLISECONDS);
 		}
 		catch (InterruptedException ex) {
+			Thread.currentThread().interrupt();
 			return false;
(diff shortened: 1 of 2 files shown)


leestana01:
One note on the test: `CountDownLatch.await(...)` clears the interrupt status when it throws `InterruptedException`, so on the current code the assertion `Thread.currentThread().isInterrupted()` is `false`. The test therefore fails without the one-line change and passes with it.

sbrannen:

Thanks

PR #36894, opened 2026-06-10
GH-36738: Document rendering-scope limitation of WebFlux RequestContext.changeLocale()

### LINKED ISSUE #36738: WebFlux RequestContext.changeLocale() lacks Javadoc for rendering-scope limitation (MVC parity gap)
## Problem

`RequestContext.changeLocale(Locale)` exists in both Spring MVC and Spring WebFlux with identical method names and signatures. In MVC, the method persists the locale change through the configured `LocaleResolver`. In WebFlux, it updates only the private field on the current `RequestContext` instance — the change is scoped to the current rendering cycle and silently discarded on the next request.

The asymmetry itself may be defensible given the reactive model. The MVC equivalent has explicit Javadoc linking `@see LocaleResolver#setLocale` and documenting persistence behavior. The WebFlux equivalent has none. A developer reading both APIs would reasonably infer symmetry that does not exist.

## MVC vs WebFlux contrast

Spring MVC delegates through the configured `LocaleResolver`:
```java
// MVC — RequestContext.changeLocale()
public void changeLocale(Locale locale) {
    LocaleResolver localeResolver =
        RequestContextUtils.getLocaleResolver(this.request);
    if (localeResolver == null) {
        throw new IllegalStateException(
            "Cannot change locale if no LocaleResolver configured");
    }
    localeResolver.setLocale(this.request, this.response, locale);
}
```
WebFlux updates a private field:
```java
// WebFlux — RequestContext.changeLocale()
public void changeLocale(Locale locale) {
    this.locale = locale;
}
```

MVC also fails loudly when no resolver is configured — the IllegalStateException tells the developer they have a wiring problem. WebFlux always succeeds silently.

## Path forward
Add Javadoc to both `changeLocale()` overloads clearly stating that the method affects only the current rendering context, does not delegate through `LocaleContextResolver`, and does not persist beyond the current request. Cross-reference the recommended approach for durable locale switching in WebFlux.

Happy to open a PR about it if this works with you

A follow-on enhancement to introduce reactive delegation through `LocaleContextResolver` will be filed separately, dependent on #36569 landing first.

## Affects
Spring WebFlux `RequestContext` — all currently supported versions.

## See also
#36569 (Add `CookieLocaleContextResolver` to Spring WebFlux) — the Javadoc fix above is independent and immediately actionable; reactive delegation as a follow-on depends on or follows from that work.

won-seoop:
## Summary

Closes #36738.

The Spring MVC `RequestContext.changeLocale()` delegates through the configured `LocaleResolver`, persisting the new locale for subsequent requests. The WebFlux equivalent updates only a private field on the current `RequestContext` instance; the change is silently discarded at the end of the rendering cycle. This asymmetry was undocumented, causing developers to expect the same durable behavior as in MVC.

This PR adds Javadoc to both `changeLocale` overloads in `org.springframework.web.reactive.result.view.RequestContext`, clearly documenting:

- The change affects only the current rendering context (not the underlying request or response).
- It does **not** delegate to a `LocaleContextResolver`.
- It is not persisted beyond the current rendering cycle.
- For durable locale changes in WebFlux, developers should configure a `LocaleContextResolver` and update it from a `WebFilter` or handler method.

Spring MVC's `RequestContext.changeLocale()` already has explicit Javadoc (including `@see LocaleResolver#setLocale`) — this PR closes the MVC/WebFlux documentation parity gap for this method.

## Test plan

- [ ] Existing tests continue to pass (no behaviour change — Javadoc only)

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/result/view/RequestContext.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/result/view/RequestContext.java
@@ -134,6 +134,16 @@ public class RequestContext {
 
 	/**
 	 * Change the current locale to the specified one.
+	 * <p><strong>Note:</strong> Unlike the Spring MVC counterpart, this method only
+	 * updates the locale within the scope of the current {@code RequestContext} instance
+	 * for view rendering purposes. It does <em>not</em> delegate to a
+	 * {@link org.springframework.web.server.i18n.LocaleContextResolver} and the change
+	 * is not persisted beyond the current rendering cycle. For a durable locale change
+	 * across requests in WebFlux, configure a
+	 * {@link org.springframework.web.server.i18n.LocaleContextResolver} and update it
+	 * from a {@link org.springframework.web.server.WebFilter} or handler method instead.
+	 * @param locale the new locale
+	 * @see #changeLocale(java.util.Locale, java.util.TimeZone)
 	 */
 	public void changeLocale(Locale locale) {
 		this.locale = locale;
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks but we already received a PR for this. Also, the DCO is not signed so we cannot accept this contribution anyway.

PR #36895, opened 2026-06-10
GH-36631: Fix MockHttpServletRequest.isRequestedSessionIdValid() to follow Servlet 6.1 spec

won-seoop:
## Summary

`MockHttpServletRequest.isRequestedSessionIdValid()` always returned `true` by default regardless of the actual session state, which violates the Servlet 6.1 specification:

> **`isRequestedSessionIdValid()`** — returns `false` if the client did not specify any session ID; otherwise reflects whether the requested ID corresponds to a valid session in the current context.

This causes two concrete failures:
1. A fresh `MockHttpServletRequest` (no session ID set) has `isRequestedSessionIdValid() == true` even though `getRequestedSessionId() == null`.
2. After `changeSessionId()` rotates the session ID, the previously-requested ID no longer matches the new session ID, but `isRequestedSessionIdValid()` still returned `true` unless manually reset.

## Fix

Changed the backing field from `boolean requestedSessionIdValid = true` to `@Nullable Boolean requestedSessionIdValid` (default `null`).

- When **explicitly set** via `setRequestedSessionIdValid(boolean)`, that value is returned verbatim (backwards-compatible override for existing tests that control the flag manually).
- When **null** (the default), `isRequestedSessionIdValid()` derives its value from the actual request state:
  - Returns `false` if `getRequestedSessionId()` is `null` (no session ID submitted by the client).
  - Returns `true` only when a session exists and `requestedSessionId` matches `session.getId()`.
  - Returns `false` in all other cases, including post-`changeSessionId()` scenarios.

## Test plan

- [ ] Existing `MockHttpServletRequestTests` continues to pass (no tests assert on `isRequestedSessionIdValid()`)
- [ ] The two minimal reproductions from the issue should now pass:

```java
// 1. No client session id
assertThat(request.isRequestedSessionIdValid()).isFalse(); // was: true

// 2. After session id rotation
request.setSession(session);
request.changeSessionId();
assertThat(request.isRequestedSessionIdValid()).isFalse(); // was: true
```

## Related Issues

- #36631

### DIFF
--- a/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
+++ b/spring-test/src/main/java/org/springframework/mock/web/MockHttpServletRequest.java
@@ -250,7 +250,7 @@ public class MockHttpServletRequest implements HttpServletRequest {
 
 	private @Nullable HttpSession session;
 
-	private boolean requestedSessionIdValid = true;
+	private @Nullable Boolean requestedSessionIdValid;
 
 	private boolean requestedSessionIdFromCookie = true;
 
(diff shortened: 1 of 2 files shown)


sbrannen:
I have not reviewed the PR, but in any case it should not contain the "Document rendering-scope limitation of WebFlux RequestContext.changeLocale()" commit. Please remove that and force push the changes to the PR's branch.

sbrannen:
All commits in PRs must be signed off in order to pass the [DCO check](https://github.com/spring-projects/spring-framework/pull/36895/checks?check_run_id=80459506107).

That applies to this PR as well as the 4 other PRs you submitted.

bclozel:
Closing because DCO is not signed.

Issue #36684, opened 2026-04-21
Improve documentation to clarify coordination of multiple transaction managers


quaff:
Given two transaction managers `tm1` and `tm2`, it's unclear that a transaction created by `tm2` will or not participate in an existing transaction created by `tm1`. According to my investigation, the answer is yes if they share the same underlying resource, such as `DataSourceTransactionManager` for example: 

```java
	@Test
	public void test(@Autowired DataSource dataSource) {
		
		DataSourceTransactionManager tm1 = new DataSourceTransactionManager(dataSource);
		DataSourceTransactionManager tm2 = new DataSourceTransactionManager(dataSource);
		
		TransactionTemplate tt1 = new TransactionTemplate(tm1);
		TransactionTemplate tt2 = new TransactionTemplate(tm2);
		
		tt1.executeWithoutResult(s1 -> {
			assertThat(s1.isNewTransaction()).isTrue();
			tt2.executeWithoutResult(s2 -> {
				assertThat(s2.isNewTransaction()).isFalse();
			});
		});
	}
```

It would be nice if the reference manual expounded on it and the Javadoc mentioned it.

PR #36896, opened 2026-06-10
GH-36684: Document transaction coordination between DataSourceTransactionManager instances sharing a DataSource

won-seoop:
## Summary

Addresses #36684.

When two `DataSourceTransactionManager` beans are configured with the **same** `DataSource` object, the second manager's `getTransaction()` will detect and join the connection already bound to the thread by the first manager — because both use the same `TransactionSynchronizationManager` key (the `DataSource` instance). The result is that both managers share the same JDBC connection and physical transaction.

This behavior was not documented anywhere in the class-level Javadoc, causing confusion for developers trying to understand whether multiple managers operating on the same underlying resource can/will coordinate. The [issue](https://github.com/spring-projects/spring-framework/issues/36684) specifically asks for this to be documented.

This PR adds a **Transaction manager coordination** paragraph to `DataSourceTransactionManager`'s Javadoc explaining:
- Why two managers sharing a `DataSource` object will participate in the same physical transaction.
- The underlying mechanism (`TransactionSynchronizationManager` keyed by `DataSource` instance).
- When to use a distinct `DataSource` (or JTA) if truly independent transactions are required.

## Test plan

- [ ] No behaviour change (Javadoc only); existing tests pass

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceTransactionManager.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/DataSourceTransactionManager.java
@@ -106,6 +106,17 @@ import org.springframework.util.Assert;
  * is available as an extended subclass which includes commit/rollback exception
  * translation, aligned with {@link org.springframework.jdbc.core.JdbcTemplate}.</b>
  *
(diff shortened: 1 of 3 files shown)


bclozel:

Issue #30706, opened 2023-06-20
Document: When is mandatory define each value for SqlConfig.TransactionMode for @SqlConfig?


manueljordan:
Hello Spring Team

The question is available on SO at:

* [When is mandatory define each value for SqlConfig.TransactionMode for @SqlConfig?](https://stackoverflow.com/questions/76382634/when-is-mandatory-define-each-value-for-sqlconfig-transactionmode-for-sqlconfig)

But the current and unique "answer" is not neither valid nor _realistic_ (AI is involved)

Therefore more details for the [SqlConfig.TransactionMode](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/jdbc/SqlConfig.TransactionMode.html) enum would valuable to understand when choice each value explicitly.

Thanks for your understanding.

snicoll:
I agree that there are a lot of information and rules about how this behaves. The reference guide could perhaps use some more examples.

cc @sbrannen

PR #36897, opened 2026-06-10
GH-30706: Document SqlConfig.TransactionMode values with examples

won-seoop:
## Summary

Expands the _Transaction management for `@Sql`_ section in the reference manual with practical guidance on choosing between `INFERRED` and `ISOLATED` transaction modes.

**Changes:**
- Splits the existing single section into two named sub-sections: `INFERRED` (the default) and `ISOLATED`
- Describes the exact inference rules applied under `INFERRED`: participates in the active test transaction when one is present, otherwise starts and immediately commits a new one
- Explains the key use case for `ISOLATED`: when the code under test opens its own transaction and can therefore only see _committed_ data — making the `INFERRED` mode invisible to it
- Adds a complete Java + Kotlin example showing the `ISOLATED` + `AFTER_TEST_METHOD` cleanup pattern
- Updates the note on static imports to cover `INFERRED` in addition to `ISOLATED`

## Related issues

- #30706


bclozel:

Issue #28632, opened 2022-06-15
Add example using @NestedTestConfiguration


sbrannen:
See https://github.com/spring-projects/spring-framework/issues/28611#issuecomment-1154708483 for background.

KetanParmar02:
hey @sbrannen I want to add sample code snippet in this documentation so can you please provide me this issue??

KetanParmar02:
And please tell me further where I have to add the code snippet..

sbrannen:
Hi @KetanParmar02,

Thanks for the offer!

And apologies for the late reply.

> And please tell me further where I have to add the code snippet..

That should probably go in this section:

https://github.com/spring-projects/spring-framework/blob/main/framework-docs/modules/ROOT/pages/testing/testcontext-framework/support-classes.adoc#nested-test-class-configuration

Feel free to submit a PR.

Cheers,

Sam

PR #36898, opened 2026-06-10
GH-28632: Add @NestedTestConfiguration(OVERRIDE) example to docs

won-seoop:
## Summary

Adds a Java and Kotlin code example to the `@Nested` test class configuration section of the reference manual (`support-classes.adoc`) demonstrating `@NestedTestConfiguration(EnclosingConfiguration.OVERRIDE)`.

The existing example only shows the default `INHERIT` behavior (configuration inherited from the enclosing class). The new example shows the complementary `OVERRIDE` case: a `@Nested` class that opts out of inheriting from its enclosing class and declares its own `@SpringJUnitConfig` instead.

The example follows the same "Hello World" `GreetingServiceTests` context established in the section, adding a `SpecialCaseTests` inner class to illustrate the use of `OVERRIDE`.

Also adds a note that `EnclosingConfiguration` is statically imported from `NestedTestConfiguration.EnclosingConfiguration`.

## Related issues

- #28632


bclozel:

PR #36900, opened 2026-06-11
Add regression test for type variable collision across interfaces

pavanB247:
This commit adds a regression test for a type variable collision scenario across multiple interfaces.

The test verifies generic type resolution when multiple interfaces declare type variables with the same name and are implemented by the same class.

The added test ensures that GenericTypeResolver.resolveType(...) continues to resolve the expected type in this scenario.


jhoeller:
We usually include tests in the commit with the actual fix, so we're not independently merging test-only PRs.

Thanks for your efforts in any case!

PR #36901, opened 2026-06-11
Playwright integration with mockmvc

chomnoue:
Provides integration between [MockMvc](https://github.com/spring-projects/spring-framework/compare/overview.html) and [Playwright](https://playwright.dev/) via MockMvcPlaywrightHandler. This allows browser-style end-to-end testing with Playwright against MVC controllers without starting a server.


bclozel:
Thanks for the proposal but we're not interested in providing a playwright integration at this time. It looks like the changes are neatly separated from the existing MockMvc code, so maybe you can consider shipping this as a separate project? Cheers!

PR #36905, opened 2026-06-11
validate samesite attribute in ResponseCookie

dxbjavid:
ResponseCookie validates the cookie name, value, domain and path but the SameSite attribute is appended to the Set-Cookie value as-is, so a value carrying control characters or a semicolon (for instance one taken from an upstream Set-Cookie by the reactive client cookie parsers) can add further attributes or split the header. It seemed safer to check it the same way as the other attributes, so this adds a SameSite check alongside the existing ones in the constructor. Control characters, non-ASCII and ';' are rejected while the usual Strict, Lax and None values still pass.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ResponseCookie.java
+++ b/spring-web/src/main/java/org/springframework/http/ResponseCookie.java
@@ -75,6 +75,7 @@ public final class ResponseCookie extends HttpCookie {
 		Rfc6265Utils.validateCookieValue(value);
 		Rfc6265Utils.validateDomain(domain);
 		Rfc6265Utils.validatePath(path);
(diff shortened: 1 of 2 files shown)


dxbjavid:
Hi maintainers,

Just a friendly follow-up on this PR. When you have a chance, could you please provide an update on its review status?

Happy to make any additional changes if needed.

Thank you!

dxbjavid:
Closing this pull request due to inactivity

bclozel:
we'll have a look in due time.

bclozel:
Thanks @dxbjavid and sorry for the delayed feedback.

dxbjavid:
Thank you so much, @bclozel! No need to apologize at all. I am  just really glad to see this merged and happy to contribute to the project.

PR #36912, opened 2026-06-13
Fix reserveMethodNames to reserve each supplied name

junhyeong9812:
## Overview

`GeneratedClass.reserveMethodNames(String...)` reserves a set of method names so that
generated methods will not collide with them.

## Problem

Inside the per-name loop, the whole varargs array was passed to `MethodName.of(...)` instead
of the current element:

```java
for (String reservedMethodName : reservedMethodNames) {
    String generatedName = generateSequencedMethodName(MethodName.of(reservedMethodNames)); // whole array
    Assert.state(generatedName.equals(reservedMethodName), ...);
}
```

`MethodName.of(String...)` joins all parts into a single camel-case name, so
`reserveMethodNames("apply", "test")` produced `"applyTest"`, and the per-element check
`Assert.state("applyTest".equals("apply"))` failed with `IllegalStateException`. Single-name
calls worked only by accident (a one-element array round-trips to the same name). The only
in-tree caller reserves a single name, so the multi-name path was never exercised.

## Fix

Pass the loop variable so that each supplied name is reserved individually. A regression test
in `GeneratedClassTests` reserves two names.

## Note on intent

This fix assumes `reserveMethodNames` is meant to reserve **each** supplied name individually.
The per-element assertion `Assert.state(generatedName.equals(reservedMethodName))` and the
javadoc ("a set of reserved method names") both indicate this. If the intent was instead to
reserve a single combined name, the loop and the per-element assertion would be inconsistent
and a different fix would be appropriate — happy to adjust if I have misread the intent.

### DIFF
--- a/spring-core/src/main/java/org/springframework/aot/generate/GeneratedClass.java
+++ b/spring-core/src/main/java/org/springframework/aot/generate/GeneratedClass.java
@@ -81,7 +81,7 @@ public final class GeneratedClass {
 	 */
 	public void reserveMethodNames(String... reservedMethodNames) {
 		for (String reservedMethodName : reservedMethodNames) {
-			String generatedName = generateSequencedMethodName(MethodName.of(reservedMethodNames));
+			String generatedName = generateSequencedMethodName(MethodName.of(reservedMethodName));
 			Assert.state(generatedName.equals(reservedMethodName),
 					() -> String.format("Unable to reserve method name '%s'", reservedMethodName));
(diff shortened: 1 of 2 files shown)


bclozel:
You were right on the intent @junhyeong9812 , this is now fixed for the next maintenance release thanks to you!

junhyeong9812:
Thanks for confirming the intent and merging this!

PR #36913, opened 2026-06-13
Fix `OptionalToObjectConverter` applicability check

junhyeong9812:
## Overview

`OptionalToObjectConverter` (registered by `DefaultConversionService` for `Optional` -> `Object`) unwraps an `Optional` and delegates the contained value to the `ConversionService`.

## Problem

`matches()` derived the contained type with `TypeDescriptor.getElementTypeDescriptor()`:

```java
return ConversionUtils.canConvertElements(sourceType.getElementTypeDescriptor(), targetType, this.conversionService);
```

`TypeDescriptor.getElementTypeDescriptor()` only resolves element types for arrays, streams and collections, so for an `Optional` it returns `null`. `ConversionUtils.canConvertElements()` then treats a `null` source element type as "maybe" and returns `true` unconditionally. As a result `matches()` always matched, and `ConversionService.canConvert(Optional<X>, target)` returned `true` even when `X` cannot be converted to `target` — violating the `canConvert` contract (a `true` result implies `convert` is capable of converting), since the subsequent conversion fails.

For example, with the out-of-the-box `DefaultConversionService`:

```java
TypeDescriptor optionalInteger = new TypeDescriptor(
        ResolvableType.forClassWithGenerics(Optional.class, Integer.class), null, null);
cs.canConvert(optionalInteger, TypeDescriptor.valueOf(LocalDate.class)); // true (incorrect)
cs.convert(Optional.of(42), optionalInteger, TypeDescriptor.valueOf(LocalDate.class)); // fails
```

## Fix

Resolve the `Optional`'s element type from its generic and check it against the target, mirroring the sibling `ObjectToOptionalConverter`. A raw `Optional` (or otherwise unresolved element type) stays permissive, since the element type is unknown. A regression test is added to `DefaultConversionServiceTests`.

## Note on impact

This is a contract-correctness fix rather than a fix for data corruption: the conversion itself already fails loudly (the subsequent `convert` throws), so nothing is silently mis-converted. The value is that `canConvert`/`matches` is a *predicate* callers rely on to decide whether to attempt a conversion or take a different path. Returning an incorrect `true` defeats that — it turns what should be a clean negative answer into a deferred exception. Restoring an accurate `matches()` lets callers get the correct answer up front. Impact is bounded, since it only affects `Optional` used as a conversion *source* whose element type is not convertible to the target.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/OptionalToObjectConverter.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/OptionalToObjectConverter.java
@@ -21,6 +21,7 @@ import java.util.Set;
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.core.ResolvableType;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.TypeDescriptor;
 import org.springframework.core.convert.converter.ConditionalGenericConverter;
@@ -51,7 +52,14 @@ final class OptionalToObjectConverter implements ConditionalGenericConverter {
 
 	@Override
 	public boolean matches(TypeDescriptor sourceType, TypeDescriptor targetType) {
-		return ConversionUtils.canConvertElements(sourceType.getElementTypeDescriptor(), targetType, this.conversionService);
+		ResolvableType elementType = sourceType.getResolvableType().getGeneric();
+		if (elementType.resolve() == null) {
+			// Unknown Optional element type (raw Optional, wildcard, or unresolved
+			// type variable): remain permissive.
+			return true;
+		}
+		TypeDescriptor sourceElementType = new TypeDescriptor(elementType, null, null);
+		return ConversionUtils.canConvertElements(sourceElementType, targetType, this.conversionService);
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


sbrannen (changes requested):
Good catch, and thanks for submitting the PR to fix it! 👍 

Can you please add tests that cover the `Optional<?>` and `Optional<? extends Number>` use cases in addition to the raw `Optional` case?

junhyeong9812:
Thanks for the review! I have added the requested tests, plus two closely related cases:

- `canConvertOptionalWithUnboundedWildcardRemainsPermissive`: `Optional<?>` has no usable element type (the `Object` upper bound resolves to `null`), so the converter stays permissive even for a target with no matching converter.
- `canConvertOptionalWithBoundedWildcardReflectsUpperBound`: `Optional<? extends Number>` is decided against the `Number` upper bound — convertible to `String` (asserted end-to-end down to the `"42"` result), not convertible to `LocalDate`, with `convert()` consistently reporting `ConverterNotFoundException`.
- `canConvertRawOptionalRemainsPermissive`: the raw `Optional` case as its own test for visibility.
- `canConvertOptionalWithUnresolvedTypeVariableRemainsPermissive`: an unresolved `Optional<T>`, covering the third shape mentioned in the code comment.

The wildcard descriptors are built from fixture fields via `TypeDescriptor(Field)`, since wildcard types cannot be expressed with `ResolvableType.forClassWithGenerics`. The branch has also been rebased onto the latest main.

sbrannen (approved):

sbrannen:

Thanks

junhyeong9812:
Thank you for reviewing and merging this. The wildcard test suggestions made the coverage much more complete.

PR #36914, opened 2026-06-13
Honor `sourceStart` offset in `AbstractXMLStreamReader#getTextCharacters`

junhyeong9812:
## Overview

`AbstractXMLStreamReader` (the base for `XMLEventStreamReader`, reachable via the public
`StaxUtils.createEventStreamReader(XMLEventReader)`) implements
`getTextCharacters(int sourceStart, char[] target, int targetStart, int length)`.

## Problem

The copy length was capped with `Math.min(length, source.length)`, ignoring `sourceStart`:

```java
char[] source = getTextCharacters();
length = Math.min(length, source.length);
System.arraycopy(source, sourceStart, target, targetStart, length);
```

When `sourceStart > 0` and `sourceStart + length` exceeds the text length, `System.arraycopy`
reads past the end of `source` and throws `ArrayIndexOutOfBoundsException`. This breaks the
`XMLStreamReader#getTextCharacters(int, char[], int, int)` contract, which copies up to `length`
characters starting at `sourceStart` and returns the number actually copied — exactly the pattern
a consumer uses to read a long text node in chunks (`sourceStart += copied`).

## Fix

Cap the length by the number of characters remaining from `sourceStart`
(`Math.min(length, source.length - sourceStart)`). A regression test is added to
`XMLEventStreamReaderTests`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/xml/AbstractXMLStreamReader.java
+++ b/spring-core/src/main/java/org/springframework/util/xml/AbstractXMLStreamReader.java
@@ -190,7 +190,7 @@ abstract class AbstractXMLStreamReader implements XMLStreamReader {
 	@Override
 	public int getTextCharacters(int sourceStart, char[] target, int targetStart, int length) {
 		char[] source = getTextCharacters();
-		length = Math.min(length, source.length);
+		length = Math.min(length, source.length - sourceStart);
 		System.arraycopy(source, sourceStart, target, targetStart, length);
 		return length;
 	}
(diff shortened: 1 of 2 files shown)


junhyeong9812:
A note for triage: this is an `ArrayIndexOutOfBoundsException` on a public API path. `StaxUtils.createEventStreamReader(XMLEventReader)` is public, and `getTextCharacters(int, char[], int, int)` exists precisely for reading a long text node in chunks - the caller advances `sourceStart` and calls again. Any such loop throws as soon as `sourceStart > 0` and the remaining text is shorter than the requested length, because the length cap ignores `sourceStart`. The fix is a one-line change to that cap.

sbrannen:
This has been merged into `7.0.x` and `main` in 2b276311ebe715354a10610c102a09e4c9be89a7, with a revision to the tests in ea2a26206cc109c9dd5c280e3250b40c90848ee0.

Thanks

junhyeong9812:
Thanks for merging this into both branches, and for upgrading the test to a parameterized one covering the boundary cases!

PR #36916, opened 2026-06-13
Balance throttle permits in SimpleAsyncTaskExecutor

junhyeong9812:
## Overview

`SimpleAsyncTaskExecutor` acquires a concurrency throttle permit (`beforeAccess()`) only on the
throttled execution branch, but `TaskTrackingRunnable.run()` released one (`afterAccess()`)
unconditionally. This unbalanced the throttle's concurrency count in two ways.

## Problem

1. **Immediate-timeout task** — with the throttle active and active-thread tracking enabled
   (`taskTerminationTimeout > 0` or `cancelRemainingTasksOnClose`), an immediate-timeout task takes
   the active-thread tracking branch without acquiring a permit, yet `run()` still released one,
   driving the count below zero and progressively defeating the configured concurrency limit. This
   is reachable through the deprecated `execute(Runnable, long)` overload; the regular
   `execute(Runnable)` / `submit(...)` paths use `TIMEOUT_INDEFINITE` and are unaffected.
2. **Cancelled-before-run task** — a throttled task cancelled before it ran (`checkCancelled()`
   threw ahead of the `try/finally` in `run()`) never released its acquired permit.

## Fix

Track whether a permit was acquired and release it in `run()`'s `finally` only in that case, with
the `finally` now also covering `checkCancelled()`, so an acquired permit is always released
exactly once. Deterministic regression tests are added to `SimpleAsyncTaskExecutorTests`.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/task/SimpleAsyncTaskExecutor.java
+++ b/spring-core/src/main/java/org/springframework/core/task/SimpleAsyncTaskExecutor.java
@@ -471,27 +471,27 @@ public class SimpleAsyncTaskExecutor extends CustomizableThreadCreator
 
 		private final @Nullable Future<?> future;
 
-		private final boolean throttled;
+		private final boolean releaseThrottle;
 
-		public TaskTrackingRunnable(Runnable task, @Nullable Future<?> future, boolean throttled) {
+		public TaskTrackingRunnable(Runnable task, @Nullable Future<?> future, boolean releaseThrottle) {
 			Assert.notNull(task, "Task must not be null");
 			this.task = task;
 			this.future = future;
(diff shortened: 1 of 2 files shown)


junhyeong9812:
A note for triage: the net effect is a concurrency limit that quietly stops holding. On the regular throttled path, a task cancelled before it ran never releases the permit it acquired, so the throttle permanently loses capacity. Separately, an immediate-timeout task releases a permit it never acquired and drives the count below zero, loosening the limit; that path is reachable through the deprecated `execute(Runnable, long)`. Either way the effective concurrency stops matching the configured `concurrencyLimit`, and nothing reports it.

The branch still merges cleanly into `main`.

PR #36917, opened 2026-06-13
Parse byte/short/char/boolean/float array attributes

junhyeong9812:
## Overview

On JDK 24+, annotation metadata is read through the `java.lang.classfile` based
`ClassFileAnnotationDelegate` (selected by `MetadataReaderFactoryDelegate`).

## Problem

`parseArrayValue()` converted only `int`, `long` and `double` array attributes into primitive
arrays. `byte`, `short`, `char`, `boolean` and `float` arrays fell through to the default branch
and were boxed into `Byte[]`, `Short[]`, `Character[]`, `Boolean[]` and `Float[]`. When such an
attribute is later accessed, `TypeMappedAnnotation` adapts the value against the declared
attribute type (e.g. `byte[]`) and throws:

```
java.lang.IllegalStateException: Attribute 'byteValue' ... should be compatible with byte[]
but a java.lang.Byte[] value was returned
```

Empty arrays returned `Object[0]` and were handled separately, which masked the issue. The
ASM-based reader returns genuine primitive arrays (see `MergedAnnotationMetadataVisitorTests`),
confirming the intended behavior; the ClassFile reader should match.

## Fix

Add explicit cases for `OfByte`/`OfShort`/`OfChar`/`OfBoolean`/`OfFloat` that build the
corresponding primitive array. (`Stream` offers no `mapToByte`/`mapToShort`/`mapToChar`/
`mapToBoolean`/`mapToFloat`, so a small loop is used, alongside the existing `OfInt`/`OfLong`/
`OfDouble` cases.)

A regression test is added under `src/test/java24` (the `java24Test` source set); it fails on
the current code with the `IllegalStateException` above and passes with the fix.

### DIFF
--- a/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileAnnotationDelegate.java
+++ b/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileAnnotationDelegate.java
@@ -122,6 +122,46 @@ abstract class ClassFileAnnotationDelegate {
 			case AnnotationValue.OfLong _ -> {
 				return stream.map(AnnotationValue.OfLong.class::cast).mapToLong(AnnotationValue.OfLong::longValue).toArray();
 			}
+			case AnnotationValue.OfByte _ -> {
+				List<AnnotationValue> values = arrayValue.values();
+				byte[] result = new byte[values.size()];
+				for (int i = 0; i < result.length; i++) {
+					result[i] = ((AnnotationValue.OfByte) values.get(i)).byteValue();
+				}
+				return result;
+			}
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks @junhyeong9812 , this has been addressed in 7de2b24d81c02681c8e8cb764de0690726eb1a8c already.

junhyeong9812:
Thanks for checking and for the pointer to 7de2b24, @bclozel. Making the pass exhaustive over the sealed `AnnotationValue` hierarchy covers more ground than my patch did - good to see it handled at the root. I appreciate the follow-up.

PR #36918, opened 2026-06-14
Add Cache-Control must-understand directive

heka1024:
This Pull Request has been created to add must-understand directive according to https://www.rfc-editor.org/rfc/rfc9111.html#name-must-understand.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/CacheControl.java
+++ b/spring-web/src/main/java/org/springframework/http/CacheControl.java
@@ -58,6 +58,8 @@ public class CacheControl {
 
 	private boolean noStore = false;
 
+	private boolean mustUnderstand = false;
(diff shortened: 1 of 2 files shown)


bclozel:
> A response that contains the must-understand directive SHOULD also contain the no-store directive.

This is not enforced by this new API in any fashion and we should try to solve this problem.

heka1024:
@bclozel How about applying no-store when must understand is called?

bclozel:
@heka1024 I don't think this would work because you could still end up with invalid combinations, whereas the current API avoids lots of those. I'll have another go at this when I'll process this PR.

bclozel:
Your contribution has been merged @heka1024 , thanks so much!

PR #36919, opened 2026-06-14
Render parameter type names in ClassFileMethodMetadata

junhyeong9812:
## Overview

On JDK 24+, method metadata is read through the `java.lang.classfile` based
`ClassFileMethodMetadata` (in `src/main/java24`).

## Problem

`ClassFileMethodMetadata`'s `toString()` formatted each parameter type as
`packageName() + "." + displayName()`. `ClassDesc.packageName()` returns an empty string for
primitive, array, and default-package types, so parameters were rendered with a leading dot —
`".int"`, `".String[]"` — and reference arrays lost their package. For example
`void sample(int, String[])` rendered as `...sample(.int,.String[])`.

The return type of the same method already uses
`ClassFileAnnotationMetadata.resolveTypeName()`, and the ASM-based reader
(`SimpleMethodMetadataReadingVisitor`) uses `Type.getClassName()`, both producing canonical names
(`int`, `java.lang.String[]`).

## Fix

Map the parameter `ClassDesc`s through `ClassFileAnnotationMetadata.resolveTypeName()` — the same
helper already used for the return type. This affects only the diagnostic `toString()` output; no
metadata logic consumes it. A regression test is added under `src/test/java24`.

## Relation to gh-36577

gh-36577 fixed the same `packageName() + "." + displayName()` problem for the **return type**
(`getReturnTypeName()` and the return-type slot in `toString()`), but the parameter rendering in
`toString()` was left unchanged (`ClassFileMethodMetadata`, parameter mapping). This PR completes
that fix for the parameter types.


bclozel:
Thanks @junhyeong9812 !

junhyeong9812:
Thanks for merging and for the polish!

PR #22638, opened 2019-03-22
Avoid duplicate registration of @ControllerAdvice implementing both RequestBodyAdvice and ResponseBodyAdvice

liquanchen9:
fixbug: duplicate add on ControllerAdvice both RequestBodyAdvice and ResponseBodyAdvice

PR #36920, opened 2026-06-14
Migrate responseBodyAdvice test to Jackson 3 converter

froggy0m0:
## Summary

- Migrate the `responseBodyAdvice` test to use the Jackson 3 converter (`JacksonJsonHttpMessageConverter`) in place of the deprecated Jackson 2 converter.
- Keep coverage for advice that participates in both request and response body handling. gh-22638


froggy0m0:
@sbrannen, could you please take a look at the direction here?

This PR addresses the existing TODO by migrating the test from 
MappingJackson2HttpMessageConverter to JacksonJsonHttpMessageConverter.

I also made sure to preserve the existing gh-22638 coverage for ResponseBodyAdvice.

Since Jackson 2 support is planned to be removed in 7.2, could you advise whether migrating 
this test now is premature, or if you would prefer a different approach?

I'm happy to adjust the PR based on your feedback.

froggy0m0:
Thanks for reviewing and merging this PR.

PR #36923, opened 2026-06-15
Let composite/filtered collections accept null elements

quaff:
1. Keep `FilteredIterator` consistent with `CompositeIterator` to accept `null`
2. Fix warning `Non-null type argument is expected` and `Assigning a class with nullable type arguments when a class with not-null type arguments is expected`


sdeleuze (approved):
I think those refinements make sense to me, just not sure if that should target 7.0.x or 7.1.x. Let's discuss that in our meeting.

PR #36932, opened 2026-06-16
Avoid divide-by-zero in `ExponentialBackOff` jitter

junhyeong9812:
## Overview

When an `ExponentialBackOff` is configured with an `initialInterval` of `0`
and a positive `jitter`, the first `nextBackOff()` throws
`ArithmeticException: / by zero`.

Both settings are individually accepted today:

- `setJitter(...)` only rejects negative values (`Assert.isTrue(jitter >= 0, ...)`).
- There is no validation on `initialInterval`, and with `jitter = 0` an
  `initialInterval` of `0` already works and yields a delay of `0`.

So the combination `initialInterval = 0` and `jitter > 0` is an accepted
configuration, yet it crashes instead of producing a delay — an inconsistency
between the two jitter branches.

## Reproduction

```java
ExponentialBackOff backOff = new ExponentialBackOff();
backOff.setInitialInterval(0);
backOff.setJitter(100);

backOff.start().nextBackOff(); // throws ArithmeticException: / by zero
```

## Root cause

In `ExponentialBackOff.ExponentialBackOffExecution#applyJitter`:

```java
long applicableJitter = jitter * (interval / initialInterval);
```

With `initialInterval == 0`, the first interval is also `0`, so this evaluates
`jitter * (0 / 0)` — an integer division by zero.

## Fix

Guard the division so no jitter scaling is applied when `initialInterval` is `0`,
leaving behavior for positive intervals unchanged:

```java
long applicableJitter = jitter * (initialInterval > 0 ? (interval / initialInterval) : 1);
```

A regression test (`jitterWithZeroInitialInterval`) is added to
`ExponentialBackOffTests`.

---

This is a narrow runtime-crash guard and is orthogonal to the validation design
discussed in #35357 — it stays compatible whichever way that validation is
ultimately implemented.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOff.java
+++ b/spring-core/src/main/java/org/springframework/util/backoff/ExponentialBackOff.java
@@ -311,7 +311,7 @@ public class ExponentialBackOff implements BackOff {
 			long jitter = getJitter();
 			if (jitter > 0) {
 				long initialInterval = getInitialInterval();
-				long applicableJitter = jitter * (interval / initialInterval);
+				long applicableJitter = jitter * (initialInterval > 0 ? (interval / initialInterval) : 1);
 				long min = Math.max(interval - applicableJitter, initialInterval);
 				long max = Math.min(interval + applicableJitter, getMaxInterval());
 				return min + (long) (Math.random() * (max - min));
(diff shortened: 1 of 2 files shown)


sbrannen:
This has been merged into `7.0.x` and `main` in 924849f55b7641ebd8eaacdb3d8c9f18e919d020 and revised in 0d706f8da60c82f0ae33ef66607a2876b7a461f7.

Thanks! 👍 

- See also: #36946

junhyeong9812:
Thank you so much, @sbrannen! I really appreciate you taking the time to review and refine the change. It's an honor to contribute to Spring Framework, and I'll keep cheering for you and the whole team. Thanks again! 🙌

PR #36933, opened 2026-06-16
Close class resource `InputStream` in `ThrowawayClassLoader`

junhyeong9812:
## Overview

`ThrowawayClassLoader#loadClassFromResource` opens an `InputStream` via
`getResourceAsStream(...)` but never closes it. The stream leaks on both the
success path (after `defineClass`) and the `IOException` path, since the
surrounding `try` block has neither a `finally` nor a try-with-resources.

```java
InputStream inputStream = this.resourceLoader.getResourceAsStream(resourceName);
if (inputStream == null) {
    return null;
}
try {
    ...
    return defineClass(name, bytes, 0, bytes.length); // stream never closed
}
catch (IOException ex) {
    throw new ClassNotFoundException(...);             // stream never closed
}
```

## Fix

Adopt the existing `inputStream` variable as a try-with-resources resource so it
is closed on every path, leaving the loading logic unchanged:

```java
try (inputStream) {
    ...
}
```

A regression test (`ThrowawayClassLoaderTests`) verifies the stream is closed by
routing a tracked `InputStream` through the resource fallback path.

### DIFF
--- a/spring-core/src/main/java/org/springframework/aot/nativex/feature/ThrowawayClassLoader.java
+++ b/spring-core/src/main/java/org/springframework/aot/nativex/feature/ThrowawayClassLoader.java
@@ -65,7 +65,7 @@ class ThrowawayClassLoader extends ClassLoader {
 		if (inputStream == null) {
 			return null;
 		}
-		try {
+		try (inputStream) {
 			ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
 			inputStream.transferTo(outputStream);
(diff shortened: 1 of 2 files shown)


sbrannen:

Thanks

junhyeong9812:
Thanks for the review and merge, @sbrannen! I opened #36938 for the related ClassNotFoundException behavior I noticed while working on this — happy to take it further if it's worth pursuing

PR #36934, opened 2026-06-16
Polish PayloadMethodArgumentResolverTests to eliminate null-safety warnings

quaff:
## Overview

1. Argument 'mono' might be null
2. Method invocation 'block' may produce 'NullPointerException'

## Related Issues

- #36319


sbrannen:
Thanks for the PR, @quaff.

Note, however, that we do not enable nullability checks for our test suite. In addition, if we were to "fix" the nullability in this one test class, we'd need to address nullability concerns consistently across the test suite.

Since we do not plan to do that, I am closing this PR.

PR #36935, opened 2026-06-16
Suppress CGLIB validation WARN for lifecycle callbacks

seonwooj0810:
Closes gh-35365

## What's going on

`CglibAopProxy.doValidateClass` walks every method of the proxy super class and, for each `public final` method that overrides an interface, logs at WARN:

```
Unable to proxy interface-implementing method [public final void X.afterPropertiesSet() throws java.lang.Exception]
because it is marked as final, consider using interface-based JDK proxies instead.
```

For methods that come from Spring's *configuration callback* interfaces (`InitializingBean`, `DisposableBean`, `Aware` sub-interfaces, `Closeable`, `AutoCloseable`) that recommendation is misleading on multiple levels:

* The user typically cannot make the method non-final - it's `final` precisely because the framework (or a base class like `WebServiceGatewaySupport`) wants to guarantee the lifecycle step runs.
* JDK proxies wouldn't help: the surrounding class is usually targeted by a class-level pointcut (e.g. `@Timed`, `@Transactional`) which means CGLIB is the only viable proxy mechanism anyway.
* Application code virtually never advises `afterPropertiesSet()` / `destroy()` / `set*Aware(...)` itself, so the missed advice the warning hints at doesn't actually exist.

The result is alarming noise during startup for users who run `WebServiceGatewaySupport`, `JmsGatewaySupport`, or any of the other framework base classes with a `final afterPropertiesSet()` together with a class-level AOP annotation. The reproducer in the linked issue (a bare-bones class with `final afterPropertiesSet()` + `@Timed`) triggers the WARN.

## The change

`doValidateClass` now uses a small helper, `implementsOnlyConfigurationCallbackInterfaces`, that returns `true` when every interface declaring the method is one of the configuration callback interfaces enumerated in `ProxyProcessorSupport#isConfigurationCallbackInterface`. The WARN about an `interface-implementing` final method is skipped in that case, and the existing DEBUG-level diagnostic in `doValidateClass` (which already explains the consequence in detail) is retained for users who want the information.

Two cases continue to log the WARN unchanged:

* Public final methods that don't implement any interface (still bad - the "Public final method [...] cannot get proxied via CGLIB" message).
* Public final methods declared by both a callback interface *and* a user-defined interface with the same signature - the user interface signal takes precedence, so the warning still fires.

The helper deliberately mirrors `ProxyProcessorSupport#isConfigurationCallbackInterface` rather than reaching across to call it, so the public API of `ProxyProcessorSupport` doesn't change and the two classes stay independent.

## Verification

* Added `CglibAopProxyConfigurationCallbackTests` with seven cases covering `InitializingBean`, `DisposableBean`, `BeanFactoryAware`, `Closeable`, a user-only interface, a method shared between a callback and a user interface, and a standalone final method without any matching interface.
* `./gradlew :spring-aop:test` (full module) - green.
* `./gradlew :spring-context:test --tests "*Proxy*Tests" --tests "org.springframework.aop.*"` - green, exercising the CGLIB validation path through the broader proxying machinery.

Verification done: (1) confirmed no in-flight PR via `gh pr list --search "35365 OR final afterPropertiesSet OR doValidateClass"` and the linked-PR query on the issue, (2) no claim comments on the issue thread, (3) fix touches `.java` only (`CglibAopProxy.java` + a new test), (4) grepped main for the `Unable to proxy interface-implementing method` string - still present in `CglibAopProxy.java:295`, (6) maintainer @snicoll transferred the issue to spring-framework with "I wonder if Spring Framework could do something about this" and replicated the noise on a bare-bones project, which is the engagement signal I targeted.

### DIFF
--- a/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java
+++ b/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java
@@ -16,6 +16,7 @@
 
 package org.springframework.aop.framework;
 
+import java.io.Closeable;
 import java.io.Serializable;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -37,6 +38,9 @@ import org.springframework.aop.RawTargetAccess;
 import org.springframework.aop.TargetSource;
 import org.springframework.aop.support.AopUtils;
 import org.springframework.aot.AotDetector;
+import org.springframework.beans.factory.Aware;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cglib.core.ClassLoaderAwareGeneratorStrategy;
 import org.springframework.cglib.core.CodeGenerationException;
 import org.springframework.cglib.core.GeneratorStrategy;
@@ -292,8 +296,10 @@ class CglibAopProxy implements AopProxy, Serializable {
 					if (Modifier.isFinal(mod)) {
 						if (logger.isWarnEnabled() && Modifier.isPublic(mod)) {
 							if (implementsInterface(method, ifcs)) {
-								logger.warn("Unable to proxy interface-implementing method [" + method + "] because " +
-										"it is marked as final, consider using interface-based JDK proxies instead.");
+								if (!implementsOnlyConfigurationCallbackInterfaces(method, ifcs)) {
+									logger.warn("Unable to proxy interface-implementing method [" + method + "] because " +
+											"it is marked as final, consider using interface-based JDK proxies instead.");
+								}
 							}
 							else {
 								logger.warn("Public final method [" + method + "] cannot get proxied via CGLIB, " +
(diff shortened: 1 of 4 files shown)


seonwooj0810:
Gentle ping — open since mid-June.

`CglibAopProxy.doValidateClass` logs at WARN for every `public final` method that overrides an interface, which includes Spring's own configuration-callback interfaces (`InitializingBean`, `DisposableBean`, `Aware` subtypes, `Closeable`/`AutoCloseable`). For those the "consider interface-based JDK proxies instead" advice is misleading and the user usually cannot unmake the method `final` (gh-35365). The PR suppresses the WARN when *every* interface declaring the method is one of those callback interfaces, leaving it intact otherwise.

Is an explicit allowlist acceptable here, or would you prefer a broader rule? Still merges cleanly, checks green.

sbrannen:
Thanks for the PR and the gentle ping @seonwooj0810. 👍 

We'll look into this.

spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java
```diff
@@ -415,6 +421,40 @@ private static boolean implementsInterface(Method method, Set<Class<?>> ifcs) {
...
+		return matched;
+	}
+
+	/**
+	 * Determine whether the given interface is a Spring configuration callback
+	 * interface (i.e. {@link InitializingBean}, {@link DisposableBean},
+	 * {@link Closeable}/{@link AutoCloseable}, or an {@link Aware} sub-interface).
+	 * <p>Mirrors {@code ProxyProcessorSupport#isConfigurationCallbackInterface}.
+	 */
+	private static boolean isConfigurationCallbackInterface(Class<?> ifc) {
```
sbrannen: Instead of duplicating this here, please extract the existing `isConfigurationCallbackInterface(Class)` method into a package-private method with the same signature in `AopProxyUtils`, have the existing `isConfigurationCallbackInterface(Class)` method in `ProxyProcessorSupport` and this class delegate to the new method, and delete this duplicate copy.

sbrannen (changes requested):
I've requested a change regarding the duplicated utility method.

sbrannen:
Hi @seonwooj0810,

Do you have time to address https://github.com/spring-projects/spring-framework/pull/36935#discussion_r3777087735?

Please let us know in either case.

Thanks,

Sam

seonwooj0810:
Hi Sam, thanks for the review — done. Extracted the check into a package-private `AopProxyUtils.isConfigurationCallbackInterface(Class<?>)` and had both `ProxyProcessorSupport` and `CglibAopProxy` delegate to it, removing the duplicate from `CglibAopProxy`. Pushed in 6e677ce.

`./gradlew :spring-aop:test` (full module) and `:spring-aop:checkstyleMain`/`checkstyleTest` are green.

sbrannen (approved):

sbrannen:

This has been merged into `main` in ac95b96c21906f644e85028fa4df96fcb8cc2c65 and slightly revised in 79a75a5762115809bc4dd8e96de39b36e21187a5.

Cheers,

Sam

PR #36939, opened 2026-06-17
Make inner classes in tests static where feasible

quaff:


quaff:
@sbrannen Please review, feel free to close if it's not worthy.

PR #36940, opened 2026-06-17
Introduce TailOrdered as opposite of PriorityOrdered

quaff:
Close GH-34670


sbrannen:
Thanks for the PR, @quaff.

I've not yet looked into the details or discussed it with the team, but at a glance, I don't think `Triviality` would be a good choice for such a feature, even if it's the opposite of "priority".

How about something along the lines of the following?

- `FallbackOrdered`
- `TailOrdered`
- `LowestPrecedenceOrdered`
- ...?

quaff:
> Thanks for the PR, @quaff.
> 
> I've not yet looked into the details or discussed it with the team, but at a glance, I don't think `Triviality` would be a good choice for such a feature, even if it's the opposite of "priority".
> 
> How about something along the lines of the following?
> 
> * `FallbackOrdered`
> * `TailOrdered`
> * `LowestPrecedenceOrdered`
> * ...?

I vote for `TailOrdered`, PR updated.

PR #36941, opened 2026-06-17
Polish DefaultListableBeanFactoryTests

quaff:
Fix that scope is not overridden and asserted.


sbrannen:

Thanks

Issue #26004, opened 2020-10-31
Order of profiles in tests is not preserved


tbalachowski:
Annotation `@ActiveProfiles` doesn't preserve the order of defined profiles in tests in spring boot 2.3.5.RELEASE
I have defined two profiles  (europe and belgium) and marked test class with `@ActiveProfiles({"europe", "belgium"})`
Profiles have been loaded in alphabetical order instead of defined in the annotation.

I wrote test for that case.
It fails due to unexpected behavior
![ffffffff](https://user-images.githubusercontent.com/22240319/97767651-b6372180-1b1d-11eb-902a-4b2c486512c9.png)

Log from context startup during test execution
`2020-10-31 02:00:03.091  INFO 15818 --- [           main] c.e.d.DemoProfilesApplicationTests       : The following profiles are active: belgium,europe`

In spring boot 2.3.4.RELEASE same test pass

sbrannen:
@tbalachowski, thanks for raising the issue.

----

@wilkinsona, since Spring Boot relies on the registration order, we agree that it is a regression and are reverting the corresponding changes made to 5.3.x, 5.2.x, and 5.1.x. 

We would also be grateful if tests could be added to the _Spring Boot Test_ test suite to catch related regressions in the future (and to verify that this has been fixed for Boot).

wilkinsona:
Thanks, Sam. We have numerous tests that verify that changing profile ordering affects property resolution, but hadn't anticipated Framework changing the ordering and doing so only in tests. It feels a little odd to be adding something to Boot to verify Framework's behaviour, but we can certainly do so if you don't feel that Framework's own test suite is sufficient to avoid the problem being reintroduced.

sbrannen:
Hi @wilkinsona,

> It feels a little odd to be adding something to Boot to verify Framework's behaviour, but we can certainly do so if you don't feel that Framework's own test suite is sufficient to avoid the problem being reintroduced.

We think it makes sense for Boot to have tests in place for behavior that is specific to Boot, and this applies to the registration order of active bean definition profiles, since the order does not matter in Spring Framework but does matter in Spring Boot.

PR #36950, opened 2026-06-19
Update Javadoc for `@ActiveProfiles` ordering

alexisgra:
## Overview

In gh-26004, `TreeSet` was replaced by `LinkedHashSet` so that the registration order of active profiles in `@ActiveProfiles` is preserved.

However, the Javadoc in MergedContextConfiguration and WebMergedContextConfiguration still described the old sorting behavior. This PR aligns it with the actual behavior.

Active profiles are part of `MergedContextConfiguration` `equals()`and `hashCode()`, so their order participates in the test context cache key. The previous Javadoc could mislead readers reasoning about why two configurations do or do not share a cached `ApplicationContext`.

I did not find any other occurrences of the outdated doc elsewhere, including in the reference documentation.

## Related Issues

- #26004


sbrannen:

Thanks

PR #36953, opened 2026-06-20
Avoid redundant registration of legacy date converters

### LINKED ISSUE #36951: Avoid redundant registration of Date/Calendar/Long converters in DefaultFormattingConversionService
While debugging `WebConversionService` initialization, I noticed that `DateFormatterRegistrar.addDateConverters(...)` appears to be invoked twice during the setup of the default formatting conversion service.

The call path is:

```
DefaultFormattingConversionService.addDefaultFormatters(FormatterRegistry formatterRegistry)
 ├─ new DateTimeFormatterRegistrar().registerFormatters(formatterRegistry)
 │   └─ DateTimeConverters.registerConverters(registry)
 │       └─ DateFormatterRegistrar.addDateConverters(registry)
 └─ new DateFormatterRegistrar().registerFormatters(formatterRegistry)
     └─ addDateConverters(registry)
```

As a result, the following converters are registered twice:

* `DateToLongConverter`
* `DateToCalendarConverter`
* `CalendarToDateConverter`
* `CalendarToLongConverter`
* `LongToDateConverter`
* `LongToCalendarConverter`

When inspecting the internal converter registry, I can see duplicate converter instances for the same source/target type pairs.

I understand that this does not appear to cause any functional issues, since conversion behavior remains unchanged. However, the duplicate registrations seem redundant and may introduce unnecessary converter instances in the registry.

I may be missing some historical context here, but I wanted to check whether this duplication is expected or if it could be simplified.

arnabnandy7:
Prior to this commit, DateTimeFormatterRegistrar registered legacy date converters via DateTimeConverters.registerConverters, which in turn called DateFormatterRegistrar.addDateConverters.

Since DefaultFormattingConversionService registers both DateTimeFormatterRegistrar and DateFormatterRegistrar, this resulted in redundant double registration of legacy converters (such as DateToLongConverter, CalendarToDateConverter, etc.).

This commit removes the redundant call to addDateConverters from DateTimeConverters.registerConverters, ensuring legacy date converters are only registered by DateFormatterRegistrar.

Closes gh-36951

### DIFF
--- a/spring-context/src/main/java/org/springframework/format/datetime/standard/DateTimeConverters.java
+++ b/spring-context/src/main/java/org/springframework/format/datetime/standard/DateTimeConverters.java
@@ -27,7 +27,6 @@ import java.util.GregorianCalendar;
 
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.core.convert.converter.ConverterRegistry;
-import org.springframework.format.datetime.DateFormatterRegistrar;
 
 /**
  * Installs lower-level type converters required to integrate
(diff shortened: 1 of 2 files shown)


bclozel:
Declining because this has been resolved differently in #36951 and this approach breaks the existing public contracts.

Issue #36929, opened 2026-06-15
Flow→Flux graphql subscription: PropagationContextElement opens Observation scopes asymmetrically, tripping ObservationThreadLocalAccessor.restore assertion


jludvice:
## Summary

In Spring Framework 7.0.8, GraphQL subscriptions returning a Kotlin `Flow<T>` over a WebSocket trip [ObservationThreadLocalAccessor.restore][mc-restore]'s contract assertion:

```
java.lang.AssertionError: Observation <secured.requests> to which we're restoring is not the same as
the one set as this scope's parent observation <graphql.datafetcher>.
Most likely a manually created Observation has a scope opened that was never closed.
```

Downgrading back to Spring Boot `4.0.6` (Spring Framework to `7.0.7`) makes the failure disappear. The change that introduced the regression is [#36667][pr] (`PropagationContextElement` added to the `Flow → Flux` adapter).

## Environment

- Spring Boot: 4.0.7
- Spring Framework: **7.0.8** (7.0.7 works)
- Spring GraphQL: 2.0.4
- Micrometer Observation: 1.16.6
- Micrometer context-propagation: 1.2.1
- kotlinx-coroutines-core / -reactor: 1.11.0
- JDK: Temurin 25 
-  WebSocket transport (graphql-transport-ws)
-  Spring Security on the request chain

## Root cause

A client opens a GraphQL-over-WebSocket connection and sends a `subscribe` frame:

```json
{"id": "1", "type": "subscribe", "payload": {"query": "subscription { events }"}}
```

The server's handler returns a Kotlin `Flow`:

```kotlin
@Controller
class EventsController(private val handler: UpdateHandler) {
    @SubscriptionMapping
    fun subscribeEvents(): Flow<Event> =
        handler.updatedEventFlow              // SharedFlow<Event>
            .filter { it.eventProperty == filterValue }
            .map { it.toGql() }
}

class UpdateHandler {
    private val _messageFlow = MutableSharedFlow<Event>(replay = 0, extraBufferCapacity = 64, onBufferOverflow = BufferOverflow.DROP_OLDEST)
    val updatedEventFlow: SharedFlow<Event> = _messageFlow.asSharedFlow()
}
```

Spring GraphQL hands the returned `Flow` to `kotlinx.coroutines.reactor.FlowAsFlux`. When the WebSocket subscriber calls `request(n)` on the resulting `FlowSubscription`, the following happens synchronously on the WebSocket worker thread:

[...]

PR #36955, opened 2026-06-20
Fix asymmetric Observation scope propagation in Kotlin Coroutines

arnabnandy7:
### Summary
This PR addresses issue #36929 where Kotlin flow-to-flux subscriptions using `PropagationContextElement` open Observation scopes asymmetrically, causing `AssertionError` crashes in `ObservationThreadLocalAccessor.restore`.

### Cause of the Issue
During undispatched coroutine context modifications (e.g., from `ChannelFlowKt.withContextUndispatched` nested inside the flow's execution path), `updateThreadContext` is called nestedly on the same thread. When a coroutine suspends within the nested block:
1. The execution pauses, meaning the `finally` block of the nested block (which calls `restoreThreadContext` to close the nested scope) is NOT executed.
2. The outer coroutine dispatcher's suspension cleanup runs and executes the outer `restoreThreadContext`.
3. Because the nested scope was never closed, the outer `restoreThreadContext` attempts to close the outer scope but finds a dirty thread-local observation stack (with the nested scope still on top), causing Micrometer's assertion to trip.

### Solution
We introduce a static `ThreadLocal<Integer> nestingDepth` tracker inside `PropagationContextElement`:
- When `updateThreadContext` is called, we increment `nestingDepth`.
- If the depth is greater than 1 (meaning propagation is already active on the current thread), we return a no-op `Scope` implementation, avoiding nested/asymmetric thread-local scope pushes.
- For the outermost update (depth = 1), we return a `WrapperScope` that delegates to the original snapshot scope and resets the `nestingDepth` to 0 when closed.

### Verification
Added a unit test `nestedInvocations` in `PropagationContextElementTests` which simulates nested context updates and outer/nested restorations, verifying that no assertion errors are thrown and the Observation stack remains balanced. All tests in `spring-core` compiled and passed.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/PropagationContextElement.java
+++ b/spring-core/src/main/java/org/springframework/core/PropagationContextElement.java
@@ -70,6 +70,8 @@ public final class PropagationContextElement extends AbstractCoroutineContextEle
 	private static final boolean coroutinesReactorPresent = ClassUtils.isPresent("kotlinx.coroutines.reactor.ReactorContext",
 			PropagationContextElement.class.getClassLoader());
 
+	private static final ThreadLocal<Integer> nestingDepth = ThreadLocal.withInitial(() -> 0);
+
 	private final ContextSnapshot threadLocalContextSnapshot;
 
 
@@ -85,6 +87,11 @@ public final class PropagationContextElement extends AbstractCoroutineContextEle
 
 	@Override
 	public ContextSnapshot.Scope updateThreadContext(CoroutineContext context) {
+		int depth = nestingDepth.get();
+		nestingDepth.set(depth + 1);
+		if (depth > 0) {
+			return new NoOpScope();
+		}
 		ContextSnapshot contextSnapshot;
 		if (coroutinesReactorPresent) {
 			contextSnapshot = ReactorDelegate.captureFrom(context);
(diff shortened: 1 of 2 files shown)


bclozel:
@arnabnandy7 I would you like to get your personal opinion on the matter.

Why do you consider this to be an issue in Spring Framework and suggest a fix here? In your experience, is this coroutines should behave with regards to the `PropagationContextElement` implemented in Spring? Can you elaborate?

Your PR currently suggests to maintain a thread local value on the carrier thread on a per element basis. Given that Framework is using `Dispatchers.Unconfined` by default, wouldn't this cause issues on a thread pool?

Thanks

arnabnandy7:
Thanks @bclozel, those are fair concerns. I focused on Spring because the report linked the regression to adding `PropagationContextElement` to the Flow-to-Flux adapter in #36667. Looking at it again, that explains where the failure became visible, but doesn’t establish that the underlying bug belongs in Spring.

I took the reported two updates followed by one restore as a reason to suppress nested updates. That was too strong an assumption without reproducing the coroutine lifecycle. Nested updates can be legitimate, and I haven’t established why the corresponding restoration is missing or delayed.

Your thread-pool concern also exposes a weakness in this approach. The counter is actually static, so it is shared across all element instances on a thread. A nested invocation carrying a different context would be skipped too. With `Dispatchers.Unconfined` allowing execution to resume on different threads, I don’t think the current test gives enough confidence in this guard.

The test only exercises a manually constructed update/restore sequence. I’d like to revisit this with an actual Flow subscription reproduction, including suspension and different nested contexts, before proposing a revised fix. I agree the current implementation needs more investigation.

arnabnandy7:
Closing this PR as we've confirmed the root cause is in `kotlinx.coroutines`'s `withContextUndispatched` scope lifetime across suspensions. 

I've submitted an upstream fix with reproduction tests here: https://github.com/Kotlin/kotlinx.coroutines/pull/4756 (addressing https://github.com/Kotlin/kotlinx.coroutines/issues/4754).

cc @bclozel

PR #36830, opened 2026-05-23
Add `DataSize` converters to `DefaultConversionService`

yeomin4242:
This PR adds default `String` and `Number` to `DataSize` converters to
`DefaultConversionService`.

Spring Boot already provides equivalent converters, and `DataSize` itself
already exposes parsing support via `DataSize.parse(...)`. This change makes
that conversion available through Spring Framework's default conversion service.

Changes:
- Add `StringToDataSizeConverter` delegating to `DataSize.parse(CharSequence)`
- Add `NumberToDataSizeConverter` delegating to `DataSize.parse(source.toString())`
- Register both converters with `DefaultConversionService`
- Add `DefaultConversionService` tests for string, number, empty, and invalid inputs

This intentionally does not move Spring Boot's `@DataSizeUnit` support into
Spring Framework.

Closes gh-28910

Issue #28910, opened 2022-08-02
Add Converter for DataSize


rstoyanchev:
There is one in Spring Boot that can move into the Spring Framework. `DataSize` even has a parse method so it should be just a matter of adapting to `Converter`.

injae-kim:
- [StringToDataSizeConverter](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/StringToDataSizeConverter.java)
- [NumberToDataSizeConverter](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/NumberToDataSizeConverter.java)

spring-boot has above 2 `*ToDataSizeConvert` now.
this issue's goal is to move above 2 `*ToDataSizeConvert` to spring-framework? 🙇

sbrannen:
Superseded by PR:

- #36830

PR #36956, opened 2026-06-22
Refactor `NumberToDataSizeConverter` to use `DataSize.ofBytes(long)` directly

quaff:
## Background

The current implementation of `NumberToDataSizeConverter` is based on Spring Boot's string-based implementation which includes support for `@DataSizeUnit` and therefore works directly with strings instead of numbers.

## Overview

Previously, a `Number` was converted to a `String` and then the string was parsed/matched using regular expressions back into a suitable `long` which was inefficient and also prevented valid data size values such as `10.0`.

To address those issues, this PR refactors `NumberToDataSizeConverter` to use `DataSize.ofBytes(long)` directly, first checking that the supplied `Number` does not have a fractional part.

## Related Issues

- #28910
- #36830

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/support/NumberToDataSizeConverter.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/support/NumberToDataSizeConverter.java
@@ -23,15 +23,20 @@ import org.springframework.util.unit.DataSize;
  * Converts from a {@link Number} to a {@link DataSize}.
  *
  * @author YeongJae Min
+ * @author Yanming Zhou
  * @since 7.1
- * @see DataSize#parse(CharSequence)
+ * @see DataSize#ofBytes(long)
  * @see StringToDataSizeConverter
  */
 final class NumberToDataSizeConverter implements Converter<Number, DataSize> {
 
 	@Override
 	public DataSize convert(Number source) {
-		return DataSize.parse(source.toString());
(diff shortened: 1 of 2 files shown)


quaff:
@sbrannen Please review.

sbrannen (changes requested):
Thanks for the PR.

This is definitely an improvement over converting a `Number` to a `String` and then using regular expressions to match/parse the string back into a suitable `long`. There might be some edge cases around precision or overflow that the new calculation/check doesn't catch, but I think that's negligible for the use case at hand and anyway better than the previous approach.

However, please introduce unit tests which demonstrate the happy path and failure path for the changes (e.g., `new BigDecimal("10")`, `new BigDecimal("10.0")`, `new BigDecimal("10.5")`).

Also, please update the PR's description and the commit message to explain the rationale for the changes, while simultaneously explaining the rationale for the original implementation (i.e., that it was based on Spring Boot's string-based implementation which included support for `@DataSizeUnit`).

sbrannen (approved):
Thanks for making the requested changes

sbrannen:
This has been merged into `main` in c35e17b038da2b99c1f6b40721c973ec976b4eff and revised in a077324670693187ea98c33591e90831e5da12c9.

Thanks

----

I also took the liberty of revising the PR's description as well as the commit message to provide further context.

Issue #36737, opened 2026-04-30
Memory leak when running a lot of JUnit & Spring Boot tests


pkernevez:
I upgraded my project from SpringBoot 3.5.14 / Spring 6.2.18 to Spring Boot 4.0.6 / Spring 7.0.7

When I run my test suite, it fails due to OOM:

```
[ERROR] See /.../target/surefire-reports for the individual test results.
[ERROR] See dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] There was an error in the forked process
[ERROR] java.lang.OutOfMemoryError: Java heap space
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] java.lang.OutOfMemoryError: Java heap space
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:631)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:285)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:250)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeProvider(AbstractSurefireMojo.java:1347)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeAfterPreconditionsChecked(AbstractSurefireMojo.java:1145)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.execute(AbstractSurefireMojo.java:979)
[ERROR] 	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:126)
```

This OOM happens first in our CI (our runners have 4GB of total RAM). I can reproduce it locally by defining a parameter 'Xmx900m' for the forked JVM.

In production, the real Spring Boot application (version 3.5.x) is working perfectly with a `Xmx500m`.

The forked process is consuming more and more memory. This is the graph from gc logs of the test fork JVM.

I dump the memory file on the OOM. This is the first analysis I do:

<img width="1210" height="622" alt="Image" src="https://github.com/user-attachments/assets/c1d7c1cb-4f15-446a-b715-0ac418f6de84" />

**Main problem: 2, 3, 4, and 5**

[...]

PR #36977, opened 2026-06-29
Cache MetadataReader by class name to avoid duplicate metadata

LordKay-sudo:
## Summary

Addresses **duplicate metadata caching** in `CachingMetadataReaderFactory` when the same `.class` file is discovered through different `Resource` representations during classpath scanning (e.g. `ClassPathResource` with a relative path vs `UrlResource` with a filesystem path).

This matches the duplication described by @pkernevez and hypothesized by @sbrannen in gh-36737, and corresponds to **issue 1** as split by @bclozel in https://github.com/spring-projects/spring-framework/issues/36737#issuecomment-4520000000.

- Adds a class-name keyed cache so only one `MetadataReader` (and its `ClassFileAnnotationMetadata` on Java 24+) is retained per class
- Restores class-name caching previously provided by Spring Boot's removed `ConcurrentReferenceCachingMetadataReaderFactory` (spring-boot#47687), while still delegating to the ClassFile-backed factory on Java 24+

## What this does NOT fix

gh-36737 reports several independent memory problems. This PR does **not** address:

- **ClassFileAnnotationMetadata using more memory per class than the ASM implementation** (bclozel's issue 2; workaround: shadow `MetadataReaderFactoryDelegate`)
- **Micrometer global registry pinning contexts** (`management.metrics.use-global-registry=false`; Spring Boot)
- **PostgreSQL JDBC SharedTimer pinning contexts via InheritableThreadLocals** (reported by @pkernevez)
- **Multiple cached ApplicationContext instances** in large test suites

## Reproduction

On `main`, the new tests fail:

```
CachingMetadataReaderFactoryTests > shouldNotDuplicateMetadataForSameClassFromDifferentResources() FAILED
    org.mockito.exceptions.verification.TooManyActualInvocations

CachingMetadataReaderFactoryTests > shouldReuseMetadataReaderForSameClassFromDifferentResources() FAILED
    java.lang.AssertionError
```

With this PR, all four tests in `CachingMetadataReaderFactoryTests` pass.

## Test plan

- [x] `./gradlew :spring-core:test --tests "org.springframework.core.type.classreading.CachingMetadataReaderFactoryTests"`
- [x] `./gradlew :spring-core:test --tests "org.springframework.core.type.classreading.*"`
- [x] Verified reproduction tests fail on `main` before applying the fix

Related to gh-36737

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/type/classreading/CachingMetadataReaderFactory.java
+++ b/spring-core/src/main/java/org/springframework/core/type/classreading/CachingMetadataReaderFactory.java
@@ -17,20 +17,27 @@
 package org.springframework.core.type.classreading;
 
 import java.io.IOException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.DefaultResourceLoader;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.ResourceUtils;
 
 /**
  * Caching implementation of the {@link MetadataReaderFactory} interface,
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal but this PR adds a lot to the implementation while we don't have yet a clear understanding of the source of the duplication. This is also hard coding resource paths in the implementation.

PR #36978, opened 2026-06-29
Add ASM metadata reader fallback and slim ClassFile method metadata

LordKay-sudo:
## Summary

Addresses **bclozel's issue 2** and the ASM workaround discussed in gh-36737:

- Adds `-Dspring.classformat.metadatareader.asm=true` to select the ASM-based `SimpleMetadataReaderFactory` on Java 24+ (official alternative to shadowing `MetadataReaderFactoryDelegate`)
- Reduces retained memory per method in `ClassFileMethodMetadata` by storing int access flags and JVM descriptor strings instead of `AccessFlags` / `MethodTypeDesc` objects

## Context

Several reporters (@bherczyk, @lbenedetto, @pkernevez) resolved OOM by switching back to ASM metadata reading. This provides a supported configuration flag for that workaround while ClassFile memory use is further optimized.

## Test plan

- [x] `./gradlew :spring-core:test --tests "org.springframework.core.type.classreading.*"`
- [x] `./gradlew :spring-core:test --tests "org.springframework.core.type.classreading.MetadataReaderFactoryDelegateTests"`

Related to gh-36737

### DIFF
--- a/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileMethodMetadata.java
+++ b/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileMethodMetadata.java
@@ -20,17 +20,14 @@ import java.lang.classfile.AccessFlags;
 import java.lang.classfile.MethodModel;
 import java.lang.classfile.attribute.RuntimeVisibleAnnotationsAttribute;
 import java.lang.constant.ClassDesc;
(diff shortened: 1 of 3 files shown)


bclozel:
Thanks for the proposal, but I don't think we can accept this: this introduces a dependency on ASM in the ClassFile implementation. The entire goal here is to not rely on ASM for class metadata introspection.

PR #36981, opened 2026-06-30
Introduce `ResolvableType::resolveRequiredGeneric` to return casted non-null value

quaff:
Use case:
```java
abstract class AbstractRepository<T> {

	final Class<T> entityClass;

	AbstractRepository() {
		entityClass = ResolvableType.forClass(getClass()).as(AbstractRepository.class).resolveRequiredGeneric();
	}
}
```


quaff:
@sbrannen Could you take a look?

Issue #36339, opened 2026-02-17
Allow registering serialized lambda metadata through RuntimeHints


odrotbohm:
If you need to declare GraalVM metadata for serialized lambdas, you currently would need that through a custom `reachability-metadata.json` in your application (example [here](https://github.com/sampledump/native-lambdas/blob/main/src/main/resources/META-INF/native-image/foo/bar/reachability-metadata.json)). It would be nice if `RuntimeHints` allowed that through the API.

snicoll:
See https://www.graalvm.org/latest/reference-manual/native-image/metadata/#specifying-metadata-with-json

PR #36989, opened 2026-07-01
Write native configuration file when only lambda hints are present

junhyeong9812:
## Overview

`FileNativeConfigurationWriter` fails to write `reachability-metadata.json` when a `RuntimeHints` instance contains only lambda hints, silently dropping the lambda metadata.

## Problem

`NativeConfigurationWriter.write(RuntimeHints)` only writes the configuration file when `hasAnyHint(hints)` returns `true`:

```java
public void write(RuntimeHints hints) {
	if (hasAnyHint(hints)) {
		writeTo("reachability-metadata.json",
				writer -> new RuntimeHintsWriter().write(writer, hints));
	}
}
```

However `hasAnyHint(hints)` does not check `ReflectionHints.lambdaHints()`:

```java
private boolean hasAnyHint(RuntimeHints hints) {
	return (hints.proxies().jdkProxyHints().findAny().isPresent() ||
			hints.reflection().typeHints().findAny().isPresent() ||
			hints.resources().resourcePatternHints().findAny().isPresent() ||
			hints.resources().resourceBundleHints().findAny().isPresent() ||
			hints.jni().typeHints().findAny().isPresent() ||
			hasAnyDeprecatedHint(hints));
}
```

`RuntimeHintsWriter` already serializes lambda hints (see the `oneLambda` test in `RuntimeHintsWriterTests`), so when a `RuntimeHints` holds only lambda hints registered via `ReflectionHints.registerLambda(...)`, `hasAnyHint` returns `false`, the file is never written, and the lambda metadata is lost. This is not caught on the JVM; it only surfaces as a native-image runtime failure.

## Fix

Include lambda hints in `hasAnyHint()` so the configuration file is written whenever lambda hints are present:

```java
hints.reflection().lambdaHints().findAny().isPresent() ||
```

Adds a `lambdaConfig()` test that registers only a lambda hint and asserts the file is written with the expected JSON. Without the fix the test fails with `NoSuchFileException` because no file is produced.

### DIFF
--- a/spring-core/src/main/java/org/springframework/aot/nativex/NativeConfigurationWriter.java
+++ b/spring-core/src/main/java/org/springframework/aot/nativex/NativeConfigurationWriter.java
@@ -46,6 +46,7 @@ public abstract class NativeConfigurationWriter {
 	private boolean hasAnyHint(RuntimeHints hints) {
 		return (hints.proxies().jdkProxyHints().findAny().isPresent() ||
 				hints.reflection().typeHints().findAny().isPresent() ||
+				hints.reflection().lambdaHints().findAny().isPresent() ||
 				hints.resources().resourcePatternHints().findAny().isPresent() ||
 				hints.resources().resourceBundleHints().findAny().isPresent() ||
 				hints.jni().typeHints().findAny().isPresent() ||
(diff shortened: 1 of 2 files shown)


sbrannen:
Good catch! 👍 

This is effectively a bug that was introduced when the support for lambda hints was added in 7.0.6.

- #36339

junhyeong9812:
Thank you for the review and merge, @sbrannen. Thanks to your work, I get to use a great framework every day. I really appreciate it.

sbrannen:
@junhyeong9812, you're very welcome, and thanks for the kinds words!

PR #36990, opened 2026-07-01
doc: Add RestTestClient setup with @AutoConfigureRestTestClient

Nicklas2751:
When searching for how to set up the new ` RestTestClient`  for a ` @SpringBootTest`  I was writing, I missed an hint to the ` @AutoConfigureRestTestClient`  in the docs.

So I added it. ^^

I am aware that this is the spring framework doc and not the spring boot doc but since the `RestTestClient` is documentet here, I thought that this would be the best place. 

Feel free to throw away this PR if you don't want Spring Boot related stuff in it. :)

### DIFF
--- a/framework-docs/modules/ROOT/pages/testing/resttestclient.adoc
+++ b/framework-docs/modules/ROOT/pages/testing/resttestclient.adoc
@@ -77,6 +77,41 @@ Kotlin::
 ----
 ======
 
+[[resttestclient.springboottest-config]]
+=== Bind to `@SpringBootTest`
+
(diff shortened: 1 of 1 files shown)


sbrannen:
Thanks for the proposal, @Nicklas2751.

Though, as you mentioned, `@AutoConfigureRestTestClient` resides in Spring Boot. Thus, such documentation would need to be added to the Spring Boot project.

Feel free to open an issue or PR in the [Spring Boot project](https://github.com/spring-projects/spring-boot) if you would like to pursue this further.

PR #36994, opened 2026-07-02
Address EditorConfig violations

BaumiCoder:
This Pull Request is part of an academic research project. More preciously it is for my master's thesis, more background information [here](https://codeberg.org/BaumiCoder/Masterthesis).

## Definition: EditorConfig violations

The goal of this Pull Request is to address all [EditorConfig](https://editorconfig.org) violations in the project. With EditorConfig you have set for example `indent_style = tab` for many files, but there were files which use spaces for indentation. I call this an EditorConfig violation as a file content violates the respective EditorConfig property. They can have different kind of impacts. The mixture tabs and spaces can lead to visually wrong indentation.
For example if the file with authored with a tab width of 2 and one line has spaces in use instead of tabs, the visual indentation of these lines are now different as `indent_size = 4` lead to a tab width of 4.

## How to address them?

To check for such violations and fix them I developed the CLI tool [ecformat](https://codeberg.org/BaumiCoder/ecformat). However, there are different ways to address the violations:

1. **Adjust EditorConfig:** The EditorConfig properties for a file can be wrong. For example two trailing spaces in Markdown represent a line break. Therefore, `trim_trailing_whitespace = true` is often wrong for them. The correct way to address that is to set `trim_trailing_whitespace = false`  for them, which avoids automatic removal of such spaces. (I found no such cases in your repository.)
2. **Fix file content:** Another way is to fix the content of the file. Which means for example for `insert_final_newline = true` to insert the final newlines. To do so `ecformat fix` can be helpful.
3. **Ignore files:** For some files it is not meaningful to consider the EditorConfig violations, which `ecformat check` finds. 
   (See the help for `ecformat check --ignore-file` how this would be possible.) There are different reasons to ignore files:
   1. As EditorConfig is only for text files, considering binary files does not make sense.
   2. EditorConfig supports to set the property on the file level. However, there can be files with intentional differences inside. Such files need to be ignored fully by `ecformat` as it currently (version 0.2.0) has no support to ignore only a single property for some files.

## Changes in this Pull Request

For this Pull Request, I focused on the Properties of the current [EditorConfig specification](https://spec.editorconfig.org/#supported-pairs) 0.17.2. You only have some violations regarding indentation. In most cases, these are alignments and not indentation. As `ecformat` is programming language independent, it cannot distinguish between and indentation and alignment. Therefore, it detects them as indentation violations, but these are false-positives.

Common examples for alignment in Java, Kotlin and Groovy are multi line comments, where the `*` are aligned like this:

```java
/*
 * That is some multi line comment and we want to have the * symbols
 * all in the same text column for a nice border.
 */
```

In `.xml` files and `.xsd` files the attributes are often aligned (but not always) like this:

```xml
	   xmlns:context="http://www.springframework.org/schema/context">
```

Therefore, `ecformat` cannot be very helpful here. However, there were some real indentation errors in the `.xml` and `.xsd` files. Mostly due to the use of only spaces for indentation in some lines of a file. For these cases I manually reviewed the changes for `ecformat fix` making adjustments in them were necessary.

If I fixed the indentations in the wrong way for some files, please let me know with a respective Review comment. Then I will adjust it.

## Avoid EditorConfig violations in the future

Such EditorConfig violations can creep in again. Possible reasons can be that the editor of some contributors has no EditorConfig support, or it would be required to install a respective plugin for the support. Furthermore, some people may sometimes use the space key for indentations, not knowing that tab characters are in use.

If you want to avoid them in the future, I would suggest integrating a respective utility. As EditorConfig does not distinguish between indentation and alignment, tools for it, like `ecformat` or [editorconfig-checker](https://editorconfig-checker.github.io), cannot handle these situations. Therefore, a tool specific for formatting XML files would be required, which can be integrated in the CI or directly in your styling related Gradle tasks.

Let me know if you are interested using such a tool and what kind of tool you would prefer. I could add one in this Pull Request or a follow-up Pull Request.

### DIFF
--- a/integration-tests/src/test/resources/org/springframework/beans/factory/xml/component-config.xml
+++ b/integration-tests/src/test/resources/org/springframework/beans/factory/xml/component-config.xml
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <beans xmlns="http://www.springframework.org/schema/beans"
-      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-      xmlns:foo="http://www.foo.example/schema/component"
-      xsi:schemaLocation="
+	   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	   xmlns:foo="http://www.foo.example/schema/component"
+	   xsi:schemaLocation="
 http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd
 http://www.foo.example/schema/component http://www.foo.example/schema/component/component.xsd">
 
-   <foo:component id="bionic-family" name="Bionic-1">
-      <foo:component name="Mother-1">
-      	<foo:component name="Karate-1"/>
-  		<foo:component name="Sport-1"/>
-      </foo:component>
-      <foo:component name="Rock-1"/>
-   </foo:component>
+	<foo:component id="bionic-family" name="Bionic-1">
+		<foo:component name="Mother-1">
+			<foo:component name="Karate-1"/>
+			<foo:component name="Sport-1"/>
+		</foo:component>
+		<foo:component name="Rock-1"/>
+	</foo:component>
 
 </beans>
--- a/integration-tests/src/test/resources/org/springframework/beans/factory/xml/component.xsd
+++ b/integration-tests/src/test/resources/org/springframework/beans/factory/xml/component.xsd
@@ -1,19 +1,19 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 
 <xsd:schema xmlns="http://www.foo.example/schema/component"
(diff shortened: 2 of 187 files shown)


bclozel:
Thanks for raising this, and sorry for the delayed feedback.

While this PR looks valid, I'm going to decline it for the following reasons:

* this PR is quite large and we tend to favor PRs that are contained to a limited number of modules.
* we aren't working on these XML files anymore; they're mostly used for testing our heritage XML configuration model
* we never had editor config issues or complaints so far
* we don't have a way to automatically enforce/check this in the future as far as I know (there is no Gradle plugin for this and we don't have the resources/appetite to create one).

While we're all for consistency, I'm not sure the tradeoffs are in favor of accepting this now.
Thanks for creating this tool! Cheers.

PR #36996, opened 2026-07-02
Publish ApplicationEvents for cache operations

senocak:
This PR introduces ApplicationEvent publishing for cache write operations (put, evict, clear) in Spring's caching infrastructure. Currently, Spring's cache abstraction provides no mechanism to observe or react to cache mutations triggered by `@Cacheable`, `@CachePut`, and `@CacheEvict` annotations.

Spring's cache infrastructure intercepts annotated methods and performs cache operations, but:
- No events are published when an entry is inserted, updated, or evicted
- No method-level context is available to cache decorators or wrappers
- No hooks exist for reactive or asynchronous side effects after cache mutations

Modified Classes
- CacheAspectSupport: Holds an optional ApplicationEventPublisher, publishes events after successful doPut(), doEvict(), and doClear() calls in performCachePut() and performCacheEvicts().
- CacheInterceptor: Implements ApplicationEventPublisherAware to receive the publisher from the ApplicationContext.

All existing cache interceptor tests continue to pass. If no ApplicationEventPublisher is available (e.g., AspectJ mode), events are silently skipped. Existing Cache, CacheManager, and configuration interfaces are unchanged.

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
@@ -35,6 +35,11 @@ import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
+import org.springframework.cache.interceptor.events.CacheClearEvent;
+import org.springframework.cache.interceptor.events.CacheEvictEvent;
+import org.springframework.cache.interceptor.events.CacheOperationEvent;
+import org.springframework.cache.interceptor.events.CachePutEvent;
+import org.springframework.context.ApplicationEventPublisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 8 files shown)


senocak:
@sdeleuze kindly reminder please

bclozel:
Sorry for the late feedback @senocak , but we'd like to understand more before processing this PR.

Can you explain the main use case behind this? Why can't this be done in another way (through a native cache listener for example?). We must consider tradeoffs here, as publishing application events on the application thread for all caching-related events will consume a significant amount of resources during time-sensitive operations.

senocak:
Hello @bclozel 

Thanks for the feedback, and no worries about the delay. My main use case was to have a provider-independent way of observing caches. For example, via Redis, i can use redis listener to observe when something happens the data but that doesn't really help when the underlying cache is Caffeine or other Cache implementation.

My idea is to put a common Spring level mechanism for observing cache operations regardless of the cache provider. It also gives me some useful Spring context, such as which method triggered the cache operation. I agree that publishing an events in application thread for every cache operation could add overhead, especially in high-throughput applications.  I'm happy to update the pr if you have a better approach, or close it if you think this would make things unnecessarily complicated.

Thank you

PR #37005, opened 2026-07-05
Document combining `@Retryable` with proxy-based features

### LINKED ISSUE #35584: Document semantics for combining `@Retryable` with other proxy-based features
Since we introduced `@Retryable`, we have received questions regarding how `@Retryable` works when combined with other proxy-based features.

We should therefore document (and test) the semantics for the combination of such features.

Candidate features to document/test:

- `@Retryable`
- `@Async`
- `@Transactional`
- `@Cacheable`

jhan0121:
Document the semantics of combining `@Retryable` with other proxy-based annotations (`@Transactional`, `@Cacheable`, and `@Async`), addressing the advice chain ordering that determines retry behavior.

### Tests
- Add `withCacheableAnnotation()` to `RetryInterceptorTests` verifying that retry is applied outside the cache interceptor (Retry OUTER -> Cache INNER).
- Add `RetryableTransactionTests` in `spring-tx` verifying that each retry attempt starts a fresh transaction (Retry OUTER -> TX INNER). Placed in `spring-tx` to avoid a circular module dependency.

### Javadoc
- Add a paragraph to `@Retryable` documenting the proxy interaction semantics and pointing to `EnableResilientMethods#order()` and `EnableAsync#order()` for customizing the order relative to `@Async` (the order relative to `@Transactional`/`@Cacheable` is fixed).

### Reference documentation
- Add a "Combining `@Retryable` with Other Proxy-Based Features" section to `resilience.adoc`, covering the default advice chain order for each combination, and advice order customization between `@Retryable` and `@Async` via `@EnableResilientMethods(order)`/`@EnableAsync(order)` (this customization does not apply to `@Transactional`/`@Cacheable`,
  whose shared `InfrastructureAdvisorAutoProxyCreator` always runs first).
- Add cross-reference TIP blocks to the `@Transactional` and `@Async` reference sections pointing to the new combining section.

## Spring Boot compatibility

Verified with a dedicated validation repository (https://github.com/jhan0121/spring-resilience-boot-validation) that the advice chain order is consistent across all Spring Boot 4.x GA releases (4.0.0 through 4.1.0) under auto-configuration:

| Scenario                        | Order                    | Result |
|---------------------------------|--------------------------|--------|
| `@Retryable` + `@Transactional` | Retry OUTER → TX INNER   | PASS   |
| `@Retryable` + `@Cacheable`     | Retry OUTER → Cache INNER | PASS   |
| `@Retryable` + `@Async`         | Async OUTER → Retry INNER | PASS   |

Closes gh-35584


jakubhavlik (approved):

jhan0121:
Hi @sbrannen, just a gentle follow-up the PR has been open for about 10 weeks and CI is green. @jakubhavlik kindly left a review as well. Let me know if there's anything I can clarify or adjust to help with the review!

sbrannen (changes requested):
Thanks for the PR. 👍 

It looks like you forgot to add a tip for `@Cacheable` to the reference docs.

So, please add something like the following in `framework-docs/modules/ROOT/pages/integration/cache/annotations.adoc`.

```markdown
TIP: When `@Cacheable` is combined with `@Retryable`, the retry advice is applied
outermost, so each retry attempt checks the cache before invoking the method. See
xref:core/resilience.adoc#resilience-annotations-retryable-combining-cacheable[Combining `@Retryable` with `@Cacheable`]
for details.
```

Also, please rebase this PR on `7.0.x` and force-push the changes.

spring-context/src/test/java/org/springframework/resilience/RetryInterceptorTests.java
```diff
@@ -639,6 +673,92 @@ public CompletableFuture<Void> retryOperation() {
...
+		public String retryOperation() {
+			if (counter.incrementAndGet() < 2) {
+				throw new IllegalStateException();
+			}
+			return "result";
+		}
+	}
+
+
+	static class CountingCacheManager implements CacheManager {
```
sbrannen: Actually, we probably don't need this custom cache manager.

What about something similar to the following instead?

```java
@Test
void withCacheableAnnotation() {
	AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext();
	ctx.registerBeanDefinition("bean", new RootBeanDefinition(CacheableAnnotatedBean.class));
	ctx.registerBeanDefinition("cacheManager", new RootBeanDefinition(ConcurrentMapCacheManager.class));
	ctx.registerBeanDefinition("config", new RootBeanDefinition(EnablingConfigWithCaching.class));
	ctx.refresh();
	CacheableAnnotatedBean proxy = ctx.getBean(CacheableAnnotatedBean.class);
	CacheableAnnotatedBean target = (CacheableAnnotatedBean) AopProxyUtils.getSingletonTarget(proxy);
	target.cache = ctx.getBean(CacheManager.class).getCache("tests");

	// Simulates the cache being concurrently populated between retry attempts.
	String result = proxy.retryOperation();
	assertThat(result).isEqualTo("concurrently cached value");
	// Only invoked once: the 2nd attempt found the cached value instead of
	// invoking the target again, proving that retry wraps the cache check.
	assertThat(target.counter).hasValue(1);
}
```

```java
static class CacheableAnnotatedBean {
	AtomicInteger counter = new AtomicInteger();
	@Nullable Cache cache;

	@Cacheable("tests")
	@Retryable(maxRetries = 2, delay = 10)
	public String retryOperation() {
		if (counter.incrementAndGet() == 1) {
			this.cache.put(SimpleKey.EMPTY, "concurrently cached value");
			throw new IllegalStateException();
		}
		return "result";
	}
}
```

sbrannen (changes requested):

spring-context/src/test/java/org/springframework/resilience/RetryInterceptorTests.java
jhan0121: Thank you for pointing that out! That's much simpler. I've replaced the custom `CountingCacheManager` with `ConcurrentMapCacheManager` and reworked the test to simulate a concurrently-populated cache by writing into it directly from within the target method, as you suggested.

jhan0121:
Thanks for the review, @sbrannen!
I've added the suggested TIP to `annotations.adoc`, right after the `@Cacheable` overview examples and before the "Default Key Generation" section, linking back to the `resilience.adoc` section you referenced.
also, I rebased onto the latest main instead of 7.0.x directly. (is that okay, or would you prefer I redo it against 7.0.x? I've force-pushed the result)

sbrannen (approved):

sbrannen:

> I rebased onto the latest main instead of 7.0.x directly. (is that okay, or would you prefer I redo it against 7.0.x? I've force-pushed the result)

No worries. I rebased on `7.0.x` for you and re-targeted this PR for `7.0.x` as well, so that you received "Merged" credit in GitHub.

This has been merged into `7.0.x` and `main` with additional cross-references introduced in 2aa36fea64d580d0e2e531ac96e6ed31660bd412.

PR #37006, opened 2026-07-05
Ignore concrete classes in ImportHttpServices scanning

### LINKED ISSUE #36993: ImportHttpServices scans concrete classes implementing @HttpExchange interfaces
spring-web:7.0.8

Image

Image   

When currently attempting to adapt and extend based on an httpclient interface, it will inject this class into the httpServiceProxyRegistry-bean. However, in reality, it is a regular class rather than an interface. Can non-interface classes be filtered out during the filtering process.

Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'httpServiceProxyRegistry': [com.feeyo.common.client.limitx.ClientLimitxService] is not an interface
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1815)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:603)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:525)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:333)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:371)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:331)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:201)
	at org.springframework.web.service.registry.AbstractHttpServiceRegistrar.getProxyInstance(AbstractHttpServiceRegistrar.java:204)

karunsehwag:
## Summary
`AbstractHttpServiceRegistrar`'s classpath scanner uses `AnnotationTypeFilter` with `considerInterfaces=true` so that types can be discovered through an `@HttpExchange`-annotated interface. However, `isCandidateComponent()` did not verify that the discovered type itself was an interface.

As a result, concrete classes implementing an `@HttpExchange` interface were also registered as HTTP service clients, causing proxy creation to fail later with `IllegalArgumentException: [...] is not an interface`.

Fix: `isCandidateComponent()` now also requires `metadata.isInterface()`.

Fixes gh-36993

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/service/registry/AbstractHttpServiceRegistrar.java
+++ b/spring-web/src/main/java/org/springframework/web/service/registry/AbstractHttpServiceRegistrar.java
@@ -348,7 +348,7 @@ public abstract class AbstractHttpServiceRegistrar implements
 		@Override
 		protected boolean isCandidateComponent(AnnotatedBeanDefinition beanDefinition) {
 			AnnotationMetadata metadata = beanDefinition.getMetadata();
-			return (metadata.isIndependent() && !metadata.isAnnotation());
+			return (metadata.isInterface() && metadata.isIndependent() && !metadata.isAnnotation());
 		}
 
 		/**
--- a/spring-web/src/test/java/org/springframework/web/service/registry/HttpServiceRegistrarTests.java
+++ b/spring-web/src/test/java/org/springframework/web/service/registry/HttpServiceRegistrarTests.java
@@ -70,6 +70,16 @@ class HttpServiceRegistrarTests {
 		assertBeanDefinitionCount(3);
 	}
 
+	@Test  // gh-36993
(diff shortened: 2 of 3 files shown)


yashmehla (approved):
The change correctly restricts HTTP Service scanning to interfaces by checking metadata.isInterface() which prevents implementation classes from being registered as HTTP Service clients. The  regression test using EchoAImpl verifies the reported scenario which ensures the intended interfaces are only registered. Looks good to me. Approved.

PR #37012, opened 2026-07-07
Fix missing nullability in JdbcTemplate.batchUpdate

### LINKED ISSUE #37011: Missing Nullability annotation on JdbcTemplate.batchUpdate
From Kotlin, calling JdbcTemplate.batchUpdate with an array containing nulls is impossible because of a missing nullability annotation:

```kotlin
jdbcTemplate.batchUpdate(
  """
  INSERT INTO "my_table" ("non_null", "can_be_null") VALUES (?, ?)
  """,
  listOf(
    arrayOf<Any?>(
      "non-null",
      null,
    )
  )
)
```

Error:
```
None of the following candidates is applicable:
<SNIP>
                                                             vvvvvvvvv
fun batchUpdate(sql: String, batchArgs: (Mutable)List<Array<(out) Any>>): IntArray:
Argument type mismatch: actual type is 'List<Array<Any?>>', but '(Mutable)List<Array<(out) Any>>' was expected.
```

## Proposed solution

Change to ` List<@Nullable Object[]>` here:

https://github.com/spring-projects/spring-framework/blob/1700fad16d2a3d32c8d45159507f79f0b1cdf747/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java#L1048-L1055

alshain:
Updated batchUpdate method signatures to allow nullable Object arrays.

Fixes #37011

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/JdbcTemplate.java
@@ -1046,12 +1046,12 @@ public class JdbcTemplate extends JdbcAccessor implements JdbcOperations {
 	}
 
 	@Override
-	public int[] batchUpdate(String sql, List<Object[]> batchArgs) throws DataAccessException {
+	public int[] batchUpdate(String sql, List<@Nullable Object[]> batchArgs) throws DataAccessException {
 		return batchUpdate(sql, batchArgs, new int[0]);
 	}
 
 	@Override
-	public int[] batchUpdate(String sql, List<Object[]> batchArgs, int[] argTypes) throws DataAccessException {
+	public int[] batchUpdate(String sql, List<@Nullable Object[]> batchArgs, int[] argTypes) throws DataAccessException {
 		if (batchArgs.isEmpty()) {
 			return new int[0];
 		}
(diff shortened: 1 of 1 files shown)


sdeleuze:
@alshain The build seems to be broken by this change, could you please double check locally?

alshain:
Thanks for the near instant reply @sdeleuze . Unfortuntely, our corporate proxy setup would have me chnage everything to our Nexus, that seems infeasible.

I've taken a guess at the needed local nullability annotations. Could you trigger the build again?

sdeleuze:
I am running another build on the CI to check.

All commits must include a Signed-off-by trailer at the end of each commit message, see https://github.com/spring-projects/spring-framework/blob/main/CONTRIBUTING.md#submit-a-pull-request for more details. Maybe squash the 2 commits to a single one with the required trailer and force-push?

sdeleuze:
Still broken, I think you should find a way to run the test locally.

alshain:
@sdeleuze I'll try to do it at home, there it should be no problem.

sdeleuze:
I am currently merging various PRs, so don't bother, I will take it from there and refine.

sdeleuze:
Also note that I will modify your contribution to use `List<? extends @Nullable Object[]>` in order to allow both nullable and non-nullable elements in the generic type. I think this is the first time we have a use case for a list of array, but I think that applies there. I will add a Kotlin test case to double check it works as expected.

PR #37021, opened 2026-07-09
Support arbitrary levels of nested class names in `ClassUtils.forName()`

guanchengang:
Prior to this commit, `ClassUtils#forName` only performed a single-level conversion when resolving nested class names in Java source style (e.g. "`java.lang.Thread.State`"). For multi-level nested classes such as "`com.example.Outer.Middle.Inner`", the method would fail to resolve the actual class "`com.example.Outer$Middle$Inner`" because only the last dot was converted to a dollar sign.

This commit improves the resolution by introducing a loop that iteratively converts dots to dollar signs from right to left, until the class is found or no further conversion is possible. Additionally, if the given name already contains a dollar sign, the method now fails fast without attempting unnecessary conversions.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/ClassUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ClassUtils.java
@@ -67,6 +67,7 @@ import org.springframework.lang.Contract;
  * @author Sam Brannen
  * @author Sebastien Deleuze
  * @author Sungbin Yang
+ * @author Chengang Guan
  * @since 1.1
  * @see TypeUtils
  * @see ReflectionUtils
@@ -308,16 +309,27 @@ public abstract class ClassUtils {
 			return Class.forName(name, false, clToUse);
 		}
 		catch (ClassNotFoundException ex) {
-			int lastDotIndex = name.lastIndexOf(PACKAGE_SEPARATOR);
-			int previousDotIndex = name.lastIndexOf(PACKAGE_SEPARATOR, lastDotIndex - 1);
-			if (lastDotIndex != -1 && previousDotIndex != -1 && Character.isUpperCase(name.charAt(previousDotIndex + 1))) {
-				String nestedClassName =
-						name.substring(0, lastDotIndex) + NESTED_CLASS_SEPARATOR + name.substring(lastDotIndex + 1);
-				try {
-					return Class.forName(nestedClassName, false, clToUse);
+			if (name.lastIndexOf(NESTED_CLASS_SEPARATOR) > 0) {
+				// not java source style
+				throw ex;
+			}
+			String curName = name;
+			for (;;) {
+				int lastDotIndex = curName.lastIndexOf(PACKAGE_SEPARATOR);
+				int previousDotIndex = curName.lastIndexOf(PACKAGE_SEPARATOR, lastDotIndex - 1);
+				if (lastDotIndex != -1 && previousDotIndex != -1 &&
+						Character.isUpperCase(curName.charAt(previousDotIndex + 1))) {
+					curName = curName.substring(0, lastDotIndex) + NESTED_CLASS_SEPARATOR +
+							curName.substring(lastDotIndex + 1);
+					try {
+						return Class.forName(curName, false, clToUse);
+					}
+					catch (ClassNotFoundException ex2) {
+						// Swallow - let original exception get through
+					}
 				}
-				catch (ClassNotFoundException ex2) {
-					// Swallow - let original exception get through
+				else {
+					break;
 				}
 			}
 			throw ex;
--- a/spring-core/src/main/java/org/springframework/util/ClassUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ClassUtils.java
@@ -309,7 +309,7 @@ public abstract class ClassUtils {
 			return Class.forName(name, false, clToUse);
 		}
 		catch (ClassNotFoundException ex) {
-			if (name.charAt(name.length() - 1) != PACKAGE_SEPARATOR) {
+			if (!name.isEmpty() && name.charAt(name.length() - 1) != PACKAGE_SEPARATOR) {
 				StringBuilder nestedClassName = new StringBuilder(name);
 				int lastDotIndex = -1;
 				for (int i = nestedClassName.length() - 1; i >= 0; i--) {
--- a/spring-core/src/main/java/org/springframework/util/ClassUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ClassUtils.java
@@ -273,6 +273,10 @@ public abstract class ClassUtils {
 	 * for primitives (for example, "int") and array class names (for example, "String[]").
 	 * Furthermore, it is also capable of resolving nested class names in Java source
 	 * style (for example, "java.lang.Thread.State" instead of "java.lang.Thread$State").
+	 * <p>Note: When using the Java source style, it assumes standard naming conventions,
+	 * package names all lowercase and class names starting with uppercase. If your class
+	 * or package name deviates from these conventions, use the fully qualified binary
+	 * name instead to ensure correct resolution.
 	 * @param name the name of the Class
 	 * @param classLoader the class loader to use
 	 * (can be {@code null}, which indicates the default class loader)
--- a/spring-core/src/main/java/org/springframework/util/ClassUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ClassUtils.java
@@ -312,7 +312,8 @@ public abstract class ClassUtils {
 			if (!name.isEmpty() && name.charAt(name.length() - 1) != PACKAGE_SEPARATOR) {
 				StringBuilder nestedClassName = new StringBuilder(name);
 				int lastDotIndex = -1;
-				for (int i = nestedClassName.length() - 1; i >= 0; i--) {
+				int i = nestedClassName.length() - 1;
+				for (; i >= 0; i--) {
 					if (nestedClassName.charAt(i) == PACKAGE_SEPARATOR) {
 						char next = nestedClassName.charAt(i + 1);
 						if (Character.isUpperCase(next)) {
@@ -331,6 +332,10 @@ public abstract class ClassUtils {
 							break;
 						}
 					}
+				}
+				if (i == -1 && lastDotIndex != -1 && !Character.isUpperCase(nestedClassName.charAt(0))) {
+					// Single package name, such as a.ClassA
+					nestedClassName.setCharAt(lastDotIndex, PACKAGE_SEPARATOR);
 				}
 				try {
 					return Class.forName(nestedClassName.toString(), false, clToUse);
--- a/spring-core/src/main/java/org/springframework/util/ClassUtils.java
+++ b/spring-core/src/main/java/org/springframework/util/ClassUtils.java
@@ -309,27 +309,34 @@ public abstract class ClassUtils {
 			return Class.forName(name, false, clToUse);
 		}
 		catch (ClassNotFoundException ex) {
-			if (name.lastIndexOf(NESTED_CLASS_SEPARATOR) > 0) {
-				// not java source style
-				throw ex;
-			}
-			String curName = name;
-			for (;;) {
-				int lastDotIndex = curName.lastIndexOf(PACKAGE_SEPARATOR);
-				int previousDotIndex = curName.lastIndexOf(PACKAGE_SEPARATOR, lastDotIndex - 1);
-				if (lastDotIndex != -1 && previousDotIndex != -1 &&
-						Character.isUpperCase(curName.charAt(previousDotIndex + 1))) {
-					curName = curName.substring(0, lastDotIndex) + NESTED_CLASS_SEPARATOR +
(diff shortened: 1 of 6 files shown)


guanchengang:
Hi spring-team,
I'd like to open a design discussion on my PR.
# The issue:
This modification caused `CandidateComponentsIndexerTests#embeddedCandidatesAreDetected` to fail in the spring-context-indexer module.

After investigation, I found the root cause: the test uses mixed-format nested class names like `p.A.B$C` – a format that contains both `$` (JVM binary name style) and `.` (Java source style). The original implementation would convert the last dot to `$`, yielding `p.A$B$C`, and successfully load the class.

However, my PR added a fast-fail logic that rejects names already containing `$` because they are technically not "Java source style" – they are JVM binary names.

The inconsistency in the original implementation:

Under the current implementation, class loading behaves inconsistently:
| Input format | Expected behavior | Actual behavior |
|---|---|---|
| "p.A.B.C" (pure source style, 3-level nested) | Should resolve to p.A$B$C | ❌ Fails to load |
| "p.A$B.C" (mixed format) | Should fail? (not a valid source style) | ✅ Loads successfully | 
| "p.A.B$C" (mixed format) | Should fail? (not a valid source style) | ✅ Loads successfully | 

`p.A$B.C` and `p.A.B$C` work only because the single-level conversion happens to convert the last dot to `$`, which coincidentally aligns with the actual binary name `p.A$B$C` after the second conversion. This is an accidental success, not an intentional design.

This is clearly inconsistent: if we support mixed formats, we should also support pure source-style multi-level names. If we don't, then mixed formats should not work either.

# Questions for discussion:
Should mixed-format names like `p.A$B.C` be supported in` ClassUtils.forName()`?
If yes:
* Multi-level pure source-style names (e.g., `p.A.B.C`) should also be supported for consistency.
* The fast-fail logic should be removed.

If no:
* Mixed-format names should not be supported.
* Internal usages (including `CandidateComponentsIndexerTests`) should be updated to use either pure binary names or pure source-style names.

I'm happy to adapt the PR based on the team's decision. Please let me know your thoughts.

Thanks!

bclozel:
Hello @guanchengang 

First, can you elaborate more on the reason you're sending this PR? Do you have a specific use case for this in your application?

The current changes fail the build, so we cannot accept this contribution at this time, please advise: https://github.com/spring-projects/spring-framework/actions/runs/29003538580/job/86340709623?pr=37021#step:3:576

guanchengang:
Hi @bclozel 
In my current project, I have a number of classes that are not managed by the Spring container (i.e., they are neither `@Component` nor registered as beans). Before using them, I need to inspect their metadata (e.g., checking for specific annotations, implemented interfaces, etc.) at runtime. To do this, I have to load them by their fully qualified name as a string and obtain the Class object. So I naturally turned to `ClassUtils.forName()` as a more convenient alternative to plain `Class.forName()`.
While reading the Javadoc, I noticed this statement:
`Furthermore, it is also capable of resolving nested class names in Java source style (for example, "java.lang.Thread.State" instead of "java.lang.Thread$State").`
That led me to believe I could pass nested class names in the usual Java source format, e.g., `com.example.A.B.C`. So I started using that style directly.
However, I quickly discovered that when the nesting depth is greater than two, the resolution fails. Digging into the implementation, I found that ClassUtils.forName() currently only converts the last dot to a $ – meaning it only supports exactly one level of nesting (e.g., `A.B` → `A$B`). For deeper nesting like `A.B.C`, it produces `A.B$C`, which is not a valid binary name for the JVM, and Class.forName() throws ClassNotFoundException.
What made this even more puzzling is that, under this "replace-only-the-last-dot" logic, mixed‑style names like `p.A$B.C` actually succeed. This happens because the single‑dot replacement coincidentally turns it into the correct JVM binary name `p.A$B$C`. So we end up with an inconsistent behaviour: the standard multi‑dot Java source style (which the Javadoc promises) fails, while a non‑standard ad‑hoc mixture of $ and . works by chance. This seems contrary to the documented intent of supporting "Java source style" nested class names.

I'd also like to provide some context about the specific test failure that caused the build to break.
The failing test is in `org.springframework.context.index.processor.CandidateComponentsIndexerTests#embeddedCandidatesAreDetected`. It attempts to load a class with the name `org.springframework.context.index.sample.Another$AnotherPublicCandidate`. However, the actual binary name of that nested class (as compiled by the JVM) is `org.springframework.context.index.sample$Another$AnotherPublicCandidate`.
The failure was triggered by the fast‑fail logic I initially added. The intention behind that fast‑fail was sound – `Class.forName()` is a relatively heavy operation, and early rejection of obviously malformed names can save unnecessary lookups. But in this case, it rejected a name that could have been resolved if we simply tried the right transformation.
Consider a name like `org.springframework.context.index.sample$Another.AnotherPublicCandidate` (notice the mixed $ and .). Under the current implementation – which only replaces the last dot with a $ – this resolves successfully by sheer coincidence, since the last dot is in exactly the right place. But for `org.springframework.context.index.sample.Another.AnotherPublicCandidate` (pure Java source style, multi‑level), the same logic produces `org.springframework.context.index.sample.Another$AnotherPublicCandidate`, which is still invalid and fails.
For details, please refer to my previous comment.

guanchengang:
I think we should first discuss whether ClassUtils.forName() should support multi-level nested class names in Java source style at all, and then decide which formats should be rejected, or whether we should instead try all possible combinations (i.e., attempt every `.` to`$` conversion permutation).

bclozel:
Indeed the situation is not ideal. I am not sure which solution we should choose. Changing the behavior to be more strict is likely to cause lots of regressions because the behavior has been out so long. What initially was meant as a helpful feature can be seen as inconsistent.

Supporting arbitrary levels of nesting might cause issues: Spring Boot and Spring Framework perform lots of "negative lookups" to check whether a library is present and we should enable its support. Did you check what would happen for lots of negative lookups? Would that cause lots of additional IO and checks?

I'll discuss that with the team but let's keep the discussion going in the meantime.

guanchengang:
You're absolutely right that the situation is far from ideal, and I fully understand the risk of introducing regressions given how long this behavior has been in place.
Regarding the performance concern, repeated negative lookups would be costly. Each failed `Class.forName()` attempt triggers two major overheads:
1. Full Classpath traversal: scanning all directories and JARs .
2. Exception stack trace filling: fillInStackTrace() captures the entire call stack.

I'd like to propose a more performant alternative:
Instead of trying multiple conversions progressively, we could assume the input class name follows the standard Java source style and simply convert all remaining dots after the package name to `$` in a single transformation, then make only one `Class.forName` attempt.
For example:
- Input: `com.example.A.B.C.D` → `com.example.A$B$C$D`
- Input: `com.example.A.B$C.D` (mixed style) → `com.example.A$B$C$D`
- Input: `com.example.A$B.C.D` (mixed style) → `com.example.A$B$C$D`

Benefits of this approach:
- Supports arbitrary nesting depth (any number of A.B.C.D.E...)
- Maintains backward compatibility with existing mixed-style usage
- Only one `Class.forName` attempt – no cascading failures, no multiplied I/O overhead
- Performance remains consistent with the current implementation (just a single string transformation + one lookup)

I think this strikes a good balance between correctness, backward compatibility, and performance. I'd be happy to update the PR with this implementation if the team agrees with the direction.
Looking forward to the team's discussion – thanks again for your time!

bclozel:
Yes please update the PR with that intent. No rush on that!

guanchengang:
@bclozel 
I have updated the PR.
Currently, the logic has covered the parsing of both pure Java source code style (a.b.C.D) and mixed style (a.b$C.D), and the test cases have also been passed.

spring-core/src/main/java/org/springframework/util/ClassUtils.java
```diff
@@ -308,13 +309,31 @@ public static Class<?> forName(String name, @Nullable ClassLoader classLoader)
 			return Class.forName(name, false, clToUse);
 		}
 		catch (ClassNotFoundException ex) {
-			int lastDotIndex = name.lastIndexOf(PACKAGE_SEPARATOR);
-			int previousDotIndex = name.lastIndexOf(PACKAGE_SEPARATOR, lastDotIndex - 1);
-			if (lastDotIndex != -1 && previousDotIndex != -1 && Character.isUpperCase(name.charAt(previousDotIndex + 1))) {
-				String nestedClassName =
-						name.substring(0, lastDotIndex) + NESTED_CLASS_SEPARATOR + name.substring(lastDotIndex + 1);
+			if (name.charAt(name.length() - 1) != PACKAGE_SEPARATOR) {
```
sbrannen: It seems that would now throw a `StringIndexOutOfBoundsException` instead of `ClassNotFoundException` if the `name` were an empty string.

spring-core/src/main/java/org/springframework/util/ClassUtils.java
```diff
@@ -308,13 +309,31 @@ public static Class<?> forName(String name, @Nullable ClassLoader classLoader)
...
+				StringBuilder nestedClassName = new StringBuilder(name);
+				int lastDotIndex = -1;
+				for (int i = nestedClassName.length() - 1; i >= 0; i--) {
+					if (nestedClassName.charAt(i) == PACKAGE_SEPARATOR) {
+						char next = nestedClassName.charAt(i + 1);
+						if (Character.isUpperCase(next)) {
```
sbrannen, line 323: How would this new algorithm handle package names that start with uppercase letters like `org.example.A`, where `A` is a package instead of a type?

For example, would `org.example.A.MyClass.NestedClass` still resolve to `org.example.A.MyClass$NestedClass`?

In any case, we'll need to put some thought into the topic of package name conventions where we assume package names are always lowercase. One option would be to require that, but then we'd need to clearly document and test that.

guanchengang: Currently, the algorithm does not handle package names that start with uppercase letters.
To fully support uppercase package names, the algorithm would need to repeatedly attempt `Class.forName` at every possible split point to determine whether a segment is a package or a class. While this is technically feasible (like the first commit), it would introduce significant I/O overhead and performance degradation, especially when resolving deeply nested types.

Given this trade-off, I think this requires broader team discussion to decide whether we want to:

1. Accept the performance cost and support all valid Java package names, regardless of naming conventions, or
2. Require package names to be all lowercase (consistent with JLS §6.1 recommendations) and document this constraint clearly.

Personally, I lean toward option 2, as it keeps the implementation simple and performant while aligning with widely adopted Java conventions. If the team agrees with this approach, I will update the Javadoc for this method and add corresponding tests.

spring-core/src/main/java/org/springframework/util/ClassUtils.java
guanchengang: I fixed this case

spring-core/src/main/java/org/springframework/util/ClassUtils.java
guanchengang: I have submitted new tests and supplemented the Javadoc. During this process, I noticed a few other issues. Since current implementation searches from right to left, some special cases can still be loaded successfully.

|Package Name|Class Name|Java Source Style|
|---|---|---|
|a.B.c|Cls$Inner|a.B.c.Cls.Inner|
|_a|Cls$Inner|_a.Cls.Inner|

However, we are only given a single string and cannot reliably distinguish the package part from the class name. As a result, it is extremely challenging to guard against all non‑standard naming conventions.
For such non‑standard classes, the fully qualified binary name should be used directly to avoid resolution ambiguity.

bclozel:
We discussed this topic today as a team and came up with the following.

First, we agree that the current situation is not right. The Javadoc is at the very least misleading because it states that "it can resolve nested class names" (inferring, all nesting levels when it only works for one level).

The main use cases for this utility are:
* Framework loading class from `spring.factories` files
* Framework loading application classes from XML bean definitions
* direct usage from applications/libraries

The current implementation:
* will resolve `"org.example.Type.SubType"` to `"org.example.Type$SubType"`
* but it won't resolve `"org.example.Type.SubType.OtherType"` to `"org.example.Type$SubType$OtherType"`.
* failing to load `"org.example.Type"` will not try to load `"org.example$Type"` because the implementation looks at the previous token and considers it a package name 

If we assume that class names are JLS compliant in this utility (and turning "." into "$" in the expected places), this could have the following effects:
* a bit less lookups for the inner classes cases, since we will load directly `"org.example.Type$SubType"` without even trying `"org.example.Type.SubType"`
* asking for `"org.example.PackageName.Type.SubType"` (note the uppercase package name) will fail trying to load `"org.example.PackageName$Type$SubType"` whereas currently this "works" by loading `"org.example.PackageName.Type$SubType"`

So there is one risk of regression if applications try to load an inner class with an uppercase package name through XML bean definitions or Spring factories.

We considered several options:
1. update the Javadoc to reflect the current behavior
2. support any level of nested inner classes, but keep compatibility at the cost of some extra lookups
3. change the behavior completely and risk regressions for inner classes + upper case package names
4. use a package private copy of that code for the most sensitive "heritage" code paths like bean definitions and change the behavior for ClassUtils itself

The group wasn't full today so we'll get the opinion of other team members before we move forward with a concrete plan.

PR #37037, opened 2026-07-10
Avoid deprecated extractor in StatusHandler

### LINKED ISSUE #37010: Stop using deprecated `HttpMessageConverterExtractor` in `StatusHandler`
## Overview

The `initBodyConvertFunction(...)` method in `StatusHandler` still uses the recently deprecated `HttpMessageConverterExtractor` (which has no documented replacement).

We should consider replacing that with something custom to avoid using an officially deprecated API in production code.

## Related Issues

- #36574

sakshichitnis27:
## Summary
- Replace the deprecated `HttpMessageConverterExtractor` usage in `StatusHandler` with local response body conversion.
- Preserve support for standard, generic, and smart message converters, including existing empty-body and error handling behavior.
- Add focused tests for the replacement conversion paths.

## Testing
- `./gradlew :spring-web:test --tests org.springframework.web.client.StatusHandlerTests`
- `./gradlew :spring-web:check`

Closes #37010

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/StatusHandler.java
+++ b/spring-web/src/main/java/org/springframework/web/client/StatusHandler.java
@@ -19,12 +19,15 @@ package org.springframework.web.client;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.Type;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal @sakshichitnis27 - I went a different route in 51c4539bb61f3c86cbc59b22f002d011056c1e7c - instead of copying the now deprecated implementation, I extracted the relevant code from `DefaultRestClient` and shared it. This also aligns the behavior between proper body extraction and the status handler code path. Cheers!

PR #37059, opened 2026-07-16
Release Jackson BufferRecycler to its pool in encoders

samueldlightfoot:
## Summary

Both encoders acquire a pooled BufferRecycler via factory._getBufferRecycler() but never return it. Jackson 3 changed the default pool from a ThreadLocal to a per-factory ConcurrentDequePool that only refills on an explicit releaseToPool(), so the gap leaves the pool empty and every encode allocates a fresh recycler and buffers.

Hold the recycler in a local and release it once the generator and byte builder are done: in encodeValue's finally, and in the streaming path's doAfterTerminate after generator.close() and byteBuilder.release(). releaseToPool() is idempotent, so sharing the recycler with the generator cannot double-release.

## Allocation profile

**Setup:** Bytes allocated per `encodeValue` call, measured with HotSpot `ThreadMXBean.getThreadAllocatedBytes()` single-threaded at steady state (200k warmup, 500k ops/round × 7, median), using a default Jackson 3 `JsonMapper`; the only difference between arms is the added `BufferRecycler.releaseToPool()`.

| Payload (JSON size) | Before | After | Saving |
|---|---:|---:|---:|
| Single Pojo (25 B) | 18,720 B/op | 800 B/op | −17,920 B/op (−96%) |
| 200-Pojo list (9,381 B) | 45,992 B/op | 27,578 B/op | −18,414 B/op (−40%) |

The fix removes a fixed ~18 KB of recycler-buffer allocation per encode — constant in absolute terms, so the percentage just tracks payload size.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/AbstractJacksonEncoder.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/AbstractJacksonEncoder.java
@@ -32,6 +32,7 @@ import tools.jackson.core.JacksonException;
 import tools.jackson.core.JsonEncoding;
 import tools.jackson.core.JsonGenerator;
 import tools.jackson.core.exc.JacksonIOException;
+import tools.jackson.core.util.BufferRecycler;
 import tools.jackson.core.util.ByteArrayBuilder;
 import tools.jackson.databind.JavaType;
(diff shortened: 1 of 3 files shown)


bclozel:
Thanks so much @samueldlightfoot for your contribution!

PR #37067, opened 2026-07-18
Add `@Nullable` annotations when treating `Map.remove()` as returning `@Nullable`

msridhar:
The next release of NullAway treats `Map.remove` as returning `@Nullable`, leading to errors at these locations.  Fix them by marking the return type as `@Nullable`.


msridhar:
This PR will fix the build failures when updating to NullAway 0.13.8

sbrannen:

Thanks

PR #37069, opened 2026-07-19
Encode quoted pairs in ContentDisposition name

### LINKED ISSUE #37064: Ensure parsing/tostring symmetry in ContentDisposition
Currently the parsing/toString process aren't strictly symmetric when it comes to encoded pairs. Wre should ensure that this is always the case, otherwise building and parsing two similar instances won't result in the same value or would not honor the `equals` contract.

kamilkrzywanski:
## Overview

Fixes #37064.

`ContentDisposition` already applied quoted-pair encoding/decoding for the `filename` parameter, but the `name` parameter was written and read as a raw quoted string. As a result, a name containing `"` or `\` produced an invalid header and failed to round-trip through `toString()` / `parse()`, which also broke `equals` for otherwise equivalent instances.

## Changes

- Encode `name` with `encodeQuotedPairs` in `toString()`
- Decode `name` with `decodeQuotedPairs` in `parse()` when backslashes are present (same approach as `filename`)
- Add regression tests for formatting, parsing, and build/parse symmetry

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
+++ b/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
@@ -172,7 +172,7 @@ public final class ContentDisposition {
 			sb.append(this.type);
 		}
 		if (this.name != null) {
-			sb.append("; name=\"").append(this.name).append('\"');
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks but this issue is already assigned and I have the changes ready.

PR #37070, opened 2026-07-19
Isolate known attribute names for type-based @SessionAttributes by handler type

froggy0m0:
## Problem

`SessionAttributesHandler` manages the session attributes that a specific handler declares through `@SessionAttributes`.

With type-based `@SessionAttributes`, the actual model attribute names are discovered at runtime.
`SessionAttributesHandler` therefore stores these known attribute names in the session 
so they can be restored in distributed session scenarios.

However, all handler types currently store their known attribute names under the same `SESSION_KNOWN_ATTRIBUTE` key. 
As a result, one handler can treat names recorded by another handler as its own.

Consequently, one handler can restore session attributes managed by another.
If that handler subsequently calls `SessionStatus#setComplete()`, 
the cleanup that follows can also remove session attributes managed by the original handler.

## Changes

Include the handler type name in the session key used to store and restore known attribute names.

This keeps the known attribute names for type-based `@SessionAttributes` isolated by handler type.
It does not change how the actual session attributes are stored or how name-based 
`@SessionAttributes` declarations behave.

## Tests

Add regression tests that verify the following:

- Session attributes managed by a different handler type are not restored.
- Cleanup for one handler type does not remove session attributes managed by another handler type.

---

I would appreciate your review of these changes. I can make changes if needed.

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/method/annotation/SessionAttributesHandler.java
+++ b/spring-web/src/main/java/org/springframework/web/method/annotation/SessionAttributesHandler.java
@@ -69,6 +69,8 @@ public class SessionAttributesHandler {
 
 	private final SessionAttributeStore sessionAttributeStore;
 
+	private final String sessionKnownAttribute;
+
 
 	/**
 	 * Create a new session attributes handler. Session attribute names and types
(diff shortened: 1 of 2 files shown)


yashsiwacha:
This is an elegant and robust solution. Isolating the known session attributes by appending `handlerType.getName()` to the session key prevents side-effects in multi-controller/distributed session scenarios.

I verified the changes:
1. The new class-level key (`SessionAttributesHandler.class.getName() + ".KNOWN." + handlerType.getName()`) provides complete isolation per controller type.
2. The regression tests in `SessionAttributesHandlerTests` comprehensively cover both the leakage on retrieval and the accidental removal during cleanup.

Looks good to merge!

PR #37087, opened 2026-07-25
Use HttpHeaders.copyOf in DefaultClientResponseBuilder

### LINKED ISSUE #37086: ClientResponse.mutate() builder can modify originalResponse headers.
** Bug Reports **

Not sure if this is a bug, but

```java
var mutatedClientResponse = originalClientResponse.mutate().header("xyz", "pqrst").build();
```
updates the headers in the original `originalClientResponse`, too.

---

Doc says
> Return a builder to mutate this response, for example to change
> the status, headers, cookies, and replace or transform the body.

So maybe is this expected behavior.  IMO `mutate()` returning a _builder_ looks like immutable semantics apply.

<details><summary>analysis</summary>

`ClientResponse.mutate()` returns a `DefaultClientResponseBuilder` with a null `headers` field.
https://github.com/spring-projects/spring-framework/blob/9bdeadcfbd32745d790957d7814e4e4389cc8277/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultClientResponseBuilder.java#L101-L115

Headers are implied from `this.originalResponse = other;` and lazily created via getHeaders:
https://github.com/spring-projects/spring-framework/blob/9bdeadcfbd32745d790957d7814e4e4389cc8277/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultClientResponseBuilder.java#L144-L150

This refers to the original headers because `this.headers = unwrap(httpHeaders)` just copies the backing map reference.
https://github.com/spring-projects/spring-framework/blob/9bdeadcfbd32745d790957d7814e4e4389cc8277/spring-web/src/main/java/org/springframework/http/HttpHeaders.java#L477-L482

So updates to the headers goes to the original (shared) backing map reference.
</details>

yashmehla:
Fixes #37086.

`ClientResponse.mutate()` currently modifies the original response's headers because `DefaultClientResponseBuilder` initializes its headers using the `HttpHeaders(HttpHeaders)` constructor, which shares the underlying headers.

This change uses `HttpHeaders.copyOf(...)` so that the builder works with an independent copy of the original response's headers.

A regression test has been added to verify that mutating the builder does not modify the original response's headers or cookies.

### DIFF
--- a/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultClientResponseBuilder.java
+++ b/spring-webflux/src/main/java/org/springframework/web/reactive/function/client/DefaultClientResponseBuilder.java
@@ -144,7 +144,7 @@ final class DefaultClientResponseBuilder implements ClientResponse.Builder {
 	@SuppressWarnings({"ConstantConditions", "NullAway"})
 	private HttpHeaders getHeaders() {
 		if (this.headers == null) {
-			this.headers = new HttpHeaders(this.originalResponse.headers().asHttpHeaders());
+			this.headers = HttpHeaders.copyOf(this.originalResponse.headers().asHttpHeaders());
 		}
 		return this.headers;
 	}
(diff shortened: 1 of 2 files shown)


bclozel:
Thanks for the proposal, but this was submitted before the discussion ended its course and is missing the Cookie fixed I was mentioning in the original issue.

PR #37096, opened 2026-07-31
Preserve async API version resolver order

Gimini-3:
This fixes reactive API version resolution so that resolver registration order is honored.

`DefaultApiVersionStrategy.resolveApiVersion` currently uses `flatMap`, which subscribes to resolver publishers concurrently. As a result, `next()` selects the first emitted value instead of the first non-empty resolver in the configured list. A later resolver can therefore win when it completes faster, and an error from an earlier resolver can be cancelled after a later resolver emits.

This change uses `concatMap` to subscribe to one resolver at a time. That matches the synchronous `resolveVersion` implementation and the documented first-non-null resolver behavior.

The added tests deterministically verify that:

- an earlier resolver wins even while a later resolver can emit immediately
- an error from an earlier resolver is propagated instead of being masked by a later value

The tests use `TestPublisher` rather than wall-clock delays.

Tests:

- `./gradlew :spring-webflux:test --tests org.springframework.web.reactive.accept.DefaultApiVersionStrategiesTests`
- `./gradlew :spring-webflux:check`


bclozel:
Thanks @Gimini-3 this will be shipped in the next 7.0.x maintenance version.

PR #37105, opened 2026-08-03
Fix buffer leak in RSocket SETUP frame handling

yashsiwacha:
Remove the explicit `payload.retain()` call in `MessagingRSocket#handleConnectionSetupPayload` because the setup payload is fully owned by the SocketAcceptor, and `MessagingRSocket#handle` properly strips metadata, wraps the data in a DataBuffer, and eventually releases the initial buffer. The extra `.retain()` caused the underlying Netty buffer reference count to drop to 1 rather than 0 on success.

Additionally, handle early failures in `RSocketMessageHandler#responder` by releasing the connection setup payload explicitly if responder creation fails (e.g. invalid mime types).

Issue: #37026


bclozel:
Thanks but this was already assigned.

Issue #37111, opened 2026-08-04
Class-File metadata reader retains more heap than the ASM variant on Java 24+


perovic:
This is case 2 from https://github.com/spring-projects/spring-framework/issues/36737#issuecomment-4544765280, filed as a separate issue with a minimal reproducer as suggested there.

Affects 7.0.8.

**Expected:** the Class-File and ASM metadata readers retain a comparable amount of heap for the same set of classes.

**Actual:** on Java 24+ the Class-File reader retains considerably more, because the metadata of an annotated class also retains its parsed class file.

## Reproducer

The example below can be run on JDK 25 with `spring-core` 7.0.8 as the only dependency. Each reader reads the same class file repeatedly and retains the resulting metadata, and the heap that stays occupied is reported. `SimpleMetadataReaderFactory` is the ASM reader and serves as a baseline, while `MetadataReaderFactory.create(...)` goes through `MetadataReaderFactoryDelegate`, which on Java 24+ selects `ClassFileMetadataReaderFactory`. The output prints the implementation used in each case.

```java
import org.springframework.core.SpringVersion;
import org.springframework.core.type.AnnotationMetadata;
import org.springframework.core.type.classreading.MetadataReaderFactory;
import org.springframework.core.type.classreading.SimpleMetadataReaderFactory;

import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.util.ArrayList;
import java.util.List;

@Marker
public class Repro {

    public static void main(String[] args) throws Exception {
        ClassLoader cl = Repro.class.getClassLoader();
        System.out.printf("\nJDK %s, spring-core %s\n", Runtime.version(), SpringVersion.getVersion());
        for (MetadataReaderFactory factory :
                List.of(new SimpleMetadataReaderFactory(cl), MetadataReaderFactory.create(cl))) {
            retainedHeap(factory, 100);
            System.out.printf("%-32s %7.1f MB\n", factory.getClass().getSimpleName(),
                    retainedHeap(factory, 10000) / (1024.0 * 1024.0));
        }
    }

    static long retainedHeap(MetadataReaderFactory factory, int count) throws Exception {
        List<AnnotationMetadata> retained = new ArrayList<>(count);
        long before = usedHeap();
        for (int i = 0; i < count; i++) {
            retained.add(factory.getMetadataReader("Repro").getAnnotationMetadata());
        }
        return usedHeap() - before;
    }

    static long usedHeap() throws Exception {
[...]

PR #37112, opened 2026-08-04
Avoid retaining class files in annotation metadata

perovic:
This is the Class-File versus ASM metadata overhead case from #36737, reported with a reproducer in #37111. On Java 24+ `MetadataReaderFactoryDelegate` selects the Class-File reader, which retains considerably more heap than the ASM one for the same classes.

`ClassFileAnnotationDelegate` keeps the raw `java.lang.classfile.Annotation` as the `MergedAnnotation` source. That annotation holds a `Utf8Entry`, so retaining the metadata of an annotated class also retains its class file `byte[]`, the parsed constant pool, and the class model. This stores the declaring class name instead, which is what `SimpleAnnotationMetadataReadingVisitor.Source` holds on the ASM path.

Measured with the reproducer (Temurin 25.0.3, spring-core 7.0.8):

| | retained heap |
|---|---|
| `SimpleMetadataReaderFactory` (ASM) | 12.3 MB |
| `ClassFileMetadataReaderFactory` before | 84.5 MB |
| `ClassFileMetadataReaderFactory` after | 16.7 MB |

In a large application's test context this took the retained size of a single `DefaultListableBeanFactory` from 128.4 MB to 53.4 MB. Forcing the ASM reader in the same build, by copying `MetadataReaderFactoryDelegate` into the project [as suggested here](https://github.com/spring-projects/spring-framework/issues/36737#issuecomment-4395495629), gives 46.1 MB.

The difference that is left is not addressed here. `ClassFileMethodMetadata` holds `AccessFlags` and `MethodTypeDesc` objects and `ClassFileAnnotationMetadata` one `AccessFlags`, where `SimpleMethodMetadata` and `SimpleAnnotationMetadata` keep plain `int` and `String` values instead. #36978 already makes that change for `ClassFileMethodMetadata`. Once both are changed, the reproducer matches the ASM baseline of 12.3 MB and the test context mentioned above comes to 46.9 MB.

### DIFF
--- a/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileAnnotationDelegate.java
+++ b/spring-core/src/main/java24/org/springframework/core/type/classreading/ClassFileAnnotationDelegate.java
@@ -76,7 +76,7 @@ abstract class ClassFileAnnotationDelegate {
 				}
 			}
 			Map<String, Object> compactedAttributes = (attributes.isEmpty() ? Collections.emptyMap() : attributes);
-			return MergedAnnotation.of(classLoader, new Source(annotation), annotationType, compactedAttributes);
+			return MergedAnnotation.of(classLoader, new Source(className), annotationType, compactedAttributes);
 		}
 		catch (ClassNotFoundException | LinkageError ex) {
 			// Non-loadable annotation type -> ignore.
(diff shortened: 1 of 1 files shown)


bclozel:
Thanks @perovic !

PR #37119, opened 2026-08-07
Provide MediaType to AbstractJacksonHttpMessageConverter.customizeReader

### LINKED ISSUE #37094: Provide MediaType to AbstractJacksonHttpMessageConverter.customizeReader
I am implementing a CSV message converter by extending `AbstractJacksonHttpMessageConverter`.

While `customizerWriter` supplies the target `MediaType`, `customizeReader` does not supply the source `MediaType`.

As a result, I cannot check [the "header" parameter](https://www.rfc-editor.org/info/rfc4180/#section-3) to decide the value for `CsvSchema.withUseHeader`.

A new extension point could be added without breaking anything:

```java
protected ObjectReader customizeReader(
        ObjectReader reader, JavaType javaType, @Nullable MediaType contentType) {

    return customizeReader(reader, javaType);
}
```

This could also be of general use for other formats, as media types such as `application/x-my-thing+json` are commonly used, and knowing the base type may be useful in configuring a reader.

yashsiwacha:
Closes gh-37094

This commit adds an overloaded customizeReader method to both AbstractJacksonHttpMessageConverter and AbstractJackson2HttpMessageConverter that accepts a MediaType parameter. This allows subclasses to apply configuration dynamically based on the input content type. The existing 2-argument method is preserved for backward compatibility.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/AbstractJacksonHttpMessageConverter.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/AbstractJacksonHttpMessageConverter.java
@@ -352,7 +352,7 @@ public abstract class AbstractJacksonHttpMessageConverter<T extends ObjectMapper
 			if (hints != null && hints.containsKey(JSON_VIEW_HINT)) {
 				objectReader = objectReader.withView((Class<?>) hints.get(JSON_VIEW_HINT));
 			}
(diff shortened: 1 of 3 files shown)


bclozel:
This issue is now triaged yet, let's consider the issue first before suggesting changes. Thanks.

Issue #36774, opened 2026-05-09
Support generating Leyden AOT cache from integration tests


vpelikh:
## Problem Statement
Spring Boot 3.3 introduced CDS support and laid the groundwork for Project Leyden. Java 24 brought the AOT cache (JEP 483) to production, and Java 25 / 26 have stabilised and enhanced it further — including compiled-code caching via JEP 515 in Java 25. These JDK versions are already released and widely adopted. Today the recommended way to generate an AOT cache is a dedicated training run (e.g. `-XX:AOTMode=record -Dspring.context.exit=onRefresh`) or via Buildpacks environment variables. This workflow adds extra steps to CI/CD pipelines — developers already write thorough integration tests that exercise realistic code paths, yet those tests cannot be used to produce the cache.

Quarkus successfully demonstrated that integration tests are a natural and effective training workload: `@QuarkusIntegrationTest` tests can be run with a single flag to generate `app.aot`. Spring Boot should offer a similarly streamlined experience, allowing teams to leverage their existing integration-test suites to create the AOT cache without additional deployment tooling.

## Use Case
1. A Spring Boot project has a suite of `@SpringBootTest` integration tests that exercise REST endpoints, database access, messaging, etc.
2. The developer (or CI system) runs those tests with a special JVM flag or Maven/Gradle property.
3. During test execution, the JVM records class-loading, method-profiling data, and compiled-code information (Leyden recording mode, with full support from Java 25’s JEP 515).
4. At the end of the test run, an optimised `.aot` (or `.jsa`) cache file is produced.
5. The same cache file is then shipped with the container image, giving the production application the full benefit of Leyden’s startup and warmup improvements.

## Proposed Solution
Spring Boot could provide:

1. **A dedicated test-listener / AotCacheTestExecutionListener** that hooks into the Spring TestContext Framework and sets the correct JVM flags for AOT recording. The Spring Framework already has `TestContextAotGenerator` for build-time AOT processing; a similar runtime listener for Leyden could be introduced.

2. **Maven/Gradle task integration** — analogous to how `BP_JVM_AOTCACHE_ENABLED` works for Buildpacks — so that a command like:

   ```bash
   ./mvnw verify -Dspring.aot.cache.record=true
   ```

   or

   ```bash
   ./gradlew test -PspringAotCache record
   ```

   would automatically set the necessary JVM arguments (`-XX:AOTMode=record`, `-XX:AOTConfiguration=...`, etc.) for the test task.

[...]

PR #37124, opened 2026-08-10
Detect and verify JVM AOT cache recording during integration tests

vpelikh:
Introduce `AotCacheTestExecutionListener`, registered by default in the TestContext framework. It activates automatically when the test JVM is started with AOT cache recording enabled via the JDK 25+ single-step `-XX:AOTCacheOutput=<path>` flag, injected by the Spring Boot Maven or Gradle plugins.

The JVM performs the actual cache generation. This listener coordinates the training run and verifies the outcome:

- Validates the JDK version (JDK 25+, JEP 514).
- Eagerly loads the `ApplicationContext` so that context creation is captured as part of the training workload.
- Warns when the application class loader is not a standard JDK class loader (AOT cache only caches classes loaded by JDK built-in class loaders).
- Warns when `-Dspring.context.exit=onRefresh` is set, since it would terminate the test JVM mid-run.
- Verifies on JVM exit that the cache file was produced at the path specified by the `-XX:AOTCacheOutput` flag.

## Why

This PR is part of #36774. Generating the JVM AOT cache from integration tests lets a realistic test suite exercise a broader portion of the application's startup and request-handling workflow than a simple `-Dspring.context.exit=onRefresh` run, producing a more effective cache. The cache also stores method profiling information, so exercising realistic code paths improves warmup behavior as well.

## How it works

- No Spring configuration or property is required: the JVM flags are the single source of truth for whether recording is enabled and where the cache is written. The listener only detects flags via `RuntimeMXBean.getInputArguments()`.
- By convention, the cache is written to `aot-cache/application.aot` relative to the test working directory, matching the location detected by the Paketo Spring Boot buildpack and bundled into the image by the Spring Boot build plugins.
- The listener is registered by default and is a no-op unless AOT cache recording is enabled, so it does not affect regular test runs.

## Documentation

- `integration/aot-cache.adoc`: new "Generating the cache from integration tests" section.
- `testcontext-framework/tel-config.adoc` and `TestExecutionListener` javadoc: added the listener to the default listener list.

## Notes

- JDK 24 is not supported. It is not an LTS release and it has a different flow for AOT cache generation (JEP 483), so it's not worth adding support for it now; it can be done as a follow-up if needed.

## Related PRs

- Spring Boot PR https://github.com/spring-projects/spring-boot/pull/51101 adds AOT cache support to the Spring Boot Maven and Gradle plugins: a new `aotCacheRecord` option that injects the `-XX:AOTCacheOutput` JVM flag into the test JVM forked by the plugin, so that running tests also records the AOT cache. The PR was closed and needs to be recreated to provide full support for generating the cache from integration tests.
- Paketo buildpack PR https://github.com/paketo-buildpacks/spring-boot/pull/609 adds support in the buildpacks for a pre-recorded AOT cache.

### DIFF
--- a/framework-docs/modules/ROOT/pages/integration/aot-cache.adoc
+++ b/framework-docs/modules/ROOT/pages/integration/aot-cache.adoc
@@ -3,12 +3,11 @@
 :page-aliases: integration/class-data-sharing.adoc
 :page-aliases: integration/cds.adoc
 
-The ahead-of-time cache is a JVM feature introduced in Java 24 via
-https://openjdk.org/jeps/483[JEP 483] that can help reduce the startup time and memory
+The ahead-of-time cache is a JVM feature that can help reduce the startup time and memory
 footprint of Java applications. AOT cache is a natural evolution of
 https://docs.oracle.com/en/java/javase/17/vm/class-data-sharing.html[Class Data Sharing (CDS)].
 Spring Framework supports both CDS and AOT cache, and it is recommended that you use the
-latter if available in the JVM version you are using (Java 24+).
+latter if available in the JVM version you are using (Java 25+).
 
 To use this feature, an AOT cache should be created for the particular classpath of the
 application. It is possible to create this cache on the deployed instance, or during a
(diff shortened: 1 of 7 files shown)


vpelikh:
CC @sdeleuze

vpelikh:
Fixed CI failures in [57d0ee0](https://github.com/spring-projects/spring-framework/pull/37124/commits/57d0ee08dcfbebdc4320cc3ec62266d901c1db06).

sdeleuze:
Declining the PR per https://github.com/spring-projects/spring-framework/issues/36774#issuecomment-5508529440.

PR #37126, opened 2026-08-10
feat(http): support HTTP QUERY method

AchyuthMohan:
Add support for the HTTP QUERY method across Spring's HTTP infrastructure. This change introduces QUERY as a standard HTTP method and integrates it with the relevant request mapping, client, server, and test infrastructure while preserving its safe and idempotent semantics. Includes tests covering QUERY method handling and request processing.


bclozel:
Duplicates #34993

Issue #37123, opened 2026-08-10
Use the 4-argument `Property` constructor in SpEL's `ReflectivePropertyAccessor`


sbrannen:
## Overview

While reviewing #36911, I noticed that `ReflectivePropertyAccessor` currently invokes the 3-argument `Property` constructor which does not accept a predetermined property name and therefore ends up invoking `org.springframework.core.convert.Property.resolveName()`.

However, the invocation of `resolveName()` seems unnecessary, since `ReflectivePropertyAccessor` already knows the name of the property.

## Background

The 3-argument `Property` constructor has been in use since SpEL was introduced; whereas, the 4-argument `Property` constructor was added in 2012, in commit ca017a488056 ("Introduce strategy for BeanInfo creation", SPR-9677) whose commit message states it was added specifically "allowing BeanWrapperImpl to specify the property name at construction time (as opposed to using Property#resolveName(), which supports the JavaBeans contract only)". So, it seems like it was simply an oversight not to migrate to the new `Property` constructor back then.

## Related Issues

- #14311
- #26029
- #36911

lbkulinski:
@sbrannen I'm looking to contribute more to Spring and would like to take this issue on unless you were planning to

sbrannen:
Hi @lbkulinski,

> I'm looking to contribute more to Spring and would like to take this issue on unless you were planning to

Thanks for the offer! 👍 

However, I created the issue and assigned it to myself. So, yes, that indicates that I will do it.

In fact, I completed the work two days ago: https://github.com/sbrannen/spring-framework/commit/570250b7cc9dc709633c20082f08e5cdfe15299c

Regards,

Sam

sbrannen:
After having written the corresponding tests, I determined that this is actually a "bug" rather than a mere "enhancement" due to the following:

- Record accessors whose component names embed or start with "get"/"is" (e.g., `budget()`, `issue()`, etc.) resolve to an incorrect and sometimes empty property name, causing `Property#getField()` to fail to locate the backing field — so its annotations are silently omitted from `Property#getAnnotations()`.
- The same failure occurs for non-record "data classes" using the same bare-accessor convention (e.g., a hand-written `island()` accessor), causing `Property#getField()` to fail to locate the backing field — so its annotations are silently omitted from `Property#getAnnotations()`.
[...]

PR #37129, opened 2026-08-12
Migrate ReflectivePropertyAccessor to use Property's four param constructor to avoid unneeded resolveName() call

lbkulinski:
This PR implements the constructor migration mentioned in #37123. Since we already know the name of the property in `ReflectivePropertyAccessor`, `Property`'s four argument constructor is used to avoid the `resolveName()` call.


sbrannen:
Closing for the reasons given in https://github.com/spring-projects/spring-framework/issues/37123#issuecomment-5265658119.

For future reference, PRs like this must include tests that fail before the changes and pass after the changes.

lbkulinski:
Noted! Thanks!

PR #37132, opened 2026-08-12
Use 4-argument Property constructor in ReflectivePropertyAccessor

### LINKED ISSUE #37123: Use the 4-argument `Property` constructor in SpEL's `ReflectivePropertyAccessor`

yashsiwacha:
Closes gh-37123

### DIFF
--- a/spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectivePropertyAccessor.java
+++ b/spring-expression/src/main/java/org/springframework/expression/spel/support/ReflectivePropertyAccessor.java
@@ -149,7 +149,7 @@ public class ReflectivePropertyAccessor implements PropertyAccessor {
 		if (method != null) {
 			// Treat it like a property...
 			// The readerCache will only contain gettable properties (let's not worry about setters for now).
-			Property property = new Property(type, method, null);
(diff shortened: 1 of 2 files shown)


lbkulinski:
Already completed per https://github.com/spring-projects/spring-framework/issues/37123#issuecomment-5265658119. I also tried to fix it haha

PR #37133, opened 2026-08-13
Ensure parsing/toString symmetry in ContentDisposition name

### LINKED ISSUE #37064: Ensure parsing/tostring symmetry in ContentDisposition

yashsiwacha:
Closes gh-37064

### Overview
Currently, the parsing/`toString()` process for the `name` parameter in `ContentDisposition` is not symmetric when it comes to quoted-pairs (characters like `\"` and `\\`). 

While `ContentDisposition` already correctly encodes and decodes quoted-pairs for the `filename` parameter, the `name` parameter was written and read as a raw quoted string. As a result, formatting a name containing a quote or backslash produces malformed HTTP headers, and parsing them back yields corrupted names containing unescaped backslashes.

### Changes
- **Formatting**: Updated `ContentDisposition.toString()` to encode the `name` parameter using `encodeQuotedPairs(this.name)`.
- **Parsing**: Updated `ContentDisposition.parse()` to decode the parsed `name` value using `decodeQuotedPairs(value)` if a backslash is present.
- **Testing**: Added `formatAndParseWithNameWithQuotes` in `ContentDispositionTests` to assert formatting and parsing symmetry under nested quotes and backslashes.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
+++ b/spring-web/src/main/java/org/springframework/http/ContentDisposition.java
@@ -172,7 +172,7 @@ public final class ContentDisposition {
 			sb.append(this.type);
 		}
 		if (this.name != null) {
-			sb.append("; name=\"").append(this.name).append('\"');
+			sb.append("; name=\"").append(encodeQuotedPairs(this.name)).append('\"');
 		}
 		if (this.filename != null) {
 			if (this.charset == null || StandardCharsets.US_ASCII.equals(this.charset)) {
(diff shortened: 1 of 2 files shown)


bclozel:
See https://github.com/spring-projects/spring-framework/pull/37069#issuecomment-5016417421

PR #37134, opened 2026-08-13
Update `MockCookie#parse(String)` validation

ngocnhan-tran1996:
https://github.com/spring-projects/spring-framework/blob/68e6acd37ed0c12e395ef96d170971028e727383/spring-test/src/main/java/org/springframework/mock/web/MockCookie.java#L130-L132

Based on the Javadoc, I think it should throw `Set-Cookie header must not be null or empty` instead of `Invalid Set-Cookie header ''` when setCookieHeader is empty.

### DIFF
--- a/spring-test/src/test/java/org/springframework/mock/web/MockCookieTests.java
+++ b/spring-test/src/test/java/org/springframework/mock/web/MockCookieTests.java
@@ -121,13 +121,11 @@ class MockCookieTests {
 			.withMessageContaining("Set-Cookie header must not be null or empty");
 	}
 
-	@Test
-	void parseEmptyHeader() {
+	@ParameterizedTest
+	@ValueSource(strings = {"", "  "})
+	void parseEmptyHeader(String header) {
 		assertThatIllegalArgumentException()
-			.isThrownBy(() -> MockCookie.parse(""))
-			.withMessageContaining("Set-Cookie header must not be null or empty");
-		assertThatIllegalArgumentException()
-			.isThrownBy(() -> MockCookie.parse("  "))
+			.isThrownBy(() -> MockCookie.parse(header))
 			.withMessageContaining("Set-Cookie header must not be null or empty");
 	}
 
--- a/spring-test/src/test/java/org/springframework/mock/web/MockCookieTests.java
+++ b/spring-test/src/test/java/org/springframework/mock/web/MockCookieTests.java
@@ -118,7 +118,17 @@ class MockCookieTests {
 	void parseNullHeader() {
 		assertThatIllegalArgumentException()
 			.isThrownBy(() -> MockCookie.parse(null))
-			.withMessageContaining("Set-Cookie header must not be null");
+			.withMessageContaining("Set-Cookie header must not be null or empty");
+	}
+
+	@Test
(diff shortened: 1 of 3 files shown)


spring-test/src/test/java/org/springframework/mock/web/MockCookieTests.java
```diff
@@ -118,7 +118,17 @@ private void assertCookie(MockCookie cookie, String name, String value) {
...
+	}
+
+	@Test
+	void parseEmptyHeader() {
+		assertThatIllegalArgumentException()
+			.isThrownBy(() -> MockCookie.parse(""))
+			.withMessageContaining("Set-Cookie header must not be null or empty");
+		assertThatIllegalArgumentException()
+			.isThrownBy(() -> MockCookie.parse("  "))
+			.withMessageContaining("Set-Cookie header must not be null or empty");
```
sbrannen:
```suggestion
	@ParameterizedTest
	@ValueSource(strings = {"", "  "})
	void parseEmptyHeader(String header) {
		assertThatIllegalArgumentException()
			.isThrownBy(() -> MockCookie.parse(header))
			.withMessageContaining("Set-Cookie header must not be null or empty");
	}
```

A `@ParameterizedTest` would be better here.

sbrannen (changes requested):
Good catch! 👍 

I've requested a minor change to the test.

Also, please update the duplicate variant of this class: `spring-web/src/testFixtures/java/org/springframework/web/testfixture/servlet/MockCookie.java`

Thanks

sbrannen (approved):

Looks good now.

sbrannen:
This has been merged into `main` in 8b894933aef40b8662f25ae60ea56cb136e22895 and slightly revised in 27a85be4e074d763e157d3393c72f34fa372df44 (since I realized after the merge that we can actually combine the "null" and "empty" tests into a single test method).

Thanks

sbrannen:
Hi @ngocnhan-tran1996,

Due to a code freeze on `main`, I reverted those two commits via 951c1f306a20149c5dc3cc65b5169a08aced51ca and e78d3df566a30786d586702fb91c6f2e79f91eba.

After the 7.1 M1 release, I will cherry pick those two commits back on top of `main` for inclusion in 7.1 M2, retaining you as the author of your commit.

In light of that, can you please create a new issue (not a PR) with the title and body of this PR, referencing the two original commits (8b894933aef40b8662f25ae60ea56cb136e22895 and 27a85be4e074d763e157d3393c72f34fa372df44) to be applied?

Or if you'd prefer that I do that, just let me know.

I apologize for the inconvenience.

Regards,

Sam

ngocnhan-tran1996:
Hi @sbrannen 

I created issue https://github.com/spring-projects/spring-framework/issues/37136. Please let me know if the content needs to be changed.

sbrannen:
> I created issue #37136. Please let me know if the content needs to be changed.

That looks perfect. Thanks, @ngocnhan-tran1996. 👍 

I've now labeled this PR as:

- Superseded by #37136

And I've already assigned that issue to the 7.1 M2 milestone.

Cheers,

Sam

PR #37139, opened 2026-08-14
Reject write methods not starting with "set" in `Property`

junhyeong9812:
## Overview

`org.springframework.core.convert.Property#resolveName()` derives a property
name from a write `Method` when no explicit name is given. It located the
`set` prefix with `String#indexOf`, which matches the token **anywhere** in
the method name.

## Problem

A write method whose name merely contains `set` is silently accepted and
resolves to a meaningless property name derived from whatever follows the
token:

| write method | current result | note |
| --- | --- | --- |
| `setName(String)` | `name` | intended behavior |
| `offsetX(String)` | `x` | not a setter, silently accepted |
| `upset(String)` | `""` (empty) | not a setter, empty name |
| `updateName(String)` | `IllegalArgumentException` | no `set` token at all |

An empty or wrong name then misleads `Property#getField()` and any consumer of
`Property#getName()`. Only names with no `set` token at all were rejected.

## Fix

Match the `set` prefix only at the start of the method name via `startsWith`.
Write methods that are not setters now consistently throw the existing
`IllegalArgumentException("Not a setter method")` instead of silently
producing a wrong name.

This was originally bundled into gh-36911 and has been extracted into this
dedicated PR as requested there, since it is a behavior change: write methods
that embed `set` mid-name were previously accepted and now throw.

## Note on impact

A remaining known gap, unchanged by this PR: a write method whose name starts
with `set` followed by a lowercase letter (for example `settle(String)`)
resolves to the stripped remainder (`tle`) before and after this change, since
prefix matching alone cannot tell it apart from a setter.

New exceptions surface at `Property` construction time for write methods
that are not setters (previously `offsetX`/`upset` shapes constructed
successfully with a wrong name). Framework-internal call sites are largely
unaffected: spring-beans passes an explicit property name, and SpEL searches
for `set`-prefixed methods — with one exotic exception: SpEL's Kotlin support
can surface a `@JvmName`-renamed setter, whose name previously either already
threw (no `set` token anywhere) or resolved to a coincidental name (`set`
mid-name) and now consistently throws. gh-37123 passes verified property
names explicitly at those SpEL call sites, which takes them off this code
path entirely.

## Tests

`PropertyTests` covers the intact `setName` happy path, the previously
rejected `updateName` shape, and the two newly rejected shapes (`offsetX`,
`upset`), asserting the exception message as well.

Note: gh-36911 also adds a `PropertyTests` class on its branch; whichever PR
lands second will be rebased to merge the two test classes.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/convert/Property.java
+++ b/spring-core/src/main/java/org/springframework/core/convert/Property.java
@@ -150,12 +150,11 @@ public final class Property {
 			return StringUtils.uncapitalize(this.readMethod.getName().substring(index));
 		}
 		else if (this.writeMethod != null) {
-			int index = this.writeMethod.getName().indexOf("set");
-			if (index == -1) {
+			String methodName = this.writeMethod.getName();
+			if (!methodName.startsWith("set")) {
 				throw new IllegalArgumentException("Not a setter method");
 			}
-			index += 3;
-			return StringUtils.uncapitalize(this.writeMethod.getName().substring(index));
+			return StringUtils.uncapitalize(methodName.substring(3));
 		}
 		else {
 			throw new IllegalStateException("Property is neither readable nor writable");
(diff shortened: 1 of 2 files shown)


sbrannen (approved):

sbrannen:

Thanks

junhyeong9812:
Thank you for merging this, and for taking care of the overlap with the test class introduced in #36911.

PR #37142, opened 2026-08-15
Cache generic bean type lookups

gregjotau:
Repeated `getBeanNamesForType(ResolvableType)` calls currently rescan every bean definition even after configuration is frozen. This differs from the `Class` overload, which caches its results.

This showed up in an allocation profile of a large Spring Boot application. Its startup made 271 generic type queries for only 51 distinct types; 220 were repeated after configuration had been frozen. In particular, creating seven `HttpSecurity` instances repeated the same top-level `Customizer<T>` lookups.

A backport of this change into the application's Spring Framework 7.0.8 dependency reduced sampled allocation under `doGetBeanNamesForType` from an average of 490.1 MiB to 308.3 MiB (-37.1%). Allocation below `isTypeMatch` and `ResolvableType` fell from 283.3 MiB to 112.8 MiB (-60.2%). Whole-application startup time was too noisy to claim a reliable improvement because Hibernate query parsing and concurrent JIT compilation dominate that measurement.

The included JMH benchmark performs a repeated generic lookup in a frozen bean factory containing 1,001 bean definitions. On JDK 25.0.4, Apple Silicon:

| | Throughput | Allocation |
| --- | ---: | ---: |
| Before | 32,333 ops/s | 192,648 B/op |
| After | 422,750,679 ops/s | approximately 0 B/op |

The new caches mirror the existing raw-type caches. They are used only for frozen configuration with eager initialization allowed, reject keys containing classes that are unsafe for the bean class loader, invalidate matching entries when a singleton appears, and are cleared with the other by-type caches. Tests cover both all-bean and singleton-only caches, invalidation by matching and nonmatching singletons, and recursive generic bounds.

### DIFF
--- a/spring-beans/src/jmh/java/org/springframework/beans/factory/DefaultListableBeanFactoryBenchmark.java
+++ b/spring-beans/src/jmh/java/org/springframework/beans/factory/DefaultListableBeanFactoryBenchmark.java
@@ -30,11 +30,13 @@ import org.springframework.beans.factory.support.DefaultListableBeanFactory;
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.beans.testfixture.beans.LifecycleBean;
 import org.springframework.beans.testfixture.beans.TestBean;
+import org.springframework.core.ResolvableType;
 
 /**
  * Benchmark for retrieving various bean types from the {@link DefaultListableBeanFactory}.
  *
  * @author Brian Clozel
+ * @author Greg Taube
  */
 @BenchmarkMode(Mode.Throughput)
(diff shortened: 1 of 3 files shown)


gregjotau:
I am closing this and deleting my fork so I do not keep a long-lived copy of the Spring codebase.

I do not need ownership or authorship of the change. If the patch is useful, please feel free to re-open it from a Spring team branch, with or without modifications. The only goal on our side is for the generic bean type lookup cache to land so our applications resolve types with less allocation.

The commits remain on this closed PR if that is a convenient source. Happy to answer questions if anything in the patch is unclear.

PR #37149, opened 2026-08-18
Rethrow IOException as ResourceAccessException in RestClient response…

### LINKED ISSUE #37078: RestClient readWithMessageConverters method hiding I/O Exceptions with incorrect error message.
The `org.springframework.web.client.DefaultRestClient#readWithMessageConverters` method from spring-web:7.0.8 is hiding I/O Exceptions (SocketTimeouts from connectTimeout/readTimeout) as a `RestClientException` with an unrelated content-type incompatibility message:

```java
throw new RestClientException("Error while extracting response for type [" + ResolvableType.forType(bodyType) + "] and content type [" + contentType + "]", cause);
```

For example a SocketTimeout exception triggered by a readTimeout set in a client configured with Accept/Content-Type headers for "application/json", throws a `RestClientException` with the message:

> "Error while extracting response for type [MyServiceResponse] and content type [application/octet-stream]"

This message is misleading, causing confusion in the actual underlying issue. Ideally if the issue is I/O related, the `readWithMessageConverters` method should throw a `ResourceAccessException` with the proper message.

yashsiwacha:
… reading

readWithMessageConverters in DefaultRestClient and RestClientUtils was catching IOException (e.g. SocketTimeoutException) together with HttpMessageNotReadableException and wrapping both in a generic RestClientException with the misleading message:

  'Error while extracting response for type [...] and content type [...]'

This caused developers to investigate content-type mismatches when the actual issue was a network I/O failure, making debugging significantly harder.

Fix: split the combined catch so that:
- IOException / UncheckedIOException → ResourceAccessException, which is the semantically correct exception already used elsewhere in RestClient and RestTemplate for low-level I/O failures.
- HttpMessageNotReadableException → RestClientException with the original content-type diagnostic message (unchanged behavior).

Closes gh-37078

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
+++ b/spring-web/src/main/java/org/springframework/web/client/DefaultRestClient.java
@@ -231,16 +231,19 @@ final class DefaultRestClient implements RestClient {
 			return RestClientUtils.readWithMessageConverters(
 					clientResponse, this.messageConverters, bodyType, bodyClass, hints);
 		}
+		catch (RestClientException ex) {
+			throw ex;
+		}
 		catch (UncheckedIOException | IOException exc) {
-			Throwable cause;
+			IOException cause;
 			if (exc instanceof UncheckedIOException uncheckedIOException) {
-				cause = uncheckedIOException.getCause();
+				IOException unwrapped = uncheckedIOException.getCause();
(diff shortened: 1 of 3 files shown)


bclozel:
This was not triaged yet and we already received a PR for this. Thanks!

PR #37150, opened 2026-08-18
Accept @Nullable contextPath in ServerHttpRequest.Builder

### LINKED ISSUE #37098: ServerHttpRequest.Builder.contextPath() does not accept null but the implementation is nullable
While working on spring-cloud/spring-cloud-gateway#4129 I needed to clear the context path
of a request through `mutate()`. My first try was `contextPath(null)` but NullAway fails
the build with:

`passing @Nullable parameter 'null' where @NonNull is required`

The package is `@NullMarked` and the builder method dosen't have `@Nullable` on the
parameter, so null is not allowed. I ended up using `contextPath("")` as a workarround.
In the review it was asked if null could work here instead, discussion is here:
https://github.com/spring-cloud/spring-cloud-gateway/pull/4129#discussion_r3023408590

The confusing part is that everything below the builder method already accepts null:

- `DefaultServerHttpRequestBuilder` stores it in a `@Nullable String contextPath` field
- `MutatedServerHttpRequest` constructor also takes `@Nullable String contextPath`
- `DefaultRequestPath.initContextPath()` handles null and treats it same as empty string

So only the public builder method forbids null. Is this intentional or an oversight from
the JSpecify migration? If its an oversight, can the parameter be marked `@Nullable`?

Checked on 7.0.8 and also on current main, same behaviour in both.

yashsiwacha:
Mark the `contextPath` parameter as `@Nullable` in both the `ServerHttpRequest.Builder` interface and `DefaultServerHttpRequestBuilder` implementation.

Previously, although `DefaultServerHttpRequestBuilder` handled null values internally (e.g., in `MutatedServerHttpRequest` and `DefaultRequestPath.initContextPath()`), the public interface lacked the `@Nullable` annotation. In a `@NullMarked` package, this caused compile-time validation issues (such as under NullAway) for callers trying to pass `null` to reset/clear the context path.

Closes gh-37098

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/server/reactive/DefaultServerHttpRequestBuilder.java
+++ b/spring-web/src/main/java/org/springframework/http/server/reactive/DefaultServerHttpRequestBuilder.java
@@ -103,7 +103,7 @@ class DefaultServerHttpRequestBuilder implements ServerHttpRequest.Builder {
 	}
 
 	@Override
-	public ServerHttpRequest.Builder contextPath(String contextPath) {
+	public ServerHttpRequest.Builder contextPath(@Nullable String contextPath) {
 		this.contextPath = contextPath;
 		return this;
 	}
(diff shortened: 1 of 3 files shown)


sbrannen:
Closing as a duplicate of:

- #37099

In the future, please:

1. First ask the team if a PR is desired by commenting on the related issue.
2. Ensure there are no competing PRs before submitting a duplicate.

Thanks

PR #37159, opened 2026-08-20
Fix busy-spin in SubscriberInputStream await()

flex-donggyu:
`SubscriberInputStream.await()` parks the reader with `LockSupport.park()` after storing the current thread in `parkedThread`. When `park()` returns without `resume()` having set `READY` — spurious wakeup or interrupt — the reference still points to the current thread, so the loop can neither exit nor park again (`compareAndSet(null, ...)` always fails). The thread spins at 100% CPU until the next upstream signal. And since `park()` doesn't consume the interrupt flag, every later `await()` call on that thread spins the same way.

The interrupt case is easy to hit through Spring MVC async handling: both the async timeout and a client disconnect cancel the task with `Future.cancel(true)` (`CallableInterceptorChain#cancelTask`). With `StreamingResponseBody` proxying a slow upstream, the worker is usually parked in `await()` at that moment.

On virtual threads a spinning thread never yields its carrier, so a few cancelled streaming requests can starve the whole scheduler — we traced a production outage (liveness probe kills) to this.

Reproduced on JDK 21 and 24: after the interrupt the thread stays `RUNNABLE`, burning a full core (~13M loop iterations in 200ms).

## Fix

- Park again when the wakeup was spurious and the reference is still ours.
- On interrupt, restore the interrupt status, clear the reference, and propagate the cancellation as an `InterruptedIOException` — unless data arrived concurrently, in which case it is delivered first and the thread stays interrupted for the next call. The `read()` methods rethrow `IOException`s as-is instead of wrapping them, so the interruption surfaces to callers as a regular `IOException`.

Applied to both near-duplicate classes (spring-web and spring-core), with a regression test that fails on the old code.

Related: #35978 fixed a separate defect in `resume()` in this class; the `await()` interrupt path wasn't covered by it.

### DIFF
--- a/spring-core/src/main/java/org/springframework/core/io/buffer/SubscriberInputStream.java
+++ b/spring-core/src/main/java/org/springframework/core/io/buffer/SubscriberInputStream.java
@@ -18,6 +18,7 @@ package org.springframework.core.io.buffer;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InterruptedIOException;
 import java.util.ConcurrentModificationException;
 import java.util.Objects;
 import java.util.Queue;
@@ -228,6 +229,9 @@ final class SubscriberInputStream extends InputStream implements Subscriber<Data
 			this.closed = true;
 			requiredSubscriber().cancel();
 			cleanAndFinalize();
+			if (ex instanceof IOException ioEx) {
+				throw ioEx;
+			}
 			throw Exceptions.propagate(ex);
 		}
 		finally {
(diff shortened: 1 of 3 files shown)


skrcode:
JAIPilot Cloud reviewed exact head 7e537998b603952eb2505564af1cea5e6daef74f and found one focused gap: the interrupt fix changes both SubscriberInputStream implementations, while the PR regression test covers only spring-web.

I opened a test-only draft against the contributor branch: https://github.com/flex-donggyu/spring-framework/pull/1

The identical focused baseline/candidate command passed for DataBufferUtilsTests and SubscriberInputStreamTests, and spring-core:check plus spring-web:check passed in the managed sandbox. No production refactor is included in the offer. AI-assisted; please review and squash/cherry-pick with the required DCO sign-off if useful.

flex-donggyu:
Added the same regression test for the spring-core copy (through `DataBufferUtils.subscriberInputStream`), since the fix touches both classes. Squashed into the original commit.

PR #37192, opened 2026-08-22
Handle MIME type parameter names case-insensitively

lArtiquel:
`MimeType` keeps parameters in a `LinkedCaseInsensitiveMap`, so `equals()` matches names case-insensitively. `parametersHashCode()` hashes the raw key, so it doesn't — which breaks the `equals`/`hashCode` contract:

```java
MimeType a = MimeType.valueOf("text/plain;FOO=bar");
MimeType b = MimeType.valueOf("text/plain;foo=bar");

a.equals(b);                            // true
a.hashCode() == b.hashCode();           // false
new HashSet<>(List.of(a)).contains(b);  // false
```

The charset branch in `equals`, `hashCode` and `compareTo` keys off `PARAM_CHARSET.equals(key)`, so a parameter spelled `Charset` skips it and is compared as an opaque string, even though the constructor already resolved it through the same case-insensitive map:

```java
MimeType.valueOf("text/plain;charset=UTF-8").equals(MimeType.valueOf("text/plain;charset=utf-8")); // true
MimeType.valueOf("text/plain;Charset=UTF-8").equals(MimeType.valueOf("text/plain;Charset=utf-8")); // false
```

`MediaType` has the same assumption in two more places: an out-of-range quality value escapes validation when spelled `Q=1.1`, and `removeQualityValue()` leaves a `Q=` parameter in place.

Fix is six lines — `equalsIgnoreCase` in the four comparisons, and a normalized key before hashing.

Two behavior changes, both bringing the uppercase spelling in line with the lowercase one: `Charset=UTF-8` now equals `Charset=utf-8`, and `Q=1.1` is now rejected at parse time. Four tests added, each failing before the change; `spring-core`, `-web`, `-webmvc`, `-webflux`, `-messaging` and `-test` suites pass.

### DIFF
--- a/spring-core/src/main/java/org/springframework/util/MimeType.java
+++ b/spring-core/src/main/java/org/springframework/util/MimeType.java
@@ -486,9 +486,9 @@ public class MimeType implements Comparable<MimeType>, Serializable {
 	/**
 	 * Determine if the parameters in this {@code MimeType} and the supplied
 	 * {@code MimeType} are equal, performing case-insensitive comparisons
-	 * for {@link Charset Charsets} and disregarding quoting of parameter
-	 * values, so that, for example, {@code spring="framework"} and
-	 * {@code spring=framework} are considered equal.
+	 * for parameter names and {@link Charset Charsets}, and disregarding
+	 * quoting of parameter values, so that, for example, {@code spring="framework"}
+	 * and {@code spring=framework} are considered equal.
 	 * @since 4.2
 	 */
 	private boolean parametersAreEqual(MimeType other) {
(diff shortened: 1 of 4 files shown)


bclozel:
Duplicates #37008?

lArtiquel:
No. #37008 rejects duplicate params within one header string; this is equals/hashCode across two instances. I applied it on main and the case is unchanged:

```java
MimeType a = MimeType.valueOf("text/plain;FOO=bar");
MimeType b = MimeType.valueOf("text/plain;foo=bar");
a.equals(b);                   // true
a.hashCode() == b.hashCode();  // false
```

`parametersAreEqual` resolves the key through the `LinkedCaseInsensitiveMap`, `parametersHashCode` hashes it verbatim.

They only overlap in `MimeTypeTests`, so one needs a rebase. Can split the `MediaType` part out if you'd rather.

PR #37193, opened 2026-08-22
Ignore invalid SSE retry field

lArtiquel:
The [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) says a `retry` field must be ignored unless its value is only ASCII digits. `ServerSentEventHttpMessageReader` passes it straight to `Long.parseLong`, so one malformed field kills the whole stream:

```
retry:none
data:foo

```
```
NumberFormatException: For input string: "none"
```

Same for an empty `retry:` and for a value too large for a `long`. A client can't control what the server sends, so an unusable reconnection hint takes down an otherwise healthy subscription — and `retry` is a hint the reader never acts on itself.

This ignores the field in those cases. Test added; it fails before the change. `spring-web` and `spring-webflux` suites pass.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/codec/ServerSentEventHttpMessageReader.java
+++ b/spring-web/src/main/java/org/springframework/http/codec/ServerSentEventHttpMessageReader.java
@@ -163,7 +163,10 @@ public class ServerSentEventHttpMessageReader implements HttpMessageReader<Objec
 					sseBuilder.event(line.substring(6).trim());
 				}
 				else if (line.startsWith("retry:")) {
-					sseBuilder.retry(Duration.ofMillis(Long.parseLong(line.substring(6).trim())));
+					Long retry = parseRetry(line.substring(6).trim());
+					if (retry != null) {
+						sseBuilder.retry(Duration.ofMillis(retry));
+					}
 				}
 				else if (line.startsWith(":")) {
 					comment = (comment != null ? comment : new StringBuilder());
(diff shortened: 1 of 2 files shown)


spring-web/src/main/java/org/springframework/http/codec/ServerSentEventHttpMessageReader.java
```diff
@@ -188,6 +191,32 @@ else if (line.startsWith(":")) {
...
+	 * consists solely of ASCII digits and fits into a {@code long}.
+	 * @return the reconnection time in milliseconds, or {@code null} to ignore the field
+	 * @see <a href="https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation">
+	 * HTML Living Standard: interpreting an event stream</a>
+	 */
+	private static @Nullable Long parseRetry(String value) {
+		if (value.isEmpty()) {
+			return null;
+		}
+		for (int i = 0; i < value.length(); i++) {
```
bclozel, line 207: Why check for ascii characters when Long parseLong performs stricter checks right after?

bclozel (changes requested):

lArtiquel:
`Long.parseLong` is actually more lenient than the spec here — it accepts a leading sign and non-ASCII digits:

```java
Long.parseLong("-1")     // -1     -> Duration.ofMillis(-1)
Long.parseLong("+5000")  // 5000
Long.parseLong("٥٠٠٠")   // 5000   (Arabic-Indic)
```

None of those "consist of only ASCII digits", so the spec says ignore the field. The try/catch is still there for the overflow case, which `parseLong` does reject.

I've added a comment saying so, plus test cases for `+5000` and the Arabic-Indic form. Dropping the loop makes `ignoreInvalidRetry` fail on `retry:-1`.

bclozel:
Thanks @lArtiquel !

PR #37196, opened 2026-08-24
Upgrade actions/setup-java to v6

brunoborges:
Upgrade all active workflow references to `actions/setup-java` v6. SHA-pinned references now use the v6.0.0 release SHA, while the semantic reference uses `v6`.

Tests were not run (workflow-only change).


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #37197, opened 2026-08-25
fix(web): fallback to normalized URL path if URI template is unavailable

zanarellidev:
Fixes #36547 by falling back to the URL path for the uri observation tag when the URI template is missing, preventing unbounded cardinality when URLs are passed without templates.


bclozel:
Isn't this change creating a DoS because of a metric cardinality explosion? The new test contributed here seems to indicate that. For any path like `"/resource/42"` this will create a metric dimension, overwhelming the application memory and the observability backend.

Please advise.

zanarellidev:
@bclozel You're absolutely right, thank you for pointing that out.

The intention was to give path visibility when requests don't provide a template, but because the framework cannot safely know whether a raw path contains dynamic variables (like `/resource/42`, IDs, or UUIDs) without a template, placing the unparameterized path in `LowCardinalityKeyNames.URI` exposes metric backends to unbounded cardinality explosion.

Since the full URL is already captured in the high-cardinality key-values (`http.url`), keeping `uri: none` for low-cardinality metrics when no URI template is available is indeed the safer and intended design.

I will close this PR. Thanks again for the review and explanation!

PR #37202, opened 2026-08-27
Emit multipart parts with empty bodies in PartGenerator

### LINKED ISSUE #37200: Multipart HttpMessageConverter drops parts with empty bodies
## Description
The new servlet `MultipartHttpMessageConverter` (spring-web, 7.1) silently drops every part whose body is empty, so a blank form field is indistinguishable from an absent field.

## Analysis
Parts are emitted exclusively from `PartListener.onBody(buffer, last=true)` (`PartGenerator`: FormFieldState, InMemoryState, FileState). For an empty part, `MultipartParser.BodyState` computes `len == 0` and transitions to the next headers state without ever invoking `onBody`, so the part is never added to the resulting `MultiValueMap`.

This is the servlet-side equivalent of the issue fixed for the reactive stack in #30953 (regression test: `DefaultPartHttpMessageReaderTests.emptyLastPart` asserts empty parts *are* emitted). The new servlet port did not carry the fix over; `MultipartParserTests.noBody()` currently pins the broken event sequence.

## Impact
An HTML form with a blank text input yields "field absent" instead of "field = empty string", which can flip application logic that distinguishes the two (including auth/business-rule branches).

## Reproduction
`GET`-style body:
```
--boundary
Content-Disposition: form-data; name="text1"

<empty>
--boundary
Content-Disposition: form-data; name="text2"

a
--boundary--
```
Reading via `MultipartHttpMessageConverter` returns only `text2`; `text1` is missing.

shaggyinsomniac:
Fixes #37200

## Description
Parts were emitted only from `PartListener.onBody(buffer, last=true)`, so a part with an empty body (for example a blank form field, or a trailing empty part) was silently dropped from the resulting `MultiValueMap`, and was indistinguishable from an absent field.

This carries over the fix from the reactive `DefaultPartHttpMessageReader` (#30953): `State` gains an `onComplete()` callback that emits the part — also when it has an empty body. It is invoked when a new part begins (previous part completes without a final body chunk) and when parsing completes (final part).

Emission is guarded per state so a part completed via `onBody(last=true)` is not emitted twice.

## Test coverage
- `readMultipartEmptyPart()` — empty form field followed by a non-empty one
- `readMultipartEmptyLastPart()` — non-empty field followed by a trailing empty part
- Both fail on `main` (part missing from the result) and pass with this change; the full `:spring-web:test` suite (3,919 tests) passes

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/converter/multipart/PartGenerator.java
+++ b/spring-web/src/main/java/org/springframework/http/converter/multipart/PartGenerator.java
@@ -86,6 +86,7 @@ final class PartGenerator implements MultipartParser.PartListener {
 
 	@Override
 	public void onHeaders(HttpHeaders headers) {
+		this.state.onComplete();
 		if (isFormField(headers)) {
 			this.state = new FormFieldState(headers);
 		}
@@ -127,6 +128,7 @@ final class PartGenerator implements MultipartParser.PartListener {
 
 	@Override
 	public void onComplete() {
+		this.state.onComplete();
 		if (logger.isTraceEnabled()) {
 			logger.trace("Finished reading " + this.partCount + " part(s)");
 		}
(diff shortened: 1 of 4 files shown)


bclozel:
Please [sign the DCO](https://github.com/spring-projects/spring-framework/pull/37202/checks?check_run_id=98520670906)

bclozel:
Thanks @shaggyinsomniac !

PR #37203, opened 2026-08-27
Fix out-of-bounds read for truncated percent-escape in opaque host

### LINKED ISSUE #37201: WhatWgUrlParser throws StringIndexOutOfBoundsException for truncated percent-escape in opaque host
## Description
`WhatWgUrlParser.parse("foo://%4", ...)` throws an unchecked `StringIndexOutOfBoundsException` when URL validation is enabled.

## Analysis
In `OpaqueHost.parse` the percent-escape validation guard reads `input.codePointAt(i + 2)` after only checking `input.length() - i < 2` — two code points must follow `%`, so the guard must require `< 3`. Additionally it checks `isAsciiDigit` where the URL spec (and the adjacent code comment) require ASCII *hex* digits.

Per the [URL spec opaque-host parser](https://url.spec.whatwg.org/#concept-opaque-host-parser), an invalid percent-escape is a (non-fatal) validation error: the host `%4` is accepted.

## Impact
Any caller that catches `InvalidUrlException` to return a 400 instead sees a raw runtime exception (500 / error-handling bypass).

## Reproduction
```java
WhatWgUrlParser.parse("foo://%4", UrlRecord.EMPTY, null, error -> {});  // StringIndexOutOfBoundsException
```

shaggyinsomniac:
Fixes #37201

## Description
The opaque-host percent-escape validation guard reads `input.codePointAt(i + 2)` after only checking `input.length() - i < 2`, so an input such as `foo://%4` throws `StringIndexOutOfBoundsException` instead of reporting a validation error.

The guard now requires two code points after `%` and checks ASCII hex digits rather than ASCII digits, matching the [URL spec](https://url.spec.whatwg.org/#concept-opaque-host-parser) and the existing code comment: invalid percent-escapes in opaque hosts are validation errors, not failures — the host `%4` is accepted.

## Test coverage
- `parseOpaqueHostTruncatedPercentEscape()` — parses `foo://%4` / `foo://%4x` with validation enabled, asserts the resulting host and that validation errors were reported
- Fails on `main` with `StringIndexOutOfBoundsException` and passes with this change; the full `:spring-web:test` suite (3,919 tests) passes

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/util/WhatWgUrlParser.java
+++ b/spring-web/src/main/java/org/springframework/web/util/WhatWgUrlParser.java
@@ -2241,8 +2241,8 @@ final class WhatWgUrlParser {
 				// If input contains a U+0025 (%) and the two code points following it
 				// are not ASCII hex digits, invalid-URL-unit validation error.
 				if (p.validate() && ch == '%' &&
-						(input.length() - i < 2 || !isAsciiDigit(input.codePointAt(i + 1)) ||
-								!isAsciiDigit(input.codePointAt(i + 2)))) {
+						(input.length() - i < 3 || !isAsciiHexDigit(input.codePointAt(i + 1)) ||
+								!isAsciiHexDigit(input.codePointAt(i + 2)))) {
 					p.validationError("Code point \"" + ch + "\" is not a URL unit.");
 				}
 			}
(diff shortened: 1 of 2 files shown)


bclozel:
Please [sign the DCO](https://github.com/spring-projects/spring-framework/pull/37203/checks?check_run_id=98520725188).

Issue #25707, opened 2020-09-04
CallMetaDataContext.reconcileParameters overwrites cursor name


javadoug4:
**Affects:** \SB2
org.springframework:spring-jdbc:5.1.9.RELEASE
Explanation. Existing code is using a named cursor.
When the Resultset comes back, it is always empty, even though when I debug through the spring-jdbc I see it getting created with correct results. The issue is with the named cursor.

Example:

> this.simpleJdbcCall = new SimpleJdbcCall(jdbcTemplate)
>                 .withSchemaName("schema")
>                 .withFunctionName("functionName")
>                 .returningResultSet("myCursorName", mapperInstance);

if I name the cursor 'returnValue' then it works. Searching through the code, I saw that it is over writing 'myCursorName' with 'returnValue'. 'returnValue' seems to be the default name used internally.
In SimpleJdbcCall:

> 	public <T> T executeFunction(Class<T> returnType, SqlParameterSource args) {
> 		return (T) doExecute(args).get(getScalarOutParameterName());
> 	}

'getScalarOutParameterName()' is returning 'returnValue' even though I named it as '.returningResultSet("myCursorName",'

But I also traced this to `CallMetaDataContext.reconcileParameters` around line 396:

> 	param = declaredParams.get(getFunctionReturnName());
> 	if (param == null && !getOutParameterNames().isEmpty()) {
> 		param = declaredParams.get(getOutParameterNames().get(0).toLowerCase());
> 	}
> 	if (param == null) {
> 		throw new InvalidDataAccessApiUsageException(
> 				"Unable to locate declared parameter for function return value - " +
> 				" add an SqlOutParameter with name '" + getFunctionReturnName() + "'");
> 	}
> 	else if (paramName != null) {
> 		setFunctionReturnName(paramName);
> 	}

In older version it used to have (the last 3 lines):

> 	else {
> 		setFunctionReturnName(param.getName());
> 	}

paramName is 'returnValue', but it should be using what I supplied which is 'param.getName()'.
Then this would return the correct ruleset: return (T) doExecute(args).get(getScalarOutParameterName());

Please advise. Thanks.

jhoeller:
This seems to be an accidental regression introduced by our nullability efforts in 5.0. I'll fix this for 5.2.9, with a backport to 5.1.18 and 5.0.19 at the same time (for release on Sep 15).

javadoug4:
Thanks for the quick response.

PR #37206, opened 2026-08-27
Normalize function return parameter lookup in `CallMetaDataContext`

junhyeong9812:
## Overview

`CallMetaDataContext.reconcileParameters` matches the return parameter reported by the database metadata against the declared parameters with lookup keys that do not follow the map's own key rule. When a function has an additional out parameter declared before the return parameter, the return slot is silently bound to the wrong parameter and `executeFunction` returns the wrong value; on SQL Server the same declaration order with `withReturnValue()` fails with `InvalidDataAccessApiUsageException`. This PR normalizes the lookups with the same rule used to build the map.

## Problem

The declared parameter map is keyed by `lowerCase(provider.parameterNameToUse(name))`, so a declared `RESULT` is stored under `result` and a SQL Server `@out_total` under `out_total`. The return parameter branch, however, looked up `getFunctionReturnName()` as declared (original case, so it never matched on Oracle) and then fell back to `getOutParameterNames().get(0).toLowerCase()`, i.e. the first declared out parameter without the provider transformation.

Because the first lookup always missed, the fallback decided the result, and it only happens to be right when the return parameter is the first declared out parameter:

| Declaration (Oracle function `GET_TOTAL`) | Before | After |
|---|---|---|
| `SqlOutParameter("RESULT")`, `SqlOutParameter("out_status")` | `[RESULT, AMOUNT, out_status]` | unchanged |
| `SqlOutParameter("out_status")`, `SqlOutParameter("RESULT")` | `[out_status, AMOUNT, out_status]`, `executeFunction` returns the `out_status` value | `[RESULT, AMOUNT, out_status]` |

On SQL Server, `withProcedureName(...).withReturnValue()` with `SqlOutParameter("@out_total")` declared before `SqlOutParameter("RETURN_VALUE")` threw `Unable to locate declared parameter for function return value - add an SqlOutParameter with name 'return'`, since `"@out_total".toLowerCase()` does not match the stored key `out_total`.

The first-out-parameter fallback itself dates back to the 4.x behavior restored in #25707 and is kept as is; only the lookup keys change.

## Fix

The return parameter branch now looks up the metadata-derived name (`paramNameToCheck`) first, then the function return name, then the first declared out parameter, with all three keys normalized by `lowerCase(provider.parameterNameToUse(...))` exactly like the map keys. Parameter names themselves are not modified, so existing call parameter names and call strings are unchanged. Applications that declared an additional out parameter before the return parameter of a function now receive the actual return value under the declared return parameter name instead of the value of that other out parameter. As before, this lookup only decides which declared parameter represents the return slot; whether a return slot is bound at all is still governed by `withFunctionName()` and `withReturnValue()`.

`SimpleJdbcCallTests` gains four tests: an Oracle function and a SQL Server procedure with return value, each with the additional out parameter declared before and after the return parameter. The two "before" cases failed prior to this change; the two "after" cases guard the previously working declaration order. The full `spring-jdbc` test suite passes.

### DIFF
--- a/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/CallMetaDataContext.java
+++ b/spring-jdbc/src/main/java/org/springframework/jdbc/core/metadata/CallMetaDataContext.java
@@ -374,9 +374,14 @@ public class CallMetaDataContext {
 			if (declaredParams.containsKey(paramNameToCheck) || (meta.isReturnParameter() && returnDeclared)) {
 				SqlParameter param;
 				if (meta.isReturnParameter()) {
-					param = declaredParams.get(getFunctionReturnName());
+					// Same normalization as the declaredParams keys above; the function
+					// return name may have been adopted from a declared out parameter
+					param = declaredParams.get(paramNameToCheck);
+					if (param == null) {
+						param = declaredParams.get(lowerCase(provider.parameterNameToUse(getFunctionReturnName())));
+					}
 					if (param == null && !getOutParameterNames().isEmpty()) {
-						param = declaredParams.get(getOutParameterNames().get(0).toLowerCase(Locale.ROOT));
+						param = declaredParams.get(lowerCase(provider.parameterNameToUse(getOutParameterNames().get(0))));
 					}
 					if (param == null) {
 						throw new InvalidDataAccessApiUsageException(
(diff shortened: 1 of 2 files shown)


sbrannen:

Thanks 👍

junhyeong9812:
Thanks for reviewing and merging this, and for applying it to 7.0.x as well!

Issue #25613, opened 2020-08-19
Potential integer overflow in ExecutorConfigurationSupport.setAwaitTerminationSeconds(int)


TomaszPierzchala:
**Affects:**  5.2.8.RELEASE
There is **a small** bug in the class  `org.springframework.scheduling.concurrent.ExecutorConfigurationSupport`
in the method :
```java 
public void setAwaitTerminationSeconds(int awaitTerminationSeconds) {
		this.awaitTerminationMillis = awaitTerminationSeconds * 1000;
}
```

When the input parameter `awaitTerminationSeconds` is bigger then `Integer.MAX_VALUE/1000`
the value asigned to `long this.awaitTerminationMillis` can even be **negative** !

I recommend changing the `1000` to the `1000l` long version as below
```java 
public void setAwaitTerminationSeconds(int awaitTerminationSeconds) {
		this.awaitTerminationMillis = awaitTerminationSeconds * 1000l;
}
```

jhoeller:
Thanks for spotting this! It may be rare to specify such a large value there but it still needs to work correctly.

This also applies to `AbstractResourceBasedMessageSource.setCacheSeconds` where I'll fix it the same way.

PR #37208, opened 2026-08-28
Avoid int overflow in expiration time calculations in MockMvc and FlashMap

a-kogun:
Two places still compute an expiration time by multiplying an `int`
number of seconds by `1000` without widening to `long`:

- `FlashMap.startExpirationPeriod(int)` (spring-webmvc)
- `MockMvcWebConnection.createCookie(...)` (spring-test)

Above `2_147_483` seconds (~24.9 days) the multiplication overflows and
yields a negative offset, so the expiration time lands in the **past**.
With `Integer.MAX_VALUE`, `timeToLive * 1000` evaluates to `-1000`, i.e.
one second ago.

**Impact**

- `FlashMap`: a timeout set via the public
  `AbstractFlashMapManager.setFlashMapTimeout(int)` is treated as
  already expired, so flash attributes are silently dropped.
- `MockMvcWebConnection`: because `storeCookies` compares the computed
  expiry against now, a cookie with a large `max-age` is passed to
  `cookieManager.removeCookie(...)` instead of `addCookie(...)` — the
  cookie is discarded rather than stored.

**Fix**

Widen the literal to `1000L` in both places. This is the same change
already applied to this exact pattern in gh-25613
(`ExecutorConfigurationSupport.setAwaitTerminationSeconds(int)` and
`AbstractResourceBasedMessageSource.setCacheSeconds(int)`), and matches
how those methods read today.

**Tests**

Two regression tests are included, both of which fail before the change:

- `FlashMapTests.notExpiredWithLargeTimeToLive()` — fails with
  `Expecting actual: <now-1000> to be greater than: <now>`.
- `MockMvcWebClientBuilderTests.cookieWithLargeMaxAgeIsStored()` — fails
  with `expected: "foo" but was: "NA"`, showing the cookie was dropped.

A scan of `src/main/java` found no other live instances of this pattern:
`AbstractSockJsService.ONE_YEAR` and `DurationFormatterUtils.micros` are
already declared `long`.

### DIFF
--- a/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockMvcWebConnection.java
+++ b/spring-test/src/main/java/org/springframework/test/web/servlet/htmlunit/MockMvcWebConnection.java
@@ -183,7 +183,7 @@ public final class MockMvcWebConnection implements WebConnection {
 	private static Cookie createCookie(jakarta.servlet.http.Cookie cookie) {
 		Date expires = null;
 		if (cookie.getMaxAge() > -1) {
-			expires = new Date(System.currentTimeMillis() + cookie.getMaxAge() * 1000);
+			expires = new Date(System.currentTimeMillis() + cookie.getMaxAge() * 1000L);
 		}
 		BasicClientCookie result = new BasicClientCookie(cookie.getName(), cookie.getValue());
 		result.setDomain(cookie.getDomain());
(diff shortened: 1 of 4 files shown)


sbrannen:
Hi @a-kogun,

Good catch, and congratulations on submitting your first PR for the Spring Framework! 👍 

Thanks

PR #37214, opened 2026-08-30
Prefer reason messages over response status fallback

### LINKED ISSUE #36984: Custom reason is overwritten by i18n message in ResponseStatusException.updateAndGetBody()
In cases where `reason` on `ResponseStatusException` is not manually set, I expect localization key equal to `problemDetail.org.springframework.web.server.ResponseStatusException` to act as a fallback; but since super.updateAndGetBody() is called before instead of after custom logic, this localization acts as an overwrite instead.

blackdurumi:
Fixes #36984

## Problem

When a `ResponseStatusException` has a reason that can be resolved through a `MessageSource`, that localized reason should be used as the `ProblemDetail` detail.

This can be surprising when the message source also contains the default detail message for `ResponseStatusException`. The general `ErrorResponse` localization step may replace the original reason first, so the exception no longer recognizes that it should resolve the reason message. In that case, the default detail is returned even though a localized reason is available.

## What this change does

- Keeps the reason as the initial `ProblemDetail` detail for `ResponseStatusException`.
- Resolves the reason after the common `ErrorResponse` localization has completed when the default detail message code is being used.
- Preserves an explicitly configured detail message code as the source of truth.
- Preserves details supplied by specialized `ResponseStatusException` subclasses, such as `MethodNotAllowedException`.
- Keeps repeated updates and locale changes consistent.

In short, a localized reason is preferred when the exception is using the default response-status detail, while explicit or specialized details remain authoritative.

## Tests

The regression coverage verifies:

- localized reason messages take precedence over the default detail;
- repeated updates do not lose the reason localization;
- changing the locale re-resolves the reason;
- the default detail remains the fallback when the reason cannot be resolved;
- custom detail message codes are not overridden by the reason;
- subclass-specific details remain unchanged.

## Verification

- `JAVA_HOME=/opt/homebrew/opt/openjdk@25 ./gradlew :spring-web:test --tests org.springframework.web.ErrorResponseExceptionTests`
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25 ./gradlew :spring-webflux:test --tests org.springframework.web.reactive.result.method.annotation.ResponseEntityExceptionHandlerTests`
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25 ./gradlew :spring-web:check`
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25 ./gradlew build`

### DIFF
--- a/spring-web/src/main/java/org/springframework/web/server/ResponseStatusException.java
+++ b/spring-web/src/main/java/org/springframework/web/server/ResponseStatusException.java
@@ -24,6 +24,7 @@ import org.springframework.context.MessageSource;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatusCode;
 import org.springframework.http.ProblemDetail;
+import org.springframework.web.ErrorResponse;
 import org.springframework.web.ErrorResponseException;
 
 /**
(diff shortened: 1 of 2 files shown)


blackdurumi:
I submitted this PR before confirming whether a contribution would be welcome. Based on the maintainer guidance, I’m closing it for now and will wait for the issue to be triaged. I’d be happy to revisit it if a PR is requested.

PR #36944, opened 2026-06-17
Add batch update support to JdbcClient

Empatixx:
This adds batch update support to `JdbcClient`, as requested in gh-33843.

Batch entries are accumulated in a fluent way through a new `batch()` method on `StatementSpec`:

```java
int[] rowsAffected = client.sql("INSERT INTO user (first_name, last_name) VALUES (:first, :last)")
        .batch()
            .param("first", "Jane").param("last", "Smith").add()
            .param("first", "John").param("last", "Doe")
        .batchUpdate();
```

Each entry is bound with `param(...)` / `params(...)` / `paramSource(...)` exactly like a single update; `add()` separates entries and `batchUpdate()` executes them as one JDBC batch (the last entry is completed implicitly). Both positional and named parameters are supported (mixing the two in one batch is rejected). It delegates to `JdbcOperations.batchUpdate` / `NamedParameterJdbcOperations.batchUpdate`.

Reference docs and integration tests are included. The `spring-jdbc` build, Checkstyle and tests pass.

I know this is assigned to @sdeleuze — feel free to take whatever is useful here.

Issue #33843, opened 2024-11-05
Add batch operations support to `JdbcClient`


sdeleuze:
Similar to #33812 but for `JdbcClient`.

matt-locker:
@sdeleuze 
For replacing JdbcTemplate, this is the last remaining bit :)
Any chance this will land also on 6.x?

sdeleuze:
Likely not, too involved, but let's confirm when implemented in 7.x.

fmbenhassine:
This would be beneficial for Spring Batch as we intend to replace `JdbcTemplate` usage with `JdbcClient`, see https://github.com/spring-projects/spring-batch/pull/4897

Great to see that this is planned for SF 7.1.x, it would be great to have it in Batch 6.1 which will be based on SF 7.1.

jhoeller:
Superseded by #37216.

PR #37216, opened 2026-08-31
Add batch update support to JdbcClient

quaff:
See #36944

### DIFF
--- a/framework-docs/modules/ROOT/pages/data-access/jdbc/core.adoc
+++ b/framework-docs/modules/ROOT/pages/data-access/jdbc/core.adoc
@@ -680,6 +680,21 @@ provides `firstName` and `lastName` properties, such as the `Actor` class from a
 			.update();
 ----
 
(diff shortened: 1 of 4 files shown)


jhoeller:
This looks good for the purposes of #33843, thanks for all the efforts! We're considering inclusion in 7.1 M2.

In terms of API naming, pretty much my only concern is the repetition of the "batch" term in a `batch()...batchUpdate()` sequence. I suppose we could simply name the latter `update()`, even if it returns an `int[]` since it naturally executes a batch at the end of a fluent batch call sequence?

quaff:
> This looks good for the purposes of #33843, thanks for all the efforts! We're considering inclusion in 7.1 M2.
> 
> In terms of API naming, pretty much my only concern is the repetition of the "batch" term in a `batch()...batchUpdate()` sequence. I suppose we could simply name the latter `update()`, even if it returns an `int[]` since it naturally executes a batch at the end of a fluent batch call sequence?

PR updated for renaming, please review @jhoeller

PR #37220, opened 2026-08-31
Support CompletableFuture<Optional<?>> cache values

### LINKED ISSUE #37162: Allow @Cacheable method to return CompletableFuture<java.util.Optional> variant of cached value
```java
class MyService {

    @Cacheable("myCache")
    public CompletableFuture<Optional<String>> doSomethingExpensive(String id) {
        return CompletableFuture.completedFuture(Optional.ofNullable(id).map(i -> "PREFIX-" + i));
    }

}
```
In the above example, a ClassCastException is thrown when `doSomethingExpensive` is called a second time to retrieve the cached value but the cache actually returns `CompletableFuture<String>` instead of expected `CompletableFuture<Optional<String>>`.

Both CompletableFuture and Optional are currently supported separately, it would be nice to have support for their combination.

spring version: 7.0.8

blackdurumi:
Fixes #37162

## What does this PR do?

This change adds support for `@Cacheable` methods that return `CompletableFuture<Optional<T>>`.

The cache now stores the value inside the `Optional` (or the cache-compatible null representation for `Optional.empty()`), while callers continue to receive the declared `CompletableFuture<Optional<T>>` result on both cache misses and cache hits. The same adaptation is applied to the existing `sync = true` path.

The cache-put completion stage remains part of the returned future, so cache put failures and completion ordering retain the existing `CompletableFuture` behavior.

## Why is this change needed?

Before this change, a cached result could be returned with the wrong inner type, causing a `ClassCastException` when a `CompletableFuture<Optional<T>>` method was called again.

## How is this tested?

- Added regression coverage for present and empty `Optional` values.
- Covered regular asynchronous caching and `sync = true` caching.
- Asserted that the second call is served from the cache.
- Ran `JAVA_HOME=/opt/homebrew/opt/openjdk@25 ./gradlew :spring-context:check`.

### DIFF
--- a/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
+++ b/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java
@@ -55,6 +55,7 @@ import org.springframework.core.BridgeMethodResolver;
 import org.springframework.core.KotlinDetector;
 import org.springframework.core.ReactiveAdapter;
 import org.springframework.core.ReactiveAdapterRegistry;
+import org.springframework.core.ResolvableType;
 import org.springframework.core.SpringProperties;
 import org.springframework.expression.EvaluationContext;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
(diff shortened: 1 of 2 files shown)


bclozel:
Original issue has been declined.

sbrannen:
As a side note, please refrain from submitting PRs for issues that are still labeled as "waiting for triage". You should instead first participate in the discussion on the corresponding issue and inquire whether a PR is welcome.

PR #37224, opened 2026-09-01
Parse RFC 9651-like date headers

sabberworm:
RFC 9651 specifies @«timestamp» as a new format for date headers. The Deprecation header as specified in RFC 9745, for example, makes use of it. Make sure this can be parsed using `HttpHeaders#getFirstDate` and `HttpHeaders#getFirstZonedDateTime`.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
+++ b/spring-web/src/main/java/org/springframework/http/HttpHeaders.java
@@ -29,7 +29,9 @@ import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
+import java.time.temporal.ChronoField;
 import java.util.AbstractSet;
 import java.util.ArrayList;
 import java.util.Base64;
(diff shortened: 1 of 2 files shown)


bclozel:
RFC 9651 ships way more than that. Is there a particular need to support this feature in a standalone fashion?
Are you encountering a particular case in production that requires such support? Can you share which client or server is involved ?

sabberworm:
> RFC 9651 ships way more than that

@bclozel true, but `HttpHeaders` currently has no convenience methods for data types specified in RFC 9651 other than dates. Whether it makes sense to add these is a separate discussion. The case for extending the existing methods for dates to parse all possible date formats including RFC 9651 is much stronger, IMHO.

My use-case is that I want to alert if the API we consume is deprecated. I want to increase the alert severity depending on how far in the future the deprecation is scheduled for, which is why I need to parse the date in the `Deprecated` header.

Some of the API I consume uses the draft version of the `Deprecated` header, which uses a date format currently supported by `HttpHeaders`. Ideally I would like to parse both using the same code path, which is why I think this belongs in the framework.

bclozel:
Thanks @sabberworm - with this change, we're not committing to RFC 9651 (because there is a lot to unpack there), but the use case makes sense.

PR #37219, opened 2026-08-31
Polish ScheduledAnnotationBeanPostProcessor to refine null-safety

quaff:

PR #37225, opened 2026-09-01
Revise null-safety within `DateTimeFormatterRegistrar`

DevJunz:
## Overview

Narrow the scope of the `@SuppressWarnings("NullAway")` annotation in `DateTimeFormatterRegistrar` from the class level to a single `getFactory(Type)` accessor.

The `factories` map is a private, final `EnumMap` that is fully populated for every `Type` in the constructor and never mutated afterward, so the suppression only needs to cover that one lookup instead of masking unrelated issues across the whole class.

## Related Issues

- #37219
- #36921

### DIFF
--- a/spring-context/src/main/java/org/springframework/format/datetime/standard/DateTimeFormatterRegistrar.java
+++ b/spring-context/src/main/java/org/springframework/format/datetime/standard/DateTimeFormatterRegistrar.java
@@ -43,6 +43,7 @@ import org.springframework.format.annotation.DateTimeFormat.ISO;
  *
  * @author Juergen Hoeller
  * @author Phillip Webb
+ * @author Junhwan Choi
  * @since 4.0
  * @see #setDateStyle
  * @see #setTimeStyle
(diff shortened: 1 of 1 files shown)


sbrannen (changes requested):
Hi @DevJunz,

And thanks for the proposal.

However, introducing `getFactory(Type)` with an `Assert.state()` check doesn't really buy us anything here. `this.factories` is a private, final `EnumMap` that's fully populated for every `Type` in the constructor and never mutated afterward. So, the assertion can never actually fail: it adds a bit of overhead and indirection without providing any real null-safety benefit.

I'd suggest keeping the `getFactory(Type)` helper like you've proposed, but instead of performing a not-null assertion, just move the class-level suppression down to that one method:

```java
@SuppressWarnings("NullAway") // Well-known map keys
private DateTimeFormatterFactory getFactory(Type type) {
	return this.factories.get(type);
}
```

That achieves the actual goal: no more class-wide suppression masking unrelated issues.

DevJunz:
Thanks for the review, @sbrannen 
I've applied your suggestion and pushed the updated commit. Please take another look when you get a chance.

sbrannen (changes requested):

sbrannen (approved):

sbrannen:

Thanks

PR #37233, opened 2026-09-03
Fix thrown exception setup in ExceptionHandlingWebHandlerTests

noojung:
`handleErrorSignal()` and `thrownExceptionBecomesErrorSignal()` previously had identical bodies, so both exercised the `Mono.error()` path in `StubWebHandler`.

`StubWebHandler` already had a `raise` flag to simulate a synchronously thrown exception, but no test enabled it.

Enable the flag in `thrownExceptionBecomesErrorSignal()` to cover the [catch (Throwable) block](https://github.com/spring-projects/spring-framework/blob/16edf9867a143f95cefadb7b2115dcbbf624e176/spring-web/src/main/java/org/springframework/web/server/handler/ExceptionHandlingWebHandler.java#L79-L81) in `ExceptionHandlingWebHandler`.

### DIFF
--- a/spring-web/src/test/java/org/springframework/web/server/handler/ExceptionHandlingWebHandlerTests.java
+++ b/spring-web/src/test/java/org/springframework/web/server/handler/ExceptionHandlingWebHandlerTests.java
@@ -17,6 +17,7 @@
 package org.springframework.web.server.handler;
 
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


noojung:
Superseded by #37317, which includes this fix along with other duplicated test corrections.

PR #37236, opened 2026-09-04
Resolve Gradle deprecation warnings for Project object as a dependency notation

noojung:
Currently, running the build with `--warning-mode=all` reports the following deprecation warning:

```text
> Configure project :integration-tests
Using a Project object as a dependency notation has been deprecated. This will fail with an error in Gradle 10. Please use the project(String) method on DependencyHandler or the createProjectDependency(String) method on DependencyFactory instead. Consult the upgrading guide for further information: https://docs.gradle.org/9.7.1/userguide/upgrading_version_9.html#dependency_project_notation
```

This PR updates `KotlinConventions` and `RuntimeHintsAgentPlugin` to create explicit project dependencies with `DependencyFactory#createProjectDependency(...)`.

Reference: https://docs.gradle.org/9.7.1/userguide/upgrading_version_9.html#dependency_project_notation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ plugins {
 	// kotlinVersion is managed in gradle.properties
 	id 'org.jetbrains.kotlin.plugin.serialization' version "${kotlinVersion}" apply false
 	id 'org.jetbrains.dokka'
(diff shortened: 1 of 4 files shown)


bclozel:
Thanks @noojung !

PR #37239, opened 2026-09-04
 Release the body of an unconsumed `ReactorClientHttpResponse` once it becomes unreachable

raccoonback:
## Motivation

`ReactorClientHttpConnector.connect()` emits a `ClientHttpResponse` before the body is read.
If the response is dropped without subscribing to `getBody()`, the buffered content stays in the Reactor Netty inbound queue.
The connector releases the body only on `doOnCancel`, but a cancel signal does not always arrive.
For example `Mono.zip` marks a source as cancelled once it has emitted, so a later cancel is not propagated upstream and the emitted response is discarded instead.
Reactor Netty does not release the queued content when the connection is closed either.

We receives many client cancellations (HTTP 499). When a request is cancelled after the upstream response was already emitted but before its body was read, the leak reproduces. In production it appears as:

```
LEAK: ByteBuf.release() was not called before it's garbage-collected.
  Hint: [..., R:upstream:80] Buffered ByteBufHolder in the inbound buffer queue
  reactor.netty.channel.FluxReceive.onInboundNext(FluxReceive.java:397)
```

Reproduce: call `connector.connect(...)` against a local server, then drop the response without reading the body, or cancel a `Mono.zip` after the response was emitted. `PooledByteBufAllocator` metrics show one allocation that stays active.

> The body cannot be released on a timer or on connection close because a late subscriber is still valid at that point.
> It can only be released once nobody can subscribe anymore, that is when the response object is unreachable.

## Key changed

* Register `ReactorClientHttpResponse` with a `java.lang.ref.Cleaner` when the response may have a body.
* When the response becomes unreachable with the body still not subscribed, drain the inbound and release the buffers.
* Deregister as soon as `getBody()` is subscribed, so normal responses add no work for the GC.
* `mayHaveBody()` now uses the request method from the Reactor Netty response, so a `HEAD` redirected to `GET` is handled by the final method.
* Add `ReactorClientHttpResponseReleaseTests` (consumed, cancelled before subscribe, dropped, cancelled after emission, HEAD redirected to GET).

No behavior change for responses whose body is subscribed or cancelled before emission.

### DIFF
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpConnector.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpConnector.java
@@ -48,6 +48,7 @@ import org.springframework.util.Assert;
  * @author Rossen Stoyanchev
  * @author Sebastien Deleuze
  * @author Juergen Hoeller
+ * @author Seungbin Ko
  * @since 5.0
  * @see reactor.netty.http.client.HttpClient
  */
@@ -175,7 +176,7 @@ public class ReactorClientHttpConnector implements ClientHttpConnector, SmartLif
 				.doOnCancel(() -> {
 					ReactorClientHttpResponse response = responseRef.get();
 					if (response != null) {
-						response.releaseAfterCancel(method);
+						response.releaseAfterCancel();
 					}
 				});
 	}
--- a/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpResponse.java
+++ b/spring-web/src/main/java/org/springframework/http/client/reactive/ReactorClientHttpResponse.java
@@ -16,6 +16,7 @@
 
 package org.springframework.http.client.reactive;
 
+import java.lang.ref.Cleaner;
 import java.util.Collection;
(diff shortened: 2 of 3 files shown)


bclozel:
Hello @raccoonback 

I think we need to take a step back here. Can you create a new issue that describes the usage pattern that leads to this problem. Especially that part:

> ReactorClientHttpConnector.connect() emits a ClientHttpResponse before the body is read.
If the response is dropped without subscribing to getBody(), the buffered content stays in the Reactor Netty inbound queue.
The connector releases the body only on doOnCancel, but a cancel signal does not always arrive.
For example Mono.zip marks a source as cancelled once it has emitted, so a later cancel is not propagated upstream and the emitted response is discarded instead.
Reactor Netty does not release the queued content when the connection is closed either.

Can you share code snippets or a sample application we can use? Cancelling the request before the server has a chance to write to the response should help us to reproduce the problem.

I'm closing this PR for now because `java.lang.ref.Cleaner` are fallbacks to try-with-resources structures, and reactive streams has its own mechanism for cleaning up resources. At this stage I'm not sure this is a problem in the application itself or a bug in Spring, but in both case I don't think we'll use this proposal. Cheers.

PR #37248, opened 2026-09-07
Fix missing assertion in DefaultClientResponseTests

noojung:
Add the missing `.isTrue()` to the header assertion in `createError()`.

Also remove obsolete deprecation suppressions from `createError()` and `createException()`.


sbrannen:

Thanks

PR #37254, opened 2026-09-07
Fix and strengthen tests in CollectionToCollectionConverterTests

noojung:
Despite their names, `convertEmptyVector_shouldReturnEmptyArrayList()` and `convertNonEmptyVector_shouldReturnNonEmptyArrayList()` had identical logic:
```java
	@Test
	void convertEmptyVector_shouldReturnEmptyArrayList() {
		Vector<String> vector = new Vector<>();
		vector.add("Element");
		testCollectionConversionToArrayList(vector);
	}

	@Test
	void convertNonEmptyVector_shouldReturnNonEmptyArrayList() {
		Vector<String> vector = new Vector<>();
		vector.add("Element");
		testCollectionConversionToArrayList(vector);
	}
```

As a result, the empty Vector case was never tested.

Update `convertEmptyVector_shouldReturnEmptyArrayList()` to use an empty Vector and assert that the result is an empty ArrayList. In `convertNonEmptyVector_shouldReturnNonEmptyArrayList()`, explicitly assert that the result is a non-empty ArrayList.

Also strengthen `collectionsEmptyList()` to verify that conversion returns the original instance, in addition to checking that it completes without an exception.


sbrannen:
This has been merged into `7.0.x` and `main` in 01a23e32b55666786a3d1155498d9d9e76f82089 and revised in f768641c080a1352e2bb2b5cf64ad1a867769046.

Thanks

PR #37256, opened 2026-09-08
Avoid redundant filtering in FilteredMap.size()

noojung:
`FilteredMap.size()` iterates over `keySet()`, which already applies the predicate through a `FilteredSet` wrapper.
The loop then tests each accepted key a second time.

This PR updates `FilteredMap.size()` to iterate over `delegate.keySet()` instead, avoiding redundant evaluations and the creation of a `FilteredSet` wrapper on every call.


sbrannen (approved):

sbrannen:

Thanks

Issue #26104, opened 2020-11-17
PathMatchingResourcePatternResolver fails to resolve resources from JAR located in path with '#' on JDK 11+


rweisleder:
**Affects:** Spring Framework 5.3.1 on JRE 11+

---

I have a simple JAR which contains the resource `demo/dummy.txt` and a class with this code:
```java
ResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
Resource[] resources = resolver.getResources("classpath*:**/dummy.txt");
System.out.println("resources = " + Arrays.toString(resources));
```

If I put this JAR into `c:\tmp\demo` and run the code, the resource is found.
If I put this JAR into `c:\tmp\#demo` and run the code, no resources are found.  Instead, an exception is logged at debug level:
```
Cannot search for matching files underneath [c:\tmp\#demo\resource-resolver-demo-1.0-SNAPSHOT.jar] because it cannot be converted to a valid 'jar:' URL: no !/ in spec

java.net.MalformedURLException: no !/ in spec
	at java.base/java.net.URL.<init>(URL.java:679)
	at java.base/java.net.URL.<init>(URL.java:541)
	at java.base/java.net.URL.<init>(URL.java:488)
	at org.springframework.core.io.UrlResource.<init>(UrlResource.java:95)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.addClassPathManifestEntries(PathMatchingResourcePatternResolver.java:435)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.addAllClassLoaderJarRoots(PathMatchingResourcePatternResolver.java:400)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.doFindAllClassPathResources(PathMatchingResourcePatternResolver.java:346)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.findAllClassPathResources(PathMatchingResourcePatternResolver.java:321)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.getResources(PathMatchingResourcePatternResolver.java:288)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.findPathMatchingResources(PathMatchingResourcePatternResolver.java:494)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.getResources(PathMatchingResourcePatternResolver.java:284)
	at demo.ResourceResolverDemo.main(ResourceResolverDemo.java:13)
[...]

PR #37281, opened 2026-09-16
Encode percent signs in Class-Path jar URLs

### LINKED ISSUE #37280: `PathMatchingResourcePatternResolver` misses jars on `java.class.path` whose path contains `%`
**Affects:** spring-core 7.0.9 and 6.2.19 (reproduced), code unchanged on `main` as of 2026-09-15. JDK 21, macOS.

`ClassPathManifestEntry.fixPath` escapes `#` (see #26104) but not `%`. `asJarFileResource` builds `jar:file:<path>!/` from the raw file path, and opening that URL decodes the path again. A `%` in a directory name therefore makes the resolver look for a different file:

- `%` followed by two hex digits (`host%3A8099`) is decoded to `host:8099`, so the jar is not found. `doFindPathMatchingJarResources` swallows the `NoSuchFileException`, and the jar is silently missing from `classpath*:` results rooted at the class path root, such as `classpath*:**/*.class`.
- Any other `%` (`100%`) makes `getResources` throw `IllegalArgumentException: Malformed escape pair`.

Both happen only for jar roots from `addClassPathManifestEntries`, i.e. jars on the `java.class.path` of the system class loader. That is how IDEs and build tools usually launch an application.

**How we hit it:** Coursier, the resolver used by sbt and Mill, percent-encodes the host in its cache path. Jars from a repository with a port (`http://localhost:8099/repository/…`) end up in `~/Library/Caches/Coursier/v1/http/localhost%3A8099/…`. Vaadin's dev mode scans the whole class path. Its `FilterableResourceResolver` extends `PathMatchingResourcePatternResolver` and opens the same jar root URL without catching, so the application fails at startup:

```
Caused by: java.nio.file.NoSuchFileException: /Users/…/Library/Caches/Coursier/v1/http/localhost:8099/repository/…/togglebutton-3.0.0.jar
	at java.base/sun.net.www.protocol.jar.JarURLConnection.getJarFile(JarURLConnection.java:71)
	at com.vaadin.flow.spring.io.FilterableResourceResolver.readPackageProperties(FilterableResourceResolver.java:417)
	at com.vaadin.flow.spring.io.FilterableResourceResolver.cachePackageProperties(FilterableResourceResolver.java:329)
	at com.vaadin.flow.spring.io.FilterableResourceResolver.doFindPathMatchingJarResources(FilterableResourceResolver.java:207)
	at org.springframework.core.io.support.PathMatchingResourcePatternResolver.findPathMatchingResources(PathMatchingResourcePatternResolver.java:779)
```

**Reproducer** (bash or zsh; `SPRING_CORE` and `COMMONS_LOGGING` point to `spring-core-7.0.9.jar` and `commons-logging-1.3.5.jar`):

```bash
mkdir -p repro/content/demo 'repro/lib/plain' 'repro/lib/host%3A8099' 'repro/lib/100%' && cd repro
echo hello > content/demo/hello.txt
jar cf lib/plain/demo.jar -C content demo
cp lib/plain/demo.jar 'lib/host%3A8099/demo.jar'
cp lib/plain/demo.jar 'lib/100%/demo.jar'
cat > Repro.java <<'EOF'
import org.springframework.core.io.Resource;
import org.springframework.core.io.support.PathMatchingResourcePatternResolver;

public class Repro {
        Resource[] resources = new PathMatchingResourcePatternResolver().getResources("classpath*:**/hello.txt");
        System.out.println(resources.length + " found");
        for (Resource resource : resources) {
            System.out.println(resource.getURL());
        }
    }
}
EOF
javac -cp "${SPRING_CORE}:${COMMONS_LOGGING}" -d classes Repro.java
for jar in lib/plain/demo.jar 'lib/host%3A8099/demo.jar' 'lib/100%/demo.jar'; do
  echo "== ${jar}"
  java -cp "classes:${SPRING_CORE}:${COMMONS_LOGGING}:${jar}" Repro 2>&1 | head -3
done
```

Output:

```
== lib/plain/demo.jar
1 found
jar:file:/…/repro/lib/plain/demo.jar!/demo/hello.txt
== lib/host%3A8099/demo.jar
0 found
== lib/100%/demo.jar
Exception in thread "main" java.lang.IllegalArgumentException: Malformed escape pair: /…/repro/lib/100%/demo.jar
	at java.base/sun.net.www.ParseUtil.decode(ParseUtil.java:207)
	at java.base/sun.net.www.protocol.file.Handler.openConnection(Handler.java:69)
```

The `IllegalArgumentException` leaves Spring through `doFindPathMatchingJarResources(PathMatchingResourcePatternResolver.java:903)` ← `findPathMatchingResources(:779)` ← `getResources(:375)`.

Expected: `1 found` for all three jars.

**Suggested fix:** escape `%` before `#` in `ClassPathManifestEntry.fixPath`:

```java
path = StringUtils.replace(path, "%", "%25");
// Since '#' can appear in directories/filenames, java.net.URL should not treat it as a fragment
return StringUtils.replace(path, "#", "%23");
```

With this change applied to 7.0.9, the reproducer finds the resource in all three jars, for example `jar:file:/…/repro/lib/host%253A8099/demo.jar!/demo/hello.txt`. That is the same URL the JDK returns from `ClassLoader.getResource("demo/hello.txt")` for that jar.

kalayciburak:
## Motivation

`ClassPathManifestEntry.fixPath` already encodes `#` so `java.net.URL` does not treat it as a fragment (gh-26104). It does not encode `%`, so a jar on `java.class.path` whose directory name contains `%` is either skipped (`host%3A8099` is decoded to `host:8099`) or causes `IllegalArgumentException: Malformed escape pair` (`100%`).

That path shape appears in practice with Coursier caches (`~/Library/Caches/Coursier/v1/http/localhost%3A8099/...`).

Closes gh-37280

## Change

Encode `%` to `%25` before encoding `#` in `fixPath`, matching the JDK `ClassLoader.getResource` URL form.

## Tests

- `PathMatchingResourcePatternResolverTests$ClassPathManifestEntries.javaDashJarFindsClassPathManifestEntriesWhenPathContainsPercentHex`
- `PathMatchingResourcePatternResolverTests$ClassPathManifestEntries.javaDashJarFindsClassPathManifestEntriesWhenPathContainsLonePercent`
- full `PathMatchingResourcePatternResolverTests`
- `:spring-core:checkstyleMain` and `:spring-core:checkstyleTest`


jhoeller:
This is being worken on separately, under the original #37280, aiming to resolve the issue in a more defensive fashion.

In general, we recommend to wait a bit before implementing a solution - until a core maintainer triaged the issue. In particular for core framework issues, it is rather unlikely that a PR for an untriaged issue will be accepted as-is.

PR #37283, opened 2026-09-16
Make message decoding in ProtobufDecoder more extensible

### LINKED ISSUE #37265: Make message decoding in ProtobufDecoder more extensible
Recent changes in `ProtobufDecoder` have made it more open to extension. One thing that is missing still is the ability to override the `Message` creation step. The base class does all the interesting stuff, and comes up with a byte buffer with the message contents, but then doesn't allow you to insert your own logic for decoding. A protected method in `ProtobufDecoder` would be great, or a helper interface like `MessageSizeReader`.

Would enable us to more easily (without copy-pasting the whole of `ProtobufDecoder`) support grpc-web in https://github.com/spring-projects/spring-grpc/issues/427.

AbhishekLaddha54:
Fixes gh-37265

### Summary
`ProtobufDecoder` recently gained extensibility for size reading via `MessageSizeReader` (7.0), but message creation is still hard-coded in a private inner class. This makes gRPC-Web support impossible without copy-paste.

gRPC-Web uses:
- 5-byte header: 1 byte flag (0x80 = trailer, 0x01 = compressed) + 4 bytes BE length
- Trailer frames that must be skipped

### Changes
- Add `protected ExtensionRegistry getExtensionRegistry()`
- Add `protected @Nullable Message decodeMessage(CodedInputStream, ResolvableType)` and overload for `Class<?>` — returning null skips frame
- Refactor `decode(DataBuffer, ...)` and `MessageDecoderFunction` to use `decodeMessage()`
- Make `MessageDecoderFunction` protected with getters and add `createMessageDecoderFunction()` factory

This mirrors the existing `initMessageSizeReader()` pattern.

### Enables (example)
```java
class GrpcWebDecoder extends ProtobufDecoder {
  @Override protected MessageSizeReader initMessageSizeReader() {
    return new GrpcWebReader(); // reads 5-byte header
  }
  @Override protected Message decodeMessage(CodedInputStream s, ResolvableType t) throws Exception {
    if (isTrailer) return null;
    return super.decodeMessage(s, t);
  }
}
```
See spring-projects/spring-grpc#427

### Testing
./gradlew :spring-web:test --tests "*ProtobufDecoderTests*" → BUILD SUCCESSFUL, 0 failures

Existing tests unchanged, new hook returns same as old logic by default
text


bclozel:
Thanks for the proposal but we haven't decidec yet how to triage or handle this issue so I'll decline this for now.

PR #37285, opened 2026-09-16
Enforce disk usage limit when spilling part to disk

### LINKED ISSUE #35099: `max-disk-usage-per-part` allows upload of bigger files
I would like to restrict the size of the file during multipart uploads. WebFlux provides configuration options:
```properties
spring.webflux.multipart.max-in-memory-size=1KB
spring.webflux.multipart.max-disk-usage-per-part=1KB
```

I would expect the same behaviour as with `spring.servlet.multipart.max-file-size` in WebMVC. However, WebFlux still allows the uploads of data greater than 1 KB. In some cases the size of file can reach up to 4KB with this setup, but I haven't found a stable reproducer for it yet. For small sizes still bigger than 1KB it is reproducible most of the times.

<details>
<summary>Example</summary>
Controller:

```java

  @RestController
  public class UploadController {

    @PostMapping
    public Mono<Long> upload(@RequestPart("file") Mono<FilePart> file) {
        return file.flatMapMany(Part::content)
                .map(DataBuffer::readableByteCount)
                .reduce(0L, Long::sum);
    }
  }

```

Invocation:
```bash
# 1030b is a file of size 1030b
curl -XPOST -F"file=@1030b" localhost:8080  
```

</details>

It looks like `DefaultDataBuffer` does not consider the properties during allocation. Thus it can fit the whole request after a single read from the underlying stream, and in this case the limit will never be checked.

harrisleesh:
Previously, `PartGenerator` only enforced `maxDiskUsagePerPart` for body buffers that arrived after a part had switched to file storage (`IdleFileState#body`). The content accumulated in memory that triggered the switch was carried into `CreateFileState` and written to disk without any check against `maxDiskUsagePerPart`. As a result, a part whose last body buffer caused the in-memory overflow was accepted in full, even when its total size exceeded the configured disk usage limit — matching the reproduction in gh-35099 (`max-in-memory-size=1KB`, `max-disk-usage-per-part=1KB`, 1030-byte upload accepted).

This PR checks the accumulated byte count against `maxDiskUsagePerPart` in `InMemoryState#switchToFile` before the temp file is created, and emits a `DataBufferLimitException` with the same message as the existing check for subsequent buffers. Buffers held by the state are released before emitting the error, keeping the leak checks in `DefaultPartHttpMessageReaderTests` green.

The added regression test drives `files.multipart` (100-byte parts, 64-byte buffers) with `maxInMemorySize=90` / `maxDiskUsagePerPart=99`, so the spill-over itself exceeds the disk limit; it fails on `main` with `onNext(DefaultFilePart)` instead of the expected error.

Closes gh-35099


bclozel:
Thanks @harrisleesh !

PR #37286, opened 2026-09-17
Document weak consistency of `ConcurrentLruCache.clear()`

guanchengang:
The Javadoc previously stated that clear() would "immediately remove all entries from this cache". This is accurate for single-threaded usage but does not hold under concurrent modification:

- The implementation iterates the eviction queue, not the backing ConcurrentHashMap. Entries whose AddTask is still pending in the write queue (i.e. a concurrent put whose value is already visible via get() but has not yet been linked into the eviction queue) are not removed.
- Concurrent puts are neither blocked nor synchronized with, so entries added during clear() may survive the call.

This behavior is intentional: clear() is a weak-consistency operation, consistent with the cache's overall design as a high-concurrency, eventually-consistent data structure. The fix updates the Javadoc to describe this accurately.


spring-core/src/main/java/org/springframework/util/ConcurrentLruCache.java
```diff
@@ -179,7 +179,11 @@ public int size() {
 	}
 
 	/**
-	 * Immediately remove all entries from this cache.
+	 * Remove all entries currently tracked by the eviction queue.
+	 * <p>Concurrent writes whose {@code AddTask} is still pending in
+	 * the write queue may survive this call, they are treated as if
+	 * they arrived after the {@code clear} operation. This method does
+	 * not block or synchronize with in-flight {@code put} operations.
```
sbrannen:
```suggestion
 * Remove all entries currently tracked by the eviction queue.
 * <p>This method does not block or synchronize with in-flight
 * {@link #put(Object) put} operations. A concurrent {@code put} whose
 * entry has not yet been linked into the eviction queue at the time
 * this method runs may not be removed, and can therefore survive
 * this call as if it had occurred immediately afterward.
```

sbrannen (changes requested):
Let's document this behavior in terms of the public API instead of the private implementation (i.e, don't mention `AddTask`).

Also, please rebase on `7.0.x`.

Thanks

spring-core/src/main/java/org/springframework/util/ConcurrentLruCache.java
```diff
@@ -179,7 +179,12 @@ public int size() {
 	}
 
 	/**
-	 * Immediately remove all entries from this cache.
+ * Remove all entries currently tracked by the eviction queue.
+ * <p>This method does not block or synchronize with in-flight
+ * {@link #put(Object) put} operations. A concurrent {@code put} whose
+ * entry has not yet been linked into the eviction queue at the time
+ * this method runs may not be removed, and can therefore survive
+ * this call as if it had occurred immediately afterward.
```
sbrannen, line 187: Oops. The leading tabs are missing now.

guanchengang: I'm sorry, I will force this branch again when I have free time later.

sbrannen: No worries. My "suggestion" didn't contain the necessary tabs, so I take the blame for that. 

Before you force push, please rebase on 7.0.x as well.

Thanks

sbrannen:
While reviewing the wording here, I dug a bit deeper into `clear()` and found that the underlying behavior it describes is actually a fixable ordering issue, not just an inherent consequence of weak consistency: `clear()` drains its pending write queue *after* removing entries, so a write that was already queued before `clear()` even started ends up being re-applied and survives the call. I've filed #37287 to track that.

Since fixing #37287 narrows exactly the behavior this PR's Javadoc describes (an already-pending write will no longer survive `clear()` -- only a write that's genuinely concurrent with an in-progress `clear()` call can), the final wording in the Javadoc depends on the outcome of that fix. So, rather than merge this PR's Javadoc now and immediately revise it again once I've completed #37287, I'm going to close this PR and create a combined/updated version of the documentation in #37287, so that it accurately reflects the fixed behavior in one step.

I'll make sure to credit you in the commit that resolves #37287.

Thanks again for catching this and taking the time to open the PR. It led to a real fix, not just a doc clarification! 👍

Issue #37287, opened 2026-09-17
`ConcurrentLruCache.clear()` can miss entries added while it's running


sbrannen:
## Overview

While reviewing #37286, I noticed an issue with `clear()` that's a bit more than plain _eventual consistency_.

Specifically, `clear()` holds the eviction lock for its whole duration, so a `put()` happening at the same time can't finish its own bookkeeping and ends up queued as a pending write operation. After `clear()` finishes removing everything it can currently see, it processes that leftover queue itself (via `this.writeOperations.drainAll()`), which completes the pending "add" that should have been taken care of by `clear()`.

The practical effect is that a `put()` which completes _while `clear()` is running_ (i.e, its entry is already visible via `get()`/`contains()`) can still be present in the cache right after `clear()` returns.

It seems like we can fix this by having `clear()` process that leftover queue _before_ removing entries rather than after, without reintroducing any blocking between `put()` and `clear()`.

## Related Issues
 
- #37286

guanchengang:
Hi @sbrannen 
Perhaps we could consider changing it to something like:
```java
public void clear() {
		this.evictionLock.lock();
		try {
			Iterator<Map.Entry<K, Node<K, V>>> it = this.cache.entrySet().iterator();
			while (it.hasNext()) {
				Node<K, V> node = it.next().getValue();
				markAsRemoved(node);
				this.evictionQueue.remove(node);
				it.remove();
			}
			this.readOperations.clear();
			this.writeOperations.drainAll();
		}
		finally {
			this.evictionLock.unlock();
		}
	}
```
This iterates directly from the cache, avoiding unnecessary evictionQueue operations for already-removed nodes. Since `AddTask`  fail silently after a node is marked removed, this should reduce no-op work and improve performance.

sbrannen:
@guanchengang,

What about `this.cache.remove(node.key, node)` instead of `it.remove()` to preserve current runtime semantics?

guanchengang:
Yes, I think that’s better. Let’s go with `this.cache.remove(node.key, node)`.

PR #37292, opened 2026-09-18
Avoid useless queue ops in `ConcurrentLruCache.clear()`

guanchengang:
`ConcurrentLruCache.clear()` previously drained write operations before cleaning up the cache. This could re-enqueue nodes that `clear()` was about to remove, causing useless evictionQueue operations. Now `clear()` iterates the cache values directly, removes and marks nodes as removed first, and drains write operations afterward. Since AddTask fails silently after a node is marked removed, this avoids the no-op work and improves performance.

See gh-37287


sbrannen:

Thanks

PR #37293, opened 2026-09-18
Remove unchecked cast in DispatcherServlet

cookie-meringue:
## Summary

The private attribute restoration method declares its snapshot parameter as `Map<?, ?>`, although its only caller passes a `Map<String, Object>`. Match the parameter type to the caller and remove the unnecessary unchecked cast and warning suppression.


bclozel:
Thanks @cookie-meringue

PR #37295, opened 2026-09-18
Use "isEmpty" to check if a collection is empty

quaff:


quaff:
Here is the flaky test:
```
  > Task :spring-core:java21Test
  
  RetryTemplateTests > TimeoutTests > retryableWithTimeoutExceededAfterSecondRetry() FAILED
      java.lang.AssertionError at RetryTemplateTests.java:802
```

bclozel:
Thanks @quaff !

Issue #35164, opened 2025-07-07
Add SSE Support to RestClient


asaikali:
In Spring WebClient is the only supported way to consume SSE streams. However, using WebClient introduce a reactive streams increasing application complexity.

Ideally RestClient can support the consumption of SSE streams allowing developers to keep codebases simple.

Spring offers reactive and non-reactive facilities for producing SSE events. Adding the ability to consume SSE streams in a reactive and non-reactive ways bring the server side and client side experience to parity.

tzolov:
I believe this is related: https://github.com/spring-projects/spring-framework/issues/34290

rstoyanchev:
Historically `RestTemplate` exposed low level streaming via `InputStream` and `OutputStream`. For `WebClient`, with Reactive Streams (error/completion/cancel/demand signals) and a functional, `java.util.stream`-like API, it was natural natural to provide higher level streaming. `RestClient` picks up from where `RestTemplate` was.

How to consume an SSE stream is a key question. Historically, an API of this kind is callback-based to avoid blocking. In the JDK, the `HttpClient` has `HttpResponse.BodyHandler` (essentially Reactive Streams `Flow.Subscriber`) including one that splits the response by line. We need the Loom friendly equivalent.

In modern Java with Loom, it's natural to use a pull-style API that blocks, and integrates well with Structured Concurrency for responsiveness to cancellation. A basic version with rendezvous hand-off is straight-forward enough, but a more advanced version that supports prefetching raises plenty of questions. An even more advanced, functional API, is even more ambitious.

There is currently nothing in the JDK for this, and it's not going to come soon (e.g. see Structured Concurrency [JEP 505](https://openjdk.org/jeps/505, which lists "sharing streams of data among threads" as a non-goal). We need to explore alternatives in the mean time.

pro-peller:
I am very upset that there is no SSE support in SpringBoot, my current projects do not allow updating SpringBoot 4, so the upgrade process has stopped. Please add your work with SSE.

bclozel:
@pro-peller I don't understand the link with Spring Boot 4. Why can't you upgrade to Spring Boot 4 and why is that about SSE support in RestClient?

axma331:
Spring Boot 4 lacks SSE support, blocking our upgrade path. Our applications depend on Server-Sent Events, so we can't migrate until this feature is available. Please add SSE support to Spring Boot 4

bclozel:
[...]

PR #37301, opened 2026-09-19
Add SSE support to RestClient

ABin-Huang:
## Summary

This adds blocking (non-reactive) consumption of Server-Sent Events (SSE) to `RestClient`, addressing #35164.

Today, `WebClient` is the only supported way to consume an SSE stream, which forces applications that otherwise use the synchronous `RestClient` to pull in the entire reactive stack. This change lets non-reactive code consume `text/event-stream` directly.

## Proposal

A new method on `RestClient.ResponseSpec`:

```java
<T> void bodyToServerSentEvents(Class<T> dataType, Consumer<ServerSentEvent<T>> consumer);
```

It reads the response incrementally and invokes the consumer for each decoded `ServerSentEvent`, blocking until the stream is exhausted. The `data` payload is decoded to `dataType` via the configured `HttpMessageConverter`s (String built-in; JSON via the existing converters).

### Usage

```java
List<TradeEvent> events = new CopyOnWriteArrayList<>();

restClient.get()
    .uri("/events")
    .accept(MediaType.TEXT_EVENT_STREAM)
    .retrieve()
    .bodyToServerSentEvents(TradeEvent.class, events::add);
```

## Implementation

- `RestClient.ResponseSpec#bodyToServerSentEvents(Class, Consumer)` — new API.
- `DefaultRestClient.DefaultResponseSpec` — runs the request, applies status handlers (4xx/5xx), and delegates to the new reader.
- New package-private `SseEventReader` — a blocking counterpart to `ServerSentEventHttpMessageReader`. It parses the stream line-by-line (blank-line delimited event blocks), mirrors the existing `buildEvent` logic for `data`/`id`/`event`/`retry`/`comment`, and decodes each event's `data` through the configured message converters.

## Tests

Added two parameterized integration tests in `RestClientIntegrationTests` (run across all five `ClientHttpRequestFactory` implementations):
- `serverSentEventsAsString` — verifies comment, multi-line `data`, `id`, `event`, `retry`, and the `Accept: text/event-stream` header.
- `serverSentEventsAsPojo` — verifies JSON `data` decoding into a POJO.

All 10 cases pass, and `checkstyleMain`/`checkstyleTest` are clean.

## Notes

- This is the blocking/callback variant, which directly addresses the issue's goal of keeping codebases simple without reactive streams. A reactive `Flux`-returning variant could be layered on top later if desired.
- Marking as draft initially for feedback on the API shape.


bclozel:
Thanks for the proposal, but we are not going to accept this contribution.

Besides the parsing security issues, the main problem here is the programming model. Accumulating elements in a list with a callback style interface does not fit with a modern programming model and won't work with modern infrastructure like structured concurrency.

The Spring team is experimenting right now and we'll update the issue when we have results that we can share. For the time being I've assigned the issue and reschedule it since it's not likely to make our 7.1 release.

PR #37304, opened 2026-09-19
Fix `@Nullable` placement in ConcurrentReferenceHashMap

ngocnhan-tran1996:
Apply `@Nullable` to the type argument instead of the wildcard.


ngocnhan-tran1996:
I misunderstood the nullability annotation here.

PR #37306, opened 2026-09-19
Fix RestClient I/O exception handling during response extraction

### LINKED ISSUE #37078: RestClient readWithMessageConverters method hiding I/O Exceptions with incorrect error message.

dhanush-rao-k:
Closes gh-37078.

### Summary

`DefaultRestClient.readWithMessageConverters(...)` currently wraps I/O exceptions encountered during response extraction in a generic `RestClientException` with a misleading content-type extraction message.

This change:

- propagates I/O failures as `ResourceAccessException`
- preserves the original `IOException` as the cause
- keeps the existing `RestClientException` behavior for non-I/O failures
- adds a regression test covering a `SocketTimeoutException` while reading the response body

### Testing

- `./gradlew :spring-web:test --tests org.springframework.web.client.DefaultRestClientTests`


bclozel:
https://github.com/spring-projects/spring-framework/issues/37078#issuecomment-5386635571

PR #37307, opened 2026-09-20
Improve JdbcClient to align BatchSpec with StatementSpec

quaff:
Add missing methods:
1. `param(int jdbcIndex, @Nullable Object value)`
2. `param(int jdbcIndex, @Nullable Object value, int sqlType)`
3. `param(String name, @Nullable Object value, int sqlType)`


spring-jdbc/src/main/java/org/springframework/jdbc/core/simple/DefaultJdbcClient.java
```diff
@@ -355,12 +355,36 @@ public BatchSpec param(@Nullable Object value) {
...
 
+			@Override
+			public BatchSpec param(int jdbcIndex, @Nullable Object value) {
+				Assert.state(jdbcIndex > 0, "jdbcIndex must greater than 0");
+				int index = jdbcIndex - 1; // one-based -> zero-based
+				if (index >= this.currentIndexedParams.size()) {
+					// padding
+					for (int i = 0; i <= jdbcIndex - this.currentIndexedParams.size(); i++) {
+						this.currentIndexedParams.add(null);
+					}
```
sbrannen: Iterating based on the current `size()` and adding to the list within the loop will lead to errors.

Perhaps you can take inspiration from `DefaultStatementSpec.param(int, Object)`'s handling of `indexedParams`?

sbrannen (changes requested):
Thanks for the suggestions and the PR.

I've requested a small fix.

spring-jdbc/src/main/java/org/springframework/jdbc/core/simple/DefaultJdbcClient.java
```diff
@@ -161,25 +161,29 @@ public StatementSpec param(@Nullable Object value) {
 
 		@Override
 		public StatementSpec param(int jdbcIndex, @Nullable Object value) {
+			addIndexedParam(this.indexedParams, jdbcIndex, value);
+			return this;
+		}
+
+		private static void addIndexedParam(List<@Nullable Object> indexedParams, int jdbcIndex, @Nullable Object value) {
```
quaff: @sbrannen I make this util method as static to avoid misusing `this.indexedParams`, is it recommended?

sbrannen: Yes, making it static is a good call, but please move `addIndexedParam()` and `validateIndexedParamValue()` below `statementCreatorForIndexedParamsWithKeys()`.

sbrannen (changes requested):

sbrannen (changes requested):
The new tests in `JdbcClientIntegrationTests` use `add()` more than once, resulting in a multi-statement batch update. However, the new unit tests in `JdbcClientIndexedParameterTests` and `JdbcClientNamedParameterTests` only have a single `add()`, resulting in a single-statement batch.

Please introduce additional unit tests analogous to the current new ones but with multiple statements (e.g., 2) per batch update, and verify `setObject(...)` is invoked with the right values for both rows in order, plus `containsExactly(...)`.

quaff:
> The new tests in `JdbcClientIntegrationTests` use `add()` more than once, resulting in a multi-statement batch update. However, the new unit tests in `JdbcClientIndexedParameterTests` and `JdbcClientNamedParameterTests` only have a single `add()`, resulting in a single-statement batch.
> 
> Please introduce additional unit tests analogous to the current new ones but with multiple statements (e.g., 2) per batch update, and verify `setObject(...)` is invoked with the right values for both rows in order, plus `containsExactly(...)`.

Updated tests with `@ParameterizedTest`.

PR #37308, opened 2026-09-20
Set first batch arg once while executing batch with generated keys

quaff:
`batchArgs[0]` is already set by `pscf::newPreparedStatementCreator`, it is unnecessary to set it again by `BatchPreparedStatementSetter::setValues`.


quaff:
Should I create a separated PR for second commit or keep it as is? it's related to new `JdbcClient` feature since 7.1.

sbrannen:
> Should I create a separated PR for second commit or keep it as is? it's related to new `JdbcClient` feature since 7.1.

Since we typically squash commits in a PR before merging, it would be better to pre-squash them into a single commit with an all-encompassing title such as  "Set first batch arg once while executing batch with generated keys". The commit message could then be a combination of the current two commit messages.

I'll go ahead and update the current PR title to reflect that.

quaff:
Should the first commit be backported to 6.x?

sbrannen:
> Should the first commit be backported to 6.x?

6.x is no longer under open source support, and this is more of an _enhancement_ than a _bug fix_, so we don't plan to backport this.

sbrannen:

Thanks

